PR #3152, opened 2020-09-20
wasm2js: Support exported tables

sbc100:
Depends on https://github.com/emscripten-core/emscripten/pull/12291

PR #3163, opened 2020-09-23
Remove unnecessary "new" on Table generation in wasm2js

kripken:
Using `new Table()` is not what we want, since we get an input and want to
modify and return it.

(View one of the modified testcases's full file, looking for `Table`, to see
the issue.)


kripken:
Sadly this is not enough, minification causes this:
```javascript
     ret.get = function(i) {
      return this[kb];
     };
```
Note the `kb`. I think this `Table` function needs to not go in the area where we minify things at the start.

sbc100:
So what tests to fail the?  I certainly ran all the wasm2js tests in binaryen before landing this change?

How is it that this only broke in the binaryen+emcc test?  It worked fine in the binaryen+clang tests of wasm2js and in all the emscripten-side tests.

sbc100:
I suggest we just disable the binaryen.js tests until we find a fix.

kripken:
Hmm, I think we should revert https://github.com/WebAssembly/binaryen/pull/3152 - it reached the `tot`, and it's just broken for wasm2js. We'll have broken builds on emscripten CI soon.

kripken:
I think the problem is binaryen CI doesn't test actually running wasm2js outputs. So the problem was missed there.

The Chromium CI doesn't run wasm2js, so it was missed there too.

kripken:
But actually we should run at least `wasm2.test_wasm2js` on chromium CI, so I may be wrong...

sbc100:
But I've already landed the emscripten change that depends on #3152... so we would need to revert the emscripten-side change first.

Also the combined change passes all the wasm2js tests on the emscripten side, which to run the output right?  

What tests fail exacaly?  Is it just binaryen.js?

kripken:
> Also the combined change passes all the wasm2js tests on the emscripten side, which to run the output right?

Have we run on emscripten CI since the binaryen roll?

kripken:
It does look like `wasm2.test_wasm2js` passes, while the binaryen build fails. I suspect the minification issue only comes up in big enough projects.

sbc100:
> > Also the combined change passes all the wasm2js tests on the emscripten side, which to run the output right?
> 
> Have we run on emscripten CI since the binaryen roll?

I'm pretty sure we have... plus I can run the entire `runner.py wasm2js2` suite locally with no failures.

sbc100:
> > > Also the combined change passes all the wasm2js tests on the emscripten side, which to run the output right?
> > 
> > 
> > Have we run on emscripten CI since the binaryen roll?
> 
> I'm pretty sure we have... plus I can run the entire `runner.py wasm2js2` suite locally with no failures.

Do you mean emscripten circle-ci?  Yes for sure since otherwise I couldn't have landed my change that depends on this.

kripken:
I see.

Ok, maybe this really just hits binaryen.js then. I am having a hard time narrowing it down. Maybe I'm just missing something?

kripken:
One necessary thing is to actually call the `.get()` method. That only happens from an invoke. So this only happens on projects with exceptions, really - like binaryen.js.

kripken:
In any case we should land this and maybe disable the emscripten tests here for now, if we can't easily revert. But if we can't fix the minification issue quickly, that may get worrying.

kripken:
Ah I was wrong, this is not the main issue. `new` does still respect the `return` of `ret`. So we can remove the `new` as a code size improvement in this PR, but it's not going to fix anything. The only problem is the minification issue.

sbc100 (approved):

sbc100:
Yeah we should land this anyway.. although it doesn't have to be right now.  Go ahead and land if you like.

PR #3164, opened 2020-09-24
When looking for node binary, search for 'node' first

sbc100:
`node` is the name used by the upstream project.  `nodejs` is a legacy
name used on older debian/ubunru systems.

Searching for `nodejs` first meant it was finding my local (old)
`nodejs` package even those I have a more recent `node` in my $PATH.

### DIFF
--- a/scripts/test/shared.py
+++ b/scripts/test/shared.py
@@ -191,7 +191,7 @@ NATIVECC = (os.environ.get('CC') or which('mingw32-gcc') or
             which('gcc') or which('clang'))
 NATIVEXX = (os.environ.get('CXX') or which('mingw32-g++') or
             which('g++') or which('clang++'))
-NODEJS = os.getenv('NODE', which('nodejs') or which('node'))
+NODEJS = os.getenv('NODE', which('node') or which('nodejs'))
 MOZJS = which('mozjs') or which('spidermonkey')
 V8 = which('v8') or which('d8')
 


scripts/test/shared.py
```diff
@@ -182,7 +182,7 @@ def is_exe(fpath):
             which('gcc') or which('clang'))
 NATIVEXX = (os.environ.get('CXX') or which('mingw32-g++') or
             which('g++') or which('clang++'))
-NODEJS = os.getenv('NODE', which('nodejs') or which('node'))
+NODEJS = os.getenv('NODE', which('node'))
```
kripken, line 185: maybe `node or nodejs` (flip the order)? that would still handle old/odd distros where there is just `nodejs`.

sbc100: I thought about that.

Such users probably don't have a new  enough node anyway to run the binaryen tests.  

Remember this is only for running binaryen tests.  

Also, Debian and Ubuntu systems have /usr/bin/node being the nodejs binary anyway when nodejs is installed.

Also we have the "export NODE=xxx" method to override this is anyone really wants to.

Also, we have not such logic in emscripten .. we only look for `node` there, and emscripten is targeting a much wider group that the binaryen test suite.  If just looking for `node` is good enough there I think its good enough here.

Also, the original point of `nodejs` symlink I believe we because there was another package with a binary called node and this was a way to be sure you were getting nodejs over the other thing... and if we look for `node` first subvert the the entire point of that and risk getting the other thing called `node` which is nothing to do with JS. 

So, I think its cleaner all around.

kripken: Ok, sgtm if you are sure this is just an issue in older distros. I worry maybe odd new ones do the same, but I don't have evidence of that.

kripken (approved):

scripts/test/shared.py
sbc100: Actually I was wrong.. we do have logic in emscripten to look for nodejs..  I will hold off on this for a second.

PR #3167, opened 2020-09-24
wasm2js: Skip heap creation in the absence of wasm memory. NFC

sbc100:
Also, format the asmFunc call to make it more readable in the ES6
modules case.

### DIFF
--- a/src/wasm2js.h
+++ b/src/wasm2js.h
@@ -272,7 +272,7 @@ private:
   // on operations.
   std::unordered_set<Name> functionsCallableFromOutside;
 
-  void addBasics(Ref ast);
+  void addBasics(Ref ast, Module* wasm);
   void addFunctionImport(Ref ast, Function* import);
   void addGlobalImport(Ref ast, Global* import);
   void addTable(Ref ast, Module* wasm);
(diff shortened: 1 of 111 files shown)


sbc100:
Previously we would create an 1 page wasm memory wasting 64k of memory, but I can't see any reason to do this.

src/wasm2js.h
```diff
@@ -2419,41 +2423,40 @@ void Wasm2JSGlue::emitPostES6() {
   //
   // Note that the translation here expects that the lower values of this memory
   // can be used for conversions, so make sure there's at least one page.
-  {
-    auto pages = wasm.memory.initial == 0 ? 1 : wasm.memory.initial.addr;
```
kripken, line 2423: I suspect this was because creating an ArrayBuffer (or a typed array or whatever we create here) with a size of 0 is not valid. It would be good to verify this works, as I'm not sure we have a test for that specifically.

kripken (approved):
In general lgtm but for the size 0 issue.

src/wasm2js.h
kripken: Testing this on node, both old and new, it seems ok... so maybe the worry was irrelevant?

sbc100: Seems to work in chrome too.

PR #3169, opened 2020-09-24
GC: Fuzzing support for i31, assuming i31.new is constant

dcodeIO:
Integrates `i31ref` types and instructions into the fuzzer, by assuming that `(i31.new (i32.const N))` is constant [[1]](https://github.com/WebAssembly/binaryen/pull/3154#issuecomment-698478851) [[2]](https://github.com/WebAssembly/gc/issues/141#issuecomment-698296908). Might make sense to delay landing this until we know more.


dcodeIO:
```
Invocations so far:
   FuzzExec: 12024
   CompareVMs: 3277
   CheckDeterminism: 1030
   Wasm2JS: 2870
   Asyncify: 3052

ITERATION: 14108
```

Going to do more fuzzing tomorrow.

src/tools/fuzzing.h
```diff
@@ -2652,6 +2660,20 @@ class TranslateToFuzzReader {
...
+    assert(wasm.features.hasReferenceTypes() && wasm.features.hasGC());
+    auto* value = make(Type::i32);
+    return builder.makeI31New(value);
+  }
+
+  Expression* makeI31Get(Type type) {
+    assert(type == Type::i32);
+    assert(wasm.features.hasReferenceTypes() && wasm.features.hasGC());
+    auto* i31 = make(Type::i31ref);
+    return builder.makeI31Get(i31, oneIn(2));
```
tlively, line 2674: This might make it clearer that this is choosing a random bool.

```suggestion
    return builder.makeI31Get(i31, bool(oneIn(2)));
```

src/wasm/wasm.cpp
```diff
@@ -217,15 +217,8 @@ const char* getExpressionName(Expression* curr) {
...
-    return c->value;
-  } else if (auto* n = curr->dynCast<RefNull>()) {
-    return Literal::makeNull(n->type);
-  } else if (auto* r = curr->dynCast<RefFunc>()) {
-    return Literal::makeFunc(r->func);
-  } else {
-    WASM_UNREACHABLE("Not a constant expression");
-  }
+  assert(Properties::isConstantExpression(curr));
+  return Properties::getSingleLiteral(curr);
```
tlively, line 221: Huh, it doesn't seem useful to have two functions that do the same thing. It would be good to add a TODO here to remove one of them.

tlively (approved):

src/wasm/wasm.cpp
dcodeIO: This function's body was a duplicate of `Properties::getSingleLiteral(curr)`, yeah, but I added an additional assert here to make sure that it is only used where `Properties::isConstantExpression(curr)` is true, in case `Properties::getSingleLiteral` ever diverges to support more than what's valid in global initializers. Might well be that it never does, and is actually intended to do the exact same thing anyway?

tlively:
Thanks! LGTM

dcodeIO:
```
Invocations so far:
   FuzzExec: 20829
   CompareVMs: 5655
   CheckDeterminism: 1773
   Wasm2JS: 4884
   Asyncify: 5214

ITERATION: 24375
```

PR #3172, opened 2020-09-25
Allow atomics to validate with unshared memory

tlively:
This relaxation has made it to Chrome stable, so it makes sense that we would
allow it in the tools.


kripken (approved):
Relaxation sgtm.

Do we ever emit such code in the toolchain, btw? I don't think emscripten can. Is there a specific future use case for this?

tlively:
Yes, it will now be possible to build objects or libraries with `-pthread` or `-matomics` and reuse them for both pthreads and single threaded builds.

PR #3173, opened 2020-09-25
Restore minification in emscripten builds

kripken:
The emscripten bug has been fixed in https://github.com/emscripten-core/emscripten/pull/12329


aheejin (approved):
Don't know much context, but it seems just reenabling a feature after a bugfix? Rubberstamped.

kripken:
Yes, it's just a re-enabling.

PR #3174, opened 2020-09-25
GC: Add stubs for remaining instructions

dcodeIO:
NFC, except adding most of the boilerplate for the remaining GC instructions. Will significantly reduce noise in follow-up PRs. Each implementation site is marked with a respective `TODO (gc): theInstruction` in between the typical boilerplate code.


tlively (approved):
Awesome! This will make future PRs very pleasant 👍

PR #3177, opened 2020-09-29
Clean up support/bits.h

tlively:
Use overloads instead of templates where applicable and change function names
from PascalCase to camelCase. Also puts the functions in the Bits namespace to
avoid naming conflicts.

cc @MaxGraey


MaxGraey (approved):

kripken (approved):
Nice!

PR #3162, opened 2020-09-23
Prototype extended-name-section proposal

dcodeIO:
Implements the parts of the [Extended Name Section Proposal for WebAssembly](https://github.com/WebAssembly/extended-name-section) that are trivially applicable to Binaryen, in particular:

* Table name<del>s</del>
* Memory name<del>s</del>
* Global names

Does not yet implement

* Label names
* Type names
* Elem names
* Data names

Given that unknown name section subsections will be ignored anyway, I didn't add a feature flag, but can add one if preferred.

Background: @binji [pointed out](https://github.com/WebAssembly/design/issues/1377#issuecomment-694424414) that there is already an existing proposal when I asked about support for global names as a follow-up to https://github.com/WebAssembly/binaryen/pull/3115#discussion_r486748807, and that

> to advance we really need only tools + engine support

As such, this PR, in case it is accepted, adds another tool to the list, if there is one already, with partial support. Will already be useful for us at AssemblyScript since we are using Binaryen outputs exclusively in testing, so we will have proper names, and I hope that more tools and engines will follow.

PR #3179, opened 2020-09-29
Fix applying default / align SExpr and Wasm builder names

dcodeIO:
SExpressionWasmBuilder was not applying default memory and table import names on the memory and table, unlike on functions, globals and events, where it applies them. Also aligns default import names to use the same shorter forms as in binary parsing.

Noticed during https://github.com/WebAssembly/binaryen/pull/3162, where something seemed off.


tlively (approved):
👍

PR #3180, opened 2020-09-30
Stop generating __growWasmMemory

sbc100:
See: https://github.com/emscripten-core/emscripten/pull/12378


scripts/wasm2js.js
```diff
@@ -6,12 +6,6 @@ var WebAssembly = {
   Memory: function(opts) {
     return {
       buffer: new ArrayBuffer(opts['initial'] * 64 * 1024),
-      grow: function(amount) {
-        var oldBuffer = this.buffer;
-        var ret = __growWasmMemory(amount);
-        assert(this.buffer !== oldBuffer); // the call should have updated us
-        return ret;
-      }
```
kripken, line 14: I think it's good to remove this from `wasm-emscripten-finalize`, but I'm not as happy with the emscripten side here - that adds a difference between wasm2js and non-wasm2js already at the link stage, which adds complexity and may make debugging trickier.

Can we leave this all to `wasm2js.h`? That is, as this PR does, remove work in `wasm-emscripten-finalize`, and then when a growth occurs the outside calls the Table `.grow()` method? (why is that removed here?)

sbc100: I agree.. I just wasn't sure how to make this memory grow function accessible from the generated wasm2js to to this outside world wrapper core.

I guess I removed it here because this helper is only used for testing it seems we have any tests here that create the memory on the outside and then grow it?

sbc100: (otherwise CI would have failed I right?)

kripken: Yes, we may not have a test that uses that path. In particular testing here doesn't use the emscripten output of wasm2js.

I think in the wasm2js case we can either import the memory, in which case we can replace any existing `grow` with our own, or we create and export it, in which case the current code should just work. In both cases the outside calls `wasmMemory.grow`, so we just need to set that from wasm2js output, unless I'm missing something.

sbc100: I see .. so the memory object that is received by wasm2js should get monkey patched to have our wasm2js-specific grow function?    So that the wasm2js module can effectively observe when the outside memory grow is attempted?

kripken: Yes, exactly.

In fact, reading `src/wasm2js.js` now in emscripten, literally all that call does is to call `__growWasmMemory`. So we can just erase it from emscripten. That is, when creating the Memory in JS, and then importing to wasm, the initial Memory would have no `.grow`, and the wasm2js output will fill that part in, adding `.grow = ..`.

kripken (approved):

sbc100:
This is going to cause a log of size improvements on the emscripten side so I guess I'll need to PR there to disable some codesize tests... looking into it now.

PR #3182, opened 2020-09-30
Update lld tests

aheejin:
This updates lld tests by running generate_lld_tests.py with the latest
toolchain.


aheejin:
I think the import name change is due to #3162, but not sure what the custom section size change is about. Is this expected?

sbc100 (approved):

tlively (approved):

tlively:
I'm not sure what the producer's section change is about either, but AFAIK that is only generated by upstream clang, so maybe something changed there?

PR #3185, opened 2020-09-30
wasm2js: override incoming memory's grow method

sbc100:
This will allow for the completely removal of
`__growWasmMemory` as a followup.  We currently
unconditionally generate this function in 
`generateMemoryGrowthFunction`.

See #3180


kripken (approved):
I think this is good, but you're making the JS output even more interesting, so might be worth running emscripten tests before landing, to avoid a minification surprise...

sbc100:
Yeah I've been ruing wasm2js0 wasm2 and other pretty consistently.

Issue #3081, opened 2020-08-27
Rename __invoke_* functions in wasm backend?


kripken:
Currently wasm-emscripten-finalize finds functions starting with `__invoke_` and renames them. This is for asm.js-style C++ exceptions and longjmp support. The renaming is important because the type in the wasm is the LLVM IR type, as this comment from `wasm-emscripten.cpp` says:
```cpp
  // LLVM backend lowers
  //   invoke @func(arg1, arg2) to label %invoke.cont unwind label %lpad
  // into
  // ... (some code)
  //   call @invoke_SIG(func, arg1, arg2)
  // ... (some code)
  // SIG is a mangled string generated based on the LLVM IR-level function
  // signature. In LLVM IR, types are not lowered yet, so this mangling scheme
  // simply takes LLVM's string representtion of parameter types and concatenate
  // them with '_'. For example, the name of an invoke wrapper for function
  // void foo(struct mystruct*, int) will be
  // "__invoke_void_%struct.mystruct*_int".
  // This function converts the names of invoke wrappers based on their lowered
  // argument types and a return type. In the example above, the resulting new
  // wrapper name becomes "invoke_vii".
```
So there can be many invokes with different IR names that turn into the same invoke on wasm types. The ratio can be very large, for example all LLVM IR pointers (`i32*, mystruct*`, etc.) would get a different name in IR, but all we need in wasm is an i32 for them.

In #3043 we are working to remove the work done in wasm-emscripten-finalize, so we'd like to remove this. We could support all the invokes in emscripten directly, but the large number would be a significant regression in debug builds (where we don't do anything after link). Instead, could we rename them in the wasm backend? That is, I think I understand what that comment says about using LLVM IR types, as the emscripten-exceptions and -longjmp passes are on LLVM IR. But could we add a backend pass that renames them?

(The names don't need to match anything - it can be any naming scheme whatsoever. The only goal is to avoid one invoke per LLVM IR type, it's that amount that is the problem.)

cc @aheejin @tlively

tlively:
Yes, I think this should be possible. What does wasm-emscripten-finalize rename them to? And what motivates the renaming?

kripken:
[...]

Issue #3043, opened 2020-08-14
Reduce modifications done in wasm-emscripten-finalize


kripken:
Currently wasm-emscripten-finalize both writes out some metadata and modifies the wasm. If we can reduce the amount of modifications, we can eventually get to a place where there are no modifications, at least in some build types, which would be simpler and faster.

kripken:
Not running wasm-emscripten-finalize uncovered what looks like an LLVM bug with wasm exceptions,

https://bugs.llvm.org/show_bug.cgi?id=47413

But that doesn't need to block this, as an experimental feature I think we could stop running `-fwasm-exceptions` on `test_exceptions_3` where it happens.

tlively:
@kripken, @sbc100, can we declare mission accomplished on this, or is there more to do?

sbc100:
By default we don't even run finalize anymore.  Closing.

PR #3191, opened 2020-10-01
Rename Emscripten EHSjLj functions in wasm backend

aheejin:
Now that we are renaming invoke wrappers and `emscripten_longjmp_jmpbuf`
in the wasm backend, this deletes all related renaming routines and
relevant tests. Depends on #3192.

Addresses: #3043 and #3081

Companions:
https://reviews.llvm.org/D88697
emscripten-core/emscripten#12399


test/lld/duplicate_imports.wat.out
```diff
@@ -32,6 +36,20 @@
...
+   (local.get $fptr)
+  )
+ )
+ (func $dynCall_ffd (param $fptr i32) (param $0 f32) (param $1 f64) (result f32)
+  (call_indirect (type $f32_f64_=>_f32)
+   (local.get $0)
+   (local.get $1)
+   (local.get $fptr)
+  )
+ )
```
aheejin, line 52: These dynCalls are generated because the input file contains `invoke_` functions. Before they were not targets for dynCall generations because `EmscriptenGlueGenerator` looked for the format generated by the LLVM backend, like `__invoke_void_i32`, and not `invoke_vi`.

src/passes/GenerateDynCalls.cpp
```diff
@@ -36,21 +36,103 @@ namespace wasm {
...
+  for (const auto& param : sig.params) {
+    args.push_back(builder.makeLocalGet(++i, param));
+  }
+  Expression* call = builder.makeCallIndirect(fptr, args, sig);
+  f->body = call;
+
+  wasm->addFunction(f);
+  exportFunction(*wasm, f->name, true);
+}
+
```
sbc100, line 135: How about making this its own PR that would be NFC?   Seems like it might be nice intermediate step?

src/passes/GenerateDynCalls.cpp
```diff
@@ -36,21 +36,103 @@ namespace wasm {
...
-        generator.generateDynCallThunk(
-          getModule()->getFunction(indirectFunc)->sig);
+        generateDynCallThunk(getModule()->getFunction(indirectFunc)->sig);
       }
     }
   }
 
+  void visitFunction(Function* func) {
+    // Generate dynCalls for invokes
+    if (func->imported() && func->base.startsWith("invoke_")) {
```
kripken, line 58: please check also that `func->module == ENV`

kripken (approved):

src/passes/GenerateDynCalls.cpp
aheejin: Done in #3192. Will rebase this PR on top of that. It's inconvenient that Github doesn't support dependent PRs... (We can emulate it if I used branches in this repo and not my own fork though..)

src/passes/GenerateDynCalls.cpp
aheejin: Done

aheejin:
I incorporated invoke wrapper renaming we discussed in the meeting this morning here; I thought about doing it as a separate patch but it will take three more patches, each from LLVM, Binaryen, and Emscripten, and rolling them will be a pain, so...

aheejin:
Sorry for changing things again; I reverted the invoke renaming part. Renaming this turns out to change a lot of things in Emscripten side, which I think is really better be a separate patch.

src/passes/GenerateDynCalls.cpp
```diff
@@ -55,7 +55,8 @@ struct GenerateDynCalls : public WalkerPass<PostWalker<GenerateDynCalls>> {
 
   void visitFunction(Function* func) {
     // Generate dynCalls for invokes
-    if (func->imported() && func->base.startsWith("invoke_")) {
+    if (func->imported() && func->module == ENV &&
+        func->base.startsWith("invoke_")) {
```
sbc100, line 59: This seems like maybe unneeded/unrelated?

If you do want to make this change you you should change the logic in src/wasm/wasm-emscripten.cpp which seems do duplicate this when generating invokeFuncs mabye?

sbc100 (approved):
I'm curious, can the llvm change land before this on its own? 

It seems like it should be possible .. it would just mean the fixInvokeFunctionNames would becoming inactive.. and then change to roll after?

src/passes/GenerateDynCalls.cpp
aheejin: It was suggested in https://github.com/WebAssembly/binaryen/pull/3191#discussion_r498557152, and yeah, I think the condition in wasm/wasm-emscripten.cpp should be the same. Should I duplicate the condition there? @kripken

aheejin:
> I'm curious, can the llvm change land before this on its own?
> 
> It seems like it should be possible .. it would just mean the fixInvokeFunctionNames would becoming inactive.. and then change to roll after?

Yeah I think llvm change can land on its own; maybe I was not clear when I was asking for the double roll thing.. What I meant was, Emscripten CI will pass only after both the LLVM and Binaryen patches land. But landing those two will break the roller because Emscripten patch hasn't landed.

src/passes/GenerateDynCalls.cpp
aheejin: Added the same condition in wasm-emscripten.cpp.

PR #3192, opened 2020-10-01
Let GenerateDynCalls generate dynCalls for invokes

aheejin:
This moves dynCall generating functionaity for invokes from
`EmscriptenGlueGenerator` to `GenerateDynCalls` pass. So now
`GenerateDynCalls` pass will take care of all cases we need dynCalls:
functions in tables and invokes.


test/lld/duplicate_imports.wat.out
```diff
@@ -32,6 +36,20 @@
...
+   (local.get $fptr)
+  )
+ )
+ (func $dynCall_ffd (param $fptr i32) (param $0 f32) (param $1 f64) (result f32)
+  (call_indirect (type $f32_f64_=>_f32)
+   (local.get $0)
+   (local.get $1)
+   (local.get $fptr)
+  )
+ )
```
aheejin, line 52: These dynCalls are generated because the input file contains `invoke_` functions. Before they were not targets for dynCall generations because `EmscriptenGlueGenerator` looked for the format generated by the LLVM backend, like `__invoke_void_i32`, and not `invoke_vi`.

sbc100 (approved):
Nice!

PR #3200, opened 2020-10-08
Remove old attempted DSL beginnings from OptimizeInstructions

kripken:
Wasm turned out to not be that good for a DSL for such peephole optimizations,
so that never made progress. Meanwhile we have the new matcher stuff which
works well.

In the future perhaps we can find a useful DSL to make this even easier, but
I have no ideas or plans for that atm.


tlively (approved):
Maybe a proper DSL that emits optimized C++ could make a good intern project sometime.

PR #3204, opened 2020-10-08
Fuzz fix for DuplicateFunctionElimination

kripken:
The `replaceFunctions` utility replaced exports by name, but did not check
the kind, so it could get confused when names happen to overlap (see
testcase).

### DIFF
--- a/src/passes/opt-utils.h
+++ b/src/passes/opt-utils.h
@@ -97,7 +97,9 @@ inline void replaceFunctions(PassRunner* runner,
   }
   // replace in exports
   for (auto& exp : module.exports) {
-    maybeReplace(exp->value);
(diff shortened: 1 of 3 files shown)


aheejin (approved):
Other calls for `maybeReplace` don't need to be guarded in the same way? If so, would moving the guard inside `maybeReplace` be simpler?

kripken:
The other places don't need such a guard - they are all on things that can only be a function, like the `start` function of the module, or the Table (although I guess that's changing with the ability to put externrefs in Tables?). The export check is specific to exports, so it can't be moved into `maybeReplace` - if we need another check, we'd need something special for each case (like if Table elements can be things other than functions, we'd need a way to check that - but right now Table elements are just names).

PR #3205, opened 2020-10-08
Asyncify fuzzing harness fix for growth

kripken:
Fuzzing Asyncify on data with a memory growth showed that the harness
did not handle a growth. When growth happens we must recreate the view.

### DIFF
--- a/scripts/fuzz_shell.js
+++ b/scripts/fuzz_shell.js
@@ -55,6 +55,7 @@ var Asyncify = {
         if (typeof imports[module][i] === 'function') {
           (function(module, i) {
             ret[module][i] = function() {
+              refreshView();
               if (!Asyncify.sleeping) {
                 // Sleep if asyncify support is present (which also requires
                 // that the memory be exported), and at a certain probability.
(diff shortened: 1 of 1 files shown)


sbc100 (approved):
I don't know this code at all.  Happy to rubberstamp it unless there is someone else like @tlively who might know it?

kripken:
(This is mostly a JS question, about the need for refreshing typed array views after a growth - nothing much in terms of interaction with the existing code itself.)

scripts/fuzz_shell.js
```diff
@@ -179,8 +180,14 @@ var instance = new WebAssembly.Instance(new WebAssembly.Module(binary), imports)
...
-if (exports.memory) {
-  var view = new Int32Array(exports.memory.buffer);
+
+var view;
+
+// Recreate the view. This is important both initially and after a growth.
+function refreshView() {
+  if (exports.memory) {
+    view = new Int32Array(exports.memory.buffer);
+  }
```
sbc100, line 190: It looks like is not immediately run whereas it was before.  I assume that is indended.

sbc100 (approved):

scripts/fuzz_shell.js
kripken: Yes, I put it in a place that dominates all the uses.

tlively (approved):
LGTM, although I also have only a tenuous grasp of what this file does.

PR #3206, opened 2020-10-09
Fix split_wast on asserts before the first module

kripken:
Normally a wast file has a module and then asserts on it, but
some tests have just asserts without a module. Before this PR
we would break if such asserts happened before the first
module, as we didn't know what module to attach them to.

### DIFF
--- a/scripts/test/support.py
+++ b/scripts/test/support.py
@@ -146,7 +146,10 @@ def split_wast(wastFile):
         elif chunk.startswith('(assert_invalid'):
             continue
         elif chunk.startswith(('(assert', '(invoke')):
-            ret[-1][1].append(chunk)
+            # ret may be empty if there are some asserts before the first
+            # module
+            if ret:
+                ret[-1][1].append(chunk)
     return ret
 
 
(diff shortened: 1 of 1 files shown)


scripts/test/support.py
```diff
@@ -146,7 +146,10 @@ def to_end(j):
         elif chunk.startswith('(assert_invalid'):
             continue
         elif chunk.startswith(('(assert', '(invoke')):
-            ret[-1][1].append(chunk)
+            # ret may be empty if there are some asserts before the first
+            # module
+            if ret:
+                ret[-1][1].append(chunk)
```
aheejin, line 152: So we just ignore those asserts? Should we error out instead?

aheejin (approved):

scripts/test/support.py
kripken: No, they are valid. It's ok to have an assert without a module, something like `assert_invalid( some bad wast without a module reference )`

kripken: Oh sorry, I misunderstood the question.

Yes, right now we just ignore them. I actually need this fixed for getting the modules. But yes, this means this method will ignore some asserts. I'll add a FIXME, as in theory we want that too.

kripken: I made it return a None module in that case, which does seem more correct. Code calling this will need to handle that.

PR #2880, opened 2020-05-27
Add EH support for fuzzer

aheejin:
This adds exception handling instruction generation support for the
fuzzer. One thing this has to handle is, EH instructions do not support
Flatten and dependent passes (several passes including ReReloop and
LocalCSE), so they have to be disabled in the fuzzer when EH is enabled.
Also this adds `--disable-flatten` option to wasm-reduce, which makes
the reducer do not use Flatten and dependent passes when reducing.

Instruction generation is straightforward except for `catch`. Due to the
restriction that `exnref.pop` follows right after `catch`, we cannot do
the usual top-down generation and have to choose among several common
scenarios that start with `exnref.pop`. Also we need to exclude the
'dangling pop' expression, which contains `exnref.pop` from certain
churns in the late stage of fuzzing, in order to prevent it from being
separated from `catch`.

PR #3207, opened 2020-10-09
Add a little code to prepare exception handling support in fuzzer

kripken:
`fixLabels()` in the fuzzer looks for invalid labels and fixes them up,
after doing some random changes to existing wasm (which checks
for types while doing so, but it may invalidate labels if we remove the
target of a branch, for example). This adds trivial support for
`BrOnExn` and `Try` there.

I wasn't working on exceptions, just a fuzz testcase happened to hit
this, and this was enough to get past it.

### DIFF
--- a/src/tools/fuzzing.h
+++ b/src/tools/fuzzing.h
@@ -729,6 +729,8 @@ private:
 
       void visitBreak(Break* curr) { replaceIfInvalid(curr->name); }
 
+      void visitBrOnExn(BrOnExn* curr) { replaceIfInvalid(curr->name); }
+
       bool replaceIfInvalid(Name target) {
         if (!hasBreakTarget(target)) {
           // There is no valid parent, replace with something trivially safe.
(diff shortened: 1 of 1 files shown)


aheejin (approved):
Thanks! This was also included in #2880, but I was holding off on landing it because I wasn't able to debug a fuzzer failure on `CheckDeterminism` found after like a million iterations ;(

kripken:
@aheejin Ok, let's land this for now, but if the determinism issue in the other PR is that rare, maybe we can land it too? As this is an experimental feature anyhow. We could also disable the determinism check in the fuzzer when exceptions are enabled.

aheejin:
We have to land a major change soon, and having the fuzzer on makes it difficult. If you agree with the timeline that "enable EH fuzzer" -> "disable the fuzzer for EH before we land the major change" -> "Add EH changes piece by piece" -> "Once everything is done reenable the EH fuzzer", that might work.

kripken:
@aheejin oh, it sounds fine to me either way. I'm doing some experiments locally with turning more things in the fuzzer, but there is no rush to do so in general. can wait for whenever things stabilize.

PR #3208, opened 2020-10-09
Optimize getLoggableTypes in fuzzer, and add isLoggableType

kripken:
Previously `getLoggableType` would allocate a vector every time. This
caches it.

Also add `isLoggableType` which will need this caching as I will be
calling it quite a lot in a future patch to the fuzzer.


aheejin (approved):
I guess the caching thing can happen for other similar `get***Types` methods.. should we do that?

kripken:
I think caching matters if we think we'll be calling it constantly. So we may need to think about each one separately. Or maybe just run the profiler a little to see if anything is really obvious.

PR #3211, opened 2020-10-09
EmscriptenPIC: Remove internalization of GOT entries

sbc100:
wasm-ld now has a `-Bsymbolic` flag whihc allows these
this to be done directly in the linker.

See https://reviews.llvm.org/D89152


src/passes/EmscriptenPIC.cpp
```diff
@@ -110,34 +93,8 @@ struct EmscriptenPIC : public WalkerPass<PostWalker<EmscriptenPIC>> {
...
-      module->memory.exists && module->memory.segments.size() == 1;
-
     for (Global* g : gotMemEntries) {
-      // If this global is defined in this module, we export its address
-      // relative to the relocatable memory. If we are in a main module, we can
-      // just use that location (since if other modules have this symbol too, we
-      // will "win" as we are loaded first). Otherwise, import a g$ getter. Note
-      // that this depends on memory having a single segment, so we know the
-      // offset, and that the export is a global.
       auto base = g->base;
```
kripken, line 97: Sounds good to let wasm-ld do this work instead of us. Does it do the optimizations mentioned in these comments?

kripken (approved):

src/passes/EmscriptenPIC.cpp
sbc100: Essentially yes, it avoid the GOT.mem or GOT.func import when we are the main module and that symbol is defined in the main module.

This means that gotMemEntries and gotFuncEntries for things we export will simply not be present anymore.

I will try replacing with an assert just to be 100% sure...

PR #3214, opened 2020-10-09
Log nested pass names in BINARYEN_PASS_DEBUG=2

kripken:
This makes it easier to debug problems with nested passes.


aheejin:
I might be mistaken, but shouldn't this be added to [this part](https://github.com/WebAssembly/binaryen/blob/f4d074fc3ee8184cd89c870612fc316adaeb8e1a/src/passes/pass.cpp#L545-L614) of the code instead?

kripken:
I think that part of the code is for the non-nested case (when running non-nested passes, we already had good logging.)

aheejin:
Sorry I might be mistaken and you wrote that part of the code, but help me understand. So the `if` condition there is
```cpp
if (!isNested && (options.debug || passDebug))
```
I interpreted this as, `if (debug mode)`. But we haven't logged this for nested passes so far. That's the reason we have `!isNested`.

And the its matching `else` part is
```cpp
else {
    // non-debug normal mode, run them in an optimal manner - for locality it is
    // better to run as many passes as possible on a single function before
    // moving to the next
```
So it says "non-debug normal mode", and we are not running passes one by one here for easy debugging, but rather running as many passes as possible on a single function before moving to the next.
This is the part you added the nested pass logging. Is this what you intended?

kripken:
> I interpreted this as, if (debug mode). But we haven't logged this for nested passes so far. That's the reason we have !isNested. So I thought we should change this part to log nested case too.

The confusing thing is maybe that the old logging code *doesn't work* with nested code. Basically, the old logging code does things like print out the module, which requires running a pass (as Print is a pass). So the old logging code was just done on non-nested passes, which avoids the printing issue.

This PR adds very minimal logging to the nested case, and does not do any printing, etc.

You're right that this is confusing. Two options might be:

1. Add more comments explaining what's going on.
2. Refactor the existing code to add ifs around each of the print operations, but otherwise keep all the logging in the same place (unlike this PR which adds logging to a new place).

Thoughts?

src/passes/pass.cpp
```diff
@@ -546,27 +546,33 @@ static void dumpWast(Name name, Module* wasm) {
 
 void PassRunner::run() {
   static const int passDebug = getPassDebug();
-  if (!isNested && (options.debug || passDebug)) {
+  // Emit logging information when asked for. At passDebug level 1+ we log
+  // the main passes, while in 2 we also log nested ones. Note that for
+  // nested ones we can only emit their name - we can't validate, or save the
+  // file, or print, as the wasm may be in an intermediate state that is not
+  // valid.
+  if (options.debug || (passDebug == 2 || (passDebug && !isNested))) {
```
aheejin, line 554: What does `options.debug` do here?

src/passes/pass.cpp
```diff
@@ -546,27 +546,33 @@ static void dumpWast(Name name, Module* wasm) {
...
     // for debug logging purposes, run each pass in full before running the
     // other
     auto totalTime = std::chrono::duration<double>(0);
-    size_t padding = 0;
     WasmValidator::Flags validationFlags = WasmValidator::Minimal;
     if (options.validateGlobally) {
       validationFlags = validationFlags | WasmValidator::Globally;
     }
-    std::cerr << "[PassRunner] running passes..." << std::endl;
+    auto what = isNested ? "nested passes" : "passes";
```
aheejin, line 562:
```suggestion
    auto what = isNested ? "  nested passes" : "passes";
```
We talked offline and it is hard to track depths of nested passes, but maybe it'd be good to just add a uniform indentation for nested passes?

aheejin (approved):
Sorry and thank you for answer my many questions (here and offline)!

src/passes/pass.cpp
kripken: It's a way for a specific PassRunner object itself to be marked as debug (instead of globally using the env var). Not sure if it's useful actually.

src/passes/pass.cpp
kripken: That would be good, but I think this change would just make it print `Running    nested passes` (that is, the extra indentation is not at the beginning of the line).

kripken: Looking around, it doesn't seem like there is an easy and obvious way to add indentation here (it would need to affect more than one place), so let's leave that for work later maybe.

PR #3216, opened 2020-10-09
wasm2js: Avoid mangling export names

kripken:
If accessed as strings (see change to fuzz harness), there seems to be no reason to
mangle them. This way the original name is used, which is nicer, as the test suite
diff shows, e.g.
```diff
 return {
  "atomic_fence": $0
  "atomic-fence": $0  // the original name in the wasm
 };
}
```
This seems to work for corner cases like an empty string too.

Maybe there's some JS reason I'm missing though, like some form of JS modules
works with JS identifiers and not strings..? @RReverser @dcodeIO


RReverser:
> Maybe there's some JS reason I'm missing though, like some form of JS modules
> works with JS identifiers and not strings..?

ESM integration does that, yeah, but I'm not sure if wasm2js output is intended to be used in such contexts?

dcodeIO:
What one can't do then is to

```js
import { atomic-fence } from "./myModule"`;
         ~~~~~~~~~~~~
```

because import syntax requires identifiers. Not sure of the implications, because seeing the exact names has advantages as well (as long as there is an `export default` to obtain the object with the non-identifier keys one can access with `["atomic-fence"]`).

kripken:
I see, thanks... then this is a bad idea then, as I think we do want the output here to be as ESM-friendly as possible. Unless we want to revisit that decision, but it's the default output atm.

This will make fuzzing harder, sadly, but I'll see if I can figure that out.

sbc100:
> What one can't do then is to
> 
> ```js
> import { atomic-fence } from "./myModule"`;
>          ~~~~~~~~~~~~
> ```
> 
> because import syntax requires identifiers. Not sure of the implications, because seeing the exact names has advantages as well (as long as there is an `export default` to obtain the object with the non-identifier keys one can access with `["atomic-fence"]`).

But if one wants to that that aren't there other ways to do it:

1. Design your module with JS-compatible names?
2. Have some kind of other separate pass to do the name mangling: "mangleExportsForJsCompat"

It seems that (2) would be useful regardless of wasm2js perhaps when directly importing wasm sybmols.

I like this change because its means that wasm2js users don't need to learn a name mangling scheme.. they just just use the same names as that were in the original wasm.

kripken:
@sbc100 good ideas, I like that plan, reopening.

kripken:
I opened #3281 to track this - the new approach would not use the same code, so doesn't make sense to leave this PR open.

PR #3217, opened 2020-10-09
Handle uncaught wasm exceptions in shell-interface and execution-results

kripken:
Very similar to the existing handling of traps.

This is needed for fuzzing.

### DIFF
--- a/src/shell-interface.h
+++ b/src/shell-interface.h
@@ -226,7 +226,10 @@ struct ShellExternalInterface : ModuleInstance::ExternalInterface {
     throw TrapException();
   }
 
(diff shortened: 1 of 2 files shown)


aheejin (approved):
Are you adding EH fuzzing to Binaryen now? Then why don't we just land #2880? (This was included in #2880 too)

kripken:
Sorry, yeah, I didn't remember that PR, and so it seems I've recreated small parts of it here and there...

There's no need to land this, if it's not tested for now. I'll make the new fuzzer approach avoid EH for now.

PR #3218, opened 2020-10-09
wasm-reduce: Don't try to replace a non-number (like a reference) with a Const

kripken:


aheejin (approved):

kripken:
Actually I see this was wrong, I should have put it a few lines lower down (as it can handle a few non-number things). Should be fixed now.

PR #3219, opened 2020-10-09
Fuzzing: Don't recombine (copy to various places) Pops

kripken:
A Pop has very limitation valid locations atm.

### DIFF
--- a/src/tools/fuzzing.h
+++ b/src/tools/fuzzing.h
@@ -605,6 +605,10 @@ private:
       std::map<Type, std::vector<Expression*>> exprsByType;
 
       void visitExpression(Expression* curr) {
+        // Some things cannot be moved around easily, give up on them.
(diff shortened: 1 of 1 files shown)


aheejin:
This was also a part of #2880 😂... I did a similar thing for `recombine::Modder::visitExpression`, `mutate::Modder::visitExpression` there. I used 
```cpp
if (contains<Pop>(curr, 2)) {
  return;
}
```
because all patterns the EH fuzzer (#2880) generates has `pop` within the depth of 2.

By the way I'm wondering, without the EH fuzzer, why `pop`s are generated in the first place?

kripken:
I'm only kind-of adding EH support... I'm running the fuzzer in a way that starts with an existing testcase, then adds fuzzing on top. So if the testcase had a pop, it would come from there (and maybe be moved around, copied, etc.).

PR #3220, opened 2020-10-10
RemoveUnusedBrs fuzz fix

kripken:
We turn a `br_table` with a single target into a `br`, but we reverse the order of the
condition and the value when doing so, which we forgot to take into account.


tlively (approved):
Oh wow, how long has that bug been there?

kripken:
> Oh wow, how long has that bug been there?

Looks like almost 2 years, #1753

The tricky thing is that the bug only happens with a switch with a value, which LLVM never emits, and apparently the (old) fuzzer doesn't emit enough...

PR #3228, opened 2020-10-12
Fuzzer: Fix wasm2c name mangling

kripken:
This is necessary for cases where the input has an export name that
needs mangling, like a name with `-` (common in the spec test suite).

### DIFF
--- a/src/tools/wasm2c-wrapper.h
+++ b/src/tools/wasm2c-wrapper.h
@@ -26,9 +26,9 @@
 namespace wasm {
 
 // Mangle a name in (hopefully) exactly the same way wasm2c does.
-static std::string wasm2cMangle(Name name) {
+static std::string wasm2cMangle(Name name, Signature sig) {
   const char escapePrefix = 'Z';
-  std::string mangled;
+  std::string mangled = "Z_";
   const char* original = name.str;
   unsigned char c;
   while ((c = *original++)) {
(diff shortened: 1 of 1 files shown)


src/tools/wasm2c-wrapper.h
```diff
@@ -25,6 +25,27 @@
...
+      mangled += c;
+    } else {
+      // This must be escaped, as prefix + hex character code.
+      mangled += escapePrefix;
+      std::stringstream ss;
+      ss << std::hex << std::uppercase << unsigned(c);
+      mangled += ss.str();
+    }
+  }
+  return std::string("(*Z_") + mangled + "Z_";
```
aheejin, line 46: I compared this code with wasm2c's [`MangleName`](https://github.com/WebAssembly/wabt/blob/1eb5e2c5b7409a64c6e610fb74df3e5b61394cfc/src/c-writer.cc#L408-L425) method and

kripken: Thanks, I think you're right, not sure how this worked in my local testing... fixed.

kripken: Oh, actually the code worked before, but was confusing, sorry about that. I refactored it now so it's hopefully clearer.

What was unclear was that the "mangle" function only mangled part of the name, leaving the params etc. for later. There is another `Z_` before them. I moved all the mangling logic into a single place now

binji (approved):
Generally looks right to me

aheejin (approved):
Sorry, I was confused ;)

PR #3229, opened 2020-10-12
Fuzzer: Do not emit random global.get/sets of the hang limit global

kripken:
That global is for internal use. If we emit random sets to it, we could prevent
it from doing its job of preventing an infinite loop (normally it decreases each
time a loop runs or we recurse, until we reach 0 - if we set it to a nonzero
value in that code, that would be bad).

Random sets are less of a problem, but may be confusing when debugging
a testcase.

### DIFF
--- a/src/dataflow/graph.h
+++ b/src/dataflow/graph.h
@@ -157,9 +157,7 @@ struct Graph : public UnifiedExpressionVisitor<Graph, Node*> {
     return ret;
   }
 
-  Node* makeZero(wasm::Type type) {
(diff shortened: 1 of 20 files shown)


aheejin (approved):
It's surprising this hasn't caused problems so far..?

> Random sets are less of a problem, but may be confusing when debugging a testcase.
Do you mean random gets?

kripken:
I think this is either very very unlikely, or maybe it's more likely with the new fuzzer variation somehow, not sure...

Oh, yes, I meant gets in the second paragraph, not sets.

PR #3230, opened 2020-10-12
Added Initial Memory64Lowering pass

aardappel:
This pass will convert a module with 64-bit loads and stores accessing a 64-bit memory to a regular 32-bit one.
Pointers remain 64-bit but are truncated just before use.


src/passes/Memory64Lowering.cpp
```diff
@@ -0,0 +1,82 @@
...
+    auto& module = *getModule();
+    if (!module.memory.is64()) {
+      return;
+    }
+    if (ptr->type == Type::unreachable) {
+      return;
+    }
+    assert(ptr->type == Type::i64);
+    Builder builder(*getModule());
+    ptr = builder.makeUnary(UnaryOp::WrapInt64, ptr);
```
sbc100, line 41: Is this something that was just pre-existing in binaryen?  Very cool.

I love how small it makes this patch.

sbc100 (approved):
lgtm, although @kripken should probably give the final OK as he knows the code way better than me.

src/passes/Memory64Lowering.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+namespace wasm {
+
+struct Memory64Lowering : public WalkerPass<PostWalker<Memory64Lowering>> {
+
+  void WrapIfAddress64(Expression*& ptr) {
+    auto& module = *getModule();
+    if (!module.memory.is64()) {
+      return;
+    }
```
tlively, line 35: It would be good to do this in a `doWalkFunction` implementation to avoid the traversal entirely when possible.

```c++
void doWalkFunction(Function* func) {
  if (getModule()->memory.is64()) {
    super::doWalkFunction(func);
  }
}
```

sbc100: Or skip the entire module?

aardappel: This is not a pass that is by default on, it will be something people select because they specifically want this lowering. But either way works for me.

How do I skip the entire module? `doWalkFunction(Module*)` I guess?

tlively: You'll want to override `void run(PassRunner* runner, Module* module)`. If wasm64 is enabled, you can have it call `super::run(runner, module)`.

aardappel: I guess this is all going to blow up when we support accessing multiple memories at once, but ok..

src/passes/Memory64Lowering.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include "ir/bits.h"
+#include "pass.h"
+#include "wasm-builder.h"
+#include "wasm.h"
+
+namespace wasm {
+
+struct Memory64Lowering : public WalkerPass<PostWalker<Memory64Lowering>> {
+
+  void run(PassRunner* runner, Module* module) {
```
kripken, line 31:
```suggestion
  void run(PassRunner* runner, Module* module) override {
```

kripken (approved):
Nice!

This could in theory run the function parts in parallel, but as this is will not be in the default optimziation pipeline, that doesn't seem urgent. But if we notice people using this on huge modules in the future, we may want to do that. Could be worth a TODO.

src/passes/Memory64Lowering.cpp
```diff
@@ -0,0 +1,86 @@
...
+    if (module->memory.is64()) {
+      super::run(runner, module);
+    }
+  }
+
+  void WrapIfAddress64(Expression*& ptr) {
```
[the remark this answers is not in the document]
aardappel: Done

src/passes/Memory64Lowering.cpp
aardappel: Done.

aardappel:
@kripken added todo

PR #3238, opened 2020-10-14
Assign import names consistently between text and binary reader

sbc100:
The s-parser was assigning numbers names per-type where as
the binaryn reader was using the global import count as the
number to append.

This change switches to use per-element count which I think
it preferable as it increases the stability of the auto-generated
names.  e.g. memory is now always named `$mimport0`.


tlively (approved):

kripken:
This seems to have broken the binaryen roll: https://logs.chromium.org/logs/emscripten-releases/buildbucket/cr-buildbucket.appspot.com/8866373888698788080/+/steps/Emscripten_testsuite__upstream__other_/0/stdout

sbc100:
Taking a look now.

sbc100:
The followup changes fixes the problem: https://github.com/WebAssembly/binaryen/pull/3241

PR #3239, opened 2020-10-14
[MemoryPacking] Emit the correct segment indices on memory.init

### LINKED ISSUE #3225: Fuzz bug in MemoryPacking
```wat
(module
 (type $none_=>_externref_f64_i64_f64_i32_exnref (func (result externref f64 i64 f64 i32 exnref)))
 (type $v128_=>_externref_f64_i64_f64_i32_exnref (func (param v128) (result externref f64 i64 f64 i32 exnref)))
 (memory $0 (shared 2048 2049))
 (data passive "h\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00v\00n\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00")
 (data passive "")
 (data passive "")
 (data passive "")
 (data passive "")
 (data (i32.const 0) "")
 (data passive "")
 (data (i32.const 5) "")
 (data (i32.const 13) "")
 (export "func_54" (func $0))
 (func $0 (param $0 v128) (result externref f64 i64 f64 i32 exnref)
  (memory.init 0
   (i32.const 6)
   (i32.const 2)
   (i32.const 51)
  )
  (tuple.make
   (ref.null extern)
   (f64.const -nan:0xfffffffffbb56)
   (i64.const -130625735552240)
   (f64.const 0.235)
   (i32.const 102)
   (ref.null exn)
  )
 )
)
```
STR
```
$ wasm-opt w.wat --memory-packing --fuzz-exec
[fuzz-exec] calling func_54
[fuzz-exec] note result: func_54 => (externref(null), -nan:0xfffffffffbb56, -130625735552240, 0.235, 102, exnref(null))
[fuzz-exec] calling func_54
[trap out of bounds segment access in memory.init]
[fuzz-exec] comparing func_54
not identical! () != (externref(null), -nan:0xfffffffffbb56, -130625735552240, 0.235, 102, exnref(null))
[fuzz-exec] optimization passes changed execution results
```

tlively:
This PR fixes a bug in which the segment index of a memory.init instruction was
incorrect in some circumstances. Specifically, the first segment index used in
output memory.init instructions was always the index of the first segment
created from splitting up the corresponding input segment. This was incorrect
when the input memory.init had an offset that caused it to skip over that first
emitted segment so that the first output memory.init should have referred to a
subsequent output segment.

Fixes #3225.

### DIFF
--- a/src/passes/MemoryPacking.cpp
+++ b/src/passes/MemoryPacking.cpp
@@ -514,10 +514,16 @@ void MemoryPacking::createReplacements(Module* module,
     size_t start = init->offset->cast<Const>()->value.geti32();
     size_t end = start + init->size->cast<Const>()->value.geti32();
 
+    // Segment index used in emitted memory.init instructions
+    size_t initIndex = segmentIndex;
+
     // Index of the range from which this memory.init starts reading
     size_t firstRangeIdx = 0;
     while (firstRangeIdx < ranges.size() &&
            ranges[firstRangeIdx].end <= start) {
+      if (!ranges[firstRangeIdx].isZero) {
+        ++initIndex;
+      }
       ++firstRangeIdx;
     }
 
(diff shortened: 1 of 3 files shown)


kripken (approved):
Nice!

Issue #3226, opened 2020-10-12
wasm-opt adds names that are not present in the input 


sbc100:
As of #3162 it seems that emscripten output files include a names for tables/memories/globals that are not present in the input:

```
$ ./emcc -g ~/test/hello.c -o out.wasm
$ wasm-objdump -x out.wasm
 - table[0] <0>
 - memory[0] <0>
 - global[0] <global$0>
 - global[1] <global$1>
```

@dcodeIO please take a look.  We've had users of emscripten who have been stung by this unexpected change.

sbc100:
I looks like this problem actually predates #3162.. wasm-opt will also have made up names for functions if they are not found in the input.

I think the problem is that binaryen doesn't distinguish between something that a user-defined name, and something that was given and internal name simply because binaryen needs to assign a name to everything.

kripken:
Why is this a problem? (a few more bytes in the names section doesn't seem bad, as it isn't present in release builds anyhow..?)

sbc100:
The name section is supposed to be for user-visible names, not these internal names that we make up for binaryen-specific reasons.      I won't expect tools like binaryen and wabt to add names to my binaries that were not present in the input.

I'd rather turn the question around: what useful purpose does adding these number-based named serve?   For functions it just gives them the names `0`.. `1` .. `2`.  This is not helpful, and a little misleading because it might suggest to the user that the `1` is the name of the function in the source language which is clearly isn't.

sbc100:
Isn't the fact that binaryen assigns numeric names to things as it reads them in is an implementation detail that wasm-opt should not persist in the output file?

kripken:
I agree. Sorry if I wasn't clear before. Yes, it would be better to avoid this. I was just trying to say, it seems like a very minor aesthetic issue? When you said

> We've had users of emscripten who have been stung by this unexpected change.

That sounds like something more serious to me, but maybe I'm misreading it.

sbc100:
> I agree. Sorry if I wasn't clear before. Yes, it would be better to avoid this. I was just trying to say, it seems like a very minor aesthetic issue? When you said
> 
> > We've had users of emscripten who have been stung by this unexpected change.
> 
> That sounds like something more serious to me, but maybe I'm misreading it.

This part is because most wasm parsers out there don't yet handle the names for memories/tables/globals there were recently added.  Clearly it would be good to fix those parsers too (they should be ignoring those failures anyway).   But also good to make this fix to binaryen.

[...]

PR #3241, opened 2020-10-15
Only write explicit names to name section

sbc100:
Fixes: #3226

### DIFF
--- a/src/wasm.h
+++ b/src/wasm.h
@@ -1332,30 +1332,28 @@ public:
   void finalize() { WASM_UNREACHABLE("TODO (gc): array.len"); }
 };
 
-// A named element whose name can end up in the name section.
-// Explicit names are ones that we read from the input file and
-// will be written the name section in the output file.
-// Implicit names are names that binaryen generated for internal
-// use only and will not be written the name section.
-struct Namable {
-  Name name;
-  bool explicitName = false;
-
-  void setName(Name name_, bool explicitName_) {
-    name = name_;
-    explicitName = explicitName_;
-  }
-
-  void setExplicitName(Name name_) { setName(name_, true); }
-};
-
 // Globals
 
-struct Importable : public Namable {
+struct Importable {
+  Name name;
+
+  // Explicit names are ones that we read from the input file and
+  // will be written the name section in the output file.
+  // Implicit names are names that binaryen generated for internal
+  // use only and will not be written the name section.
+  bool hasExplicitName = false;
+
   // If these are set, then this is an import, as module.base
   Name module, base;
 
   bool imported() { return module.is(); }
+
+  void setName(Name name_, bool hasExplicitName_) {
+    name = name_;
+    hasExplicitName = hasExplicitName_;
+  }
+
+  void setExplicitName(Name name_) { setName(name_, true); }
 };
 
 class Function;
--- a/src/wasm.h
+++ b/src/wasm.h
@@ -1332,9 +1332,26 @@ public:
   void finalize() { WASM_UNREACHABLE("TODO (gc): array.len"); }
 };
 
+// A named element whose name can end up in the name section.
+// Explicit names are ones that we read from the input file and
+// will be written the name section in the output file.
+// Implicit names are names that binaryen generated for internal
+// use only and will not be written the name section.
+struct Namable {
+  Name name;
+  bool explicitName = false;
+
+  void setName(Name name_, bool explicitName_) {
+    name = name_;
+    explicitName = explicitName_;
+  }
+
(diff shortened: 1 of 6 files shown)


sbc100:
Adding tests for this is a little tricky since its hard to observe using just wasm-dis .. since having the default/internal names in the name section looks exactly like not having them once loaded into binaryen.

I did verify using emscripten that this has the desired effect.

src/wasm.h
```diff
@@ -1332,9 +1332,26 @@ class ArrayLen : public SpecificExpression<Expression::ArrayLenId> {
   void finalize() { WASM_UNREACHABLE("TODO (gc): array.len"); }
 };
 
+// A named element whose name can end up in the name section.
+// Explicit names are ones that we read from the input file and
+// will be written the name section in the output file.
+// Implicit names are names that binaryen generated for internal
+// use only and will not be written the name section.
+struct Namable {
```
tlively, line 1340: `Namable` => `Nameable`

Also, do we plan to use this for anything that's not `Importable`? If so, it might make sense to just make this interface part of `Importable`.

src/wasm/wasm-binary.cpp
```diff
@@ -628,7 +628,7 @@ void WasmBinaryWriter::writeNames() {
   }
 
   // table names
-  if (wasm->table.exists && wasm->table.name.is()) {
+  if (wasm->table.exists && wasm->table.explicitName) {
```
tlively, line 631: Renaming `explicitName` to `isNameExplicit` would help make it clear that this is a boolean and not the actual name.

src/wasm.h
sbc100: Good point..  went I wrote this I had assumable there would be other nameable thing but I guess not yet.

We probably want to rename `Importable` to something like `External` (i.e. element that can be imported and/or exports)  to match the spec language and that used in the wasm C API, but I'll leave it for now.

src/wasm/wasm-binary.cpp
sbc100: Sounds good.  I went with `hasExplicitName`

src/wasm.h
```diff
@@ -1335,10 +1335,25 @@ class ArrayLen : public SpecificExpression<Expression::ArrayLenId> {
 // Globals
 
 struct Importable {
+  Name name;
+
+  // Explicit names are ones that we read from the input file and
+  // will be written the name section in the output file.
```
kripken, line 1341:
```suggestion
  // will be written into the "Names" section in the output file.
```

kripken (commented):
Something about "explicit" seems wrong to me. Maybe "debug"? Not great either.

Why was this a problem, though? In what way did this break emscripten?

src/wasm.h
```diff
@@ -1335,10 +1335,25 @@ class ArrayLen : public SpecificExpression<Expression::ArrayLenId> {
 // Globals
 
 struct Importable {
+  Name name;
```
kripken, line 1338: Not optimal that this is on `Importable`, as this is not really a property of an imported thing, semantically.

However, you are using `Importable` essentially as "global thing on the module", and this would make sense there. So maybe the problem is that `Importable` needs to be renamed. But can be separate from this PR of course.

tlively (approved):

src/wasm.h
sbc100: Indeed, see my earlier comment.   The wasm-c-api uses "extern" for this to mean something that can be imported and exports to/from the module.   "extern" and "external" are bit overloaded but I think maybe the spec uses those terms too.. which which case we should rename as a followup.

sbc100:
> Something about "explicit" seems wrong to me. Maybe "debug"? Not great either.
> 
> Why was this a problem, though? In what way did this break emscripten?

My comment tries to explain the choice of term here.    "Explicit" means specified by the user, "Implicit" means created on the fly based on some index or other internal binaryen thing.   I could replace with "External" name vs "Internal" name?

tlively:
FWIW, I like the `Explicit` name.

kripken:
Ok, let's keep it "explicit" then since the consensus is around that, lgtm.

sbc100:
> Why was this a problem, though? In what way did this break emscripten?

The reason this is particularly problematic is that if we persist these names like `0`, `1`, `2`, `fimport$1` etc... then we read them in.. and then inject a new function or re-order the functions we can end up with the third function being named `0`...  which is then ripe for collision if the zero'th function doesn't have a name.. does that make sense?

I guess it would be good to write some tests for this kind of thing.. I'm just not sure how.

sbc100:
OK to land as is and figure out precise tests later?  (since it unblocks CI)

PR #3248, opened 2020-10-15
finalize: move more functionality behind legacyPIC

sbc100:
Internalizing of the stack pointer is only needed in legacy
PIC mode, since in the new PIC mode we support mutable globals.

Also the additional ASSIGN_GOT_ENTRIES function only exists
in support of the legacy mode.


tlively (approved):
I'm not as familiar with the relevant PIC compilation schemes as I would like to be, but this lgtm!

PR #3250, opened 2020-10-16
finalize: add --mutable-sp flag

sbc100:
This flag disables the features of wasm-emscripten-finalize
the replace the mutable global import of `__stack_pointer`.

See the corresponding emscripten change that depends on this
one: https://github.com/emscripten-core/emscripten/pull/12536

### DIFF
--- a/src/tools/wasm-emscripten-finalize.cpp
+++ b/src/tools/wasm-emscripten-finalize.cpp
@@ -52,6 +52,7 @@ int main(int argc, const char* argv[]) {
   bool DWARF = false;
   bool sideModule = false;
   bool legacyPIC = true;
+  bool mutableSP = false;
   bool legalizeJavaScriptFFI = true;
   bool bigInt = false;
   bool checkStackOverflow = false;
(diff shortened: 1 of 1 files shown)


sbc100:
One the emscripten side lands the followup change here is: https://github.com/WebAssembly/binaryen/pull/3251

src/tools/wasm-emscripten-finalize.cpp
```diff
@@ -110,6 +111,13 @@ int main(int argc, const char* argv[]) {
          [&sideModule](Options* o, const std::string& argument) {
            sideModule = true;
          })
+    .add("--mutable-sp",
+         "",
+         "Import SP as a mutable global",
```
tlively, line 116: It would be good to spell out SP as "stack pointer" at least in the help text.

src/tools/wasm-emscripten-finalize.cpp
```diff
@@ -278,7 +286,7 @@ int main(int argc, const char* argv[]) {
   passRunner.setDebug(options.debug);
   passRunner.setDebugInfo(debugInfo);
 
-  if (checkStackOverflow && !sideModule) {
+  if (checkStackOverflow) {
```
tlively, line 289: Can side modules now support stack checks when they couldn't before?

sbc100: Without `--mutable-sp` the side module is forced to manipulate SP indirectly via the `stackSave`/`stackRestore` functions that live in the main module.  

So we go away without running this pass on the side module previously.

src/tools/wasm-emscripten-finalize.cpp
sbc100: Sure, although I don't imagine this option lasting more than a week so maybe this isn't too important.

tlively (approved):

PR #3251, opened 2020-10-16
Remove now-redundant stack pointer manipulation passes

sbc100:
The use of these passes was removed on the emscripten side
in https://github.com/emscripten-core/emscripten/pull/12536.


src/tools/wasm-emscripten-finalize.cpp
```diff
@@ -110,6 +111,13 @@ int main(int argc, const char* argv[]) {
          [&sideModule](Options* o, const std::string& argument) {
            sideModule = true;
          })
+    .add("--mutable-sp",
+         "",
+         "Import SP as a mutable global",
+         Options::Arguments::Zero,
+         [&mutableSP](Options* o, const std::string& argument) {
+           mutableSP = true;
+         })
```
kripken, line 120: This may be in another PR already judging by the title I remember reading?

kripken (approved):

PR #3253, opened 2020-10-16
Fix validateGlobally usage in validator

kripken:
validateGlobally means that we can't do lookups on the module. A few places
were missing that, or had it wrong. I think the reason for the wrong usages is
that we used to have types on the module, and then removed that, so more is
now validatable actually.


kripken:
This uncovered a real bug, where i64-to-32 would ignore an unreachable
parameter of a `call_indirect`. That's bad, since if the type is i64, we need
to replace it with two parameters. To fix that, just handle unreachability
there, using the existing logic (which skips the `call_indirect` entirely in
this case).

aheejin (approved):

PR #3254, opened 2020-10-16
wasm-reduce: When trying to remove a function, try to replace ref.func usages too

kripken:


src/tools/wasm-reduce.cpp
```diff
@@ -937,6 +937,14 @@ struct Reducer
...
         }
       }
+      void visitRefFunc(RefFunc* curr) {
+        if (names.count(curr->func)) {
+          auto& functions = getModule()->functions;
+          if (!functions.empty()) {
+            curr->func = functions[0]->name;
+          }
+        }
+      }
```
aheejin, line 947: - Do we need to check if the replacing function has the same type? We don't have a way to call function references directly, but after [typed function reference proposal](https://github.com/WebAssembly/function-references/blob/master/proposals/function-references/Overview.md) is added later, we would have `call_ref` instruction, then I guess it can matter?
- Is there a possibility that `functions[0]->name == curr->func`?

kripken: Good points, both of those corner cases can happen. In general it's ok if this fails - the specific attempted reduction just won't succeed - but at least the second case seems like it could be common (if after reduction just one function remains). I changed it to do what we do for calls, which is to replace with something trivial like a const, that seems simpler and safer.

aheejin (approved):

PR #3261, opened 2020-10-18
Remove old/non-working SpillPointers pass

sbc100:
And associated stack.h.  The current stack.h clearly doesn't work with
the llvm back as it assumes the stack grows up, which means non of these
has been working or used in a long time.

Rather than trying to fix this unused features its probably cleaner to
just remove it for now and restore it rom git history if its someone
that anyone actually wants to use in the future.


kripken (approved):
lgtm, although I slightly worry we'll need something like this and forget this existed, and end up doing more work. In theory getting it from git is fine, but it is less searchable there.

Maybe mention this in the changelog? Then it would be searchable in the current tip of tree at least.

sbc100:
I guess the alternative would be to keep it around but error if someone uses it?

sbc100:
Added CHANGELOG entry.

PR #3265, opened 2020-10-20
Avoid name collisions in MemoryPacking

kripken:
Such a collision can happen if we run the pass twice, and somehow it finds
more to optimize.

To make this easy, add a general utility for getting a unique name based on
a root + a numeric suffix to avoid collisions.

Fixes the testcase in https://github.com/WebAssembly/binaryen/issues/3225#issuecomment-709395467

### DIFF
--- a/src/ir/names.h
+++ b/src/ir/names.h
@@ -46,6 +46,33 @@ inline void ensureNames(Function* func) {
   }
 }
 
+// Given a root of a name, finds a valid name with perhaps a number appended
+// to it, by calling a function to check if a name is valid.
+inline Name
+getValidName(Module& module, Name root, std::function<bool(Name)> check) {
+  if (check(root)) {
+    return root;
+  }
+  auto prefixed = std::string(root.str) + '_';
+  Index num = 0;
+  while (1) {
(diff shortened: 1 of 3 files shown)


src/ir/names.h
```diff
@@ -46,6 +46,33 @@ inline void ensureNames(Function* func) {
...
+}
+
+inline Name getValidGlobalName(Module& module, Name root) {
+  return getValidName(
+    module, root, [&](Name test) { return !module.getGlobalOrNull(test); });
+}
+inline Name getValidFunctionName(Module& module, Name root) {
+  return getValidName(
+    module, root, [&](Name test) { return !module.getFunctionOrNull(test); });
+}
```
tlively, line 74: Are there any users of `getValidFunctionName`? It would be good to add one or remove this function, possibly with a TODO to add it back with users in the future.

tlively (approved):

src/ir/names.h
kripken: I have one in another PR (the fuzzing with initial testcases). Do you object to landing this as-is?

PR #3267, opened 2020-10-20
SimplifyLocals fuzz fix: Don't be confused by subtype assigns. 

### LINKED ISSUE #3266: Eqref fuzz bug?
```wat
(module
 (export "func_145" (func $0))
 (func $0 (param $0 eqref) (param $1 i31ref) (result anyref)
  (local $2 i32)
  (local $3 anyref)
  (local $4 eqref)
  (local.set $4
   (local.get $0)
  )
  (local.set $0
   (local.get $1)
  )
  (local.set $2
   (ref.eq
    (local.get $4)
    (local.get $0)
   )
  )
  (if
   (local.get $2)
   (unreachable)
  )
  (local.get $3)
 )
)
```

turns into

```wat
(module
 (export "func_145" (func $0))
 (func $0 (param $0 eqref) (param $1 i31ref) (result anyref)
  (local $2 i32)
  (local $3 anyref)
  (local $4 eqref)
  (nop)
  (local.set $0
   (local.get $1)
  )
  (nop)
  (if
   (ref.eq
    (local.get $0)
    (local.get $0)
   )
   (unreachable)
  )
  (local.get $3)
 )
)
```

by `--simplify-locals-notee`. `--fuzz-exec` finds a different result in them. The first returns `anyref(null)` and the second hits that `unreachable`.

What seems to happen is that in the first case we do `ref.eq` on an `eqref` and on an `i31ref`. Both are null, and we get that they are not equal. Is that correct?

In the second case, we end up comparing the `i31ref` to itself, and find they are equal.

(edit: fuzz seed `11824809954583288331` on `initialfuzz` branch)

kripken:
We checked if the type matches when deciding if two locals are equivalent,
but if the type didn't match, we forgot to reset any previously equivalent
things. So we thought something was equivalent when it wasn't, see the
reduced testcase.

Fixes #3266

### DIFF
--- a/src/passes/SimplifyLocals.cpp
+++ b/src/passes/SimplifyLocals.cpp
@@ -964,16 +964,23 @@ struct SimplifyLocals
               }
               anotherCycle = true;
             }
+            // Nothing more to do, ignore the copy.
+            return;
           } else if (func->getLocalType(curr->index) ==
                      func->getLocalType(get->index)) {
-            // There is a new equivalence now.
+            // There is a new equivalence now. Remove all the old ones, and add
+            // the new one.
+            // Note that we ignore the case of subtyping here, to keep this
+            // optimization simple by assuming all equivalent indexes also have
+            // the same type. TODO: consider optimizing this.
             equivalences.reset(curr->index);
             equivalences.add(curr->index, get->index);
+            return;
           }
-        } else {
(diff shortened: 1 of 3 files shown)


aheejin (approved):
Ah, sorry, this appears to be my bug ;( Thanks for fixing it!

PR #3268, opened 2020-10-20
Fuzzer: improve random selection of features

kripken:
We used to either apply all, or pick each at random. Also add a chance
to pick none at all.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -106,7 +106,7 @@ def randomize_feature_opts():
     # coverage both ways, and 1/3 pick each one randomly
     if random.random() < 0.33333:
         FEATURE_OPTS += POSSIBLE_FEATURE_OPTS
-    elif random.random() < 0.33333:
+    elif random.random() < 0.5:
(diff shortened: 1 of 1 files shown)


scripts/fuzz_opt.py
```diff
@@ -102,12 +102,11 @@ def no_pass_debug():
...
-    # half the time apply all the possible opts. this lets all test runners work at max
-    # capacity at least half the time, as otherwise if they need almost all the opts, the
-    # chance of getting them is exponentially small.
-    if random.random() < 0.5:
+    # 1/3 the time apply all the possible opts, 1/3 none of them, to maximize
+    # coverage both ways, and 1/3 pick each one randomly
+    if random.random() < 0.33333:
         FEATURE_OPTS += POSSIBLE_FEATURE_OPTS
-    else:
+    elif random.random() < 0.33333:
```
aheejin, line 109: This sounds like not 1/3? Isn't this 2/3 * 1/3, given that this is within `elif`?

kripken: Oh, heh, yes, how silly of me...

tlively (approved):

PR #3270, opened 2020-10-20
Fuzzer: Handle the case where we can't even generate the wasm file

kripken:
If we can't run the `-ttf` stage, there is no point in printing out the
instructions to reduce things - we can't reduce without a wasm.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -952,7 +952,9 @@ if __name__ == '__main__':
             # longer working on the original test case but modified one, which
             # is likely to be called within wasm-reduce script itself, so
             # original.wasm and reduce.sh should not be overwritten.
-            if not given_wasm:
+            # Note that we can't do this if a.wasm doesn't exist, which can be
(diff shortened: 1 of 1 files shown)


aheejin (approved):

scripts/fuzz_opt.py
```diff
@@ -952,7 +952,9 @@ def randomize_opt_flags():
             # longer working on the original test case but modified one, which
             # is likely to be called within wasm-reduce script itself, so
             # original.wasm and reduce.sh should not be overwritten.
-            if not given_wasm:
+            # Note that we can't do this if a.wasm doesn't exist, which can be
+            # the case if we failed to even generate the wasm.
+            if not given_wasm and os.path.exists('a.wasm'):
```
aheejin, line 957: It might be better to print a warning that wasm generation failed so we don't have a reduced test case?

kripken: Good idea, added!

tlively (approved):

PR #3271, opened 2020-10-20
Fuzzer: Handle log_execution instrumentation, and warn on unknown imports

kripken:
`log_execution` can be useful to run on a fuzz testcase for debugging purposes. This
adds support to `--fuzz-exec` for printing out those values.

In general `--fuzz-exec` ignores unknown imports (so that it can be run on more wasm
files, for at least some fuzz testing there). This also adds a warning in that case, so it
is less surprising (the behavior does not change in this PR).

### DIFF
--- a/src/tools/execution-results.h
+++ b/src/tools/execution-results.h
@@ -40,7 +40,19 @@ struct LoggingExternalInterface : public ShellExternalInterface {
         loggings.push_back(argument);
       }
       std::cout << "]\n";
+      return {};
+    } else if (import->module == ENV) {
+      if (import->base == "log_execution") {
+        std::cout << "[LoggingExternalInterface log-execution";
+        for (auto argument : arguments) {
+          std::cout << ' ' << argument;
+        }
+        std::cout << "]\n";
+        return {};
+      }
     }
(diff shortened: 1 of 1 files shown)


aheejin (approved):

tlively:
What's the difference between the new log_execution import and the previous fuzzing-support imports?

kripken:
@tlively This isn't a new import - the PR adds support to the fuzzer for the existing import.

(As always the `log_execution` import receives a number, and should log it out, to indicate an interesting point in a program.)

tlively:
I guess I'm just blanking on where else `log_execution` is used. Anyway, this LGTM.

kripken:
@tlively Oh, sorry, I should have said: the `LogExecution` pass will add this instrumentation (at function entry/exit, and loop headers, etc.). One of the family of instrumention-adding passes, others of which instrument local operations, or memory operations, etc.

IOW, this PR allows the following debugging flow: you run `wasm-opt --log-execution` on the wasm, then run it in the fuzzer, `wasm-opt --fuzz-exec[-before]` and you get that logging.

tlively:
Oh neat! I didn't know about that pass. That sounds super useful.

PR #3278, opened 2020-10-22
Implement v128.{load,store}{8,16,32,64}_lane instructions

tlively:
These instructions are proposed in https://github.com/WebAssembly/simd/pull/350.
This PR implements them throughout Binaryen except in the C/JS APIs and in the
fuzzer, where it leaves TODOs instead. Right now these instructions are just
being implemented for prototyping so adding them to the APIs isn't critical and
they aren't generally available to be fuzzed in Wasm engines.


src/wasm-builder.h
```diff
@@ -434,6 +434,21 @@ class Builder {
...
+                                           uint8_t index,
+                                           Expression* ptr,
+                                           Expression* vec) {
+    auto* ret = wasm.allocator.alloc<SIMDLoadStoreLane>();
+    ret->op = op;
+    ret->offset = offset;
+    ret->align = align;
+    ret->index = index;
+    ret->ptr = ptr;
+    ret->finalize();
```
kripken, line 449: It looks like this does not assign `vec`?

(I wonder if an automatic code generation tool could help in the future here... :thinking:)

src/wasm/wasm-s-parser.cpp
```diff
@@ -1302,6 +1302,9 @@ static size_t parseMemAttributes(Element& s,
   align = fallbackAlign;
   while (!s[i]->isList()) {
     const char* str = s[i]->c_str();
+    if (strncmp(str, "align", 5) != 0 && strncmp(str, "offset", 6) != 0) {
```
kripken, line 1305: I'm confused, what does this do?

src/wasm-builder.h
tlively: Oops! Yes, it certainly would.

src/wasm/wasm-s-parser.cpp
tlively: These instructions are the first to have a memarg (align and offset) followed by another immediate argument. This parsing routine previously expected the align and offset to be the only non-expression operands (see the loop condition), so it wasn't able to parse the new instructions correctly. This addition makes the loop bail out instead of raise an error when it encounters a non-memarg immediate argument.

kripken: I see, thanks. Maybe a comment on the function (saying what it parses, and that it bails on anything else)?

kripken (approved):

PR #3282, opened 2020-10-23
Make pops valid in tests

### LINKED ISSUE #3213: pop()s in test/passes/instrument-locals_all-features.wast are not written correctly to binary
STR:

```
wasm-opt test/passes/instrument-locals_all-features.wast -o a.wasm
wasm-opt a.wasm
```

error:
```
[parse exception: attempted pop from empty stack / beyond block start boundary at 161 (at 0:161)]
```

aheejin:
These two tests had pops in invalid locations; pops are valid only after
`catch`. This fixes those invalid wasm files.
This removes pops from Os_print-stack-ir_all-features.wast too.

Fixes #3213 and #3283.


test/passes/remove-unused-names_code-folding_all-features.wast
```diff
@@ -1273,7 +1273,6 @@
       (try
         (do
           ;; Expressions that can throw should NOT be taken out of 'try' scope.
-          (local.set $exn (pop exnref))
```
aheejin, line 1276: I'm not sure why we had this; I think this was a copy-paste error.

tlively (approved):
LGTM, but for what it's worth, I would also be fine just deleting these parts of the tests. They are from back when I originally added Pop with the idea that it could be used anywhere (along with Push, which we've since removed).

kripken (approved):
lgtm.

Removing them would be even better, but I'm not sure myself if these don't help somehow. If you you think it's ok to remove then sgtm.

PR #3284, opened 2020-10-23
[NFC] `using namespace Abstract` to make matchers more compact

tlively:
This change makes matchers in OptimizeInstructions more compact and readable by
removing the explicit `Abstract::` namespace from individual operations. In some
cases, this makes multi-line matcher expressions fit on a single line.

This change is only possible because it also adds an explicit "RMW" prefix to
each element of the `AtomicRMWOp` enumeration. Without that, their names
conflicted with the names of Abstract ops.

cc @MaxGraey this will help make matchers more readable in your PRs.


src/passes/OptimizeInstructions.cpp
```diff
@@ -221,6 +221,7 @@ struct OptimizeInstructions
       // `using namespace Match` can be hoisted to function scope and this extra
       // block scope can be removed.
       using namespace Match;
+      using namespace Abstract;
```
aheejin, line 224: Would using it once and for all at the top of the file be problematic?

aheejin (approved):

kripken (approved):

MaxGraey (approved):

src/passes/OptimizeInstructions.cpp
tlively: Probably not, but I do want to tie this change specifically to the use of matchers, since I don't think removing the explicit `Abstract::` prefix is a readability win in other cases.

PR #3286, opened 2020-10-27
Implement i8x16.popcnt

tlively:
As proposed in https://github.com/WebAssembly/simd/pull/379. Since this
instruction is still being evaluated for inclusion in the SIMD proposal, this PR
does not add support for it to the C/JS APIs or to the fuzzer. This PR also
performs a drive-by fix for unrelated instructions in c-api-kitchen-sink.c


src/wasm/literal.cpp
```diff
@@ -1653,6 +1653,9 @@ Literal Literal::absI32x4() const {
 Literal Literal::negI8x16() const {
   return unary<16, &Literal::getLanesUI8x16, &Literal::neg>(*this);
 }
+Literal Literal::popcntI8x16() const {
+  return unary<16, &Literal::getLanesUI8x16, &Literal::popCount>(*this);
```
aheejin, line 1657: Is each lane signed or unsigned? The LLVM builtin [says](https://github.com/llvm/llvm-project/blob/6335446c99eb7263b8ae89a071df8a2cd66940d5/clang/include/clang/Basic/BuiltinsWebAssembly.def#L117) it's signed though

aheejin (approved):

kripken (approved):

src/wasm/literal.cpp
tlively: Signedness doesn't really make a difference for popcnt since it doesn't interpret the bits as a number. However, it's important to use the unsigned lane extraction function here because it will zero-extend each lane. If we instead used the signed lane extraction function, it would sign-extend each lane which could change the population counts.

Issue #3039, opened 2020-08-12
Can't build with MinGW?


phated:
Recently, I've been trying to build Binaryen with MinGW to link into Grain's OCaml codebase, but the build fails halfway through with something like:
```
D:/a/binaryen.ml/binaryen.ml/src/binaryen/third_party/llvm-project/include/llvm/Object/SymbolicFile.h: In function 'OStream& llvm::object::operator<<(OStream&, const llvm::object::DataRefImpl&)':
D:/a/binaryen.ml/binaryen.ml/src/binaryen/third_party/llvm-project/include/llvm/Object/SymbolicFile.h:47:30: error: expected ')' before 'PRIxPTR'
   OS << "(" << format("0x%08" PRIxPTR, D.p) << " (" << format("0x%08x", D.d.a)
                      ~       ^~~~~~~~
                              )
```

Should Binaryen build with MinGW? I noticed some commits from 2016-2017 that mentioned support for MinGW, but it seems like it was removed from the CI matrix. Maybe I'm doing something wrong.

gerald-dotcom:
Hit the same. 

Just edit cmake file and add following

add_compile_flag("-D__STDC_FORMAT_MACROS")

phated:
@gerald-dotcom whoa! Thank you! Where exactly did you add that?

gerald-dotcom:
On your text editor, Search for "add_compile_flag" and you will find a function named add_compile_flag then put this code under the function (Not inside).

phated:
Fixed by #3287 - There is further work to be done to get the tests working, but at least it builds now.

PR #3287, opened 2020-10-27
Fix MinGW build & add to GitHub Actions

phated:
This PR adds a failing MinGW build to the CI (as I mentioned in #3039) and then a separate PR that adds the recommended solution from the comments.

I was reading in the `check.py` that Windows + gcc tests need additional work, so I didn't enable tests for the MinGW build, but it at least builds successfully now.

I'm hoping this helps our integration with Binaryen into Grain.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,3 +205,19 @@ jobs:
       run: |
         source $HOME/emsdk/emsdk_env.sh
         ./scripts/emcc-tests.sh
+
+  # Windows + gcc needs work before the tests will run, so just test the compile
+  build-mingw:
+    name: mingw
+    runs-on: windows-latest
+    steps:
+    - uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
+    - uses: actions/checkout@v1
+    - name: cmake
+      run: |
+        mkdir -p out
+        cmake -S . -B out -G "MSYS Makefiles"
+    - name: build
+      run: cmake --build out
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -222,6 +222,7 @@ else()
   add_compile_flag("-Wswitch") # we explicitly expect this in the code
   if(WIN32)
     add_compile_flag("-D_GNU_SOURCE")
+    add_compile_flag("-D__STDC_FORMAT_MACROS")
```
sbc100, line 225: Why this this needed only on win32?

.github/workflows/ci.yml
```diff
@@ -205,3 +205,19 @@ jobs:
...
+    - uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
+    - uses: actions/checkout@v1
+    - name: cmake
+      run: |
+        mkdir -p out
+        cmake -S . -B out -G "MSYS Makefiles"
+    - name: build
+      run: cmake --build out
```
sbc100, line 223: Shouldn't this run in the mingw shell?

phated: The MSYS Makefiles seem to pick up the correct GNU compiler. I tested using `shell: bash` and there was no difference.

CMakeLists.txt
phated: I have no idea. I just know that it won't build in MinGW on Windows and it successfully builds on all other platforms. I just took the suggestion from the issue and applied it to only Windows + MinGW.

sbc100 (approved):

sbc100:
Can you make the PR description a little more descriptive?

phated:
Done! Let me know if you'd like me to put something else.

PR #3264, opened 2020-10-19
Autogenerate boilerplate code for expressions from a single tool

kripken:
This proposes a new way of developing Binaryen: Instead of hand-writing the
boilerplate for expressions - like the logic for walking, comparing, etc. - we
have a single declaration of each expression, and a script that generates
the code.

The goal is to reduce the hand-written boilerplate code, mainly to make it
easier to work on Binaryen - reduce the time to add new instructions, or
change them - and to reduce the risk for bugs in those places. A minor
goal is to also improve the speed of the code.

This PR implements the declaration and emitting in Python. This can't
be done in C++ itself - in theory macros or templates can get close, but
we want a lot more power here than macros provide even in other
languages (AFAIK - but there is lisp...). Fundamentally we want the power to
process the declarations of expressions and to generate something
pretty arbitrary from that, using a full programming language. A possible
future use of this, for example, is to automatically reorder our instructions
in some optimal way, so that it's a single comparison to check if something
is control flow structure or not; or to reorder (reorderable) fields on an
expression; and more complex things are possible.

Example of what this PR does: We declare `call` once,

```python
class Call(Expression):
    operands = ChildList()
    target = Name()
    isReturn = Bool(init='false')
```

This reads as: Call is an expression. It has a field "operands" which is
a list of child expressions. It has a field "target" which is a name (the
called function). And it has a field "isReturn" which is a bool that has
an initial value.

The declarations are all ported from the existing header, with initial
values and comments and so forth all copied over, so this should be
NFC (hopefully!)

We then run the tool and it autogenerates the C++ header:

```cpp
class Call : public SpecificExpression<Expression::CallId> {
public:
  Call(MixedArena& allocator) : operands(allocator) {}
  ExpressionList operands;
  Name target;
  bool isReturn = false;
  void finalize();
};
```

And also it autogenerates C++ code to compare calls:

```cpp
case Expression::CallId: {
  auto* castLeft = left->cast<Call>();
  auto* castRight = right->cast<Call>();
  if (castLeft->operands.size() != castRight->operands.size()) {
    return false;
  }
  for (auto* child : castLeft->operands) {
    leftStack.push_back(child);
  }
  for (auto* child : castRight->operands) {
    rightStack.push_back(child);
  }
  if (castLeft->target != castRight->target) {
    return false;
  }
  if (castLeft->isReturn != castRight->isReturn) {
    return false;
  }
  break;
}
```

That code shows how this can prevent bugs: the tool won't forget a field
or anything like that.

[...]

PR #3290, opened 2020-10-27
Reduce boilerplate with a C macro for class delegations. NFC

kripken:
Split off from #3264, this part is hopefully non-controversial.


tlively (approved):
Strongly LGTM 🥇

PR #3294, opened 2020-10-28
Avoid boilerplate code with a big universal delegation header

kripken:
This is an alternative approach to #3264 . Instead of autogenerating code using a
python script, and checking those in, this creates a single "universal" delegation
mechanism for all classes and all fields, that everything can be built on top of.

To get an idea, this is a sample from the header:
```cpp
switch (DELEGATE_ID) {
[..]
  case Expression::Id::BlockId: {
    DELEGATE_START(Block);
    DELEGATE_FIELD_CHILD_LIST(Block, list);
    DELEGATE_FIELD_SCOPE_NAME(Block, name);
    DELEGATE_END(Block);
    break;
  }
```
A switch is defined (on some `DELEGATE_ID` that is provided). There is then
a case for each expression type, calling a delegate for each field type. This allows
a user to define the delegates it wants, then include this header.

This can reduce boilerplate similarly to #3264, but has different tradeoffs. Personally
I prefer the python script, as working with C macros like this can be annoying - error
messages are not great, for one - you can build with `-E` to see the preprocessor
output, which I had to do to get this example working, but it's not a great experience.
The python script approach checks in the emitted code as simple cpp, so errors are
clearer. However, a big benefit to this approach is that it just uses C++ features, so
there is no python, no build system changes, etc.

Thoughts?

Side note, this found that we were missing `BrOnExn::sent` scanning in the
existing hashing/comparing code (which is what I used as a basis) - if this doesn't
land, we should land that piece separately. This was noticeable because I
used the new header in a rewrite of the copying logic (where missing a field
is more obvious).

Side note 2, the rewrite of the copying logic now avoids recursion.


tlively:
This could be made safer by putting the giant switch into a templated function that takes a separate template parameter for each of `DELEGATE_START`, `DELEGATE_FIELD_*`, etc. instead of using raw macros, The benefit would be that it would be impossible to accidentally forget to handle a certain kind of field, but the downside would be that it couldn't be used for declarations such as in the C API. It might be useful to have both, building the template version on top of the macro version.

kripken:
That's an option too. The reason that made me write it this way is that with a C macro you can refer to locals in the current scope, which is very convenient. (That wouldn't be an issue in e.g. JS...) But yes, good point that we could build a safer function version from the macro one, and use that wherever possible.

tlively:
Any state would have to be hoisted into the calling context and closed over by lambdas used as the delegate template parameters, which I don't think would be so bad either.

kripken:
@tlively Hmm, I did some work along those lines now, but personally it worries me. Consider this:
```cpp
DELEGATE_FIELD_CHILD_LIST(Block, list);
```
If we want to replace that with a template that expands to something like
```cpp
copyBlock->list = originalBlock->list;
```
then we need to use something like this:
```cpp
template<typename Class>
class Handler {
  Class* copy;
  Class* original;

  template<typename Field>
  void handleChildList(Field field) {
    copy->*field = original->*field;
  }
};
```
This is more C++ trickery than I would like, with two layers of templates and a pointer to a class member that the other side would initialize as `&Block::list`. I'm also not 100% sure all compilers would fully optimize that field access. The corresponding C macro as in this PR would be
```cpp
#define DELEGATE_START(Class) \
  Class* copy; \
  Class* original;
#define DELEGATE_FIELD_CHILD_LIST(Class, name) \
  copy->name = original->name;
```
Maybe just a matter of personal taste, of course. I still prefer the Python in the other PR most :wink:

aheejin:
Maybe I'm mistaken, but if we have something like `FieldIterator`, do we even need to delegate each expression case, such as `Block`, `Unary`, `Break`, ...? Can we have a unified routine to copy/hash all of those in the same manner?

kripken:
@aheejin A FieldIterator would be defined using this. The idea is that this would be the single place for writing out the list of fields, and then it could be used to create a FieldIterator, a walker (which just looks at child fields), and so forth.

Also, this differs from the previously existing ChildIterator, and a possible FieldIterator that is similar to it, in performance: the ChildIterator uses walking under the hood, first doing a walk and saving children to a vector, and then iterating on the vector. This PR would allow writing something like a walker, comparer, hasher, etc. without going through a vector in the middle (unless necessary). The copying logic in this PR shows an example of that.

aardappel:
Thanks for at least trying this out!

It's.. a macro visitor :) I hadn't seen that before. It has some advantages though, being very general, and likely collapsing to very efficient code, especially in cases where not all macros/nodes are needed, compared to a visitor that uses dynamic dispatch.

And personally I'd rather debug errors caused by this than ones from an external Python program.

I would personally attempt to do it all with templates, but I can understand if that's not your cup of tea.

kripken:
@aardappel I'm still open to the idea of templates. The best I can think of there is mentioned in the previous comment, https://github.com/WebAssembly/binaryen/pull/3294#issuecomment-718307324

kripken:
Friendly review ping, if you have time @aheejin / @aardappel / @tlively

src/wasm-delegations-fields.h
```diff
@@ -0,0 +1,582 @@
...
+// are optional, so you can just provide what you want, and no code will be
+// emitted for the others.
+//
+// (The only mandatory thing to define is DELEGATE_ID which is the key for the
+// switch.)
+//
+// All #defines used here are undefed automatically at the end for you.
+//
+// Child pointers are emitted in reverse order (which is convenient for walking
+// by pushing them to a stack first).
```
aheejin, line 28: Does this order matter? I don't think it matters in copying, but I guess you plan to use this in some other place that matters..?

src/ir/ExpressionManipulator.cpp
```diff
@@ -23,306 +23,90 @@ namespace ExpressionManipulator {
...
+  castCopy->name.resize(castOriginal->name.size());                            \
+  for (Index i = 0; i < castOriginal->name.size(); i++) {                      \
+    castCopy->name[i] = castOriginal->name[i];                                 \
+  }
+
+#define DELEGATE_FIELD_SIGNATURE(id, name) castCopy->name = castOriginal->name;
+
+#define DELEGATE_FIELD_TYPE(id, name) castCopy->name = castOriginal->name;
+
+#define DELEGATE_FIELD_ADDRESS(id, name) castCopy->name = castOriginal->name;
```
aheejin, line 99: It looks we can reduce the number of `define`s here by doing something like
```cpp
#define COPY_FIELD(id, name) castCopy->name = castOriginal->name;

#define COPY_FIELD_LIST(id, name)                                              \
  for (Index i = 0; i < castOriginal->name.size(); i++) {                      \
    castCopy->name[i] = castOriginal->name[i];                                 \
  }

#define DELEGATE_FIELD_INT COPY_FIELD
#define DELEGATE_FIELD_INT_ARRAY(id, name)                                     \
  assert(castCopy->name.size() == castOriginal->name.size());                  \
  COPY_FIELD_LIST(id, name)
#define DELEGATE_FIELD_LITERAL COPY_FIELD
#define DELEGATE_FIELD_NAME COPY_FIELD
#define DELEGATE_FIELD_SCOPE_NAME COPY_FIELD
#define DELEGATE_FIELD_SCOPE_NAME_LIST(id, name)                               \
  castCopy->name.resize(castOriginal->name.size());                            \
  COPY_FIELD_LIST(id, name)
#define DELEGATE_FIELD_SIGNATURE COPY_FIELD
#define DELEGATE_FIELD_TYPE COPY_FIELD
#define DELEGATE_FIELD_ADDRESS COPY_FIELD
```

aheejin (commented):
Sorry for the delayed reply. I think this direction looks good after all 🎉

src/ir/ExpressionManipulator.cpp
kripken: Good idea, thanks!

I also renamed the variable-sized one "vector", to match the fixed-size one which is "array", which is hopefully more like C++ and less confusing.

src/wasm-delegations-fields.h
kripken: Yes, it will matter in walking, but does not matter yet, in copying.

src/wasm-delegations-fields.h
```diff
@@ -0,0 +1,582 @@
...
+    DELEGATE_START(Try);
+    DELEGATE_FIELD_CHILD(Try, catchBody);
+    DELEGATE_FIELD_CHILD(Try, body);
+    DELEGATE_END();
+    break;
+  }
+  case Expression::Id::ThrowId: {
+    DELEGATE_START(Throw);
+    DELEGATE_FIELD_CHILD_LIST(Throw, operands);
+    DELEGATE_FIELD_INT(Throw, event);
```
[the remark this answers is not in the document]
kripken: Good catch!

aheejin (approved):

src/wasm-delegations-fields.h
```diff
@@ -0,0 +1,584 @@
...
+// switch.)
+//
+// All #defines used here are undefed automatically at the end for you.
+//
+// Child pointers are emitted in reverse order (which is convenient for walking
+// by pushing them to a stack first).
+
+// Emits code at the start of the case for a class.
+#ifndef DELEGATE_START
+#define DELEGATE_START(id)
```
tlively, line 32: It might be safer not to provide default empty definitions for the delegates, since the default definitions make it possible to forget to handle a particular field type. Do you think the inconvenience would be worth the safety?

tlively (commented):
Besides that safety question, LGTM!

src/wasm-delegations-fields.h
kripken: That's a good point. Yeah, I'll make these error in a followup, as it's better to force people to write out an empty handler (there aren't that many) than to forget one, or have a typo etc.

PR #3296, opened 2020-10-28
Use RAII in fuzzer for function context creation

kripken:
Followup to #3276 where this was suggested by @tlively


tlively (commented):
Would it be worthwhile to move the state (e.g. `func`, `labelIndex`, etc.) from the parent into the `FunctionCreationContext` to make it impossible to forget to create a context as well? I guess the downside would be potentially having to thread the context through a bunch of function calls as an argument.

kripken:
Fair point, it's probably worth the extra refactoring. I did that now.

tlively (approved):
👍

PR #3302, opened 2020-10-29
[Memory64]

aardappel:
Fixed bug in memory64-lowering pass for memory.size/grow

### DIFF
--- a/src/passes/Memory64Lowering.cpp
+++ b/src/passes/Memory64Lowering.cpp
@@ -35,7 +35,7 @@ struct Memory64Lowering : public WalkerPass<PostWalker<Memory64Lowering>> {
     }
   }
 
-  void WrapAddress64(Expression*& ptr) {
+  void wrapAddress64(Expression*& ptr) {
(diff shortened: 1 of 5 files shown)


src/passes/Memory64Lowering.cpp
```diff
@@ -46,20 +46,34 @@ struct Memory64Lowering : public WalkerPass<PostWalker<Memory64Lowering>> {
     ptr = builder.makeUnary(UnaryOp::WrapInt64, ptr);
   }
 
+  void ExtendAddress64(Expression*& ptr) {
```
kripken, line 49:
```suggestion
  void extendAddress64(Expression*& ptr) {
```
Sorry, I see I missed this in a past review in WrapAddress64, which should be `wrapAddress64` as well.

kripken (approved):
lgtm with naming convention fixes

src/passes/Memory64Lowering.cpp
aardappel: renamed both

PR #3306, opened 2020-10-30
MemoryPacking: Properly notice zeroFilledMemory

kripken:
We can only pack memory if we know it is zero-filled before us.


src/passes/MemoryPacking.cpp
```diff
@@ -182,7 +178,21 @@ void MemoryPacking::run(PassRunner* runner, Module* module) {
...
+  if (!memory.exists) {
+    return false;
+  }
+
+  // We must optimize under the assumption that memory has been initialized to
+  // zero. That is the case for a memory declared in the module, but for a
+  // memory that is imported, we must be told that it is zero-initialized.
+  if (memory.imported() && !passOptions.zeroFilledMemory) {
+    return false;
+  }
```
tlively, line 192: This is slightly pessimistic. For example, all passive segments can still be optimized if the memory is imported since their zeroes would still be written out by memory.fill instructions. This might make a difference in practice, too, since many thread-enabled modules will import their memories and use primarily passive segments. Could we instead have a separate `canOptimizeActiveSegments` variable that gets set to false in this case?

kripken: I don't think this would make a difference in practice right now, as the `zero-filled-memory` option allows us to keep optimizing in that case.

But it sounds good to do that as well. Would we then need to do more refactoring to split up the passive and active segments? You know this code best.

tlively: Actually, this is pretty simple. You would just have to move this logic from `MemoryPacking::canOptimize` to `MemoryPacking::canSplit`, where it can be applied on a per-segment basis to active segments only.

tlively: I guess this could apply to a bunch of the logic that's currently in `canOptimize`, but that would be a less-related change. I can handle that as a follow-up if you'd like. In that case, this looks good to land as-is.

kripken: Sounds good to me to keep this PR simple and land it as is, and leave that for a followup. It wouldn't be an urgent followup since there is no observable effect in his PR currently anyhow.

tlively (approved):

PR #3311, opened 2020-11-01
Fix usage of `allocate` in JS (no longer takes a type argument)

dcodeIO:
During debugging of a related problem using a debug build of binaryen.js, I was hitting the following error:

```
Error: abort(Error: abort(Assertion failed: allocate no longer takes a type argument)
    at jsStackTrace (path\to\binaryen.js:1279776:19)
    at stackTrace (path\to\binaryen.js:1279793:16)
    at abort (path\to\binaryen.js:1279488:44)
    at assert (path\to\binaryen.js:1278645:5)
    at allocate (path\to\binaryen.js:1278734:3)
    at strToStack (path\to\binaryen.js:1287254:16)
```

triggered [around here](https://github.com/emscripten-core/emscripten/blob/437266ac3dfd61747e53bc6e4ae57b81c92dc093/src/preamble.js#L212-L217), introduced in https://github.com/emscripten-core/emscripten/pull/12279.

Is this the right fix?

### DIFF
--- a/src/js/binaryen.js-post.js
+++ b/src/js/binaryen.js-post.js
@@ -9,7 +9,7 @@ function preserveStack(func) {
 }
 
 function strToStack(str) {
-  return str ? allocate(intArrayFromString(str), 'i8', ALLOC_STACK) : 0;
+  return str ? allocate(intArrayFromString(str), ALLOC_STACK) : 0;
 }
 
 function i32sToStack(i32s) {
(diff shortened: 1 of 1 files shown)


dcodeIO:
Background: Compiling AS with the full runtime yielded a very strange abort on `Module::addExport` with an empty name previously, a fairly new problem only surfacing say a month ago or so, and I ultimately traced the problem back here. With the change, the AS side of things works as expected.

kripken (approved):
Yes, this is the right fix, the type was removed.

There are assertions for this, but I guess we don't run tests on a non-optimized build? Maybe we should.

PR #3323, opened 2020-11-05
wasm2js: Support for exported memory

sbc100:
This change is required for https://github.com/emscripten-core/emscripten/pull/12323


sbc100:
I think this good to land now.

src/wasm2js.h
```diff
@@ -2445,6 +2459,18 @@ void Wasm2JSGlue::emitPreES6() {
 }
 
 void Wasm2JSGlue::emitPost() {
+  // Create a helper bufferView to access the buffer if we need one. We use it
+  // for creating memory segments if we have any (we may not if the segments are
+  // shipped in a side .mem file, for example), and also in bulk memory
+  // operations.
+  // This will get assigned during `asmFunc` (and potentailly re-assiged
```
kripken, line 2466:
```suggestion
  // This will get assigned during `asmFunc` (and potentially re-assigned
```

src/wasm2js.h
```diff
@@ -2445,6 +2459,18 @@ void Wasm2JSGlue::emitPreES6() {
...
+  // during __wasm_memory_grow
```
kripken, line 2467:
```suggestion
  // during __wasm_memory_grow).
```

src/wasm2js.h
```diff
@@ -503,6 +512,10 @@ void Wasm2JSBuilder::addBasics(Ref ast, Module* wasm) {
     addHeap(HEAPU32, UINT32ARRAY);
     addHeap(HEAPF32, FLOAT32ARRAY);
     addHeap(HEAPF64, FLOAT64ARRAY);
+    ast->push_back(
+      ValueBuilder::makeBinary(ValueBuilder::makeName("bufferView"),
+                               SET,
+                               ValueBuilder::makeName(HEAPU8)));
```
kripken, line 518: this creation of `bufferView` looks unconditional - whenever there is a memory at all. why is another one needed elsewhere?

test/wasm2js/atomic_fence.2asm.js
```diff
@@ -55,10 +57,7 @@ function asmFunc(env, buffer) {
  };
 }
 
-var memasmFunc = new ArrayBuffer(1507328);
-var bufferView = new Uint8Array(memasmFunc);
+var bufferView;
```
kripken, line 60: can we not remove this `bufferView` entirely, and just use `HEAP8` which is guaranteed to exist?

sbc100: See my `TODO: We should probably just share a single HEAPU8 var.` above. 

I was thinking that could be a followup.     Keeping this change slightly more limited.

kripken: This does make it harder to review, though, see the comment above with `unconditional` - it looks like the current PR is not emitting `bufferView` properly, but I may be missing something?

sbc100: If we start to share HEAPU8 with the containing scope is that going to interfere with the emscripten-side that also defines a HEAPU8 itself?   I mean that should always point to the same thing anyway so maybe its fine?

src/wasm2js.h
sbc100: This is not a creation of anything but an assignment to the outer scoped `bufferView`

But you are right I think it needs to be conditional in the same way that bufferView itself is conditionally declared in the outer scope.

test/wasm2js/atomic_fence.2asm.js
sbc100: Yes it looks like the wasm2js output shares a scope with the rest of emscripten, but `asmFunc` currently creates it own function-local versions of HEAP8, etc.   This means that in emscripten output there are two copies of all these views today.   

We should possibly merge those two and (i.e. have the output HEAP8 be the same as the inner HEAP8), but I'd rather not worry about that as part of this change.

src/wasm2js.h
kripken: Oh, I see, I was misreading it, sorry.

kripken (approved):
This is fairly confusing (not because of you, but because the wasm2js output is complex with the different modes) but reading the outputs I've convinced myself it is correct.

PR #3326, opened 2020-11-05
Remove OptimizeCalls from PostEmscripten. 

sbc100:
We no longer build modules that import `global.Math`.


tlively (approved):
Code looks good (yay deleting things!), but I wouldn't call this NFC. Why don't we need this optimization, again?

sbc100:
emscripten-built modules no longer use any `Math_*` imports so there is nothing to optimize.

sbc100:
Removed NFC.

sbc100:
@kripken can you take a look at the changes to test input/output.. is that first test case still relevant at all?

kripken (approved):
lgtm but not sure what you are asking in

> is that first test case still relevant at all?

Do you mean the entire first test file with a diff here? The removed parts look good to remove, and the remaining ones look good to remain (like the invoke optimization).

PR #3331, opened 2020-11-09
build/linux: more correct rpath handling

Akaricchi:
Don't hardcode `lib`, it's incorrect for many systems (e.g. `lib64` is common on some x86_64 distros like openSUSE). Use `CMAKE_INSTALL_LIBDIR` instead.

The `APPLE` branch may need the same change, but I don't have any apple systems to confirm this, so I left it as-is.


kripken (approved):
Thanks!

PR #3337, opened 2020-11-11
wasm2js: Declare data segments before calling asmFunc

sbc100:
This is because we maybe need to reference the segments
during the start function.  For example in the case of
pthreads we conditionally load passive segments during
start.

Tested in emscripten with: tests/runner.py wasm2js1


src/wasm2js.h
```diff
@@ -92,6 +92,39 @@ bool isTableExported(Module& wasm) {
   return false;
 }
 
+bool hasActiveSegments(Module& wasm) {
+  // If there are any active segments, initActiveSegments needs access
```
kripken, line 96:
```suggestion
```

src/wasm2js.h
```diff
@@ -92,6 +92,39 @@ bool isTableExported(Module& wasm) {
...
+  // to bufferView.
```
kripken, line 97:
```suggestion
```

src/wasm2js.h
kripken: (comment is duplicated from below, and not needed here)

src/wasm2js.h
```diff
@@ -512,11 +545,23 @@ void Wasm2JSBuilder::addBasics(Ref ast, Module* wasm) {
...
-    if ((!wasm->memory.segments.empty()) || wasm->features.hasBulkMemory()) {
+    if (needsBufferView(*wasm)) {
       ast->push_back(
         ValueBuilder::makeBinary(ValueBuilder::makeName("bufferView"),
                                  SET,
                                  ValueBuilder::makeName(HEAPU8)));
+
+    }
+    if (hasActiveSegments(*wasm)) {
+      // If we have at least on active segment call `initActiveSegments` now.
```
kripken, line 556:
```suggestion
      // If we have at least one active segment, call `initActiveSegments` now.
```

src/wasm2js.h
```diff
@@ -2574,16 +2598,35 @@ void Wasm2JSGlue::emitPostES6() {
...
+    if (seg.isPassive) {
+      out << "  var memorySegments = {};\n";
+      break;
+    }
+  }
+
   out <<
-    R"(for (var base64ReverseLookup = new Uint8Array(123/*'z'+1*/), i = 25; i >= 0; --i) {
+    R"(  var base64ReverseLookup = new Uint8Array(123/*'z'+1*/);
+  for (var i = 25; i >= 0; --i) {
```
kripken, line 2629: The old way is more compact (a single `var`).

kripken (approved):

src/wasm2js.h
sbc100: But it was horrendous to read.. I makes it look like `base64ReverseLookup` is scoped the for loop which it most certainly is not.  I don't think its worth the 4 bytes... if we want to go after 4 byte saving here I think there way much lower hanging fruit.

sbc100: If you feel strongly I'd rather find other vars i this file to collapse.

sbc100: For example couldn't we collapse all of these:

```
 var Math_imul = Math.imul;                                                      
 var Math_fround = Math.fround;                                                  
 var Math_abs = Math.abs;                                                        
 var Math_clz32 = Math.clz32;                                                    
 var Math_min = Math.min;                                                        
 var Math_max = Math.max;                                                        
 var Math_floor = Math.floor;                                                    
 var Math_ceil = Math.ceil;                                                      
 var Math_trunc = Math.trunc;                                                    
 var Math_sqrt = Math.sqrt;                                                      
 var abort = env.abort;   
```

?

kripken: Fair enough, yeah, I don't feel strongly.

We should probably leave such micro-opts for closure anyhow. (which I think it does in closure=2)

PR #3341, opened 2020-11-12
Create a pass to log functions

awtcode:
Create a log function pass that assigns an index to every function. This index is not guaranteed to be consistent if other passes modify the functions in the wasm binary.


awtcode:
@sbc100 and @tlively . This is the initial drop for the log function pass.

src/passes/LogFunction.cpp
```diff
@@ -0,0 +1,73 @@
...
+struct LogFunction : public WalkerPass<PostWalker<LogFunction>> {
+  void visitFunction(Function* curr) {
+    if (curr->imported()) {
+      return;
+    }
+    if (auto* block = curr->body->dynCast<Block>()) {
+      if (!block->list.empty()) {
+        block->list.back() = makeLogCall(block->list.back());
+      }
+    }
```
tlively, line 46: Why add a log call at the end of the body? I would expect the call at the beginning of the body would be sufficient. A call at the end of the body also isn't guaranteed to be run, since control flow in the body could skip over it.

src/passes/LogFunction.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include <wasm-builder.h>
+#include <wasm.h>
+
+namespace wasm {
+
+Name FNLOGGER("log_function");
+
+struct LogFunction : public WalkerPass<PostWalker<LogFunction>> {
+  void visitFunction(Function* curr) {
+    if (curr->imported()) {
```
tlively, line 39: Since this doesn't really need to walk the full bodies of each function, you could make this a regular `Pass` and use `ModuleUtils::iterDefinedFunctions` (from module-utils.h) inside its `run` method rather than using `visitFunction`.

src/passes/LogFunction.cpp
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright 2017 WebAssembly Community Group participants
```
kripken, line 2:
```suggestion
 * Copyright 2020 WebAssembly Community Group participants
```

kripken (commented):
Please also add a test, see e.g. `test/passes/log-execution.*` for the ones for the similar pass.

src/passes/LogFunction.cpp
awtcode: Oops you are right, will remove it.

src/passes/LogFunction.cpp
```diff
@@ -0,0 +1,69 @@
...
+    import->name = FNLOGGER;
+    import->module = ENV;
+    import->base = FNLOGGER;
+    import->sig = Signature(Type::i32, Type::none);
+    curr->addFunction(import);
+  }
+
+private:
+  Expression* makeLogCall(Expression* curr) {
+    static Index id = 0;
```
juj, line 58: Because each function has an index, would it be possible for this log call to call out with that specific index? That would make matching up with `--emit-symbol-map` immediately possible.

src/passes/LogFunction.cpp
awtcode: > 
> 
> Since this doesn't really need to walk the full bodies of each function, you could make this a regular `Pass` and use `ModuleUtils::iterDefinedFunctions` (from module-utils.h) inside its `run` method rather than using `visitFunction`.

We will be instrumenting at the block level moving forward so is `ModuleUtils::iterDefinedFunctions` sufficient moving forward?
Also, we are using `visitModule` to register the JS import for now. We can rework this when you are done with handcrafting the Wasm function to create the profile data?

tlively:
@awtcode If you want, we could skip this prototype instrumentation and move directly to instrumentation built into wasm-split that uses the profile dumping API we discussed. I've already started on that, so it should be ready to go soon.

awtcode:
> 
> 
> @awtcode If you want, we could skip this prototype instrumentation and move directly to instrumentation built into wasm-split that uses the profile dumping API we discussed. I've already started on that, so it should be ready to go soon.

Sounds good @tlively, please go ahead :) Thanks for your help!

awtcode:
Not required anymore

PR #3342, opened 2020-11-12
[Fuzzer] Compare D8 on Liftoff and TurboFan

kripken:
Previously we picked one of the two compilers at the top level. But that doesn't
actually compare between them directly - each entire run used one of the two.

Instead, add separate "VMs" for each of them, and keep the existing D8 VM as
well (which tests tiering up).

The code also seems nicer this way.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -429,7 +429,7 @@ class CompareVMs(TestCaseHandler):
 
             def run(self, wasm, extra_d8_flags=[]):
                 run([in_bin('wasm-opt'), wasm, '--emit-js-wrapper=' + wasm + '.js'] + FEATURE_OPTS)
-                return run_vm([shared.V8, wasm + '.js'] + shared.V8_OPTS + ['--', wasm])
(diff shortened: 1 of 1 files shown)


scripts/fuzz_opt.py
```diff
@@ -556,7 +549,8 @@ def can_compare_to_others(self):
                 # NaNs can differ from wasm VMs
                 return not NANS
 
-        self.vms = [BinaryenInterpreter(), D8(), Wasm2C(), Wasm2C2Wasm()]
+        self.vms = [BinaryenInterpreter(), D8(), D8Liftoff(), D8TurboFan(),
```
tlively, line 552: Is plain `D8()` still useful?

tlively (approved):
Nice! I agree this looks cleaner.

scripts/fuzz_opt.py
kripken: I think so. That is how V8 normally runs, with (nondeterministic) tiering up, so it's good to test that, even if it is less valuable now.

PR #3343, opened 2020-11-12
[Fuzzer] Add a chance to pick particularly important initial contents

kripken:
`OptimizeInstructions` is seeing the most work these days (thanks @MaxGraey !)
so it's good for the fuzzer to focus on that some more.

Also move some code around in the main test wast: it's useful to put each
feature in its own module to maximize the chance of getting them to be used.
That is, if a module has a single use of atomics, then if atomics are disabled
in the current run, we can't use any of the module and we skip initial contents
entirely. Moving each feature to it's own module reduces that risk. (We do
pick randomly between the modules, and atm a small module has the same
chance as a big one, but this still seems worth it.)

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -169,6 +169,13 @@ def randomize_fuzz_settings():
     print('randomized settings (NaNs, OOB, legalize, extra V8_OPTS):', NANS, OOB, LEGALIZE, extra_v8_opts)
 
 
+IMPORTANT_INITIAL_CONTENTS = [
+    os.path.join('passes', 'optimize-instructions_all-features.wast'),
+    os.path.join('passes', 'optimize-instructions_fuzz-exec.wast'),
+]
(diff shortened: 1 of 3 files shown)


tlively (approved):

scripts/fuzz_opt.py
```diff
@@ -178,7 +185,12 @@ def pick_initial_contents():
     # half the time don't use any initial contents
     if random.random() < 0.5:
         return
-    test_name = random.choice(all_tests)
+    # some of the time use initial contents that are known to be especially
+    # important
+    if random.random() < 0.5:
+        test_name = random.choice(IMPORTANT_INITIAL_CONTENTS)
```
aheejin, line 191: Is that single pass worth half initial contents testing? (I'm not very familiar with things there myself, just wondering..)

kripken: Maybe half is a little high, yeah, I'm not sure. But it is the pass that is getting a lot of new commits, every day basically - and we've found bugs. The other initial contents have not found bugs for weeks now.

PR #3344, opened 2020-11-12
Remove boilerplate in walking logic

kripken:
Also fix the order of walking children, which was wrong in the macro. What's nice
is that fixing it there would fix anything else using the macro automatically
(however, so far nothing else was affected by it).

This is an example where we just care about children, and not immediates,
so we need to emit a line for each immediate that does nothing. Maybe that
justifies adding a way to not error on missing defines. However, if it's just here,
then I think it may not be worth it - I suggest waiting to see if more places
need such a thing (and it's just a few lines here, and something that will need
very few updates).


src/wasm-delegations-fields.h
```diff
@@ -93,7 +93,7 @@
 #ifndef DELEGATE_FIELD_CHILD_VECTOR
 #ifdef DELEGATE_GET_FIELD
 #define DELEGATE_FIELD_CHILD_VECTOR(id, name)                                  \
-  for (Index i = 0; i < (DELEGATE_GET_FIELD(id, name)).size(); i++) {          \
+  for (int i = int((DELEGATE_GET_FIELD(id, name)).size()) - 1; i >= 0; i--) {  \
```
tlively, line 96: Are we confident that iterating through the vector in reverse is always going to be the expected default behavior? I guess this is consistent with the order in which other children are visited, so it shouldn't be a problem.

tlively (approved):

src/wasm-delegations-fields.h
kripken: Yes, if we need the opposite order we may want to add another option. I can't think of a use case atm for that though.

PR #3345, opened 2020-11-12
Update tests following conflicting landings

kripken:


kripken:
Landing quickly to fix breakage.

PR #3348, opened 2020-11-13
[wasm64] fix for Memory64Lowering affecting DWARF data

aardappel:
We change the AddrSize which causes all DW_FORM_addr to be written differently.
Depends on https://reviews.llvm.org/D91395

This may need a test.. sadly dwarfdump doesn't show the size of DW_FORM_addr, they're always 64-bit. So its only `addr_size = 0x04` in the header thats evidence of this code working.


kripken (approved):
This is simpler than I had thought!

PR #3351, opened 2020-11-13
Fix warnings in delegations-fields header

### LINKED ISSUE #3350: warning C4003: not enough arguments for function-like macro invocation 'DELEGATE_END'
There is a flood of such warnings generated when building on Windows. 

The macro is defined as `#define DELEGATE_END(id)` then called as `DELEGATE_END();`, not sure what the preferred solution is, maybe define a separate `DELEGATE_END_ID` ?

kripken:
The end macro is not actually used yet, it is just for symmetry. And apparently
gcc/clang do not warn on this, so it was missed...

Fixes #3350


aardappel (approved):

aardappel:
Confirmed warnings are gone.

PR #3352, opened 2020-11-13
Make getExpressionName resemble instruction names

aheejin:
This function does not return exact instruction names but more of
category names. But when there is a matching instruction, as in case of
`global.get/set` or `local.get/set`, it seems to return instruction
names. In that regard, this makes `getExpressionName`'s return values to
similar to that of real instruction names when possible, in case of
some atomic instructions and `memory.init/copy` and `data.drop`.

It is hard to make a test for this because this function is used in a
very limited way in the codebase, such as:
- When printing error messages
- When printing a stack instruction names, but only for control flow
  instructions
- When printing instruction names in Metrics

### DIFF
--- a/src/wasm/wasm.cpp
+++ b/src/wasm/wasm.cpp
@@ -154,15 +154,15 @@ const char* getExpressionName(Expression* curr) {
     case Expression::Id::UnreachableId:
       return "unreachable";
     case Expression::Id::AtomicCmpxchgId:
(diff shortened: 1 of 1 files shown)


kripken (approved):
In principle we could write a C++ unit test for this under `test/example/` but I don't think it's worth it as it's mainly for debugging.

PR #3353, opened 2020-11-13
Rename atomic.notify and *.atomic.wait

aheejin:
- atomic.notify -> memory.atomic.notify
- i32.atomic.wait -> memory.atomic.wait32
- i64.atomic.wait -> memory.atomic.wait64

See WebAssembly/threads#149.

This renames instruction name printing but not the internal data
structure names, such as `AtomicNotify`, which are not always the same
as printed instruction names anyway. This also does not modify C API.
But this fixes interface functions in binaryen.js because it seems
binaryen.js's interface functions all follow the corresponding
instruction names.


kripken (approved):
Please mention the JS API change in `CHANGELOG.md`.

@dcodeIO , any concerns about the JS API change?

tlively (approved):

dcodeIO:
LGTM, thanks for the heads up :)

PR #3355, opened 2020-11-13
Rename Indirect to NonDirect in CallGraphPropertyAnalysis

kripken:
This is in preparation for [`CallRef`](https://github.com/WebAssembly/function-references/blob/master/proposals/function-references/Overview.md), which takes a reference to a function
and calls it.

CallGraphPropertyAnalysis needs to be aware of anything that is not a
direct call, and the best name I can think of that covers `CallIndirect` and
`CallRef` is "non-direct". Maybe there's a better idea though?

### DIFF
--- a/src/ir/module-utils.h
+++ b/src/ir/module-utils.h
@@ -308,7 +308,9 @@ template<typename T> struct CallGraphPropertyAnalysis {
   struct FunctionInfo {
     std::set<Function*> callsTo;
     std::set<Function*> calledBy;
-    bool hasIndirectCall = false;
+    // A non-direct call is any call that is not direct. That includes
+    // CallIndirect and CallRef.
+    bool hasNonDirectCall = false;
   };
 
   typedef std::map<Function*, T> Map;
@@ -331,7 +333,7 @@ template<typename T> struct CallGraphPropertyAnalysis {
         }
 
         void visitCallIndirect(CallIndirect* curr) {
-          info.hasIndirectCall = true;
+          info.hasNonDirectCall = true;
         }
 
       private:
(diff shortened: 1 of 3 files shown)


aheejin:
In theory I think `call.ref` also can be called an indirect call, so I think that should be fine to leave it as is, but given that we have an instruction `call_indirect`, I understand why you would want a non-ambiguous name... I think either is fine, maybe modulo the asyncify flag you mentioned..?

aheejin (approved):
Anyway I think the name is OK either way

kripken:
@aheejin You led me to think about another option here. Since `call_ref` is really a form of indirect call, maybe we could actually reuse the same class for both? That is, `CallIndirect` would be `call_indirect` if the target is an `int32`, and if it's a typed function reference then it would be a `call_ref`. This could almost just work, since the C++ class would be the same, anyhow. It would be a little like how `local.tee` and `local.set` are the same class.

This seems like less work and less code, and the downsides seem pretty minor. What do you think?

(And if we do that, then there's no need for this PR, but that's less important.)

kripken:
Reading more carefully, `CallIndirect` has a function type signature, which would not be used for a reference call. So this isn't perfectly symmetrical. If we reuse the same class, that field would not be used for reference calls.

kripken:
After experimenting a bit with both options, I think the idea to use a single class for both would save some code, but increase complexity. We'd have to add a field to identify it as which of the two it is, since we can't use the type for that (it might be unreachable), and other issues.

So I'll land this and move forward with the straightforward approach of a new `CallRef` class.

aheejin:
Yeah I agree using separate classes for them is better.

PR #3357, opened 2020-11-14
[Types] Identify subtyping between funcref and all typed function references

kripken:
This is necessary to support typed function references.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -522,15 +522,13 @@ bool Type::isSubType(Type left, Type right) {
   }
   if (left.isRef() && right.isRef()) {
     if (right == Type::anyref) {
-      // Everything is a subtype of anyref.
       return true;
     }
-    if (left == Type::i31ref && right == Type::eqref) {
-      // i31 is a subtype of eqref.
+    if ((left == Type::i31ref || left == Type::array || left == Type::struct) && right == Type::eqref) {
       return true;
     }
     if (left.getHeapType().isSignature() && right == Type::funcref) {
-      // All function signatures are subtypes of funcref.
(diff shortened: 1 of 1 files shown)


src/wasm/wasm-type.cpp
```diff
@@ -521,8 +521,19 @@ bool Type::isSubType(Type left, Type right) {
...
-           (left == Type::i31ref && right == Type::eqref);
+    if (right == Type::anyref) {
+      // Everything is a subtype of anyref.
+      return true;
+    }
+    if (left == Type::i31ref && right == Type::eqref) {
+      // i31 is a subtype of eqref.
+      return true;
+    }
+    if (right == Type::funcref && left.getHeapType().isSignature()) {
```
tlively, line 532:
```suggestion
    if (left.getHeapType().isSignature() && right == Type::funcref) {
```
I would put the left on the left and the right on the right.

tlively (commented):
This could use a test in test/reference-types.wast. (if it's possible to express in a wast?)

Also, while you're editing this code you could also implement `array <: eqref` and `struct <: eqref` if you wanted.

src/wasm/wasm-type.cpp
kripken: I guess there's no reason not to, yeah, there's no speed reason here (which is I think why my brain preferred it).

kripken:
I don't think this is expressible/testable until I actually add the new feature for typed function references. All the stuff will get tested then. I was trying to split off small pieces where possible, but the downside is that then they don't arrive with tests... happy to do either way.

kripken:
Added array and struct.

tlively (approved):

PR #3358, opened 2020-11-14
[TypedFunctionReferences] Allow creation of literals with typed function types

kripken:
Before this, all function literals would have `funcref` type. This allows a custom type
to be provided.

The interpreter passes in the type of the `ref.func` expression. Currently this has
type `funcref` so there is no change, but when typed function references lands and
is enabled then the proper subtype will be used.

### DIFF
--- a/src/literal.h
+++ b/src/literal.h
@@ -71,7 +71,8 @@ public:
   explicit Literal(const std::array<Literal, 8>&);
   explicit Literal(const std::array<Literal, 4>&);
   explicit Literal(const std::array<Literal, 2>&);
-  explicit Literal(Name func) : func(func), type(Type::funcref) {}
+  explicit Literal(Name func, Type type = Type::funcref)
+    : func(func), type(type) {}
   explicit Literal(std::unique_ptr<ExceptionPackage>&& exn)
     : exn(std::move(exn)), type(Type::exnref) {}
   Literal(const Literal& other);
@@ -233,7 +234,9 @@ public:
     assert(type.isNullable());
     return Literal(type);
   }
-  static Literal makeFunc(Name func) { return Literal(func.c_str()); }
+  static Literal makeFunc(Name func, Type type = Type::funcref) {
+    return Literal(func.c_str(), type);
+  }
   static Literal makeExn(std::unique_ptr<ExceptionPackage>&& exn) {
     return Literal(std::move(exn));
   }
--- a/src/literal.h
+++ b/src/literal.h
@@ -235,7 +235,7 @@ public:
     return Literal(type);
   }
   static Literal makeFunc(Name func, Type type = Type::funcref) {
-    return Literal(func.c_str(), type);
+    return Literal(func, type);
   }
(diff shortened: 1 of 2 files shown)


src/literal.h
```diff
@@ -233,7 +234,9 @@ class Literal {
     assert(type.isNullable());
     return Literal(type);
   }
-  static Literal makeFunc(Name func) { return Literal(func.c_str()); }
+  static Literal makeFunc(Name func, Type type = Type::funcref) {
+    return Literal(func.c_str(), type);
```
tlively, line 238: Is this `c_str()` just implicitly made back into a `Name`? I don't see a `Literal` constructor that takes `char*`.

tlively (approved):

aheejin:
What are other types that can be custom-provided? Function type references proposal seems to introduce named nominal types, is my understanding correct? In that case, can types be expanded indefinitely?

And given that it's hard to provide a test for now for this, would it be better to add support for this later?

tlively:
> What are other types that can be custom-provided?

Any function reference type like `(ref (func (params i32) (results f64)))`. Only `funcref` and typed function reference types would be allowed here.

> Function type references proposal seems to introduce named nominal types, is my understanding correct?

Not quite; function types are structural rather than nominal, so if `(ref (func (params i32) (results f64)))` is declared twice in a type section, the two type indices refer to the same type and can be used interchangeably. If the types were nominal, the indices with the same definitions would not be interchangeable.

The `Name` used in the `Literal` identifies the function that this literal `funcref` or typed function reference refers to and is not part of the type.

> In that case, can types be expanded indefinitely?

Yes, even without being nominal, function types could in principle be recursive and therefore expand infinitely. An example would be a function type that takes an `i32` as an argument returns itself as a result. Binaryen doesn't have a mechanism for constructing recursive types yet, though.

src/literal.h
kripken: Good point, looks like that `.c_str()` was unnecessary before, fixing.

kripken:
@aheejin 

> And given that it's hard to provide a test for now for this, would it be better to add support for this later?

The benefit of splitting out parts of the future typed-function-references PR is to avoid that one PR being really big. I think it's better to have small pieces. But if you disagree let's discuss that.

There is no risk this won't get tested later - this (and other small pieces) are literally things that the current test suite will fail on when enabling the new feature.

aheejin (approved):
@tlively Thanks for the answer!

@kripken
No I am not opinionated in particular; I was just wondering because there was no tests.

> There is no risk this won't get tested later - this (and other small pieces) are literally things that the current test suite will fail on when enabling the new feature.

Which new feature are you talking about? Do we already have the typed function reference feature flag, and if we enable it, does this crash?

kripken:
@aheejin The new feature is typed function references. We don't have it yet. In a PR soon I'll add it, with extra tests (but existing ones cover a lot actually, just because we run tests with `-all` which turns it on, and it affects the behavior of `ref.func` that already existed).

PR #3359, opened 2020-11-14
Initial wasm-split tool

tlively:
Implement an initial version of the wasm-split tool, which splits modules into a
primary module and a secondary module that can be instantiated after the primary
module. Eventually, this tool will be able to not only split modules, but also
instrument modules to collect profiles that will be able to guide later
splitting. In this initial version, however, wasm-split can neither perform
instrumentation nor consume any kind of profile data.

Despite those shortcomings, this initial version of the tool is already able to
perform module splitting according to function lists manually provided by the
user via the command line. Follow-up PRs will implement the stubbed out
instrumentation and profile consumption functionality.

### DIFF
--- a/src/tools/wasm-split.cpp
+++ b/src/tools/wasm-split.cpp
@@ -1,0 +1,367 @@
+/*
+ * Copyright 2020 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// wasm-split: Split a module in two or instrument a module to inform future
+// splitting.
+
+#include "ir/module-splitting.h"
+#include "ir/module-utils.h"
+#include "support/name.h"
+#include "support/utilities.h"
+#include "tool-options.h"
+#include "wasm-io.h"
+#include "wasm-validator.h"
+#include <sstream>
+
+using namespace wasm;
+
+namespace {
+
+std::set<Name> parseNameList(const std::string& list) {
+  std::set<Name> names;
+  std::istringstream stream(list);
+  for (std::string name; std::getline(stream, name, ',');) {
+    names.insert(name);
+  }
+  return names;
+}
+
+struct WasmSplitOptions : ToolOptions {
+  bool verbose = false;
+
+  bool instrument = false;
+
+  std::string profileFile;
+  std::string profileExport = "__write_profile";
+
+  std::set<Name> keepFuncs;
+  std::set<Name> splitFuncs;
+
+  std::string input;
+  std::string output;
+  std::string primaryOutput;
+  std::string secondaryOutput;
+
+  std::string importNamespace;
+  std::string placeholderNamespace;
+  std::string exportPrefix;
+
+  WasmSplitOptions();
+  bool validate();
+  void parse(int argc, const char* argv[]);
+};
+
+WasmSplitOptions::WasmSplitOptions()
+  : ToolOptions("wasm-split",
+                "Split a module into a primary module and a secondary "
+                "module or instrument a module to gather a profile that "
+                "can inform future splitting.") {
+  (*this)
+    .add("--instrument",
+         "",
+         "Instrument the module to generate a profile that can be used to "
+         "guide splitting",
+         Options::Arguments::Zero,
+         [&](Options* o, const std::string& argument) { instrument = true; })
+    .add(
+      "--profile",
+      "",
+      "The profile to use to guide splitting. May not be used with "
+      "--instrument.",
+      Options::Arguments::One,
+      [&](Options* o, const std::string& argument) { profileFile = argument; })
+    .add("--profile-export",
+         "",
+         "The export name of the function the embedder calls to write the "
+         "profile into memory. Defaults to `__write_profile`. Must be used "
+         "with --instrument.",
+         Options::Arguments::One,
+         [&](Options* o, const std::string& argument) {
+           profileExport = argument;
+         })
+    .add("--keep-funcs",
+         "",
+         "Comma-separated list of functions to keep in the primary module, "
+         "regardless of any profile.",
+         Options::Arguments::One,
+         [&](Options* o, const std::string& argument) {
+           keepFuncs = parseNameList(argument);
+         })
+    .add("--split-funcs",
+         "",
+         "Comma-separated list of functions to split into the secondary "
+         "module, regardless of any profile. If there is no profile, then "
+         "this defaults to all functions defined in the module.",
+         Options::Arguments::One,
+         [&](Options* o, const std::string& argument) {
+           splitFuncs = parseNameList(argument);
+         })
+    .add("--output",
+         "-o",
+         "Output file. Only usable with --instrument.",
+         Options::Arguments::One,
+         [&](Options* o, const std::string& argument) { output = argument; })
+    .add("--primary-output",
+         "-o1",
+         "Output file for the primary module. Not usable with --instrument.",
+         Options::Arguments::One,
+         [&](Options* o, const std::string& argument) {
+           primaryOutput = argument;
+         })
+    .add("--secondary-output",
+         "-o2",
+         "Output file for the secondary module. Not usable with --instrument.",
+         Options::Arguments::One,
+         [&](Options* o, const std::string& argument) {
+           secondaryOutput = argument;
+         })
+    .add("--import-namespace",
+         "",
+         "The namespace from which to import objects from the primary "
+         "module into the secondary module.",
+         Options::Arguments::One,
+         [&](Options* o, const std::string& argument) {
+           importNamespace = argument;
+         })
+    .add("--placeholder-namespace",
+         "",
+         "The namespace from which to import placeholder functions into "
+         "the primary module.",
+         Options::Arguments::One,
+         [&](Options* o, const std::string& argument) {
+           placeholderNamespace = argument;
+         })
+    .add(
+      "--export-prefix",
+      "",
+      "An identifying prefix to prepend to new export names created "
+      "by module splitting.",
+      Options::Arguments::One,
+      [&](Options* o, const std::string& argument) { exportPrefix = argument; })
+    .add("--verbose",
+         "-v",
+         "Verbose output mode. Prints the functions that will be kept "
+         "and split out when splitting a module.",
+         Options::Arguments::Zero,
+         [&](Options* o, const std::string& argument) {
+           verbose = true;
+           quiet = false;
+         })
+    .add_positional(
+      "INFILE",
+      Options::Arguments::One,
+      [&](Options* o, const std::string& argument) { input = argument; });
+}
+
+bool WasmSplitOptions::validate() {
+  bool valid = true;
+  auto fail = [&](auto msg) {
+    std::cerr << "error: " << msg << "\n";
+    valid = false;
+  };
+
+  if (!input.size()) {
+    fail("no input file");
+  }
+  if (instrument) {
+    using Opt = std::pair<const std::string&, const std::string&>;
+    for (auto& opt : {Opt{profileFile, "--profile"},
+                      Opt{primaryOutput, "primary output"},
+                      Opt{secondaryOutput, "secondary output"},
+                      Opt{importNamespace, "--import-namespace"},
+                      Opt{placeholderNamespace, "--placeholder-namespace"},
+                      Opt{exportPrefix, "--export-prefix"}}) {
+      if (opt.first.size()) {
+        fail(opt.second + " cannot be used with --instrument");
+      }
+    }
+    if (keepFuncs.size()) {
+      fail("--keep-funcs cannot be used with --instrument");
+    }
+    if (splitFuncs.size()) {
+      fail("--split-funcs cannot be used with --instrument");
+    }
+  } else {
+    if (output.size()) {
+      fail(
+        "must provide separate primary and secondary output with -o1 and -o2");
+    }
+  }
+
+  std::vector<Name> impossible;
+  std::set_intersection(keepFuncs.begin(),
+                        keepFuncs.end(),
+                        splitFuncs.begin(),
+                        splitFuncs.end(),
+                        std::inserter(impossible, impossible.end()));
+  for (auto& func : impossible) {
+    fail(std::string("Cannot both keep and split out function ") +
+         func.c_str());
+  }
+
+  return valid;
+}
+
+void WasmSplitOptions::parse(int argc, const char* argv[]) {
+  ToolOptions::parse(argc, argv);
+  // Since --quiet is defined in ToolOptions but --verbose is defined here,
+  // --quiet doesn't know to unset --verbose. Fix it up here.
+  if (quiet && verbose) {
+    verbose = false;
+  }
+}
+
+void parseInput(Module& wasm, const WasmSplitOptions& options) {
+  ModuleReader reader;
+  reader.setDWARF(false);
+  reader.setProfile(options.profile);
+  try {
+    reader.read(options.input, wasm);
+  } catch (ParseException& p) {
+    p.dump(std::cerr);
+    std::cerr << '\n';
+    Fatal() << "error parsing wasm";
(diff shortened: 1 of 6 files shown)


tlively:
cc @kripken @sbc100 @awtcode @dschuff 

This is a lot more concrete than anything we've talked about so far, so I am open to feedback about large changes we should make. How to test this is also an open question, since none of the current testing infrastructure is set up to handle multiple output files. I'm thinking this might be a good opportunity to experiment with using FileCheck + lit in Binaryen.

test/lit/wasm-split/basic.wast
```diff
@@ -0,0 +1,141 @@
+;; RUN: %wasm-split %s --export-prefix='%' -o1 %t.none.1.wasm -o2 %t.none.2.wasm -v 2>&1 \
```
aheejin, line 1: Haven't read the PR yet, but are we introducing lit tests here? Are we supposed to add tests in this format or the current txt format from now? Are we planning to eventually migrate all existing tests to lit tests? Maybe I missed some previous discussions on this.

tlively: We’ve thought about it before, but you haven’t missed any recent discussions or decisions. I just wanted to see what would be possible. I expect we will discuss this and try to answer all those questions together on Monday. I think it would be nice to move to lit and FileCheck for most of our tests, but I don’t know what anyone else thinks about that.

.github/workflows/ci.yml
```diff
@@ -103,14 +106,16 @@ jobs:
     - uses: actions/checkout@v1
     - name: install ninja
       run: sudo apt-get install ninja-build
+    - name: install lit and FileCheck
+      run: pip3 install lit FileCheck
```
sbc100, line 110: If we decide to go with this new test framework, can we make this a seperate PR.  Just add lit + FileCheck and perhaps convert one or two trivial tests?

.gitignore
```diff
@@ -37,6 +37,10 @@ CMakeFiles
 # files commonly related to building out-of-tree
 /build/
 
+# files produced by lit testing
+**/Output
```
sbc100, line 41: We already a have `out/test` for test output.  Can we teach lit to use the same directory?

scripts/test/shared.py
```diff
@@ -54,6 +54,10 @@ def parse_args(args):
               ' Default: bin/ of current directory (i.e. assume an in-tree build).'
               ' If not specified, the environment variable BINARYEN_ROOT= can also'
               ' be used to adjust this.'))
+    parser.add_argument(
+        '--binaryen-build', dest='binaryen_build', default='',
+        help=('Specifies a path to the Binaryen build directory. '
+              'Default: parent of `binaryen_bin`'))
```
sbc100, line 60: Why do we need this in addition `--binaryen-bin`?   It would be nice if we could avoid adding more arguments here.

.gitignore
tlively: Possibly. I will investigate.

scripts/test/shared.py
tlively: Answered [here](https://github.com/WebAssembly/binaryen/pull/3367#issuecomment-727641092) so we can keep this discussion focused more on the wasm-split code.

.github/workflows/ci.yml
tlively: Yes, I agree 👍 see #3367 for what that would look like.

.gitignore
tlively: It turns out that this "Output" directory is hard-coded into lit.

sbc100: Oh.. that is odd.. and uppercase output non-the-less.   I guess I'd never realized that "Output" was part of all my tests intermediates in llvm.

In any case this `Output` directory looks like lives under the build directory so this is only needs for in-tree builds so it can go under the `files related to building in-tree` category. 

It would be good if we could make the current check.py tests also write to the build director and use this same sub-directory name (maybe as a precursor to this change).

src/tools/wasm-split.cpp
```diff
@@ -0,0 +1,372 @@
...
+  void parse(int argc, const char* argv[]);
+};
+
+WasmSplitOptions::WasmSplitOptions()
+  : ToolOptions("wasm-split",
+                "Split a module into a primary module and a secondary "
+                "module or instrument a module to gather a profile that "
+                "can inform future splitting.") {
+  (*this)
+    .add("--instrument",
```
kripken, line 75: maybe instead of `--instrument` there could be
```
wasm-split instrument --flag
wasm-split split --otherflag
```
? That is, there are really two main modes for the tool, so passing that in as a first argument feels nicer somehow. But just bikeshedding.

src/tools/wasm-split.cpp
```diff
@@ -0,0 +1,372 @@
...
+  // Since --quiet is defined in ToolOptions but --verbose is defined here,
+  // --quiet doesn't know to unset --verbose. Fix it up here.
+  if (quiet && verbose) {
+    verbose = false;
+  }
+}
+
+void parseInput(Module& wasm, const WasmSplitOptions& options) {
+  ModuleReader reader;
+  reader.setDWARF(false);
```
kripken, line 234: the default is false
```suggestion
```

src/tools/wasm-split.cpp
```diff
@@ -0,0 +1,372 @@
...
+  reader.setProfile(options.profile);
+  try {
+    reader.read(options.input, wasm);
+  } catch (ParseException& p) {
+    p.dump(std::cerr);
+    std::cerr << '\n';
+    Fatal() << "error parsing wasm";
+  } catch (MapParseException& p) {
```
kripken, line 242: we'll hopefully be removing `MapParseException` soon as DWARF is replacing source maps. Meanwhile, it fine to not check for it here, and not add more code to remove later.

src/tools/wasm-split.cpp
```diff
@@ -0,0 +1,372 @@
...
+    } else {
+      keepFuncs.erase(func);
+    }
+  }
+
+  if (!options.quiet && keepFuncs.size() == 0) {
+    std::cerr << "warning: not keeping any functions in the primary module\n";
+  }
+
+  // If warnings are enabled, check that any functions are being split out
```
kripken, line 293:
```suggestion
  // If warnings are enabled, check that any functions are being split out.
```

kripken: We are not very consistent about this, but might be worth it.

test/lit/wasm-split/invalid-options.wast
```diff
@@ -0,0 +1,64 @@
...
+;; RUN: not wasm-split %s -o %t 2>&1 \
+;; RUN:   | filecheck %s --check-prefix NO-INSTRUMENT-OUT
+
+;; --instrument is required to use --profile-export
+;; RUN: not wasm-split %s --profile-export=foo 2>&1 \
+;; RUN:   | filecheck %s --check-prefix NO-INSTRUMENT-PROFILE-EXPORT
+
+;; INSTRUMENT-PROFILE: error: --profile cannot be used with --instrument
+
+;; INSTRUMENT-OUT1: error: primary output cannot be used with --instrument
```
kripken, line 46: I thought the big benefit of lit is that the output is next to the case. why are all the outputs at the end?

src/tools/wasm-split.cpp
tlively: Yes, I agree that this seems nicer. The only downside is the ugly repetition in `wasm-split split`. Maybe that's not a huge problem, though.

sbc100: How about `wasm-split cleave` ?  (joking)

test/lit/wasm-split/invalid-options.wast
tlively: The more general benefit of lit is its flexibility and ability to put multiple test invocations in a single file ;)

In this case it seemed simpler to put the error messages all together because they are so simple to read. Do you think this would be more readable with the error messages next to their corresponding RUN lines?

src/tools/wasm-split.cpp
tlively: Actually, would you mind if we made this change in a follow up? I think the cleanest solution would be to build the concept of subcommands directly into the command line parsing in support/command-line.h.

src/tools/wasm-split.cpp
```diff
@@ -0,0 +1,367 @@
...
+  };
+
+  if (!input.size()) {
+    fail("no input file");
+  }
+  if (instrument) {
+    using Opt = std::pair<const std::string&, const std::string>;
+    for (auto& opt : {Opt{profileFile, "--profile"},
+                      Opt{primaryOutput, "primary output"},
+                      Opt{secondaryOutput, "secondary output"},
```
aheejin, line 185:
```suggestion
                      Opt{primaryOutput, "--primary-output"},
                      Opt{secondaryOutput, "--secondary-output"},
```
Any reason these are not options names?

src/tools/wasm-split.cpp
```diff
@@ -0,0 +1,367 @@
...
+      if (opt.first.size()) {
+        fail(opt.second + " cannot be used with --instrument");
+      }
+    }
+    if (keepFuncs.size()) {
+      fail("--keep-funcs cannot be used with --instrument");
+    }
+    if (splitFuncs.size()) {
+      fail("--split-funcs cannot be used with --instrument");
+    }
```
aheejin, line 198: Can't we also put these two options within the `for` loops above?

src/tools/wasm-split.cpp
```diff
@@ -0,0 +1,367 @@
...
+  } else {
+    if (output.size()) {
+      fail(
+        "must provide separate primary and secondary output with -o1 and -o2");
```
aheejin, line 202: Is this supposed to be used with optimization flags? I thought this is a standalone tool. (And is there any reason only -o1 and -o2 are possible?)

src/tools/wasm-split.cpp
```diff
@@ -0,0 +1,367 @@
...
+    }
+    if (profileExport != DEFAULT_PROFILE_EXPORT) {
+      fail("--profile-export must be used with --instrument");
+    }
```
aheejin, line 206: Isn't it possible that a user explicitly gives `--profile-export` which is the same as `DEFAULT_PROFILE_EXPORT`?

aheejin (approved):
LGTM % comments

src/tools/wasm-split.cpp
kripken: sure, let's leave it for now. it's probably fine as-is.

test/lit/wasm-split/invalid-options.wast
kripken: I think that might be clearer. but it might be that i'm just not used to lit in general since I don't do much work on LLVM. it's probably fine as is and I'll get used to it.

kripken (approved):

src/tools/wasm-split.cpp
aheejin: Hmm, I guess this was because we also have shorthands for those options? I think this should be fine then.

test/lit/wasm-split/invalid-options.wast
```diff
@@ -0,0 +1,64 @@
...
+
+;; --instrument cannot be used with --keep-funcs
+;; RUN: not wasm-split %s --instrument --keep-funcs=foo 2>&1 \
+;; RUN:   | filecheck %s --check-prefix INSTRUMENT-KEEP-FUNCS
+
+;; --instrument cannot be used with --split-funcs
+;; RUN: not wasm-split %s --instrument --split-funcs=foo 2>&1 \
+;; RUN:   | filecheck %s --check-prefix INSTRUMENT-SPLIT-FUNCS
+
+;; --instrument is required to use -o
```
aheejin, line 36: Nit: This comment seems to be hinting that user is missing `--instrument` but the error message seems to be saying the user is missing `-o1` and `-o2`. Would it be better to be consistent?

aheejin (commented):
One thing I'm wondering is, given that we have a disjoint set of instrumentation and split functionalities, do we need to keep both in a single tool? If we use two separate tools there wouldn't be any need to check all different combinations of options for errors.

src/tools/wasm-split.cpp
```diff
@@ -0,0 +1,367 @@
...
+std::set<Name> parseNameList(const std::string& list) {
+  std::set<Name> names;
+  std::istringstream stream(list);
+  for (std::string name; std::getline(stream, name, ',');) {
+    names.insert(name);
+  }
+  return names;
+}
+
+struct WasmSplitOptions : ToolOptions {
```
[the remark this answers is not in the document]
tlively: I believe that `public` is the default for structs, so specifying it wouldn't make a difference.

src/tools/wasm-split.cpp
```diff
@@ -0,0 +1,367 @@
...
+      Options::Arguments::One,
+      [&](Options* o, const std::string& argument) { exportPrefix = argument; })
+    .add("--verbose",
+         "-v",
+         "Verbose output mode. Prints the functions that will be kept "
+         "and split out when splitting a module.",
+         Options::Arguments::Zero,
+         [&](Options* o, const std::string& argument) {
+           verbose = true;
+           quiet = false;
```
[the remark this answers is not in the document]
tlively: Yes, I would like to have three levels of output:

 - `--quiet` means no warnings
 - default means warnings
 - `--verbose` means warnings and extra information

`--verbose` specifically dumps out whether each function is being kept in the primary module or split out into the secondary module. I will add a test for that functionality.

src/tools/wasm-split.cpp
tlively: No, unfortunately that would not type check because `keepFuncs` and `splitFuncs` are not strings.

src/tools/wasm-split.cpp
tlively: Yes, but I wasn't sure that fixing it by adding a separate variable like `bool profileExportSet` was worth the cost. What do you think?

tlively:
> One thing I'm wondering is, given that we have a disjoint set of instrumentation and split functionalities, do we need to keep both in a single tool? If we use two separate tools there wouldn't be any need to check all different combinations of options for errors.

The reason to keep both in the same tool is so that the user can use just a single tool for the entire workflow of instrumenting and splitting a module. My hypothesis is that this will be a nicer user experience than need to use multiple tools would be. I do think it would be nice to not have to manually validate all the disjoint options, though. As a follow-up, I want to investigate improvements we can make to the option parsing mechanism to make that part easier and to improve the `--help` documentation.

src/tools/wasm-split.cpp
aheejin: I think that should be fine not to check it, which will be a pretty marginal case anyway.

test/lit/wasm-split/verbose.wast
```diff
@@ -0,0 +1,14 @@
+;; Test that --verbose mode correctly prints the kept and split funcs
```
aheejin, line 1: Can't we do this by adding one more command in an existing test, such as basic.wast?

test/lit/wasm-split/invalid-options.wast
```diff
@@ -33,7 +33,7 @@
 ;; RUN: not wasm-split %s --instrument --split-funcs=foo 2>&1 \
 ;; RUN:   | filecheck %s --check-prefix INSTRUMENT-SPLIT-FUNCS
 
-;; --instrument is required to use -o
+;; --instrument requires -o1 and -o2 rather than -o
```
aheejin, line 36: Sorry, but isn't this the opposite? `--instrument` requires `-o` and not having `--instrument` requires `-o1` and `-o2`..

test/lit/wasm-split/verbose.wast
aheejin: Oh, right, we can't handle multiple modules...

test/lit/wasm-split/invalid-options.wast
aheejin: Gentle ping :) This is a pretty minor question, but just in case you haven't had a chance to see it. (The same for the question above)

test/lit/wasm-split/verbose.wast
tlively: Yes, and I use this functionality in the new test in #3400 as well. Do you think it would be a good idea to remove this test?

test/lit/wasm-split/invalid-options.wast
tlively: Sorry for missing these questions, and thank you for your fix!

test/lit/wasm-split/verbose.wast
aheejin: Oh no, I was just wondering if this is the only thing we test in this file, probably we could've added this test in basic.wast by adding one more `RUN` line. But I don't feel strongly anyway and this looks fine as is.

tlively: 👍 (I fixed my email filters and am getting these notifications now, btw)

PR #3360, opened 2020-11-14
Add include-what-you-use GitHub action

tlively:


tlively:
This was an interesting experiment, but:

 - It didn't save any meaningful time for a fresh build (didn't measure incremental builds)
 - The action takes an unpleasantly long time to run
 - This is a very large change for the benefit it brings
 - There is slightly more manual fixing than I would like
 - The suggestions and fixes don't seem portable enough
 - The GH action doesn't have an option for only running on src (we don't want to change third_party)

It might be worthwhile to go through and make a smaller patch to only remove unused `#include`s.

PR #3363, opened 2020-11-14
[cost.h] Require all expressions to define a cost

kripken:
Almost NFC, but adds a cost to `DataDrop` - just a number I made up, not sure
how much it makes sense.

### DIFF
--- a/src/ir/cost.h
+++ b/src/ir/cost.h
@@ -24,7 +24,7 @@ namespace wasm {
 
 // Measure the execution cost of an AST. Very handwave-ey
 
-struct CostAnalyzer : public Visitor<CostAnalyzer, Index> {
+struct CostAnalyzer : public OverriddenVisitor<CostAnalyzer, Index> {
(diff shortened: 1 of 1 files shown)


tlively:
We might want to make `data.drop` "cheap" since it is in principle supposed to save resources in the engine, but I'm not sure it matters in practice. Do we have a better way of getting these costs than just guessing? How much does the accuracy here matter in practice?

kripken:
The numbers are pretty hand-wavey. But it would be good if they are generally in the right order between each other. However there is a lot of uncertainty, so there is really no way to make them "accurate" (differences between VMs, between tiers, and other factors).

Overall I think `5` might make sense here as it does cause a call into the VM's internals - so it's not fast like an add or subtract which is guaranteed inline. And good point it should be fairly small as it should be favored given it helps later. So I still think 5 is ok - it's just slightly more than a call, basically.

tlively (approved):

PR #3365, opened 2020-11-14
[Tools] Apply features to the wasm module before loading it

kripken:
This normally does not matter, but will matter for typed function references,
where the presence of the feature will actually change how the IR is built. But,
it makes sense to do this early anyhow - we know the features already.

`wasm-dis` does not accept feature flags, because the idea is that it just
disassembles. This didn't matter much til now, but for consistency with the
other tools, set the features early, and in this case just set them all (which will
allow typed function references to just work, and I don't see a downside?).

### DIFF
--- a/src/tools/wasm-as.cpp
+++ b/src/tools/wasm-as.cpp
@@ -99,6 +99,7 @@ int main(int argc, const char* argv[]) {
   auto input(read_file<std::string>(options.extra["infile"], Flags::Text));
 
   Module wasm;
+  options.applyFeatures(wasm);
 
   try {
     if (options.debug) {
@@ -114,8 +115,6 @@ int main(int argc, const char* argv[]) {
     p.dump(std::cerr);
     Fatal() << "error in parsing input";
   }
-
-  options.applyFeatures(wasm);
 
   if (options.extra["validate"] != "none") {
     if (options.debug) {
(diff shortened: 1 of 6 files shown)


tlively:
With this change, the binary reader will override the user-provided features when it reads the target features section, removing the user's ability to override the target feature section. I'm not sure any users override the target feature section in practice, though, so perhaps we should reconsider this whole mechanism. In addition, `options.applyFeatures` not only applies the command line features to the module, but it also checks for consistency with the features section in the module. That check is a no-op after this change, for better or worse.

kripken:
Oh, interesting... I didn't think about interactions with the features section.

It does seem like typed function references making the features actually matter for loading the wasm can lead to weird interactions with things. It's the first time wasm adds such a feature in way that doesn't have a workaround, I believe.

What do you suggest doing here? Maybe it's not an important use case to let users override the feature section, but I'm not totally sure.

tlively:
What if we just always parse `ref.func` as using the more specific typed function references and allow that usage even without the typed-function-references feature enabled? We still would not allow typed function references to appear anywhere else, such as local, block, or function types without that feature enabled, so I don't think the difference would be observable once we write the binary. Passes that introduce new locals, blocks, globals, or functions would have to look at the features to determine whether to use `funcref` or more specific types, but that's not too onerous.

kripken:
Thanks, I think I can get that to work. It worries me that we may be missing something (to use newer types in older code) but it seems like only newer instructions would notice, once we fix all the subtyping validation issues. Closing in favor of that.

PR #3368, opened 2020-11-15
[wasm-emscripten-finalize] Don't emit a global ctor that is not needed

kripken:
If a ctor has no side effects, we don't need to call it.

Also add a missing check for a ctor being a function (and not something else,
which never happens in practice).

Fixes https://github.com/emscripten-core/emscripten/issues/12777

This will need to temporarily disable code size and metadce tests on emscripten as looks
like the improvement makes them fail.

cc @juj

### DIFF
--- a/src/tools/wasm-emscripten-finalize.cpp
+++ b/src/tools/wasm-emscripten-finalize.cpp
@@ -1,3 +1,4 @@
+#include <wasm-printing.h>
 /*
  * Copyright 2017 WebAssembly Community Group participants
  *
@@ -22,6 +23,7 @@
 #include <exception>
 
 #include "abi/js.h"
+#include "ir/effects.h"
 #include "ir/trapping.h"
 #include "support/colors.h"
 #include "support/debug.h"
@@ -296,7 +298,8 @@ int main(int argc, const char* argv[]) {
       if (auto* e = wasm.getExportOrNull(WASM_CALL_CTORS)) {
         if (e->kind == ExternalKind::Function) {
           // If the initializer does nothing, we don't need it.
-          if (wasm.getFunction(e->value)->body->is<Nop>()) {
+          std::cerr << wasm.getFunction(e->value)->body << '\n';
+          if (!EffectAnalyzer(options.passOptions, wasm.features, wasm.getFunction(e->value)->body).hasSideEffects()) {
             // We can also remove the unneeded export (later optimizations can
             // also remove the function, if it is not used elsewhere).
             wasm.removeExport(WASM_CALL_CTORS);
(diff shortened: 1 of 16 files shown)


src/tools/wasm-emscripten-finalize.cpp
```diff
@@ -294,7 +295,21 @@ int main(int argc, const char* argv[]) {
...
       if (auto* e = wasm.getExportOrNull(WASM_CALL_CTORS)) {
-        initializerFunctions.push_back(e->name);
+        if (e->kind == ExternalKind::Function) {
+          if (!EffectAnalyzer(options.passOptions,
+                              wasm.features,
+                              wasm.getFunction(e->value)->body)
+                 .hasSideEffects()) {
+            // The initializer exists, but doesn't do anything we could notice.
+            // Remove the export, and don't call it from JS (later
+            // optimizations can also remove the function, if it is not used
```
sbc100, line 305: I guess we could see this as the most trivial form of `eval-ctors`.. I wonder if its time to revive that pass?   We should not block this change on that, but maybe node to remove this once we get `eval-ctors` working?   

How does this relative to minimally modifying the wasm file?   Presmably in debug builds we would not want to modify the wasm file in this way because it would require the DWARF to get re-written.   As long as we don't re-serialize the file I guess we don't re-generate the DWARF.. but should be guard these kind of mutations with the kind of "only-if-we-are-writing-the-file" back out flag?

sbc100: Should we delay this change until `--post-emscripten` when other binaryen optimizations might have run which could increase the chances of removing this?

kripken: `eval-ctors` is worth doing, but not sure if I'll have time for it anytime soon.

Yes, this could be behind a check if `outfile` is defined. That would save a little work in the non-optimizing case, and also be clearer.

Yes, this should interact with modifying the wasm. On the emscripten side we will need to pass in `outfile` when optimizing.

Interesting idea to do this at `post-emscripten` time. Even further this could be done at metadce time, which is another option I considered. But the downside of doing it at those times is that we would need to have a mechanism to modify the JS to not call the ctor. Doing it here is simplest for that reason. But you're right that if the ctor has no effects only thanks to binaryen opts then we miss that.

sbc100 (approved):

juj:
Nice! Thanks for the productive Sunday <3

kripken:
Actually let me close this, as the metadce approach is superior. It is important to do this after optimizations run because linking modules together with LTO means that we need opts just to see the effects. That approach will take more work, though.

PR #2631, opened 2020-01-28
Add EH support for EffectAnalyzer

aheejin:
This adds EH support to `EffectAnalyzer`. Before `throw` and `rethrow`
conservatively set property. Now `EffectAnalyzer` has a new property
`mayThrow` to represent an expression that can throw, and expression
that can throw sets `mayThrow` correctly.

When EH is enabled, any calls can throw too, so we cannot reorder them
with another expression with any side effects, meaning all calls should
be treated in the same way as branches when evaluating `invalidate`.
This prevents many reorderings, so this patch sets `mayThrow` for calls
only when the exception handling features is enabled. This is also why I
passed `--disable-exception-handling` to `wasm2js` tests. Most of code
changes outside of `EffectAnalyzer` class was made in order to pass
`FeatureSet` to it.

`mayThrow` isn't always set whenever an expression contains a throwable
instruction. When an throwable instruction is within an inner try, it
will be caught by the corresponding inner catch, so it does not set
`mayThrow`.

PR #3370, opened 2020-11-15
[effects.h] Refactor hasGlobalSideEffects and throw handling. NFC

kripken:
The new `writesGlobalState` has a name that more clearly indicates what it
actually does. This removes `throw` from there, and handles it directly in the
single caller of the method.

Add `hasExternallyNoticeableEffects` which is what the old `hasGlobalSideEffects`
was maybe trying to be, but does it in a clearer way. There are currently no
users but a wip PR of mine will need this very soon.

I wrote this PR because that future PR tried to use `hasGlobalSideEffects` but
it really wasn't well-defined enough. The logic there is somewhat specific to
the single caller of the old method, I guess.

### DIFF
--- a/src/ir/effects.h
+++ b/src/ir/effects.h
@@ -140,7 +140,7 @@ struct EffectAnalyzer
 
   // Changes something in globally-stored state.
   bool writesGlobalState() const {
-    return globalsWritten.size() > 0 || writesMemory || isAtomic || calls;
+    return globalsWritten.size() || writesMemory || isAtomic || calls;
   }
   bool readsGlobalState() const {
     return globalsRead.size() || readsMemory || isAtomic || calls;
@@ -148,7 +148,7 @@ struct EffectAnalyzer
 
   // Whether this has any effect that could be noticeable from someone outside
   // the current function call. That does not include a write to a local, for
-  // example, but  does include any writes to global state as well as trapping
+  // example, but does include any writes to global state as well as trapping
   // and throwing. Note that this does not consider the return value from the
   // function (which we do not have access to directly here).
   bool hasExternallyNoticeableEffects() const {
@@ -217,7 +217,7 @@ struct EffectAnalyzer
     // function, so transfersControlFlow would be true) - while we allow the
     // reordering of traps with each other, we do not reorder exceptions with
     // anything.
-    assert(!((implicitTrap && other.throws) || (throws && other.implicitTrap)));
+    assert(!(implicitTrap && other.throws) && !(throws && other.implicitTrap));
     // We can't reorder an implicit trap in a way that could alter what global
     // state is modified.
     if ((implicitTrap && other.writesGlobalState()) ||
--- a/src/ir/effects.h
+++ b/src/ir/effects.h
@@ -138,21 +138,29 @@ struct EffectAnalyzer
     return branchesOut || throws || hasExternalBreakTargets();
   }
 
-  bool hasGlobalSideEffects() const {
-    return calls || globalsWritten.size() > 0 || writesMemory || isAtomic ||
-           throws;
+  // Changes something in globally-stored state.
+  bool writesGlobalState() const {
+    return globalsWritten.size() > 0 || writesMemory || isAtomic || calls;
+  }
+  bool readsGlobalState() const {
+    return globalsRead.size() || readsMemory || isAtomic || calls;
+  }
+
+  // Whether this has any effect that could be noticeable from someone outside
+  // the current function call. That does not include a write to a local, for
+  // example, but  does include any writes to global state as well as trapping
+  // and throwing. Note that this does not consider the return value from the
+  // function (which we do not have access to directly here).
+  bool hasExternallyNoticeableEffects() const {
+    return writesGlobalState() || implicitTrap || throws;
   }
   bool hasSideEffects() const {
-    return implicitTrap || localsWritten.size() > 0 || danglingPop ||
-           hasGlobalSideEffects() || transfersControlFlow();
+    return localsWritten.size() > 0 || danglingPop ||
+           hasExternallyNoticeableEffects() || transfersControlFlow();
   }
   bool hasAnything() const {
     return hasSideEffects() || accessesLocal() || readsMemory ||
-           accessesGlobal() || isAtomic;
-  }
-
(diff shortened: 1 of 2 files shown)


src/ir/effects.h
```diff
@@ -199,14 +207,21 @@ struct EffectAnalyzer
...
     }
-    // we can't reorder an implicit trap in a way that alters global state
-    if ((implicitTrap && other.hasGlobalSideEffects()) ||
-        (other.implicitTrap && hasGlobalSideEffects())) {
+    // Note that the above includes disallowing the reordering of a trap with an
+    // exception (as an exception can transfer control flow inside the current
+    // function, so transfersControlFlow would be true) - while we allow the
+    // reordering of traps with each other, we do not reorder exceptions with
+    // anything.
+    assert(!((implicitTrap && other.throws) || (throws && other.implicitTrap)));
```
aheejin, line 220:
```suggestion
    assert(!(implicitTrap && other.throws) && !(throws && other.implicitTrap));
```
Nit: I personally find this expansion easier to read..

src/passes/LoopInvariantCodeMotion.cpp
```diff
@@ -114,17 +114,20 @@ struct LoopInvariantCodeMotion
...
-                            effectsSoFar.invalidates(effects) ||
-                            (effects.noticesGlobalSideEffects() &&
-                             loopEffects.hasGlobalSideEffects());
+        bool unsafeToMove =
+          effects.writesGlobalState() || effectsSoFar.invalidates(effects) ||
+          (effects.readsGlobalState() && loopEffects.writesGlobalState());
+        // TODO: look into optimizing this with exceptions. for now, disallow
+        if (effects.throws || loopEffects.throws) {
+          unsafeToMove = true;
+        }
```
aheejin, line 130: So before it seemed that we allowed the current expression to be moved out of the loop even when the loop threw as long as the current expression does not read any global states, but we don't allow it anymore, right?

What we were doing before seems safe to me, but I guess the difference is pretty marginal anyway, but anyway it is not technically NFC then.. If it's intended this looks fine though.

aheejin (commented):
This layer of methods were somewhat confusing to me before, and this seems to add one more layer on that. Not that it's bad or anything, but I'd like to understand a little more on intended uses of each layer: `writesGlobalState`, `hasExternallyNoticeableEffects`, and `hasSideEffects`. And also why is `throws` separately?

> This removes `throw` from there, and handles it directly in the single caller of the method.

I'm not sure if I understand what this means... what is the "single caller"?

kripken:
The "single caller" is the single pass that is modified here, `LoopInvariantCodeMotion.cpp`.

src/passes/LoopInvariantCodeMotion.cpp
kripken: No, I think this is NFC. The old calls to `hasGlobalSideEffects` (for both `effects` and `loopEffects`) would check for `throws`. The refactoring removed that. So to make it NFC, I added them explicitly here.

aheejin: The old code checked for `loopEffects.hasGlobalSideEffects()` only when `effects.noticesGlobalSideEffects()` held; that's what I meant by 
> we allowed the current expression to be moved out of the loop even when the loop threw as long as the current expression did not read any global states

kripken: Oh, yes, sorry - you're right, I missed that.

I guess I can change that. But would you be ok with leaving it as a TODO for now? It seems low priority, and I do think we can allow full exceptions support there, once we spend the time, anyhow.

aheejin: Yeah that sounds fine. I was mostly just checking.

aheejin (approved):
LGTM, but I'd appreciate a little more explanation on this:
> This layer of methods was somewhat confusing to me before, and this seems to add one more layer on that. Not that it's bad or anything, but I'd like to understand a little more on intended uses of each layer: `writesGlobalState`, `hasExternallyNoticeableEffects`, and `hasSideEffects`. And also why is `throws` handled separately?

kripken:
> This layer of methods was somewhat confusing to me before, and this seems to add one more layer on that. Not that it's bad or anything, but I'd like to understand a little more on intended uses of each layer: writesGlobalState, hasExternallyNoticeableEffects, and hasSideEffects. And also why is throws handled separately?

If it's still not clear with the comments in this PR, then maybe this needs rethinking. I was hoping this was much clearer than before.

My idea here was that `writes/readsGlobalState` is self-explanatory: the global state is memory and globals (and the table, but nothing can change it yet).

And `hasExternallyNoticeableEffects` is any change observable from outside the current function call. That ignores a change to a local, but anything to global state, or trapping or exceptions.

And `hasSideEffects` is unchanged from before: that the expression has side effects of *some* form, noticeable in the current function even, so even a local change. If something has any type of side effect, it cannot be removed, in general.

If those don't make sense to you, maybe we can find something better?

kripken:
One option is that if `hasExternallyNoticeableEffects` is not clear, I can remove it, and figure something else out in the future PR that needs something like it.

aheejin:
Thanks, that's clearer. Any reason you renamed `hasGlobalSideEffects` to `hasExternallyNoticeableEffects`? (The semantics is different from the old `hasGlobalSideEffects` though) We use global- and local- in many places and this introduces a new name, so. (Also I'm little hazy on why we separately introduced `transfersControlFlow` even though apparently I was the one who did it.. I traced the reason back and it [says](https://github.com/WebAssembly/binaryen/pull/2631#issuecomment-580058339) "we talked offline" 😂... good old days when talking offline was a thing...) But anyway this PR LGTM.

kripken:
My goal wasn't to rename it, but to do something different, so I changed the name. But you're right that the name isn't good enough. I removed it, and I'll figure out something else in the future PR.

aheejin:
I didn't necessarily mean that was not clear enough.. I meant more like, it would be fine to change the existing `hasGlobalSideEffects`'s semantics as long as we are clear about the new meanings.

kripken:
I felt the old name was confusing :smile: so I preferred to change it. But without consensus I preferred not to add a new name. I do think I can avoid using it later anyhow.

PR #3373, opened 2020-11-16
[cost.h] Define costs for I31New and I31Get which the fuzzer emits

kripken:
cost.h now requires costs be defined, or it halts at runtime. But the fuzzer
already emits those two, so define them to unbreak the fuzzer.

Both access a reference, so they should be at least as costly as a load.
I31New also allocates, which is very cheap on a generational GC, so do not
add much for that.


kripken:
Landing quickly to unbreak fuzzer.

tlively:
LGTM

PR #3376, opened 2020-11-17
[OptimizeInstructions] Fix fuzz bug with shifts

kripken:
The code there looks for a "sign-extend": `(x << a) >> b` where the
right shift is signed. If `a = b = 24` for example then that is a sign
extend of an 8-bit value (it works by shifting the 8-bit value's sign bit
to the position of the 32-bit value's sign bit, then shifting all the way
back, which fills everything above 8 bits with the sign bit). The tricky
thing is that in some cases we can handle `a != b` - but we forgot a
place to check that. Specifically, a repeated sign-extend is not
necessary, but if the outer one has extra shifts, we can't do it.

This is annoyingly complex code, but for purposes of reviewing this
PR, you can see (unless I messed up) that the only change is to
ensure that when we look for a repeated sign extend, then we
only optimize that case when there are no extra shifts. And a
repeated sign-extend is obviously ok to remove,
```
(((x << a) >> a) << a) >> a   =>   (x << a) >> a
```

This is an ancient bug, showing how hard it can be to find certain
patterns either by fuzzing or in the real world...

Fixes https://github.com/WebAssembly/binaryen/issues/3362


tlively (commented):
Would it be possible to move the relevant code out of `properties.h` and into this file? I can't imagine this is used anywhere else and having all the code in the same place would make it easier to clean up and understand.

kripken:
I looked into it and it's not easy to move that code, as some of it is used in `PickLoadSigns`.

src/passes/OptimizeInstructions.cpp
```diff
@@ -488,11 +488,17 @@ struct OptimizeInstructions
...
         }
-        // if the sign-extend input cannot have a sign bit, we don't need it
-        // we also don't need it if it already has an identical-sized sign
-        // extend
-        if (Bits::getMaxBits(ext, this) + extraShifts < bits ||
-            isSignExted(ext, bits)) {
+        // We can in some cases remove part of a sign extend, that is,
+        //   (x << A) >> B   =>   x << (A - B)
+        // If the sign-extend input cannot have a sign bit, we don't need it.
+        if (Bits::getMaxBits(ext, this) + extraShifts < bits) {
```
tlively, line 494: Can you rename `extraShifts` to something like `extraLeftShifts`? IIRC, that's the direction of the extra shifts.

test/passes/optimize-instructions_fuzz-exec.txt
```diff
@@ -321,13 +324,35 @@
...
+     (i32.shr_s
+      (i32.shl
+       (local.get $0)
+       (i32.const 24)
+      )
+      (i32.const 24)
+     )
+     (i32.const 30)
+    )
+    (i32.const 24)
```
tlively, line 340: Wouldn't it be possible to turn this into a left shift by 6?

tlively (approved):
LGTM with that variable name change if my suggested further optimization would be incorrect or you don't want to do it in this PR.

test/passes/optimize-instructions_fuzz-exec.txt
kripken: No, because of the possible effective sign bit. Consider this:
```wat
00000100000
;; shift left by x+y (x, y > 0)
10000000000
;; signed shift right by y
11100000000

;; start again
00000100000
;; shift left by just x and that's it:
00100000000
```

PR #3381, opened 2020-11-17
[Types] Handle function types fully in more places

kripken:
Call `isFunction` to check for a general function type instead of just
a `funcref`, in places where we care about both, and some other minor
miscellaneous typing fixes in preparation for typed function references
(this will be tested fully at that time).

Change is mostly whitespace.

### DIFF
--- a/src/wasm-builder.h
+++ b/src/wasm-builder.h
@@ -762,9 +762,8 @@ public:
         return makeRefNull(type);
       case Type::i31ref:
         return makeI31New(makeConst(value.geti31()));
-      default: {
+      default:
         WASM_UNREACHABLE("invalid constant expression");
-      }
     }
   }
 
--- a/src/wasm-builder.h
+++ b/src/wasm-builder.h
@@ -741,24 +741,29 @@ public:
   // Make a constant expression. This might be a wasm Const, or something
   // else of constant value like ref.null.
   Expression* makeConstantExpression(Literal value) {
-    TODO_SINGLE_COMPOUND(value.type);
-    switch (value.type.getBasic()) {
-      case Type::funcref:
-        if (!value.isNull()) {
-          return makeRefFunc(value.getFunc());
-        }
-        return makeRefNull(value.type);
+    auto type = value.type;
+    if (type.isNumber()) {
+      return makeConst(value);
+    }
+    if (type.isFunction()) {
+      if (!value.isNull()) {
+        return makeRefFunc(value.getFunc(), type);
+      }
+      return makeRefNull(type);
+    }
+    TODO_SINGLE_COMPOUND(type);
+    switch (type.getBasic()) {
       case Type::externref:
       case Type::exnref: // TODO: ExceptionPackage?
       case Type::anyref:
       case Type::eqref:
         assert(value.isNull() && "unexpected non-null reference type literal");
-        return makeRefNull(value.type);
+        return makeRefNull(type);
       case Type::i31ref:
         return makeI31New(makeConst(value.geti31()));
-      default:
-        assert(value.type.isNumber());
-        return makeConst(value);
+      default: {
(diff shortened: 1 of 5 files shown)


src/wasm-builder.h
```diff
@@ -741,24 +741,29 @@ class Builder {
...
-          return makeRefFunc(value.getFunc());
-        }
-        return makeRefNull(value.type);
+    auto type = value.type;
+    if (type.isNumber()) {
+      return makeConst(value);
+    }
+    if (type.isFunction()) {
+      if (!value.isNull()) {
+        // TODO: with typed function references we need to do more for the type
```
tlively, line 750: What functionality does this new code add over the old code?

src/wasm/literal.cpp
```diff
@@ -422,58 +422,59 @@ void Literal::printVec128(std::ostream& o, const std::array<uint8_t, 16>& v) {
...
-      o << "i31ref(" << literal.geti31() << ")";
-      break;
-    case Type::unreachable:
-      WASM_UNREACHABLE("invalid type");
+  if (literal.type.isFunction()) {
+    if (literal.isNull()) {
+      o << "funcref(null)";
+    } else {
+      o << "funcref(" << literal.getFunc() << ")";
+    }
```
tlively, line 430: I guess we don't need to print the signature because it's redundant with the function name we print. Is that what you were thinking?

src/wasm/wasm-validator.cpp
```diff
@@ -795,10 +795,11 @@ void FunctionValidator::visitCall(Call* curr) {
         curr,
         "calls may only be unreachable if they have unreachable operands");
     } else {
-      shouldBeEqual(curr->type,
-                    target->sig.results,
-                    curr,
-                    "call type must match callee return type");
+      shouldBeSubTypeOrFirstIsUnreachable(
```
tlively, line 798: What is the benefit of allowing `Call` expressions to have imprecise types?

src/wasm/wasm-validator.cpp
```diff
@@ -2030,6 +2031,9 @@ void FunctionValidator::visitRefFunc(RefFunc* curr) {
   }
   auto* func = getModule()->getFunctionOrNull(curr->func);
   shouldBeTrue(!!func, curr, "function argument of ref.func must exist");
+  shouldBeTrue(curr->type.isFunction(),
+               curr,
+               "ref.func must have a function reference type");
```
tlively, line 2036: IIRC, the type should also not be nullable. It would be good to check that here.

aheejin (approved):

src/wasm-builder.h
kripken: Before, a function type would reach `TODO_SINGLE_COMPOUND` and trap. This does not trap, and creates a valid value (either a `ref.null` or `ref.func`).

src/wasm/literal.cpp
kripken: The printing is not changed here. This just moves the code around so that all `isFunction` types get printed properly. Before, we would trap on anything but a plain `funcref`.

(But also, it seems fine to not print the type, I think.)

src/wasm/wasm-validator.cpp
kripken: Oh, good point, it doesn't need to be a subtype...

src/wasm/wasm-validator.cpp
kripken: Adding, thanks.

tlively (approved):

kripken:
The extra non-nullability check caused a failure here. As discussed offline with @tlively , let's ignore non-nullability for now, and try to get everything else working. Then adding non-nullability will be a separate step after that.

Issue #3378, opened 2020-11-17
Should the fuzzer compare references?


kripken:
Consider this:
```wat
(module
 (func $0
  (nop)
 )
 (func $1
  (nop)
 )
 (func "export" (result funcref)
  (ref.func $1)
 )
)
```
Running duplicate function elimination on this, it will merge `$1` into `$0`. It will turn `(ref.func $1)` into `(ref.func $0)` as a result. But then the fuzzer will see that before opts this returned a reference to the name "1" and afterwards to "0" which looks like a bug.

But is it..? Calling the function will show no difference in behavior. OTOH, comparing references on the outside would notice a difference if we return `(ref.func $0)` from another funciton - before optimizations the two references are different, but not after.

If this is not a valid optimization, it seems like duplicate function elimination and other passes need to be very careful about any function whose reference is taken - similar to how they are careful about references to them in the table (but worse in some cases, as the table may be internal sometimes).

tlively:
This is related to the issue @RossTate raised in https://github.com/WebAssembly/design/issues/1351. IMO, this seems like one of those cases where theoretically the optimization is unsafe, but in practice we should do it anyway. Maybe we should control the behavior via a flag that Emscripten would turn on by default? Note that once you have `table.get`, it also becomes much harder to prove that changes even to internal tables will not become externally visible.

tlively:
We have settled on not comparing the contents of function references to preserve our ability to deduplicate or otherwise transform functions.

PR #3384, opened 2020-11-17
[Fuzzer] Don't compare references, just their types

kripken:
There are several issues here that we can't fully handle, see https://github.com/WebAssembly/binaryen/issues/3378, but basically
we are comparing results between two separate wasm modules (and
a separate instance of each) - we can't really identify an identical
reference between such things. We can only compare things structurally,
for which we compare the types.

This is a little unfortunate, but I don't see a better solution. For some
types of bugs we'll need another approach.

### DIFF
--- a/src/tools/execution-results.h
+++ b/src/tools/execution-results.h
@@ -123,9 +123,42 @@ struct ExecutionResults {
     ExecutionResults optimizedResults;
     optimizedResults.get(wasm);
     if (optimizedResults != *this) {
-      std::cout << "[fuzz-exec] optimization passes changed execution results";
+      std::cout << "[fuzz-exec] optimization passes changed results\n";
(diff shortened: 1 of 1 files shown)


tlively (approved):

aheejin (approved):
We used to do this to avoid this, but I guess your approach is better. But I'm wondering why the fuzzer errored with this..? Maybe because I only handled `funcref` and you are testing with other function references, right?

Anyway I guess if we go with your code it'd be better to delete my code here:
https://github.com/WebAssembly/binaryen/blob/a3d2a4d7854b8c14c7c7d222307b2af0c48b1275/src/tools/execution-results.h#L97-L104

kripken:
@aheejin Thanks! Removed.

PR #3387, opened 2020-11-18
[effects.h] Add a trap effect for unreachable

kripken:
We did not really model the effects of `unreachable` properly before. It
always traps, so it's not an implicit trap, but we didn't do anything but mark
it as "branches out", which is not really enough, as while yes it does
branch inside the current function, it also traps which is noticeable outside.

To fix that, add a `trap` effect to track this. `implicitTrap` will set `trap` as well,
automatically, if we do not ignore implicit traps, so it is enough to check just
that (unless one cares about the difference between implicit and explicit
ones).

### DIFF
--- a/src/ir/effects.h
+++ b/src/ir/effects.h
@@ -77,14 +77,9 @@ public:
   // each other, but it is not ok to remove them or reorder them with other
   // effects in a noticeable way.
   bool trap = false;
-
-private:
   // A trap from an instruction like a load or div/rem, which may trap on corner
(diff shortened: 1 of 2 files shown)


aheejin:
Then do we need `implicitTrap` as a separate variable at all? Nothing outside of `EffectAnalyzer` queries if this is an implicit trap or an explicit one.

kripken:
In theory some future pass might. I agree that's not likely, but I'm not sure it's worth removing now.

aheejin (approved):

PR #3388, opened 2020-11-18
[TypedFunctionReferences] Add Typed Function References feature and use the types

kripken:
This adds the new feature and starts to use the new types where relevant. We
use them even without the feature being enabled, as we don't know the features
during wasm loading - but the hope is that given the type is a subtype, it should
all work out. In practice, if you print out the internal type you may see a typed
function reference-specific type for a `ref.func` for example, instead of a generic
`funcref`, but it should not affect anything else.

This PR does **not** support non-nullable types, that is, everything is nullable
for now. As suggested by @tlively this is simpler for now and leaves nullability
for later work (which will apparently require `let` or something else, and many
passes may need to be changed).

To allow this PR to work, we need to provide a type on creating a `RefFunc`. The
`wasm-builder.h` internal API is updated for this, as are the C  and JS APIs,
which are breaking changes. cc @dcodeIO 

We must also write and read function types properly. This PR improves
`collectSignatures` to find all the types, and also to sort them by the
dependencies between them (as we can't emit X in the binary if it depends
on Y, and Y has not been emitted - we need to give Y's index). This sorting
ends up changing a few test outputs.

InstrumentLocals support for printing function types that are not `funcref`
is disabled for now, until we figure out how to make that work and/or
decide if it's important enough to work on.

The fuzzer has various fixes to emit valid types for things (mostly
whitespace there). Also two drive-by fixes to call `makeTrivial` where it
should be (when we fail to create a specific node, we can't just try to make
another node, in theory it could infinitely recurse).

Binary writing changes here to replace calls to a standalone function to
write out a type with one that is called on the binary writer object itself,
which maintains a mapping of type indexes (`getFunctionSignatureByIndex`).

This looks ok after fuzzing.


src/ir/module-utils.h
```diff
@@ -444,10 +461,57 @@ collectSignatures(Module& wasm,
       counts[innerPair.first] += innerPair.second;
     }
   }
+  // We must sort all the dependencies of a signature before it. For example,
```
tlively, line 464: I don't think this is actually true in the proposal, but I opened https://github.com/WebAssembly/function-references/issues/41 to clarify that. I do recognize that this is required by the current parser in Binaryen, though, because I haven't done the work to allow using undefined types in type definitions yet. I will hop on that.

src/ir/module-utils.h
```diff
@@ -444,10 +461,57 @@ collectSignatures(Module& wasm,
...
+  // To calculate the depth of dependencies, we'll do a flow analysis, visiting
+  // each signature as we find out new things about it.
+  std::set<Signature> toVisit;
+  for (auto& pair : counts) {
+    auto sig = pair.first;
+    depthOfDependencies[sig] = 0;
+    toVisit.insert(sig);
+    for (Type type : {sig.params, sig.results}) {
+      for (auto subType : type) {
+        if (subType.isRef()) {
```
tlively, line 480: Perhaps "component" or "element" rather than "subtype" here, since this isn't actually about subtyping.

src/ir/module-utils.h
```diff
@@ -444,10 +461,57 @@ collectSignatures(Module& wasm,
...
+  }
+  // Sort by frequency and then simplicity, and also keeping every signature
+  // before things that depend on it.
   std::vector<std::pair<Signature, size_t>> sorted(counts.begin(),
                                                    counts.end());
   std::sort(sorted.begin(), sorted.end(), [&](auto a, auto b) {
-    // order by frequency then simplicity
+    if (depthOfDependencies[a.first] != depthOfDependencies[b.first]) {
+      return depthOfDependencies[a.first] < depthOfDependencies[b.first];
+    }
```
tlively, line 514: It would be more optimal to calculate the transitive closure of the dependency graph and use that directly in the sort function rather than approximating it with dependency path lengths, but perhaps it doesn't matter too much if we're going to remove all this code once we support building recursive types.

Are there other places where we need to calculate dependency graphs or topological sorts? I'm wondering if it would be worthwhile to pull this out as a generic graph processing routine.

src/passes/InstrumentLocals.cpp
```diff
@@ -135,45 +135,48 @@ struct InstrumentLocals : public WalkerPass<PostWalker<InstrumentLocals>> {
...
-        case Type::i31ref:
-          import = set_i31ref;
-          break;
-        case Type::unreachable:
-          return; // nothing to do here
-        default:
-          WASM_UNREACHABLE("unexpected type");
-      }
+    if (type.isFunction() && type != Type::funcref) {
+      // FIXME: support typed function references
```
tlively, line 139: Why not continue aborting in this case since it still isn't implemented?

src/tools/fuzzing.h
```diff
@@ -323,28 +323,33 @@ class TranslateToFuzzReader {
...
-    switch (type.getBasic()) {
-      case Type::anyref:
-        if (wasm.features.hasReferenceTypes()) {
-          options.push_back(Type::funcref);
-          options.push_back(Type::externref);
-          if (wasm.features.hasExceptionHandling()) {
-            options.push_back(Type::exnref);
+    if (type.isFunction()) {
+      // TODO: specific typed function references types
+      options.push_back(Type::funcref);
```
tlively, line 328: What's this change for? If `type` is a typed function reference, then this isn't correct. If `type` is a funcref, then this change isn't necessary.

src/tools/fuzzing.h
```diff
@@ -2018,12 +2026,25 @@ class TranslateToFuzzReader {
...
       if (type == Type::i31ref) {
         return builder.makeI31New(makeConst(Type::i32));
       }
-      return builder.makeRefNull(type);
+      if (type.isNullable()) {
+        return builder.makeRefNull(type);
+      }
+      // Well this is quite the pickle, we are asked to create a non-nullable
+      // type, and so can't just make a null. Look for an existing thing with
+      // the right type.
```
tlively, line 2040: Can we just change the fuzzer to never request a non-nullable type?

src/tools/fuzzing.h
```diff
@@ -2018,12 +2026,25 @@ class TranslateToFuzzReader {
...
+        auto type = Type(HeapType(target->sig), /* nullable = */ true);
+        return builder.makeRefFunc(target->name, type);
       }
       if (type == Type::i31ref) {
         return builder.makeI31New(makeConst(Type::i32));
       }
-      return builder.makeRefNull(type);
+      if (type.isNullable()) {
+        return builder.makeRefNull(type);
+      }
```
tlively, line 2037: Why prefer to create a null over any other constant value?

src/tools/fuzzing.h
```diff
@@ -2997,12 +3018,27 @@ class TranslateToFuzzReader {
...
-    for (size_t i = 0; i < numElements; ++i) {
-      elements[i] = getSingleConcreteType();
+    size_t maxElements = 2 + upTo(MAX_TUPLE_SIZE - 1);
+    for (size_t i = 0; i < maxElements; ++i) {
+      auto type = getSingleConcreteType();
+      // Don't add ref types into a tuple for now, as if they are non-nullable
+      // then they can't be spilled into locals. And even if they are nullable,
+      // like a funcref, but we create a subtype that is *not* nullable (like
+      // a ref.func) then that would be the effective type here, and again we'd
+      // have a non-nullable type.
```
tlively, line 3035: I recommend disallowing `funcref` everywhere because it is non-nullable and enforcing the use of `(ref null func)` instead.

src/wasm/wasm-binary.cpp
```diff
@@ -1581,6 +1682,18 @@ void WasmBinaryBuilder::readFunctionSignatures() {
...
 }
 
+Signature WasmBinaryBuilder::getFunctionSignatureByIndex(Index index) {
+  Signature sig;
+  if (index < functionImports.size()) {
+    return functionImports[index]->sig;
+  }
+  Index adjustedIndex = index - functionImports.size();
+  if (adjustedIndex >= functionSignatures.size()) {
+    throwError("invalid call index");
```
tlively, line 1692: I know this is pre-existing, but
```suggestion
    throwError("invalid function index");
```

src/wasm/wasm-binary.cpp
```diff
@@ -5169,7 +5275,10 @@ void WasmBinaryBuilder::visitRefFunc(RefFunc* curr) {
     throwError("ref.func: invalid call index");
   }
   functionRefs[index].push_back(curr); // we don't know function names yet
-  curr->finalize();
+  // To support typed function refs, we give the reference not just a general
+  // funcref, but a specific subtype with the actual signature.
+  curr->finalize(
+    Type(HeapType(getFunctionSignatureByIndex(index)), /* nullable = */ true));
```
tlively, line 5281: Shouldn't `ref.func` have a non-nullable type? I believe this should be the case even if we disallow non-nullable types everywhere else.

src/wasm/wasm-type.cpp
```diff
@@ -460,6 +460,13 @@ Type Type::reinterpret() const {
 
 FeatureSet Type::getFeatures() const {
   auto getSingleFeatures = [](Type t) -> FeatureSet {
+    if (t != Type::funcref && t.isFunction()) {
+      // Strictly speaking, typed function references require the typed function
+      // references feature, however, we use these types internally regardless
+      // of the presence of features (in particular, since during load of the
+      // wasm we don't know the features yet, so we apply the "higher" types).
```
tlively, line 467: "higher" => "more refined"

src/wasm/wasm-validator.cpp
```diff
@@ -2313,6 +2313,7 @@ void FunctionValidator::visitFunction(Function* curr) {
   for (const auto& var : curr->vars) {
     features |= var.getFeatures();
     shouldBeTrue(var.isConcrete(), curr, "vars must be concretely typed");
+    // TODO: check for nullability
```
tlively, line 2316: We should enforce that locals are non-nullable now, as well as enforcing that types in block signatures and function signatures are non-nullable.

test/binaryen.js/kitchen-sink.js
```diff
@@ -539,8 +539,8 @@ function test_core() {
     // Reference types
     module.ref.is_null(module.ref.null(binaryen.externref)),
     module.ref.is_null(module.ref.null(binaryen.funcref)),
-    module.ref.is_null(module.ref.func("kitchen()sinker")),
-    module.select(temp10, module.ref.null(binaryen.funcref), module.ref.func("kitchen()sinker"), binaryen.funcref),
+    module.ref.is_null(module.ref.func("kitchen()sinker", binaryen.funcref)),
```
tlively, line 542: We should prefer to give `ref.func` a precise type rather than just `funcref`, to the point that we might want to consider enforcing that in the API. Being imprecise with the type always leaves useful information on the table.

test/passes/translate-to-fuzz_all-features.txt
```diff
@@ -1,24 +1,23 @@
 (module
  (type $none_=>_none (func))
+ (type $funcref_f64_=>_i32 (func (param funcref f64) (result i32)))
```
tlively, line 3: We should really replace this test with something that's more precise about what it is we're actually testing.

src/wasm/wasm-binary.cpp
```diff
@@ -1253,6 +1348,12 @@ Type WasmBinaryBuilder::getType() {
       return Type::anyref;
     case BinaryConsts::EncodedType::eqref:
       return Type::eqref;
+    case BinaryConsts::EncodedType::nullable:
+      return Type(getHeapType(), /* nullable = */ true);
+    case BinaryConsts::EncodedType::nonnullable:
+      // FIXME: support non-nullable types. search for all "nullable = "
+      //        comments.
+      return Type(getHeapType(), /* nullable = */ true);
```
tlively, line 1356: I think it would be better to error out here rather than silently change the users' program.

kripken: I thought the plan we agreed on was to just do everything with nullable types for now, which is definitely broken, but it's a stepping stone? Erroring out makes sense in production-ready code, I agree, but this is all very much wip.

src/wasm/wasm-binary.cpp
kripken: As in the previous comment, I thought we agreed we would only support nullable types for now as a first step? Allowing them on `ref.func` would mean allowing them in enough places as to be as hard to support as to support them everywhere. (That is, a single `ref.func` introduces a non-nullable type, which then cannot be spilled to a local, which then...)

test/binaryen.js/kitchen-sink.js
kripken: Agreed, but there is no way to do that atm in the JS API?

dcodeIO: Yes, the JS API only has a mechanism to create Tuple types currently, via `createType` (and expanding via `expandType`). Something similar will be needed for functions, arrays and structs eventually. Perhaps wrappers like `TupleType` (backed by what's currently `createType`), `FunctionType`, `ArrayType` and `StructType` on top of the respective new C API?

kripken: Do you mean `createType(SignatureType(..)) as opposed to `createSignatureType(..)`? I don't have an opinion myself, not sure what's more JS-ey...

test/passes/translate-to-fuzz_all-features.txt
kripken: This test could possibly just print out the metrics, and not the actual body. Seems like that's good enough I can do that separately.

src/wasm/wasm-binary.cpp
tlively: I agree that we should only support nullable types for now, but IMO that should include parsing as well. In general, this makes our level of support explicit to our users and will help us find the places we need to change later when we do want to support nullable types.

test/binaryen.js/kitchen-sink.js
tlively: > Agreed, but there is no way to do that atm in the JS API?

Ah, maybe that would be good to fix first, then. I don't feel too strongly about it, though.

src/passes/InstrumentLocals.cpp
tlively: I see. How about checking the features and aborting only if `typed-function-references` is enabled?

src/ir/module-utils.h
tlively: Talked about this offline - the missing piece was that it is important for this code to order types by usage to minimize the cumulative size of lebs in type annotations. Once we support recursive types (and more generally, once we support referring to undefined types in type definitions) this ordering code will no longer be necessary.

src/wasm/wasm-binary.cpp
kripken: Do you mean parsing of custom user types (not parsing of `ref.func` etc.)? I guess there's no harm to allow non-nullable types there for now, as long as we don't test them, sgtm.

src/passes/InstrumentLocals.cpp
kripken: Doing that would still break people running with `-all`, including our existing tests. I can work around it in the tests, but I'm not sure it's worth it?

test/binaryen.js/kitchen-sink.js
kripken: I don't think it should block landing this - we can get very far in testing and using typed function references, and GC, without C or JS API. But I agree it is necessary, obviously.

kripken:
Updated to properly parse non-nullable types in the binary format (we will almost certainly error if they are used in a binary, and they are not tested).

src/wasm/wasm-binary.cpp
```diff
@@ -950,6 +952,99 @@ void WasmBinaryWriter::finishUp() {
...
+void WasmBinaryWriter::writeType(Type type) {
+  if (type.isRef()) {
+    auto heapType = type.getHeapType();
+    if (heapType.isSignature()) {
+      if (type.isNullable()) {
+        o << S32LEB(BinaryConsts::EncodedType::nullable);
+      } else {
+        o << S32LEB(BinaryConsts::EncodedType::nonnullable);
+      }
+      writeHeapType(heapType);
```
aheejin, line 964: We don't call `writeHeapType` when `heapType` is not a signature?

aheejin (commented):
Haven't read the whole PR yet, but some basic questions to improve my understanding:
- According to the [spec](https://github.com/WebAssembly/function-references/blob/master/proposals/function-references/Overview.md), regardless of the current Binaryen implementation, is `ref.func`'s return type a nullable type or a non-nullable type?
- Why is the feature set not available at loading type?
- What do we do if we assume more refined types for function references, and later it turns out we only have the basic reference types enabled but not typed function references?

src/ir/module-utils.h
```diff
@@ -414,7 +414,16 @@ collectSignatures(Module& wasm,
...
+        // A signature type could be anywhere.
+        if (curr->type.isRef()) {
+          auto heapType = curr->type.getHeapType();
+          if (heapType.isSignature()) {
+            counts[heapType.getSignature()]++;
+          }
+        }
+        // Specific places also have a special signature.
         if (auto* call = curr->dynCast<CallIndirect>()) {
           counts[call->sig]++;
```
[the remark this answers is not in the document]
kripken: Yes, each param can be a typed function reference type. But I'm not sure what you mean in the second sentence - what do you mean by "collect signatures"? (I think this is the one place that collects them, so there is no other place to worry about?)

src/wasm/wasm-binary.cpp
kripken: I think maybe we should for GC. But for the typed function references proposal, all we have are signatures.

Maybe there's no harm in generalizing it here though, I'll update with that. Seems like it can't hurt, may save time later, and is shorter anyhow.

kripken:
> According to the spec, regardless of the current Binaryen implementation, is ref.func's return type a nullable type or a non-nullable type?

I think the spec says it is non-nullable. That is, it takes advantage of the fact that it definitely is not null when you take a reference to an actual function by name.

>  Why is the feature set not available at loading type?

It's a little complicated because of the feature section, which applies features, and that the user can also apply features through flags. My initial thought was to apply the user features before loading the wasm - so if the user said "use X", we'd know X is present - but @tlively pointed out that that would affect how the features section works. Specifically, we want to error if the user flags don't match the feature section flags, and if we apply the user flags first to the wasm, we'd need to change how we do that comparison (because we look at what's applied on the wasm).

It seems like there is complexity both ways. I'm not 100% sure what's best to be honest.

> What do we do if we assume more refined types for function references, and later it turns out we only have the basic reference types enabled but not typed function references?

It should "just work". Without the newer feature for typed function references enabled, there is no instruction that can notice the difference. Or so at least we hope :smile: It definitely has observable effects for binaryen, though. For example, if a pass wants to save a value to a local, and creates a local of that type, then that is always valid with reference types (it's ok to make a local of type `funcref`) but **not** ok if it's a non-nullable typed function reference type (since a local has an initial value which would be null). But, for now we are ignoring non-nullable types anyhow for other reasons - that they require `let` and other work, that maybe we don't need immediately - so this is hopefully not an issue.

But this does add some oddness, to use newer types than the features. But avoiding this complexity would require a  change to the features section handling, as in the last answer...

src/ir/module-utils.h
```diff
@@ -414,7 +414,16 @@ collectSignatures(Module& wasm,
       Counts& counts;
 
       TypeCounter(Counts& counts) : counts(counts) {}
+
       void visitExpression(Expression* curr) {
+        // A signature type could be anywhere.
```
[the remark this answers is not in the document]
tlively: Since this type counting and sorting is meant to be a binary size optimization, I still think that we should only be counting the locations in which a type index needs to be written into the binary.

src/passes/InstrumentLocals.cpp
tlively: Ok, that's reasonable. I still think this isn't great, but I agree that having to change all the tests to not use `-all` is worse, so there's not anything to be done about it.

src/tools/fuzzing.h
tlively: As a reader, it can be hard to tell what is supposed to work when the code is future-facing like this. I think it would be clearer to draw a bright line and not try to anticipate our future needs.

tlively: I guess I want to go beyond not supporting non-nullability and not even mention it in the code. Perhaps this would be beyond the scope of this PR, but it would be good to remove the concept of nullability from wasm-type.h completely, now that we've decided not to support it in the near future.

kripken: I think we have a clear line: non-nullable types do not work, for now. And I think we know that the future needs include non-nullability, don't we?

(If somehow the wasm spec decides to remove non-nullability, then at that time of course I would agree to remove all the relevant code.)

Assuming we agree on those, I disagree that means we should delete all the nullability code from wasm-type.h, and other places with nullability code in this PR. We will need to support non-nullability pretty soon. Removing that code and adding it all back later would be a significant amount of extra work, and the only benefit seems to be to keep the code more "clean" in the short interval in between?

Maybe I'm more of a pragmatist than you. To me, the long-term goal is to support non-nullability, and I'm just trying to find the shortest path to there. Especially since "long-term" here may be weeks, I am not trying to optimize for code clarity in those interim weeks necessarily. I agree with you that that is a compromise on code clarity meanwhile, but I think we'll get to perfect clarity eventually (and I agree that that matters!). Does that make sense?

src/ir/module-utils.h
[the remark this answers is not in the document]
kripken: Thanks, right, this is counting unnecessary things atm. Pushing a fix now.

src/tools/fuzzing.h
tlively: I see. I was thinking that we might want to consider supporting non-nullability in half a year or so :P

src/ir/module-utils.h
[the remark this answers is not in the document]
kripken: This turned out to be a little tricky, but I think I have it fixed now.

src/ir/module-utils.h
```diff
@@ -416,20 +416,27 @@ collectSignatures(Module& wasm,
       TypeCounter(Counts& counts) : counts(counts) {}
 
       void visitExpression(Expression* curr) {
-        if (curr->is<RefNull>()) {
-          auto heapType = curr->type.getHeapType();
-          if (heapType.isSignature()) {
-            counts[heapType.getSignature()]++;
-          }
+        if (curr->is<RefNull>() || curr->is<RefFunc>()) {
```
tlively, line 419: It doesn't look like `ref.func` has a type annotation: https://webassembly.github.io/reference-types/core/binary/instructions.html#reference-instructions/

kripken: I may be confused. While it doesn't have a type in the binary directly, it does potentially define a new type - say, a nullable reference to a function of type X, which differs from the type X itself. That new type may appear in no other places. Is that not a problem?

kripken: Seems like it shouldn't be, as you say. Fuzzer agrees. Fixed!

src/wasm/wasm-binary.cpp
kripken: I had to revert that - it turns out we already have a testcase in the test suite that breaks because of the change, presumably because we need to fix the other side (reading and writing) as well. Let's leave GC for later then.

src/tools/fuzzing.h
tlively: If I understand this code correctly, `ref.func` of any function with the correct signature would also be an option for nullable typed function reference types, which you take advantage of in the next loop. But if the type is nullable, that loop will never run because you always choose `ref.null` here.

src/tools/fuzzing.h
tlively: I was wrong when I wrote that `funcref` is non-nullable, and I think our offline discussions have resolved the disconnect here. Still, could you simplify this code by just having `getSingleConcreteType` return only nullable reference types? I don't understand the issue you describe here:

>  And even if they are nullable, like a funcref, but we create a subtype that is *not* nullable (like a ref.func) then that would be the effective type here, and again we'd have a non-nullable type.

src/wasm/wasm-binary.cpp
tlively: I believe we resolved this offline. We agreed that it was simpler to pretend that even producers of non-nullable types produce nullable types for now, rather than apply that relaxation only to consumers of non-nullable types, as I had been envisioning.

src/tools/fuzzing.h
tlively: We resolved this in our offline discussion. We will land the forward-facing code for now, but we will go back and remove it if it looks like we won't implement nullable types for a long time.

src/ir/module-utils.h
tlively: Resolved offline.

src/ir/module-utils.h
tlively: Yeah, it's kind of weird that there are runtime types that do not appear in the type section, now that I think about it. Glad the fuzzer agrees that this is correct, though!

src/wasm/wasm-binary.cpp
tlively: > I thought the plan we agreed on was to just do everything with nullable types for now.

I think we interpreted that differently. My interpretation was that we should only _support_ nullable types for now, which to me means erroring out when encountering a non-nullable type in the user input. It looks like your interpretation allows for "supporting" non-nullable types from the user's point of view by treating them as nullable types. I feel less strongly about this particular instance now because some of the implementation strategies we were brainstorming earlier did involve treating non-nullable types as nullable in Binaryen IR, but on a related note, do you plan to add validation that local, block, and function types do not contain non-nullable references in this PR?

src/wasm/wasm-validator.cpp
aheejin: Why should locals be non-nullable? Shouldn't they be rather nullable, given that they need a null initialization value?

src/wasm/wasm-binary.cpp
aheejin: That sounds fine; it'd be better to leave a TODO here maybe, because when we currently call it only for signatures.

src/ir/module-utils.h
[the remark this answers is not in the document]
aheejin: There are several places we count signatures, each of which consists of params and results, and each param or result can be a typed reference type that contains a signature inside. What I meant was, when we count signatures call `call_indirect`, block types, or loop types, ..., should we check if each param/result is in turn also a typed reference type, in case it contains a signature? Maybe I'm misunderstanding something.

aheejin:
> > Why is the feature set not available at loading type?
> 
> It's a little complicated because of the feature section, which applies features, and that the user can also apply features through flags. My initial thought was to apply the user features before loading the wasm - so if the user said "use X", we'd know X is present - but @tlively pointed out that that would affect how the features section works. Specifically, we want to error if the user flags don't match the feature section flags, and if we apply the user flags first to the wasm, we'd need to change how we do that comparison (because we look at what's applied on the wasm).
> 
> It seems like there is complexity both ways. I'm not 100% sure what's best to be honest.

Sorry I think I still lack some understanding 😅 So we can know user flags before loading, but we can't know what feature section contains before loading, is this what you're saying? If so, is that because feature sections are read at the end?

And I'm not sure what you mean by when user flags and the feature section do not match. Do we error out? If so, can't we just use user-given features anyway from the beginning, given that we are gonna error out if they turn out to be incorrect later?

Also, I'm not sure what you mean by "both ways". Sorry for so many questions...


> > What do we do if we assume more refined types for function references, and later it turns out we only have the basic reference types enabled but not typed function references?
> 
> It should "just work". Without the newer feature for typed function references enabled, there is no instruction that can notice the difference. Or so at least we hope 😄 It definitely has observable effects for binaryen, though. For example, if a pass wants to save a value to a local, and creates a local of that type, then that is always valid with reference types (it's ok to make a local of type `funcref`) but **not** ok if it's a non-nullable typed function reference type (since a local has an initial value which would be null). But, for now we are ignoring non-nullable types anyhow for other reasons - that they require `let` and other work, that maybe we don't need immediately - so this is hopefully not an issue.
> 
> But this does add some oddness, to use newer types than the features. But avoiding this complexity would require a change to the features section handling, as in the last answer...

Isn't the output (both binary and text) different for `funcref` and typed function references? If we don't have TFR enabled but we have TFRs in-memory, aren't they gonna make differences when they are finally written out to output files? For example, `writeType` in wasm-binary.cpp doesn't seem to check if TFR feature is enabled. It seems I am not understanding some core assumptions..

kripken:
@tlively 

> do you plan to add validation that local, block, and function types do not contain non-nullable references in this PR?

I didn't plan on adding anything else to this PR :smile: My goal here is not to be 100% comprehensive. We have some unknowns about our plans about about the spec, so doing it all here would just add unnecessary work now or require unnecessary work later to remove.

What I see this PR as doing is adding the minimal amount of support for the typed function references feature, and use the types, and get the fuzzer passing (and maybe doing a little more, if it's already been done, and is code that is easier to leave in than do the work to remove it and remember to add it later). That leaves quite a lot of future work, but unblocks work on `call_ref` and on GC, which is my main goal.

I don't see thiis PR as creating technical debt or otherwise slowing us down later. I'd agree that would be something to avoid, if I missed such a thing here.

@aheejin 

> Sorry I think I still lack some understanding 😅 So we can know user flags before loading, but we can't know what feature section contains before loading, is this what you're saying? If so, is that because feature sections are read at the end?

Probably @tlively can explain that better as I don't fully understand how our feature section handling works.

However, I'd like to leave that as a separate question from this PR, if that's ok? If we decide to change the approach (and check the feature before using the types, which means changing how we use the feature section) that would not be hard to do later, but would be a larger discussion on changing how the feature section works (which I'd rather not block GC work on).

> Isn't the output (both binary and text) different for funcref and typed function references? If we don't have TFR enabled but we have TFRs in-memory, aren't they gonna make differences when they are finally written out to output files? For example, writeType in wasm-binary.cpp doesn't seem to check if TFR feature is enabled. It seems I am not understanding some core assumptions..

It does seem like it shouldn't work :wink: but it does, I think. If TFR is disabled then a `ref.func` would still have a more refined type that is part of TFR. But that is not actually written to the binary - luckily, the binary format is concise and does not write a type for `ref.func`. The type is not needed since we know the type of the function already. And it seems to work out for `ref.null` too, which **does** provide a type, but if TFR is disabled then that type would not be a TFR type - we don't create a new type in-memory there.

kripken:
@tlively 

Maybe a shorter way to state my position: I don't consider this PR to be "correct" in terms of nullability. I'm suggesting that we don't care about nullability for now, and leave it for later. The benefit is that that can save us work given the uncertainties here. But I sense that you want there to be a proper mental model for nullability even in this PR? If so, the options are to either remove it entirely, or something else that would be more work either now or later, I'm pretty sure. So I hope we can agree to having "wrong" nullability for now,

tlively:
It's definitely not the way I would have written it, but I see where you're coming from :) I know you're not about to stop work on this and leave nullability half-done, so I'm fine with landing it this way. I'll give it one last pass through to make sure I don't have any other concerns.

tlively (approved):

tlively:
Looks like the fuzz test just has to be updated again.

@aheejin, I posted #3393 to more fully explain what we are currently doing with features and start a discussion about whether we should make any changes to how we handle them.

aheejin:
@kripken
> > According to the spec, regardless of the current Binaryen implementation, is ref.func's return type a nullable type or a non-nullable type?
> 
> I think the spec says it is non-nullable. That is, it takes advantage of the fact that it definitely is not null when you take a reference to an actual function by name.

Then is it OK make it a non-nullable type in the implementation? I agree we should start with non-nullable types, but I'm just wondering if this is spec-compliant. For example, this does not seem to be valid in the spec, but we might treat this valid..? 
```wasm
(ref.as_non_null (ref.func $foo))
```
I guess it doesn't matter now because we don't have `ref.as_non_null` yet in the impl, but I'm wondering generally making `ref.func`'s return type more wider (i.e., supertype) of its real return type would be OK.

---

> > Isn't the output (both binary and text) different for funcref and typed function references? If we don't have TFR enabled but we have TFRs in-memory, aren't they gonna make differences when they are finally written out to output files? For example, writeType in wasm-binary.cpp doesn't seem to check if TFR feature is enabled. It seems I am not understanding some core assumptions..
> 
> It does seem like it shouldn't work 😉 but it does, I think. If TFR is disabled then a `ref.func` would still have a more refined type that is part of TFR. But that is not actually written to the binary - luckily, the binary format is concise and does not write a type for `ref.func`. The type is not needed since we know the type of the function already. And it seems to work out for `ref.null` too, which **does** provide a type, but if TFR is disabled then that type would not be a TFR type - we don't create a new type in-memory there.

So you mean, even though we don't know the feature set while reading a module (I asked a question above that below), we know the feature set when we write back the module to either text of binary, so we can emit the write types that conform to the current feature set? That makes sense, thanks.

---

@kripken and @tlively 
My question is, regardless of how we should handle features input, either from user or the features section or both, can't we determine the feature set for Binaryen before loading the module? @kripken said this PR was done under the assumption that we don't know the feature set while we read a module. I'm wondering why that's the case. (I also read #3393, but I'm still not sure why all those feature set decision can't precede module loading)

tlively:
> My question is, regardless of how we should handle features input, either from user or the features section or both, can't we determine the feature set for Binaryen before loading the module?

No, because the feature set is determined in part by the target features section. The only way to read the target features section right now is to load the module.

aheejin:
> > My question is, regardless of how we should handle features input, either from user or the features section or both, can't we determine the feature set for Binaryen before loading the module?
> 
> No, because the feature set is determined in part by the target features section. The only way to read the target features section right now is to load the module.

So we parse and read code before reading the features section? Is that because the features section occurs after the code section?

tlively:
> > > According to the spec, regardless of the current Binaryen implementation, is ref.func's return type a nullable type or a non-nullable type?
> > 
> > 
> > I think the spec says it is non-nullable. That is, it takes advantage of the fact that it definitely is not null when you take a reference to an actual function by name.
> 
> Then is it OK make it a non-nullable type in the implementation? I agree we should start with non-nullable types, but I'm just wondering if this is spec-compliant.

(I think you wrote "non-nullable" in this message where you meant "nullable". Is that correct?)

Since non-nullable references are subtypes of their corresponding nullable references, they can be used anywhere the nullable references can be used. That means that treating instructions that produce non-nullable references as though they produced nullable references only lets them be used in locations where they could already be used. It therefore does not allow any invalid modules to validate.

> For example, this does not seem to be valid in the spec, but we might treat this valid..?
> 
> ```webassembly
> (ref.as_non_null (ref.func $foo))
> ```

This is valid in the spec because `ref.func` produces a non-nullable reference, which is a subtype of a nullable reference and can therefore be passed to `ref.as_non_null`, which expects a nullable reference.

tlively:
> > > My question is, regardless of how we should handle features input, either from user or the features section or both, can't we determine the feature set for Binaryen before loading the module?
> > 
> > 
> > No, because the feature set is determined in part by the target features section. The only way to read the target features section right now is to load the module.
> 
> So we parse and read code before reading the features section? Is that because the features section occurs after the code section?

Yes, that's correct. In theory we could skip ahead and read the target features section first, but that wouldn't buy us much because using the more refined types internally works out fine.

kripken:
(I do suspect this is not spec-compliant for some reason or other, detectable in a spec test, but this is wip - we'll fix those issues later.)

PR #3389, opened 2020-11-19
[wasm-split] Initial instrumentation

tlively:
Implement an instrumentation pass that records the timestamp at which each
defined function is first called. Timestamps are not actual time, but rather
snapshots of a monotonically increasing counter. The instrumentation exports a
function that the embedder can call to dump the profile data into a memory
buffer at a given offset and size. The function returns the total size of the
profile data so the embedder can know how much to read out of the buffer or how
much it needs to grow the buffer.

Parsing and using the profile is left as future work, as is recording a hash of
the input file that will be used to guard against accidentally instrumenting one
module and trying to use the resulting profile to split a different module.

Since GitHub doesn't support stacked diffs yet this PR includes content from #3359.
Please review only the commits beginning with the eponymous commit.

cc @sbc100 @awtcode


src/tools/wasm-split.cpp
```diff
@@ -245,8 +261,195 @@ void parseInput(Module& wasm, const WasmSplitOptions& options) {
   options.applyFeatures(wasm);
 }
 
+// Add a global monotonic counter and a timestamp global for each function, code
+// at the beginning of each function to set its timestamp, and a new exported
+// function for dumping the profile data.
+struct FuncInstrumentor : public Pass {
```
kripken, line 267: This name seems too generic - how about `TimeInstrumentor`?

src/tools/wasm-split.cpp
```diff
@@ -245,8 +261,195 @@ void parseInput(Module& wasm, const WasmSplitOptions& options) {
...
+  counterGlobal = Names::getValidGlobalName(*wasm, "monotonic_counter");
+  functionGlobals.reserve(wasm->functions.size());
+  ModuleUtils::iterDefinedFunctions(*wasm, [&](Function* func) {
+    functionGlobals.push_back(Names::getValidGlobalName(
+      *wasm, std::string(func->name.c_str()) + "_timestamp"));
+  });
+
+  // Create and add new globals
+  auto addGlobal = [&](Name name) {
+    auto global = std::make_unique<Global>();
```
kripken, line 308: please use `builder.makeGlobal`

src/tools/wasm-split.cpp
```diff
@@ -245,8 +261,195 @@ void parseInput(Module& wasm, const WasmSplitOptions& options) {
...
+  // Inject the following code at the beginning of each function to advance the
+  // monotonic counter and set the function's timestamp if it hasn't already
+  // been set.
+  //
+  //   (if (i32.eqz (global.get $timestamp))
+  //     (block
+  //       (global.set $monotonic_counter
+  //         (i32.add
+  //           (global.get $monotonic_counter)
+  //           (i32.const)
```
kripken, line 332: missing `1`

src/tools/wasm-split.cpp
```diff
@@ -245,8 +261,195 @@ void parseInput(Module& wasm, const WasmSplitOptions& options) {
...
+  //     )
+  //   )
+  Builder builder(*wasm);
+  auto globalIt = functionGlobals.begin();
+  ModuleUtils::iterDefinedFunctions(*wasm, [&](Function* func) {
+    func->body = builder.makeSequence(
+      builder.makeIf(
+        builder.makeUnary(EqZInt32,
+                          builder.makeGlobalGet(*globalIt, Type::i32)),
+        builder.blockify(
```
kripken, line 347: could be `makeSequence` (`blockify` does more things, which you do not need here)

src/tools/wasm-split.cpp
```diff
@@ -245,8 +261,195 @@ void parseInput(Module& wasm, const WasmSplitOptions& options) {
...
+    offset += 4;
+  }
+
+  writeProfile->body = builder.blockify(
+    builder.makeIf(builder.makeBinary(GeUInt32, getSize(), profileSizeConst()),
+                   writeData),
+    profileSizeConst());
+
+  // Create an export for the function
+  auto ex = std::make_unique<Export>();
```
kripken, line 413: please use `builder.makeExport`

src/tools/wasm-split.cpp
```diff
@@ -245,8 +261,195 @@ void parseInput(Module& wasm, const WasmSplitOptions& options) {
...
+  if (wasm.getExportOrNull(options.profileExport) != nullptr) {
+    Fatal() << "error: Export " << options.profileExport << " already exists.";
+  }
+
+  // TODO: calculate module hash.
+  uint64_t moduleHash = 0;
+  PassRunner runner(&wasm, options.passOptions);
+  runner.add(
+    std::make_unique<FuncInstrumentor>(options.profileExport, moduleHash));
+  runner.run();
```
kripken, line 446: I think you can replace these 3 lines with `FuncInstrumentor(..).run(runner, module);`

kripken (approved):
lgtm % comments

src/tools/wasm-split.cpp
tlively: Sounds good, but I'll wait for #3391 to land first ;)

src/tools/wasm-split.cpp
tlively: `Func` is supposed to differentiate it from a hypothetical future `BlockInstrumenter`. But given that this is the only instrumentation in the tool right now, how about doubling down on the generality and changing the name to "Instrumenter"? This is entirely unambiguous in the context of wasm-split.

src/tools/wasm-split.cpp
```diff
@@ -245,8 +261,189 @@ void parseInput(Module& wasm, const WasmSplitOptions& options) {
...
+  this->wasm = wasm;
+  addGlobals();
+  instrumentFuncs();
+  addProfileExport();
+}
+
+void Instrumenter::addGlobals() {
+  // Create fresh global names
+  counterGlobal = Names::getValidGlobalName(*wasm, "monotonic_counter");
+  functionGlobals.reserve(wasm->functions.size());
```
aheejin, line 300: Wouldn't this be the size for both imported and defined functions?

src/tools/wasm-split.cpp
```diff
@@ -245,8 +261,189 @@ void parseInput(Module& wasm, const WasmSplitOptions& options) {
...
+  writeProfile->hasExplicitName = true;
+  writeProfile->setLocalName(0, "addr");
+  writeProfile->setLocalName(1, "size");
+
+  // Calculate the size of the profile:
+  //   8 bytes module hash +
+  //   4 bytes for the timestamp for each function
+  size_t numDefinedFunctions = 0;
+  ModuleUtils::iterDefinedFunctions(*wasm,
+                                    [&](Function*) { numDefinedFunctions++; });
```
aheejin, line 378: Maybe it would be better to have a helper function that returns the number of imported / defined functions...

aheejin (approved):
Nice!

src/tools/wasm-split.cpp
```diff
@@ -245,8 +261,189 @@ void parseInput(Module& wasm, const WasmSplitOptions& options) {
...
+    return builder.makeConst(int32_t(profileSize));
+  };
+
+  // Write the hash followed by all the time stamps
+  Expression* writeData =
+    builder.makeStore(8, 0, 1, getAddr(), hashConst(), Type::i64);
+
+  uint32_t offset = 8;
+  for (const auto& global : functionGlobals) {
+    writeData = builder.blockify(
```
aheejin, line 396:
```suggestion
    writeData = builder.makeSequence(
```
I guess this can be `makeSequence` too?

tlively: No, this one should be a `blockify`. The difference is that `makeSequence` always introduces a new block, but `blockify` appends to the first item if it is already a block.

src/tools/wasm-split.cpp
kripken: sgtm. I didn't realize the intended comparison to blocks.

src/tools/wasm-split.cpp
aheejin: Ah you're right; I thought `writeData` was always a `store`, but this is within a for loop.

src/tools/wasm-split.cpp
tlively: Yes, but it's probably okay to over-allocate a bit here.

src/tools/wasm-split.cpp
tlively: I changed this to just use `functionGlobals.size()`.

PR #3390, opened 2020-11-19
[StackCheck] Handle colliding names

kripken:
This just moves code around and adds some `Names::getValid*Name` calls to
ensure we don't collide with an existing name.

Fixes emscripten-core/emscripten#12834


src/passes/StackCheck.cpp
```diff
@@ -55,23 +48,12 @@ static void importStackOverflowHandler(Module& module, Name name) {
 static void addExportedFunction(Module& module, Function* function) {
   module.addFunction(function);
   auto export_ = new Export;
-  export_->name = export_->value = function->name;
+  export_->name = Names::getValidExportName(module, function->name);
```
tlively, line 51: Will users of the generated code know to look for the new export name if there was a conflict?

kripken: It seems like this can only happen by running the pass more than once, so it's probably someone running it manually anyhow. But, fair point that erroring on the export might be safer.

src/passes/StackCheck.cpp
```diff
@@ -135,13 +117,28 @@ struct EnforceStackLimits : public WalkerPass<PostWalker<EnforceStackLimits>> {
...
 
 struct StackCheck : public Pass {
+  // The base is where the stack begins. As it goes down, that is the highest
+  // valid address.
+  Name STACK_BASE;
+  // The limit is the farthest it can grow to, which is the lowest valid
+  // address.
+  Name STACK_LIMIT;
+  // Exported function to set the base and the limit.
+  Name SET_STACK_LIMITS;
```
sbc100, line 127: These probably shouldn't be in ALL_CAPS anymore if they are no longer fixed.   

I guess `SET_STACK_LIMITS` can stay as a global since we are going to error out on conflict?

sbc100:
I guess we are missing an emscripten test here too right?

sbc100:
Shall we get this landed: https://github.com/emscripten-core/emscripten/issues/12834

sbc100 (approved):

PR #3391, opened 2020-11-20
[wasm-builder] Construct module elements as unique_ptrs

tlively:
When Functions, Globals, Events, and Exports are added to a module, if they are
not already in std::unique_ptrs, they are wrapped in a new std::unique_ptr owned
by the Module. This adds an extra layer of indirection when accessing those
elements that can be avoided by allocating those elements as std::unique_ptrs.
This PR updates wasm-builder to allocate module elements via std::make_unique
rather than `new`. In the future, we should remove the raw pointer versions of
Module::add* to encourage using std::unique_ptrs more broadly.

### DIFF
--- a/src/passes/StackCheck.cpp
+++ b/src/passes/StackCheck.cpp
@@ -52,30 +52,30 @@ static void importStackOverflowHandler(Module& module, Name name) {
   }
 }
 
-static void addExportedFunction(Module& module, Function* function) {
-  module.addFunction(function);
-  auto export_ = new Export;
-  export_->name = export_->value = function->name;
-  export_->kind = ExternalKind::Function;
-  module.addExport(export_);
+static void addExportedFunction(Module& module,
+                                std::unique_ptr<Function> function) {
+  auto export_ =
+    Builder::makeExport(function->name, function->name, ExternalKind::Function);
+  module.addFunction(std::move(function));
+  module.addExport(std::move(export_));
 }
 
 static void generateSetStackLimitFunctions(Module& module) {
   Builder builder(module);
-  Function* limitsFunc = builder.makeFunction(
+  auto limitsFunc = builder.makeFunction(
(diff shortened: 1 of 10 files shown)


kripken (approved):
This is great, but my very very minor concern is that there are at least 2 open PRs that will conflict with this... but we will manage.

src/passes/StackCheck.cpp
```diff
@@ -127,9 +127,9 @@ struct EnforceStackLimits : public WalkerPass<PostWalker<EnforceStackLimits>> {
   }
 
 private:
-  Global* stackPointer;
-  Global* stackBase;
-  Global* stackLimit;
+  const Global* stackPointer;
+  const std::unique_ptr<Global>& stackBase;
+  const std::unique_ptr<Global>& stackLimit;
```
aheejin, line 132: Why are these different from `stackPointer`?

aheejin (approved):
Nice!

src/passes/StackCheck.cpp
tlively: `stackPointer` is initialized via `getStackPointerGlobal` in wasm-emscripten.cpp. I looked into making that function return `const std::unique_ptr<Global>&` rather than `Global*`, but ran into the problem the function returns `nullptr` on failure and there's no clean way to do that using references. With C++17, I could use `std::optional<std::reference_wrapper<Global>>`, but even that's so gross that I would rather just use `Global*` 😅

aheejin: What I was asking is about `stackBase` and `stackLimit`, that why can't they also just be `const Global*`, by passing the raw pointer using `get()`, when we call `EnforceStackLimits` constructor?

aheejin: Gentle ping :) This is a pretty minor question, but just in case you haven't had a chance to see it.

PR #3395, opened 2020-11-23
[s-expression parsing] Parse typed function reference types

kripken:
This does not add testing yet - my next PR, which adds `call_ref`, will make
it much easier to test typed function references, and will include tests that
cover this.


aheejin:
I understand you want to make PRs small and I like small PRs too, but I think this kind of things, like parsing, would be better with tests.. This PR is quite small anyway so I guess adding this to the next PR that has tests wouldn't increase the size of the PR too much..?

kripken:
Sure, I can make it part of the next one then. I agree this is pretty small.

PR #3396, opened 2020-11-23
[TypedFunctionReferences] Implement call_ref

kripken:
Includes minimal support in various passes. Also includes actual optimization
work in Directize, which was easy to add.

Almost has fuzzer support, but the actual `makeCallRef` is just a stub so far.

Includes s-parser support for parsing typed function references types.


src/passes/Directize.cpp
```diff
@@ -92,23 +104,29 @@ struct FunctionDirectizer : public WalkerPass<PostWalker<FunctionDirectizer>> {
 
 struct Directize : public Pass {
   void run(PassRunner* runner, Module* module) override {
+    TableUtils::FlatTable flatTable(module->table);
+    TableUtils::FlatTable* usedFlatTable = &flatTable;
```
tlively, line 108: I think it would be clearer to make this a separate `bool` variable like `canOptimizeCallIndirects`.

src/passes/MergeBlocks.cpp
```diff
@@ -564,7 +564,7 @@ struct MergeBlocks : public WalkerPass<PostWalker<MergeBlocks>> {
 
   void visitCall(Call* curr) { handleCall(curr); }
 
-  void visitCallIndirect(CallIndirect* curr) {
+  template<typename T> void handleNonDirectCall(T* curr) {
```
tlively, line 567: This seems like a code smell that suggests `CallIndirect` and `CallRef` should be rolled into a single expression class (like how `call` and `return_call` are combined), but I'm not sure that would be better overall. Did you consider that approach?

src/wasm-interpreter.h
```diff
@@ -1375,6 +1375,10 @@ class ExpressionRunner : public OverriddenVisitor<SubType, Flow> {
     NOTE_EVAL1(value);
     return Literal(value.geti31(curr->signed_));
   }
+  Flow visitCallRef(CallRef* curr) {
+    NOTE_ENTER("CallRef");
+    WASM_UNREACHABLE("TODO (gc): ref.call");
+  }
```
tlively, line 1381: It looks like this should be up with the other nonconstant expressions with body `{ WASM_UNREACHABLE("unimp"); }`.

src/wasm-s-parser.h
```diff
@@ -77,6 +77,11 @@ class Element {
   Element* setString(cashew::IString str__, bool dollared__, bool quoted__);
   Element* setMetadata(size_t line_, size_t col_, SourceLocation* startLoc_);
 
+  // comparisons
+  bool operator==(Name name) { return isStr() && str() == name; }
+
+  template<typename T> bool operator!=(T t) { return !(*this == t); }
```
tlively, line 83: Is there an `operator==` besides the one you added here? If not, it would be simpler to not make this generic.

src/wasm/wasm-validator.cpp
```diff
@@ -406,6 +407,44 @@ struct FunctionValidator : public WalkerPass<PostWalker<FunctionValidator>> {
...
+                    curr,
+                    "return_call* should have unreachable type");
+      shouldBeEqual(
+        getFunction()->sig.results,
+        sig.results,
+        curr,
+        "return_call* callee return type must match caller return type");
+    } else {
+      if (curr->type != Type::unreachable) {
+        shouldBeEqual(curr->type,
```
tlively, line 440: This looks like an opportunity to use `shouldBeEqualOrFirstIsUnreachable`.

test/typed-function-references.wast
```diff
@@ -0,0 +1,21 @@
+(module
```
tlively, line 1: It would be good to add `return_call_ref` variants to this testing as well.

src/wasm/wasm-validator.cpp
```diff
@@ -2199,6 +2152,19 @@ void FunctionValidator::visitTupleExtract(TupleExtract* curr) {
   }
 }
 
+void FunctionValidator::visitCallRef(CallRef* curr) {
```
tlively, line 2155: It would be good to check that the tail-call feature is enabled if this is a `return_call_ref`, like `visitCall` and `visitCallIndirect` do. It may be worth abstracting that out, too.

src/wasm/wasm-validator.cpp
kripken: Sounds good, improving.

src/wasm/wasm-validator.cpp
kripken: Good idea, refactoring.

test/typed-function-references.wast
kripken: Good idea, adding.

src/wasm-s-parser.h
aheejin: Where do we need to use these operators? I thought `Element` as a more higher container-like entity than `Name`, so it feels a bit weird to me. If we don't use it in too many places then calling `str()` to compare names wouldn't be too bad..

src/wasm/wasm-binary.cpp
```diff
@@ -5426,6 +5436,27 @@ void WasmBinaryBuilder::visitBrOnExn(BrOnExn* curr) {
...
+  BYN_TRACE("zz node: CallRef\n");
+  curr->target = popNonVoidExpression();
+  auto type = curr->target->type;
+  if (!type.isRef()) {
+    throwError("Non-ref type for a call_ref");
+  }
+  auto heapType = type.getHeapType();
+  if (!heapType.isSignature()) {
+    std::cerr << type << '\n';
+    throwError("Invalid reference type for a call_ref");
```
aheejin, line 5449: How about printing these in a single line, like, `Invalid reference type for a call_ref: type`?

src/wasm/wasm-s-parser.cpp
```diff
@@ -925,10 +925,34 @@ Type SExpressionWasmBuilder::elementToType(Element& s) {
...
+  auto size = list.size();
+  if (size > 0 && elementStartsWith(s, REF)) {
+    // It's a reference. It should be in the form
+    //   (ref $name)
+    // or
+    //   (ref null $name)
+    if ((size != 2 && size != 3) || !list[1]->isStr()) {
+      throw ParseException(
+        std::string("invalid reference type"), s.line, s.col);
+    }
```
aheejin, line 938: While we are checking, how about checking both ways, and all at the top, like
```suggestion
    if (size != 2 && size != 3) {
      throw ParseException(
        std::string("invalid reference type"), s.line, s.col);
    }
    if (size == 2 && !list[1]->isStr()) {
      throw ParseException(
        std::string("invalid reference type"), s.line, s.col);
    }
    if (size == 3 && (*list[1] != NULL_ || !list[2]->isStr())) {
      throw ParseException(
        std::string("invalid reference type"), s.line, s.col);
}
```

src/passes/Directize.cpp
tlively: I found this confusing because it is largely redundant with the `flatTable` variable. It took me a bit to figure out that the only extra information `usedFlatTable` was encoding was whether the `flatTable` should be used. I believe that intent could be more simply (and therefore more clearly) represented as a bool. This seems different from other situations in which we use pointers to indicate optionality because in most of those situations, the optional pointee is either available or not present at all.

src/wasm-s-parser.h
tlively: I think avoiding templates would have the additional benefit of improving error messages, but I don't feel too strongly about this either.

tlively (approved):
Beyond those two remaining comments, which I will leave up to you, and the current test failure, this LGTM!

aheejin (approved):

src/passes/Directize.cpp
kripken: Oh, I think we were talking about different things. Do you mean in this function itself, `run`? I agree it would make sense to add a bool there, I'll do that.

src/wasm-s-parser.h
kripken: It's not many places. But it is pretty nice to do `if (x == REF)` instead of `if (x.isStr() && x.str() == REF)` I think?

src/wasm/wasm-binary.cpp
kripken: I don't see an easy way to do that. We should really have an easy way to convert a type into a string...

I agree it's a little odd as it is. Let me just remove the type for now, to match other errors.

kripken:
Last commits should address the review comments.

src/wasm/wasm-binary.cpp
aheejin: It seems we have [`Type::toString`](https://github.com/WebAssembly/binaryen/blob/b2d797f1f9f1192b8f4d2664f76a8d0b9278a0ef/src/wasm/wasm-type.cpp#L117)?

kripken:
Fixing the error requires a non-trivial amount of work it turns out - fixing s-expression parsing and emitting of function ref type names. I pushed a fix for emitting them, will look at parsing next...

src/wasm-s-parser.h
aheejin: Hmm, yeah, that doesn't look bad... I'm OK either way 😅

src/wasm/wasm-binary.cpp
kripken: Thanks! I missed that, will add...

kripken:
Last commit refactors the s-expression parsing code to allow us to parse an "inline" function signature, that is, it now supports not just `(ref $typename)` but also `(ref (func (result i32)))`.

PR #3397, opened 2020-11-24
Let EnforceStackLimits have pointers to Globals (NFC)

aheejin:
These can be simply raw pointers, given that they are stored in modules
using `unique_ptr`s.

Context:
https://github.com/WebAssembly/binaryen/pull/3391#discussion_r527495451


tlively (approved):
This does look cleaner, thanks. Sorry about missing this question on the original PR, too. I'll look into why my email filters were burying the notification emails for these post-merge comments.

PR #3398, opened 2020-11-24
Fix comment in invalid-options.wast

aheejin:
Context:
https://github.com/WebAssembly/binaryen/pull/3359#discussion_r527484301


tlively (approved):
Oops! Thanks for the fix :)

PR #3399, opened 2020-11-24
[OptimizeInstructions] Fix a fuzz bug with comparing signed and unsigned values

kripken:
The old code there was buggy, it turns out. I rewrote it to handle only the cases
that are obviously correct, which removes the bug. I believe this does not lose
any valid optimization cases, but frankly the old code was hard to read - best
to just read the new code there, I think (very unlikely this optimization is that
important anyhow).

### DIFF
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -515,10 +515,10 @@ struct OptimizeInstructions
               Bits::popCount(right >> uint32_t(bits - 1));
             // If all the relevant bits on C are zero
             // then we can mask off the high bits instead of sign-extending x.
-            // This is valid because if x is signed, then the comparison was
-            // false before (signed vs unsigned value), and will still be false
+            // This is valid because if x is negative, then the comparison was
+            // false before (negative vs positive), and will still be false
             // as the sign bit will remain to cause a difference. And if x is
-            // unsigned then the upper bits would be zero anyhow.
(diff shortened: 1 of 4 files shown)


kripken:
Fuzzer found a bug in the new code, fixed.

kripken:
Fuzzer found another bug. Rewritten to be much simpler and hopefully correct now.

(I wish the fuzzer were this good back when I wrote this code originally years ago...)

src/passes/OptimizeInstructions.cpp
```diff
@@ -496,47 +496,50 @@ struct OptimizeInstructions
...
+            uint32_t right = c->value.geti32();
+            uint32_t numRelevantBits = 32 - bits + 1;
+            uint32_t setRelevantBits =
+              Bits::popCount(right >> uint32_t(bits - 1));
+            // If all the relevant bits on C are zero
+            // then we can mask off the high bits instead of sign-extending x.
+            // This is valid because if x is signed, then the comparison was
+            // false before (signed vs unsigned value), and will still be false
+            // as the sign bit will remain to cause a difference. And if x is
+            // unsigned then the upper bits would be zero anyhow.
```
tlively, line 521: I think this would be clearer if you replaced "signed" with "negative" and "unsigned" with "positive."

tlively (approved):
As you suggested, I did not read the old code, but the new code looks good to me modulo a couple comments on the comments.

PR #3400, opened 2020-11-24
[wasm-split] Read and use profiles

tlively:
Read the profiles produced by wasm-split's instrumentation to guide splitting.
In this initial implementation, all functions that the profile shows to have
been called are kept in the initial module. In the future, users may be able to
tune this so that functions that are run later will still be split out.

cc @sbc100 @awtcode

### DIFF
--- a/src/tools/wasm-split.cpp
+++ b/src/tools/wasm-split.cpp
@@ -359,6 +359,21 @@ void Instrumenter::instrumentFuncs() {
   });
 }
 
+// wasm-split profile format::
+//
+// The wasm-split profile is a binary format designed to be simple to produce
+// and consume. It is comprised of:
+//
+//   1. An 8-byte module hash
+//
+//   2. A 4-byte timestamp for each defined function
+//
+// The module hash is meant to guard against bugs where the module that was
+// instrumented and the module that is being split are different. The timestamps
+// are non-zero for functions that were called during the instrumented run and 0
+// otherwise. Functions with smaller non-zero timestamps were called earlier in
+// the instrumented run than funtions with larger timestamps.
+
 void Instrumenter::addProfileExport() {
   // Create and export a function to dump the profile into a given memory
   // buffer. The function takes the available address and buffer size as
@@ -446,6 +461,7 @@ void instrumentModule(Module& wasm, const WasmSplitOptions& options) {
   writer.write(wasm, options.output);
 }
 
+// See "wasm-split profile format" above for more information.
 std::set<Name> readProfile(Module& wasm, const WasmSplitOptions& options) {
   auto profileData =
     read_file<std::vector<char>>(options.profileFile, Flags::Binary);
--- a/src/tools/wasm-split.cpp
+++ b/src/tools/wasm-split.cpp
@@ -20,6 +20,7 @@
 #include "ir/module-splitting.h"
 #include "ir/module-utils.h"
 #include "ir/names.h"
+#include "support/file.h"
 #include "support/name.h"
 #include "support/utilities.h"
 #include "tool-options.h"
@@ -423,6 +424,8 @@ void Instrumenter::addProfileExport() {
       wasm->memory.max = pages;
     }
   }
+
+  // TODO: export the memory if it is not already exported.
 }
 
 void instrumentModule(Module& wasm, const WasmSplitOptions& options) {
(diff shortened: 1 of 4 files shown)


src/tools/wasm-split.cpp
```diff
@@ -443,12 +446,50 @@ void instrumentModule(Module& wasm, const WasmSplitOptions& options) {
   writer.write(wasm, options.output);
 }
 
+std::set<Name> readProfile(Module& wasm, const WasmSplitOptions& options) {
```
kripken, line 449: where is the profile file documented? I'm unsure what to compare the code here against.

test/lit/wasm-split/call_exports.mjs
```diff
@@ -0,0 +1,25 @@
+// Instantiates an instrumented module, calls the given exports, then collects
+// its wasm-split profile and writes it to a given file.
+//
+// Usage:
+//
+//  node call_exports.mjs <module> <profile> <export>*
```
kripken, line 6: what does this lit test do? I don't see any checks for comparing anything... maybe I just don't know enough about lit.

test/lit/wasm-split/profile-guided.wast
```diff
@@ -0,0 +1,67 @@
+;; Instrument the binary
+
+;; RUN: wasm-split --instrument %s -o %t.instrumented.wasm
```
kripken, line 3: what does such a line do without a check later? just verify it doesn't crash?

test/lit/wasm-split/profile-guided.wast
```diff
@@ -0,0 +1,67 @@
...
+;; RUN: node %S/call_exports.mjs %t.instrumented.wasm %t.bar.prof bar
+;; RUN: node %S/call_exports.mjs %t.instrumented.wasm %t.both.prof foo bar
+;; RUN: node %S/call_exports.mjs %t.instrumented.wasm %t.none.prof
+
+;; Create profile-guided splits
+
+;; RUN: wasm-split %s --profile=%t.foo.prof -v -o1 %t.foo.1.wasm -o2 %t.foo.2.wasm \
+;; RUN:   | filecheck %s --check-prefix FOO
+
+;; FOO: Keeping functions: deep_foo_callee, foo, foo_callee, shared_callee
```
kripken, line 17: what is `FOO` here and above? I'm not sure how to read this...

test/lit/wasm-split/call_exports.mjs
tlively: This is not a lit test, this is a script invoked by profile-guided.wast (which _is_ a lit test).

src/tools/wasm-split.cpp
tlively: For now the only thing to compare against is the code that writes the profile, which is also in this file.

test/lit/wasm-split/profile-guided.wast
tlively: That's correct, but it also sets up state that is used in the subsequent `RUN` lines. Specifically, it writes the file %t.instrumented.wasm. The "%t" is replaced with a unique filename at runtime.

test/lit/wasm-split/profile-guided.wast
tlively: `FOO` takes the place of `CHECK` (via the --check-prefix option above). This option allows multiple different test outputs to be checked in the same file with each using a different check prefix.

aheejin: Reference: https://llvm.org/docs/CommandGuide/FileCheck.html#the-filecheck-check-prefix-option

src/tools/wasm-split.cpp
aheejin: Maybe it'd be good to add comments on the file format, however simple it is.

aheejin (approved):

test/lit/wasm-split/profile-guided.wast
kripken: Thanks! I see, so the relevant docs are for FileCheck and not lit... TIL

I understand the benefit of the prefix in the example there, but not here. It seems like FOO, BAR, are each used once and used immediately. Could they all have been CHECK?

tlively: If the had all been CHECK, then every invocation of filecheck would have checked all of them against the input, which wouldn't have worked. Remember that filecheck matches against the entire file, not just the lines after or between RUN lines. filecheck doesn't know anything about RUN lines at all.

kripken: Thanks, I had assumed it did understand about RUN lines from the examples, but now I get it..

PR #3401, opened 2020-11-25
[TypedFunctionReferences] Enable call_ref in fuzzer, and fix minor misc fuzz bugs

kripken:
* Count signatures in tuple locals.
* Count nested signature types (confirming @aheejin was right, that was missing).
* Inlining was using the wrong type.
* OptimizeInstructions should return -1 for unhandled types, not error.
* The fuzzer should check for ref types as well, not just typed function references,
  similar to what GC does.
* The fuzzer now creates a function if it has no other option for creating a constant
  expression of a function type, then does a ref.func of that.
* Handle unreachability in call_ref binary reading.
* S-expression parsing fixes in more places, and add a tiny fuzzer for it.
* Switch fuzzer test to just have the metrics, and not print all the fuzz output which
  changes a lot. Also fix `noprint` handling which only  worked on binaries before.
* Fix `Properties::getLiteral()` to use the specific function type properly, and make
  Literal's function constructor require that, to prevent future bugs.
* Turn all input types into nullable types, for now.


src/wasm/wasm-type.cpp
```diff
@@ -98,8 +98,8 @@ struct TypeInfo {
       case TupleKind:
         return tuple == other.tuple;
       case RefKind:
-        return ref.heapType == other.ref.heapType &&
-               ref.nullable == other.ref.nullable;
+        // TODO: compare nullability as well, once we properly support that
+        return ref.heapType == other.ref.heapType;
```
tlively, line 102: What made this change necessary? It seems dangerous to me. For one, type equality no longer implies hash equality, and I wouldn't be surprised if that subtly breaks something.

test/passes/inline_return_call_ref.bin.txt
```diff
@@ -0,0 +1,15 @@
+(module
```
tlively, line 1: Why is there no .wast file for this test?

test/passes/optimize-instructions_all-features.txt
```diff
@@ -5698,3 +5698,10 @@
   )
  )
 )
+(module
+ (type $i32_=>_i32 (func (param i32) (result i32)))
+ (type $ref_null<i32_->_i32>_=>_i32 (func (param (ref null (func (param i32) (result i32)))) (result i32)))
```
tlively, line 5703: Hmm, should this be using `$i32_=>_i32`?

test/passes/translate-to-fuzz_all-features_metrics_noprint.txt
```diff
@@ -0,0 +1,42 @@
+total
```
tlively, line 1: Hmm, even these numbers are still going to change frequently so diffs will still not be very useful. How about we make this a lit/filecheck test that can just test that all of the expected element types are present without checking their numbers? Even that seems concerning because it looks like there are a large number of expression types missing below.

test/typed-function-references.wast.fromBinary.noDebugInfo
```diff
@@ -1,9 +1,15 @@
...
  (type $i32_=>_i32 (func (param i32) (result i32)))
+ (type $none_=>_none (func))
+ (type $none_=>_eqref (func (result eqref)))
+ (type $i32_i32_i32_i32_i32_=>_none (func (param i32 i32 i32 i32 i32)))
  (type $none_=>_i32 (func (result i32)))
+ (type $none_=>_anyref (func (result anyref)))
  (type $ref_null<i32_->_i32>_=>_i32 (func (param (ref null (func (param i32) (result i32)))) (result i32)))
  (type $ref<i32_->_i32>_=>_i32 (func (param (ref (func (param i32) (result i32)))) (result i32)))
+ (type $f64_=>_ref_null<_->_eqref> (func (param f64) (result (ref null (func (result eqref))))))
+ (type $none_=>_ref<i32i32i32i32i32_->_> (func (result (ref (func (param i32 i32 i32 i32 i32))))))
```
tlively, line 11: It looks like this type name generation could be improved (but not necessarily in this PR).

src/passes/OptimizeInstructions.cpp
```diff
@@ -121,7 +121,9 @@ struct LocalScanner : PostWalker<LocalScanner> {
   Index getMaxBitsForLocal(LocalGet* get) { return getBitsForType(get->type); }
 
   Index getBitsForType(Type type) {
-    TODO_SINGLE_COMPOUND(type);
+    if (!type.isBasic()) {
+      return -1;
```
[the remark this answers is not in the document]
kripken: It's safer this way, as "0" means no bits are used, which might lead an optimization to think that it can do certain things to it. But it would be a bug either way that we should fix, so it doesn't matter much - this just keeps using "-1" as the existing code lower down does.

src/tools/fuzzing.h
```diff
@@ -1449,7 +1452,31 @@ class TranslateToFuzzReader {
...
+      if (i == TRIES) {
+        return makeTrivial(type);
+      }
+    }
+    std::vector<Expression*> args;
+    for (const auto& type : target->sig.params) {
+      args.push_back(make(type));
+    }
+    auto targetType = Type(HeapType(target->sig), /* nullable = */ true);
+    // TODO: half the time make a completely random item with that type.
```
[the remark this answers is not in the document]
kripken: This would take some interesting extra work in the fuzzer, I think, as `make(type)` doesn't handle typed function references yet. I'd rather defer it.

src/wasm/wasm-s-parser.cpp
```diff
@@ -1911,9 +1907,25 @@ Expression* SExpressionWasmBuilder::makeRefNull(Element& s) {
...
   }
-  auto heapType = stringToHeapType(s[1]->str());
   auto ret = allocator.alloc<RefNull>();
-  ret->finalize(heapType);
+  if (s[1]->isStr()) {
+    ret->finalize(stringToHeapType(s[1]->str()));
+  } else {
+    // To parse a heap type, create an element around it, and call that method.
+    // That is, given (func) we wrap to (ref (func)).
+    // TODO add a helper method, but this is the only user atm.
```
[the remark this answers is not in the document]
kripken: `ref.null is generalised to take a <heaptype> immediate` in https://github.com/WebAssembly/function-references/blob/master/proposals/function-references/Overview.md implies that, I believe?

It makes sense since we don't need the full ref type with nullability - just the heap type.

Adding a comment.

src/wasm/wasm-type.cpp
kripken: without this, the fuzzer runs into issues, because we emit nullable types in places that are not technically valid. but we do still emit a non-nullable type for a `ref.func` for example. and we can end up comparing those.

for now, I think we should just ignore the issue of nullability, which is what this moves us towards.

test/passes/inline_return_call_ref.bin.txt
kripken: it's a wasm binary file atm - but yeah, it could be a wat, let me fix that...

test/passes/optimize-instructions_all-features.txt
kripken: I think both are valid, and I'm not sure which is best actually. The benefit of printing the full structure of the type is that we can print a type anywhere in the codebase - that is, without collecting all the types and giving them names etc.

But, with some work maybe we could special-case the type section declarations to be more concise?

test/passes/translate-to-fuzz_all-features_metrics_noprint.txt
kripken: Yeah, I don't think that would quite work - this is random data, so we can't guarantee that all instructions actually appear.

I do think it's still useful to have the metrics. They would show a big unexpected change in a clear way, at least.

kripken:
Pushed some commits following the review feedback.

test/typed-function-references.wast.fromBinary.noDebugInfo
kripken: I think I found a reasonable way to improve this, pushed now. It emits things like `$none_=>_ref[i32_i32_->_]` (proper separation between elements, and using `[ ]` it is less confusing with the `>` in the `->`).

src/wasm/wasm-s-parser.cpp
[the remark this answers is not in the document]
tlively: But heaptype is defined as `heaptype ::= <typeidx> | func | extern`. And that's the abstract syntax, not the concrete text syntax. I took a look at wasp and wabt but couldn't quickly figure out what those tools expected.

src/wasm/wasm-type.cpp
tlively: I'm still worried that ignoring issues of nullability by hacking around it whenever it comes up is going to leave us in a real mess where we are totally inconsistent in how we treat nullability. It would be better to ignore non-nullability at its source by either having the parsers translate all non-nullable types to nullable types or making the validator reject all non-nullable types. If we ensure no non-nullable types get into our system, we are free to write code that simply asserts that they are not present. Then, later when we want to support them, we only have to make changes to a few key places and all the asserts will tell us exactly what needs fixing rather than having silently broken code (albeit marked with TODOs) all over the place.

src/wasm/wasm-s-parser.cpp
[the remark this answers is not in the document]
tlively: I filed https://github.com/WebAssembly/function-references/issues/42 to clarify this.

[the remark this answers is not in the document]
tlively: But I'm fine landing this more powerful syntax for now, so there's nothing here to stop the PR from landing.

test/passes/optimize-instructions_all-features.txt
tlively: This seems fine. We can change it if we have to depending on how https://github.com/WebAssembly/function-references/issues/42 is resolved.

test/passes/translate-to-fuzz_all-features_metrics_noprint.txt
tlively: I think the value of this test is to demonstrate that all of the interesting things we want the fuzzer to generate are in fact generated. I would want the onus to be on whoever changes the fuzzer to find a permutation of the input such that everything is generated at least once. Without that, how can we differentiate random changes in the test output from real bugs?

src/wasm/wasm-type.cpp
kripken: That's true, but the total work of now + later would still be the same, I think.

But it sounds like you feel strongly about this? If so I can do the work now.

Note though that the work would not just be changing the parsers, but all other places that generate a type, like a `ref.func` execution in the interpreter, etc.

test/passes/translate-to-fuzz_all-features_metrics_noprint.txt
kripken: I agree the coverage is not complete now. But finding a random input that generates all the things is only slightly less hard than solving NP-completeness. Well not really, but the method might be "keep adding random stuff until everything is emitted", which may well lead to a very large input.

If you dislike this, I can just remove the `--metrics` part, so it prints nothing. Then it would just test that it doesn't crash. And we'd need to be careful when landing any change to the fuzzer (basically, we'd want to do some manual testing at those times). I think the tradeoff is better the way it is now, but I don't feel strongly, and I see that this is a test that doesn't cleanly cover all the things we'd want, which is I guess what annoys you, and I can sort of see that too.

src/wasm/literal.cpp
```diff
@@ -34,7 +34,8 @@ Literal::Literal(Type type) : type(type) {
     // i31ref is special in that it is non-nullable, so we construct with zero
     i32 = 0;
   } else {
-    assert(type != Type::unreachable && (!type.isRef() || type.isNullable()));
+    assert(type != Type::unreachable);
+    // TODO: assert on non-nullability when we support that
```
aheejin, line 38: Why should the type non-nullable here? Similar question with my another comment, I think

src/wasm-interpreter.h
```diff
@@ -2110,6 +2110,10 @@ template<typename GlobalManager, typename SubType> class ModuleInstanceBase {
       if (target.breaking()) {
         return target;
       }
+      if (target.getSingleValue().isNull()) {
+        // TODO: remove this when we fully support non-nullable types
```
aheejin, line 2114: According to the [spec](https://github.com/WebAssembly/function-references/blob/master/proposals/function-references/Overview.md#functions), `call_ref` takes a nullable reference and traps on the null value, so I think this can be the final state and we don't need this TODO?

src/tools/fuzzing.h
```diff
@@ -1449,7 +1452,31 @@ class TranslateToFuzzReader {
...
 
   Expression* makeCallRef(Type type) {
-    return makeTrivial(type); // FIXME
+    // look for a call target with the right type
+    Function* target;
+    bool isReturn;
+    size_t i = 0;
+    while (1) {
+      // TODO: handle unreachable
+      target = wasm.functions[upTo(wasm.functions.size())].get();
```
aheejin, line 1461: Can `wasm.functions` be empty?

src/tools/fuzzing.h
```diff
@@ -1449,7 +1452,31 @@ class TranslateToFuzzReader {
   }
 
   Expression* makeCallRef(Type type) {
-    return makeTrivial(type); // FIXME
+    // look for a call target with the right type
+    Function* target;
```
aheejin, line 1456: Should we initialize this with the current function or `nullptr`, in case `wasm.functions` is empty?

test/passes/translate-to-fuzz_all-features_metrics_noprint.txt
tlively: No, I think this is a big improvement; thank you :) Let's keep this as-is for now.

scripts/test/wasm_opt.py
```diff
@@ -171,7 +172,8 @@ def update_wasm_opt_tests():
         passes_file = os.path.join(shared.get_test_dir('passes'), passname + '.passes')
         if os.path.exists(passes_file):
             passname = open(passes_file).read().strip()
-        opts = [('--' + p if not p.startswith('O') and p != 'g' else '-' + p) for p in passname.split('_')]
+        passes = [p for p in passname.split('_') if p != 'noprint']
+        opts = [('--' + p if not p.startswith('O') and p != 'g' else '-' + p) for p in passes]
```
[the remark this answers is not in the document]
kripken: I think what happens there is that the list of passes is retrieved from a separate `.passes` file. The problem is that when not using a `.passes` file we get the passes from the filename, and so then we need to realize that `noprint` is not a pass.

src/wasm/wasm-type.cpp
tlively: I would appreciate that, thank you! I think it will be much easier to reason about that code with that change.

aheejin: Can we pretend `ref.func` generates a nullable type, and assume/assert that all types are nullable elsewhere? That way, we need to only fix `ref.func` later at the source. By the way, I don't think the interpreter needs to change? (I commented there too) The spec's `ref.func` takes a nullable reference and traps on null anyway.

src/ir/module-utils.h
```diff
@@ -445,13 +445,21 @@ collectSignatures(Module& wasm,
...
   // Collect all the counts.
   Counts counts;
+  auto maybeAdd = [&](Type type) {
+    if (type.isRef()) {
+      auto heapType = type.getHeapType();
+      if (heapType.isSignature()) {
+        counts[heapType.getSignature()]++;
+      }
+    }
+  };
```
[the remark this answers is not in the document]
kripken: Good idea to share that method, will refactor.

I don't think that method needs to handle tuple types, though? That's handled explicitly in visitExpression which is the one place where it is relevant, I believe (and we need to check if the expression is a control flow structure first there).

src/literal.h
```diff
@@ -231,7 +230,7 @@ class Literal {
     }
   }
   static Literal makeNull(Type type) {
-    assert(type.isNullable());
+    // TODO: assert on non-nullability when we support that
```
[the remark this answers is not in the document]
kripken: Yes, the type here should be nullable, sorry - the comment is reversed...

src/wasm/wasm.cpp
```diff
@@ -973,12 +973,10 @@ void MemoryGrow::finalize() {
 void RefNull::finalize(HeapType heapType) { type = Type(heapType, true); }
 
 void RefNull::finalize(Type type_) {
-  assert(type_ == Type::unreachable || type_.isNullable());
```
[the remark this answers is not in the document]
kripken: We shouldn't be asserting on stuff like this, we should be checking it in the validator. (We don't atm because nullability is a wip, though - removing this line matches the behavior there.)

src/wasm-interpreter.h
kripken: Good point, I thought it takes a non-null reference... fixing.

src/tools/fuzzing.h
kripken: Good catch, yes!

src/tools/fuzzing.h
kripken: I don't think we need to. We only exit the loop (via the break) and actually use that variable if we set it to a proper value before exiting the loop.

kripken:
Ok, all review comments are hopefully now addressed.

src/ir/module-utils.h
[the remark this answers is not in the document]
aheejin: If block types contain tuples, shouldn't we check each of the tuple element for heap signature too?

src/ir/module-utils.h
```diff
@@ -417,23 +427,14 @@ collectSignatures(Module& wasm,
...
-          }
-        }
-      }
-
-      void maybeNote(Type type) {
-        if (type.isRef()) {
-          auto heapType = type.getHeapType();
-          if (heapType.isSignature()) {
-            counts[heapType.getSignature()]++;
+            counts.note(Signature(Type::none, curr->type));
```
aheejin, line 437: Shouldn't we do this here too?
```cpp
if (type.isTuple()) {
  for (auto t : type) {
    counts.maybeNote(t);
  }
}
```
This is what I meant by doing tuple handling in `maybeNote` too... but I might be mistaken or something.

test/passes/optimize-instructions_all-features.wast
```diff
@@ -6162,3 +6162,10 @@
     )
   )
 )
+;; typed function references
+(module
+  (type $i32-i32 (func (param i32) (result i32)))
+  (func $call_from-param (param $f (ref null $i32-i32)) (result i32)
+    (unreachable)
+  )
+)
```
aheejin, line 6171: This is not supposed to be optimized, right? It would be good to comment on what the expected output is and why it shouldn't be changed.

test/passes/simplify-globals_all-features_fuzz-exec.wast
```diff
@@ -0,0 +1,14 @@
...
+ )
+ (func "export" (result funcref)
+  ;; this set's value will be applied to the get right after it. we should carry
+  ;; over the specific typed function reference type properly while doing so.
+  (global.set $global$0
+   (ref.func $0)
+  )
+  (global.get $global$0)
+ )
+)
```
aheejin, line 14: What does this test for TFRs? This doesn't seem to have TFRs so...

src/wasm/wasm-type.cpp
tlively: > The spec's ref.func takes a nullable reference and traps on null anyway

`ref.func` => `call_ref`, I think.

src/ir/module-utils.h
```diff
@@ -402,7 +402,17 @@ inline void
...
                   std::vector<Signature>& signatures,
                   std::unordered_map<Signature, Index>& sigIndices) {
-  using Counts = std::unordered_map<Signature, size_t>;
+  struct Counts : public std::unordered_map<Signature, size_t> {
+    void note(Signature sig) { (*this)[sig]++; }
+    void maybeNote(Type type) {
+      if (type.isRef()) {
+        auto heapType = type.getHeapType();
+        if (heapType.isSignature()) {
+          note(heapType.getSignature());
```
tlively, line 411: IIUC, this will miss some more interesting cases. For example, in this module:

```webassembly
(module
  (type $g (func))
  (type $f (func (param (ref $g))))
  (func $foo (param (ref $f) (ref $f)))
)
```

This code will properly note `$f` twice, but it will not note `$g`. I guess in general this will require a type walk. I'll throw up a PR adding a type walking function I have written locally that you can use.

src/wasm/wasm-s-parser.cpp
```diff
@@ -1911,9 +1908,30 @@ Expression* SExpressionWasmBuilder::makeRefNull(Element& s) {
...
+    list.resize(3);
+    Element ref(wasm.allocator);
+    ref.setString(REF, false, false);
+    Element null(wasm.allocator);
+    null.setString(NULL_, false, false);
+    list[0] = &ref;
+    list[1] = &null;
+    list[2] = s[1];
+    ret->finalize(elementToType(wrapper));
+  }
```
aheejin, line 1934: I understand it is not clear whether we allow inline TFRs for `ref.null`, but even in case we do, do we write in as `(func (param i32))` or `(ref (func (param i32))`? If it's the latter, I don't think we need this?

src/ir/module-utils.h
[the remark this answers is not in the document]
kripken: I think that is what the code on line 470+ does. We would have the block type there, and then recursively traverse its elements, and then we'd get the signature types in there.

src/ir/module-utils.h
aheejin: I think it will note `$g`? Doesn't the recursive traversing from line 470 do that?

src/ir/module-utils.h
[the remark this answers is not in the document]
aheejin: Ah, I think you're right. Then I guess we don't need the separate tuple handling for `vars`(locals)?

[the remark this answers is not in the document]
kripken: I added a testcase for this now. Does that match the problem you were thinking about?

src/ir/module-utils.h
kripken: I think it works too. The roundtripped output contains this, which matches `$g`'s structure:
```wat
(module
 (type $none_=>_none (func))
```

src/ir/module-utils.h
kripken: I think on line 470 we should handle it? The count we add here adds a signature whose `results` is what your loop would traverse, and we traverse those values there.

test/passes/optimize-instructions_all-features.wast
kripken: Sure, adding a comment.

test/passes/simplify-globals_all-features_fuzz-exec.wast
kripken: Not in the source code, correct. But we make a TFR type for the `ref.func`. The comment tries to explain the bug that this prevents from regressing again:
```
  ;; this set's value will be applied to the get right after it. we should carry
  ;; over the specific typed function reference type properly while doing so.
```

src/wasm/wasm-s-parser.cpp
kripken: It's the former. IIUC, the former is a "heap type". The latter is a "reference type". A reference type also has nullability.

I could refactor that into a method to parse heap types specifically. But maybe let's see what the spec discussion returns, as maybe this is all wrong anyhow...

src/ir/module-utils.h
tlively: Oh right, I forgot about that code 👍

src/ir/module-utils.h
tlively: oh right, I forgot about that code 👍 
`

src/ir/module-utils.h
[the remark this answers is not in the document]
kripken: I think we do need the tuple handling there. The code lower down does recursive analysis of signatures - not types.

But thinking about it now, I am not sure why this is different from the other tuple handling code in `visitExpression`. I think they both achieve the same result - getting a signature from a type, which is then analyzed later - but one also creates a signature (with no inputs, and with those outputs) that is counted. I *think* that is because in the binary format, such a thing appears for control flow structures as `visitExpression` handles, but for a tuple in a local, we actually lower those down anyhow, they are not in the wasm binary format, so we don't need to count them (just to note their children). So I think it's correct, but I'm not as familiar with the tuple stuff as @tlively ?

test/passes/optimize-instructions_all-features.wast
aheejin: Sorry, this is maybe because I'm not very famliar with `OptimizeInstructions`, but I think it's worth a little more explanation because that pass is doing so many things anyway. If this has not been a ref type, what kind of optimization would have happened? I'm still not very sure what this tries to test....

src/ir/module-utils.h
```diff
@@ -466,8 +467,36 @@ collectSignatures(Module& wasm,
...
+  // we just need to visit it once.
+  // TODO: handle struct and array fields
+  std::unordered_set<Signature> newSigs;
+  for (auto& pair : counts) {
+    newSigs.insert(pair.first);
+  }
+  while (!newSigs.empty()) {
+    auto iter = newSigs.begin();
+    auto sig = *iter;
+    newSigs.erase(iter);
```
tlively, line 484: It would be slightly more efficient to pop from the back here.

src/ir/module-utils.h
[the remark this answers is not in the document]
aheejin: Yes I think so, thanks!

src/ir/module-utils.h
aheejin: Is it possible? It is an `unordered_set`..

src/ir/module-utils.h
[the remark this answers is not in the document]
tlively: Yes, that all sounds correct.

src/ir/module-utils.h
```diff
@@ -448,10 +449,10 @@ collectSignatures(Module& wasm,
...
-      if (type.isRef()) {
-        auto heapType = type.getHeapType();
-        if (heapType.isSignature()) {
-          counts[heapType.getSignature()]++;
+      counts.maybeNote(type);
+      if (type.isTuple()) {
+        for (auto t : type) {
+          counts.maybeNote(t);
         }
       }
```
aheejin, line 457:
```suggestion
```
Because we are doing tuple handling below, I guess we don't need to do this here again?

src/ir/module-utils.h
tlively: Oops, ignore me 😬

src/ir/module-utils.h
tlively: I don't think that's right (but I could be confused again).

If we have a tuple local that contains two different TFRs, `maybeNote` on the local type won't do anything because `maybeNote` only increments counts if the type is a TFR, not if it is a tuple. But we still want to count those two TFRs in the tuple because in the output binary they will be separate locals that refer to their type indices. (Although aren't local types run length encoded or something? I don't think we've been accounting for that, but I think that's fine for now because that would be complicated. Overcounting is better than undercounting here.)

aheejin: From the line 470+, we have
```cpp
    for (Type type : {sig.params, sig.results}) {
      for (auto element : type) {
        ...
```
Doesn't the `for (auto element : type)` take care of tuples?

aheejin: Oh no nevermind, locals may not be itself a signature but contain tuples I guess...

test/passes/optimize-instructions_all-features.wast
kripken: Oh, no optimization is possible there. The bug was just a very silly one. We always scan the parameters. We just need to not crash while doing that, if the type is a ref type. I'll add more to the comment, sorry for not being clear.

aheejin (approved):
I'm not sure if we want to wait for WebAssembly/function-references#42, but if not, I guess we can land it?

src/ir/module-utils.h
kripken: Yes, this line handles a local that is a tuple. So `(i32 f32 f64)`, something like that. We just look at the elements here for any signatures, and note those. Tuples aren't recursive so we can stop looking for tuples at that point.

The loop `for (auto element : type)` lower down handles multiple return values and multiple parameters, in a signature (which can be recursive). I actually don't know if each of those is an actual tuple type - would `isTuple` return true there..? If it would, then maybe this can all be simplified somehow...

tlively (approved):
Thanks for you patience with all the back-and-forth!

src/ir/module-utils.h
tlively: The params and results of a signature may or may not be tuples. For example, if a signature has no params, then the params would be `Type::none`, which is not a tuple. Similarly, if there is just one result, the results would be that type rather than a tuple.

kripken:
> I'm not sure if we want to wait for WebAssembly/function-references#42, but if not, I guess we can land it?

Not sure how long that will take... let's land this, and fix up later as necessary.

PR #3404, opened 2020-11-25
[NFC] Clean up sign/zero extension code in properties.h

tlively:
Use matchers and more descriptive variable names to clarify the intent of the
functions for finding and inspecting sign extension patterns.

### DIFF
--- a/src/ir/properties.h
+++ b/src/ir/properties.h
@@ -133,8 +133,8 @@ inline Literals getLiterals(const Expression* curr) {
 // that is extended, otherwise nullptr
 inline Expression* getSignExtValue(Expression* curr) {
   using namespace Match;
-  int32_t leftShift, rightShift;
-  Expression* extended;
+  int32_t leftShift = 0, rightShift = 0;
+  Expression* extended = nullptr;
   if (matches(curr,
               binary(ShrSInt32,
                      binary(ShlInt32, any(&extended), i32(&leftShift)),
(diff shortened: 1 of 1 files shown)


src/ir/properties.h
```diff
@@ -131,84 +132,73 @@ inline Literals getLiterals(const Expression* curr) {
...
-    }
+  using namespace Match;
+  int32_t leftShift, rightShift;
+  Expression* extended;
+  if (matches(curr,
+              binary(ShrSInt32,
+                     binary(ShlInt32, any(&extended), i32(&leftShift)),
+                     i32(&rightShift))) &&
+      Bits::getEffectiveShifts(rightShift, Type::i32) <=
+        Bits::getEffectiveShifts(leftShift, Type::i32) &&
```
tlively, line 167: To be honest, I have no idea why we are checking that the right shift is less than the left shift. Is it not still "almost" a sign extension if the result is shifted farther right rather than farther left? I would like to figure this out and improve the comment and possibly the function name as well.

tlively:
@kripken, I was reviewing #3376, but I had no idea what was going on. This is part of my efforts to properly review that PR.

src/ir/properties.h
kripken: There may be more patterns that are close to a sign-extend, yeah. In practice I noticed that it's common to have this one (`x << l >> r` where `l >= r`) so I optimized that, and didn't try to generalize any further.

(This was significant for code size back in the asm2wasm days.)

kripken (approved):
Sorry for the code quality here, this is all very old code.

PR #3405, opened 2020-11-25
[wasm-type] Introduce a function for walking type trees

tlively:
Since types can be arbitrarily deeply nested, even discovering all the types
used in a module requires being able to traverse type definition trees. This PR
introduces a function to do just that. It takes two callbacks, a function to
call on each visited type and a function to call whenever a recursion is
encountered. Note that it is not currently possible to construct a recursive
type, but the recursion callback will be useful in an upcoming PR to allow
constructing recursive types.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -16,6 +16,7 @@
 
 #include <array>
 #include <cassert>
+#include <set>
 #include <shared_mutex>
 #include <sstream>
 #include <unordered_map>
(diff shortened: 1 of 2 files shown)


tlively:
@kripken, feel free to steal as much of this as you want for use in #3401. That will probably get you going faster than if I actually making this PR ready to land by ripping out the recursion callback and adding tests.

src/wasm/wasm-type.cpp
```diff
@@ -390,6 +391,71 @@ bool Type::isRtt() const {
...
+      case HeapType::ArrayKind:
+        scanList.push_back(&ht.array.element.type);
+        break;
+    }
+  };
+
+  // Scan the type tree, noting recursion when we find it
+  while (scanList.size() != 0) {
+    Type* curr = scanList.back();
+    scanList.erase(scanList.end() - 1);
```
kripken, line 427:
```suggestion
    scanList.pop_back();
```

src/wasm/wasm-type.cpp
```diff
@@ -390,6 +391,71 @@ bool Type::isRtt() const {
   }
 }
 
+void Type::walk(std::function<void(const Type&)> noteRecursion,
+                std::function<void(const Type&)> visitType) {
+  std::set<Type> scanned;
+  std::vector<Type*> scanList = {this};
+  std::vector<Type*> visitList;
```
kripken, line 398: why use `Type*` for these? would it not work with `Type`?

kripken (approved):
lgtm to land now and test later, or if we end up using it in a key place, maybe it won't need separate tests.

src/wasm/wasm-type.cpp
tlively: Isn't pop_back introduced in C++20?

src/wasm/wasm-type.cpp
tlively: Ah, I adapted this from some code that needs to be able to modify the types in-place as they are visited, so I needed to reference the types' original locations. If you won't need this very soon, I'll probably make some changes to share code between the mutable and immutable versions and just include this as part of a larger PR.

src/wasm/wasm-type.cpp
kripken: No, I think it's been there pretty much all along? https://en.cppreference.com/w/cpp/container/vector/pop_back (looks like c++20 also added a constexpr of it though)

src/wasm/wasm-type.cpp
kripken: Take your time, I have other stuff to work on meanwhile then. I'll delay writing of GC stuff and focus on reading.

src/wasm/wasm-type.cpp
tlively: Yep, you're totally right, thanks :)

PR #3408, opened 2020-12-01
[module-splitting] Allow splitting with non-const table offsets

tlively:
Extend the splitting logic to handle splitting modules with a single table
segment with a non-const offset. In this situation the placeholder function
names are interpreted as offsets from the table base global rather than absolute
indices into the table. Since addition is not allowed in segment offset
expressions, the secondary module's segment must start at the same place as the
first table's segment. That means that some primary functions must be duplicated
in the secondary segment to fill any gaps. They are exported and imported as
necessary.

cc @sbc100 @awtcode I hope this is sufficient to get wasm-split working with
dynamic linking and PIC more generally.

### DIFF
--- a/src/emscripten-optimizer/istring.h
+++ b/src/emscripten-optimizer/istring.h
@@ -127,6 +127,8 @@ struct IString {
     return !str || str[0] == 0;
   }
 
+  operator bool() const { return !!*this; }
+
   const char* c_str() const { return str; }
   bool equals(const char* other) const { return !strcmp(str, other); }
 
--- a/src/emscripten-optimizer/istring.h
+++ b/src/emscripten-optimizer/istring.h
@@ -127,8 +127,6 @@ struct IString {
     return !str || str[0] == 0;
   }
 
-  operator bool() const { return !!*this; }
-
   const char* c_str() const { return str; }
   bool equals(const char* other) const { return !strcmp(str, other); }
 
--- a/src/ir/module-splitting.cpp
+++ b/src/ir/module-splitting.cpp
@@ -175,7 +175,7 @@ TableSlotManager::TableSlotManager(Module& module)
   });
 }
 
-auto TableSlotManager::getSlot(Name func) -> Slot {
+TableSlotManager::Slot TableSlotManager::getSlot(Name func) {
   auto slotIt = funcIndices.find(func);
   if (slotIt != funcIndices.end()) {
     return slotIt->second;
--- a/src/ir/module-splitting.cpp
+++ b/src/ir/module-splitting.cpp
@@ -126,7 +126,7 @@ struct TableSlotManager {
 Expression* TableSlotManager::Slot::makeExpr(Module& module) {
   Builder builder(module);
   auto makeIndex = [&]() { return builder.makeConst(int32_t(index)); };
-  if (global.is()) {
+  if (global.size()) {
     Expression* getBase = builder.makeGlobalGet(global, Type::i32);
     return index == 0 ? getBase
                       : builder.makeBinary(AddInt32, getBase, makeIndex());
@@ -434,7 +434,7 @@ void ModuleSplitter::setupTablePatching() {
     return;
   }
 
-  if (tableManager.activeBase.global.is()) {
+  if (tableManager.activeBase.global.size()) {
     assert(primary.table.segments.size() == 1 &&
            "Unexpected number of segments with non-const base");
     assert(secondary.table.segments.size() == 0);
--- a/src/ir/module-splitting.cpp
+++ b/src/ir/module-splitting.cpp
@@ -126,7 +126,7 @@ struct TableSlotManager {
 Expression* TableSlotManager::Slot::makeExpr(Module& module) {
   Builder builder(module);
   auto makeIndex = [&]() { return builder.makeConst(int32_t(index)); };
-  if (global) {
+  if (global.is()) {
     Expression* getBase = builder.makeGlobalGet(global, Type::i32);
     return index == 0 ? getBase
                       : builder.makeBinary(AddInt32, getBase, makeIndex());
@@ -434,7 +434,7 @@ void ModuleSplitter::setupTablePatching() {
     return;
   }
 
-  if (tableManager.activeBase.global) {
+  if (tableManager.activeBase.global.is()) {
     assert(primary.table.segments.size() == 1 &&
            "Unexpected number of segments with non-const base");
     assert(secondary.table.segments.size() == 0);
--- a/src/ir/module-splitting.cpp
+++ b/src/ir/module-splitting.cpp
@@ -160,7 +160,7 @@ TableSlotManager::TableSlotManager(Module& module)
     Index maxIndex = 0;
     for (auto& segment : table.segments) {
       assert(segment.offset->is<Const>() &&
-             "Unexpected non-const segment offset");
+             "Unexpected non-const segment offset with multiple segments");
       Index segmentBase = segment.offset->cast<Const>()->value.geti32();
       if (segmentBase + segment.data.size() >= maxIndex) {
         maxIndex = segmentBase + segment.data.size();
@@ -237,12 +237,12 @@ struct ModuleSplitter {
   void shareImportableItems();
 
   ModuleSplitter(Module& primary, const Config& config)
-    : config(config), secondaryPtr(ModuleSplitter::initSecondary(primary)),
-      primary(primary), secondary(*secondaryPtr),
-      classifiedFuncs(ModuleSplitter::classifyFunctions(primary, config)),
+    : config(config), secondaryPtr(initSecondary(primary)), primary(primary),
+      secondary(*secondaryPtr),
+      classifiedFuncs(classifyFunctions(primary, config)),
       primaryFuncs(classifiedFuncs.first),
       secondaryFuncs(classifiedFuncs.second), tableManager(primary),
-      exportedPrimaryFuncs(ModuleSplitter::initExportedPrimaryFuncs(primary)) {
+      exportedPrimaryFuncs(initExportedPrimaryFuncs(primary)) {
     moveSecondaryFunctions();
     thunkExportedSecondaryFunctions();
     indirectCallsToSecondaryFunctions();
@@ -457,13 +457,12 @@ void ModuleSplitter::setupTablePatching() {
          i < primarySeg.data.size() && replacement != replacedElems.end();
          ++i) {
       if (replacement->first == i) {
+        // primarySeg.data[i] is a placeholder, so use the secondary function.
         secondaryElems.push_back(replacement->second);
+        ++replacement;
       } else {
         exportImportFunction(primarySeg.data[i]);
         secondaryElems.push_back(primarySeg.data[i]);
-      }
-      if (replacement->first == i) {
-        ++replacement;
       }
     }
 
--- a/src/ir/module-splitting.cpp
+++ b/src/ir/module-splitting.cpp
@@ -62,12 +62,10 @@
 //      code that accesses them, but now that mutable-globals is shipped on all
 //      browsers, hopefully that extra complexity won't be necessary.
 //
-//   2. It assumes that all table segment offsets are constants. This simplifies
-//      the generation of segments to actively patch in the secondary functions
-//      without overwriting any other table slots. This assumption could be
-//      relaxed by 1) having secondary segments re-write primary function slots
-//      as well, 2) allowing addition in segment offsets, or 3) synthesizing a
-//      start function to modify the table instead of using segments.
+//   2. It assumes that either all table segment offsets are constants or there
+//      is exactly one segment that may have a non-constant offset. It also
+//      assumes that all segments are active segments (although Binaryen does
+//      not yet support passive table segments anyway).
 //
 //   3. It assumes that each function appears in the table at most once. This
 //      isn't necessarily true in general or even for LLVM output after function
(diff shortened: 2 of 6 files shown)


src/emscripten-optimizer/istring.h
```diff
@@ -127,6 +127,8 @@ struct IString {
     return !str || str[0] == 0;
   }
 
+  operator bool() const { return !!*this; }
```
kripken, line 130: Shouldn't this check `str`? Or am I missing something?

If I'm not missing something then I'm fine with an API like this. If so then we should be consistent, though, and remove `is()` and `isNull()`.

src/ir/module-splitting.cpp
```diff
@@ -90,81 +88,118 @@ namespace {
...
+
+    // Generate code to compute the index of this table slot
+    Expression* makeExpr(Module& module);
+  };
   Module& module;
   Table& table;
   Table::Segment* activeSegment = nullptr;
-  Index activeBase = 0;
-  std::map<Name, Index> funcIndices;
+  Slot activeBase = {"", 0};
```
kripken, line 116: may be nicer to remove the `= ..` stuff and add a ` = 0` in the struct definition? (the Name is initialized anyhow)

src/ir/module-splitting.cpp
```diff
@@ -90,81 +88,118 @@ namespace {
...
-  assert(it.second && "Function already has multiple indices");
+Expression* TableSlotManager::Slot::makeExpr(Module& module) {
+  Builder builder(module);
+  if (global) {
+    Expression* getBase = builder.makeGlobalGet(global, Type::i32);
+    return index == 0 ? getBase
+                      : builder.makeBinary(
+                          AddInt32, getBase, builder.makeConst(int32_t(index)));
+  } else {
+    return builder.makeConst(int32_t(index));
```
kripken, line 134: can deduplicate the `makeConst` into a local?

src/ir/module-splitting.cpp
```diff
@@ -228,6 +274,41 @@ ModuleSplitter::classifyFunctions(const Module& primary, const Config& config) {
...
+  if (exportIt != exportedPrimaryFuncs.end()) {
+    exportName = exportIt->second;
+  } else {
+    exportName = Names::getValidExportName(
+      primary, config.newExportPrefix + funcName.c_str());
+    primary.addExport(
+      Builder::makeExport(exportName, funcName, ExternalKind::Function));
+    exportedPrimaryFuncs[funcName] = exportName;
+  }
+  // Import the function it is not already imported into the secondary module.
```
kripken, line 302:
```suggestion
  // Import the function if it is not already imported into the secondary module.
```

src/ir/module-splitting.cpp
```diff
@@ -369,17 +428,57 @@ void ModuleSplitter::setupTablePatching() {
...
+
+  if (tableManager.activeBase.global) {
+    assert(primary.table.segments.size() == 1 &&
+           "Unexpected number of segments with non-const base");
+    assert(secondary.table.segments.size() == 0);
+    // Since addition is not currently allowed in initializer expressions, we
+    // need to start the new secondary segment where the primary segment starts.
+    // The secondary segment will contain the same primary functions as the
+    // primary module except in positions where it needs to overwrite a
+    // placeholder function. All primary functions in the table therefore need
```
kripken, line 444: Another option might be to put the secondary segment before the primary one. Then it doesn't matter what it has in places defined in the primary one, which will overwrite those.

src/emscripten-optimizer/istring.h
tlively: This uses `operator!` from above, which checks `str`. So this returns `true` iff the string exists and is not empty.

src/ir/module-splitting.cpp
tlively: Not without leaking it in cases where it is not used, but perhaps that doesn't matter so much. For now I'll deduplicate it into a small lambda to avoid leaks. Let me know what you think; happy to iterate further.

src/ir/module-splitting.cpp
tlively: The secondary segment is in the secondary module, so there's no way to load it before the primary segment.

src/emscripten-optimizer/istring.h
kripken: Oh, I see...

Ok, sounds good, assuming you are on board with the followup plan?

src/ir/module-splitting.cpp
kripken: Oh, right...

kripken (approved):

src/ir/module-splitting.cpp
kripken: Ah, I missed the leak issue...

I guess it's the same amount of code. Reading the lambda, I don't have a preference, so lgtm either way.

aheejin:
>  That means that some primary functions must be duplicated in the secondary segment to fill any gaps. They are exported and imported as necessary.

Can't we define a single dummy function in the second module in case of any gaps and use it to fill all them? For example, if we have three gap elements from the base, in the second module,
```wasm
(elem (global.get $base) $dummy $dummy $dummy $foo)
(func $dummy)
```

src/emscripten-optimizer/istring.h
tlively: I'm a little concerned about replacing `is` and `isNull` because they're not technically the same as this `operator bool`. Does the distinction between an empty string and a null string ever matter in practice? Should I be using `.is()` in this PR rather than treating null strings the same as empty strings?

tlively:
> > That means that some primary functions must be duplicated in the secondary segment to fill any gaps. They are exported and imported as necessary.
> 
> Can't we define a single dummy function in the second module in case of any gaps and use it to fill all them? For example, if we have three gap elements from the base, in the second module,
> 
> ```webassembly
> (elem (global.get $base) $dummy $dummy $dummy $foo)
> (func $dummy)
> ```

Unfortunately not. If we did that, then after the second module was loaded the table would be full of those dummy functions and indirect calls would start calling the dummy functions rather than the real functions.

aheejin:
Sorry, but I still don't get it.. Aren't indirect calls are to call placeholder functions from the first module? Why do they end up calling dummy functions in the second module, when those functions are not deferred and still in the second module?

If exporting and importing doesn't cost more, it should be fine anyway; I am just curious about the rationale.

tlively:
The primary module has placeholder functions in place of secondary functions in the table, but the idea is that after the secondary module is loaded, those placeholder functions should be patched out and the table should be exactly like it is in the original unsplit module. If the secondary module's segment contained a bunch of dummy functions, then those dummy functions would end up in the table after the secondary module is loaded. The table would never be the same as the original table in the unsplit module.

aheejin:
Ah right. They share the table. I somehow forgot that and kind of thought the tables are separate... Sorry my memory must be really bad. 😢

tlively:
Haha no worries :) There's no way your memory is worse than mine!

src/ir/module-splitting.cpp
```diff
@@ -228,6 +274,42 @@ ModuleSplitter::classifyFunctions(const Module& primary, const Config& config) {
...
+    exportName = exportIt->second;
+  } else {
+    exportName = Names::getValidExportName(
+      primary, config.newExportPrefix + funcName.c_str());
+    primary.addExport(
+      Builder::makeExport(exportName, funcName, ExternalKind::Function));
+    exportedPrimaryFuncs[funcName] = exportName;
+  }
+  // Import the function if it is not already imported into the secondary
+  // module.
```
aheejin, line 303: In which case the function can be already imported into the secondary module at this point?

src/ir/module-splitting.cpp
```diff
@@ -369,17 +429,57 @@ void ModuleSplitter::setupTablePatching() {
...
+           "Unexpected number of segments with non-const base");
+    assert(secondary.table.segments.size() == 0);
+    // Since addition is not currently allowed in initializer expressions, we
+    // need to start the new secondary segment where the primary segment starts.
+    // The secondary segment will contain the same primary functions as the
+    // primary module except in positions where it needs to overwrite a
+    // placeholder function. All primary functions in the table therefore need
+    // to be imported into the second module. TODO: use better strategies here,
+    // such as using ref.func in the start function or standardizing addition in
+    // initializer expressions.
```
aheejin, line 448: In theory we could only export functions in the gap slots, not all functions within the segment. But I'm not sure this is an important enough thing to optimize. (Does exporting / importing more functions cost us more?)

src/ir/module-splitting.cpp
```diff
@@ -369,17 +429,57 @@ void ModuleSplitter::setupTablePatching() {
...
+         i < primarySeg.data.size() && replacement != replacedElems.end();
+         ++i) {
+      if (replacement->first == i) {
+        secondaryElems.push_back(replacement->second);
+      } else {
+        exportImportFunction(primarySeg.data[i]);
+        secondaryElems.push_back(primarySeg.data[i]);
+      }
+      if (replacement->first == i) {
+        ++replacement;
```
aheejin, line 466: Can't this be in the same `if` statement in line 459?

src/ir/module-splitting.cpp
```diff
@@ -369,17 +429,57 @@ void ModuleSplitter::setupTablePatching() {
...
+    secondaryElems.reserve(primarySeg.data.size());
+
+    // Copy functions from the primary segment to the secondary segment,
+    // replacing placeholders and creating new exports and imports as necessary.
+    auto replacement = replacedElems.begin();
+    for (Index i = 0;
+         i < primarySeg.data.size() && replacement != replacedElems.end();
+         ++i) {
+      if (replacement->first == i) {
+        secondaryElems.push_back(replacement->second);
```
aheejin, line 460: At this point, are `replacement->second` and `primatySeg.data[i]` different? If they are the same, I guess changing this to `primarySeg.data[i]` to be consistent with `else` clause below will make it clear that we are doing the same action.

src/emscripten-optimizer/istring.h
tlively: @kripken what do you think about this? Does the difference between this `operator bool` and `is` matter?

src/ir/module-splitting.cpp
```diff
@@ -90,81 +88,118 @@ namespace {
...
+  forEachElement(table, [&](Name base, Index offset, Name func) {
+    addSlot(func, {base, offset});
+  });
 }
 
-Index TableSlotManager::getIndex(Name func) {
-  auto indexIt = funcIndices.find(func);
-  if (indexIt != funcIndices.end()) {
-    return indexIt->second;
+auto TableSlotManager::getSlot(Name func) -> Slot {
```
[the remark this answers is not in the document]
tlively: Otherwise the return type would have been `TableSlotManager::Slot`, which seemed overly verbose.

src/ir/module-splitting.cpp
```diff
@@ -90,81 +88,118 @@ namespace {
...
 
-  Index newIndex = activeBase + activeSegment->data.size();
+  Slot newSlot = {activeBase.global,
+                  activeBase.index + Index(activeSegment->data.size())};
   activeSegment->data.push_back(func);
-  addIndex(func, newIndex);
-  if (table.initial <= newIndex) {
-    table.initial = newIndex + 1;
+  addSlot(func, newSlot);
+  if (table.initial <= newSlot.index) {
```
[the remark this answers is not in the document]
tlively: Right, there's not really anything to do for an unknown global base. Keeping this code path means that at least setting the base to 0 will work. I don't know if that's useful in practice, but it doesn't seem like it would hurt anything either.

src/ir/module-splitting.cpp
```diff
@@ -196,7 +241,8 @@ struct ModuleSplitter {
       primary(primary), secondary(*secondaryPtr),
       classifiedFuncs(ModuleSplitter::classifyFunctions(primary, config)),
       primaryFuncs(classifiedFuncs.first),
-      secondaryFuncs(classifiedFuncs.second), tableManager(primary) {
+      secondaryFuncs(classifiedFuncs.second), tableManager(primary),
+      exportedPrimaryFuncs(ModuleSplitter::initExportedPrimaryFuncs(primary)) {
```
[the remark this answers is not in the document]
tlively: No we don't, thanks!

src/ir/module-splitting.cpp
tlively: For example if `exportImportFunction` has already been called on it. This can happen for a primary function that is called from a secondary function and is also in the table segment with a non-const offset. In that case both `exportImportCalledPrimaryFunctions` and `setupTablePatching` will call `exportImportFunction` on it.

src/ir/module-splitting.cpp
tlively: This early exit is what the `&& replacement != replacedElems.end()` condition in the loop below is doing. You can see this tested in module-splitting.txt. Search for "(elem (global.get $base) $foo $bar $baz)" and notice that it does not include $quux in the secondary segment.

My impression is that exporting and importing isn't too expensive, but we still want to avoid it where we can for code size.

src/ir/module-splitting.cpp
tlively: Yes, they are different. In the `if` case, `primarySeg.data[i]` would be a placeholder function but we want to use the original secondary function. In the `else` case, `primarySeg.data[i]` is the primary function we need to import/export and place in the secondary table.

I will add a comment about this.

src/ir/module-splitting.cpp
tlively: Oh yes, good point. I will combine them.

src/emscripten-optimizer/istring.h
kripken: I don't think the difference matters, but if it does then the test suite should catch it.

src/ir/module-splitting.cpp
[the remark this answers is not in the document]
tlively: FWIW, I changed this back to normal notation. A little extra verbosity seems less bad than strange syntax we don't use anywhere else.

aheejin (approved):
LGTM!

PR #3409, opened 2020-12-01
[GC types] Refactoring to allow future heap type parsing. NFC

kripken:
Defined types in wasm are really one of the "heap types": a signature type, or
(with GC) a struct or an array type. This refactors the binary and text parsers
to load the defined types into an array of heap types, so that we can start to
parse GC types. This replaces the existing array of signature types (which
could not support a struct or an array).

Also start to parse heap types in the s-expression code, enough to be able
to test this locally. For full testing we also need writing (since we have no
tests that just read but don't write - we could add one, but it would be very
temporary, so I'd rather not). The writing is waiting on @tlively 's work I believe.

Locally this PR can parse and print as text simple GC types. For that it was
necessary to also fix `Type::getFeatures` for GC.


tlively:
Is the writing blocked on me because writing the types will require traversing them?

Nit: the PR title suggests that arrays are parsed, but it doesn't look like that's part of the implementation, so it would be good to have a more specific title.

kripken:
@tlively 

Title changed, is that better? (Not sure what you meant about the title implying arrays were parsed, though - "array" wasn't in the title before?)

> Is the writing blocked on me because writing the types will require traversing them?

Yes. I'm pretty sure I'll need to make changes to `module-utils.h` (to collect not just signatures but all heap types etc.), and I'm waiting on your work there, so we don't conflict?

tlively:
Yes, that seems better, thanks. I had interpreted "Heap type parsing" to mean parsing _all_ heap types, including arrays.

I'll try to get my type building PR up ASAP to get you unblocked. Sorry for the blockage!

src/wasm/wasm-s-parser.cpp
```diff
@@ -2784,36 +2758,89 @@ void SExpressionWasmBuilder::parseInnerElem(Element& s,
...
+      if (last->isList() && *(*last)[0] == MUT) {
+        mutable_ = true;
+        last = (*last)[1];
+      }
+      fields.emplace_back(elementToType(*last), mutable_);
     }
+    return Struct(fields);
   }
-  return Signature(Type(params), Type(results));
+  throw ParseException("invalid heap type", s.line, s.col);
```
aheejin, line 2830: I think by taking this struct and array parsing part out to the next PR (+ and `getFeatures` function change in wasm-type.cpp), this PR will become purely NFC refactoring and will not be blocked by anything else, and we can land this array/struct parsing hopefully with tests. I think that would make this PR simpler, WDYT?

src/wasm/wasm-binary.cpp
```diff
@@ -5293,8 +5287,8 @@ void WasmBinaryBuilder::visitRefFunc(RefFunc* curr) {
   // To support typed function refs, we give the reference not just a general
   // funcref, but a specific subtype with the actual signature.
   // FIXME: for now, emit a nullable type here
-  curr->finalize(
-    Type(HeapType(getFunctionSignatureByIndex(index)), /* nullable = */ true));
+  curr->finalize(Type(HeapType(getSignatureByFunctionIndex(index)),
+                      /* nullable = */ true));
```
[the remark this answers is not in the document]
kripken: I think this would be incorrect. The index here is the index of the *function*, hence we call "get signature by function index". And `types` is the list of types by type index.

src/wasm/wasm-s-parser.cpp
kripken: I agree it could be simpler that way. Happy to refactor it if you prefer. However, the benefit to doing it this way is that it is actually testable. This is, I think, the minimal amount of code to be testable. (But a test is not added, since adding a new test mode temporarily seems a bad idea.)

kripken: I removed the struct parsing now.

tlively (approved):

src/wasm/wasm-binary.cpp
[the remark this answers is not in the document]
aheejin: Oh, you're right. Sorry!

aheejin (approved):
I think that refactoring PRs like this don't have tests changed is fine; it is the very definition of refactoring, no? 😅 The correctness is validated by that all existing tests pass.

PR #3410, opened 2020-12-02
[Printing] Print type names where possible.

kripken:
For a nested type, we used to print e.g.
```wat
(param $x (ref (func (param i32))))
```
Instead of expanding the full type inline, which can get long for
a deeply nested type, print a name when running the Print pass.
In this example that would be something like
```wat
(param $x (ref $i32_=>_none))
```

### DIFF
--- a/src/passes/Print.cpp
+++ b/src/passes/Print.cpp
@@ -133,6 +133,24 @@ std::ostream& operator<<(std::ostream& os, SigName sigName) {
   return os;
 }
 
+// Wrapper for printing type names when they are available. When not, the basic
+// type name (i32, etc.) is used.
+struct TypeName {
+  Type type;
+  TypeName(Type type) : type(type) {}
+};
+
+std::ostream& operator<<(std::ostream& os, TypeName typeName) {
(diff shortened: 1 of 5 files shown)


src/passes/Print.cpp
```diff
@@ -2328,11 +2355,23 @@ struct PrintSExpression : public OverriddenVisitor<PrintSExpression> {
     }
     if (curr.params.size() > 0) {
       o << maybeSpace;
-      o << ParamType(curr.params);
```
tlively, line 2331: Were these the only places in which we used `ParamType` and `ResultType`? Can we remove them?

tlively (approved):
I was thinking the other day that even these type names may become unwieldy because they are generated from the structure of the types, but we can cross that bridge when we get to it.

src/passes/Print.cpp
kripken: They are still used in event printing, which I'm not sure what to do about, but if we decide to do it there as well then we can remove them. However, they may also be useful for debugging.

PR #3411, opened 2020-12-02
Make NUM_PARAMS in FuncCastEmulation a configuration option

dalcde:
Compiling scipy requires a `NUM_PARAMS` of at least 61; [https://github.com/iodide-project/pyodide](pyodide) patches emsdk in order to compile.


src/passes/FuncCastEmulation.cpp
```diff
@@ -38,7 +38,7 @@ namespace wasm {
 // This should be enough for everybody. (As described above, we need this
 // to match when dynamically linking, and also dynamic linking is why we
 // can't just detect this automatically in the module we see.)
-static const int NUM_PARAMS = 16;
+static const int NUM_PARAMS = 61;
```
kripken, line 41: (Wow, SciPy needs that many...? Couldn't it be filling in a struct and passing a pointer to it maybe..? Probably they have a good reason though.)

This has a downside for both code size and runtime speed. How about making it a parameter to the pass? See the `ExtractFunction` pass for an example of a tiny pass that has a parameter. And it's then easy to tell emcc to pass the parameter to wasm-opt when it runs, with the `BINARYEN_EXTRA_PASSES` flag.

dalcde:
Now done. No new tests have been added for the configuration option.

dalcde:
Tests for new feature added. The new test, being generated by `auto_update_tests`, by definition passes, but diffing with the version without `max-func-params=5` shows it to be quite reasonable.

src/passes/FuncCastEmulation.cpp
```diff
@@ -183,12 +185,12 @@ struct FuncCastEmulation : public Pass {
...
     }
     // update call_indirects
-    ParallelFuncCastEmulation(ABIType).run(runner, module);
+    ParallelFuncCastEmulation(ABIType, num_params).run(runner, module);
   }
 
 private:
   // Creates a thunk for a function, casting args and return value as needed.
-  Name makeThunk(Name name, Module* module) {
+  Name makeThunk(Name name, Module* module, unsigned int num_params) {
```
kripken, line 193: please change the `unsigned int` to `Index`, and `num_params` to `numParams`

test/passes/fpcast-emu_pass-arg=max-func-params@5.wast
```diff
@@ -0,0 +1,80 @@
+(module
+  (type $vijfd (func (param i32) (param i64) (param f32) (param f64)))
+  (type $jii (func (param i32) (param i32) (result i64)))
+  (type $fjj (func (param i64) (param i64) (result f32)))
+  (type $dff (func (param f32) (param f32) (result f64)))
+  (type $idd (func (param f64) (param f64) (result i32)))
+  (import "env" "imported_func" (func $imported-func (param i32 i64 f32 f64) (result f32)))
+  (table 10 10 funcref)
+  (elem (i32.const 0) $a $b $c $d $e $e $imported-func)
```
kripken, line 9: this test could be much shorter and more focused I think. How about having just a module with a table and a single function in that table? Then we'd see it's instrumented to have five params, and that would show it works.

dalcde: Sounds reasonable. Should I change the original fpcast-emu test too?

src/passes/FuncCastEmulation.cpp
```diff
@@ -30,16 +30,12 @@
 
 #include <ir/literal-utils.h>
```
kripken, line 31:
```suggestion
#include <string>

#include <ir/literal-utils.h>
```

src/passes/FuncCastEmulation.cpp
```diff
@@ -30,16 +30,12 @@
...
 #include <pass.h>
+#include <string>
```
kripken, line 33:
```suggestion
```

src/passes/FuncCastEmulation.cpp
kripken: (standard C++ imports should go before the binaryen ones)

kripken (approved):

PR #3412, opened 2020-12-02
[wasm-split] Record checksums in profiles

tlively:
Calculate a checksum of the original uninstrumented module and emit it as part
of the profile data. When reading the profile, compare the checksum it contains
to the checksum of the module that is being split. Error out if the module being
split is not the same as the module that was originally instrumented.

Also fixes a bug in how the profile data was being read. When `char` is signed,
bytes read from the profile were being incorrectly sign extended. We had not
noticed this before because the profiles we have tested have contained only
small-valued counts.

cc @sbc100 @awtcode.

### DIFF
--- a/src/tools/wasm-split.cpp
+++ b/src/tools/wasm-split.cpp
@@ -359,7 +359,7 @@ void Instrumenter::instrumentFuncs() {
   });
 }
 
-// wasm-split profile format::
+// wasm-split profile format:
 //
 // The wasm-split profile is a binary format designed to be simple to produce
 // and consume. It is comprised of:
@@ -443,14 +443,23 @@ void Instrumenter::addProfileExport() {
   // TODO: export the memory if it is not already exported.
 }
 
+uint64_t hashFile(const std::string& filename) {
+  auto contents(read_file<std::vector<char>>(filename, Flags::Binary));
+  size_t digest = 0;
+  // Don't use `hash` or `rehash` - they aren't deterministic between executions
+  for (char c : contents) {
+    hash_combine(digest, c);
+  }
+  return uint64_t(digest);
+}
+
 void instrumentModule(Module& wasm, const WasmSplitOptions& options) {
   // Check that the profile export name is not already taken
   if (wasm.getExportOrNull(options.profileExport) != nullptr) {
     Fatal() << "error: Export " << options.profileExport << " already exists.";
   }
 
-  // TODO: calculate module hash.
-  uint64_t moduleHash = 0;
+  uint64_t moduleHash = hashFile(options.input);
   PassRunner runner(&wasm, options.passOptions);
   Instrumenter(options.profileExport, moduleHash).run(&runner, &wasm);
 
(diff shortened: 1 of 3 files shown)


src/tools/wasm-split.cpp
```diff
@@ -443,14 +443,23 @@ void Instrumenter::addProfileExport() {
...
 }
 
+uint64_t hashFile(const std::string& filename) {
+  auto contents(read_file<std::vector<char>>(filename, Flags::Binary));
+  size_t digest = 0;
+  // Don't use `hash` or `rehash` - they aren't deterministic between executions
+  for (char c : contents) {
+    hash_combine(digest, c);
+  }
+  return uint64_t(digest);
```
aheejin, line 453: `digest` is `size_t`, then why do we need `uint64_t` to store a module hash?

test/lit/wasm-split/mismatched-hashes.wast
```diff
@@ -0,0 +1,20 @@
...
+
+;; Instrument the module
+;; RUN: wasm-split --instrument %s -o %t.instrumented.wasm
+
+;; Generate a profile
+;; RUN: node %S/call_exports.mjs %t.instrumented.wasm %t.prof
+
+;; Attempt to split the instrumented module
+;; RUN: not wasm-split %t.instrumented.wasm --profile=%t.prof -o1 %t.1.wasm -o2 %t.2.wasm \
+;; RUN:   2>&1 | filecheck %s
```
aheejin, line 12: How about adding a `RUN` line that runs successfully too?

aheejin (approved):

src/tools/wasm-split.cpp
tlively: Using uint64_t makes the profile parsing simpler because we can unconditionally use 64 bits in the profile rather than having to determine whether to use 32 or 64 bits depending on whether we are running on a 32-bit platform.

test/lit/wasm-split/mismatched-hashes.wast
```diff
@@ -15,6 +15,9 @@
 ;; CHECK-SAME: The split module must be the original module that was instrumented
 ;; CHECK-SAME: to generate the profile.
 
+;; Check that the matcing module succeeds
```
aheejin, line 18: matching* (But given that it is already merged, maybe let's fix this in the next PR, or nevermind...)

src/tools/wasm-split.cpp
aheejin: But it seems we are using only 32 bits, then can't we use 32 bits unconditionally?

test/lit/wasm-split/mismatched-hashes.wast
tlively: Oops, thanks for the catch. I'll keep it in mind to fix in the next PR.

src/tools/wasm-split.cpp
aheejin: Oh, sorry, I somehow assumed `size_t`'s size is the same as `unsigned int`... it seems to be implementation-defined. Nevermind.

PR #3413, opened 2020-12-02
[OptimizeInstructions] Fix a fuzz bug with getting the shifts of an unreachable

kripken:
This code was edited recently but it seems to not be a regression, rather a
bug that was there all along.


aheejin (approved):
I don't know enough about its usage, but it seems correct given that the comment says it returns the expression only when it is able to extend the value. Shouldn't we do the same check in `getZeroExtValue` too?

kripken:
@aheejin good point, added!

PR #3415, opened 2020-12-02
[Print] Add support for printing GC type names. NFC

kripken:
Generalize `SigName` to `HeapTypeName` so it can support all heap types.

NFC as we cannot parse GC types yet anyhow.

### DIFF
--- a/src/passes/Print.cpp
+++ b/src/passes/Print.cpp
@@ -80,13 +80,20 @@ static std::ostream& operator<<(std::ostream& o, const SExprType& localType) {
   return o;
 }
 
-// Wrapper for printing signature names
-struct SigName {
-  Signature sig;
-  SigName(Signature sig) : sig(sig) {}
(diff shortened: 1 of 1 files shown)


aheejin:
Wouldn't this kind of PR (especially parsing) be better with real array/struct tests..? It's hard know what the code is printing when there's no wast files...

kripken:
I agree, but this is also necessary for that parsing to be tested, as the test will need to both print and parse. So there is a chicken and egg problem.

I can submit one big PR at the end if you prefer.

kripken:
I checked and @tlively also prefers tested but possibly larger PRs, so I'll close this for now, sorry for the noise.

aheejin:
I think it depends on the situation, but I personally prefer tested PRs especially in case of parsing and printing, because it is clear we know what we are printing and parsing.. 😅

PR #3417, opened 2020-12-03
[Types] Return const& for HeapTypes and their Structs and Arrays. NFC

kripken:
I am starting to write lines like
```cpp
curr->value->type.getHeapType().getStruct().fields[curr->index].type
```
and it scares me to think that there may be copies going on there.
The issue is that Types are interned, but HeapTypes and their components
are not.

Talking to @tlively , we think this should avoid the perf issue and also
allow for possible interning of more things in the future. I am happy to see
that it compiles without needing any changes.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -513,24 +513,29 @@ FeatureSet Type::getFeatures() const {
   return getSingleFeatures(*this);
 }
 
-HeapType Type::getHeapType() const {
+static HeapType funcHeapType(HeapType::FuncKind),
+  externHeapType(HeapType::ExternKind), exnHeapType(HeapType::ExnKind),
+  anyHeapType(HeapType::AnyKind), eqHeapType(HeapType::EqKind),
+  i31HeapType(HeapType::I31Kind);
+
+const HeapType& Type::getHeapType() const {
   if (isRef()) {
     if (isCompound()) {
       return getTypeInfo(*this)->ref.heapType;
     }
     switch (getBasic()) {
       case funcref:
-        return HeapType::FuncKind;
+        return funcHeapType;
       case externref:
-        return HeapType::ExternKind;
+        return externHeapType;
       case exnref:
-        return HeapType::ExnKind;
+        return exnHeapType;
       case anyref:
-        return HeapType::AnyKind;
+        return anyHeapType;
       case eqref:
-        return HeapType::EqKind;
+        return eqHeapType;
       case i31ref:
-        return HeapType::I31Kind;
+        return i31HeapType;
       default:
         break;
     }
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -513,10 +513,17 @@ FeatureSet Type::getFeatures() const {
   return getSingleFeatures(*this);
 }
 
+// getHeapType() returns a const HeapType&, so we need a canonical object to
+// return for the basic types, so that we don't create a temporary copy on each
+// call.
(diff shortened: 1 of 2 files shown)


src/wasm/wasm-type.cpp
```diff
@@ -513,24 +513,29 @@ FeatureSet Type::getFeatures() const {
...
-HeapType Type::getHeapType() const {
+static HeapType funcHeapType(HeapType::FuncKind),
+  externHeapType(HeapType::ExternKind), exnHeapType(HeapType::ExnKind),
+  anyHeapType(HeapType::AnyKind), eqHeapType(HeapType::EqKind),
+  i31HeapType(HeapType::I31Kind);
+
+const HeapType& Type::getHeapType() const {
   if (isRef()) {
     if (isCompound()) {
       return getTypeInfo(*this)->ref.heapType;
```
kripken, line 524: I believe the type interning allows this line to return a reference to a HeapType that stays around forever.

aheejin:
How about making these also a part of [`TypeStore`](https://github.com/WebAssembly/binaryen/blob/00f96c854ed3691b01fa35bba0a1d010d08958bd/src/wasm/wasm-type.cpp#L235)?

src/wasm/wasm-type.cpp
```diff
@@ -513,24 +513,29 @@ FeatureSet Type::getFeatures() const {
   return getSingleFeatures(*this);
 }
 
-HeapType Type::getHeapType() const {
+static HeapType funcHeapType(HeapType::FuncKind),
+  externHeapType(HeapType::ExternKind), exnHeapType(HeapType::ExnKind),
+  anyHeapType(HeapType::AnyKind), eqHeapType(HeapType::EqKind),
+  i31HeapType(HeapType::I31Kind);
```
tlively, line 519: Can you move these declarations into `Type::getHeapType` (but keep them static) so they don't pollute the global namespace in this file?

tlively (approved):
@aheejin, I'm not sure what you have in mind.

src/wasm/wasm-type.cpp
aheejin: This is more or less what I meant. I thought putting these into a class will not clutter the global namespace. I guess, as you suggested, `Type` is better than `TypeStore` for this, given that it already has all primitive types.

aheejin (approved):

src/wasm/wasm-type.cpp
kripken: Putting these inside the function would be less efficient, though? It would use atomic operations to create the object on the first call, and check constantly if it is the first call.

How about putting it in a `statics` local namespace? I pushed a commit with that.

tlively: Oh, TIL!

tlively (approved):

src/wasm/wasm-type.cpp
aheejin: Um, aren't file static variables only visible within that file anyway? (I said something about the global namespace but nevermind that, it was not accurate; I guess I thought if it was within `Type::` it would be more cleaner.)

tlively: Yeah, I was just hoping to hide them from the rest of wasm-type.cpp, but that's not worth the performance impact putting them in the function would have. I'm fine with the way it is now in the `statics` namespace 👍

kripken: I'd be fine with any idea to improve this in a followup. But I do think it's reasonably encapsulated now.

PR #3418, opened 2020-12-03
TypeBuilder

tlively:
Introduce TypeBuilder, a utility for constructing heap types in terms of other
heap types that may have not yet been defined. Internally, it works by using new
TypeInfo variants that contain pointers to HeapTypes rather than HeapType
values. That lets the TypeBuilder create temporary types that can refer to
uninitialized heap types owned by the TypeBuilder. Those temporary types can in
turn be used to initialize the very heap types they refer to. Since temporary
types are only valid for the lifetime of their TypeBuilder, there is a
canonicalization step that converts the temporary types reachable from the
TypeBuilder's heap types into globally interned canonical types. Only after that
canonicalization step can the heap types be used to construct non-temporary
types.

This PR allows Types to be built in terms of as of yet undefined heap types, but
it currently errors out in the presence of recursive types. Supporting recursive
types will require further work to canonicalize them into finite, acyclic
representations. Currently any attempt to compare recursive types would
infinitely recurse.

cc @dcodeIO


src/wasm-type.h
```diff
@@ -396,6 +397,37 @@ struct Rtt {
...
+  ~TypeBuilder();
+
+  TypeBuilder(TypeBuilder& other) = delete;
+  TypeBuilder(TypeBuilder&& other) = delete;
+  TypeBuilder& operator=(TypeBuilder&) = delete;
+
+  HeapType& operator[](size_t i) { return heapTypes[i]; }
+
+  Type getTempTupleType(const Tuple&);
+  Type getTempRefType(size_t i, bool nullable);
```
kripken, line 423: A comment explaining what `i` is might help. Reading this the first time, I have no guesses as to what it could be.

src/wasm/wasm-type.cpp
```diff
@@ -32,25 +33,44 @@ struct TypeInfo {
...
     bool nullable;
   };
+  struct TempRef {
+    HeapType* heapType;
+    bool nullable;
+  };
+  struct TempRtt {
+    uint32_t depth;
+    HeapType* heapType;
+  };
```
kripken, line 50: This makes me wonder once again if maybe HeapTypes should always be by a pointer. That is, that on line 37/40 we'd have `HeapType*` instead of `HeapType`.

I may be missing something, but it seems like that would make it simpler to implement the TypeBuilder, as we wouldn't need to add `TempRef` and `TempRtt`. And, separately, it would also avoid the risk of copies being passed around.

If memory safety is a concern here, then our usage pattern is very simple - allocate types when creating a wasm, and then use them. We could just store them in a global arena that is freed on process exit, like strings are. Note that that is a separate question from whether we intern them or not (which I am not sure about either way).

One of my concerns about `TempRef` and `TempRtt`, more specifically, is that they end up intruding in various places throughout this file. If somehow the TypeBuilder could encapsulate that complexity into itself then that would be another option to think about.

src/wasm/wasm-type.cpp
```diff
@@ -950,8 +1030,229 @@ std::ostream& operator<<(std::ostream& os, TypeInfo info) {
...
+  return impl->store.makeType(TypeInfo::TempRef{&heapTypes[i], nullable});
+}
+
+Type TypeBuilder::getTempRttType(size_t i, uint32_t depth) {
+  assert(i < heapTypes.size() && "Index out of bounds");
+  return impl->store.makeType(TypeInfo::TempRtt{depth, &heapTypes[i]});
+}
+
+// Implements the algorithm to canonicalize the HeapTypes in a TypeBuilder
+struct Canonicalizer {
```
kripken, line 1077: Reading this the first time, I am unclear on what "canonicalization" means. Perhaps a concrete example in a comment could help?

Even reading a second time I am still unsure. I realize that canonicalization is important for recursive types, but as this appears in this PR, it seems it matters in an additional way. My intuition is that without recursion we just allocate N types, if there are N types in the type section, and then set up pointers between them (which is why we need the TypeBuilder). Those pointers could be "final", that is, we allocated N types up front, and those are all of our types. We just need to replace the pointers with copies (but see the comment earlier about pointers), essentially. That seems like it should be possible since each Type provides an indirection (as types are internalized and used by reference). That is, since HeapTypes do not have direct references to each other, the Type indirection gives us all we need. Clearly I'm missing something though :smile: What is it?

kripken (commented):
Oh, and just to make sure about the overall plan: This PR does not modify `module-utils.h`, but your next PR would?

src/wasm/wasm-type.cpp
tlively: Maybe the missing piece is that the TypeInfos in the TypeBuilder's TypeStore need to be merged into the globalTypeStore? It can't be as simple as moving the `std::unique_ptr<TypeInfo>`s from one TypeStore to the other because canonical versions of the TypeBuilder's types might already exist in the globalTypeStore. That's why all the temporary types in the type graph need to be replaced with their globally canonical versions.

tlively:
> Oh, and just to make sure about the overall plan: This PR does not modify `module-utils.h`, but your next PR would?

The next PR would use TypeBuilder in the binary and text parsers. The PR after that would update the type finding code in module-utils.h to do a full type traversal and would remove the code that does the almost-topographical sort of the types.

src/wasm/wasm-type.cpp
tlively: I'll experiment with this.

tlively:
@kripken I rewrote this on top of #3428 and expanded many of the comments. I think it's much cleaner now 👍 PTAL!

src/wasm/wasm-type.cpp
```diff
@@ -961,6 +991,267 @@ std::ostream& operator<<(std::ostream& os, HeapTypeInfo info) {
...
+    case TypeInfo::RefKind:
+      noteChild(*type, &info->ref.heapType);
+      break;
+    case TypeInfo::RttKind:
+      noteChild(*type, &info->rtt.heapType);
+      break;
+  }
+}
+
+void Canonicalizer::makeReachabilityFixpoint() {
```
kripken, line 1215: Isn't ["Fixed Point"](https://en.wikipedia.org/wiki/Fixed_point_(mathematics)) the term? Or is "fixpoint" a common shorthand?
```suggestion
void Canonicalizer::makeReachabilityFixedPoint() {
```

src/wasm/wasm-type.cpp
```diff
@@ -961,6 +991,267 @@ std::ostream& operator<<(std::ostream& os, HeapTypeInfo info) {
...
+  // The fully canonicalized heap types.
+  std::vector<HeapType> results;
+
+  Canonicalizer(TypeBuilder& builder);
+  template<typename T1, typename T2> void noteChild(T1 parent, T2* child);
+  void scanHeapType(HeapType* ht);
+  void scanType(Type* child);
+  void makeReachabilityFixpoint();
+  template<typename T>
+  void canonicalize(T* type, std::unordered_map<T, T>& canonicals);
```
kripken, line 1100: maybe a comment on this method? the out param is not immediately obvious.

src/wasm/wasm-type.cpp
```diff
@@ -961,6 +991,267 @@ std::ostream& operator<<(std::ostream& os, HeapTypeInfo info) {
...
+  impl->entries[i].set(std::move(struct_));
+}
+
+void TypeBuilder::setHeapType(size_t i, Array array) {
+  assert(i < impl->entries.size() && "index out of bounds");
+  impl->entries[i].set(array);
+}
+
+Type TypeBuilder::getTempTupleType(const Tuple& tuple) {
+  return impl->typeStore.canonicalize(tuple);
```
kripken, line 1039: How does this work? For a Ref type there is an index i, which gives us the necessary indirection, but here the Tuple contains actual Types..? (There also doesn't seem to be a test for `getTempTupleType`.)

src/wasm/wasm-type.cpp
tlively: That wikipedia article starts "In mathematics, a fixed point (sometimes shortened to fixpoint" so I think we could go either way.

kripken: Oh, sorry, not sure how I missed that in the link...

src/wasm/wasm-type.cpp
tlively: This is for when you need to create a tuple of temporary types, for example when constructing a function signature. You can't create a normal tuple type for that because that would allow references to the temporary types to leak out into the global type store. In the worst case, a signature type created later could be canonicalized to a type that refers to the original temporary types which have long since become invalid.

src/wasm/wasm-type.cpp
```diff
@@ -309,6 +326,19 @@ using HeapTypeStore = Store<HeapTypeInfo>;
 TypeStore globalTypeStore;
 HeapTypeStore globalHeapTypeStore;
 
+// Specialized to simplify programming generically over Types and HeapTypes.
+template<typename T> struct MetaTypeInfo {};
```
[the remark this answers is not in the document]
tlively: I think it is useful to maintain the distinction between them, despite their similarities. They're certainly not interchangeable in the public API.

src/wasm/wasm-type.cpp
```diff
@@ -961,6 +991,267 @@ std::ostream& operator<<(std::ostream& os, HeapTypeInfo info) {
   WASM_UNREACHABLE("unexpected kind");
 }
 
+struct TypeBuilder::Impl {
+  TypeStore typeStore;
+  HeapTypeStore heapTypeStore;
+
+  struct Entry {
+    bool initialized = false;
+    HeapTypeInfo info = Signature();
```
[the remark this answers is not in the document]
tlively: `HeapTypeInfo` does not have a default initializer, but we want to be able to default initialize `Entry`, so we have to give `info` an explicit default value. I'll add a comment clarifying that it is arbitrary.

src/wasm/wasm-type.cpp
tlively: no problem :) I'll go with your suggestion since it is indeed the more standard name.

src/wasm/wasm-type.cpp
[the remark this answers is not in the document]
kripken: That's a fair point that they matter in the public API. But OTOH, the public API also differentiates struct from array types, but we don't have a StructType or ArrayType. In other words, we could have a single Type, with a `isHeap()` like we have `isTuple(),isRef(),isRtt()` etc. Maybe I'm overthinking this though.

src/wasm/wasm-type.cpp
kripken: I think I see. Is this an internal API then (and therefore not tested as part of the public API)?

kripken: No, wait, the user needs to create those tuples, for each function..?

tlively: Right, users need to do this. I forgot to say that I will add a test for this 👍

kripken: Hmm, why are tuples different than everything else, though? The others all get `size_t i`.

src/wasm/wasm-type.cpp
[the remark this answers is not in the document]
aheejin: I personally think having them separately is clearer. They represent fundamentally different kinds of type: `Type` is what every value and expression has, and `HeapType` is what we list in the type section, and they cannot be used interchangeably and one is not a subtype of the other. On the other hand, all of tuple, ref, and rtt are subtypes of `Type`.

src/wasm/wasm-type.cpp
tlively: Tuples don't have associated HeapTypes, so they don't need to take a heap type index. They still need to be backed by the TypeBuilder's TypeStore, though.

test/example/type-builder.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+using namespace wasm;
+
+// Construct Signature, Struct, and Array heap types using undefined types.
+void test_builder() {
+  std::cout << ";; Test TypeBuilder\n";
+
+  // (type $sig (func (param (ref $struct)) (result (ref $array) i32)))
+  // (type $struct (struct (field (ref null $array) (mut ref extern))))
+  // (type $array (array (mut ref null extern)))
```
aheejin, line 14:
```suggestion
  // (type $struct (struct (field (ref null $array) (mut externref))))
  // (type $array (array (mut externref)))
```
Nit: For simplicity (and also to match the output below)

aheejin (commented):
Haven't finished reading, but some initial questions:

test/example/type-builder.txt
```diff
@@ -0,0 +1,21 @@
...
+(ref $sig) => (ref (func (param (ref (struct (field (ref null (array (mut externref))) (mut (ref extern)))))) (result (ref (array (mut externref))) i32)))
+(ref $struct) => (ref (struct (field (ref null (array (mut externref))) (mut (ref extern)))))
+(ref $array) => (ref (array (mut externref)))
+(ref null $array) => (ref null (array (mut externref)))
+
+After building types:
+(ref $sig) => (ref (func (param (ref (struct (field (ref null (array (mut externref))) (mut (ref extern)))))) (result (ref (array (mut externref))) i32)))
+(ref $struct) => (ref (struct (field (ref null (array (mut externref))) (mut (ref extern)))))
+(ref $array) => (ref (array (mut externref)))
+(ref null $array) => (ref null (array (mut externref)))
```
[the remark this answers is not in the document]
tlively: The "building" step replaces the temporary types and heap types with permanent types and heap types. There is no way to differentiate between temporary and permanent types while printing, unfortunately.

`refStruct` was initialized by line `builder.setHeapType(1, struct_);`, which replaces the HeapTypeInfo backing the HeapType that `refStruct` refers to. This mutability of HeapTypeInfos is the key feature of the TypeBuilder that allows it to create types before the types they depend on have been created.

test/example/type-builder.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+  TypeBuilder builder(3);
+
+  Type refSig = builder.getTempRefType(0, false);
+  Type refStruct = builder.getTempRefType(1, false);
+  Type refArray = builder.getTempRefType(2, false);
+  Type refNullArray = builder.getTempRefType(2, true);
+  Type refExt(HeapType::ext, false);
```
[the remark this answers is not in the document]
tlively: Unfortunately we can't use `extern` because it is a keyword. Between the alternatives `extern_` and `ext` I thought `ext` looked better, but perhaps the ugliness of `extern` would be worth the clarity. @kripken, what do you think about this? I'd be happy to change it if we collectively think `extern_` would be better.

test/example/type-builder.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+  Type newRefSig = Type(built[0], false);
+  Type newRefStruct = Type(built[1], false);
+  Type newRefArray = Type(built[2], false);
+  Type newRefNullArray = Type(built[2], true);
+
+  assert(refSig != newRefSig);
+  assert(refStruct != newRefStruct);
+  assert(refArray != newRefArray);
+  assert(refNullArray != newRefNullArray);
```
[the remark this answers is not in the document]
tlively: The types on the left are temporary types whose lifetime ends with the TypeBuilder and he types on the right are permanent globally canonical types.

src/wasm-type.h
```diff
@@ -449,6 +455,42 @@ struct Rtt {
...
+// TypeBuilder - allows for the construction of recursive types. Contains a
+// table of `n` mutable HeapTypes and can construct temporary types that are
+// backed by those HeapTypes, refering to them by reference. Those temporary
+// types are owned by the TypeBuilder and should only be used in the
+// construction of HeapTypes to insert into the TypeBuilder. Temporary types
+// should never be used in the construction of normal Types, only other
+// temporary types.
+struct TypeBuilder {
+  struct Impl;
+  std::unique_ptr<Impl> impl;
```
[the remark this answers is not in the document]
tlively: We need to hide the definition of `Impl` in the .cpp file because it depends on implementation details like `HeapTypeInfo`. We can't do that and have `Impl impl;` because hiding the definition necessarily makes `Impl` and incomplete type at this point. This is the classic [`PImpl` idiom](https://en.cppreference.com/w/cpp/language/pimpl).

src/wasm/wasm-type.cpp
```diff
@@ -974,6 +1004,276 @@ std::ostream& operator<<(std::ostream& os, HeapTypeInfo info) {
...
+  struct Item {
+    enum Kind {
+      TypeKind,
+      HeapTypeKind,
+    } kind;
+    union {
+      Type* type;
+      HeapType* heapType;
+    };
+    Item(Type* type) : kind(TypeKind), type(type) {}
```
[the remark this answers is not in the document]
tlively: Yes, notice that `scanHeapType` records `Type*`s to scan in the future and `scanType` records both `HeapType*` and `Type*` (the latter in the tuple case).

src/wasm/wasm-type.cpp
```diff
@@ -974,6 +1004,276 @@ std::ostream& operator<<(std::ostream& os, HeapTypeInfo info) {
...
+    Item(HeapType* heapType) : kind(HeapTypeKind), heapType(heapType) {}
+  };
+
+  // IDs of scanned Types and HeapTypes, used to prevent repeated scanning.
+  std::unordered_set<TypeID> scanned;
+
+  // The work list of Types and HeapTypes remaining to be scanned.
+  std::vector<Item> scanList;
+
+  // The list of Types and HeapTypes to visit (in forward preorder).
```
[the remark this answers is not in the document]
tlively: Yeah sorry I can clear that up. The list stored in forward preorder as this comment says, but then it is finally traversed from back to front, giving reverse postorder. I'll clarify that here and use "reverse postorder" everywhere I am currently just using "postorder."

src/wasm/wasm-type.cpp
```diff
@@ -974,6 +1004,276 @@ std::ostream& operator<<(std::ostream& os, HeapTypeInfo info) {
...
+    union {
+      Type* type;
+      HeapType* heapType;
+    };
+    Item(Type* type) : kind(TypeKind), type(type) {}
+    Item(HeapType* heapType) : kind(HeapTypeKind), heapType(heapType) {}
+  };
+
+  // IDs of scanned Types and HeapTypes, used to prevent repeated scanning.
+  std::unordered_set<TypeID> scanned;
```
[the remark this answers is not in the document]
tlively: Because it needs to refer to `HeapType`s as well. This gives a simple way to store both in the same collection where we don't need to be able to differentiate them because we know that no compound `Type` and `HeapType` will ever have the same ID. I used the same trick in `reaches`.

test/example/type-builder.cpp
tlively: Good catch, thanks! I'll replace the second one (`ref null extern`), but the first one `ref extern` is not the same as `externref` because it is non-nullable.

test/example/type-builder.cpp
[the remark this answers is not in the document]
kripken: I have a slight preference to `extern_`.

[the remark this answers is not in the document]
tlively: Sounds good. I'll make this change in a followup 👍

tlively:
I believe I have addressed all of the feedback so far 👍

kripken (approved):

kripken:
Nice work!

I have some thoughts for possible followups as mentioned earlier, but nothing urgent, and we can discuss those separately.

PR #3419, opened 2020-12-03
[NFC] Rename BasicID to BasicType

tlively:
This name is more descriptive and paves the way toward interning HeapTypes as
well, with an analogous BasicHeapType.


kripken (approved):
But what is a "basic heap type"? Aren't they all non-basic?

tlively:
No, basic heap types include `extern`, `func`, `any`, and (until we remove it) `exn`. Remeber that `externref` is just an abbreviation for `(ref null extern)`.

PR #3420, opened 2020-12-03
[Types] Refactor signature collection to heap type collection. NFC.

kripken:
This will allow writing GC types in the future, which are non-signature
heap types.

To allow this PR to work, it adds `operator<` for `HeapType` so that it
can be used in the data structures that collect uses.

Drive-by fix of a weird hack with sending a `Name*` in Print.

### DIFF
--- a/src/ir/module-utils.h
+++ b/src/ir/module-utils.h
@@ -404,7 +404,7 @@ inline void collectHeapTypes(Module& wasm,
                              std::vector<HeapType>& types,
                              std::unordered_map<HeapType, Index>& typeIndices) {
   struct Counts : public std::unordered_map<HeapType, size_t> {
-    bool isRelevant(Type type) { return !type.isBasic(); }
+    bool isRelevant(Type type) { return !type.isBasic() && type.isRef(); }
     void note(HeapType type) { (*this)[type]++; }
     void maybeNote(Type type) {
       if (isRelevant(type)) {
--- a/src/ir/module-utils.h
+++ b/src/ir/module-utils.h
@@ -400,14 +400,11 @@ template<typename T> struct CallGraphPropertyAnalysis {
 // indices. HeapTypes are sorted in order of decreasing frequency to minize the
 // size of their collective encoding. Both a vector mapping indices to
 // HeapTypes and a map mapping HeapTypes to indices are produced.
-inline void
-collectHeapTypes(Module& wasm,
-                 std::vector<HeapType>& types,
-                 std::unordered_map<HeapType, Index>& typeIndices) {
+inline void collectHeapTypes(Module& wasm,
+                             std::vector<HeapType>& types,
+                             std::unordered_map<HeapType, Index>& typeIndices) {
   struct Counts : public std::unordered_map<HeapType, size_t> {
-    bool isRelevant(Type type) {
-      return !type.isBasic() && type.isRef();
-    }
+    bool isRelevant(Type type) { return !type.isBasic() && type.isRef(); }
     void note(HeapType type) { (*this)[type]++; }
     void note(Signature sig) { note(HeapType(sig)); }
     void maybeNote(Type type) {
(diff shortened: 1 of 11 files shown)


src/ir/module-utils.h
```diff
@@ -392,29 +392,29 @@ template<typename T> struct CallGraphPropertyAnalysis {
...
-                  std::vector<Signature>& signatures,
-                  std::unordered_map<Signature, Index>& sigIndices) {
-  struct Counts : public std::unordered_map<Signature, size_t> {
-    void note(Signature sig) { (*this)[sig]++; }
+// HeapTypes and a map mapping HeapTypes to indices are produced.
+inline void collectHeapTypes(Module& wasm,
+                             std::vector<HeapType>& types,
+                             std::unordered_map<HeapType, Index>& typeIndices) {
+  struct Counts : public std::unordered_map<HeapType, size_t> {
+    bool isRelevant(Type type) { return !type.isBasic() && type.isRef(); }
```
tlively, line 407: What does the `&& type.isRef()` add?

src/ir/module-utils.h
```diff
@@ -392,29 +392,29 @@ template<typename T> struct CallGraphPropertyAnalysis {
...
+    void note(HeapType type) { (*this)[type]++; }
+    void note(Signature sig) { note(HeapType(sig)); }
```
tlively, line 409: IIRC, `Signature` intentionally has an implicit conversion to `HeapType`, so we shouldn't need a separate overload for it.

src/passes/Print.cpp
```diff
@@ -2375,6 +2390,13 @@ struct PrintSExpression : public OverriddenVisitor<PrintSExpression> {
     }
     o << ")";
   }
+  void handleHeapType(HeapType type) {
+    if (type.isSignature()) {
+      handleSignature(type.getSignature());
+    } else {
+      WASM_UNREACHABLE("bad heap type");
```
tlively, line 2397: Maybe "Unsupported heap type" would be more correct.

src/wasm-type.h
```diff
@@ -316,6 +317,7 @@ struct Struct {
   Struct(FieldList&& fields) : fields(std::move(fields)) {}
   bool operator==(const Struct& other) const { return fields == other.fields; }
   bool operator!=(const Struct& other) const { return !(*this == other); }
+  bool operator<(const Struct& other) const { return fields < other.fields; }
```
tlively, line 320: I wish we could use C++20's spaceship operator here instead of having to define both `operator==` and `operator<` :(

src/wasm/wasm-type.cpp
```diff
@@ -746,6 +755,26 @@ HeapType& HeapType::operator=(const HeapType& other) {
   return *this;
 }
 
+bool Field::operator<(const Field& other) const {
+  if (type < other.type) {
+    return true;
+  }
+  if (other.type < type) {
+    return false;
+  }
```
tlively, line 764:
```suggestion
  if (type != other.type) {
    return type < other.type;
  }
```

src/wasm/wasm-type.cpp
```diff
@@ -746,6 +755,26 @@ HeapType& HeapType::operator=(const HeapType& other) {
...
+  if (mutable_ < other.mutable_) {
+    return true;
+  }
+  if (other.mutable_ < mutable_) {
+    return false;
+  }
```
tlively, line 770: This one's slightly weirder but should still work:
```suggestion
  if (mutable_ != other.mutable_) {
    return mutable < other.mutable;
  }
```
It also might be worth comparing the type last to avoid having to compare types in as many cases as possible.

src/ir/module-utils.h
kripken: Hmm, isn't `funcref` both basic and reference? We need to ignore funcref here.

kripken:
Thanks for the feedback, applied.

src/ir/module-utils.h
kripken: Oh, this compiles, but actually it's wrong I see... it will infinitely recurse without the explicit cast :rofl:

kripken: Oh actually I misunderstood you, did you mean to remove the line entirely? Yes, that would both compile and be correct :smile:

src/ir/module-utils.h
tlively: Would't `!type.isBasic()` alone be enough to ignore `funcref`, though? Are you meaning to exclude RTT types here?

tlively (approved):

src/ir/module-utils.h
kripken: Oh, I see what you mean now. Yes, ignoring all basic ones should be enough.

kripken: Oh, no, that breaks all the tests, it lets something invalid through.

I don't think it's RTT though, since there are none?

Anyhow, l find it clearer as it is, plus it does seem good to disallow anything that is not a reference.

tlively: Oh, maybe tuples? Anyway this is fine.

PR #3421, opened 2020-12-03
[GC] Refactor Vacuum pass. NFC.

kripken:
This is shorter by using ChildIterator, so it's more general. The greater generality
should not be noticeable atm, but it will allow this pass to just work on GC
instructions once we have them (without this, in my testing the optimizer does not
do very well on GC).

### DIFF
--- a/src/passes/Vacuum.cpp
+++ b/src/passes/Vacuum.cpp
@@ -71,6 +71,10 @@ struct Vacuum : public WalkerPass<ExpressionStackWalker<Vacuum>> {
   Expression* optimize(Expression* curr, bool resultUsed, bool typeMatters) {
     FeatureSet features = getModule()->features;
     auto type = curr->type;
+    // If the type is none, then we can never replace it with another type.
+    if (type == Type::none) {
+      typeMatters = true;
+    }
     // An unreachable node must not be changed. DCE will remove those.
     if (type == Type::unreachable) {
       return curr;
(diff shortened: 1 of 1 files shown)


src/passes/Vacuum.cpp
```diff
@@ -49,155 +50,85 @@ struct Vacuum : public WalkerPass<ExpressionStackWalker<Vacuum>> {
...
+  //  (drop
+  //   (i32.eqz
+  //    (call ..)))
+  //
+  // The drop means that the value is not used later. And while the call has
+  // side effects, the i32.eqz does not. So when we are called on the i32.eqz,
+  // and told the result does not matter, we can return the call. Note that in
+  // this case the type does not matter either, as drop doesn't care, and anyhow
+  // i32.eqz returns the same type as it receives. But for an expression that
+  // returns a different type, if the type matter then we cannot replace it.
```
tlively, line 70:
```suggestion
  // returns a different type, if the type matters then we cannot replace it.
```

src/passes/Vacuum.cpp
```diff
@@ -49,155 +50,85 @@ struct Vacuum : public WalkerPass<ExpressionStackWalker<Vacuum>> {
...
+      typeMatters = true;
+    }
+    // An unreachable node must not be changed. DCE will remove those.
     if (type == Type::unreachable) {
       return curr;
     }
-    // We iterate on possible replacements. If a replacement changes the type,
-    // stop and go back.
+    // resultUsed only makes sense when the type is concrete
+    assert(!(resultUsed && curr->type == Type::none));
```
tlively, line 83: Nit: I think the comment makes the intention clear so that this is more readable.
```suggestion
    assert(!resultUsed || curr->type != Type::none));
```

src/passes/Vacuum.cpp
```diff
@@ -49,155 +50,85 @@ struct Vacuum : public WalkerPass<ExpressionStackWalker<Vacuum>> {
...
+    // If we actually need the result, then we must not change anything.
+    // TODO: clever things
```
tlively, line 85: Can you give some examples of potential clever things?

tlively (approved):
Holy cow, I can't believe all that code was just checking for children with side effects 😯

src/passes/Vacuum.cpp
kripken: That's funny, for me it's the opposite... :smile: but let's go with your way, to favor people reading the code.

src/passes/Vacuum.cpp
kripken: We could maybe find unused code in the middle somehow... maybe that's not practical though.

kripken:
As background, this is one of the first passes ever written. Back then we didn't have ChildIterator etc. and it seemed like all we had was unary/binary/select as things that we can do this optimization on, so it sort of made sense to hard-code it...

PR #3422, opened 2020-12-04
wasm-emscripten-finalize: Support PIC + threads

sbc100:
With PIC + threads the offset of passive segments is not constant but
relative to `__memory_base`.

When trying to find passive segment offset based on the target of the
`memory.init` instruction we need to consider this possilbity as well as
the regular constant one.

For the llvm side of this that generates the calles to memory.init
see:  https://reviews.llvm.org/D92620


sbc100:
Lacks testing ...

src/wasm/wasm-emscripten.cpp
```diff
@@ -170,9 +180,7 @@ std::string codeForConstAddr(Module& wasm,
                              int32_t address) {
   const char* str = stringAtAddr(wasm, segmentOffsets, address);
   if (!str) {
-    // If we can't find the segment corresponding with the address, then we
-    // omitted the segment and the address points to an empty string.
-    return escape("");
+    Fatal() << "unable to find data for ASM/EM_JS const at: " << address;
```
tlively, line 183: What's this change about?

tlively (approved):

src/wasm/wasm-emscripten.cpp
sbc100: I notices that if the segments offset are not calcualted the fallback of stringAtAddr is currently to return empty string which is really not useful.  

This would mean that emscripten gets passed the empty string for its EM_ASM constants..   I think its better to fail hard here.

I tests this with all the emscripten tests to make we don't hit this in any normal use cases.

sbc100: I'm somewhat worried by the comment I'm removing here though:  `// omitted the segment and the address points to an empty string.`

I don't see who a string could simultaneously be accessed for code and also `omitted`.    If the string was omitted it is a by definition not refernced.  If a string is referenced by omitted that would be bug and we want to fail as early as possible (IIUC).

tlively: I agree that failing hard and fast sounds much better here 👍

PR #3423, opened 2020-12-04
[GC] Support reading and writing of Struct and Array types

kripken:
This adds support in the text and binary format handling, which allows us
to have a full test of reading and writing the types.

This adds a "name" field to struct fields, which the text format supports.


kripken:
Testing found a small bug with printing of a function's `(result)`, which should also use the type name. Fixed.

src/passes/Print.cpp
```diff
@@ -119,59 +131,104 @@ std::ostream& operator<<(std::ostream& os, SigName sigName) {
...
+            continue;
+          } else if (heapType.isArray()) {
+            os << "[";
+            auto element = heapType.getArray().element;
+            if (element.mutable_) {
+              os << "mut:";
+            }
+            printType(element.type);
+            os << "]";
+            continue;
```
aheejin, line 156: Do we support packed types here? We seem to support it in wasm-type.cpp, but in type name printing here I mean.

src/passes/Print.cpp
```diff
@@ -119,59 +131,104 @@ std::ostream& operator<<(std::ostream& os, SigName sigName) {
...
+    os << "[";
+    auto element = type.getArray().element;
+    if (element.mutable_) {
+      os << "mut:";
+    }
+    printType(element.type);
+    os << "]";
+  } else {
+    os << type;
+  }
```
aheejin, line 198: I think contents of this function overlaps with the `printType` lambda function above... Can we unify them?

src/passes/Print.cpp
```diff
@@ -119,59 +131,104 @@ std::ostream& operator<<(std::ostream& os, SigName sigName) {
...
-    }
+    os << HeapTypeName(type.getHeapType());
     os << ')';
     return os;
   }
   return os << SExprType(typeName.type);
 }
 
-std::ostream& operator<<(std::ostream& os, HeapTypeName typeName) {
+std::ostream& operator<<(std::ostream& os, ResultTypeName typeName) {
```
aheejin, line 216: I'm confused about printing "***Type" vs. "***TypeName" here. This routine seems to be the same as [`ResultType`'s `operator<<`](https://github.com/WebAssembly/binaryen/blob/ef8c9f6d06b29fe2c7fe63236d709e3278a55041/src/wasm/wasm-type.cpp#L871-L873), so it doesn't need to be duplicated..? And in my understanding (correct me if I'm wrong), printing `***TypeName` is something like `${i32_f32}`, while printing `***Type` is something like `(struct (field i32) (field f32))`. And this routine seems to be doing the latter..?

aheejin (commented):
Nice that we are now printing GC types! Haven't read all of them yet, but some questions first to help my understanding:

test/heap-types.wast
```diff
@@ -0,0 +1,24 @@
+;; Test that we can roundtrip struct and array types
+(module
+  (type $struct.A (struct
+    i32
```
tlively, line 4: I'm surprised that `field` is optional. Does the explainer have an example like this?

test/typed-function-references.wast.fromBinary
```diff
@@ -93,7 +93,7 @@
    )
   )
  )
- (func $nested-type-only-there (result (ref null (func (result (ref null (func (param i32 i32 i32 i32 i32)))))))
+ (func $nested-type-only-there (result (ref null $none_=>_ref_null[i32_i32_i32_i32_i32_->_]))
```
tlively, line 96: What would you think of using `?` instead of `_null` to make the names shorter? (I forget if '?' is even an allowed character in names...)

tlively (commented):
Initial comments:

src/passes/Print.cpp
kripken: It may be missing here, yes. In a later PR I'll be adding it, when we add an instruction that can also use them (and has `_s/_u` variants).

src/passes/Print.cpp
```diff
@@ -119,62 +126,88 @@ std::ostream& operator<<(std::ostream& os, SigName sigName) {
...
+    os << "{";
+    auto sep = "";
+    for (auto& field : struct_.fields) {
+      os << sep;
+      sep = "_";
+      if (field.mutable_) {
+        os << "mut:";
+      }
+      printType(field.type);
+    }
```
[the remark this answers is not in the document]
kripken: I think that could be done probably, but it's not trivial. I'd also like to remove the `sep` duplication which is very common across this file. How about as a followup?

src/passes/Print.cpp
kripken: I think this is the same as @tlively 's comment above? Yes, there is duplication with some changes. I'd like to refactor them together in a followup, if that's ok.

src/passes/Print.cpp
kripken: This routine prints an inline type at the "top level", and names inside it. That is, it will print a tuple like this: `(result $foo $bar)`. The internal components of the tuple have type names, not the expanded type contents.

The difference is that `ResultType` doesn't use a `TypeName` internally for printing the sub-components. This is like a mixture of `ResultType` plus `TypeName`, so I called it `ResultTypeName`... :smile: 

This could all probably be simplified. I can take a look at the whole set of Printing helper classes and stuff later if that's ok. At least some of this complexity is because tuple types do not always have an `()` around them (here they don't, as in the example above, because it's in a result - this is not a problem in a param, oddly...) - maybe that could be simplified too.

src/wasm-type.h
```diff
@@ -301,7 +303,7 @@ struct Field {
 
   bool operator==(const Field& other) const {
     return type == other.type && packedType == other.packedType &&
-           mutable_ == other.mutable_;
+           mutable_ == other.mutable_ && name == other.name;
```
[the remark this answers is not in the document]
kripken: Thanks, fixing to ignore the name here.

test/heap-types.wast
kripken: The [very first example](https://github.com/WebAssembly/gc/blob/master/proposals/gc/Overview.md#type-grammar) does:
```wat
(type $tup (struct i64 i64 i32))
```
I agree in general that the large amount of sugar in the wasm text format can be kind of annoying when writing a parser... but I guess it helps when writing code which is more frequent.

test/heap-types.wast
```diff
@@ -0,0 +1,24 @@
...
+    (field f32)
+    (field $named f64)
+  ))
+  (type $struct.B (struct
+    (field (mut i64))
+    (field (ref $struct.A))
+    (field (mut (ref $struct.A)))
```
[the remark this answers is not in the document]
kripken: * Good point, named+mut was missing, will add now.
* (as mentioned earlier, I'd like to defer packed types for when I have an instruction that can process them, then we can fully test it.)

test/typed-function-references.wast.fromBinary
kripken: Oh, nice, I prefer `?` -  good idea.

I think it's valid. Anything but `()"'` should be fine IIRC.

kripken (commented):
Thanks for the feedback!

kripken:
I refactored the print logic to avoid duplication, and instead to have two mutually-recursive functions.

src/passes/Print.cpp
kripken: (should be fixed now)

src/passes/Print.cpp
```diff
@@ -141,37 +88,112 @@ struct TypeName {
...
+    return;
+  }
+  if (first) {
+    os << '$';
+  }
+  if (type.isSignature()) {
+    auto sig = type.getSignature();
+    printTypeName(os, sig.params);
+    if (first) {
+      os << "_=>_";
```
aheejin, line 136: In some place we use `_->_` and elsewhere we seem to use `_=>_`... Can we unify them?

src/passes/Print.cpp
```diff
@@ -141,37 +88,112 @@ struct TypeName {
...
+      os << sep;
+      sep = "_";
+      printTypeName(os, t);
+    }
+  } else if (type.isRef()) {
+    os << "ref";
+    if (type.isNullable()) {
+      os << "?";
+    }
+    os << "{";
```
aheejin, line 116: I think this used to be `[`? Any reason to change?

kripken: Oh, I just happened to notice that was ambiguous, since we use `"[ ]"` for Arrays too (and it seems to fit Arrays more).

src/passes/Print.cpp
kripken: The idea (before this PR) is to emit `=>` for the toplevel type, and `->` for inner ones. At least for me it makes things easier to read. Sometimes those types are big...

But I don't feel strongly. I can make this always emit one or the other if you prefer - let me know which one. That would make this PR change a lot of other existing tests, however, so it might be nicer as a separate followup.

src/passes/Print.cpp
kripken: I don't think we have a `ParamTypeName` in this file?

src/wasm/wasm-s-parser.cpp
```diff
@@ -2790,6 +2790,44 @@ HeapType SExpressionWasmBuilder::parseHeapType(Element& s) {
...
     return Signature(Type(params), Type(results));
   }
+  // It's a struct or an array.
+  auto parseField = [&](Element* t) {
+    bool mutable_ = false;
+    if (t->isStr()) {
+      // t is a simple string name like "i32"
+      return Field(elementToType(*t), mutable_);
+    }
+    // t is a tuple, containing either
```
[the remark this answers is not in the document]
kripken: Oh sorry, I mean to say "list" here... fixed.

src/wasm/wasm-type.cpp
```diff
@@ -186,6 +186,7 @@ size_t hash<wasm::Field>::operator()(const wasm::Field& field) const {
   auto digest = wasm::hash(field.type);
   wasm::rehash(digest, field.packedType);
   wasm::rehash(digest, field.mutable_);
+  wasm::rehash(digest, field.name);
```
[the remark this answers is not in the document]
kripken: Good catch, fixed.

src/passes/Print.cpp
aheejin: But don't we use `{}` for structs instead?

src/passes/Print.cpp
aheejin: I think unifying them wouldn't be too confusing, but I don't feel strongly either. Yeah whether we do it or not it should be a followup I think as you suggested.

src/passes/Print.cpp
kripken: Oh right... how silly of me :rofl: 

Then maybe `< >`..? Or something else?

aheejin: It might be harder to read if they are mixed with `->`s and `=>`s, but other than `()` we don't have any option left, so I'm OK either way..?

kripken: Yes, the arrows make it hard...

We can do like bash, `ref...fer` :wink:

Or maybe `ref|...|`?

Anyhow, I guess this is not actually ambiguous (since the `ref` prefix disambiguates) so it's not urgent to figure this out.

src/passes/Print.cpp
aheejin: Oh what I was wondering is, if we need `ResultTypeName`, why don't we also need `ParamTypeName`? Maybe I'm missing something.

src/passes/Print.cpp
kripken: Let's go with `ref|...|` for now, this greatly improves readability over something more ambiguous, e.g.
```wat
-  (local $2 (ref null ${mut:i64_ref?{{i32_f32_f64}}_mut:ref?{{i32_f32_f64}}}))
+  (local $2 (ref null ${mut:i64_ref?|{i32_f32_f64}|_mut:ref?|{i32_f32_f64}|}))
```
(note how the `{{ }}` was ambiguous before)

src/passes/Print.cpp
kripken: Oh, sorry for misunderstanding you before. It's because the wat text format is weird here. Both the params and the results are a tuple. But the params are written as `(param X) (param Y)` while the results are written as `(result X Y)`. (Maybe the difference is because params have names, so they each need their own element?) We can just loop over params and print them, but we can't do the same for results - we need to write them as a tuple but without the outer `()`, basically (that is, it would be wrong to print `(result (X Y))`, it should be `(result X Y)`). So we wrap the results in a special way to handle that.

There might be a better approach, I'm not sure offhand of the benefits vs alternatives. I think @tlively wrote that code originally and maybe remembers?

aheejin: Oh I see... Yeah maybe we can improve that, but I don't think it needs to block this from landing anyway.

aheejin (approved):

PR #3424, opened 2020-12-04
Don't apply SafeHeap to wasm start function

sbc100:
In relocable code (MAIN/SIDE modules) we use the start function to run
`__wasm_init_memory` which loads the data segments into place.   We
can't call get_sbkr pointer during that function because the sbrk
pointer itself lives in static data segment.


src/passes/SafeHeap.cpp
```diff
@@ -67,6 +67,10 @@ struct AccessInstrumenter : public WalkerPass<PostWalker<AccessInstrumenter>> {
   // If the getSbrkPtr function is implemented in the wasm, we must not
   // instrument that, as it would lead to infinite recursion of it calling
   // SAFE_HEAP_LOAD that calls it and so forth.
+  // As well as the getSbrkPtr function we also avoid insturmenting the
```
kripken, line 70:
```suggestion
  // As well as the getSbrkPtr function we also avoid instrumenting the
```

kripken (approved):

tlively (approved):

PR #3426, opened 2020-12-04
Improve lit support

sbc100:
This uses the same technique used in llvm-lit to enable
running on in-tree tests with out-of-tree builds.

So you can run something like this:

  ../binaryen-out/bin/binaryen-lit test/lit/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,6 +94,9 @@ jobs:
         name: build-${{ matrix.os }}
         path: out/install
 
+    - name: test binaryen-lit
+      run: python out/bin/binaryen-lit test/lit/parse-error.wast
+
     - name: test
       run: python check.py --binaryen-bin=out/bin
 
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -394,3 +394,5 @@ if(EMSCRIPTEN)
   set_property(TARGET binaryen_js PROPERTY CXX_STANDARD_REQUIRED ON)
   install(TARGETS binaryen_js DESTINATION ${CMAKE_INSTALL_BINDIR})
 endif()
+
+configure_file(scripts/binaryen-lit.in ${CMAKE_INSTALL_BINDIR}/binaryen-lit @ONLY)
```
tlively, line 398: What happens if you haven't set an install dir? Would it make more sense to generate this in the normal build bin dir instead?

.github/workflows/ci.yml
```diff
@@ -115,6 +115,8 @@ jobs:
         cmake -S . -B out -G Ninja -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
     - name: build
       run: cmake --build out
+    - name: test lit
+      run: python out/bin/binaryen-lit test/lit/parse-error.wast
```
tlively, line 119: Is this just testing that the lit wrapper works at all? Is this meant to be in the final PR?

sbc100: Yeah I thought it might be worth testing directly.. just just via check.py.

tlively (approved):

PR #3428, opened 2020-12-05
Intern HeapTypes and clean up types code

tlively:
Interns HeapTypes using the same patterns and utilities already used to intern
Types. This allows HeapTypes to efficiently be compared for equality and hashed,
which may be important for very large struct types in the future. This change
also has the benefit of increasing symmetry between the APIs of Type and
HeapType, which will make the developer experience more consistent. Finally,
this change will make TypeBuilder (#3418) much simpler because it will no longer
have to introduce TypeInfo variants to refer to HeapTypes indirectly.


src/wasm-type.h
```diff
@@ -251,17 +261,82 @@ struct ResultType {
   std::string toString() const;
 };
 
+class HeapType {
```
kripken, line 264: I think it would be good to add a comment explaining what a HeapType is, and why it's separate from Type.

src/wasm-type.h
```diff
@@ -251,17 +261,82 @@ struct ResultType {
...
+  constexpr bool isBasic() const { return id <= _last_basic_type; }
+  constexpr bool isCompound() const { return id > _last_basic_type; }
+  bool isFunction() const;
+  bool isSignature() const;
+  bool isStruct() const;
+  bool isArray() const;
+
+  Signature getSignature() const;
+  const Struct& getStruct() const;
+  Array getArray() const;
```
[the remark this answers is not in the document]
kripken: Is the reason Arrays and Structs differ in this way that Arrays are "small" and Structs are possibly of unbounded size?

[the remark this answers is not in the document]
tlively: Yes, that's what I was thinking 👍 Arrays contain only one `Field` each, which in turn contains only three items taking up two machine words. That seems small enough to pass around by value.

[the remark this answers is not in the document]
kripken: That does make sense, but it does make the API asymmetrical, which seems a little surprising. Or do you think for users of the API the difference wouldn't be noticeable?

A comment might be good either way.

src/wasm/wasm-type.cpp
```diff
@@ -45,300 +48,284 @@ struct TypeInfo {
...
-    {TypeInfo(HeapType(HeapType::I31Kind), false), Type::i31ref},
-  };
+  // Track unique_ptrs for constructed types to avoid leaks.
+  std::vector<std::unique_ptr<Info>> constructedTypes;
 
-  uintptr_t canonicalize(const TypeInfo& info) {
+  // Maps from constructed types to their canonical Type IDs.
+  std::unordered_map<Info, uintptr_t> typeIDs;
+
+  typename Info::type_t canonicalize(Info info) {
```
aheejin, line 256:
```suggestion
  typename Info::type_t canonicalize(const Info& info) {
```
Any reason for the change?

src/wasm/wasm-type.cpp
```diff
@@ -45,300 +48,284 @@ struct TypeInfo {
...
+          case HeapType::i31:
+            break;
         }
-        break;
-      case HeapType::ArrayKind:
-        assert(heapType.array.element.type.isSingle());
-        break;
+      } else {
+        if (info.ref.heapType == HeapType::i31) {
+          return Type::i31ref;
```
aheejin, line 299: Any reason to change this from map lookup to switch-case jump table? Wouldn't map lookup be cheaper?

src/wasm/wasm-type.cpp
```diff
@@ -45,300 +48,284 @@ struct TypeInfo {
...
+          case HeapType::eq:
+            return Type::eqref;
+          case HeapType::i31:
+            break;
         }
-        break;
-      case HeapType::ArrayKind:
-        assert(heapType.array.element.type.isSingle());
-        break;
+      } else {
```
aheejin, line 297: How about adding an assertion failure in case `heapType` is not nullable and is not `i31`, because we don't support it yet?

src/wasm/wasm-type.cpp
```diff
@@ -983,4 +945,115 @@ std::ostream& operator<<(std::ostream& os, TypeInfo info) {
...
+      return os << info.struct_;
+    case HeapTypeInfo::ArrayKind:
+      return os << info.array;
+  }
+  WASM_UNREACHABLE("unexpected kind");
+}
+
 } // namespace wasm
+
+namespace std {
```
aheejin, line 962: Any reason we put `hash` functions in two separate chunks? Can we merge them into one chunk?

aheejin (commented):
It's much clearer this way!

src/wasm-type.h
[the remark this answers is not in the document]
tlively: Yeah, the asymmetry is a little unfortunate, but I think it will be ok. What would you think of deleting `operator=` to prevent folks from accidentally doing `auto s = getStruct()` and making a copy of the struct? Would that be worth it or too inconvenient?

Will add comments.

src/wasm/wasm-type.cpp
tlively: There are two changes here:

1. I switched from having `canonicalize` returning an integer to having it return either `Type` or `HeapType` directly, since that's the only way it is ever used.

2. I switched from taking the `Info` (previously always a `TypeInfo`, now also possibly a `HeapTypeInfo`) by value to taking it be reference because `TypeInfo` can contain a `Tuple` and `HeapTypeInfo` can contain a `Struct`, both of which can have non-trivial copies.

tlively:
> It's much clearer this way!

Thanks, I'm glad you think so :D

src/wasm-type.h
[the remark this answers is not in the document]
kripken: That seems good, assuming I understand correctly and it would just force people to add the `&` in `auto& struct_ = type.getStruct();` - ? If so sgtm. But regardless, that could also be separate.

kripken (approved):
lgtm % comments we discussed

src/wasm/wasm-type.cpp
aheejin: Sorry, but not sure how they are related to the `const Info&` thing..?

src/wasm/wasm-type.cpp
tlively: There are a couple unfortunate things going on here.

 1. The declaration of specializations of `std::hash` for `TypeInfo` and `HeapTypeInfo` needs to be done before the declaration of `Store`, otherwise the declaration of `Store::typeIDs` tries to use the default instantiation and fails.

 2. There's no way to declare specializations from a different namespace, so there's no way to avoid leaving the anonymous and `wasm` namespaces and entering `std` just for these two declarations.

There's nothing stopping us from moving the hash definitions from down below up here, but I thought it would still be better to keep them separate despite the extra verbosity.

src/wasm/wasm-type.cpp
tlively: I'm not sure about the performance - the map lookup at least requires taking a lock and calculating a hash. I wouldn't expect this to have a measurable impact (but I haven't measured it). The reason I changed it was that what we had before was kind of a hack. It used tuple `TypeInfo`s to store individual `BasicType`s even though empty and singleton tuples were otherwise not allowed anywhere. In contrast, there is no need for `HeapTypeInfo` to be able to represent a `BasicHeapType` and it seemed cleaner to make that separation of concerns explicit for `TypeInfo` as well.

src/wasm/wasm-type.cpp
tlively: Ah, sorry, I had confused myself 😓  Thanks for this catch; there was no reason for me to have stopped using const reference here.

src/wasm/wasm-type.cpp
aheejin: Hmm, I thought it would be simpler if every `Type` has its matching `TypeInfo` and we can get one using the other, but as long as it is implementation detail and well encapsulated from outside I think either way should be fine.

aheejin (approved):
LGTM % comments too

src/wasm/wasm-type.cpp
tlively: I agree that erroring out when we use non-nullable types would make sense (including  for `i31ref`, which we don't support any better than other non-nullable types). But I think the validator would be a better place for that error since we support creating and querying non-nullable types, just not using them in the IR.

aheejin: What I was wondering is, even if the type is basic, if it is not nullable, it will fall out of all these `if`s and call `Store<TypeInfo>::canonicalize(info)`, which will create `TypeInfo` for it, in which case it can cause an unintended consequence which is not easy to catch later. Maybe I'm missing something?

tlively: I don't think the act of making a `TypeInfo` for a non-nullable type can make anything bad happen. We also depend on being able to do that in test/example/typinfo.cpp.

The bad things only start happening once you use a non-nullable type for locals in the IR.

aheejin: But didn't you say you don't create `TypeInfo`s anymore with (nullable) basic types?

tlively: Oh, I see what you're saying. We only provide `BasicType`s for types explicitly given their own names in the spec proposals, so `(ref null extern)`, which is equivalent to `externref`, is canonicalized into `Type::externref`, but on the other hand `(ref extern)` would be backed by a `TypeInfo`. This is working as intended. `i31ref` is weird because it is non-nullable by definition.

src/wasm/wasm-binary.cpp
```diff
@@ -5630,7 +5630,8 @@ bool WasmBinaryBuilder::maybeVisitStructGet(Expression*& out, uint32_t code) {
     default:
       return false;
   }
-  auto type = getHeapType();
+  // This type annotation is unused. Beware it needing to be used in the future!
+  getHeapType();
```
tlively, line 5634: cc @kripken here's where I fixed that unused variable error I was getting.

src/wasm/wasm-type.cpp
tlively: Thanks :) This change should not be observable outside of this function.

PR #3429, opened 2020-12-06
[GC] Add struct.get instruction parsing and execution

kripken:
This is the first instruction that uses a GC Struct or Array, so it's where
we start to actually need support in the interpreter for those values, which
is added here.

GC data is modeled as a `gcData` field on a Literal, which is just a
`Literals`. That is, both a struct and an array are represented as an
array of values. The type which is alongside would indicate if it's a
struct or an array. Note that the data is referred to using a `shared_ptr`
so it should "just work", but we'll only be able to really test that once we
add `struct.new` and so can verify that references are by reference and
not value, etc.

As the first instruction to care about i8/16 types (which are only possible
in a Struct or Array) this adds support for parsing and emitting them.

This PR includes fuzz fixes for some minor things the fuzzer found, including
some bad printing of not having `ResultTypeName` in necessary places
(found by the text format roundtripping fuzzer).


kripken:
Fuzzer noticed I forgot to define the new instruction in `costs.h`.

src/wasm/wasm-stack.cpp
```diff
@@ -1910,7 +1910,19 @@ void BinaryInstWriter::visitStructNew(StructNew* curr) {
...
+  const auto& field = heapType.getStruct().fields[curr->index];
+  int8_t op;
+  if (field.type != Type::i32 || field.packedType == Field::not_packed) {
+    op = BinaryConsts::StructGet;
+  } else if (curr->signed_) {
+    op = BinaryConsts::StructGetS;
+  } else {
+    op = BinaryConsts::StructGetU;
+  }
+  o << int8_t(BinaryConsts::GCPrefix) << int8_t(op);
```
tlively, line 1923: The `op` should be a uleb128 here. (Not that it makes a difference in practice, but it would be good to be consistent with the spec for readability purposes.)

src/wasm/wasm-validator.cpp
```diff
@@ -2231,7 +2231,13 @@ void FunctionValidator::visitStructGet(StructGet* curr) {
...
                curr,
                "struct.get requires gc to be enabled");
-  WASM_UNREACHABLE("TODO (gc): struct.get");
+  if (curr->value->type != Type::unreachable) {
+    shouldBeEqual(
+      curr->type,
+      curr->value->type.getHeapType().getStruct().fields[curr->index].type,
+      curr,
+      "struct.get must have the proper type");
+  }
```
tlively, line 2240: It looks like this will be out of bounds if the index is incorrect, so it would be good to check the index here first.

test/passes/inlining_all-features.wast
```diff
@@ -64,7 +64,7 @@
  (export "func_36_invoker" (func $1))
  (func $0
   (return_call_ref
-   (ref.null (func))
+   (ref.null $none_=>_none)
```
tlively, line 67: Would it make sense to add this as a new case rather than replacing the slightly different `ref.nul (func)` case?

src/wasm-binary.h
```diff
@@ -338,6 +338,8 @@ enum EncodedType {
   f32 = -0x3,  // 0x7d
   f64 = -0x4,  // 0x7c
   v128 = -0x5, // 0x7b
+  i8 = -0x6,   // 0x7a
+  i16 = 0x7,   // 0x79
```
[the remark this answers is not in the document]
kripken: Heh, good catch!

src/wasm-interpreter.h
```diff
@@ -1402,7 +1402,15 @@ class ExpressionRunner : public OverriddenVisitor<SubType, Flow> {
...
-    WASM_UNREACHABLE("TODO (gc): struct.get");
+    Flow flow = this->visit(curr->value);
+    if (flow.breaking()) {
+      return flow;
+    }
+    auto data = flow.getSingleValue().getGCData();
+    if (!data) {
+      trap("null ref");
+    }
+    return (*data)[curr->index];
```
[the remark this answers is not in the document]
kripken: It should, yes. Looks like it's not though - will fix in the validator!

src/wasm/wasm-binary.cpp
```diff
@@ -1030,7 +1030,7 @@ void WasmBinaryWriter::writeType(Type type) {
 
 void WasmBinaryWriter::writeHeapType(HeapType type) {
   if (type.isSignature() || type.isStruct() || type.isArray()) {
-    o << S32LEB(getTypeIndex(type));
+    o << S64LEB(getTypeIndex(type)); // TODO: Actually s33
```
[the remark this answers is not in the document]
kripken: It is actually an S33. That doesn't fit in an S32... while we'd need quite a lot of types to hit this limitation, it seems worth fixing so we don't forget later.

src/wasm/wasm-binary.cpp
```diff
@@ -5601,20 +5618,21 @@ bool WasmBinaryBuilder::maybeVisitStructGet(Expression*& out, uint32_t code) {
   switch (code) {
     case BinaryConsts::StructGet:
       curr = allocator.alloc<StructGet>();
```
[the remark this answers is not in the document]
kripken: That's done in `wasm.h`.

src/wasm/wasm-stack.cpp
kripken: It looks like all the other locations use `int8_t`, though. But I agree this might be a good cleanup for the whole file - later?

test/passes/inlining_all-features.wast
kripken: At least based on [the issue you filed](https://github.com/WebAssembly/function-references/issues/42) it seemed like it wasn't intended to work? This PR removes the old form and just supports the form we all agree should work.

src/wasm/wasm-stack.cpp
tlively: Specifically only _prefixed_ opcodes should be LEBs, but I see that you're right that we are inconsistent here. All the SIMD opcodes use LEBs but e.g. none of the atomics instructions use LEBs. Cleaning them all up in a separate PR sgtm.

test/passes/inlining_all-features.wast
tlively: Ok, sounds good to me in the context of GC. For the basic reference types feature, we will still need to support `(ref.null func)` and `(ref.null extern)`, but that can be rectified in a separate change. https://webassembly.github.io/reference-types/core/text/instructions.html#reference-instructions

tlively (approved):

src/literal.h
```diff
@@ -43,6 +43,11 @@ class Literal {
     Name func;
     // exnref package. `nullptr` indicates a `null` value.
     std::unique_ptr<ExceptionPackage> exn;
+    // A reference to GC data, either a Struct or an Array. For  both of those
```
aheejin, line 46:
```suggestion
    // A reference to GC data, either a Struct or an Array. For both of those
```

test/passes/inlining_all-features.wast
kripken: We do still support `(ref.null func)` - this just removes the "expanded inline" variant of `(ref.null (func))` (i.e. it disallows the extra parens around the `func`). That is tested in `test/reference-types.wast`.

src/literal.h
```diff
@@ -43,6 +43,11 @@ class Literal {
...
+    // we store the referred data as a Literals object (which is natural for an
+    // Array, and for a Struct, is just the fields in order). The type is used
+    // to indicate whether this is a Struct or an Array, and of what type.
+    std::shared_ptr<Literals> gcData;
```
aheejin, line 50: Any reason it is not a `unique_ptr` but `shared_ptr`?

src/tools/fuzzing.h
```diff
@@ -327,28 +327,31 @@ class TranslateToFuzzReader {
...
-        if (wasm.features.hasReferenceTypes()) {
-          options.push_back(Type::funcref);
-          options.push_back(Type::externref);
-          if (wasm.features.hasExceptionHandling()) {
-            options.push_back(Type::exnref);
+    // TODO: interesting uses of typed function types
+    // TODO: interesting subtypes of compound types
+    if (type.isBasic()) {
+      switch (type.getBasic()) {
+        case Type::anyref:
```
aheejin, line 334: Not related to this PR, but just a question: we don't have `dataref` yet right?

src/wasm/wasm-binary.cpp
```diff
@@ -1408,7 +1419,19 @@ HeapType WasmBinaryBuilder::getHeapType() {
...
+  // a field.
+  auto initial = getS32LEB();
+  if (initial == BinaryConsts::EncodedType::i8) {
+    auto mutable_ = getU32LEB();
+    return Field(Field::i8, mutable_);
+  }
+  if (initial == BinaryConsts::EncodedType::i16) {
+    auto mutable_ = getU32LEB();
+    return Field(Field::i16, mutable_);
+  }
```
aheejin, line 1432: Any reason for special-casing these? Can't we include `i8` and `i16` in `getType` too?

src/literal.h
kripken: Thanks, I'll fix this in the followup.

src/literal.h
kripken: This is intentional - it's supposed to properly model a reference to GC data on the heap. That is, two Literals can be references to the same data. If the data is modified using one of them, reading from the other reference will show the change, etc.

For function references we have the function name, which works out in the same way.

src/tools/fuzzing.h
kripken: I don't think so - I don't even know what that is. Which proposal is it from?

src/wasm/wasm-binary.cpp
kripken: `i8` and `i16` are only valid in a `Field`. We have to actually call a different constructor of `Field` when using one of them, as opposed to a normal wasm Type. And similarly `Type` can't contain them.

In theory though this could have all been different, I think. Type could have contained `i8, i16` and the validator would check those types are only used where allowed. I'm not sure offhand if that would be better though overall.

src/tools/fuzzing.h
aheejin: It seems to be a reference for a heap type `data`, which is "the common supertype of all compound data types, like struct and array types and possibly host-defined types, for which casts are allowed". It is in the [MVP doc](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md).

src/wasm/wasm-binary.cpp
aheejin: Ah right, I missed the type of the argument to `Field` constructor is different here.

src/tools/fuzzing.h
kripken: I see, thanks. I missed that in the proposal. So far I haven't seen a use or need for it in code samples.

PR #3431, opened 2020-12-08
Fixed wasm-emscripten-finalize AsmConstWalker not handling 64-bit ptrs

aardappel:


tlively (commented):
Looks good, but it would be good to have a test. Sam's been putting wasm-emscripten-finalize lit tests in test/lit/wasm-emscripten-finalize and there are more non-lit tests in test/lld.

aardappel:
@tlively ok, there's now a test, at the cost of some collateral damage to the test runners.. maybe this can be done better @sbc100 ?

sbc100:
Seems reasonable to me.  Sad to add that extra complexity to the running but I don't see how to avoid it.

scripts/test/lld.py
```diff
@@ -49,10 +49,24 @@ def run_test(input_path):
...
+        lowered_input_path = input_path
+        cmd = shared.WASM_EMSCRIPTEN_FINALIZE + args
+        if '64' in input_path:
+            # We're processing a wasm64 module, which normally gets lowered
+            # by emcc -S MEMORY64=2, so here we have to do that manually.
+            lowered_input_path = input_path + '.wasm32'
+            support.run_command(shared.WASM_OPT +
+                                ['--enable-memory64', '--memory64-lowering',
+                                 '-S', '-o', lowered_input_path, input_path])
+            cmd += ['--enable-memory64', '--bigint']
```
sbc100, line 61: But wait.. in emscripten we run the lowering pass *after* finalize.    So finalize should be designed to run on the non-lowered version right? 

wasm-emscripten-finalize is designed to take the direct output of lld as its input.   If you need the lowering to happen first it should be done in wasm-emscripten-finalize itself I think.

aardappel: It may not be needed for the current tests. I added it because it appeared that some code would not be able to deal with the full wasm64, and instead would need the lowered version.

I can remove it for now.

tlively:
It looks like merging master will fix the test failures.

tlively (approved):

aardappel:
Note: had to revert this diff:

```
--- /home/runner/work/binaryen/binaryen/test/lld/duplicate_imports.wat.out
+++ actual
@@ -4,8 +4,8 @@
  (type $none_=>_none (func))
+ (type $i64_=>_i32 (func (param i64) (result i32)))
  (type $i32_i32_=>_i32 (func (param i32 i32) (result i32)))
- (type $i64_=>_i32 (func (param i64) (result i32)))
  (type $f32_f64_=>_f32 (func (param f32 f64) (result f32)))
  (type $f64_f64_=>_f32 (func (param f64 f64) (result f32)))
  (import "env" "puts" (func $puts1 (param i32) (result i32)))
```

I'm guessing this is a case of an unstable sort being used for these types? The change was produced on Windows/MSVC.

tlively:
@aardappel, that's the problem that I thought merging master would fix. Did that not fix it? So it's actually exhibiting different behavior on Windows?

aardappel:
Yes that did not fix it.

PR #3432, opened 2020-12-08
[GC] Add basic RTT support

kripken:
This adds `rtt.canon` and `rtt.sub` together with RTT type support
that is necessary for them. Together this lets us test roundtripping the
instructions and types.

Also fixes a missing traversal over globals in `collectHeapTypes`,
which the example from the GC docs requires, as the RTTs are in
globals there.

This does not yet add interpreter support and other things.


src/wasm/wasm-s-parser.cpp
```diff
@@ -940,6 +940,12 @@ Type SExpressionWasmBuilder::elementToType(Element& s) {
     }
     return Type(parseHeapType(*s[i]), nullable);
   }
+  if (elementStartsWith(s, RTT)) {
+    // It's an RTT, something like (rtt N $typename)
+    auto depth = atoi(s[1]->str().c_str());
```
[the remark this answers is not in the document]
kripken: Thanks, I missed that. But I'm not sure what to do with it. How would we compute it? I think we could only leave it as "unspecified" or something like that. But I wonder if this is not a bug in the spec maybe?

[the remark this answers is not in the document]
kripken: Oh wait, yeah, it looks like the spec even has a separate opcode for this case. Fixing...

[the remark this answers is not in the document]
kripken: Looks like that constant overlaps with `exnref`. Was that removed? We still use it in binaryen it seems.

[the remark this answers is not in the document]
kripken: I made it use a different constant to avoid the exnref overlap. Aside from that, it should now support rtts without a depth.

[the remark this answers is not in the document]
aheejin: - I haven't removed `exnref` yet, but we are not testing that in VM in Emscripten tests anymore, so it'd be better to change `exnref`'s opcode I think? Maybe add a `FIXME` to `exnref` that this is incorrect but it will be removed so it should be fine.
- Should we leave the depth unspecified in case it is not specified? Can't we do +1 to its parent's depth? I think the binary format mandates it, so every `Rtt` seems to need one; only the text format has a leeway to omit it. In the same light, I think we should not omit it when printing in the text format. Am I missing something..?

src/wasm/wasm-binary.cpp
```diff
@@ -5575,21 +5587,20 @@ bool WasmBinaryBuilder::maybeVisitRttCanon(Expression*& out, uint32_t code) {
...
 bool WasmBinaryBuilder::maybeVisitRttSub(Expression*& out, uint32_t code) {
   if (code != BinaryConsts::RttSub) {
     return false;
   }
-  auto* curr = allocator.alloc<RttSub>();
-  WASM_UNREACHABLE("TODO (gc): rtt.sub");
-  curr->finalize();
-  out = curr;
+  // FIXME: the binary format may also have an extra heap type and index that
+  //        are not needed
```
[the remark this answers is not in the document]
aheejin: I didn't know about the binary format doc, thanks for pointing that out. Yeah, I found that MVP doc already uses some inconsistent notations within itself, and sometimes it does not match with the Overview doc (e.g. `rtt.canon` is `rtt.get` in the Overview doc..), and there are also discrepancies with the binary format doc...

src/wasm/wasm-s-parser.cpp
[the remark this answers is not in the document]
aheejin: Sorry oh wait, the binary format has a separate opcode for `rtt`s without a depth as you said...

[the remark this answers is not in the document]
kripken: Sounds good, I made rtt use the right code, and made exnref use an incorrect one with a TODO to remove it. I removed one wasm binary test file that used the old exnref code.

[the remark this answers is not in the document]
aheejin: Sorry, nevermind what I said about +1. We can't trace all rtts up to their roots to compute it in parsing, so we should be able to parse and write rtts without depths after all. I wonder why that doesn't exist in the binary format doc though..? It's confusing which part of GC is standardized (?) and which is not...

src/wasm/literal.cpp
```diff
@@ -60,6 +60,9 @@ Literal::Literal(const Literal& other) : type(other.type) {
     new (&gcData) std::shared_ptr<Literals>(other.gcData);
   } else if (type.isFunction()) {
     func = other.func;
+  } else if (type.isRtt()) {
+    // Nothing to do: Rtts help JITs optimize, but are not used in the
+    // interpreter yet, and they are opaque to the wasm itself.
```
aheejin, line 65: Doesn't this PR add the interpreter support too?

aheejin (approved):

src/wasm-binary.h
```diff
@@ -354,8 +354,12 @@ enum EncodedType {
   nonnullable = -0x15, // 0x6b
   // integer reference type
   i31ref = -0x16, // 0x6a
-  // exception reference type
-  exnref = -0x18, // 0x68
+  // run-time type info type, with depth index n
+  rtt_n = -0x17, // 0x69
+  // run-time type info type, without depth index n
+  rtt = -0x18, // 0x68 FIXME
```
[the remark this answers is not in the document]
kripken: Oh, I think I just forgot that in there. I'll remove it.

src/wasm/literal.cpp
kripken: I didn't plan to, but the fuzzer did actually run into this, so I added it. But there is a lot more work necessary I think, for later.

src/wasm-binary.h
kripken, line 360:
```suggestion
  rtt = -0x18, // 0x68
```

aheejin:
The MVP doc says
> Validation requires that each RTT's parent type is a representative of a static supertype; runtime subtyping hence is a sub-relation of static subtyping (a graph with fewer nodes and edges).

Should we add this to the validation? But to add this, I guess we need a way to figure out subtyping relationship between heap types, such as `(struct i32 i64)` is a subtype of `(struct i32)`, or something. It seems to be defined here: https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#defined-types

Do you think this is the right way to proceed?

By the way it also [says](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#value-types-1)
> Note: `rtt n? $t1` is not a subtype of `rtt n? $t2`, if `$t1` is merely a subtype of `$t2`; such covariant subtyping would be unsound, since RTTs are used in both co- and contravariant roles (e.g., both when constructing and consuming a reference

kripken:
@aheejin Good questions about RTTs and subtyping - I'm not sure. In general I haven't taken the time to fully understand RTTs yet as there is a lot of other work that is definitely blocking the main current use cases, and I'm not sure if RTT subtyping is.

Meanwhile, I find RTTs pretty confusing... the examples in the docs are fairly short, and there don't seem to be spec tests written yet which could have helped me at least to understand what they actually do.

Perhaps @tlively knows more?

src/wasm-interpreter.h
```diff
@@ -1388,13 +1388,13 @@ class ExpressionRunner : public OverriddenVisitor<SubType, Flow> {
...
-  }
+  Flow visitRttCanon(RttCanon* curr) { return Literal(curr->type); }
   Flow visitRttSub(RttSub* curr) {
-    NOTE_ENTER("RttSub");
-    WASM_UNREACHABLE("TODO (gc): rtt.sub");
+    Flow parent = this->visit(curr->parent);
+    if (parent.breaking()) {
+      return parent;
+    }
+    return Literal(curr->type);
```
tlively, line 1397: From the explainer:

```
In order to cast down the type of a struct or array, the aggregate itself must be equipped with a suitable RTT. Attaching runtime type information to aggregates happens at allocation time but is optional. If no RTT is attached then their runtime type is treated as if it was anyref and a down cast to a more specific type will fail.
```

IIUC, this means that `Struct` and `Array` need to own optional Rtts and that Rtts actually need to refer to two heap types: the supertype and subtype of the subtyping relation they witness. The implementation of casts will need to trap if the provided Rtt does not witness the correct subtying relation.

Although this seems to work differently than in the V8 prototype, which uses a depth instead???

src/wasm-type.h
```diff
@@ -430,14 +432,20 @@ struct Array {
 };
 
 struct Rtt {
+  enum {
+    // An Rtt can have no depth specified
+    NoDepth = -1
+  };
```
tlively, line 438: how about `static constexpr uint32_t NoDepth = -1`?

src/wasm/wasm-type.cpp
```diff
@@ -463,6 +463,8 @@ FeatureSet Type::getFeatures() const {
       if (heapType.isStruct() || heapType.isArray()) {
         return FeatureSet::ReferenceTypes | FeatureSet::GC;
       }
+    } else if (t.isRtt()) {
+      return FeatureSet::ReferenceTypes | FeatureSet::GC;
```
tlively, line 467: Why is GC alone not enough here? AFAIK, all instructions that interact with Rtts in any way are in the GC proposal.

src/wasm/wasm-type.cpp
```diff
@@ -529,6 +531,15 @@ HeapType Type::getHeapType() const {
     }
     WASM_UNREACHABLE("Unexpected type");
   }
+  if (isRtt()) {
+    return getRtt().heapType;
+  }
+  WASM_UNREACHABLE("unexpected type");
```
tlively, line 537: This is already covered above.

src/wasm/wasm-type.cpp
```diff
@@ -529,6 +531,15 @@ HeapType Type::getHeapType() const {
...
+}
+
+const Rtt& Type::getRtt() const {
```
tlively, line 540: Our (brand new) convention is to pass `Rtt`s by value rather than reference (but I'm happy to reconsider that).

src/wasm/wasm-validator.cpp
```diff
@@ -2211,13 +2211,27 @@ void FunctionValidator::visitBrOnCast(BrOnCast* curr) {
 void FunctionValidator::visitRttCanon(RttCanon* curr) {
   shouldBeTrue(
     getModule()->features.hasGC(), curr, "rtt.canon requires gc to be enabled");
-  WASM_UNREACHABLE("TODO (gc): rtt.canon");
+  shouldBeTrue(curr->type.isRtt(), curr, "rtt.canon must have RTT type");
+  auto rtt = curr->type.getRtt();
+  shouldBeEqual(rtt.depth, Index(0), curr, "rtt.canon has a depth of 0");
```
tlively, line 2216: It would also be good to check that the HeapType on the instruction matches the HeapType in the expression's type.

tlively: Oh, I see, it's not stored separately 👍

tlively:
Oh oops, I had the tab open for a while and didn't see that this had already been merged. Don't worry about my suggested changes above. I'll make a PR for them.

src/wasm-interpreter.h
tlively: Looks like this confusion is because I was reading Overview.md when I should have been reading MVP.md. I filed https://github.com/WebAssembly/gc/issues/167 to fix this.

tlively:
> And unrelated to this PR, but I have a basic question on the GC proposal itself: What is the purpose of RTTs? They seem to be derived from existing static types anyway... Why don't we just use static types in their place?

@aheejin, from what I can gather from the explainer, the most important reason to have explicit Rtt values is to make them optional. Structs and arrays that do not have Rtts attached to them at allocation time can never be casted and therefore the engine can theoretically do extra optimizations on them. I got this from reading Overview.md and it's a little unclear whether it is still the case in MVP.md.

PR #3434, opened 2020-12-09
Small RTT cleanups

tlively:


src/wasm-type.h
```diff
@@ -432,10 +434,8 @@ struct Array {
 };
 
 struct Rtt {
-  enum {
-    // An Rtt can have no depth specified
-    NoDepth = -1
-  };
+  // An Rtt can have no depth specified
+  static constexpr uint32_t NoDepth = -1;
```
kripken, line 438: I'm fine either way, but what is the benefit of this approach? The type?

src/wasm-type.h
```diff
@@ -174,10 +174,12 @@ class Type {
...
   FeatureSet getFeatures() const;
 
-  // Gets the heap type corresponding to this type
+  // Gets the heap type corresponding to this type, assuming that it is a
+  // reference or Rtt type.
   HeapType getHeapType() const;
 
-  const struct Rtt& getRtt() const;
+  // Gets the Rtt for this type, assuming that it is an Rtt type.
+  Rtt getRtt() const;
```
kripken, line 182: What's the benefit to removing `const&`? At least the const part seems nice to clarify that this does not return an item to be modified.

kripken (approved):

src/wasm-type.h
tlively: There's no concrete benefit, but since #3428 we have a documented convention of passing `Rtt`s by value. Once you're returning by value, I don't think the `const` has any effect.

src/wasm-type.h
tlively: Yeah, the type explicitly matches its uses, the code is slightly shorter, and it's consistent with other magic values defined in this file (i.e. `_last_basic_type`).

src/wasm-type.h
kripken: Isn't there a small benefit to disallowing `getRtt().depth = 5` at compile time?

tlively: I just checked locally and that is disallowed with or without the `const` with "error: expression is not assignable."

kripken: Oh nice, thanks for checking!

PR #3435, opened 2020-12-09
Read and write data segments names in names section

sbc100:


sbc100:
Still trying to figure out where/how best to test this.

sbc100:
I accidentally landed llvm support for this section which is causing warnings to be generated by binaryen.

So we now need to land this with some urgency as its blocking the llvm roll.   But I should fine a way to test it first.

tlively (approved):

PR #3436, opened 2020-12-10
[GC] Add Array operations

kripken:
array.new/get/set/len - pretty straightforward after structs and all the
infrastructure for them.

Also fixes validation of the unnecessary heapType param in the
text and binary formats in structs as well as arrays.

Fixes printing of packed types in type names, which emitted i32
for them. That broke when we emitted the same name for an array
of i8 and i32 as in the new testing here.

Also fix a bug in `Field::operator<` which was wrong for packed
types; again, this was easy to notice with the new testing.

### DIFF
--- a/src/ir/cost.h
+++ b/src/ir/cost.h
@@ -586,12 +586,20 @@ struct CostAnalyzer : public OverriddenVisitor<CostAnalyzer, Index> {
     return 1 + nullCheckCost(curr->ref) + visit(curr->ref);
   }
   Index visitStructSet(StructSet* curr) {
-    return 1 + nullCheckCost(curr->ref) + visit(curr->ref) + visit(curr->value);
+    return 2 + nullCheckCost(curr->ref) + visit(curr->ref) + visit(curr->value);
   }
-  Index visitArrayNew(ArrayNew* curr) { WASM_UNREACHABLE("TODO: GC"); }
-  Index visitArrayGet(ArrayGet* curr) { WASM_UNREACHABLE("TODO: GC"); }
-  Index visitArraySet(ArraySet* curr) { WASM_UNREACHABLE("TODO: GC"); }
-  Index visitArrayLen(ArrayLen* curr) { WASM_UNREACHABLE("TODO: GC"); }
+  Index visitArrayNew(ArrayNew* curr) {
+    return 4 + visit(curr->rtt) + visit(curr->size) + visit(curr->init);
+  }
+  Index visitArrayGet(ArrayGet* curr) {
+    return 1 + nullCheckCost(curr->ref) + visit(curr->ref) + visit(curr->index);
+  }
+  Index visitArraySet(ArraySet* curr) {
+    return 2 + nullCheckCost(curr->ref) + visit(curr->ref) + visit(curr->index) + visit(curr->value);
+  }
+  Index visitArrayLen(ArrayLen* curr) {
+    return 1 + nullCheckCost(curr->ref) + visit(curr->ref);
+  }
 
 private:
   Index nullCheckCost(Expression* ref) {
--- a/src/ir/cost.h
+++ b/src/ir/cost.h
@@ -595,7 +595,8 @@ struct CostAnalyzer : public OverriddenVisitor<CostAnalyzer, Index> {
     return 1 + nullCheckCost(curr->ref) + visit(curr->ref) + visit(curr->index);
   }
   Index visitArraySet(ArraySet* curr) {
-    return 2 + nullCheckCost(curr->ref) + visit(curr->ref) + visit(curr->index) + visit(curr->value);
+    return 2 + nullCheckCost(curr->ref) + visit(curr->ref) +
+           visit(curr->index) + visit(curr->value);
   }
   Index visitArrayLen(ArrayLen* curr) {
     return 1 + nullCheckCost(curr->ref) + visit(curr->ref);
--- a/src/ir/effects.h
+++ b/src/ir/effects.h
@@ -556,8 +556,7 @@ private:
     }
     void visitRttCanon(RttCanon* curr) {}
     void visitRttSub(RttSub* curr) {}
-    void visitStructNew(StructNew* curr) {
-    }
+    void visitStructNew(StructNew* curr) {}
     void visitStructGet(StructGet* curr) {
       // traps when the arg is null
       if (curr->ref->type.isNullable()) {
@@ -570,8 +569,7 @@ private:
         parent.implicitTrap = true;
       }
     }
-    void visitArrayNew(ArrayNew* curr) {
-    }
+    void visitArrayNew(ArrayNew* curr) {}
     void visitArrayGet(ArrayGet* curr) {
       // traps when the arg is null or the index out of bounds
       parent.implicitTrap = true;
--- a/src/ir/effects.h
+++ b/src/ir/effects.h
@@ -571,16 +571,20 @@ private:
       }
     }
     void visitArrayNew(ArrayNew* curr) {
-      WASM_UNREACHABLE("TODO (gc): array.new");
     }
     void visitArrayGet(ArrayGet* curr) {
-      WASM_UNREACHABLE("TODO (gc): array.get");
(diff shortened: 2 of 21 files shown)


src/ir/cost.h
```diff
@@ -586,12 +586,21 @@ struct CostAnalyzer : public OverriddenVisitor<CostAnalyzer, Index> {
     return 1 + nullCheckCost(curr->ref) + visit(curr->ref);
   }
   Index visitStructSet(StructSet* curr) {
-    return 1 + nullCheckCost(curr->ref) + visit(curr->ref) + visit(curr->value);
+    return 2 + nullCheckCost(curr->ref) + visit(curr->ref) + visit(curr->value);
```
kripken, line 589: this is changed to match `store`

src/wasm/wasm-validator.cpp
```diff
@@ -2304,25 +2304,72 @@ void FunctionValidator::visitStructSet(StructSet* curr) {
...
+  if (!shouldBeTrue(
+        heapType.isArray(), curr, "array.new heap type must be array")) {
+    return;
+  }
+  const auto& element = heapType.getArray().element;
+  if (curr->isWithDefault()) {
+    shouldBeTrue(
+      !curr->init, curr, "array.new_with_default should have no init");
+    // The element must be defaultable.
+    // TODO: add type.isDefaultable()?
```
aheejin, line 2326: This sounds good; so all number types can return true for `isDefaultable`

src/wasm/wasm.cpp
```diff
@@ -1123,10 +1123,39 @@ void StructSet::finalize() {
...
-// TODO (gc): array.get
-// TODO (gc): array.set
-// TODO (gc): array.len
+void ArrayNew::finalize() {
+  if (rtt->type == Type::unreachable || size->type == Type::unreachable ||
+      (init && init->type == Type::unreachable)) {
+    type = Type::unreachable;
+    return;
+  }
+  type = Type(rtt->type.getHeapType(), /* nullable = */ false);
```
aheejin, line 1132: Why `false` for `nullable`? Aren't we assuming all types are nullable for the moment?

src/wasm/wasm-validator.cpp
```diff
@@ -2304,25 +2304,72 @@ void FunctionValidator::visitStructSet(StructSet* curr) {
...
     getModule()->features.hasGC(), curr, "array.get requires gc to be enabled");
-  WASM_UNREACHABLE("TODO (gc): array.get");
+  shouldBeEqualOrFirstIsUnreachable(
+    curr->index->type, Type(Type::i32), curr, "array.get index must be an i32");
+  if (curr->type == Type::unreachable) {
+    return;
+  }
+  const auto& element = curr->ref->type.getHeapType().getArray().element;
+  shouldBeEqual(
+    curr->type, element.type, curr, "array.get must have the proper type");
```
aheejin, line 2350: Should we check if the element type is a packed type if `signed_` is true? (The same for `visitStructGet`)

aheejin (approved):
LGTM % questions

src/ir/effects.h
```diff
@@ -571,16 +571,20 @@ class EffectAnalyzer {
...
     }
     void visitArraySet(ArraySet* curr) {
-      WASM_UNREACHABLE("TODO (gc): array.set");
+      // traps when the arg is null or the index out of bounds
+      parent.implicitTrap = true;
     }
     void visitArrayLen(ArrayLen* curr) {
-      WASM_UNREACHABLE("TODO (gc): array.len");
+      // traps when the arg is null
+      if (curr->ref->type.isNullable()) {
```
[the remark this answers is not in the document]
kripken: `array.len` will only trap if the reference is null, so if we know statically that it isn't null, there is no trap. But that's not the case for `array.set` for example, which can also trap for a second reason, if the index is out of bounds. There is no way to verify the index at compile time (even if the index were constant, array types don't have a length). So we may always trap there.

src/wasm/wasm-validator.cpp
kripken: Good idea, adding.

src/wasm/wasm.cpp
kripken: Yes, until we figure out how to do non-nullability we have to assume all types are nullable.

src/wasm/wasm-stack.cpp
```diff
@@ -1940,21 +1940,38 @@ void BinaryInstWriter::visitStructSet(StructSet* curr) {
...
+    o << U32LEB(BinaryConsts::ArrayNewDefaultWithRtt);
+  } else {
+    o << U32LEB(BinaryConsts::ArrayNewWithRtt);
+  }
+  parent.writeHeapType(curr->rtt->type.getHeapType());
 }
 
 void BinaryInstWriter::visitArrayGet(ArrayGet* curr) {
-  WASM_UNREACHABLE("TODO (gc): array.get");
+  const auto& heapType = curr->ref->type.getHeapType();
```
tlively, line 1953:
```suggestion
  auto heapType = curr->ref->type.getHeapType();
```
Huh, TIL you can have a const reference to an rvalue. Not taking a reference to the HeapType should be more idiomatic, though.

src/wasm/wasm-stack.cpp
```diff
@@ -1940,21 +1940,38 @@ void BinaryInstWriter::visitStructSet(StructSet* curr) {
...
+    op = BinaryConsts::ArrayGetU;
+  }
+  o << int8_t(BinaryConsts::GCPrefix) << int8_t(op);
+  parent.writeHeapType(heapType);
 }
 
 void BinaryInstWriter::visitArraySet(ArraySet* curr) {
-  o << int8_t(BinaryConsts::GCPrefix) << U32LEB(BinaryConsts::ArraySet);
-  WASM_UNREACHABLE("TODO (gc): array.set");
+  o << int8_t(BinaryConsts::GCPrefix) << int8_t(BinaryConsts::ArraySet);
```
tlively, line 1968: It was correct to use `U32LEB` here and below. Why change? You also use `U32LEB` up in `visitArrayNew` but not in `visitArrayGet`.

src/ir/effects.h
```diff
@@ -571,16 +571,20 @@ class EffectAnalyzer {
       }
     }
     void visitArrayNew(ArrayNew* curr) {
-      WASM_UNREACHABLE("TODO (gc): array.new");
```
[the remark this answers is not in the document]
kripken: Looks like clang-format accepts both, and leaves whichever is already present, interesting. I'll make them the way you expect.

src/wasm-interpreter.h
```diff
@@ -1445,40 +1445,117 @@ class ExpressionRunner : public OverriddenVisitor<SubType, Flow> {
...
+      return rtt;
+    }
+    auto size = this->visit(curr->size);
+    if (size.breaking()) {
+      return size;
+    }
+    const auto& element = curr->rtt->type.getHeapType().getArray().element;
+    Literals data;
+    Index num = size.getSingleValue().geti32();
+    data.resize(num);
```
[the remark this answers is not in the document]
kripken: Sounds good, will add a SmallVector constructor to support this.

src/wasm/wasm-stack.cpp
kripken: Oh, due to copy-pasting... sorry, will fix.

src/wasm/wasm.cpp
aheejin: Then shouldn't this be `true`?

kripken: Oh, right, I keep mixing up non-nullability with nullability...

tlively: Another nice cleanup we could do would be to add globally available enums for `Nullable, NonNullable` and `Mutable, NonMutable` that we could use rather than bools. I can do that if you'd like.

kripken: Ah, good idea - yes, that would be great if you have time!

tlively (approved):

PR #3437, opened 2020-12-10
[wasm-reduce] Improve support for reducing on text files

kripken:
Passing `--detect-features` there doesn't work (as there is no feature
section).

Maybe there's a simpler way though?


tlively (approved):
I wonder if we should make a text format for the target features section 🤔

PR #3438, opened 2020-12-10
[GC] Fix Array optimization issues

kripken:
Precompute still tried to precompute a reference because the check was
not in the topmost place.

Also we truncated i8/i16 values, but did not extend them properly. That
was also an issue with structs.

The new test replaces the old one by moving from `-O1` to `-Oz` (which
runs more opts, and would have noticed this earlier), and adds array
operations too, including sign extends.


src/wasm-interpreter.h
```diff
@@ -1541,13 +1543,37 @@ class ExpressionRunner : public OverriddenVisitor<SubType, Flow> {
...
+  Literal extendForPacking(Literal value, const Field& field, bool signed_) {
     if (field.type == Type::i32) {
+      int32_t c = value.geti32();
       if (field.packedType == Field::i8) {
-        value = value.and_(Literal(int32_t(0xff)));
+        if (signed_) {
+          value = Literal((c << 24) >> 24);
+        } else {
+          // The stored value should already be truncated.
+          assert(c == (c & 0xff));
```
tlively, line 1569: This should be true in both cases, so how about moving this just before the `if (signed_)`? Same below.

src/passes/Precompute.cpp
```diff
@@ -228,6 +223,12 @@ struct Precompute
   // Precompute an expression, returning a flow, which may be a constant
   // (that we can replace the expression with if replaceExpression is set).
   Flow precomputeExpression(Expression* curr, bool replaceExpression = true) {
+    // Don't try to precompute a reference. We can't replace it with a constant
+    // expression, as that would make a copy of it by value.
+    // TODO: do so when safe
+    if (curr->type.isRef()) {
+      return Flow(NONCONSTANT_FLOW);
+    }
```
tlively, line 231: I don't quite understand why moving this makes a difference. What's a case where the previous code would have been incorrect? Would moving it to the top of `visitExpression` work as well? If so, that seems like a better place for this.

kripken: The previous location was in `visitExpression` right before a call to `precomputeExpression`. The new location is in `precomputeExpression` itself, which guarantees that all callers have the right behavior (which includes `precomputeValue` in particular, where the new bug was).

src/wasm-interpreter.h
kripken: Good idea, that's shorter too. Done.

tlively (approved):

src/passes/Precompute.cpp
tlively: Got it, thanks!

PR #3439, opened 2020-12-11
[GC] Add ref.test and ref.cast

kripken:
This adds enough to read and write them and test that, but leaves
interpreter support for later (for which I need to actually understand
RTTs...).

### DIFF
--- a/src/wasm-type.h
+++ b/src/wasm-type.h
@@ -346,7 +346,7 @@ struct Tuple {
   std::string toString() const;
 
   // Prevent accidental copies
-  Struct& operator=(const Struct&) = delete;
+  Tuple& operator=(const Tuple&) = delete;
 
 private:
   void validate() {
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -5599,10 +5599,13 @@ bool WasmBinaryBuilder::maybeVisitRefTest(Expression*& out, uint32_t code) {
   if (code != BinaryConsts::RefTest) {
     return false;
   }
-  auto* curr = allocator.alloc<RefTest>();
-  WASM_UNREACHABLE("TODO (gc): ref.test");
-  curr->finalize();
-  out = curr;
+  auto heapType1 = getHeapType();
+  auto heapType2 = getHeapType();
+  auto* ref = popNonVoidExpression();
+  validateHeapTypeUsingChild(ref, heapType1);
+  auto* rtt = popNonVoidExpression();
+  validateHeapTypeUsingChild(rtt, heapType2);
+  out = Builder(wasm).makeRefTest(ref, rtt);
   return true;
 }
 
@@ -5610,10 +5613,13 @@ bool WasmBinaryBuilder::maybeVisitRefCast(Expression*& out, uint32_t code) {
   if (code != BinaryConsts::RefCast) {
     return false;
   }
-  auto* curr = allocator.alloc<RefCast>();
-  WASM_UNREACHABLE("TODO (gc): ref.cast");
-  curr->finalize();
-  out = curr;
+  auto heapType1 = getHeapType();
+  auto heapType2 = getHeapType();
+  auto* ref = popNonVoidExpression();
+  validateHeapTypeUsingChild(ref, heapType1);
+  auto* rtt = popNonVoidExpression();
+  validateHeapTypeUsingChild(rtt, heapType2);
+  out = Builder(wasm).makeRefCast(ref, rtt);
   return true;
 }
 
--- a/src/wasm/wasm-s-parser.cpp
+++ b/src/wasm/wasm-s-parser.cpp
@@ -2093,17 +2093,19 @@ Expression* SExpressionWasmBuilder::makeI31Get(Element& s, bool signed_) {
 }
 
 Expression* SExpressionWasmBuilder::makeRefTest(Element& s) {
-  auto ret = allocator.alloc<RefTest>();
-  WASM_UNREACHABLE("TODO (gc): ref.test");
-  ret->finalize();
-  return ret;
+  auto heapType = parseHeapType(*s[1]);
+  auto* ref = parseExpression(*s[2]);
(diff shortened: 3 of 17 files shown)


src/wasm-type.h
```diff
@@ -346,7 +346,7 @@ struct Tuple {
   std::string toString() const;
 
   // Prevent accidental copies
-  Struct& operator=(const Struct&) = delete;
```
tlively, line 349: Oops 😅

src/wasm/wasm-binary.cpp
```diff
@@ -5599,21 +5599,27 @@ bool WasmBinaryBuilder::maybeVisitRefTest(Expression*& out, uint32_t code) {
   if (code != BinaryConsts::RefTest) {
     return false;
   }
-  auto* curr = allocator.alloc<RefTest>();
-  WASM_UNREACHABLE("TODO (gc): ref.test");
-  curr->finalize();
-  out = curr;
+  auto heapType1 = getHeapType();
+  auto heapType2 = getHeapType();
```
tlively, line 5603: So instead of taking one static heap type this instruction takes two static heap types and a dynamic heap type!? I must be missing something big here...

src/wasm/wasm-binary.cpp
```diff
@@ -5599,21 +5599,27 @@ bool WasmBinaryBuilder::maybeVisitRefTest(Expression*& out, uint32_t code) {
...
+  auto* ref = popNonVoidExpression();
+  validateHeapTypeUsingChild(ref, heapType1);
+  auto* rtt = popNonVoidExpression();
+  validateHeapTypeUsingChild(rtt, heapType2);
```
tlively, line 5607: Hmm, I wonder if `validateChildUsingHeapType` would paint a clearer picture of what this does.

src/wasm/wasm-s-parser.cpp
```diff
@@ -2093,17 +2093,19 @@ Expression* SExpressionWasmBuilder::makeI31Get(Element& s, bool signed_) {
 }
 
 Expression* SExpressionWasmBuilder::makeRefTest(Element& s) {
-  auto ret = allocator.alloc<RefTest>();
-  WASM_UNREACHABLE("TODO (gc): ref.test");
-  ret->finalize();
-  return ret;
+  auto heapType = parseHeapType(*s[1]);
```
tlively, line 2096: The text format only has one static heap type even though the binary format has two?

src/wasm/wasm-binary.cpp
kripken: It does seem excessive, yeah. Not sure what you mean by 2 and 1, though? It gets two arguments that have a type with a heap type, and it also gets those heap types as immediates.

src/wasm/wasm-binary.cpp
kripken: I find that less clear actually. The child is already known to be valid at this point. The only question is whether the extra heap type immediate matches the child. But I don't feel strongly about the name.

src/wasm/wasm-s-parser.cpp
kripken: I believe so? [MVP doc](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#casts) says
```
ref.test $t : [(ref null ht) (rtt n? $t)] -> [i32]
```
for the text, and [binary format doc](https://docs.google.com/document/d/1DklC3qVuOdLHSXB5UXghM_syCh-4cMinQ50ICiXnK3Q/edit?pli=1#) says
```
ref.test | 0xfb40 | <heaptype1> <heaptype2> | [(ref null t1) (rtt n t2)] -> [i32]
```
None of these heap types are actually necessary, so I do hope we can get everyone to agree to remove them before we ship... but maybe there is some baseline compiler angle that I am missing, that makes it better to add them? Anyhow, we can figure that out later I guess.

@jakobkummerow am I misreading the binary format doc perhaps?

src/wasm/wasm-binary.cpp
tlively: The one "dynamic heap type" I was thinking about was the rtt value. We certainly couldn't get rid of the reference typed argument!

tlively (approved):
LGTM, but I also hope we can slim down the binary format and improve consistency between the binary and text formats in the future.

src/wasm/wasm-s-parser.cpp
jakobkummerow: The "binary format doc" reflects an earlier version of MVP.md, and we're intentionally keeping that stable for now [1]. See comments on that doc for what the V8 prototype currently implements (ref.test and ref.cast with heap type immediates, rtt.sub and br_on_cast without -- it's a bit inconsistent, sorry about that).

I agree that it would be both feasible from an implementation PoV, and desirable from a module size PoV, to drop these type immediates that can easily be inferred by looking at the types of the values on the expression stack.

[1] See the very top where it says "version 2". We're only making backwards-compatible changes at the moment (e.g. Manos added `anyref` support recently), so that toolchains have a non-moving target to work with. That means our prototype does not reflect the latest changes to MVP.js, such as RTTs without depth. At some point we'll define a "version 3" of the prototype spec doc and update the prototype's implementation accordingly; timeline TBD. I'll communicate it early and widely when that happens.

kripken: I see, thanks @jakobkummerow . I may have been doing the wrong thing then, I've only read the binary format doc when something wasn't clear enough in the MVP. So there may be small differences atm between binaryen and v8/wabt. I'll do a pass eventually to reconcile all that, maybe a good time would be at version 3 of that doc, or when a user is ready to use it, whichever is earlier.

src/wasm/wasm-binary.cpp
kripken: I think we also can't get rid of the rtt - the VM checks that value at runtime. So both the ref and rtt seem equally necessary?

PR #3442, opened 2020-12-14
Refactor BranchUtils to avoid boilerplate. NFC

kripken:
This should make it easier to add `br_on_cast` for example.


src/ir/branch-utils.h
```diff
@@ -28,111 +28,108 @@ namespace BranchUtils {
...
 }
 
-inline bool isBranchReachable(Switch* sw) {
-  return !(sw->value && sw->value->type == Type::unreachable) &&
-         sw->condition->type != Type::unreachable;
+// Perform a generic operation on uses of scope names (branch targets) in an
+// expression. The provided function receives a Name& which it can modify if it
+// needs to.
+inline void operateOnScopeNameUses(Expression* expr,
+                                   std::function<void(Name&)> func) {
```
tlively, line 47: It might be good to templatize this over some callable `F` rather than using `std::function` to avoid the lambdas being converted.

tlively (approved):
Nice!

src/ir/branch-utils.h
kripken: I guess that's shorter too, sounds good.

Is there an overhead to the conversion, though? I'd hope `std::function` is a small wrapper around a function pointer?

tlively: I could imagine that the conversion could defeat the optimizer and prevent the lambda from being inlined, but perhaps the optimizer is smart enough to work around that 🤷‍♂️

PR #3443, opened 2020-12-15
Use enums for mutability and nullability

tlively:
Previously we were using bools for both of these concepts, but using enums makes
the code clearer. In particular, the PR removes many instances of
`/*nullability=*/ true`.

### DIFF
--- a/src/wasm-type.h
+++ b/src/wasm-type.h
@@ -45,7 +45,7 @@ struct Array;
 struct Rtt;
 
 enum Nullability { NonNullable, Nullable };
-enum Mutability { NonMutable, Mutable };
+enum Mutability { Immutable, Mutable };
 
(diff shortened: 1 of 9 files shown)


src/wasm-type.h
```diff
@@ -44,6 +44,9 @@ struct Struct;
 struct Array;
 struct Rtt;
 
+enum Nullability { NonNullable, Nullable };
+enum Mutability { NonMutable, Mutable };
```
kripken, line 48: Maybe Immutable? But NonMutable does capitalize better.

kripken (approved):
Nice!

src/wasm-type.h
tlively: Oh yeah, if we're going to switch to using meaningful English words, might as well get them right 👍

PR #3445, opened 2020-12-15
[GC] Add BrOnCast instruction

kripken:
This is I believe the last missing instruction in the [wasm GC prototype](https://docs.google.com/document/d/1DklC3qVuOdLHSXB5UXghM_syCh-4cMinQ50ICiXnK3Q/edit?pli=1#) doc :partying_face:

### DIFF
--- a/src/ir/ReFinalize.cpp
+++ b/src/ir/ReFinalize.cpp
@@ -155,7 +155,12 @@ void ReFinalize::visitRefTest(RefTest* curr) { curr->finalize(); }
 void ReFinalize::visitRefCast(RefCast* curr) { curr->finalize(); }
 void ReFinalize::visitBrOnCast(BrOnCast* curr) {
   curr->finalize();
-  WASM_UNREACHABLE("TODO (gc): br_on_cast");
+  if (curr->type == Type::unreachable) {
+    replaceUntaken(curr->ref, nullptr);
+  } else {
+    updateBreakValueType(curr->name,
+                         Type(curr->rtt->type.getHeapType(), Nullable));
+  }
 }
 void ReFinalize::visitRttCanon(RttCanon* curr) { curr->finalize(); }
 void ReFinalize::visitRttSub(RttSub* curr) { curr->finalize(); }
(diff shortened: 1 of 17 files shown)


aheejin:
Does this model the peculiar behavior similar to `br_on_exn`? Or is it actually the case that `br_on_cast` has the same semantics? What I'm talking about is, when taken, `br_on_exn` sends extracted values as block return values, but if not, it leaves the operand unchanged. That's why we had to come up with something like `sent` for `br_on_exn`, because there was no way to express two kinds of types in a single expression.

I don't seem to find any mention of this behavior in MVP.md, but [this section](https://github.com/WebAssembly/gc/blob/master/proposals/gc/Overview.md#casting-and-runtime-types) in the overview seems to have it:
> This branches to `$label` if the check is successful, with the operand as an argument, but using its refined type. Otherwise, the operand remains on the stack. By chaining multiple of these instructions, runtime type analysis ("typecase") can be implemented:

And the example below shows a case that you can use multiple `br_on_cast`s on a single value on the stack.

I don't think the PR reflects this behavior, because it doesn't have any notion of separate type and 'sent' type. Am I right..?

kripken:
I think your description of the semantics is correct. If the value is non-null and the cast succeeds, we take the branch with the cast value. Otherwise we return the input ref with the original type.

And I think you're right that this PR is not correct - good point! It should work in general, but it uses the rtt object's type to figure out the sent type. I forgot that that could be unreachable, in which case we'd fail... which is I guess why we have `sent` on BrOnExn.

I'll look into this some more, I wonder if maybe we can solve things better for both instructions in some way. Let me close this for now as hopefully I'll be able to do some refactoring first to make this easier.

aheejin:
> I'll look into this some more, I wonder if maybe we can solve things better for both instructions in some way. Let me close this for now as hopefully I'll be able to do some refactoring first to make this easier.

FYI, `br_on_exn` is gonna be gone soon, so we don't need to refactor code to make both instructions share code.

kripken:
@aheejin 

Thanks, good to know that about `br_on_exn`.

PR #3448, opened 2020-12-16
More refactoring of branch utility code to remove boilerplate. NFC

kripken:
This will help with BrOnCast.

### DIFF
--- a/src/wasm.h
+++ b/src/wasm.h
@@ -703,11 +703,13 @@ public:
   // needed (which may require scanning the block)
   void finalize(Type type_);
 
+  enum Breakability { Unknown, HasBreak, NoBreak };
+
   // set the type given you know its type, and you know if there is a break to
   // this block. this avoids the need to scan the contents of the block in the
   // case that it might be unreachable, so it is recommended if you already know
   // the type and breakability anyhow.
-  void finalize(Type type_, bool hasBreak);
+  void finalize(Type type_, Breakability breakability);
 };
 
 class If : public SpecificExpression<Expression::IfId> {
--- a/src/wasm.h
+++ b/src/wasm.h
@@ -1842,4 +1842,6 @@ template<> struct hash<wasm::Address> {
 };
 } // namespace std
 
+#include "wasm-printing.h"
+
 #endif // wasm_wasm_h
--- a/src/wasm.h
+++ b/src/wasm.h
@@ -1842,6 +1842,4 @@ template<> struct hash<wasm::Address> {
 };
 } // namespace std
 
-#include "wasm-printing.h"
-
 #endif // wasm_wasm_h
--- a/src/wasm/wasm.cpp
+++ b/src/wasm/wasm.cpp
@@ -305,36 +305,17 @@ void Block::finalize() {
     type = Type::none;
     return;
   }
+  // The default type is what is at the end. Next we need to see if breaks and/
+  // or unreachabitily change that.
+  type = list.back()->type;
   if (!name.is()) {
-    // nothing branches here, so this is easy
-    // normally the type is the type of the final child
-    type = list.back()->type;
-    // and even if we have an unreachable child somewhere,
-    // we still mark ourselves as having that type,
-    // (block (result i32)
-    //  (return)
-    //  (i32.const 10)
-    // )
-    if (type.isConcrete()) {
-      return;
-    }
-    // if we are unreachable, we are done
-    if (type == Type::unreachable) {
-      return;
-    }
-    // we may still be unreachable if we have an unreachable
-    // child
-    for (auto* child : list) {
-      if (child->type == Type::unreachable) {
-        type = Type::unreachable;
-        return;
-      }
-    }
+    // Nothing branches here, so this is easy.
+    handleUnreachable(
+      this, /* breakabilityKnown = */ true, /* hasBreak = */ false);
     return;
   }
 
   // The default type is according to the value that flows out.
-  type = list.back()->type;
   BranchUtils::BranchSeeker seeker(this->name);
   Expression* temp = this;
   seeker.walk(temp);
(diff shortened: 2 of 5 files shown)


src/wasm.h
```diff
@@ -1842,4 +1842,6 @@ template<> struct hash<wasm::Address> {
 };
 } // namespace std
 
+#include "wasm-printing.h"
+
```
tlively, line 1846: This looks like an unintentional addition.

src/wasm/wasm.cpp
```diff
@@ -362,41 +301,55 @@ static void handleUnreachable(Block* block,
...
+    // )
+    if (type.isConcrete()) {
+      return;
+    }
+    // if we are unreachable, we are done
+    if (type == Type::unreachable) {
+      return;
+    }
+    // we may still be unreachable if we have an unreachable
+    // child
```
tlively, line 326: It might be worth pointing out explicitly that only blocks that would otherwise have `Type::none` are eligible to become `unreachable` if they have an unreachable child. The previous two `if`s could even be replaced with `if (type != Type::none)`. Also, some of this work looks duplicated with work done in `handleUnreachable`. Can that function be used here?

src/wasm/wasm.cpp
```diff
@@ -362,41 +301,55 @@ static void handleUnreachable(Block* block,
...
   }
 
-  TypeSeeker seeker(this, this->name);
-  type = Type::mergeTypes(seeker.types);
-  handleUnreachable(this);
+  // The default type is according to the value that flows out.
+  type = list.back()->type;
+  BranchUtils::BranchSeeker seeker(this->name);
+  Expression* temp = this;
+  seeker.walk(temp);
```
tlively, line 340: Would `seeker.walk(this)` work?

src/wasm/wasm.cpp
```diff
@@ -362,41 +301,55 @@ static void handleUnreachable(Block* block,
...
+    // Take the branch values into account.
+    if (seeker.valueType != Type::none) {
+      type = Type::getLeastUpperBound(type, seeker.valueType);
+    } else {
+      // No value is sent, but as we have a branch we are not unreachable.
+      type = Type::none;
+    }
+  } else {
+    // There are no branches, so this block may be unreachable.
+    handleUnreachable(this);
```
tlively, line 351: It would be good to pass `breakabilityKnown = true, hasBreak = false` if we know there are no branches.

src/wasm.h
kripken: Oh, sorry, I meant to comment - that was intentional. I got tired of typing it in every time I debug. We do include Type, Literal, etc. printing by default, and were just missing this one, so I added it. Maybe there's a nicer place to put it though?

src/wasm/wasm.cpp
kripken: No, as that function takes a reference, which it can potentially modify (all walkers can do so in general, if you call `replaceCurrent`).

It would be nice if we could have two walkers,
```cpp
walk(Expression*&)
walk(Expression*)
```
and the latter would be a fallback for a case like this. But C++ doesn't seem to allow that...

src/wasm.h
tlively: Hmm, those other files don't depend on anything declared in wasm.h though. This is kind of weird in that it creates a cyclic dependency between the two files.

src/wasm/wasm.cpp
kripken: Good point! This is pre-existing code that was just de-indented, so I didn't read it carefully til now. Yes, it seems like it does duplicate the other pre-existing code... will fix.

kripken:
Refactoring for the feedback - even shorter now :smiley:

src/wasm.h
kripken: Maybe a nicer way is to just embed the contents of `wasm-printing.h` in `wasm.h`? That's basically what we do for all the others. (OTOH, separating the file out is more modular)

src/wasm/wasm.cpp
```diff
@@ -362,41 +301,37 @@ static void handleUnreachable(Block* block,
...
-      type = list.back()->type;
-      // and even if we have an unreachable child somewhere,
-      // we still mark ourselves as having that type,
-      // (block (result i32)
-      //  (return)
-      //  (i32.const 10)
-      // )
-      if (type.isConcrete()) {
-        return;
-      }
```
aheejin, line 378: This seems missing in the new code? So the old code uses the last element's type if it's concrete without checking other elements, and the new code rather uses the current block's type if it's concrete w/o checking other elements.

src/wasm/wasm.cpp
```diff
@@ -362,41 +301,37 @@ static void handleUnreachable(Block* block,
...
-      // child
-      for (auto* child : list) {
-        if (child->type == Type::unreachable) {
-          type = Type::unreachable;
-          return;
-        }
-      }
+    // Nothing branches here, so this is easy.
+    handleUnreachable(
+      this, /* breakabilityKnown = */ true, /* hasBreak = */ false);
```
aheejin, line 314: In this code, if the block already has a concrete type, we honor it, because this `handleUnreachable` does that. But the old code does not have that step. Is this intentional?

src/wasm/wasm.cpp
tlively: Maybe we could give `Walker` a helper method like `walkImmutable` that takes care of this temporary variable dance, but I don't love that solution either.

src/wasm.h
tlively: Oh yes, wasm-printing.h is fairly short and straightforward. I think it would make sense to put at least the declarations of the various `operator<<`s in wasm.h. It would be good to do that as a separate change, though.

src/wasm/wasm.cpp
tlively, line 314: It might be worth introducing an `enum Breakability { Unknown, HasBreak, NoBreak };` to replace these bools.

aheejin (approved):

PR #3450, opened 2020-12-17
Refactor printing code so that printing Expressions always works

kripken:
This avoids needing to add `include wasm-printing` if a file doesn't already have it.
To achieve that, add the `std::ostream` hooks in `wasm.h`, and also use them
when possible, removing the need for the special WasmPrinter object.

Stop printing in "full" (print types on each line) in error messages by default. The
user can still get that, as always, using `BINARYEN_PRINT_FULL=1` in the env.


aheejin:
`dump` method can be useful in gdb; `<<` operator is hard to use in gdb, so `dump` might be good to keep.

src/wasm.h
```diff
@@ -659,9 +660,6 @@ class Expression {
     assert(int(_id) == int(T::SpecificId));
     return (const T*)this;
   }
-
-  // Print the expression to stderr. Meant for use while debugging.
-  void dump();
```
aheejin, line 664: `dump` method can be useful in gdb; `<<` operator is hard to use in gdb, so `dump` might be good to keep.

kripken: Oh, I didn't think about that... I restored `dump` now.

src/parsing.h
```diff
@@ -26,7 +26,7 @@
 #include "shared-constants.h"
 #include "support/colors.h"
 #include "support/utilities.h"
-#include "wasm-printing.h"
+#include "wasm-traversal.h"
```
tlively, line 29: Is this because `wasm-traversal.h` was previously included indirectly through `wasm-printing.h`?

src/passes/Print.cpp
```diff
@@ -21,12 +21,22 @@
...
+static std::ostream& printExpression(Expression* expression,
+                                     std::ostream& o,
+                                     bool minify = false,
+                                     bool full = false);
+
+static std::ostream&
+printStackInst(StackInst* inst, std::ostream& o, Function* func = nullptr);
+
+static std::ostream&
+printStackIR(StackIR* ir, std::ostream& o, Function* func = nullptr);
```
tlively, line 38: Would it work to move these into the anonymous namespace below (and do the same for their definitions) rather than mark them `static`?

src/wasm/wasm-validator.cpp
```diff
@@ -44,7 +43,7 @@ inline std::ostream& printModuleComponent(T curr, std::ostream& stream) {
 // Extra overload for Expressions, to print type info too
 inline std::ostream& printModuleComponent(Expression* curr,
                                           std::ostream& stream) {
-  WasmPrinter::printExpression(curr, stream, false, true) << std::endl;
+  stream << *curr << '\n';
```
tlively, line 46: Am I correct that this will previously printed the expression in "full" mode but now it does not? If so, is this intended?

src/parsing.h
kripken: Yes, it didn't build without this.

src/wasm/wasm-validator.cpp
kripken: Yes, I think this is ok to do. (But it makes this not NFC.)

The user can always set `BINARYEN_PRINT_FULL=1` to get "full" printing (i.e. print types on each line) when necessary. I'm not sure how helpful it is to do by default - in fact it is sometimes annoying.

src/passes/Print.cpp
kripken: Not easily... the anonymous namespace starts and ends pretty early in the file, but these impls need stuff all the way at the end. So it might take a lot of code motion to use an anonymous namespace for this.

Another option might be to use a non-anonymous namespace for them, but that seems less nice I think.

tlively: Ok, no problem then 👍

tlively (approved):

PR #3451, opened 2020-12-17
[GC] Add br_on_cast

kripken:
The tricky part here, as pointed out by @aheejin in my previous attempt, is that
we need to know the type of the value we send if the branch is taken. We can
normally calculate that from the `rtt` parameter's type - we are casting to that
RTT, so we know what type that is - but if the `rtt` is unreachable, that's a problem.
We could add `sent;` like we did on `br_on_exn`, but I realized that it's simpler to
just assume the sent type is the most generic one, `anyref`. So we always have a
valid type, and of course if the RTT is unreachable then the type doesn't matter
anyhow.

This includes a test with a `br_on_cast` that succeeds and sends the cast value,
one that fails and passes through the uncast value, and also of one with an
unreachable RTT. So I think this now handles all cases.

This includes a fix for Precompute, as noticed by that new test. If a break is
taken, with a ref as a value, we can't precompute it - for the same reasons
we can't precompute a ref in general, that it is a pointer to possibly shared
data.


aheejin:
If we assume all `block` that's targeted by `br_on_cast` is `anyref`, when we should be able to pass it as an argument to another function that expects the precise casted-down RTT type, we can't do that. Is this OK?

kripken:
I guess I'm not sure - maybe I'm missing something in the subtyping. If so then we are missing a validation check.

This is in the tests in this PR:
```wat
    (drop
      (block $out (result (ref $struct.B))
        (drop
          (br_on_cast $out $struct.B (ref.null $struct.A) (rtt.canon $struct.B))
        )
        ;; an untaken br_on_cast, with unreachable rtt - so we cannot use the
        ;; RTT in binaryen IR to find the cast type (binaryen IR does not store
        ;; the extra cast type that wasm does).
        (br_on_cast $out $struct.B (ref.null $struct.A) (unreachable))
        (unreachable)
      )
    )
```
After roundtripping to text, that last br becomes
```wat
    (br_on_cast $out any
     (ref.null ${i32_f32_f64})
     (unreachable)
    )
```
Should that not be valid?

aheejin:
```wasm
(func $foo (param (ref $struct.B))

(call $foo
  (block $out (result (ref $struct.B))
    (drop
      (br_on_cast $out $struct.B (ref.null $struct.A) (rtt.canon $struct.B))
    )
    (br_on_cast $out $struct.B (ref.null $struct.A) (unreachable))
    (unreachable)
  )
)
```

I think this should work, but if the `block`'s return type is `anyref`, it wouldn't.

I haven't yet read the code and I just read your PR description and asked; so I might be misunderstanding something.

aheejin:
```wasm
(func $foo (param (ref $struct.B))

(call $foo
  (block $out (result (ref $struct.B))
    (drop
      (br_on_cast $out $struct.B (ref.null $struct.A) (rtt.canon $struct.B))
    )
    (unreachable)
  )
)
```

I think this should work, but if the `block`'s return type is `anyref`, it wouldn't.

I haven't yet read the code and I just read your PR description and asked; so I might be misunderstanding something.

kripken:
@aheejin Ah yes, you're right... I thought what I did was correct, but it isn't. One thing that confused me is that my example seemed to validate - but not when doing `BINARYEN_PASS_DEBUG=1` which enables costly validation. And that wasn't run in the test I added. Adding testing to the test in passes, which runs the interpreter and also costly validation, and it indeed breaks...

Ok, we do need to add a field to `BrOnCast`, to store the proper type we are casting to. I added that now.

src/wasm.h
```diff
@@ -1342,13 +1344,24 @@ class RefCast : public SpecificExpression<Expression::RefCastId> {
...
 
 class BrOnCast : public SpecificExpression<Expression::BrOnCastId> {
 public:
   BrOnCast(MixedArena& allocator) {}
 
-  void finalize() { WASM_UNREACHABLE("TODO (gc): br_on_cast"); }
+  Name name;
+  // The cast type cannot be inferred from rtt if rtt is unreachable, so we must
+  // store it explicitly.
+  Type castType;
```
tlively, line 1358: Would it make more sense to store a HeapType here rather than a Type? That way this would match the binary format better and would not have to involve unnecessary mutability information. Same question for the return type of `getCastType`.

kripken: It could be either, in principle, yes. I tried both ways and this seems shorter in terms of the amount of code, as the other way you need to construct the Type from the HeapType more often. (Also, this matches the current BrOnExn which likewise has a Type, but I guess that's going away soon so maybe doesn't matter.)

src/wasm/wasm.cpp
```diff
@@ -1029,16 +1029,39 @@ void RefTest::finalize() {
   }
 }
 
+// Helper to get the cast type for a cast instruction. They all look at the rtt
+// operand's type.
+template<typename T> static Type doGetCastType(T* curr) {
+  if (curr->rtt->type == Type::unreachable) {
+    // We don't have the RTT type, so just assume the most generic type as
+    // possible.
+    return Type::anyref;
```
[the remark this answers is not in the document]
kripken: Good idea, yes, unreachable seems safer.  The type here should never matter - unlike BrOnCast, where it does - but still, yes, if we end up seeing it in debug info, or using it somehow in the future, unreachable seems much less risky.

tlively (approved):

test/heap-types.wast
```diff
@@ -167,5 +167,16 @@
...
+      (block $out (result (ref $struct.B))
+        (drop
+          (br_on_cast $out $struct.B (ref.null $struct.A) (rtt.canon $struct.B))
+        )
+        ;; an untaken br_on_cast, with unreachable rtt - so we cannot use the
+        ;; RTT in binaryen IR to find the cast type.
+        (br_on_cast $out $struct.B (ref.null $struct.A) (unreachable))
+        (unreachable)
+      )
+    )
```
aheejin, line 180: I'd be good to add the test that `block`'s return type that's targeted by `br_on_cast` is correct, as in https://github.com/WebAssembly/binaryen/pull/3451#issuecomment-747171175.

test/passes/Oz_fuzz-exec_all-features.wast
```diff
@@ -141,4 +141,42 @@
...
+  (drop
+   (block $block (result ($ref $struct))
+    (drop
+     (block $extendedblock (result (ref $extendedstruct))
+      (drop
+       ;; second, try to cast our simple $struct to what it is, which will work
+       (br_on_cast $block $struct
+        ;; first, try to cast our simple $struct to an extended, which will fail
+        (br_on_cast $extendedblock $extendedstruct
+         (local.get $any) (rtt.canon $extendedstruct)
```
aheejin, line 159: Isn't this `extendedstruct`...?

src/wasm/wasm-validator.cpp
```diff
@@ -2213,19 +2213,31 @@ void FunctionValidator::visitRefCast(RefCast* curr) {
...
+      curr->ref->type.isRef(), curr, "br_on_cast ref must have ref type");
+  }
+  if (curr->rtt->type != Type::unreachable) {
+    shouldBeTrue(
+      curr->rtt->type.isRtt(), curr, "br_on_cast rtt must have rtt type");
+    shouldBeEqual(curr->rtt->type.getHeapType(),
+                  curr->castType.getHeapType(),
+                  curr,
+                  "br_on_cast rtt must have the proper heap type");
+    noteBreak(curr->name, Type(curr->rtt->type.getHeapType(), Nullable), curr);
```
aheejin, line 2239:
```
noteBreak(curr->name, curr->castType, curr);
```

test/heap-types.wast
kripken: Sorry, I must be missing the point of that comment... I thought you were saying to check that we send the right type to the block. Isn't that what's tested on https://github.com/WebAssembly/binaryen/pull/3451/files#diff-02d11b1ea6faf71dddbf6ba74127bc9022481aeb27d74e423821c029cf814ad2R156 ?

test/passes/Oz_fuzz-exec_all-features.wast
kripken: Do you mean `$any`? It's assigned a `$struct` on line 148. So casting it to an `$extendedstruct` fails.

src/wasm/wasm-validator.cpp
kripken: Thanks! Followup in https://github.com/WebAssembly/binaryen/pull/3457

test/passes/Oz_fuzz-exec_all-features.wast
aheejin: Oh right, I was confused the heap type annotation with the argument.. Sorry.

test/heap-types.wast
aheejin: That value is dropped, so even if it is not the right type, it's not gonna be caught.. no?

kripken: Is your concern that the value that falls through doesn't have the proper type, which in that case should be  `anyref`? And that something other than a drop could notice that the type is not `anyref`?

aheejin: Yeah as in my example, can we pass the block as a function argument that expects just `ref $struct.B` and not `anyref` or `ref $struct.A`? (I think we can, but anyway, what I meant was a test for that)

kripken: I'm confused then, maybe because it's late...

Your example is
```wat
(func $foo (param (ref $struct.B))

(call $foo
  (block $out (result (ref $struct.B))
    (drop
      (br_on_cast $out $struct.B (ref.null $struct.A) (rtt.canon $struct.B))
    )
    (unreachable)
  )
)
```
I'm not sure what role the `call` plays here? The block having result `$struct.B` already is enough by itself to verify that we send that type to the block. That is, all branches to `$out` must have type `$struct.B` (or an even more refined type, but there isn't one in the program) or it won't validate? So I think that was tested already, unless I'm missing something.

aheejin: Oh, right, the validator checks that... I forgot about that sorry. Sorry for making a lot of noise at night. 😅

kripken: No problem!

And good we looked at this more carefully as the other test was indeed missing for the value flowing out (as the drop accepted any type).

PR #3452, opened 2020-12-17
finalize: there can only ever be a single initializer function. NFC.

sbc100:


src/wasm/wasm-emscripten.cpp
```diff
@@ -482,13 +482,9 @@ std::string EmscriptenGlueGenerator::generateEmscriptenMetadata(
...
 
-  if (!initializerFunctions.empty()) {
+  if (initializer.is()) {
     meta << "  \"initializers\": [";
-    commaFirst = true;
-    for (const auto& func : initializerFunctions) {
-      meta << nextElement();
-      meta << "\"" << func.c_str() << "\"";
-    }
+    meta << "\n    \"" << initializer.c_str() << "\"";
```
tlively, line 487: Do you plan to update the metadata format to make this not be a list any more as well?

tlively (approved):

src/wasm/wasm-emscripten.cpp
sbc100: TBH I think we can completely remove this .. but we need to remove it from the emscripten side first.

PR #3454, opened 2020-12-17
[wasm-split] Add an --initial-table option

tlively:
Adds an option to wasm-split to allow the user to specify the initial table size
for both instrumenting and splitting use cases. In the short term this will be
used in Emscripten SPLIT_MODULE + dynamic linking workflow to ensure that the
expected table size baked into the JS works for both the instrumented and the
primary split modules. In the longer term this may be replaced with a more
elegant mechanism for making the JS works in both cases.

cc @awtcode @sbc100

### DIFF
--- a/src/tools/wasm-split.cpp
+++ b/src/tools/wasm-split.cpp
@@ -66,6 +66,11 @@ struct WasmSplitOptions : ToolOptions {
   std::string placeholderNamespace;
   std::string exportPrefix;
 
+  // A hack to ensure the split and instrumented modules have the same table
+  // size when using Emscripten's SPLIT_MODULE mode with dynamic linking. TODO:
+  // Figure out a more elegant solution for that use case and remove this.
+  int initialTableSize = -1;
+
   WasmSplitOptions();
   bool validate();
   void parse(int argc, const char* argv[]);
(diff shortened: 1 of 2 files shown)


src/tools/wasm-split.cpp
```diff
@@ -453,6 +468,24 @@ uint64_t hashFile(const std::string& filename) {
...
 }
 
+void adjustTableSize(Module& wasm, const WasmSplitOptions& options) {
+  if (options.initialTableSize < 0) {
+    return;
+  }
+  if (!wasm.table.exists) {
+    Fatal() << "--initial-table used but there is no table";
+  }
+  if ((uint64_t)options.initialTableSize < wasm.table.initial) {
```
sbc100, line 478: Just pass options.initialTableSize as the second arg since not other options are used in this function?

sbc100 (approved):
I guess ideally this size would just be the max of the initial size used by either of these modules?    If we did that an modified emscripten to allow a dyanmic  initial table size would that obviate the need to this new option?

lgtm in the mean time..

tlively:
> I guess ideally this size would just be the max of the initial size used by either of these modules? If we did that an modified emscripten to allow a dyanmic initial table size would that obviate the need to this new option?

Right. Unfortunately changing Emscripten to allow a dynamic initial table size seems complicated, but perhaps it would seem simpler if I were more familiar with the code base.

PR #3456, opened 2020-12-17
Add Type::IsDefaultable and use that to do more validation

kripken:
The extra validation noticed that we need to do something for i31ref. For
refs in general we just make them nullable, but we can't do that for a built-in
type like i31ref. So treat it as nullable for now, until we handle nullability
properly.


src/wasm/wasm-type.cpp
```diff
@@ -393,6 +393,14 @@ bool Type::isException() const {
...
 bool Type::isNullable() const {
   if (isBasic()) {
+    if (id == i31ref) {
+      // i31ref is always non-nullable - there is no way to declare a nullable
+      // i31 reference, unlike other reference types. For now, allow it in vars
+      // for testing (as we have no other way to test it, yet), by reporting
+      // that it is nullable.
+      // FIXME when we have proper non-nullability.
+      return true;
+    }
```
aheejin, line 403: I thought primitive types are defaultable too, because they can be set to 0, but this does not allow it. [This](https://github.com/WebAssembly/binaryen/blob/2257f857069faa56335d2e24d7d6853c9501fcb7/src/wasm/wasm-validator.cpp#L2282) seems to imply all basic types are defaultable.

src/wasm/wasm-type.cpp
```diff
@@ -393,6 +393,14 @@ bool Type::isException() const {
 
 bool Type::isNullable() const {
   if (isBasic()) {
+    if (id == i31ref) {
+      // i31ref is always non-nullable - there is no way to declare a nullable
+      // i31 reference, unlike other reference types. For now, allow it in vars
```
tlively, line 398: Isn't `(ref null i31)` legal?

src/wasm/wasm-type.cpp
kripken: [The GC MVP doc says i31ref is not](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#reference-types)], at least unless I'm misreading it
```
anyref == (ref null any)

eqref == (ref null eq)

dataref == (ref data)

i31ref == (ref i31)
```
So apparently dataref and i31ref are non-nullable?

src/wasm/wasm-type.cpp
kripken: Good question... the binaryen parser doesn't allow it. But maybe we should, as it seems like `i31` is a heap type...?

In that case, we could just remove `i31ref` entirely, if it's just a wat shorthand for `(ref i31)`? That is, we could remove i31ref from the entire codebase, and whenever we need one, do `Type(HeapType::i31, NonNullable)`.

tlively: Well, I wouldn't go as far as removing `i31ref` from the codebase entirely, unless we also want to remove `anyref`, `externref`, etc. for the same reasons. But we should be using `(ref null i31)` rather than `i31ref` everywhere we are currently using e.g. `externref` in place of `(ref extern)`.

src/wasm/wasm-type.cpp
tlively: I wonder if `i31ref` is meant to default to 0? I assume `(ref null i31)` defaults to `null` regardless.

aheejin: Oh sorry nevermind, I thought this method was `isDefaultable`, but it was `isNullable`....

aheejin (approved):

src/wasm/wasm-type.cpp
dcodeIO: Not sure if it helps, but I got [this response](https://github.com/WebAssembly/gc/issues/124#issuecomment-680941832) when asking about `i31ref == (ref i31)` specifically.

> Having a different nullability default for the i31ref shorthand was a pragmatic choice based on expected use cases, but it is a bit irregular. We can switch it around if folks think that's preferable. Opinions?

Potentially related, `i31.new` has recently been promoted to a [constant instruction](https://github.com/WebAssembly/gc/pull/150/files#diff-1df1ea4af362cad5073bac12a65046e4c21bdaa644f69e203b41687b933ac91dR303), which may or may not indicate that `i31ref` can now default to `(i31.new (i31.const 0))`.

kripken: Thanks @dcodeIO , interesting.

Sounds like not everything is figured out in the spec about this, so we'll revisit as necessary...

src/wasm/wasm-type.cpp
kripken: @tlively I'm not sure I follow your last sentence. What is the analogy with externref saying?

My point maybe wasn't clear before. What I am saying is this: right now we have `i31ref`, and we cannot read `(ref null i31)` (with or without `null`). If I add the ability to read that, then now we have two internal representations of the same type, `Type::i31ref` and `Type(HeapType::BasicHeapType::i31, NonNullable)`.

The latter would not count as `Type::isBasic()` right now, so we'd need to fix that. I think we'd need to fix equality checking by making sure the two are interned in the same way, but I'm not sure. Instead, it seems simpler to me to have a single internal representation (the second of the two above), and to basically turn `i31ref` into sugar in the text format, which is immediately lowered in the s-parser. Thoughts?

(That all seems like annoying work, so I'd prefer if there's a better option!)

tlively: Due to canonicalization, `Type(HeapType::BasicHeapType::i31, NonNullable) == Type::i31ref`, just like `Type(HeapType::BasicHeapType::any, Nullable) == Type::anyref`. That means that `Type::isBasic()` returns `true` either way.

`i31ref`, `anyref`, `externref`, etc. are all just shorthands for `(ref i31)`, `(ref null any)`, `(ref null extern)`, etc. We could remove all these shorthands from the codebase, but it wouldn't make sense to remove just one of them.

I'm suggesting that we parse `i31ref` (i.e. `(ref i31)`) as `(ref null i31)` just like we currently parse `(ref any)` as `(ref null any)` (i.e. `anyref`).

kripken:
Ok, after recent heap type improvements this PR does not require any hacks for i31, and just adds `isDefaultable` and uses it.

src/wasm/wasm-type.cpp
```diff
@@ -411,6 +411,13 @@ bool Type::isStruct() const { return isRef() && getHeapType().isStruct(); }
 
 bool Type::isArray() const { return isRef() && getHeapType().isArray(); }
 
+bool Type::isDefaultable() const {
+  // A variable can get a default value if its type is concrete (unreachable
+  // and none have no values, hence no default), and if it's a reference, it
+  // must be nullable.
+  return isConcrete() && (!isRef() || isNullable());
```
tlively, line 418: Do RTTs have default values?

tlively (approved):

src/wasm/wasm-type.cpp
kripken: I think so, and binaryen gives them the natural default value (`rtt.canon`). I admit I can't find an explanation of that on the GC docs, though. I opened https://github.com/WebAssembly/gc/issues/175

PR #3457, opened 2020-12-18
Followup to #3451 after feedback

kripken:
As suggested, the validation code can be further simplified after adding `castType`.

### DIFF
--- a/test/heap-types.wast
+++ b/test/heap-types.wast
@@ -161,6 +161,7 @@
   (func $rtt-param-with-depth (param $rtt (rtt 1 $parent)))
   (func $rtt-param-without-depth (param $rtt (rtt $parent)))
   (func $rtt-operations
+    (local $temp.A (ref null $struct.A))
     (drop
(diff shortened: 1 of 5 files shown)


kripken:
Also added a test for the type of the value flowing out of `br_on_cast`.

test/heap-types.wast
```diff
@@ -169,7 +170,9 @@
     )
     (drop
       (block $out (result (ref $struct.B))
-        (drop
+        ;; set the value to a local with type $struct.A, showing that the value
+        ;; flowing out has the right type
+        (local.set $temp.A
```
aheejin, line 175: Sorry what I meant was not value flowing out but the target block's return type...

aheejin: Sorry I didn't know what I was saying is [already tested](https://github.com/WebAssembly/binaryen/pull/3451#discussion_r545553499) in the validator. Nevermind.

aheejin (approved):
Thanks!

PR #3458, opened 2020-12-18
MemoryPacking: Preserve segment names

sbc100:
Also avoid packing builtin llvm segments names so that
segments such as `__llvm_covfun` (use by llvm-cov) are
preserved in the final output.

### DIFF
--- a/src/passes/MemoryPacking.cpp
+++ b/src/passes/MemoryPacking.cpp
@@ -156,6 +156,7 @@ void MemoryPacking::run(PassRunner* runner, Module* module) {
     auto& currReferrers = referrers[origIndex];
 
     std::vector<Range> ranges;
+
     if (canSplit(segment, currReferrers)) {
       calculateRanges(segment, currReferrers, ranges);
     } else {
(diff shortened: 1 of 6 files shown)


sbc100:
Updated test.

src/passes/MemoryPacking.cpp
```diff
@@ -252,6 +253,10 @@ bool MemoryPacking::canOptimize(const Memory& memory,
 
 bool MemoryPacking::canSplit(const Memory::Segment& segment,
                              const Referrers& referrers) {
+  if (segment.name.is() && segment.name.startsWith("__llvm")) {
```
tlively, line 256: Can you add a comment about why we check this?

src/passes/MemoryPacking.cpp
```diff
@@ -528,7 +534,22 @@ void MemoryPacking::createSplitSegments(Builder& builder,
       range.end = lastNonzero->end;
       ranges.erase(ranges.begin() + i + 1, lastNonzero + 1);
     }
-    packed.emplace_back(segment.isPassive,
+    Name name;
+    if (segment.name.is()) {
+      // Name the first range after the original segment and all following
+      // ranged get numbered accordingly.  This means that that for segments
```
tlively, line 540:
```suggestion
      // ranges get numbered accordingly.  This means that that for segments
```

tlively (approved):

kripken:
Looks like this broke the roller, maybe due to increasing the size of the wasm file?

https://logs.chromium.org/logs/emscripten-releases/buildbucket/cr-buildbucket.appspot.com/8859135604284104048/+/steps/Emscripten_testsuite__upstream__wasm2_/0/stdout

PR #3464, opened 2020-12-31
Improve C and JS API module inspection features

dcodeIO:
* Changes element getters to return `NULL` if an element does not exist to also allow checking for presence, which wasn't supported yet and is more useful than aborting. Technically a breaking change with a CHANGELOG entry.
  * `BinaryenGetFunction`
  * `BinaryenGetGlobal`
  * `BinaryenGetEvent`
* Also adds `BinaryenGetExport` to obtain a reference to an export by external name. Returns `NULL` like its siblings if the export does not exist to allow checking for presence.
* Adds `BinaryenGetNumGlobals` and `BinaryenGetGlobalByIndex` to iterate over globals, just like already supported for functions and exports. This is useful when writing passes interested in all expressions (incl. global initializers), using just the C-API.
* Corrects the return type of `BinaryenGetNumFunctions` to `BinaryenIndex`, the type alias that iteration with `BinaryenGetFunctionByIndex` expects and similar functions use.
* Moves `BinaryenGetNumExports` and `BinaryenGetExportByIndex` to locations in binaryen-c.h/cpp that match locations of similar functions.
* Adds related descriptions of C API functions.

For those curious, we are going to use this for [custom Binaryen passes on our end](https://github.com/AssemblyScript/assemblyscript/blob/stw/src/passes/pass.ts), which may at some point make a fine binaryen.js feature (`class Pass`) as well if there is interest.


tlively (approved):
Good idea 👍

kripken (approved):

PR #3466, opened 2021-01-05
Prototype SIMD extending pairwise add instructions

tlively:
As proposed in https://github.com/WebAssembly/simd/pull/380, using the opcodes
used in LLVM and V8. Since these opcodes overlap with the opcodes of
i64x2.all_true and i64x2.any_true, which have long since been removed from the
SIMD proposal, this PR also removes those instructions.


kripken (approved):

sbc100:
This change caused some emscripten tests to fail:

```
$ ./tests/runner.py wasm2.test_wasm_builtin_simd
runner:WARNING: use EMTEST_ALL_ENGINES=1 in the env to run against all JS engines, which is slower but provides more coverage
Test suites:
['test_core']
Running test_core: (1 tests)
(checking sanity from test runner)
shared:INFO: (Emscripten: Running sanity checks)
test_wasm_builtin_simd (test_core.wasm2) ... [wasm-validator error in function 139] function body type must match, if function returns, on 
(i16x8.extadd_pairwise_i8x16_s
 (local.get $0)
)
[wasm-validator error in function 140] function body type must match, if function returns, on 
(i16x8.extadd_pairwise_i8x16_u
 (local.get $0)
)
Fatal: error validating input
em++: error: '/usr/local/google/home/sbc/dev/wasm/binaryen-out/bin/wasm-opt --strip-dwarf --post-emscripten --no-exit-runtime -O2 --low-memory-unused --zero-filled-memory --strip-debug --strip-producers test_wasm_builtin_simd.wasm -o test_wasm_builtin_simd.wasm --mvp-features --enable-simd' failed (1)
FAIL
```

Similar error running `test_wasm_intrinsics_simd`.

Sadly this didn't prevent this change from rolling into emscripten-releases (I guess because we don't run d8 tests on the emscripten-releases waterfall).

tlively:
Thanks for disabling the tests in the meantime!

PR #3467, opened 2021-01-06
Prototype prefetch instructions

tlively:
As proposed in https://github.com/WebAssembly/simd/pull/352, using the opcodes
used in the LLVM and V8 implementations.


src/ir/cost.h
```diff
@@ -537,6 +537,7 @@ struct CostAnalyzer : public OverriddenVisitor<CostAnalyzer, Index> {
   Index visitSIMDShuffle(SIMDShuffle* curr) {
     return 1 + visit(curr->left) + visit(curr->right);
   }
+  Index visitPrefetch(Prefetch* curr) { return 1 + visit(curr->ptr); }
```
kripken, line 540: I doubt it matters, but maybe this could be 0 instead of 1, as it is meant to speed up later things?

src/wasm/wasm-binary.cpp
```diff
@@ -5372,6 +5375,27 @@ bool WasmBinaryBuilder::maybeVisitSIMDLoadStoreLane(Expression*& out,
...
+    case BinaryConsts::PrefetchT:
+      op = PrefetchTemporal;
+      break;
+    case BinaryConsts::PrefetchNT:
+      op = PrefetchNontemporal;
+      break;
+    default:
+      return false;
+  }
+  auto* curr = allocator.alloc<Prefetch>();
```
kripken, line 5390: I think it's better to use `builder.makePrefetch`, that is, to leave the direct allocation to the Builder as much as possible, both here an in the s-parser code.

kripken (approved):

PR #3469, opened 2021-01-07
[GC] Fix parsing/printing of ref types using i31

kripken:
This lets us parse `(ref null i31)` and `(ref i31)` and not just `i31ref`.

It also fixes the parsing of `i31ref`, making it nullable for now, which we
need to do until we support non-nullability.

Fix some internal handling of i31 where we had just i31ref (which meant we
just handled the non-nullable type).

After fixing a bug in printing (where we didn't print out `(ref null i31)`
properly), I found some a simplification, to remove `TypeName`.

### DIFF
--- a/src/passes/Print.cpp
+++ b/src/passes/Print.cpp
@@ -79,10 +79,13 @@ static void
 printHeapTypeName(std::ostream& os, HeapType type, bool first = true);
 
 static void printTypeName(std::ostream& os, Type type) {
-  if (type.isBasic()) {
+  // If this is simple enough that it is not defined in the type section, just
+  // print it out.
+  if (type.isBasic() || (type.isRef() && type.getHeapType().isBasic())) {
     os << type;
     return;
   }
+  // This will be defined in the type section, emit the proper name.
   if (type.isRtt()) {
     auto rtt = type.getRtt();
     os << "rtt_";
--- a/src/passes/Print.cpp
+++ b/src/passes/Print.cpp
@@ -67,14 +67,6 @@ static std::ostream& printLocal(Index index, Function* func, std::ostream& o) {
   return printName(name, o);
 }
 
-// Wrapper for printing a type when we try to print the type name as much as
-// possible. For example, for a signature we will print the signature's name,
-// not its contents.
-struct TypeName {
-  Type type;
-  TypeName(Type type) : type(type) {}
-};
-
 static void
 printHeapTypeName(std::ostream& os, HeapType type, bool first = true);
 
@@ -208,10 +200,6 @@ static std::ostream& operator<<(std::ostream& o, const SExprType& sType) {
   return o;
 }
 
-std::ostream& operator<<(std::ostream& os, TypeName typeName) {
-  return os << SExprType(typeName.type);
-}
-
 // TODO: try to simplify or even remove this, as we may be able to do the same
 //       things with SExprType
 struct ResultTypeName {
@@ -229,10 +217,10 @@ std::ostream& operator<<(std::ostream& os, ResultTypeName typeName) {
     for (auto t : type) {
       os << sep;
       sep = " ";
-      os << TypeName(t);
+      os << SExprType(t);
     }
   } else {
-    os << TypeName(type);
+    os << SExprType(type);
   }
   os << ')';
   return os;
@@ -2546,7 +2534,7 @@ struct PrintSExpression : public OverriddenVisitor<PrintSExpression> {
       o << "(param ";
       auto sep = "";
       for (auto type : curr.params) {
-        o << sep << TypeName(type);
+        o << sep << SExprType(type);
         sep = " ";
       }
       o << ')';
@@ -2556,7 +2544,7 @@ struct PrintSExpression : public OverriddenVisitor<PrintSExpression> {
       o << "(result ";
       auto sep = "";
       for (auto type : curr.results) {
-        o << sep << TypeName(type);
+        o << sep << SExprType(type);
         sep = " ";
       }
       o << ')';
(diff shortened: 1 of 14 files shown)


src/passes/Print.cpp
```diff
@@ -174,8 +169,8 @@ struct SExprType {
   SExprType(Type type) : type(type){};
 };
 
-static std::ostream& operator<<(std::ostream& o, const SExprType& localType) {
-  Type type = localType.type;
+static std::ostream& operator<<(std::ostream& o, const SExprType& sType) {
```
aheejin, line 172: I'm not very sure what is this method exactly for anymore.. It seems it was added in #2675 to support multivalue printing for local types without commas, like `(i32 i64)`, but it has since grown and its name also changed from `LocalType` to `SExprType`... Can you help me on when we just use `printType` and when we use this method?

src/support/string.h
```diff
@@ -115,6 +115,18 @@ inline std::string trim(const std::string& input) {
...
+inline bool isNumber(const std::string& str) {
+  if (str.empty()) {
+    return false;
+  }
+  for (auto c : str) {
+    if (!std::isdigit(c)) {
+      return false;
+    }
+  }
+  return true;
```
aheejin, line 127:
```suggestion
  return !str.empty() && std::all_of(str.begin(), str.end(), ::isdigit);
```

src/wasm/wasm-binary.cpp
```diff
@@ -1416,7 +1416,8 @@ Type WasmBinaryBuilder::getType(int initial) {
       // FIXME: for now, force all inputs to be nullable
       return Type(getHeapType(), Nullable);
     case BinaryConsts::EncodedType::i31ref:
-      return Type::i31ref;
+      // FIXME: for now, force all inputs to be nullable
+      return Type(HeapType::BasicHeapType::i31, Nullable);
```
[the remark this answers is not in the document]
tlively: I'm not sure this would be any simpler because all the other code needs to be prepared to deal with either `(ref null i31)` or `(ref i31)`. Which version is considered "Basic" because it has the `Type::i31ref` shorthand should really only matter for printing, unless I'm missing something.

src/wasm/wasm-type.cpp
```diff
@@ -594,8 +600,9 @@ bool Type::isSubType(Type left, Type right) {
       return true;
     }
     // Various things are subtypes of eqref.
-    if ((left == Type::i31ref || left.getHeapType().isArray() ||
-         left.getHeapType().isStruct()) &&
+    auto leftHeap = left.getHeapType();
+    if ((leftHeap == HeapType::i31 || leftHeap.isArray() ||
+         leftHeap.isStruct()) &&
```
tlively, line 605: I assume `eqref` is short for `(ref null eq)` and that there is also a `(ref eq)`. In that case this code probably should compare with `rightHeap == HeapType::eq` rather than with `right == Type::eqref`. It will have to handle differing nullability between left and right as well.

src/wasm/wasm-binary.cpp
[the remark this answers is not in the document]
kripken: I think it is slightly simpler in the current approach, since this is more consistent with our other behavior, which is to turn all inputs immediately into nullable types. It's true that it looks a little more awkward here, since it's no longer "basic", but it is still applying that simple rule.

src/wasm/wasm-type.cpp
```diff
@@ -493,22 +493,28 @@ FeatureSet Type::getFeatures() const {
...
+          case HeapType::BasicHeapType::exn:
+            return FeatureSet::ReferenceTypes | FeatureSet::ExceptionHandling;
+          case HeapType::BasicHeapType::any:
+          case HeapType::BasicHeapType::eq:
+          case HeapType::BasicHeapType::i31:
+            return FeatureSet::ReferenceTypes | FeatureSet::GC;
+          default:
+            WASM_UNREACHABLE("unknown basic type");
+        }
+      }
```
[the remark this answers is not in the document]
kripken: Yes, good point. I pushed a commit with that simplification now.

src/passes/Print.cpp
kripken: Sorry, yes, it is a little confusing. I've added comments in this PR which might help. But probably we should rename things too.

* `printTypeName` prints a "serialized" name, as a list of characters. It has no spaces, in particular. We use this to emit a name for any type, no matter what it contains. Say an array contains `(ref eq)` then we'd print some serialization of that into `[ref_eq]` or such.
* SExprType prints out a type in normal s-format. For `(ref eq)` it would be exactly that (with spaces etc.). But that can recursively call `printTypeName` for the sub-types that it is declared from, when it needs their "serialized" name.

src/wasm/wasm-type.cpp
kripken: Thanks, yes. Pushed a fix now.

src/wasm/wasm-type.cpp
```diff
@@ -601,17 +597,19 @@ bool Type::isSubType(Type left, Type right) {
     }
     // Various things are subtypes of eqref.
     auto leftHeap = left.getHeapType();
+    auto rightHeap = right.getHeapType();
     if ((leftHeap == HeapType::i31 || leftHeap.isArray() ||
          leftHeap.isStruct()) &&
-        right == Type::eqref) {
+        rightHeap == HeapType::eq &&
+        (!left.isNullable() || right.isNullable())) {
```
tlively, line 604: Just noticed that the `right == Type::funcref` check below should probably be changed in the same way.

tlively (approved):

src/passes/Print.cpp
tlively: It would be good to standardize on either using `printXXX` functions or using wrapper structs with bespoke `operator<< implementations`, but I don't think that needs to happen here. (I prefer the latter fwiw.)

src/wasm/wasm-type.cpp
kripken: Good catch, fixed.

PR #3472, opened 2021-01-07
[GC] Fix minor binary format issues of ordering and a missing immediate

kripken:
The missing immediate is actually a point of disagreement between wasp and v8,
I opened https://github.com/WebAssembly/wasp/issues/52 about it. For now it seems
good to do what v8 does as it's also what's in the doc,
https://docs.google.com/document/d/1DklC3qVuOdLHSXB5UXghM_syCh-4cMinQ50ICiXnK3Q/edit?pli=1#

Noticed by comparing to V8 and Wasp. After this things are *almost* identical,
but there is also at least https://bugs.chromium.org/p/v8/issues/detail?id=11300

Test updates are due to having an instruction with two operands of which one is
unreachable. The new order puts the non-unreachable first, so it is not removed by
round-tripping through the binary format like before (which removes all unreachable
code).


tlively (approved):
It seems brittle to have to keep the order of children consistent in so many places. I wonder if we could use wasm-delegations-fields.h to automate parsing and emitting children in the correct order.

kripken:
Sorry, I had misread the doc - I read the main part, but missed the "notes" section, which changes a few things. This PR should now be final, and binaryen should match v8/wasp with it. Specifically, there are fewer extra heap things now, and no `n` in `rtt.sub`.

> It seems brittle to have to keep the order of children consistent in so many places. I wonder if we could use wasm-delegations-fields.h to automate parsing and emitting children in the correct order.

That's a little harder to do, sadly. We need to get those and pass them to wasm-builder functions. So we'd need to change how we do that, something like a builder just for immediates, that does the non-immediates automatically. Sounds like a good idea though!

src/wasm/wasm-binary.cpp
```diff
@@ -5709,13 +5709,14 @@ bool WasmBinaryBuilder::maybeVisitBrOnCast(Expression*& out, uint32_t code) {
     return false;
   }
   auto name = getBreakTarget(getU32LEB()).name;
-  auto heapType1 = getHeapType();
-  auto heapType2 = getHeapType();
+  // TODO the spec has two heaptype immediates, but the V8 prototype does not;
+  //      match that for now.
```
tlively, line 5713:
```suggestion
  //      match V8 for now.
```

src/wasm/wasm-binary.cpp
```diff
@@ -5709,13 +5709,14 @@ bool WasmBinaryBuilder::maybeVisitBrOnCast(Expression*& out, uint32_t code) {
...
   auto* rtt = popNonVoidExpression();
-  validateHeapTypeUsingChild(rtt, heapType2);
+  if (!rtt->type.isRtt()) {
+    throwError("bad rtt for br_on_cast");
+  }
```
tlively, line 5717: Does this mean we can't parse unreachable rtts in this position? That seems bad. Would we be able to synthesize an arbitrary heap type to pass to the builder in case the rtt is unreachable instead?

kripken: I think this is not a problem because we skip unreachable code anyhow (in `WasmBinaryBuilder::skipUnreachableCode`). That is, if the operand is unreachable, we never reach the rtt, unless I'm missing something. (If I am, the fuzzer will eventually correct us :wink: )

tlively (approved):

PR #3475, opened 2021-01-08
[GC] More HeapType instead of Type

kripken:
To handle both nullable and non-nullable i31s and other heap types, we cannot
just look at the isBasic case (which is just one of the two).

This *may* fix this issue on the release builder: https://github.com/WebAssembly/binaryen/runs/1669944081?check_suite_focus=true but the issue does not reproduce locally, so I worry it is something else...

### DIFF
--- a/src/wasm/literal.cpp
+++ b/src/wasm/literal.cpp
@@ -73,39 +73,54 @@ Literal::Literal(const Literal& other) : type(other.type) {
     } else {
       new (&exn) std::unique_ptr<ExceptionPackage>();
     }
-  } else if (other.isGCData()) {
+    return;
+  }
+  if (other.isGCData()) {
     new (&gcData) std::shared_ptr<GCData>(other.gcData);
-  } else if (type.isFunction()) {
+    return;
+  }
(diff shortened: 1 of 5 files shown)


src/wasm/literal.cpp
```diff
@@ -73,40 +73,55 @@ Literal::Literal(const Literal& other) : type(other.type) {
...
+    case Type::f64:
+      i64 = other.i64;
+      break;
+    case Type::v128:
+      memcpy(&v128, other.v128, 16);
+      break;
+    case Type::none:
+      break;
+    default:
+      WASM_UNREACHABLE("unexpected type");
```
tlively, line 123: I generally prefer not having `default` cases for these switches so the compiler can tell help us find all the places to update when we add a new type. Deleting all the BasicTypes for reference types would help make the class of error this PR fixes impossible and would help keep the switches from having too many useless cases. wdyt?

kripken: You're right, that is better. Done.

tlively (approved):

PR #3476, opened 2021-01-08
Disable threads in emscripten build

kripken:
The emscripten builds on CI here started to fail, e.g. https://github.com/WebAssembly/binaryen/runs/1669953244
```
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
```
Testing locally, it works for me - it actually doesn't seem to even check for pthreads. The only difference in setups that I can see is CI uses cmake 3.19 and I have 3.18.

In any case, emscripten doesn't use pthreads in binaryen atm, so maybe just disabling it makes sense. This PR tries that.

However, this seems to indicate a problem in upstream emscripten - @sbc100 did we change anything you can think of, in the last day or so?


kripken:
This appears to have been resolved in upstream emscripten somehow as new PRs pass that job. Curious if we know why @sbc100 ?

sbc100:
Hmm.. not sure.  I can't think of anything of the top of my head.  Maybe if we could see the cmake error/output file we could get a clue.

kripken:
Yeah, that's the problem... I can't reproduce it locally, and don't know of a way to ssh into the CI.

Anyhow, maybe no big deal. It could have been a temporary LLVM regression of some form on `tot`, that they've fixed meanwhile.

PR #3479, opened 2021-01-09
finalize: remove initializers from metadata output

sbc100:
See https://github.com/emscripten-core/emscripten/pull/13213

### DIFF
--- a/src/tools/wasm-emscripten-finalize.cpp
+++ b/src/tools/wasm-emscripten-finalize.cpp
@@ -230,8 +230,6 @@ int main(int argc, const char* argv[]) {
   generator.onlyI64DynCalls = onlyI64DynCalls;
   generator.noDynCalls = noDynCalls;
 
-  Name initializerFunction;
-
   if (!standaloneWasm) {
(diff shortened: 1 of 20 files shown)


kripken:
Don't we want to know if there is the initializer function, or not, or does LLVM always emit it, even when not needed?

(linked PR number looks wrong in the first comment?)

sbc100:
Fixed linked PR: https://github.com/emscripten-core/emscripten/pull/13213

As you can see emscripten no longer reads this key in the output metadata.

sbc100:
The code both here an emscripten treats the export as optional.   I'm can't remember if the linker ever actually removes it, but this code and the code in emscripten treat it as optional.

kripken (approved):
I see, sgtm.

We may want to add an optimization to remove the call from JS to an empty initializer, at some point.

sbc100:
> I see, sgtm.
> 
> We may want to add an optimization to remove the call from JS to an empty initializer, at some point.

It looks like this is already the case.  Testing locally shows that Oz and O3 both completely remove this function from the JS and from the wasm.

PR #3483, opened 2021-01-12
[GC] Fix casts of non-GC data

kripken:
The code previously assumed it could always call `getGCData`, but
that assumes the input is an array or a struct. It could also be an
anyref etc. that contains something other than GC data.

### DIFF
--- a/src/wasm-interpreter.h
+++ b/src/wasm-interpreter.h
@@ -1433,11 +1433,17 @@ public:
       return cast;
     }
     cast.originalRef = ref.getSingleValue();
-    auto gcData = cast.originalRef.getGCData();
-    if (!gcData) {
+    if (cast.originalRef.isNull()) {
       cast.outcome = cast.Null;
       return cast;
     }
+    // The input may not be a struct or an array; for example it could be an
+    // anyref of null (already handled above) or anything else (handled here).
+    if (!cast.originalRef.isGCData()) {
+      cast.outcome = cast.Failure;
+      return cast;
+    }
+    auto gcData = cast.originalRef.getGCData();
     auto refRtt = gcData->rtt;
     auto intendedRtt = rtt.getSingleValue();
     if (!refRtt.isSubRtt(intendedRtt)) {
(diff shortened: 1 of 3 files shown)


aheejin (approved):

src/wasm-interpreter.h
```diff
@@ -1433,11 +1433,17 @@ class ExpressionRunner : public OverriddenVisitor<SubType, Flow> {
...
     }
     cast.originalRef = ref.getSingleValue();
-    auto gcData = cast.originalRef.getGCData();
-    if (!gcData) {
+    if (cast.originalRef.isNull()) {
       cast.outcome = cast.Null;
       return cast;
     }
+    // The input may not be a struct or an array; for example it could be an
+    // anyref of null (already handled above) or anything else (handled here).
```
tlively, line 1441: How could the dynamic type of the value be `anyref` without being `ref.null any`? Is there any other way to materialize an `anyref` without a more specific type?

kripken: We can't create such a value in wast, but on the Web an anyref could contain a JS object for example.

tlively: Hmm, I'm a little unclear about whether the specific type of a JS object would be `anyref` or `externref`. I think the code is fine, but I would like to be more specific about what is covered by "anything else" in the comment.

tlively (approved):

src/wasm-interpreter.h
kripken: The visible type in the wat would be anyref, or externref, or eqref or such - something general. Internally of course the JS VM would have a JS-specific type. But internally in binaryen, right now at least we can't represent such a thing, so technically "anything else" is for future use in a sense - is that your concern?

tlively: I guess it's a question of the JS API. Would an exported function that takes a JS object as an argument via an `anyref` be able to downcast that `anyref` to an `externref`? If so, the dynamic (WebAssembly) type of that value would be `externef`, not `anyref` and there would be no need to be able to represent an `anyref` value in Binaryen.

tlively (approved):

src/wasm-interpreter.h
kripken: Ah, good question, I'm not sure what the plan is for the JS API there. In any case, this PR is not specific to anyref in any way? Seems like this code will be correct regardless.

PR #3485, opened 2021-01-13
[GC] Read and lower Let instructions

kripken:
For now we don't support non-nullability, and can therefore lower a let into
simpler things. That is,
```
(let $x = ...
  ;;
)

=>

(block
  $x = ...
  ;;
)
```

This lets us handle wasm binaries with let, so that we can optimize them
(with the current downside of losing non-nullability).

This is still not trivial to do, sadly, because the indexing of lets is somewhat
odd in the binary. A let modifies the indexes of other things declared before it,
which means that index "0" means different things at different times. And this
is trickier for us because we add more locals as needed for tuples and stacky
code. So this PR makes us track the absolute local indexes for each `let` as
we create it.

The binary testcase was created from this wat using wasp:
```wat
(module
  (type $vector (array (field (mut f64))))
  (func $main
    (local $x i32)
    (local $y i32)
    (drop (local.get $x)) ;; 0 is the index appearing in the binary
    ;; first let
    (array.new_with_rtt $vector
      (f64.const 3.14159)
      (i32.const 1)
      (rtt.canon $vector)
    )
    (let (local $v (ref $vector))
      (drop (local.get $v)) ;; 0
      (drop (local.get $x)) ;; 1
      ;; another one, nested
      (array.new_with_rtt $vector
        (f64.const 1234)
        (i32.const 2)
        (rtt.canon $vector)
      )
      (let (local $w (ref $vector))
        (drop (local.get $v)) ;; 1
        (drop (local.get $w)) ;; 0
        (drop (local.get $x)) ;; 2
      )
    )
    ;; another one, later
    (array.new_with_rtt $vector
      (f64.const 2.1828)
      (i32.const 3)
      (rtt.canon $vector)
    )
    (let (local $v (ref $vector))
      (drop (local.get $v)) ;; 0
      (drop (local.get $x)) ;; 1
    )
    (drop (local.get $x)) ;; 0
  )
)
```


src/wasm-binary.h
```diff
@@ -1367,6 +1369,21 @@ class WasmBinaryBuilder {
...
   std::vector<Expression*> expressionStack;
 
+  // Each let block in the binary adds new locals to the bottom of the index
+  // space. That is, all previously-existing indexes are bumped to higher
+  // indexes. getAbsoluteLocalIndex does this computation.
+  // Note that we must track not just the number of locals added in each let,
+  // but also their absolute indexes, as binaryen will add new locals as it goes
+  // for things like stacky code and tuples (so there isn't a simple way to get
+  // to the absolute index from a relative one). Hence each entry here is a list
+  // of the absolute indexes in use by that let.
```
aheejin, line 1379: I think I'm missing something, but I still don't understand why we should track all indices by value. Even if locals added for tuples and stuff in between, when we start a new `let` scope, all `let` vars within that scope are bumped by a certain constant amount. What I mean is, if a `let` has two variables `a` and `b`, the amount bumped for `a` and `b` will be the same because they are in the same `let`. Then can't we keep just a single value per `let` and not all indices..? Please correct me if I'm missing something.

kripken: That's correct, but it is more that previous lets are the issue. That is, lets can stack/nest, and we can add stacky locals in the middle. Consider this situation:

* We add a let for `$x` (just a made up name for this discussion - there is no name in the wasm binary, of course), giving it binaryen/absolute index 15.
* We add some locals for stacky code, using up indexes 16-30.
* We add a let for $y`, giving it binaryen/absolute index 31.
* We read a `local.get` with wasm/relative index 0. That's easy, it is `$y`, the last let that was "pushed". And it seems practical to compute 31 from a table of the pushed lets, I agree.
* We read a `local.get` with wasm/relative index 1. This is where it is harder. This refers to `$x`, but what is `$x`'s absolute index? We can't just subtract the lets that were pushed after it. If we just do that, we'd get 30, but that ignores the stacky locals. So we need to know about those.

And the messy part is that lets are stacked/nested, while stacky locals are not - they are added from there til the end. It is this mixture that makes it confusing to devise some sort of mechanism to track new locals. Instead, I think it's simpler to just remember the absolute indexes of all lets.

src/wasm/wasm-binary.cpp
```diff
@@ -3324,7 +3367,7 @@ void WasmBinaryBuilder::visitLocalGet(LocalGet* curr) {
   BYN_TRACE("zz node: LocalGet " << pos << std::endl);
   ;
```
tlively, line 3368: Stray `;` probably worth deleting here.

src/wasm/wasm-binary.cpp
```diff
@@ -5645,6 +5688,33 @@ void WasmBinaryBuilder::visitCallRef(CallRef* curr) {
...
+  Builder builder(wasm);
+  std::vector<Index> absoluteIndexes;
+  for (Index i = 0; i < numNewVars; i++) {
+    auto* value = popNonVoidExpression();
+    auto absoluteIndex = varsBefore + i;
+    absoluteIndexes.push_back(absoluteIndex);
+    curr->list.push_back(builder.makeLocalSet(absoluteIndex, value));
+  }
+  // Read the body, with adjusted local indexes.
+  letStack.push_back(absoluteIndexes);
```
tlively, line 5712: Might as well move the vector into place rather than copying it.

```suggestion
  letStack.emplace_back(std::move(absoluteIndexes));
```

tlively (approved):

src/wasm-binary.h
aheejin: I might be still mistaken, but I I thought is, rather than remembering absolute indices for all let vars, we can remember only two numbers (start offset and number of vars) per `let`. In your example, when we encounter the first `let` for `$x`, we push this pair to `letStack`:
```
letStack = [<15 (offset), 1 (# of vars in let)>]
```
And then we add locals for stacky code in 16-30.
And we encounter the second `let`, and we push one more entry to the `letStack`:
```
letStack = [<15 (offset), 1 (# of vars in let)>, <31 (offset), 1 (# of vars in let)>]
```

Now we read the binary index 0. Because it is within the range of the latest `let`, we can easily figure out it is the absolute index 31.
Then we read the binary index 1. Now it exceeds the size of the latest `let`, so we subtract the size of the latest `let` (1) from the index 1. Now we have the value 0. And we go to the second last entry, <15, 1>. And we know this binary index means the absolute index 15.

kripken: Thanks! Yes, that is more efficient and no more complicated, so definitely a better solution. Done.

aheejin (approved):
LGTM!

src/wasm/wasm-binary.cpp
```diff
@@ -3056,6 +3068,38 @@ BinaryConsts::ASTNodes WasmBinaryBuilder::readExpression(Expression*& curr) {
...
+  //
+  // Compute the relative index in the let we were added. We start by looking at
+  // the last let added, and if we belong to it, we are already relative to it.
+  // We will continue relativizing as we go down, til we find our let.
+  int64_t relative = index;
+  Index totalLetItems = 0;
+  for (auto i = int64_t(letStack.size()) - 1; i >= 0; i--) {
+    auto& info = letStack[i];
+    int64_t currNum = info.num;
+    totalLetItems += currNum;
```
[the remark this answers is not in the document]
kripken: Oh yes, good catch... it's really not needed any more.

PR #3486, opened 2021-01-13
Fix an index error in ArenaVectorBase::insertAt

aheejin:
Because `resize()` [sets](https://github.com/WebAssembly/binaryen/blob/dc7184afcbdcc72d0a6d66e2b36fc5857050dd87/src/mixed_arena.h#L204) `usedElements` to its argument, we were
accessing `data[usedElements]`, which can be outside of allocated memory
depending the internal state, i.e., `allocatedElements`'s value.

It is hard to come up with a test case for this because apparently the
failure condition depends on the vector's internal state.


dcodeIO (approved):
Ouch. Good find!

PR #3129, opened 2020-09-14
Fix inner block problem with 'catch'

aheejin:
Fixes #3114.

PR #3487, opened 2021-01-13
Basic EH instrucion support for the new spec

aheejin:
This updates `try`-`catch`-`catch_all` and `rethrow` instructions to
match the new spec. `delegate` is not included. Now `Try` contains not a
single `catchBody` expression but a vector of catch
bodies and events.

This updates most existing routines, optimizations, and tests modulo the
interpreter and the CFG traversal. Because the interpreter has not been
updated yet, the EH spec test is temporarily disabled in check.py. Also,
because the CFG traversal for EH is not yet updated, several EH tests in
`rse_all-features.wast`, which uses CFG traversal, are temporarily
commented out.

Also added a few more tests in existing EH test functions in
test/passes. In the previous spec, `catch` was catching all exceptions
so it was assumed that anything `try` body throws is caught by its
`catch`, but now we can assume the same only if there is a `catch_all`.
Newly added tests test cases when there is a `catch_all` and cases there
are only `catch`es separately.


src/ir/cost.h
```diff
@@ -546,7 +546,7 @@ struct CostAnalyzer : public OverriddenVisitor<CostAnalyzer, Index> {
   }
   Index visitTry(Try* curr) {
     // We assume no exception will be thrown in most cases
-    return visit(curr->body) + maybeVisit(curr->catchBody);
```
aheejin, line 549: Not sure why we were adding `catchBody` in costs, while the comment says we assume no exception will be thrown. Fixed to only count the try body.

src/wasm/wasm-stack.cpp
```diff
@@ -1859,22 +1859,33 @@ void BinaryInstWriter::visitTry(Try* curr) {
...
+    << U32LEB(parent.getEventIndex(curr->catchEvents[i]));
+}
+
+void BinaryInstWriter::emitCatchAll(Try* curr) {
+  assert(!breakStack.empty());
+  breakStack.pop_back();
+  breakStack.emplace_back(IMPOSSIBLE_CONTINUE);
+  if (func && !sourceMap) {
+    parent.writeExtraDebugLocation(curr, func, BinaryLocations::Catch);
+  }
```
aheejin, line 1879: I don't know much about debugging support, so I just copied the code for `emitCatch`, but is this right? Should we add `BinaryLocations::CatchAll` or something? Is it correct to print this info per `catch` clause, when there are multiple `catch`/`catch_all` clauses per `try`?

test/break-within-catch.wasm.fromBinary
```diff
@@ -1,14 +1,16 @@
 (module
  (type $none_=>_none (func))
```
aheejin, line 2: It's corresponding `break-within-catch.wasm` was this function in binary:
```wasm
try
  nop
catch $event$0
  drop
  br 0
end
```
This was originally added in #3129 to make `br 0` correctly target the `try`, not the inner block created within `catch`. Because currently `try` cannot have a label, #3129 wraps the `try` with a `block` and make the `br` target it instead. This PR only updates the binary file to the new format.

test/exception-handling.wast.fromBinary
```diff
@@ -15,48 +17,82 @@
...
+   )
+  )
+  (try
+   (do
+    (throw $event$2
+     (i32.const 0)
+     (i64.const 0)
+    )
+   )
+   (catch $event$2
```
aheejin, line 44: A simple `catch` with multivalue event in exception-handling.wast turns into this big code block after binary roundtripping. Is this expected? @tlively

aheejin:
Currently binaryen.js tests fail. They will pass after #3486 lands.

src/wasm/wasm-stack.cpp
kripken: Hmm, I think there is indeed a problem here. The current model of `writeExtraDebugLocation` is that an instruction has one or two more special locations, like an else or a catch. So each call with `::Catch` here overwrites the previous one.

I think we'll need to make `DelimiterLocations` an array. However, that would be nontrivial for a few reasons. For this PR, maybe just write TODOs for debug info for try-catch?

kripken (commented):
Initial set of comments.

Great to see this update!

src/wasm/wasm-s-parser.cpp
```diff
@@ -1966,15 +1973,38 @@ Expression* SExpressionWasmBuilder::makeTry(Element& s) {
...
   if (!elementStartsWith(*s[i], "do")) {
     throw ParseException(
       "try body should start with 'do'", s[i]->line, s[i]->col);
   }
-  ret->body = makeTryOrCatchBody(*s[i++], type, true);
-  if (!elementStartsWith(*s[i], "catch")) {
-    throw ParseException("catch clause does not exist", s[i]->line, s[i]->col);
+  ret->body = makeMaybeBlock(*s[i++], 1, type);
+
+  while (i < s.size() && elementStartsWith(*s[i], "catch")) {
```
tlively, line 1983: Shouldn't this check for equality rather than for a prefix? Same below for catch_all.

src/wasm/wasm-s-parser.cpp
```diff
@@ -1993,10 +2023,11 @@ SExpressionWasmBuilder::makeTryOrCatchBody(Element& s, Type type, bool isTry) {
   if (isTry && !elementStartsWith(s, "do")) {
     throw ParseException("invalid try do clause", s.line, s.col);
   }
-  if (!isTry && !elementStartsWith(s, "catch")) {
+  if (!isTry &&
+      !(elementStartsWith(s, "catch") || elementStartsWith(s, "catch_all"))) {
```
tlively, line 2027: nit: I think this would be at least as easy to read if you distributed the `!`.

src/wasm/wasm-validator.cpp
```diff
@@ -2026,20 +2026,35 @@ void FunctionValidator::visitTry(Try* curr) {
...
-    shouldBeEqual(curr->catchBody->type,
-                  Type(Type::unreachable),
+    for (auto catchBody : curr->catchBodies) {
+      shouldBeEqual(catchBody->type,
+                    Type(Type::unreachable),
+                    curr,
+                    "unreachable try-catch must have unreachable catch body");
+    }
+  }
+  if (curr->hasCatchAll()) {
```
tlively, line 2048: `hasCatchAll` only returns true if the difference is 1, so this check will never fail. Perhaps it could be replaced with an assertion?

src/wasm/wasm-validator.cpp
```diff
@@ -2084,11 +2099,9 @@ void FunctionValidator::visitRethrow(Rethrow* curr) {
...
                 curr,
                 "rethrow's type must be unreachable");
-  shouldBeSubTypeOrFirstIsUnreachable(
-    curr->exnref->type,
-    Type::exnref,
-    curr->exnref,
-    "rethrow's argument must be exnref type or its subtype");
+  // TODO Validate depth field if it becomes necessary in future. This requires
+  // a separate stack for 'catch', and the current LLVM toolchain only generates
+  // depth 0 for C++ support.
```
tlively, line 2104: Would it be worth validating that `depth` is 0 for now? Or do you plan to resolve this TODO in the near future?

test/binaryen.js/exception-handling.js
```diff
@@ -1,7 +1,8 @@
 function cleanInfo(info) {
   var ret = {};
   for (var x in info) {
-    if (x == 'id' || x == 'type' || x == 'name' || x == 'event') {
+    if (x == 'id' || x == 'type' || x == 'name' || x == 'event' ||
+        x == 'depth' || x == 'hasCatchAll') {
```
tlively, line 5: Is this filtering out all expression children because their values are not useful? It would be good to have a comment here.

test/exception-handling.wast.fromBinary
tlively: Yes, this is expected. Until I get Poppy IR landed, we will continue to emit terrible code for any multivalue construct and that code gets worse and worse each time it is parsed into Binaryen IR or emitted to a binary

test/passes/simplify-locals_all-features.wast
```diff
@@ -1698,29 +1698,15 @@
...
+  (func $pop-cannot-be-sinked (local $0 i32)
     (try
       (do)
-      (catch
-        ;; This (local.set $0) of (pop exnref) cannot be sinked to
-        ;; (local.get $0) below, because pop exnref should follow right after
-        ;; 'catch'.
-        (local.set $0 (pop exnref))
+      (catch $e-i32
+        ;; This (local.set $0) of (pop i32) cannot be sinked to (local.get $0)
```
tlively, line 1707:
```suggestion
        ;; This (local.set $0) of (pop i32) cannot be sunk to (local.get $0)
```

test/passes/simplify-locals_all-features.wast
```diff
@@ -1698,29 +1698,15 @@
     )
   )
 
-  (func $rethrow-trap (local $0 i32)
-    ;; This dead local.set cannot be replaced with a nop because rethrow can
-    ;; trap.
```
tlively, line 1703: Is this test no longer relevant because `rethrow` only trapped when given a null exnref?

tlively (commented):
Looks good overall! Just a few questions and small suggestions.

src/wasm/wasm-stack.cpp
aheejin: Done

src/wasm.h
```diff
@@ -1277,6 +1281,7 @@ class Rethrow : public SpecificExpression<Expression::RethrowId> {
   Rethrow(MixedArena& allocator) {}
 
   Expression* exnref;
```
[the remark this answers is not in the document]
aheejin: Oh, thanks. Done.

src/binaryen-c.cpp
```diff
@@ -1200,10 +1200,20 @@ BinaryenExpressionRef BinaryenRefEq(BinaryenModuleRef module,
...
+                                  BinaryenIndex numCatchBodies) {
+  auto* ret = ((Module*)module)->allocator.alloc<Try>();
+  ret->body = (Expression*)body;
+  for (BinaryenIndex i = 0; i < numCatchEvents; i++) {
+    ret->catchEvents.push_back(catchEvents[i]);
+  }
+  for (BinaryenIndex i = 0; i < numCatchBodies; i++) {
+    ret->catchBodies.push_back((Expression*)catchBodies[i]);
+  }
+  ret->finalize();
```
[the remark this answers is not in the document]
aheejin: Done

src/binaryen-c.h
```diff
@@ -1714,12 +1718,57 @@ BinaryenTryGetBody(BinaryenExpressionRef expr);
...
+// moving all subsequent catch bodies one index down. Returns the catch
+// expression.
 BINARYEN_API BinaryenExpressionRef
-BinaryenTryGetCatchBody(BinaryenExpressionRef expr);
-// Sets the catch body expression of a `try` expression.
-BINARYEN_API void BinaryenTrySetCatchBody(BinaryenExpressionRef expr,
-                                          BinaryenExpressionRef catchBodyExpr);
+BinaryenTryRemoveCatchBodyAt(BinaryenExpressionRef expr, BinaryenIndex index);
+// Gets whether an `try` expression has a catch_all clause.
+BINARYEN_API int BinaryenTryHasCatchAll(BinaryenExpressionRef expr);
```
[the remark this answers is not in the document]
aheejin: Yeah I agree.

src/cfg/cfg-traversal.h
```diff
@@ -304,11 +304,14 @@ struct CFGWalker : public ControlFlowWalker<SubType, VisitorType> {
...
       }
       case Expression::Id::TryId: {
+        // FIXME Update the implementation to match the new spec
+        /*
         self->pushTask(SubType::doEndTry, currp);
         self->pushTask(SubType::scan, &curr->cast<Try>()->catchBody);
         self->pushTask(SubType::doStartCatch, currp);
         self->pushTask(SubType::scan, &curr->cast<Try>()->body);
         self->pushTask(SubType::doStartTry, currp);
+        */
```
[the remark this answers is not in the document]
aheejin: Done

src/wasm-interpreter.h
```diff
@@ -2943,13 +2943,17 @@ template<typename GlobalManager, typename SubType> class ModuleInstanceBase {
...
+      // FIXME Update the implementation to match the new spec
+      /*
       NOTE_ENTER("Try");
       try {
         return this->visit(curr->body);
       } catch (const WasmException& e) {
         instance.multiValues.push_back(e.exn);
         return this->visit(curr->catchBody);
       }
+      */
```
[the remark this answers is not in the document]
aheejin: Done

src/ir/ExpressionAnalyzer.cpp
```diff
@@ -306,6 +307,10 @@ size_t ExpressionAnalyzer::hash(Expression* curr) {
 #define DELEGATE_FIELD_SIGNATURE(id, name) HASH_FIELD(name)
 
 #define DELEGATE_FIELD_NAME(id, name) visitNonScopeName(cast->name)
+#define DELEGATE_FIELD_NAME_VECTOR(id, name)                                   \
+  for (auto n : cast->name) {                                                  \
+    visitNonScopeName(n);                                                      \
+  }
```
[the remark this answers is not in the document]
aheejin: Thanks. Fixed wasm-delegations-fields.h so that `DELEGATE_FIELD_NAME_VECTOR` can use `GET_FIELD` and deleted this.

test/passes/vacuum_all-features.wast
```diff
@@ -804,29 +804,50 @@
...
-        (drop (pop exnref))
+      (catch $e
+        (drop (pop i32))
+      )
+    )
+  )
+
+  ;; The exception thrown in the inner try may not be caught by the inner catch
+  ;; (here it will be caught because they use the same event, but in general we
+  ;; cannot know), so the outer try-catch cannot be removed
```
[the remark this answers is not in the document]
aheejin: Done

src/passes/CodeFolding.cpp
```diff
@@ -306,10 +306,10 @@ struct CodeFolding : public WalkerPass<ControlFlowWalker<CodeFolding>> {
       }
       if (getModule()->features.hasExceptionHandling()) {
         EffectAnalyzer effects(getPassOptions(), getModule()->features, item);
-        // Currently pop instructions are only used for exnref.pop, which is a
-        // pseudo instruction following a catch. We cannot move expressions
-        // containing pops if they are not enclosed in a 'catch' body, because a
-        // pop instruction should follow right after 'catch'.
+        // Currently pop instructions are used only after 'catch', which is a
+        // pseudo instruction to simulate its behavior. We cannot move
```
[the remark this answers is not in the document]
aheejin: Done

src/wasm/wasm-s-parser.cpp
aheejin: Not sure what you mean by the 'prefix'. Do you mean I should check if the string is exactly `catch` and not things like `catchblah`? [`elementStartsWith`](https://github.com/WebAssembly/binaryen/blob/fcbefa951e3e514f6b7555664b89fb786850f6bc/src/wasm/wasm-s-parser.cpp#L65-L67) does not check string prefixes; it checks element prefixes. For example, `elementStartsWith(expr, "catch")` returns true for any `Element` in the form of `(catch ...)`.

src/wasm/wasm-s-parser.cpp
aheejin: Done

src/wasm/wasm-validator.cpp
aheejin: Oh, good catch. In the first version I maintained `hasCatchAll` a separate boolean variable within `Try` and later removed it because it basically contained redundant info, and then I mechanically replaced `hasCatchAll` with `hasCatchAll()`, which resulted in this.. I changed the code to this, WDYT?
```cpp
  shouldBeTrue(curr->catchBodies.size() - curr->catchEvents.size() <= 1,          
               curr,                                                              
               "the number of catch blocks and events do not match");
```

src/wasm/wasm-validator.cpp
aheejin: Validating this requires maintaining a separate stack within the validator, which is not difficult but I'm not very sure if it's worth it at the moment, so I don't have a plan to add that in near future, but if you think it's important we can change priorities. But at the same time I think rejecting values other than 0 can be risky, because binaryen is not guaranteed to take wasm code produced only by our LLVM toolchain.

test/binaryen.js/exception-handling.js
aheejin: Other values are basically address pointers, which can vary by execution. Other test cases within binaryen.js directory have this function too. (I added a few more fields here). Added a comment.

test/passes/simplify-locals_all-features.wast
aheejin: Yes.

src/js/binaryen.js-post.js
```diff
@@ -4189,12 +4192,97 @@ Module['Try'] = makeExpressionWrapper({
   'setBody'(expr, bodyExpr) {
     Module['_BinaryenTrySetBody'](expr, bodyExpr);
   },
-  'getCatchBody'(expr) {
-    return Module['_BinaryenTryGetCatchBody'](expr);
+  'getNumCatchEvents'(expr) {
+    return Module['_BinaryenTryGetNumCatchEvents'](expr);
+  },
+  'getCatchEvents'(expr) {
```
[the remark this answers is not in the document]
aheejin: I think we can, but I basically just copied these get/set methods from other classes, and I think it'd be better to do the refactoring on other classes too. Do you mind if I do that as a followup?

kripken (approved):
Thanks, looks like my comments are all addressed, and I don't see anything else, lgtm. Nice work!

Btw, about debug info for catches, I can look into that eventually, unless you want to.

src/wasm/wasm-validator.cpp
tlively: My thinking here is that it would be better to reject a valid program that uses non-zero depths than to accidentally and silently mis-optimize the program because we don't reason about non-zero depths right now. Or do we handle non-zero depths correctly in the optimization passes already?

src/wasm/wasm-validator.cpp
tlively: Looks good!

src/wasm/wasm-s-parser.cpp
tlively: Oops, I misunderstood what `elementStartsWith` was doing. Looks good!

src/js/binaryen.js-post.js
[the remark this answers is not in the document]
tlively: Yes, no problem.

tlively (approved):
The only unresolved conversation from my comments is whether or not to reject programs with nonzero rethrow depths for now, but I don't feel strongly about it either way. Whatever you think is best there is fine with me 👍

src/wasm/wasm-validator.cpp
aheejin: I don't think optimization passes will change the depth field, because they don't have any way of reasoning about that field. But yeah, there can be optimizations that remove or alter try-catch structure that are targeted by `rethrow`. I can't think of any optimization pass that does that right now though.

For one, `Vacuum` [removes](https://github.com/WebAssembly/binaryen/blob/f18c18e01d03d6d293fe3d701408855bbcea58bd/src/passes/Vacuum.cpp#L338-L346) try-catch altogether if the body of `try` does not throw. But this case is OK for `rethrow` because it removes the whole catch body. But there can be other passes that might interfere with `rethrow`, as you pointed out.

I'll reject non-zero depths for now for safe measure.

aheejin:
> Btw, about debug info for catches, I can look into that eventually, unless you want to.

Oh, I would appreciate that 😀

PR #3488, opened 2021-01-13
Avoid usedElements around code that modifies that value

kripken:
Followup to #3486, I wonder if it isn't a little more clear this way,
which avoids the confusion of `usedElements` being changed
while we are using it?

In general I think it's best to only use `usedElements` in the most
internal methods, and to call `size()` otherwise. But I don't feel
strongly.


dcodeIO (approved):
Makes it more clear to me :)

aheejin (approved):

Issue #2813, opened 2020-04-28
wasm-reduce unintentionally adds a feature


kripken:
This annoying thing happens:

* The reducer finds a bug. It uses no features while doing so on that testcase, plain MVP.
* `wasm-reduce` adds `-all` to all its commands. It does that so that it can succeed to load all wasm files.
* When it writes out the wasm after doing `-all`, it writes out the DataCount section, because `bulk-memory` has been enabled, and we always emit that section when that feature is enabled.
* When we try to run wasm2c on that testcase it now fails, as wasm2c only supports MVP features.

IOW, `wasm-reduce` uses `-all` to work around feature errors, but enabling features - even when just reading and writing the wasm - has a side effect of turning the output wasm into actually using features :cry: 

@tlively I hope you'll have a good idea here...

kripken:
Ah reading the code I see this promising comment:
```cpp
        // TODO(tlively): -all should be replaced with an option to use the
        // existing feature set, once implemented.
```
So maybe this is expected and we have a plan?

tlively:
Hmm, we have an option `--detect-features` that uses the target features section, but in general that section may not be present in the input. The simplest thing to do might be to add feature flags to `wasm-reduce`? I guess we could also add a pass or flag that detects used features and shrinks the enabled feature set to be minimal, but I've so far resisted doing that in favor of being more explicit and declarative.

PR #3492, opened 2021-01-15
wasm-reduce: default to -all, and make it customizable

kripken:
This goes back to the downsides of #2813, but that seems unavoidable
as without this, testcases without the features section but that use features
did not work.

This PR at least makes it easy to customize the flags send to the commands.


tlively (approved):
Nice, this seems reasonable and useful 👍

PR #3493, opened 2021-01-15
wasm-reduce: Fix setting of feature flags after loading

kripken:
We mistakenly did not set the flags to `all`, which meant that if the
features section was not present, we'd not have the proper features
set, leading to errors on writing.


src/tools/wasm-reduce.cpp
```diff
@@ -344,10 +344,14 @@ struct Reducer
...
-    Module wasm;
     ModuleReader reader;
     reader.read(working, *module);
-    wasm.features = FeatureSet::All;
+    // If there is no features section, assume we may need them all (without
+    // this, a module with no features section but that uses e.g. atomics and
+    // bulk memory would not work).
+    if (!module->hasFeaturesSection) {
+      module->features = FeatureSet::All;
+    }
```
tlively, line 354: Do we need this, given the default behavior in #3492?

kripken: Yes, these are separate fixes. That PR adds flags for the commands issued by the reducer (when it invokes wasm-opt to run entire sets of optimization passes). This PR is in the reducer's internals itself (where it does "destructive reduction" of replacing bits and pieces).

kripken:
Added some testing to here, which is now possible as the other PR landed.

tlively (approved):

PR #3495, opened 2021-01-15
Simplify set/gets of vectors in binaryen.js

aheejin:
This uses existing `getAllNested` function in `ExpressionWrapper`
functions. Also adds `setAllNested` which works in the other direction
and uses it within `ExpressionWrapper` functions.

### DIFF
--- a/src/js/binaryen.js-post.js
+++ b/src/js/binaryen.js-post.js
@@ -2492,6 +2492,25 @@ function getAllNested(ref, numFn, getFn) {
   return ret;
 }
 
+function setAllNested(ref, values, numFn, setFn, appendFn, removeFn) {
+  const num = values.length;
+  let prevNum = numFn(ref);
+  let index = 0;
+  while (index < num) {
+    preserveStack(() => {
+      if (index < prevNum) {
+        setFn(ref, index, values[index]);
+      } else {
+        appendFn(ref, values[index]);
+      }
+    });
+    ++index;
+  }
+  while (prevNum > index) {
+    removeFn(ref, --prevNum);
+  }
+}
+
 // Gets the specific id of an 'Expression'
 Module['getExpressionId'] = function(expr) {
   return Module['_BinaryenExpressionGetId'](expr);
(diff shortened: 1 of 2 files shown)


dcodeIO (approved):
Good idea!

src/js/binaryen.js-post.js
```diff
@@ -2492,6 +2492,25 @@ function getAllNested(ref, numFn, getFn) {
   return ret;
 }
 
+function setAllNested(ref, values, numFn, setFn, appendFn, removeFn) {
+  const num = values.length;
+  let prevNum = numFn(ref);
+  let index = 0;
+  while (index < num) {
+    preserveStack(() => {
```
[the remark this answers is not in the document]
aheejin: Yes, it was for that reason. It seems this is not necessary (but I guess harmless?) for expression vectors but necessary for string vectors. For example, `Switch`'s `setNames` is this:
```js
'setNames'(expr, names) {
  setAllNested(expr, names.map(strToStack), Module['_BinaryenSwitchGetNumNames'], Module['_BinaryenSwitchSetNameAt'], Module['_BinaryenSwitchAppendName'], Module['_BinaryenSwitchRemoveNameAt']);
}
```

The current `setNames` calls `preserveStack`:
https://github.com/WebAssembly/binaryen/blob/d808e900b8b8a0bc99d77fd38e94426df73c3afa/src/js/binaryen.js-post.js#L3348-L3355

Is there another way to preserve this semantics? Or, is calling `preserveStack` all the time not correct for expression vectors, such as block children?

src/js/binaryen.js-post.js
```diff
@@ -3333,31 +3333,10 @@ Module['Switch'] = makeExpressionWrapper({
...
-        } else {
-          Module['_BinaryenSwitchAppendName'](expr, strToStack(names[index]));
-        }
-      });
-      ++index;
-    }
-    while (prevNumNames > index) {
-      Module['_BinaryenSwitchRemoveNameAt'](expr, --prevNumNames);
-    }
+    setAllNested(expr, names.map(strToStack), Module['_BinaryenSwitchGetNumNames'], Module['_BinaryenSwitchSetNameAt'], Module['_BinaryenSwitchAppendName'], Module['_BinaryenSwitchRemoveNameAt']);
```
tlively, line 3339: Shouldn't the `preserveStack` call go around this rather than inside `setAllNested` so that it captures the `strToStack` calls?

src/js/binaryen.js-post.js
```diff
@@ -4196,31 +4137,10 @@ Module['Try'] = makeExpressionWrapper({
...
-        } else {
-          Module['_BinaryenTryAppendCatchEvent'](expr, strToStack(catchEvents[index]));
-        }
-      });
-      ++index;
-    }
-    while (prevNumCatchEvents > index) {
-      Module['_BinaryenTryRemoveCatchEventAt'](expr, --prevNumCatchEvents);
-    }
+    setAllNested(expr, catchEvents.map(strToStack), Module['_BinaryenTryGetNumCatchEvents'], Module['_BinaryenTrySetCatchEventAt'], Module['_BinaryenTryAppendCatchEvent'], Module['_BinaryenTryRemoveCatchEventAt']);
```
tlively, line 4143: Same here.

tlively (commented):
This looks great! I do think the `preserveStack` call needs to move, but I might be wrong.

src/js/binaryen.js-post.js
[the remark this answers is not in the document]
dcodeIO: Hmm, `names.map(strToStack)` executes, as an argument, before `setAllNested` is invoked with it. I guess this should work:

```js
'setNames'(expr, names) {
  preserveStack(() =>
    setAllNested(expr, names.map(strToStack), Module['_BinaryenSwitchGetNumNames'], Module['_BinaryenSwitchSetNameAt'], Module['_BinaryenSwitchAppendName'], Module['_BinaryenSwitchRemoveNameAt']);
  });
}
```

with an outer `preserveStack` around the code invoking `strToStack`, and the inner `preserveStack` within `setAllNested` safe to remove.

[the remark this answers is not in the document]
aheejin: Thanks! Done.

src/js/binaryen.js-post.js
aheejin: Thanks! Done.

src/js/binaryen.js-post.js
aheejin: Done

tlively (approved):

PR #3496, opened 2021-01-15
Debug info handling for new EH try-catch

kripken:
We now have multiple catches in each try, and a possible catch-all.

This changes our "extra delimiter" storage to store either an "else"
(unchanged from before) or an arbitrary list of things - we use that
for try-catch.

I'm not sure how to test this. We can't synthesize DWARF in binaryen
itself, so it would be easiest if we can emit a small testcase from clang
using the new EH stuff?


aheejin:
I tried to generate a wasm file with debug info. I attached it here. (Github doesn't support 'wasm' extension so I compressed it)
[test_eh_debuginfo.tar.gz](https://github.com/WebAssembly/binaryen/files/5826622/test_eh_debuginfo.tar.gz)

src/wasm/wasm-binary.cpp
```diff
@@ -2828,12 +2829,28 @@ BinaryConsts::ASTNodes WasmBinaryBuilder::readExpression(Expression*& curr) {
...
+      if (DWARF && currFunction) {
+        assert(!controlFlowStack.empty());
+        auto currControlFlow = controlFlowStack.back();
+        currFunction
+          ->delimiterLocations[currControlFlow][BinaryLocations::Else] =
+          startPos - codeSectionLocation;
+      }
       break;
-    case BinaryConsts::Catch:
+    case BinaryConsts::Catch: {
```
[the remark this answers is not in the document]
kripken: Thanks, yes, I think CatchAll is indeed needed here.

src/wasm/wasm-binary.cpp
```diff
@@ -3062,18 +3079,6 @@ void WasmBinaryBuilder::startControlFlow(Expression* curr) {
   }
 }
 
-void WasmBinaryBuilder::continueControlFlow(BinaryLocations::DelimiterId id,
-                                            BinaryLocation pos) {
-  if (DWARF && currFunction) {
-    assert(!controlFlowStack.empty());
-    auto currControlFlow = controlFlowStack.back();
-    // We are called after parsing the byte, so we need to subtract one to
-    // get its position.
```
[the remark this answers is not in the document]
kripken: Oh, I think the comment about subtracting one was just stale. A refactoring I did must have removed it, and forgot to update the comment...

src/wasm/wasm-binary.cpp
```diff
@@ -2828,12 +2829,29 @@ BinaryConsts::ASTNodes WasmBinaryBuilder::readExpression(Expression*& curr) {
...
+      if (DWARF && currFunction) {
+        assert(!controlFlowStack.empty());
+        auto currControlFlow = controlFlowStack.back();
+        currFunction
+          ->delimiterLocations[currControlFlow][BinaryLocations::Else] =
+          startPos - codeSectionLocation;
+      }
       break;
     case BinaryConsts::Catch:
+    case BinaryConsts::CatchAll: {
```
kripken, line 2841: @aheejin actually this fails to compile, since the binary code for CatchAll is 5, the same as for Else, so it says we have a duplicate entry in the switch.

Is it correct that CatchAll has the same code as Else?

kripken:
@aheejin Thanks for the testcase!

That showed something was missing here. I added code in `Print` to log out delimiters in try-catch. (We have also been missing it in if-else... I'll add that later to keep this small.) With that, looks like everything works properly. For example, the first catch
```
   ;; code offset: 0x18
   (catch $event$0
```
is the third line entry in the `debug_line` dwarfdump,
```
            0x0000000000000018      5      3      1   0             0  is_stmt
```
After the roundtrip, that line entry is
```
            0x0000000000000012      5      3      1   0             0  is_stmt
```
So we moved from `0x18` to `0x12`. And indeed the catch is printed as being at that binary offset,
```
   ;; code offset: 0x12
   (catch $event$0
```

src/wasm/wasm-binary.cpp
aheejin: Yes that's correct; we are doing that to reuse the opcode because `else` cannot occur with a `try`.

src/wasm/wasm-binary.cpp
```diff
@@ -2837,8 +2837,9 @@ BinaryConsts::ASTNodes WasmBinaryBuilder::readExpression(Expression*& curr) {
           startPos - codeSectionLocation;
       }
       break;
-    case BinaryConsts::Catch:
-    case BinaryConsts::CatchAll: {
+    case BinaryConsts::Catch: {
+      // TODO: add CatchAll, but the value of the constant is identical to Else
+      // case BinaryConsts::CatchAll: {
```
aheejin, line 2842: Maybe we can add `CatchAll` handling in `case BinaryConsts::Else:`? We will know if that is not an `Else` but a `CatchAll` if we encountered a `try` but not an `end` yet. But because control flow sturctures can be nested, we would need a stack to track this, something like I used as `procesStack` in #3494...?

kripken: Thanks, yeah, a stack seems unavoidable. Looks like we already have a control flow stack here, luckily, which I was able to use, PTAL.

aheejin (approved):
Thank you!

PR #3498, opened 2021-01-18
Update interpreter for new EH spec

aheejin:
This updates the interpreter for the EH instructions (modulo `delegate`)
to match the new spec. Before we had an `exnref` type so threw a
`Literal` of `exnref` type which contained `ExceptionPackage`. But now
that we don't have `exnref` anymore, so we add the contents of
`ExceptionPackage` to `WasmException`, which is used only for the
`ExpressionRunner` class hierarchy. `exnref` and `ExceptionPackage` will
be removed in a followup CL.

This allows nonzero depths for `rethrow` for now for testing; we
disallowed that for safety measure, but given that there are no passes
that modifies that field, I think the risk is low.

### DIFF
--- a/src/wasm-interpreter.h
+++ b/src/wasm-interpreter.h
@@ -41,9 +41,10 @@
 namespace wasm {
 
 struct WasmException {
-  WasmException(Literal exn) : exn(exn) {}
-  Literal exn;
+  Name event;
+  Literals values;
 };
+std::ostream& operator<<(std::ostream& o, const WasmException& exn);
 
 using namespace cashew;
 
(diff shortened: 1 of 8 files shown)


aheejin:
Currently the EH spec test fail on `wasm-opt -O` test due to the missing CFG traversal handling. After #3494 lands it will pass.

src/wasm-interpreter.h
```diff
@@ -1338,31 +1339,15 @@ class ExpressionRunner : public OverriddenVisitor<SubType, Flow> {
...
-    }
-    const auto& value = flow.getSingleValue();
-    if (value.isNull()) {
-      trap("rethrow: argument is null");
-    }
-    throwException(value);
-    WASM_UNREACHABLE("rethrow");
-    */
-  }
+  Flow visitRethrow(Rethrow* curr) { WASM_UNREACHABLE("unimp"); }
```
aheejin, line 1350: This was moved to `RuntimeExpressionRunner`, because not `rethrow` needs an access to the exception stack.

src/wasm-interpreter.h
```diff
@@ -2947,22 +2939,51 @@ template<typename GlobalManager, typename SubType> class ModuleInstanceBase {
...
+            instance.multiValues.push_back(e.values);
+            return processCatchBody(curr->catchBodies[i]);
+          }
+        }
+        if (curr->hasCatchAll()) {
+          return processCatchBody(curr->catchBodies.back());
+        }
+        // This exception is not caught by this try-catch. Rethrow it.
+        throw;
+        WASM_UNREACHABLE("try");
```
kripken, line 2974: is this WASM_UNREACHABLE still needed?

kripken (approved):
Nice, the rethrowing adds some complexity, but otherwise this is pretty simple and neat.

src/wasm-interpreter.h
aheejin: I guess not. Thanks. Removed.

aheejin:
By the way my latest commit title should be not "Remove the unnecessary `throw;`" but "Remove the unnecessary `WASM_UNREACHABLE`" 😂

tlively (approved):

PR #3503, opened 2021-01-21
Introduce a script for updating lit tests

tlively:
And demonstrate its capabilities by porting all tests of the
optimize-instructions pass to use lit and FileCheck.


test/lit/passes/optimize-instructions-atomics.wast
```diff
@@ -0,0 +1,33 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py
```
sbc100, line 1: Does this mean they should not be updated by hand?  Perhaps mention that explictly?

scripts/update_lit_checks.py
```diff
@@ -0,0 +1,171 @@
...
+
+script_name = os.path.basename(__file__)
+NOTICE = f';; NOTE: Assertions have been generated by {script_name}'
+RUN_LINE_RE = re.compile(r'^\s*;;\s*RUN:\s*(.*)$')
+CHECK_PREFIX_RE = re.compile(r'.*--check-prefix[= ](\S+).*')
+FUNC_RE = re.compile(r'^\s*\(func \$(\S*).*$', re.MULTILINE)
+
+
+def warn(msg):
+    print(f'WARNING: {msg}', file=sys.stderr)
```
sbc100, line 39: yay f-strings

scripts/update_lit_checks.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def itertests(args):
+    'Yield (filename, lines) for each test specified in the command line args'
```
sbc100, line 43: Triple quote docstrings and use double quotes for them.

scripts/update_lit_checks.py
```diff
@@ -0,0 +1,171 @@
...
+                warn(f'Skipping test {test} which was not generated by '
+                     f'{script_name}. Use -f to override.')
+                continue
+            yield test, lines
+
+
+def find_run_lines(test, lines):
+    matches = [m.group(1)
+               for m in [RUN_LINE_RE.match(l) for l in lines]
+               if m]
```
sbc100, line 63: I would personally split such double comprehensions into two for readability, but maybe this code is coming directly from the upstream version so it makes sense to preserve it?

scripts/update_lit_checks.py
```diff
@@ -0,0 +1,171 @@
...
+        else:
+            run_lines.append(line)
+    return run_lines
+
+
+def run_command(args, test, tmp, command):
+    env = dict(os.environ)
+    env['PATH'] = args.binaryen_bin + os.pathsep + env['PATH']
+    command = command.replace('%s', test)
+    command = command.replace('%t', tmp)
```
sbc100, line 80: Are these two substitutions really all the RUN lines can do?  Won't be be playing catchup with lit here are we re-implement the RUN -line mechanism here?

I guess the other thing RUN lines do is substitute the short command for the full command name, and you have chosen instead to inject the a PATH element..

scripts/update_lit_checks.py
```diff
@@ -0,0 +1,171 @@
...
+    return subprocess.check_output(command, shell=True, env=env).decode('utf-8')
+
+
+def find_funcs(module):
+    '''Return a dict mapping each function name to lines in the function'''
```
sbc100, line 85: use triple-double quotes

test/lit/passes/optimize-instructions-atomics.wast
```diff
@@ -0,0 +1,33 @@
...
+ ;; CHECK:    (i32.shl
+ ;; CHECK:     (i32.atomic.load8_u
+ ;; CHECK:      (i32.const 100)
+ ;; CHECK:     )
+ ;; CHECK:     (i32.const 24)
+ ;; CHECK:    )
+ ;; CHECK:    (i32.const 24)
+ ;; CHECK:   )
+ ;; CHECK:  )
+ ;; CHECK: )
```
sbc100, line 19: should we use CHECK_NEXT here?

sbc100 (commented):
Great!

Should this script be automatically run by `./auto_update_tests.py`?   Seems like it fits into that category.

I didn't look too closely and the fidelity with which you converted the tests.. but in general they look great!  @kripken should probably take a closer look.

scripts/update_lit_checks.py
tlively: No, the full list of substitutions is quite extensive (and extensible): https://llvm.org/docs/CommandGuide/lit.html#substitutions. Besides the tool names, which I covered by modifying PATH, these are the only substitutions I use regularly in tests, though. It shouldn't be too onerous to write any test that requires more esoteric substitutions by hand rather than using this script.

test/lit/passes/optimize-instructions.wast
```diff
@@ -0,0 +1,11557 @@
...
+(module
+  (memory 0)
+  (type $0 (func (param i32 i64)))
+ ;; CHECK: (func $f (param $i1 i32) (param $i2 i64)
+ ;; CHECK:  (drop
+ ;; CHECK:   (i32.and
+ ;; CHECK:    (local.get $i1)
+ ;; CHECK:    (i32.const 1)
+ ;; CHECK:   )
+ ;; CHECK:  )
```
kripken, line 13: should each of these be alongside the input? (and shouldn't they be right after it, not before?)

kripken (commented):
Please add the lit tests to `scripts/fuzz_opt.py`'s list of tests to pick testcases from (see `all_tests = core_tests + passes_tests + spec_tests + wasm2js_tests + lld_tests + unit_tests` there). I think it should just work, as it ignores the comments? (please run the fuzzer enough to see that it picks up some of them - can force it to by editing the code around `initial_contents`)

test/lit/passes/optimize-instructions.wast
tlively: Styles vary, but the style we use most in our WebAssembly tests is to have the filecheck assertions for a function immediately above the function, rather than below it. The script splits up the output by function because there is no reliable way to differentiate between different logical test cases within a function. Unfortunately this file has a lot of huge functions, but I have a WIP follow up to break them up a bit. For this PR I just copied each module in the original test file into a new lit test file without any further changes.

test/lit/passes/optimize-instructions-bulk-memory.wast
```diff
@@ -0,0 +1,162 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
+;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
+;; NOTE: Assertions have been generated by update_lit_checks.py
```
kripken, line 3: I'm confused by these messages.  The first appears twice. And then there is one without "and should not be edited" - are some files meant to be edited?

kripken (approved):
I don't have a strong sense for good practices with lit, but this lgtm assuming that.

test/lit/passes/optimize-instructions-bulk-memory.wast
tlively: Oops! There was previously a bug in the script where repeated invocations would add duplicates of this message (which also changed to say "and should not be edited" in the latest update. I got sidetracked half way through going through the file and deduplicating the message. I'll clean them up now.

PR #3508, opened 2021-01-22
Partially split up functions in optimize-instructions.wast

tlively:
FileCheck test cases are much more readable when their output assertions are
close to the corresponding input lines. #3503 introduced a script for updating
FileCheck assertions, but it only knows how to place input and output lines
together at the function boundary, so smaller functions will produce more
readable tests. optimize-instructions.wast has some very large functions that
include many test cases, so this PR breaks up some of those functions into
smaller functions to make the tests more readable. This PR only changes the
first quarter of the file, leaving further improvements as future work.

### DIFF
--- a/test/lit/passes/optimize-instructions.wast
+++ b/test/lit/passes/optimize-instructions.wast
@@ -4,37 +4,83 @@
 (module
   (memory 0)
   (type $0 (func (param i32 i64)))
-  ;; CHECK:      (func $f (param $i1 i32) (param $i2 i64)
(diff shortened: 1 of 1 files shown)


tlively:
I did delete a small hand full of test cases that seemed to be redundant with other test cases, but apart from that the only change here is to break up large functions into smaller functions.

kripken (approved):
Nice!

Was hard to see the actual diff here, but after some custom filtering I see that indeed the only removed testcases were indeed redundant.

Looks much better now!

PR #3509, opened 2021-01-22
Reorder i31ref and dataref

aheejin:
The binary spec
(https://docs.google.com/document/d/1yAWU3dbs8kUa_wcnnirDxUu9nEBsNfq0Xo90OWx6yuo/edit#)
lists `dataref` after `i31ref`, and `dataref` also comes after `i31ref`
in its binary code in the value-increasing order. This reorders these
two in wasm-type.h and other places, although in most of those places
the order is irrelevant.

This also adds C and JS API for `dataref`.


kripken (approved):
Nice!

PR #3516, opened 2021-01-25
[GC] RefIsNull => RefIs. NFC

kripken:
This internal refactoring prepares us for `ref.is_func/data/i31`, by renaming
the node and adding a "what" field. For now that field must always be "Null"
which means it is a `ref.is_null`.

This does not change the C or JS APIs. That keeps this NFC, and leaves
that for consideration later (more than one option there, not obvious what is
best).


src/tools/fuzzing.h
```diff
@@ -2946,7 +2946,7 @@ class TranslateToFuzzReader {
   Expression* makeRefIsNull(Type type) {
```
tlively, line 2946: Do you expect to change this function name eventually? If so, it might be worth a TODO.

src/wasm.h
```diff
@@ -1228,9 +1228,12 @@ class RefNull : public SpecificExpression<Expression::RefNullId> {
...
 };
 
-class RefIsNull : public SpecificExpression<Expression::RefIsNullId> {
+class RefIs : public SpecificExpression<Expression::RefIsId> {
 public:
-  RefIsNull(MixedArena& allocator) {}
+  RefIs(MixedArena& allocator) {}
+
+  // RefIs can represent ref.is_null, ref.is_func, ref.is_data, and ref.is_i31.
+  enum What { Null, Func, Data, I31 } what;
```
tlively, line 1236: Why not use a typical `...Op` enum for this?

kripken: Good question. In a way the others reflect the spec, and this reflects an internal detail of Binaryen IR - at least I think that's why it feels slightly better to me. Another is that this can't fit in the global namespace ("Null", "Func", etc.) so it's actually nice to write `RefIs::Null` as in this PR.

But I'm open to other ideas. If we use a normal op, maybe
```cpp
enum RefIsOp
  RefIsNull,
  RefIsFunc,
  RefIsData,
  RefIsI31,
};
```
?

tlively: That suggested `RefIsOp` enum looks good to me. I do actually like having the public member enums to get the namespacing syntax, though. What would you think of moving all the `...Op` enums to be children of the relevant expression classes as separate change?

aheejin (approved):

src/wasm.h
kripken: Ok, sounds good. You're right, it's better to be consistent here.

And yes, let's think about moving all the Op enums separately. That could be good.

kripken:
Ok, this should now be in its final form: using a standard Op enum, and with fixed C and JS APIs.

src/binaryen-c.h
```diff
@@ -1676,10 +1678,10 @@ BINARYEN_API void BinaryenMemoryFillSetSize(BinaryenExpressionRef expr,
 
 // Gets the value expression tested to be null of a `ref.is_null` expression.
 BINARYEN_API BinaryenExpressionRef
-BinaryenRefIsNullGetValue(BinaryenExpressionRef expr);
+BinaryenRefIsGetValue(BinaryenExpressionRef expr);
 // Sets the value expression tested to be null of a `ref.is_null` expression.
-BINARYEN_API void BinaryenRefIsNullSetValue(BinaryenExpressionRef expr,
-                                            BinaryenExpressionRef valueExpr);
+BINARYEN_API void BinaryenRefIsSetValue(BinaryenExpressionRef expr,
+                                        BinaryenExpressionRef valueExpr);
```
tlively, line 1684: We probably need new `GetOp` and `SetOp` functions, too.

tlively (approved):

src/binaryen-c.h
kripken: Right, yes. That wasn't blocking tests from passing, so I'm deferring it. My main priority is to get end-to-end support of GC prototype 3 finished, to unblock commandline tool usage first.

PR #3517, opened 2021-01-26
[reference-types] remove single table restriction in IR

### LINKED ISSUE #3512: Multiple tables support from the reference-types proposal
Hi, recently I've been curious about the reference-types proposal and I was trying to play around with it in various tools when I found out that `wasm-opt` doesn't like it when I give it a wasm file with multiple tables. So for the last 2-3 days I started digging through Binaryen's codes as well as some of the communications happening on this Github repo, trying to understand the situation.

So far, I found PR #2451 which added support for reference-types, but left multiple tables support and table instructions out. I also checked PR #3084 quickly, that brought the existing implementation up to date with recent changes from the proposal.

But nowhere can I find any mention of the missing support for multiple tables, which makes me feel like I am missing something: Has it been a non-goal for Binaryen to support this, or somehow the cost hasn't been worth the effort to fully implement the remaining parts, or it's just been low on priority. This is just me being curious.

I also went through the codebase, played around with the idea of replacing the single `table` object in Binaryen's Wasm module IR, with a `std::vector<std::unique_ptr<Table>>` just like the other module elements. With some help I will be able to publish it as a PR, potentially preparing the stage for full implementation of the proposal in the future.

So I just wanted to check whether I should continue this work, or there are important notes that I'm missing. I'd also appreciate any insight or directions.

martianboy:
This will remove the single table per module assumption everywhere, and
instead introduces a vector of tables, same as functions, globals, etc.
Support for parsing and writing multiple tables is added and C/JS APIs
have been updated accordingly. A table name argument is also added to
call_indirect.

(Fixes #3512)


martianboy:
Reviews appreciated @kripken @aheejin @sbc100

src/binaryen-c.cpp
```diff
@@ -1605,6 +1605,14 @@ void BinaryenCallIndirectSetTarget(BinaryenExpressionRef expr,
...
   static_cast<CallIndirect*>(expression)->target = (Expression*)targetExpr;
 }
+void BinaryenCallIndirectSetTableName(BinaryenExpressionRef expr,
+                                      const char* tableName) {
+  Name name(tableName);
+  auto* expression = (Expression*)expr;
+
+  assert(expression->is<CallIndirect>());
+  static_cast<CallIndirect*>(expression)->tableName = name;
+}
```
aheejin, line 1615: It would be good to have `GetTableName` as a counterpart, as we have setter/getter pairs for other fields

aheejin (commented):
Thank you! Haven't read all the PR yet, but the very initial comments:
I don't think we have a test case that uses multiple tables and `call_indirect`s with a table index yet. Can we have one? Maybe multi-table.wast or something in test/ directory. You can run ./auto_update_tests.py to update other files after adding it.
(Also we would need tests for other added parts such as tests for the added C API)

src/wasm/wasm-binary.cpp
```diff
@@ -2411,29 +2455,26 @@ void WasmBinaryBuilder::readDataSegments() {
...
-  if (indexType == Type::i64) {
-    throwError("Tables may not be 64-bit");
+
+  for (size_t i = 0; i < numTables; i++) {
+    auto elemType = getS32LEB();
+    if (elemType != BinaryConsts::EncodedType::funcref) {
+      throwError("ElementType must be funcref in MVP");
+    }
+    auto table = std::make_unique<Table>();
+    table->setName("table$" + std::to_string(i), false);
```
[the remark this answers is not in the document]
martianboy: Sure! Done.

src/binaryen-c.cpp
martianboy: Thanks, I will add this. But in general I wasn't sure about the changes required to the C API. Like, we have a `BinaryenSetFunctionTable` which is based on the old assumption of having a single table. Also, should we have an overloaded `BinaryenCallIndirect` that takes the table name as well? I think we should.

martianboy:
Thanks, @aheejin.

> I don't think we have a test case that uses multiple tables and `call_indirect`s with a table index yet. Can we have one? Maybe multi-table.wast or something in test/ directory.

Right, I started adding tests today. Hopefully I can push some by tomorrow.

martianboy:
@aheejin 

> I don't think we have a test case that uses multiple tables and `call_indirect`s with a table index yet. Can we have one? Maybe multi-table.wast or something in test/ directory. You can run ./auto_update_tests.py to update other files after adding it.

I don't think we can test modules with multiple tables yet, because in this PR I tried to keep it minimal and not implement the new changes from the spec as much as possible. For instance, we cannot parse an `elem` section in wast files that belong to tables with higher indexes. It does parse elems embedded inside a `table` declaration, but roundtrip runs fail as the elem section is separated from the table, and is therefore added to table 0. I did add a test/example for the new c apis though.

My plan was to get this PR reviewed and merged, to make sure that everything is working as before and I haven't missed anything. And then, continue with the remaining parts of the reference types spec.

src/wasm.h
```diff
@@ -821,6 +821,7 @@ class CallIndirect : public SpecificExpression<Expression::CallIndirectId> {
   Signature sig;
   ExpressionList operands;
   Expression* target;
+  Name tableName;
```
kripken, line 824:
```suggestion
  Name table;
```
I think that's more consistent with other terminology.

kripken (commented):
I didn't review the entire thing yet, but the wasm.h and wasm-delegates-fields.h changes look like the right approach. Thanks for the PR!

Will try to review the rest tomorrow.

src/wasm.h
martianboy: Thanks. Done.

test/example/module-splitting.txt
```diff
@@ -283,10 +295,13 @@ Before:
 Keeping: <none>
 After:
 (module
+ (table $0 0 funcref)
+ (export "%table" (table $0))
```
aheejin, line 299: Why are we adding tables to every wasm-split outputs even when they are not necessary?

aheejin (commented):
> @aheejin
> 
> > I don't think we have a test case that uses multiple tables and `call_indirect`s with a table index yet. Can we have one? Maybe multi-table.wast or something in test/ directory. You can run ./auto_update_tests.py to update other files after adding it.
> 
> I don't think we can test modules with multiple tables yet, because in this PR I tried to keep it minimal and not implement the new changes from the spec as much as possible. For instance, we cannot parse an `elem` section in wast files that belong to tables with higher indexes. It does parse elems embedded inside a `table` declaration, but roundtrip runs fail as the elem section is separated from the table, and is therefore added to table 0. I did add a test/example for the new c apis though.
> 
> My plan was to get this PR reviewed and merged, to make sure that everything is working as before and I haven't missed anything. And then, continue with the remaining parts of the reference types spec.

Thanks. Sorry I was OOO and I have still read only a fraction of the PR so far; I think this PR is one of the fundamental changes, but the only test additions + changes I see are from metrics, wasm-split (which I don't understand well why they are necessary; I asked questions there), and some table name changes.

I understand it is sometimes hard to add tests for every single part of code changes, especially in this kind of big PRs that change one of the very fundamental assumptions in the code base, so I really appreciate your work! But I still think this kind of big PRs need at least some tests that cover the very backbone of changes, which I think should include wast reading/writing and binary reading/writing. Also I think it is more desirable for a PR to have a smaller change with tests that matches the change, but as I said, I know it is not always easy in this kind of PRs.. 😅

I think maybe we can defer implementations + tests for the new `call_indirect` for the next PR, but I still would like to see at least some tests that show basic reading/writing of multiple tables working. Can we add those parts to the PR and maybe defer some other parts of the implementation that are not essential to the reading/writing to the next PR? I think C-API also could have been deferred if you wanted to make this minimal but you already added them so that looks good.

Anyway I'll probably take a more close look within a few days but please don't think me as a blocker for LGTM as long as others are happy with it.

test/passes/remove-unused-module-elements_all-features.txt
```diff
@@ -87,7 +87,7 @@
  (import "env" "memory" (memory $0 256))
  (import "env" "table" (table $timport$0 1 funcref))
  (export "mem" (memory $0))
- (export "tab" (table $0))
+ (export "tab" (table $timport$0))
```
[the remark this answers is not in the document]
martianboy: I believe the previous result has been incorrect. A table `$timport$0` is imported 3 lines above, but then an undefined table `$0` is exported on line 90.

test/example/module-splitting.txt
martianboy: At first I was a bit confused with the module-splitter, as it relies on the existence of the table object from the very beginning. So I just gave it the table for now. But I will try to fix this, so it only adds a table to the module when it needs to.

src/wasm/wasm-binary.cpp
```diff
@@ -1694,19 +1732,20 @@ void WasmBinaryBuilder::readImports() {
       }
       case ExternalKind::Table: {
         Name name(std::string("timport$") + std::to_string(tableCounter++));
-        wasm.table.module = module;
-        wasm.table.base = base;
-        wasm.table.name = name;
+        auto table = std::make_unique<Table>();
+        table->module = module;
+        table->base = base;
+        table->name = name;
```
tlively, line 1738: We should probably add a `makeTable` method to wasm-builder.h to cover this kind of usage.

src/wasm/wasm-binary.cpp
```diff
@@ -2414,29 +2474,26 @@ void WasmBinaryBuilder::readDataSegments() {
...
-  if (is_shared) {
-    throwError("Tables may not be shared");
-  }
-  if (indexType == Type::i64) {
-    throwError("Tables may not be 64-bit");
+
+  for (size_t i = 0; i < numTables; i++) {
+    auto elemType = getS32LEB();
+    if (elemType != BinaryConsts::EncodedType::funcref) {
+      throwError("ElementType must be funcref in MVP");
```
tlively, line 2481:
```suggestion
      throwError("Non-funcref tables not yet supported");
```

src/wasm/wasm-binary.cpp
```diff
@@ -2447,13 +2504,23 @@ void WasmBinaryBuilder::readTableElements() {
     throwError("Too many segments");
   }
   for (size_t i = 0; i < numSegments; i++) {
-    auto tableIndex = getU32LEB();
-    if (tableIndex != 0) {
+    auto headerFormat = getU32LEB();
+    if (headerFormat != 0) {
       throwError("Table elements must refer to table 0 in MVP");
     }
```
tlively, line 2510: This seems like an important restriction to lift in this PR.

src/wasm/wasm-emscripten.cpp
```diff
@@ -479,7 +479,11 @@ std::string EmscriptenGlueGenerator::generateEmscriptenMetadata() {
     meta << "\n  },\n";
   }
 
-  meta << "  \"tableSize\": " << wasm.table.initial.addr << ",\n";
+  if (!wasm.tables.empty()) {
+    meta << "  \"tableSize\": " << wasm.tables[0]->initial.addr << ",\n";
+  } else {
+    meta << "  \"tableSize\": 0,\n";
+  }
```
tlively, line 486: @kripken would it make more sense to error out here for now when there are multiple tables, or would this metadata item only care about table 0 anyway?

src/wasm/wasm-s-parser.cpp
```diff
@@ -2452,17 +2468,22 @@ void SExpressionWasmBuilder::parseExport(Element& s) {
   ex->name = s[1]->str();
   if (s[2]->isList()) {
     auto& inner = *s[2];
-    ex->value = inner[1]->str();
+
     if (elementStartsWith(inner, FUNC)) {
+      ex->value = getFunctionName(*inner[1]);
```
tlively, line 2473: Why these changes to non-table exports?

src/binaryen-c.h
```diff
@@ -1041,6 +1041,12 @@ BinaryenCallIndirectGetTarget(BinaryenExpressionRef expr);
 BINARYEN_API void
 BinaryenCallIndirectSetTarget(BinaryenExpressionRef expr,
                               BinaryenExpressionRef targetExpr);
+// Gets the table name of a `call_indirect` expression.
+BINARYEN_API const char*
+BinaryenCallIndirectGetTableName(BinaryenExpressionRef expr);
+// Sets the table name of a `call_indirect` expression.
+BINARYEN_API void BinaryenCallIndirectSetTableName(BinaryenExpressionRef expr,
+                                                   const char* table);
```
tlively, line 1049: I think it would be reasonable to require a table name in the CallIndirect constructor function as well, even though that would be a breaking change. We would just have to mention it in the change log.

src/binaryen-c.cpp
tlively: I think that changing the C API in whatever way makes sense is fine as long as the changes are documented in the changelog.

test/example/module-splitting.txt
martianboy: Ok I was able to fix this, although it got hacky. Totally open to better suggestions.

src/ir/module-splitting.cpp
```diff
@@ -196,13 +215,13 @@ TableSlotManager::Slot TableSlotManager::getSlot(Name func) {
 
   // If there are no segments yet, allocate one.
   if (activeSegment == nullptr) {
```
tlively, line 201: This would be the best place to lazily allocate a new table if one does not already exist.

src/ir/module-splitting.cpp
```diff
@@ -178,13 +197,13 @@ TableSlotManager::TableSlotManager(Module& module)
...
-        activeBase = {"", segmentBase};
+        activeBase = {table.name, "", segmentBase};
       }
     }
   }
   // Initialize funcIndices with the functions already in the table.
   forEachElement(table, [&](Name base, Index offset, Name func) {
-    addSlot(func, {base, offset});
+    addSlot(func, {table.name, base, offset});
   });
```
tlively, line 191: This needs to run for each table in the module. Probably the best way to do that would be to update `forEachElement` to run over each table as well. I think all the places that use `forEachElement` will need to visit every table.

Alternatively you can just assert that there is not more than one table and not implement module splitting for multiple tables right now.

test/example/c-api-multiple-tables.txt
```diff
@@ -0,0 +1,18 @@
+(module
+ (type $i32_i32_=>_i32 (func (param i32 i32) (result i32)))
+ (table $tab 1 1 funcref)
```
tlively, line 3: Is this test supposed to have multiple tables?

tlively (commented):
It looks like a lot of code could be simplified by requiring `call_indirect` to have a real table name rather than defaulting to the first table if the table name is empty.

src/binaryen-c.h
martianboy: This was my initial approach. But then I changed it to avoid a breaking change. I agree that the assumption makes a lot of things much easier!

src/ir/module-splitting.cpp
tlively: More specifically, I would mirror the `activeSegment` mechanism with an `activeTable` mechanism that works the same way rather than using `auxiliaryTable`.

martianboy:
@tlively Thanks a lot for the review! I will try to address the points by tomorrow.

test/example/c-api-kitchen-sink.c
```diff
@@ -839,6 +841,8 @@ void test_core() {
 
 void test_unreachable() {
   BinaryenModuleRef module = BinaryenModuleCreate();
+  const char* names[1] = {"unreachable-fn"};
+
```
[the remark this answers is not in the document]
martianboy: removed.

src/wasm/wasm-validator.cpp
```diff
@@ -2756,21 +2772,27 @@ static void validateMemory(Module& module, ValidationInfo& info) {
 }
 
 static void validateTable(Module& module, ValidationInfo& info) {
```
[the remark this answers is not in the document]
martianboy: Done.

src/passes/Directize.cpp
```diff
@@ -36,25 +36,36 @@ namespace {
...
+    : tables(tables) {}
 
   void visitCallIndirect(CallIndirect* curr) {
-    if (!flatTable) {
+    std::unordered_map<Name, TableUtils::FlatTable>::const_iterator it;
+    if (curr->table.is()) {
+      it = tables.find(curr->table);
+    } else {
+      it = tables.begin();
+    }
```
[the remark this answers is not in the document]
martianboy: Done.

src/passes/ExtractFunction.cpp
```diff
@@ -46,7 +46,7 @@ struct ExtractFunction : public Pass {
     }
     // clear data
     module->memory.segments.clear();
-    module->table.segments.clear();
+    module->tables.clear();
```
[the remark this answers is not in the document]
martianboy: Done.

src/passes/LegalizeJSInterface.cpp
```diff
@@ -97,10 +97,12 @@ struct LegalizeJSInterface : public Pass {
         // we need to use the legalized version in the table, as the import from
```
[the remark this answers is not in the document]
martianboy: Done.

src/passes/RemoveUnusedModuleElements.cpp
```diff
@@ -84,7 +91,22 @@ struct ReachabilityAnalyzer : public PostWalker<ReachabilityAnalyzer> {
       queue.emplace_back(ModuleElementKind::Function, curr->target);
     }
   }
-  void visitCallIndirect(CallIndirect* curr) { usesTable = true; }
+  void visitCallIndirect(CallIndirect* curr) {
+    if (module->tables.empty()) {
+      return;
+    }
```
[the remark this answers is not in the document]
martianboy: Replaced with an `assert`.

src/shell-interface.h
```diff
@@ -154,11 +159,27 @@ struct ShellExternalInterface : ModuleInstance::ExternalInterface {
...
                      Type results,
                      ModuleInstance& instance) override {
+
+    std::unordered_map<wasm::Name, std::vector<wasm::Name>>::iterator it;
+
+    if (tableName.is()) {
+      it = tables.find(tableName);
+    } else {
+      it = tables.begin();
+    }
```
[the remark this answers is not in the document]
martianboy: Done.

src/wasm.h
```diff
@@ -1814,7 +1809,8 @@ class Module {
   std::vector<std::unique_ptr<Global>> globals;
   std::vector<std::unique_ptr<Event>> events;
```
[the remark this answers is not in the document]
martianboy: 👍

src/wasm/wasm-binary.cpp
```diff
@@ -516,38 +522,44 @@ uint32_t WasmBinaryWriter::getTypeIndex(HeapType type) const {
...
+                         table->hasMax(),
+                         /*shared=*/false,
+                         /*is64*/ false);
+  });
   finishSection(start);
 }
 
 void WasmBinaryWriter::writeTableElements() {
-  if (!wasm->table.exists || wasm->table.segments.size() == 0) {
+  if (wasm->tables.empty() || wasm->tables.front()->segments.empty()) {
```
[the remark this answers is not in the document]
martianboy: Done.

src/wasm/wasm-binary.cpp
```diff
@@ -516,38 +522,44 @@ uint32_t WasmBinaryWriter::getTypeIndex(HeapType type) const {
...
-    o << U32LEB(0);
-    writeExpression(segment.offset);
-    o << int8_t(BinaryConsts::End);
-    o << U32LEB(segment.data.size());
-    for (auto name : segment.data) {
-      o << U32LEB(getFunctionIndex(name));
+  for (auto& table : wasm->tables) {
+    o << U32LEB(table->segments.size());
+    for (auto& segment : table->segments) {
+      o << U32LEB(0);
```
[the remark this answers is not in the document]
martianboy: Done.

src/wasm/wasm-binary.cpp
martianboy: This is partially lifted now as I added support headerFormat 0x2.

src/ir/module-splitting.cpp
martianboy: You're right! Done.

test/example/c-api-multiple-tables.txt
martianboy: Updated!

src/wasm.h
```diff
@@ -1688,11 +1688,13 @@ class Table : public Importable {
   static const Index kMaxSize = Index(-1);
 
   struct Segment {
+    Name table;
```
tlively, line 1691: Why do segments need to refer to the table that owns them? Shouldn't the table be available (or easily made available) to any code the uses a `Segment`? I think it would be simpler not to introduce a conceptual reference cycle here (table owns segments, segments refer to table).

src/wasm/wasm-binary.cpp
```diff
@@ -2505,22 +2531,35 @@ void WasmBinaryBuilder::readTableElements() {
   }
   for (size_t i = 0; i < numSegments; i++) {
     auto headerFormat = getU32LEB();
-    if (headerFormat != 0) {
+    if (headerFormat != 0x0 && headerFormat != 0x2) {
```
tlively, line 2534: These constants should probably be given names in `wasm-binary.h`.

src/wasm/wasm-s-parser.cpp
```diff
@@ -2824,19 +2824,46 @@ void SExpressionWasmBuilder::parseTable(Element& s, bool preParseImport) {
...
+  Expression* offset = nullptr;
 
   if (!s[i]->isList()) {
-    auto name = getTableName(*s[i]);
-    table = wasm.getTableOrNull(name);
+    // optional segment id OR 'declare' OR start of elemList
+    i += 1;
+  }
+
+  bool oldStyle = true;
```
tlively, line 2834: It would be good to have a comment with examples of the old style and the new style.

test/example/c-api-multiple-tables.txt
tlively: It would be good to add `adder` to the second table as well to prove that that works.

src/wasm/wasm-binary.cpp
```diff
@@ -516,40 +522,66 @@ uint32_t WasmBinaryWriter::getTypeIndex(HeapType type) const {
...
+                         /*is64*/ false);
+  });
   finishSection(start);
 }
 
 void WasmBinaryWriter::writeTableElements() {
-  if (!wasm->table.exists || wasm->table.segments.size() == 0) {
+  if (wasm->tables.empty()) {
+    return;
+  }
```
tlively, line 549: In the case where there is at least one table but all the tables are empty, this early return won't happen and we will write out a useless table element section. I think it would be worth iterating through the tables and checking whether they are all empty, probably like this:
```suggestion
  if (std::all_of(wasm->tables.begin(), wasm->tables.end(), [&](auto& table) {
        return std::all_of(table->segments.begin(), table->segments.end(), [&] (auto& segment) {
          return segment.size() == 0;
        })
      })) {
    // There aren't any table segments, so skip this section.
    return;
  }
```

martianboy:
@tlively question: this PR adds the table name to CallIndirect as well, and most of the related changes have been made already, except for one thing: it can't print the table name arg yet. The reason is that I wanted to keep the old style (call_indirect without a table name) in cases where reference-types is disabled or there is only a single table. But `PrintExpressionContents` doesn't have a reference to the module and it did not seem super straightforward to add either.

So my question is, am I on the right path here?

tlively:
Actually, I think the simplest thing to do would be to always print the table name with `call_indirect`s. As long as we can still parse the old style, I think it should be ok if we always print the new style, even for modules without the reference-types feature enabled. Certainly it would be possible to design a workaround, but I don't see a big enough benefit in being able to print the old style to say that any workaround would be worth it. Perhaps I'm missing an important use case for printing the old style, though?

martianboy:
I agree. The reference-types proposal is at a very advanced stage (soon to be merged in the main spec, right?) with multiple browsers and toolchains supporting it, so printing in the old-style wouldn't matter that much.

I was mainly concerned that changing might introduce unwanted effects on the assumptions made in the test suite. I could well be overthinking it, though.

@kripken @aheejin Thoughts?

kripken:
@tlively 

> As long as we can still parse the old style, I think it should be ok if we always print the new style, even for modules without the reference-types feature enabled. 

Hmm, even if we make our parser work ok with that, it could mean our text output doesn't work in wabt or wasp.

A quick check shows it works in wabt but not wasp,
```
a.wat:19:18: Expected Rpar, got Id
  (call_indirect $z (type $f32_=>_none)
                 ^^                    
```

I think we should look into a (pre-) refactor to make `PrintExpressionContents` get the features. I can look into that myself if we want.

tlively:
Ok, I'm not opposed to such a refactor if you think it's a good idea. The default feature set when the actual features aren't available or provided should be `FeatureSet::All`. Eventually wasp will need to be updated for the new syntax, though.

@martianboy as far as the test suite goes, running `./auto_update_tests.py` will automatically update most test expectations for the new output (if we decide to make that change), so that alone wouldn't be sufficient reason to make the code more complicated.

kripken:
> Eventually wasp will need to be updated for the new syntax, though.

Sorry, I wasn't clear enough before. It does support the new syntax, just not when reference types are not enabled.

That seems reasonable to me, but another option is for us to get wasp to support the new syntax more generally, I suppose. There may be other tools, however - that's what worries me.

tlively:
@martianboy, would you be able to make a separate PR to make it possible to pass a feature set to the printer, or would you prefer that @kripken look into it?

martianboy:
@tlively sure I'll send a PR. I'll set the default to `FeatureSet::All` as you suggested.

src/wasm2js.h
```diff
@@ -622,7 +624,7 @@ void Wasm2JSBuilder::addGlobalImport(Ref ast, Global* import) {
 }
 
 void Wasm2JSBuilder::addTable(Ref ast, Module* wasm) {
-  if (!wasm->table.exists) {
+  if ((wasm->tables).size() == 0) {
```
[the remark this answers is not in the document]
martianboy: done.

src/wasm2js.h
```diff
@@ -631,69 +633,73 @@ void Wasm2JSBuilder::addTable(Ref ast, Module* wasm) {
...
-  if (isTableExported(*wasm)) {
-    // If the table is exported use a fake WebAssembly.Table object
-    // We don't handle the case where a table is both imported and exported.
-    if (wasm->table.imported()) {
-      Fatal() << "wasm2js doesn't support a table that is both imported and "
-                 "exported\n";
-    }
-    Ref theVar = ValueBuilder::makeVar();
-    ast->push_back(theVar);
+    if (isTableExported(*wasm)) {
```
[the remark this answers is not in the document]
martianboy: I did what you suggested and wasm2js should now error out as soon as a module with more than a single table is parsed.

src/tools/wasm-reduce.cpp
```diff
@@ -899,7 +899,7 @@ struct Reducer
     // If we are left with a single function that is not exported or used in
     // a table, that is useful as then we can change the return type.
     if (module->functions.size() == 1 && module->exports.empty() &&
-        module->table.segments.empty()) {
+        ((module->tables).empty() || module->tables[0]->segments.empty())) {
```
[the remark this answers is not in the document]
martianboy: Right, it now checks all tables for emptiness.

test/example/c-api-multiple-tables.txt
martianboy: Updated again.

src/ir/module-utils.h
```diff
@@ -208,6 +208,14 @@ template<typename T> inline void iterDefinedTables(Module& wasm, T visitor) {
   }
 }
 
+template<typename T> inline void iterNonemptyTables(Module& wasm, T visitor) {
```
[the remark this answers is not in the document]
martianboy: Right, I removed this.

src/wasm.h
martianboy: You're absolutely right, I totally forgot about the way we read or parse elem sections. I removed this.

src/wasm/wasm-binary.cpp
martianboy: So there are 8 kinds of header formats defined in the reference types proposal, which are a bit hard to come up with good names for; e.g. `ActiveSegmentWithRefInstructionWithoutTableIndex` would be one. The header byte is basically a bitfield, so we're probably going to parse it that way and not write separate code for each of the 8 different kinds.

Here I wrote it like this because I just wanted a minimal implementation to be able to test multi-table setups.

src/wasm/wasm-s-parser.cpp
martianboy: Done.

src/wasm/wasm-binary.cpp
martianboy: Right, this early condition was unnecessary because I'm counting segments of non-empty tables right after and return if the total count is zero. So I just removed it.

src/passes/Directize.cpp
```diff
@@ -36,25 +36,36 @@ namespace {
 struct FunctionDirectizer : public WalkerPass<PostWalker<FunctionDirectizer>> {
   bool isFunctionParallel() override { return true; }
 
-  Pass* create() override { return new FunctionDirectizer(flatTable); }
+  Pass* create() override { return new FunctionDirectizer(tables); }
```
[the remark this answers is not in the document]
martianboy: Done.

src/passes/RemoveUnusedModuleElements.cpp
```diff
@@ -196,6 +216,11 @@ struct RemoveUnusedModuleElements : public Pass {
       return analyzer.reachable.count(
                ModuleElement(ModuleElementKind::Event, curr->name)) == 0;
     });
+    module->removeTables([&](Table* curr) {
+      return (curr->segments.empty() || !curr->imported()) &&
+             analyzer.reachable.count(
+               ModuleElement(ModuleElementKind::Table, curr->name)) == 0;
+    });
```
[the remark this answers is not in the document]
martianboy: Done.

src/passes/RemoveUnusedModuleElements.cpp
```diff
@@ -214,6 +214,14 @@ struct RemoveUnusedModuleElements : public Pass {
       return analyzer.reachable.count(
                ModuleElement(ModuleElementKind::Event, curr->name)) == 0;
     });
+
+    for (auto& table : module->tables) {
+      table->segments.erase(
+        std::remove_if(table->segments.begin(),
+                       table->segments.end(),
+                       [&](auto& seg) { return seg.data.empty(); }),
```
tlively, line 222: It looks like this was already done below on line 225.

src/wasm/wasm-binary.cpp
tlively: It would be reasonable to have an enum giving names to each bit, then. Something like this (with better names, of course):

```
enum TableHeader {
  BitZeroMeaning = 1,
  BitOneMeaning = 1 << 1,
  ...
};
```

src/passes/RemoveUnusedModuleElements.cpp
martianboy: No, this removes empty segments first. Consider the case of a table with one empty segment. If we don't remove that segment first, the table would not be removed on 225. I encountered this when I was adding tests.

martianboy:
@kripken @tlively question: Apparently the `wasm-as` tool does not write the features section in the output binary, and as a result, the `.fromBinary` tests are always MVP only. Is that intentional? I was just updating the test suite with the newly ability to print with respect to features, and noticed that `.from-wast` files are updated differently from `.fromBinary`.

src/wasm/wasm-s-parser.cpp
martianboy: In all cases, it doesn't handle numerical index to the module element. Check the following:

```
a.wast:
(module
  (import "env" "f1" (func $f1 (param) (result)))
  (export "f1" (func 0))
)

$ bin/wasm-opt a.wast --print

Output:

[wasm-validator error in module] unexpected false: module function exports must be found, on 
0
(module
 (type $none_=>_none (func))
 (import "env" "f1" (func $f1))
 (export "f1" (func $0))
)

Fatal: error validating input
```

martianboy: I can make this a separate PR if that's what you meant.

src/wasm/wasm-binary.cpp
martianboy: The second bit changes meaning based on the first bit (!), but sure this would be better I guess. From the [reference types spec](https://webassembly.github.io/reference-types/core/binary/modules.html#element-section):

> The initial byte can be interpreted as a bitfield. Bit 0 indicates a passive or declarative segment, bit 1 indicates the presence of an explicit table index for an active segment and otherwise distinguishes passive from declarative segments, bit 2 indicates the use of element type and element expressions instead of element kind and element indices.

src/tools/wasm-as.cpp
```diff
@@ -116,6 +116,7 @@ int main(int argc, const char* argv[]) {
   }
 
   options.applyFeatures(wasm);
+  wasm.hasFeaturesSection = true;
```
tlively, line 119: This seems fine to me. @kripken, wdyt?

src/js/binaryen.js-post.js
```diff
@@ -576,19 +576,19 @@ function wrapModule(module, self = {}) {
   // 'callIndirect', 'returnCall', 'returnCallIndirect' are deprecated and may
   // be removed in a future release. Please use the the snake_case names
   // instead.
-  self['callIndirect'] = self['call_indirect'] = function(target, operands, params, results) {
+  self['callIndirect'] = self['call_indirect'] = function(target, table, operands, params, results) {
```
tlively, line 579: It would be good to test the changes to the JS API in the kitchen-sink.js test.

src/tools/wasm-as.cpp
martianboy: I think other tools might need this as well. I had similar problem in wasm-split today, although it's not in tests right now.

src/wasm/wasm-emscripten.cpp
tlively: ping @kripken, or maybe @sbc100 knows?

src/wasm/wasm-s-parser.cpp
tlively: No, that makes sense. Thanks for the explanation!

src/wasm/wasm-binary.cpp
tlively: Yuck, maybe a hybrid approach would work then? We could have 4 enum values covering all combinations of the first 2 bits, and a single enum value for bit 2,

src/passes/RemoveUnusedModuleElements.cpp
tlively: Got it, thanks 👍

src/wasm/wasm-binary.cpp
martianboy: Take a look at my last commit I pushed just now. Doesn't look bad, but let me know what you think.

src/tools/wasm-as.cpp
tlively: Let's make the smallest change possible to get the tests working in this PR and discuss the broader change in a separate issue. Would you mind opening an issue about this?

src/wasm-binary.h
```diff
@@ -1059,6 +1059,18 @@ enum MemoryAccess {
 
 enum MemoryFlags { HasMaximum = 1 << 0, IsShared = 1 << 1, Is64 = 1 << 2 };
 
+enum ElementFlags {
+  // Bit 0: 0 means active segment, 1 means passive or declarative
+  IsInactive = 1 << 0,
```
tlively, line 1064: I think it will be clearer to match the spec language here.

```suggestion
  IsPassive = 1 << 0,
```

src/wasm-binary.h
```diff
@@ -1059,6 +1059,18 @@ enum MemoryAccess {
...
+  // Bit 1 if passive or declarative: 0 means passive, 1 means declarative
+  IsDeclarative = 1 << 1,
```
tlively, line 1066: Can we say what it means to be "declarative" in the comment?

src/wasm/wasm-binary.cpp
```diff
@@ -2540,7 +2549,7 @@ void WasmBinaryBuilder::readTableElements() {
       throwError("Table index out of range.");
     }
 
-    if (headerFormat == 0x2) {
+    if (flags == 0x2) {
```
tlively, line 2552: It would be good to reformulate this condition in terms of `isInactive` and friends.

tlively:
This is getting really close! Thanks for your patience with all the back and forth, @martianboy :)

martianboy:
@tlively It is! Thanks for all the great comments, I really appreciate them. :) So I'll fix the broken js tests and check for the remaining items. After that, the only remaining point will be the fuzzer which might take a while!

src/wasm-binary.h
martianboy: Bit zero means "it might be passive or declarative. look at bit 1"! That's why I went for `IsInactive` as `IsPassive` might have been confusing.

martianboy: Unless I'm reading this the wrong way and a "declarative" segment is also considered "passive"? I was just reading the related section in the Bulk Memory Operations [overview doc](https://github.com/WebAssembly/bulk-memory-operations/blob/master/proposals/bulk-memory-operations/Overview.md#element-segments) and I think you're right.

kripken:
> Apparently the wasm-as tool does not write the features section in the output binary

Yes, when going from text there is no feature section to read, so I guess we don't emit one. wasm-opt has an option `--emit-target-features` but looks like wasm-as does not. Anyhow, for testing we probably already do `-all` to enable all features on those binaries - if not, we might need to add that.

src/tools/wasm-as.cpp
kripken: Sorry, I commented at the toplevel just now as I didn't realize this was the same topic as that.

I hope we can avoid adding this to a tool. Can we add `-all` to a test where we now need features enabled? If not then what specifically is the error we are running into?

martianboy:
@kripken 

> Anyhow, for testing we probably already do `-all` to enable all features on those binaries - if not, we might need to add that.

The problem is that `wasm-dis` doesn't have `-all`. It always prints with MVP features. So I was thinking we either add `-all` and other feature flags to `wasm-dis` or allow other tools to emit the features section in the output binary.

src/wasm/wasm-binary.cpp
martianboy: I merged the my enum into `SegmentFlag` as they have the exact same meaning for both table and memory segments. The only difference is the third bit which is exclusive to table segments at the moment. Also fixed this.

src/wasm-binary.h
martianboy: Added a comment above `SegmentFlag` enum explaining all three types of segments.

src/wasm-binary.h
martianboy: Went with `IsPassive` as you suggested.

src/js/binaryen.js-post.js
martianboy: Updated the js tests to always pass "0" as the table name to both `call_indirect` and `return_call_indirect`.

src/tools/wasm-as.cpp
martianboy: @kripken I tried to explain the situation in #3545.

martianboy:
@tlively I think I've addressed most of the review points here (I hope I haven't missed any), except for adding support for multiple tables in the fuzzer.

After reading `fuzzing.h` a little bit, I feel like anything I do in the fuzzer now to support multiple tables will have to be changed very soon when we add data type to tables and element segments, and convert segment data items from function names into (ref) expressions. If you agree, we can put that part for a later PR, but I'm totally open to suggestions.

Other than that, please do let me know if you have other comments about this PR. Hopefully we can finalize it very soon!

tlively:
Neat, the latest changes look good to me. Can you summarize the C/JS API changes in the change log? Beyond that, I want to just do another review pass over everything and then it should be good!

src/tools/wasm-as.cpp
tlively: I think we'll probably want to add feature options to wasm-dis instead. Would you mind doing that as part of this PR or as a separate PR @martianboy?

martianboy:
@tlively I added a few missing C/JS APIs and updated the change log. Not sure if the change log update is good, we may want to make it shorter. I also sent PR #3548 to add feature options to `wasm-dis`.

src/wasm/wasm-binary.cpp
```diff
@@ -516,38 +522,61 @@ uint32_t WasmBinaryWriter::getTypeIndex(HeapType type) const {
...
-    writeExpression(segment.offset);
-    o << int8_t(BinaryConsts::End);
-    o << U32LEB(segment.data.size());
-    for (auto name : segment.data) {
-      o << U32LEB(getFunctionIndex(name));
+      // header format
+      if (tableIdx < 1) {
+        o << U32LEB(0x0);
+      } else {
+        o << U32LEB(0x2);
```
tlively, line 567: It would be good to use the enum values here.

src/wasm/wasm-emscripten.cpp
tlively: I think this should be fine, and if it's not, we can address that once LLVM supports multiple tables.

tlively:
After those final two comments are addressed, I will be happy to merge this :)

martianboy:
@tlively Thanks a lot for catching those! I've already started working on the next PR to add a ref type to `Table`. Hopefully it will be smaller and easier to review!

src/wasm/wasm-binary.cpp
martianboy: Thanks I used constants, although using `?:` made it look a bit ugly! I can try several `if` blocks if you prefer. I've seen both styles in `wasm-binary.cpp`.

tlively: Yes, I think the `if` blocks would be better in this case, but I'm going to merge this as-is because it is so big and I don't want it to get out of sync with upstream. If you wouldn't mind fixing this in a follow-up PR, that would be great :)

tlively (approved):

tlively:
Thanks for your great work on this, @martianboy!

martianboy:
Yaay! Finally! Thanks again @tlively. :)

tlively:
You're welcome :D

sbc100:
It looks like this change broke a bunch of emscirpten tests.  At first glance it looks like any test related to dyanmic linking:

e.g.:

```
./tests/runner.py wasm2.test_relocatable_void_function
runner:WARNING: use EMTEST_ALL_ENGINES=1 in the env to run against all JS engines, which is slower but provides more coverage
Test suites:
['test_core']
Running test_core: (1 tests)
(checking sanity from test runner)
shared:INFO: (Emscripten: Running sanity checks)
test_relocatable_void_function (test_core.wasm2) ... wasm-emscripten-finalize: /usr/local/google/home/sbc/dev/wasm/binaryen/src/wasm/wasm-binary.cpp:498: uint32_t wasm::WasmBinaryWriter::getTableIndex(wasm::Name) const: Assertion `it != indexes.tableIndexes.end()' failed.
emcc: error: '/usr/local/google/home/sbc/dev/wasm/binaryen-out/bin/wasm-emscripten-finalize --detect-features --minimize-wasm-changes --dyncalls-i64 --pass-arg=legalize-js-interface-export-originals test_relocatable_void_function.wasm -o test_relocatable_void_function.wasm' failed (-6)
FAIL

```

martianboy:
@sbc100 I will look into this as soon as I can.

tlively:
I fixed it in #3560

PR #3520, opened 2021-01-26
[GC] ref.as_*

kripken:
These are similar to `is`, but instead of returning an i32 answer, they trap on
an invalid value, and return it otherwise.

These could in theory be in a single `RefDoThing`, with opcodes for both As
and Is, but as the return values are different, that seems a little odd. Also harder
to think of a name for it. But open to ideas.


src/wasm/wasm.cpp
```diff
@@ -245,6 +245,17 @@ const char* getExpressionName(Expression* curr) {
       return "array.set";
     case Expression::Id::ArrayLenId:
       return "array.len";
+    case Expression::Id::RefAsId:
```
tlively, line 248: It looks like `RefIs` and `RefAs` are the only expression classes that change their names based on their contents. Is there a particular reason that they are special here? Also, I'm not sure what this function is actually useful for, given all our printing facilities. Can we remove it in a separate change?

tlively (approved):
LGTM beyond that one comment.

src/passes/Print.cpp
```diff
@@ -1856,6 +1856,21 @@ struct PrintExpressionContents
...
+        printMedium(o, "ref.as_func");
+        break;
+      case RefAsData:
+        printMedium(o, "ref.as_data");
+        break;
+      case RefAsI31:
+        printMedium(o, "ref.as_i31");
+        break;
+      default:
+        WASM_UNREACHABLE("unimplemented ref.is_*");
```
aheejin, line 1871:
```suggestion
        WASM_UNREACHABLE("invalid ref.is_*");
```
`unimplemented` sounds like we still have some instructions unimplemeneted?

src/wasm/wasm-stack.cpp
```diff
@@ -2072,6 +2072,22 @@ void BinaryInstWriter::visitArrayLen(ArrayLen* curr) {
...
+      o << int8_t(BinaryConsts::GCPrefix) << int8_t(BinaryConsts::RefAsFunc);
+      break;
+    case RefAsData:
+      o << int8_t(BinaryConsts::GCPrefix) << int8_t(BinaryConsts::RefAsData);
+      break;
+    case RefAsI31:
+      o << int8_t(BinaryConsts::GCPrefix) << int8_t(BinaryConsts::RefAsI31);
+      break;
+    default:
+      WASM_UNREACHABLE("unimplemented ref.as_*");
```
aheejin, line 2087:
```suggestion
      WASM_UNREACHABLE("invalid ref.as_*");
```

aheejin (approved):

src/wasm/wasm.cpp
kripken: Yeah, we should remove `getExpressionName`, good point. I'll do that in my very next PR (if it it can't be removed, I'll figure out how to fix things so it makes sense).

PR #3521, opened 2021-01-27
Memcpy data instead of bytewise copies

pfaffe:
wasm-finalize currently makes byte-wise copies of section data in the user and data sections. If the section is large, that's extraordinarily expensive. With a memcpy instead I see a speedup of 1.6 for a large wasm binary with DWARF data.


kripken (approved):
Nice!

lgtm with the lint errors fixed

PR #3522, opened 2021-01-27
Simplfy getExpressionName

kripken:
This used to return a simple name like "if" for an If, but it is redundant with
our proper printing logic. This PR turns it into a trivial helper that just prints
out the name of the class, so it now prints "If" with a capital. That is useful
for some logging, like in Metrics I think it is clearer than it was earlier (since
we are actually counting the classes, and our old emitting of text-format-like
names are just confusing, as we emitted "binary" there which is not valid).

Also replace some usages of that method with proper printing.


aheejin (approved):
Nice!

tlively (approved):
🎉

PR #3523, opened 2021-01-27
[GC] Update br_on_cast: the text format also no longer has a heap type

kripken:
I removed it from the binary format earlier, but forgot the text.

As a result, we cannot handle a `br_on_cast` with an unreachable RTT. The
binary format solves the problem by ignoring unreachable code, and this makes
the text format do the same.

A nice benefit of this is that we can remove the `castType` extra field.

### DIFF
--- a/scripts/test/shared.py
+++ b/scripts/test/shared.py
@@ -493,18 +493,8 @@ def minify_check(wast, verify_final_result=True):
     print('      ', ' '.join(cmd))
     subprocess.check_call(cmd, stdout=open('a.wast', 'w'), stderr=subprocess.PIPE)
     assert os.path.exists('a.wast')
-    subprocess.check_call(WASM_OPT + ['a.wast', '--print-minified', '-all'],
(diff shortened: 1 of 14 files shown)


scripts/test/shared.py
```diff
@@ -493,18 +493,8 @@ def minify_check(wast, verify_final_result=True):
...
-    assert os.path.exists('b.wast')
-    if verify_final_result:
-        expected = open('a.wast').read()
-        actual = open('b.wast').read()
-        if actual != expected:
-            fail(actual, expected)
-    if os.path.exists('a.wast'):
-        os.unlink('a.wast')
-    if os.path.exists('b.wast'):
-        os.unlink('b.wast')
```
tlively, line 507: Why this change?

tlively (approved):

scripts/test/shared.py
kripken: Sorry, I should have left a comment. See https://github.com/WebAssembly/binaryen/pull/3524 - I added it here too to show tests pass.

PR #3524, opened 2021-01-27
Remove test suite's assumption of minify_check roundtripping perfectly

kripken:
`minify_check` checks that we can print and read minified wast. The test
also, however, assumed that we round-trip such things perfectly. That's
never been true, and only by chance did this go unnoticed until now,
in https://github.com/WebAssembly/binaryen/pull/3523

The specific issue happening there is that we create a block without a
name. Then we write that as text, then read it. When we read it, we give
all such blocks a name (and we rely on optimizations to remove it later
when possible - this avoids optimizing in the parser). The extra name
looks like a bug to minify_check.

### DIFF
--- a/scripts/test/shared.py
+++ b/scripts/test/shared.py
@@ -493,18 +493,8 @@ def minify_check(wast, verify_final_result=True):
     print('      ', ' '.join(cmd))
     subprocess.check_call(cmd, stdout=open('a.wast', 'w'), stderr=subprocess.PIPE)
     assert os.path.exists('a.wast')
-    subprocess.check_call(WASM_OPT + ['a.wast', '--print-minified', '-all'],
+    subprocess.check_call(WASM_OPT + ['a.wast', '-all'],
                           stdout=open('b.wast', 'w'), stderr=subprocess.PIPE)
-    assert os.path.exists('b.wast')
-    if verify_final_result:
-        expected = open('a.wast').read()
-        actual = open('b.wast').read()
-        if actual != expected:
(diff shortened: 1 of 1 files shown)


scripts/test/shared.py
```diff
@@ -493,18 +493,8 @@ def minify_check(wast, verify_final_result=True):
...
-    assert os.path.exists('b.wast')
-    if verify_final_result:
-        expected = open('a.wast').read()
-        actual = open('b.wast').read()
-        if actual != expected:
-            fail(actual, expected)
-    if os.path.exists('a.wast'):
-        os.unlink('a.wast')
-    if os.path.exists('b.wast'):
-        os.unlink('b.wast')
```
tlively, line 507: It looks like asserting that b.wast exists would still be useful. Also, why delete the code that removes a.wast and b.wast after running the test?

kripken: I don't think we need to remove files manually. Since @sbc100 moved us unto running tests in a test folder, things are contained anyhow.

What would checking b.wast existing be helpful? It'll definitely exist since we opened the file.

tlively: Fair enough. Asserting that b.wast exists is mostly nice for symmetry since we do the same check on a.wast. I'd be ok with removing that check, too. We could also just not write a b.wast, since we don't use it for anything anymore.

kripken: Good points, thanks. Done.

tlively (approved):

PR #3525, opened 2021-01-28
[GC] Add br_on_func/data/i31

kripken:
This expands the existing `BrOnCast` into `BrOn` that can also handle the
func/data/i31 variants. This is not as elegant as `RefIs / RefAs` in that `BrOnCast`
has an extra `rtt` field, but I think it is still the best option. We already have optional
fields on `Break` (the value and condition), so making `rtt` optional is not odd. And
it allows us to share all the behavior of `br_on_*` which aside from the cast or the
check itself, is identical - returning the value if the branch is not taken, etc.


src/wasm-interpreter.h
```diff
@@ -1469,17 +1469,54 @@ class ExpressionRunner : public OverriddenVisitor<SubType, Flow> {
...
     }
-    if (cast.outcome == cast.Null || cast.outcome == cast.Failure) {
-      return cast.originalRef;
+    const auto& value = flow.getSingleValue();
+    NOTE_EVAL1(value);
+    // Prepare to break. If we do not break, we will unset the break target, and
+    // we just flow out the value.
+    flow = Flow(curr->name, value);
+    if (value.isNull()) {
+      flow.breakTo = Name();
```
aheejin, line 1497:
```suggestion
      flow.breakTo.clear();
```
Nit: I guess this reads more intuitive..? The same for similar lines

src/wasm/wasm.cpp
```diff
@@ -1032,7 +1046,7 @@ void RefAs::finalize() {
       type = Type::i31ref;
       break;
     default:
-      WASM_UNREACHABLE("unimplemented ref.is_*");
+      WASM_UNREACHABLE("invalid ref.is_*");
```
aheejin, line 1049:
```suggestion
      WASM_UNREACHABLE("invalid ref.as_*");
```

aheejin (approved):
Nice! I agree it's better to share the `BrOn` class. There were quiet a few places that I needed to handle `BrOnExn` in various places, and duplicating all that with multiple classes will be a pain. (#3505 removes all of them, but now I think it might be able to give some info for this new `BrOn` instruction on where this needs to be additionally handled for optimization passes.)

src/wasm/wasm.cpp
```diff
@@ -931,15 +931,29 @@ void RefCast::finalize() {
 
 Type RefCast::getCastType() { return doGetCastType(this); }
 
-void BrOnCast::finalize() {
-  if (ref->type == Type::unreachable || rtt->type == Type::unreachable) {
+void BrOn::finalize() {
+  if (ref->type == Type::unreachable ||
+      (rtt && rtt->type == Type::unreachable)) {
```
tlively, line 936: Is it ever ok for `rtt` to be unreachable in practice? If not, perhaps this should be an assert.

src/ir/cost.h
```diff
@@ -582,8 +582,11 @@ struct CostAnalyzer : public OverriddenVisitor<CostAnalyzer, Index> {
   Index visitRefCast(RefCast* curr) {
     return 2 + nullCheckCost(curr->ref) + visit(curr->ref) + visit(curr->rtt);
   }
-  Index visitBrOnCast(BrOnCast* curr) {
-    return 3 + nullCheckCost(curr->ref) + visit(curr->ref) + visit(curr->rtt);
+  Index visitBrOn(BrOn* curr) {
+    // BrOnCast has more work to do with the rtt, so add a little there.
+    Index base = curr->op == BrOnCast ? 3 : 2;
+    return base + nullCheckCost(curr->ref) + visit(curr->ref) +
```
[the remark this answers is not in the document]
kripken: Yes. The output is a non-nullable type (funcref, etc.). You can basically do `br_on_func` and if it is an actual function, you get it, all ready to call with no chance of traps.

src/wasm/wasm-binary.cpp
```diff
@@ -5825,17 +5825,34 @@ bool WasmBinaryBuilder::maybeVisitRefCast(Expression*& out, uint32_t code) {
...
   auto name = getBreakTarget(getU32LEB()).name;
-  auto* rtt = popNonVoidExpression();
-  if (!rtt->type.isRtt()) {
-    throwError("bad rtt for br_on_cast");
+  Expression* rtt = nullptr;
+  if (op == BrOnCast) {
+    rtt = popNonVoidExpression();
+    if (!rtt->type.isRtt()) {
+      throwError("bad rtt for br_on_cast");
+    }
```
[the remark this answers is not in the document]
kripken: Oh, I think you're right, we can leave it for the validator.

src/wasm/wasm.cpp
kripken: This is a little tricky. The text and binary readers will not produce such a thing. But creating one as IR directly could. I think it's better to let the validator catch the error and print it nicely, rather than assert here, but I could go either way - thoughts?

tlively: Ok, that's fine with me. I guess the worst case is that someone sees a crash when they try to get the rtt type of an unreachable and that shouldn't leave much question about what the problem is.

tlively (approved):

PR #3526, opened 2021-01-28
Don't error out when EH is used for Flatten

aheejin:
We used to error out in Flatten when EH is used because Flatten makes
all block types `none` by setting their return values to locals and
getting them later, but `br_on_exn` by definition pops value from the
value stack at the end of a block so it couldn't be flattened. Now that
we don't have `br_on_exn` we don't need these restriction.

### DIFF
--- a/src/dataflow/graph.h
+++ b/src/dataflow/graph.h
@@ -228,6 +228,8 @@ struct Graph : public UnifiedExpressionVisitor<Graph, Node*> {
       return doVisitUnreachable(unreachable);
     } else if (auto* drop = curr->dynCast<Drop>()) {
       return doVisitDrop(drop);
+    } else if (curr->is<Try>() || curr->is<Throw>() || curr->is<Rethrow>()) {
+      Fatal() << "DataFlow does not support EH instructions yet";
     } else {
       return doVisitGeneric(curr);
     }
--- a/src/dataflow/graph.h
+++ b/src/dataflow/graph.h
@@ -228,8 +228,6 @@ struct Graph : public UnifiedExpressionVisitor<Graph, Node*> {
       return doVisitUnreachable(unreachable);
     } else if (auto* drop = curr->dynCast<Drop>()) {
       return doVisitDrop(drop);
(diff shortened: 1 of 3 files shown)


src/dataflow/graph.h
```diff
@@ -228,8 +228,6 @@ struct Graph : public UnifiedExpressionVisitor<Graph, Node*> {
       return doVisitUnreachable(unreachable);
     } else if (auto* drop = curr->dynCast<Drop>()) {
       return doVisitDrop(drop);
-    } else if (curr->is<Try>() || curr->is<Throw>() || curr->is<Rethrow>()) {
-      Fatal() << "DataFlow does not support EH instructions yet";
```
kripken, line 232: Should we not keep this code? DataFlow will not work on EH until we implement something custom here. I'm not sure what would happen if we fall through to the generic path after this, but a clear error seems fine.

src/passes/ReReloop.cpp
```diff
@@ -278,8 +278,6 @@ struct ReReloop final : public Pass {
       ReturnTask::handle(*this, ret);
     } else if (auto* un = curr->dynCast<Unreachable>()) {
       UnreachableTask::handle(*this, un);
-    } else if (curr->is<Try>() || curr->is<Throw>() || curr->is<Rethrow>()) {
-      Fatal() << "ReReloop does not support EH instructions yet";
```
kripken, line 282: Same question as about DataFlow.

kripken (commented):
Sgtm about Flatten, but I have questions about the other two.

src/dataflow/graph.h
aheejin: I think you're right; I just mechanically reverted the previous PR because I thought now we are OK with Flatten, but yeah I think we should keep this. Thanks.

src/passes/ReReloop.cpp
aheejin: Restored this too.

kripken (approved):

PR #3527, opened 2021-01-28
[GC] ref.as_non_null

kripken:
This is different than the other RefAs variants in that it is part of the
typed functions proposal, and not GC. But it is part of GC prototype 3.

Note: This is not useful to us yet as we don't support non-nullable types.

### DIFF
--- a/src/wasm/wasm.cpp
+++ b/src/wasm/wasm.cpp
@@ -1036,6 +1036,10 @@ void RefAs::finalize() {
     return;
   }
   switch (op) {
+    case RefAsNonNull:
+      // FIXME: when we support non-nullable types, switch to NonNullable
+      type = Type(value->type.getHeapType(), Nullable);
+      break;
     case RefAsFunc:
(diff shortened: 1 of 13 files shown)


src/wasm/wasm.cpp
```diff
@@ -1036,6 +1036,10 @@ void RefAs::finalize() {
     return;
   }
   switch (op) {
+    case RefAsNonNull:
+      // FIXME: when we support non-nullable types, switch to NonNullable
+      type = Type(value->type.getHeapType(), Nullable);
```
tlively, line 1041: What happens here if the `value` is unreachable?

kripken: In that case the return on line 1036 would have been taken, and we'd never get here.

tlively: aha, should have expanded the context 👍

tlively (approved):

PR #3528, opened 2021-01-28
[GC] br_on_null

kripken:
BrOnNull is different than the other variants by being part of the typed
functions proposal, and also it is kind of flipped: it branches if null, and it does
not flow out the existing value otherwise, rather it returns it as non-null. I have
not followed the spec discussion closely so I am not sure why this is so, but
I'm guessing it's to make it more similar to `ref.as_non_null` perhaps - it flows
out the non-null type, and this just adds a possible branch to handle the error
instead of trapping.

Note: This is not useful to us yet as we don't support non-nullable types.


src/wasm-interpreter.h
```diff
@@ -1490,6 +1490,15 @@ class ExpressionRunner : public OverriddenVisitor<SubType, Flow> {
     }
     const auto& value = flow.getSingleValue();
     NOTE_EVAL1(value);
+    if (curr->op == BrOnNull) {
+      // BrOnNull is the reverse of the others: it branches if null, and it does
+      // not flow out the existing value otherwise, rather it returns it as non-
+      // null.
```
tlively, line 1496: If I'm reading the proposal correctly, `br_on_null` does not propagate the value if it takes the branch. Also, it doesn't change the value if it doesn't take the branch, but rather refines the value's type. But the value's dynamic type doesn't change, so that doesn't matter here in the interpreter.

```suggestion
      // Unlike the others, BrOnNull does not propagate the value if it takes the
      // branch.
```

src/wasm-interpreter.h
```diff
@@ -1490,6 +1490,15 @@ class ExpressionRunner : public OverriddenVisitor<SubType, Flow> {
...
+      if (value.isNull()) {
+        return Flow(curr->name, value);
```
tlively, line 1498:
```suggestion
        return Flow(curr->name);
```

src/wasm/wasm.cpp
```diff
@@ -936,12 +936,22 @@ void BrOn::finalize() {
       (rtt && rtt->type == Type::unreachable)) {
     type = Type::unreachable;
   } else {
-    type = ref->type;
+    if (op == BrOnNull) {
+      // BrOnNull is the reverse of the others: it branches if null, and it does
+      // not flow out the existing value otherwise, rather it returns it as non-
+      // null.
```
tlively, line 942: Same edit here.

test/heap-types.wast
```diff
@@ -194,6 +194,14 @@
...
   (func $br_on_X (param $x anyref)
     (local $y anyref)
+    (drop
+      (block $null (result anyref)
+        (local.set $y
+          (br_on_null $null (local.get $x))
+        )
+        (ref.null any)
+      )
+    )
```
tlively, line 204: From the proposal:

> br_on_null $l : [t* (ref null ht)] -> [t* (ref ht)]
>  - iff $l : [t*]
>  - and ht ok

I read the condition on `$l` to mean that the branch does not carry any values to the label. Is there a more up-to-date document I should be looking at?

```suggestion
    (drop
      (block $null 
        (local.set $y
          (br_on_null $null (local.get $x))
        )
      )
    )
```

This will probably require validator changes as well.

kripken:
> From the proposal:

> br_on_null $l : [t* (ref null ht)] -> [t* (ref ht)]
> iff $l : [t*]
> and ht ok

> I read the condition on $l to mean that the branch does not carry any values to the label. Is there a more up-to-date document I > should be looking at?

We read the same thing, but are interpreting it differently :smile: 

I understood `$l: [t*]` to mean that the block we target `$l` has a type, `[t*]`. So we must pass a value there..? But that doesn't make sense, now that I think about it, as it's a null. Still, how are you interpreting `$l : [t*]`..?

tlively:
If the branch carried a value, I would have expected this to say `$l : [(ref null ht)]`. I'm not sure what to make of the  `$l: [t*]`, TBH. The `t*` is everything on the stack that's *not* the nullable value in question, so I would expect it to all be dropped on the branch, not carried along. I would have expected the condition to be `$l: []` to signify that no values at all are carried to the label. It seems clear to me that the tested value (known to be null) is not carried to the branch, though.

kripken:
Ah, I think I get it now: `$l : [t*]` are values carried with the branch. In other words, it behaves almost exactly like `br` and `br_if`.

In summary, `br_on_null`

* Pops a value, and checks if it's null.
* Pops other values off the stack based on the return values of the block we branch to (same as `br / br_if`).
* If the value was null, the branch is taken. Send the other values popped.
* If the value was not null, return that value (now with non-nullable type) as well as pushing back the other values popped.

So `br_on_null` is more like `br_if` than `br_on_*`. They both have a "condition" input that determines whether to branch, and a "value(s)" input. And if they branch, they both send the value(s) to there. However,

* `br_on_null` branches if the condition is falsey, while `br_if` branches on truthy.
* If the branch is not taken, `br_if` re-pushes the value(s), while `br_on_non_null` also re-pushes the condition, and it pushes it with a non-nullable type.

Given all that, I'm not sure it makes sense to implement `br_on_null` as part of `BrOn`. It could maybe be a variant in `Break` (which already has variations for having or not having a condition and a value), or it could be a new class. I'm not sure which I prefer atm.

kripken:
Let's pause this PR, I opened an issue to discuss the spec side: https://github.com/WebAssembly/function-references/issues/45

kripken:
Updated to support the proper semantics. This does not support an optional value (the `t*` stuff) but may handle all cases that matter in practice in the short term.

src/wasm/wasm-stack.cpp
```diff
@@ -1982,19 +1982,21 @@ void BinaryInstWriter::visitRefCast(RefCast* curr) {
...
 
 void BinaryInstWriter::visitBrOn(BrOn* curr) {
-  o << int8_t(BinaryConsts::GCPrefix);
   switch (curr->op) {
+    case BrOnNull:
+      o << int8_t(BinaryConsts::BrOnNull);
+      break;
     case BrOnCast:
-      o << U32LEB(BinaryConsts::BrOnCast);
+      o << int8_t(BinaryConsts::GCPrefix) << int8_t(BinaryConsts::BrOnCast);
```
kripken, line 1990: an `int8_t` must be used above for `BrOnNull`, or else as the opcode is large enough it will be emitted with an extra 00 byte after as an LEB. I made the same change for the others as well - I think this was a bug there before. And it now matches RefAs.

tlively: This is correct for `BrOnNull` because un-prefixed opcodes are bytes, but it is not correct to make this change for the other instructions. All prefixed opcodes are LEBs.

tlively (approved):
LGTM apart from that opcode comment.

src/wasm/wasm-stack.cpp
kripken: Thanks for the clarification, that wasn't clear to me from the spec.

PR #3533, opened 2021-02-01
Remove unnecessary breakStack pushes for if/try (NFC)

aheejin:
This break stack is maintained to compute branch depths, but it seems we
don't need to push and pop for each of if and else block or every single
catch/catch_all block; it would be sufficient to have just one stack
entry for the whole try-catch or if-else.


kripken (approved):
Good find, yes, it looks like we popped a value and pushed the exact same value there once more, so it's unnecessary... also logically there is just a single break target for the entire if/try.

PR #3535, opened 2021-02-01
Prototype i32x4.widen_i8x16_{s,u}

tlively:
As proposed in https://github.com/WebAssembly/simd/pull/395. Note that the other
instructions in the proposal have not been implemented in LLVM or in V8, so
there is no need to implement them in Binaryen right now either. This PR
introduces a new expression class for the new instructions because they uniquely
take an immediate argument identifying which portion of the input vector to
widen.


tlively:
cc @omnisnip.

@kripken, I enabled auto-merge on the PR, so make sure to disable first it if you want to LGTM+nits.

src/wasm/wasm-binary.cpp
```diff
@@ -5466,6 +5469,27 @@ bool WasmBinaryBuilder::maybeVisitSIMDLoadStoreLane(Expression*& out,
...
+    default:
+      return false;
+  }
+  auto* curr = allocator.alloc<SIMDWiden>();
+  curr->op = op;
+  curr->index = getLaneIndex(4);
+  curr->vec = popNonVoidExpression();
+  curr->finalize();
+  out = curr;
+  return true;
```
kripken, line 5490: please use a Builder and call the method on there.

src/wasm/wasm-s-parser.cpp
```diff
@@ -1666,6 +1666,15 @@ SExpressionWasmBuilder::makeSIMDLoadStoreLane(Element& s,
...
 }
 
+Expression* SExpressionWasmBuilder::makeSIMDWiden(Element& s, SIMDWidenOp op) {
+  auto* ret = allocator.alloc<SIMDWiden>();
+  ret->op = op;
+  ret->index = parseLaneIndex(s[1], 4);
+  ret->vec = parseExpression(s[2]);
+  ret->finalize();
+  return ret;
+}
```
kripken, line 1676: ditto

kripken (approved):
lgtm % comments

kripken:
Heh, I did not read the later comment and so it did an auto-merge... anyhow, my comments were not urgent.

omnisip:
What do I need to do to test this with emscripten and clang?

kripken:
In a few hours there should be a new tip of tree build with this included. You can then do `emsdk install tot` to get it.

Or, you can build binaryen manually and point emscripten to that location by editing the `.emscripten` config file.

omnisip:
Okay cool! Will let you guys know how it goes.

PR #3536, opened 2021-02-02
[GC] dataref typing fixes

kripken:
dataref was not noted as `isRef`, and we also did not handle the LUB for it,
which caused validation errors. After these two fixes, it is possible to add a
testcase that goes through the optimizer.

View without whitespace as the LUB change has a lot of that.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -370,7 +370,7 @@ bool Type::isTuple() const {
 
 bool Type::isRef() const {
   if (isBasic()) {
-    return id >= funcref && id <= i31ref;
+    return id >= funcref && id <= dataref;
   } else {
     return getTypeInfo(*this)->isRef();
   }
@@ -396,7 +396,7 @@ bool Type::isData() const {
 
 bool Type::isNullable() const {
   if (isBasic()) {
-    return id >= funcref && id <= eqref; // except i31ref
+    return id >= funcref && id <= eqref; // except i31ref and dataref
   } else {
     return getTypeInfo(*this)->isNullable();
   }
@@ -671,7 +671,7 @@ Type Type::getLeastUpperBound(Type a, Type b) {
                   a.isNullable() || b.isNullable() ? Nullable : NonNullable);
     };
     auto aHeap = a.getHeapType();
-    auto BHeap = b.getHeapType();
+    auto bHeap = b.getHeapType();
     if (aHeap.isFunction() && bHeap.isFunction()) {
       return handleNullability(HeapType::func);
     }
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -678,8 +678,10 @@ Type Type::getLeastUpperBound(Type a, Type b) {
     if (aHeap.isData() && bHeap.isData()) {
       return handleNullability(HeapType::data);
     }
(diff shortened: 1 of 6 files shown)


kripken:
(I enabled auto-merge...)

src/wasm/wasm-type.cpp
```diff
@@ -370,7 +370,7 @@ bool Type::isTuple() const {
 
 bool Type::isRef() const {
   if (isBasic()) {
-    return id >= funcref && id <= i31ref;
+    return id >= funcref && id <= dataref;
```
tlively, line 373: Maybe it would be worth defining more `static constexpr` values like `_last_basic_type` to make it easier to keep these up to date? For example, we could define `_first_reftype` and `_last_reftype`. On the other hand, this probably wouldn't scale well enough for a large number of categorizations of types.

src/wasm/wasm-type.cpp
```diff
@@ -662,23 +662,34 @@ Type Type::getLeastUpperBound(Type a, Type b) {
...
+    auto aHeap = a.getHeapType();
+    auto bHeap = b.getHeapType();
+    if (aHeap.isFunction() && bHeap.isFunction()) {
+      return handleNullability(HeapType::func);
+    }
+    if (aHeap.isData() && bHeap.isData()) {
+      return handleNullability(HeapType::data);
+    }
+    if ((aHeap == HeapType::eq || aHeap == HeapType::i31 ||
+         aHeap == HeapType::data) &&
```
tlively, line 682: There should probably be an `isEq` predicate for this case on both `Type` and `HeapType`, just like there is for `isFunction`.

src/wasm/wasm-type.cpp
```diff
@@ -662,23 +662,34 @@ Type Type::getLeastUpperBound(Type a, Type b) {
...
+        (bHeap == HeapType::eq || bHeap == HeapType::i31 ||
+         bHeap == HeapType::data)) {
+      return handleNullability(HeapType::eq);
     }
-    return Type::none;
+    // The LUB of two different reference types is anyref, which may or may
+    // not be a valid type depending on whether the anyref feature is enabled.
```
tlively, line 688: It's probably worth clarifying that `anyref` is enabled by the `gc` feature. (I think for a short time `anyref` had its own feature?)

src/wasm/wasm-type.cpp
aheejin, line 373: Oops, I should've changed it too in #3509, sorry..

aheejin (approved):
LGTM % @tlively's comments

aheejin:
Oh, it seems I approved the PR saying "LGTM % @tlively's comments", and this was thus automatically merged without having chances to address @tlively's comments... Didn't expect this side effect of auto-merge. Not sure how to address this kind of issues when using the auto-merge.

kripken:
Heh, the same happened to me yesterday... we all need to get used to automerge I guess. It basically means no more "lgtm % comments", which makes me wonder if it's actually saving us any time...

I'll work on a followup for the comments.

aheejin:
Yeah I would rather keep the option of approving with comments... I personally don't want to keep from clicking LGTM because of a typo. But if others prefer the auto merge model I can adjust.

tlively:
I sent GitHub feedback yesterday with the suggestion that you should be able to disable auto-merge from the "finishing your review" pane. Until then, we will have to remember to manually disable auto-merge when we want to LGTM with comments.

PR #3537, opened 2021-02-02
Allow printing expressions based on feature set

martianboy:
As discussed in #3517, we need a way to maintain the ability to print
expressions in a backward-compatible way. This adds a `FeatureSet`
to `PrintExpressionContents` which defaults to `FeatureSet::All` in
case a feature set is not available.

This will be used for call_indirect to decide whether to print the table
name arg.


martianboy:
@tlively @kripken

kripken (approved):
Perfect, thanks @martianboy !

PR #2784, opened 2020-04-22
Add stacky mode to BinaryenIR

tlively:
As an alternative to StackIR, adds a "stacky" mode to the normal
BinaryenIR in which normal instructions are completely flat and have
only pops as their children. This means that all instructions in
blocks may have arbitrary concrete types, not just the last
instructions.

The potential benefits of introducing a new mode for Binaryen IR are
that many passes will be agnostic to whether they run on the stacky or
unstacky code and stacky optimizations might be easier to write. This
will require further comparative exploration.

The next steps in this direction are:

 - Perform the actual tuple lowering in Stackify.cpp

 - Add a new validation mode that checks invariants for stacky code

 - Testing and fuzzing support

 - Reimplement the StackIR optimizations on the stacky BinaryenIR

 - Add checking to the pass runner to ensure passes get only code
   forms they can handle.

No more work will be done on this PR until tuples are made to work
with the current StackIR. Then maybe we can look more concretely at
the tradeoffs involved.

PR #3541, opened 2021-02-04
Poppify pass

tlively:
Adds a Poppify ("--poppify") pass for converting normal Binaryen IR to Poppy IR.
Like the existing construction of Stacky IR, Poppify depends on the
BinaryenIRWriter to drive the emitting of instructions in correct stack machine
order. As instructions are "emitted," Poppify replaces their children with pops
and collects them in a list. At the end of each scope, Poppify creates a block
containing all the collected instructions for that scope and injects that block
into the enclosing scope. All tuple globals and instructions dealing with tuples
are also expanded to remove all tuples from the program.

The validator currently fails to validate many valid Poppy IR patterns produced
in the tests, but fixing that is left as follow-on work to keep this PR focused
on the Poppify pass itself. For now the tests simply skip validation.


src/passes/Stackify.cpp
```diff
@@ -0,0 +1,497 @@
+/*
+ * Copyright 2020 WebAssembly Community Group participants
```
kripken, line 2:
```suggestion
 * Copyright 2021 WebAssembly Community Group participants
```

src/passes/Stackify.cpp
```diff
@@ -0,0 +1,497 @@
...
+//     (i32.const 5)
+//     (i32.add
+//      (i32.pop)
+//      (i32.pop)
+//     )
+//    )
+//   )
+//
+// Notice that the sequence of instructions in the block is now identical to the
+// sequence of instructions in raw WebAssembly. Also note that Poppy IR's
```
kripken, line 62:
```suggestion
// sequence of instructions in a WebAssembly binary. Also note that Poppy IR's
```

src/passes/Stackify.cpp
```diff
@@ -0,0 +1,497 @@
...
+// validation rules are largely additional on top of the normal Binaryen IR
+// validation rules, with the only exceptions being block body validation and
+// block unreahchability rules.
```
kripken, line 65:
```suggestion
// block unreachability rules.
```

src/passes/Stackify.cpp
```diff
@@ -0,0 +1,497 @@
...
+//     just the final expression in the block.
+//
+//  3. All other children must be Pops, which are used to determine the input
+//     stack type of each instruction. Pops may not have `unreachable` type.
+//
+//  4. Only control flow structures and instructions that have polymorphic
+//     unreachable behavior in WebAssembly may have unreachable type. Blocks may
+//     be unreachable when they are not branch targets and when they have an
+//     unreachable child. Note that this means a block may be unreachable even
+//     if it would otherwise have a concrete type, unlike in Binaryen IR.
```
kripken, line 37: Please add an example of such a block - this is not clear to me.

src/passes/Stackify.cpp
```diff
@@ -0,0 +1,497 @@
...
+  }
+}
+
+void Stackifier::emitLocalSet(LocalSet* curr) {
+  auto& instrs = scopeStack.back().instrs;
+  if (curr->value->type.isTuple()) {
+    auto types = func->getLocalType(curr->index);
+    const auto& elems = tupleVars[curr->index];
+    // Add the unconditional sets
+    for (size_t i = types.size() - 1; i >= 1; --i) {
```
kripken, line 370: I'm confused as to why `0` is special here. It seems like for tee you emit
```
set 3
set 2
set 1
tee 0
get 1
get 2
get 3
```
Why is that correct?

kripken (commented):
I believe I've already suggested calling this pass `Poppify` - seems clearer as we use the term "Poppy IR". What did we decide there?

src/passes/Stackify.cpp
kripken: Oh, wait, I see it, nevermind...

tlively:
> I believe I've already suggested calling this pass `Poppify` - seems clearer as we use the term "Poppy IR". What did we decide there?

Oops, I had forgotten about that. I personally prefer "Stackify," but I also suspect that objectively you're right that "Poppify" is a better name and that I will get used to it quickly. I've gone ahead and renamed everything to use the poppify terminology.

src/passes/Poppify.cpp
```diff
@@ -0,0 +1,504 @@
...
+//  3. All other children must be Pops, which are used to determine the input
+//     stack type of each instruction. Pops may not have `unreachable` type.
+//
+//  4. Only control flow structures and instructions that have polymorphic
+//     unreachable behavior in WebAssembly may have unreachable type. Blocks may
+//     be unreachable when they are not branch targets and when they have an
+//     unreachable child. Note that this means a block may be unreachable even
+//     if it would otherwise have a concrete type, unlike in Binaryen IR. For
+//     example, this block could have unreachable type in Poppy IR but would
+//     have to have type i32 in Binaryen IR:
```
kripken, line 39: Thanks, now I understand the difference.

But why is this so? It's not clear to me why it has to be this way. (If it doesn't, removing differences would be good - but I assume there is a reason...)

tlively: TBH I don't remember the original concrete motivation for making this change in Poppy IR. Allowing more things to be `unreachable` allows DCE to be more aggressive, but I don't remember if this was an opportunistic change or if doing things the Binaryen IR way caused problems. I'll remove this documentation change from the PR for now and if it was there for a good reason I will find out (again) soon enough.

tlively: Oh wait, this documentation is not new in this PR; it already existed in stack-utils.h. Would it be ok with you if we revisited this point once things are working end-to-end? I'm concerned that changes we make in this half-built state will come back to bite us before we finish.

kripken: If it can speed things up I'm not opposed. But I'd like to understand how this can work, though. The point of Poppy IR is to model the stack machine format more closely, and that format doesn't have unreachable blocks. So shouldn't it be the opposite, that is, the example block here would always have type i32 in Poppy IR?

Or am I misunderstanding Poppy IR?

src/passes/Poppify.cpp
```diff
@@ -0,0 +1,504 @@
...
+  Pass* create() override { return new PoppifyFunctionsPass; }
+};
+
+} // anonymous namespace
+
+class PoppifyPass : public Pass {
+  void run(PassRunner* runner, Module* module) {
+    PassRunner subRunner(runner);
+    subRunner.add(std::make_unique<PoppifyFunctionsPass>());
+    subRunner.add(std::make_unique<ReFinalize>());
```
aheejin, line 462: Why do we need refinalization?

src/passes/Poppify.cpp
```diff
@@ -0,0 +1,504 @@
...
+#include "pass.h"
+#include "wasm-stack.h"
+
+namespace wasm {
+
+namespace {
+
+// Generate names for the elements of tuple globals
+Name getGlobalElem(Name global, Index i) {
+  return Name(std::string(global.c_str()) + '$' + std::to_string(i));
```
aheejin, line 87: What if the original module already has two globals such as `myglobal` and `myglobal$1`?

src/passes/Poppify.cpp
```diff
@@ -0,0 +1,504 @@
...
+  } else {
+    instrs.push_back(curr);
+  }
+}
+
+void Poppifier::emitGlobalSet(GlobalSet* curr) {
+  auto& instrs = scopeStack.back().instrs;
+  if (curr->value->type.isTuple()) {
+    auto types = module->getGlobal(curr->name)->type;
+    for (Index i = types.size(); i > 0; --i) {
```
aheejin, line 416: Nit: Do we need to iterate in the reverse order? (`emitLocalSet` does though)

src/passes/Poppify.cpp
```diff
@@ -0,0 +1,504 @@
...
+    subRunner.run();
+    lowerTupleGlobals(module);
+  }
+
+  void lowerTupleGlobals(Module* module) {
+    Builder builder(*module);
+    std::vector<std::unique_ptr<Global>> newGlobals;
+    for (int g = module->globals.size() - 1; g >= 0; --g) {
+      const auto& global = *module->globals[g];
+      if (global.type.isTuple()) {
```
aheejin, line 472: Nit: Maybe we can make this an early exit in case it is not a tuple, given that this loop has also other for and ifs below

test/lit/passes/poppify.wast
```diff
@@ -0,0 +1,411 @@
...
+    (drop
+      (tuple.make
+        (i32.const 0)
+        (i64.const 1)
+      )
+    )
+  )
+
+  ;; CHECK:      (func $local-get-tuple (result i32 i64)
+  ;; CHECK-NEXT:  (local $x (i32 i64))
```
aheejin, line 304: When splitting a global, we seem to remove the original tuple global. Would it be good to do that for locals too? (Not necessarily in this PR though)

aheejin (commented):
Nice! The comments are mostly nits or questions:

src/passes/Poppify.cpp
```diff
@@ -0,0 +1,504 @@
...
+
+// Poppify.cpp - Transform Binaryen IR to Poppy IR.
+//
+// Poppy IR represents stack machine code using normal Binaryen IR types by
+// imposing the following constraints:
+//
+//  1. Function bodies and children of control flow (except If conditions) must
+//     be blocks.
+//
+//  2. Blocks may have any Expressions except for Pops as children. The sequence
```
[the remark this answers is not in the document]
tlively: Once we support blocks taking values, I expect that more blocks will have pop children. Right now catch blocks are somewhat limited because this binary cannot be directly expressed in Binaryen or Poppy IR because our we need to have some instruction to consume the pop of the caught value:

```
try i32
  do
    i32.const 0
    throw $e ;; throws i32
  catch $e
    ;; i32 on stack
end
```

Eventually I would like to implement that catch with this Poppy IR block:

```
(block (params i32) (results i32)
  (pop i32)
)
```

src/passes/Poppify.cpp
```diff
@@ -0,0 +1,504 @@
...
+//
+//   (func $foo (result i32)
+//    (block
+//     (i32.const 42)
+//     (i32.const 5)
+//     (i32.add
+//      (i32.pop)
+//      (i32.pop)
+//     )
+//    )
```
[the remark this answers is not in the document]
tlively: We do generate the block, but then the normal printing logic that elides the blocks that make up function bodies kicks in so the block is not printed. You know it's there because there are multiple instructions in the function body, though. For the purposes of this documentation, I think it's important to show that the block is there even if it won't be explicitly printed.

src/passes/Poppify.cpp
```diff
@@ -0,0 +1,504 @@
...
+          labelTypes[block->name] = block->type;
+        }
+        break;
+      }
+      case Expression::LoopId: {
+        kind = Scope::Loop;
+        auto* loop = curr->cast<Loop>();
+        if (loop->name.is()) {
+          // TODO: support loop parameters
+          labelTypes[loop->name] = Type::none;
```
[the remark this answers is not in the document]
tlively: Good catch, it is not. I briefly went through a period where I was recomputing types as I traversed the module, but it's much simpler to just use `ReFinalize` like any other pass would.

src/passes/Poppify.cpp
```diff
@@ -0,0 +1,504 @@
...
+    // happen when emitting if/else or try/catch arms and function bodies.
+    if (instrs.size() == 0 || instrs[0] != block) {
+      block->list.set(instrs);
+    }
+  } else {
+    // Otherwise create a new block, even if we have just a single
+    // expression. We want blocks in every new scope rather than other
+    // instructions because Poppy IR optimizations only look at the children of
+    // blocks.
+    expr = builder.makeBlock(instrs, expr->type);
```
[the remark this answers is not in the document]
tlively: If we know that every non-pop instruction is the child of a block, then Poppy IR passes can visit each block and its immediate children and know that it did not miss any instructions. [Here](https://github.com/WebAssembly/binaryen/pull/2784/files#diff-42cdb0683b33b025d354ddf11ecbe8c4d9859d596e033a550afd29a3996c9581R25) is an example of a pass that follows this pattern. Poppy IR passes will tend to have this structure because the ability to deal with flat blocks of instructions is what sets Poppy IR apart from Binaryen IR.

src/passes/Poppify.cpp
```diff
@@ -0,0 +1,504 @@
...
+  }
+  instrs.clear();
```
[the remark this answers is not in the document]
tlively: Yes, because we don't always pop the `Scope`. For example, in `emitIfElse` we keep the scope but change its `kind`. If we didn't do an `instrs.clear()` here, the instructions from the first arm would incorrectly also appear in the second arm.

src/passes/Poppify.cpp
tlively: Consider this Binaryen IR block:

```
(block $l i32
  (br $l
    (unreachable)
  (i32.const 0)
)
```

In Binaryen IR, this block would have to have type `i32` because it ends in an `i32.const`. However, the `BinaryenIRWriter` would not emit anything after the unreachable, so in Poppy IR we would end up with:

```
(block $l i32
  (unreachable)
)
```

We want to refinalize this block to give it type `unreachable` because there is no reason to keep it typed as `i32`. It's better to give additional optimizations the flexibility and power to do things to this block that do not preserve its `i32` type.

src/passes/Poppify.cpp
tlively: Yes. Consider this:

```
(global.set $tuple
  (tuple.make
     (i32.const 0)
     (i32.const 1)
     (i32.const 2)
  )
)
```

The `tuple.make` is eliminated by Poppification, so we end up with 0, 1, and 2 on the stack. Since the 2 is at the top of stack, it needs to be consumed first, so we need to work through the values from last pushed to first pushed.

src/passes/Poppify.cpp
tlively: I'd say the point of Poppy IR is to make it easy to write optimizations and other transformations that take the stack machine structure of instructions into account. Part of that is necessarily modeling the stack machine syntax more closely, but `unreachable` types still give optimizations useful flexibility and information, so they are worth preserving in Poppy IR.

src/passes/Poppify.cpp
tlively: Good point, this code will do the wrong thing in that case. I will make this more robust.

test/lit/passes/poppify.wast
tlively: I plan to use existing passes like `CoalesceLocals` to do that as a follow-up optimization. I guess in principle we could use `RemoveUnusedModuleElements` to clean up the globals as well, but that's simple enough to do here.

src/passes/Poppify.cpp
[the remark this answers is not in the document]
aheejin: > Once we support blocks taking values, I expect that more blocks will have pop children.

Then this sounds like not an inherent characteristic of Poppy IR but a tentative restriction that will be removed later..? How about clarifying that in the comment?

> Right now catch blocks are somewhat limited because this binary cannot be directly expressed in Binaryen or Poppy IR because our we need to have some instruction to consume the pop of the caught value:
> 
> ```
> try i32
>   do
>     i32.const 0
>     throw $e ;; throws i32
>   catch $e
>     ;; i32 on stack
> end
> ```

I'm not sure if I understand what you mean by 'limited' and we have to 'consume' the caught value. Given that we don't support block inputs at the moment, `catch` seems to be an exception, but why is it limited? And I think it is possible not to consume the caught value if we flow it out of `try`, as in your example?

src/passes/Poppify.cpp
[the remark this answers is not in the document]
aheejin: Yeah but that block, which is not printed, exists before the Poppify pass, and your comment looks like Poppify generates it because it doesn't show in the 'before' example but it does in the 'after' example. I think it'd be consistent to show it in both or show it in neither.

src/passes/Poppify.cpp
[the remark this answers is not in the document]
aheejin: I see. I personally think treating `Scope` like unit in that in `emitIfElse` we pop an old scope and push a new scope, is a little more intuitive, but I'm not very opinionated on that.

aheejin (approved):
I posted a few remaining questions but they are mostly minor. LGTM!

src/passes/Poppify.cpp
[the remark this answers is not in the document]
tlively: In this case because the function body is just a single `i32.add` expression, there won't be a block containing the body before the Poppify pass. Poppify really does generate a new block in this case.

src/passes/Poppify.cpp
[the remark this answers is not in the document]
tlively: I will update this comment to make clear that this restriction might be changed in the future.

I double checked and I was wrong about this example being invalid in Binaryen IR; sorry about that. There is no limitation in Binaryen IR after all. Right now the validator does improperly reject the Poppy IR version of that example, but I will fix that in a follow-up along with other validation issues with Poppy IR.

src/passes/Poppify.cpp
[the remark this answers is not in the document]
tlively: Ack. I like that `patchInstrs` clears the instructions because each instruction should logically be "consumed" by `patchInstrs` exactly once, so having `patchInstrs` actually consume them by clearing them from the scope stack ensures that this logic cannot be messed up. I think I can take this farther and actually have `patchInstrs` pop the entire scope so it matches your intuition as well.

PR #3542, opened 2021-02-04
[GC] Do not crash on unreachable inputs to struct.get/set

kripken:
If the reference is unreachable then we cannot find the heap type to print
in the text format. Instead of crashing or emitting something invalid, print
a block instead -  the block contains the children so they are emitted, and
as the instruction was unreachable anyhow, this has no noticeable effect.
It also parallels what we do in the binary format - skip unreachable code.

### DIFF
--- a/src/passes/Print.cpp
+++ b/src/passes/Print.cpp
@@ -1840,10 +1840,16 @@ struct PrintExpressionContents
     o << "with_rtt ";
     printHeapTypeName(o, curr->rtt->type.getHeapType());
   }
+  void printUnreachableReplacement() {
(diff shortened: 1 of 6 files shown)


tlively (approved):
This seems like something that could be fixed at the spec level by not including the heap type in the text format. I've lost track; did we get rid of the heap type immediates for these instructions in the binary format? If so, we should definitely remove them from the text format as well.

kripken:
I think these heap types are still in the text and binary formats atm, but they may be removed in the future.

In Binaryen IR we don't have them. We could add them to avoid this problem, but there is little benefit, and we'd have more work if the spec changes.

PR #3548, opened 2021-02-05
Add feature options to wasm-dis

### LINKED ISSUE #3545: .fromBinary tests are always executed with MVP features
We recently allowed printing expression contents with respect to features in the Print pass (#3537) and we needed that as part of enabling multi-table modules in the Binaryen IR (#3517). But after I updated the tests, I noticed there's a difference between `.from-wast` and `.fromBinary` test files, in that `.from-wast` files were generated correctly with the correct features, but `.fromBinary` files were always generated with all features disabled.

Now, for `.fromBinary` tests we use `wasm-as` to compile a .wasm binary, and then use `wasm-dis` to check the text output. But since `wasm-dis` does not have the feature options, the only way it can print with features is if the module has a features section, `wasm-as` does not have an option to emit.

Two solutions are possible in my view:

1. Add feature options to `wasm-dis`, so we have the flexibility to print with the desired features.
2. Allow tools like `wasm-as` and `wasm-split` to emit a features section, so the features are read and applied while parsing the binary in `wasm-dis`.

martianboy:
This will allow .fromBinary tests be executed with the desired features
so there will be no difference between those tests and .from-wast tests.

Fixes #3545, unblocks #3517

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ full changeset diff at the end of each section.
 Current Trunk
 -------------
 
+- `wasm-dis` now supports options to enable or disable Wasm features.
+
 v99
 ---
 
--- a/scripts/test/shared.py
+++ b/scripts/test/shared.py
@@ -466,7 +466,7 @@ def binary_format_check(wast, verify_final_result=True, wasm_as_args=['-g'],
     subprocess.check_call(cmd, stdout=subprocess.PIPE)
     assert os.path.exists('a.wasm')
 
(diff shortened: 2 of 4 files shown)


tlively (commented):
You'll also have to add `options.applyFeatures(wasm)` somewhere for this to have any effect. Otherwise lgtm!

martianboy:
@tlively damn I was testing this on the other branch, and I forgot this line!! Thanks for noticing.

tlively (approved):

PR #3550, opened 2021-02-06
[GC] Support RTT constants in Builder::makeConstantExpression

kripken:
This allows the reducer to operate on RTTs, but could help other
things too.

### DIFF
--- a/src/wasm-builder.h
+++ b/src/wasm-builder.h
@@ -826,6 +826,9 @@ public:
     if (type.isFunction()) {
       return makeRefFunc(value.getFunc(), type);
     }
+    if (type.isRtt()) {
+      return makeRtt(value.type);
+    }
     TODO_SINGLE_COMPOUND(type);
     switch (type.getBasic()) {
       case Type::externref:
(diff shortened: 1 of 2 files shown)


src/wasm-builder.h
```diff
@@ -853,6 +856,18 @@ class Builder {
     }
   }
 
+  // Given a type, creates an RTT expression of that type, using a combination
+  // of rtt.canon and rtt.subs.
+  Expression* makeRtt(Type type) {
+    Expression* ret = makeRttCanon(type.getHeapType());
+    if (type.getRtt().hasDepth()) {
+      for (Index i = 0; i < type.getRtt().depth; i++) {
+        ret = makeRttSub(type.getHeapType(), ret);
```
tlively, line 865: Is it legal to have a chain of Rtts with the same heap type? I would have assumed that each successive Rtt would have had to have a strict sub-heaptype of the previous Rtt. Does it even make sense to construct an Rtt with depth without saying what its super-Rtt should be?

kripken: I see no limitation in the spec that requires it be a strict subtype.

In fact, it is useful to do this type of thing in order to get actually different RTTs of the same type (as the "history" of `rtt.sub` operations defines identity). Although, I'm not sure how important that is in practice.

(Note btw that this PR only moves the code around. But it may have been wrong :wink: )

tlively (approved):
Moving the code LGTM, but I am still unsure of the validity and utility of this API. I asked for clarification upstream in https://github.com/WebAssembly/gc/pull/194.

kripken:
IIUC in the upstream bug it was agreed that this PR's assumptions are correct.

PR #3554, opened 2021-02-08
[GC] Support casts of function types

kripken:
I had completely missed that the spec allows `ref.cast` etc. of function types,
and not just data. Function types do not have an RTT, unlike GC data, but we
can still cast them. A function reference has the canonical RTT of the signature
for that type, so it's like a simplified case of the GC world, without a hierarchy
of RTTs.

cc @jakobkummerow who pointed this out - thanks!

As it turns out, our validation did not rule out `rtt.canon` of a function type,
nor `ref.cast` of one, so we unintentionally already had all the support for this
aside from the actual casting, which this PR adds.

The addition is mostly trivial, except that we now need a Module in the base
ExpressionRunner class, so that we can go from a function name to the actual
function. This PR refactors things to allow that.

### DIFF
--- a/src/wasm-interpreter.h
+++ b/src/wasm-interpreter.h
@@ -157,6 +157,10 @@ public:
 template<typename SubType>
 class ExpressionRunner : public OverriddenVisitor<SubType, Flow> {
 protected:
+  // Optional module context to search for globals and called functions. NULL if
+  // we are not interested in any context.
+  Module* module = nullptr;
+
   // Maximum depth before giving up.
   Index maxDepth;
   Index depth = 0;
@@ -209,6 +213,9 @@ public:
     depth--;
     return ret;
   }
+
+  // Gets the module this runner is operating on.
+  Module* getModule() { return module; }
 
   Flow visitBlock(Block* curr) {
     NOTE_ENTER("Block");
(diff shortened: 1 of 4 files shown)


src/wasm-interpreter.h
```diff
@@ -1426,24 +1435,36 @@ class ExpressionRunner : public OverriddenVisitor<SubType, Flow> {
...
-    auto refRtt = gcData->rtt;
-    auto intendedRtt = rtt.getSingleValue();
-    if (!refRtt.isSubRtt(intendedRtt)) {
-      cast.outcome = cast.Failure;
+    Literal seenRtt;
+    Literal intendedRtt = rtt.getSingleValue();
+    if (cast.originalRef.isFunction()) {
+      // Function casts are simple in that they have no RTT hierarchies; instead
+      // each reference has the canonical RTT for the signature.
+      auto* func = module->getFunction(cast.originalRef.getFunc());
```
tlively, line 1451: Does this need to handle the case where `module` is null?

src/wasm-interpreter.h
```diff
@@ -1858,8 +1872,8 @@ class ConstantExpressionRunner : public ExpressionRunner<SubType> {
   Flow visitGlobalGet(GlobalGet* curr) {
     NOTE_ENTER("GlobalGet");
     NOTE_NAME(curr->name);
-    if (module != nullptr) {
-      auto* global = module->getGlobal(curr->name);
+    if (this->module != nullptr) {
```
tlively, line 1875: Why is the `this->` necessary?

src/wasm-interpreter.h
kripken: I don't think we can handle that case. I can add an assertion here, would be better than a crash I suppose, though I believe all relevant code paths are known to be safe statically.

src/wasm-interpreter.h
kripken: It's the C++ thing where one template is the subclass of another template. It won't compile without this, as it says "`'module' was not declared in this scope`.

tlively: Yuck 😷 Thanks for the explanation!

src/wasm-interpreter.h
tlively: should we do something like `return (NONCONSTANT_FLOw)` so the program can continue?

kripken: In case of not having a module? I don't think we should try to handle that case, it should be a fatal error. We'd need to fix the code path leading to here so it provides a module.

tlively (approved):

PR #3557, opened 2021-02-08
finalize: strip segments that contain only EM_ASM/EM_JS data

sbc100:
If we find a data segment whose entire contents is EM_JS or EM_ASM
strings then strip it from the binary.

This change also introduces a new way to exact EM_ASM strings based
on the __start_ and _stop_ sections symbols.   This code will no be active
until the emscripten-side change lands.

See: https://github.com/emscripten-core/emscripten/pull/13443

### DIFF
--- a/src/wasm/wasm-emscripten.cpp
+++ b/src/wasm/wasm-emscripten.cpp
@@ -76,8 +76,7 @@ Global* getStackPointerGlobal(Module& wasm) {
 
 const Address UNKNOWN_OFFSET(uint32_t(-1));
 
-std::string escape(const char* input) {
-  std::string code = input;
+std::string escape(std::string code) {
   // replace newlines quotes with escaped newlines
   size_t curr = 0;
   while ((curr = code.find("\\n", curr)) != std::string::npos) {
@@ -109,13 +108,20 @@ class StringConstantTracker {
 public:
   StringConstantTracker(Module& wasm) : wasm(wasm) { calcSegmentOffsets(); }
 
-  std::string codeForConstAddr(int64_t address) {
-    const char* str = stringAtAddr(address);
-    if (!str) {
-      Fatal() << "unable to find data for ASM/EM_JS const at: " << address;
+  const char* stringAtAddr(Address address) {
+    for (unsigned i = 0; i < wasm.memory.segments.size(); ++i) {
+      Memory::Segment& segment = wasm.memory.segments[i];
+      Address offset = segmentOffsets[i];
+      if (offset != UNKNOWN_OFFSET && address >= offset &&
+          address < offset + segment.data.size()) {
+        return &segment.data[address - offset];
+      }
     }
-    return escape(str);
+    Fatal() << "unable to find data for ASM/EM_JS const at: " << address;
+    return nullptr;
   }
+
+  std::vector<Address> segmentOffsets; // segment index => address offset
 
 private:
   void calcSegmentOffsets() {
@@ -172,31 +178,18 @@ private:
     }
   }
 
-  const char* stringAtAddr(Address address) {
-    for (unsigned i = 0; i < wasm.memory.segments.size(); ++i) {
-      Memory::Segment& segment = wasm.memory.segments[i];
-      Address offset = segmentOffsets[i];
-      if (offset != UNKNOWN_OFFSET && address >= offset &&
-          address < offset + segment.data.size()) {
-        return &segment.data[address - offset];
-      }
-    }
-    return nullptr;
-  }
-
   Module& wasm;
-  std::vector<Address> segmentOffsets; // segment index => address offset
+};
+
+struct AsmConst {
+  Address id;
+  std::string code;
 };
 
 struct AsmConstWalker : public LinearExecutionWalker<AsmConstWalker> {
   Module& wasm;
   bool minimizeWasmChanges;
   StringConstantTracker stringTracker;
-
-  struct AsmConst {
-    Address id;
-    std::string code;
-  };
 
   std::vector<AsmConst> asmConsts;
   // last sets in the current basic block, per index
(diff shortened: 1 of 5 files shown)


sbc100:
Hmm.. how come the UI is saying I can merge this without a review?   Does binaryen not require an lgtm?

tlively:
That's weird, it should require an LGTM

sbc100:
Ah.. its because I wasn't targeting main.   Fixed.

src/wasm/wasm-emscripten.cpp
```diff
@@ -404,7 +406,32 @@ struct EmJsWalker : public PostWalker<EmJsWalker> {
...
+
+  bool isFunctionParallel() override { return true; }
+
+  Pass* create() override { return new SegmentRemover(segment); }
+
+  void visitMemoryInit(MemoryInit* curr) {
+    if (segment == curr->segment) {
+      Builder builder(*getModule());
+      replaceCurrent(builder.makeNop());
+    }
```
tlively, line 424: This would be more robust against side effects if it created a block that evaluated and dropped all the operands of the `MemoryInit`, but perhaps that's overkill considering how much we can assume about the output of wasm-ld here.

src/wasm/wasm-emscripten.cpp
```diff
@@ -415,6 +442,32 @@ EmJsWalker fixEmJsFuncsAndReturnWalker(Module& wasm) {
...
+
+    while (cur < start + wasm.memory.segments[i].data.size()) {
+      if (walker.codeAddresses.count(cur) == 0) {
+        break;
+      }
+      cur.addr += walker.codeAddresses[cur];
+    }
+
+    if (cur == start + wasm.memory.segments[i].data.size()) {
+      // Enture segment is containes JS strings.  Remove it.
```
tlively, line 462:
```suggestion
      // Entire segment is contains JS strings.  Remove it.
```

tlively (approved):

src/wasm/wasm-emscripten.cpp
sbc100: I see... I wonder if there are other places this is done?  Maybe there is some kind of helper for this already?

tlively: I don't know of a general utility for this, but here's an example where this is done: https://github.com/WebAssembly/binaryen/blob/74ccdb87b3390e003b658b847526dad8722083f8/src/passes/MemoryPacking.cpp#L413-L416

`Builder::blockify` makes this simple enough that we probably don't need a separate utility for it (although a dedicated utility couldn't hurt).

sbc100:
This change now handle EM_ASM strings too.

src/wasm/wasm-emscripten.cpp
```diff
@@ -305,32 +297,114 @@ void AsmConstWalker::process() {
...
+  Export* end = wasm.getExportOrNull("__stop_em_asm");
+
+  // Older versions of emscripten don't export these symbols.  Instead
+  // we run AsmConstWalker in an attempt to derive the string addresses
+  // from the code.
+  if (!start || !end) {
+    AsmConstWalker walker(wasm, minimizeWasmChanges);
+    walker.process();
+    return walker.asmConsts;
+  }
```
tlively, line 359: Are planning to remove this fallback code in the future? What is stopping us from removing it now?

test/lit/wasm-emscripten-finalize/em_asm.wat
```diff
@@ -0,0 +1,29 @@
+;; Test that em_asm string are extraced correctly when the __start_em_asm
```
tlively, line 1:
```suggestion
;; Test that em_asm string are extracted correctly when the __start_em_asm
```

test/lit/wasm-emscripten-finalize/em_js.wat
```diff
@@ -3,20 +3,32 @@
...
 ;; RUN: wasm-emscripten-finalize %s -S | filecheck %s
 
-;; Both functions should be stripped from the binary
+;; All functions should be stripped from the binary, regardless
+;; of internal name
 ;; CHECK-NOT:  (func
 
+;; The data section that contains only em_js strings should
+;; be stripped.
+;; CHECK-NOT: (i32.const 512) "Only em_js strings here\00")
```
tlively, line 12: Would this change correctly test that the entire segment is stripped, rather than just changed?
```suggestion
;; CHECK-NOT: (i32.const 512)
```

test/lit/wasm-emscripten-finalize/em_js.wat
```diff
@@ -3,20 +3,32 @@
...
+
+;; Data sections that also contain other stuff should not be stripped
+;; CHECK: (data (i32.const 1024) "some JS string data\00xxx")
+;; CHECK: (data (i32.const 2048) "more JS string data\00yyy")
```
tlively, line 16: Why aren't the JS strings zeroed out?

sbc100: I've only done the zeroing out thing for the em_asm path so far.     The reason is that I don't currently use the start/end symbols for the em_js path at all.   I think we can do that a followup perhaps, and maybe even completely remove the current 1-export-per-em-js thing.

src/wasm/wasm-emscripten.cpp
sbc100: Yes.   Mostly its a transition thing.    I can't land the emscripten-side change first because it actually changes the way that arg0 gets passed the em_asm functions.   llvm ends up storing that address of the string in memory and then loading it back (at least in -O0) which is not currently supported by the AsmConstWalker.   So if I land the emscripten side first it would break this pass.

test/lit/wasm-emscripten-finalize/em_js.wat
tlively: Gotcha. The difference between EM_JS and EM_ASM still confuses me, but that's ok.

tlively (approved):

PR #3558, opened 2021-02-09
Simplify asmConst handling. NFC.

sbc100:
Support for multiple signatures per JS code string was removed in #2422.

emscripten now only needs to know that address and the body of the JS
function.
    
See https://github.com/emscripten-core/emscripten/pull/13452 for the
emscripten side that ignores this data.


kripken (approved):
Nice!

PR #3559, opened 2021-02-09
[GC] Do not crash in GC allocations in interpreter; trap

kripken:
The spec does not mention traps here, but this is like a JS VM trapping on
OOM - a runtime limitation is reached.

As these are not specced traps, I did not add them to `effects.h`. Note how
as a result the optimizer happily optimizes into a nop an unused allocation of an
array of size `unsigned(-1)`, which is the behavior we want.


tlively (commented):
I'm concerned about this approach for a couple reasons. First, it introduces non-determinism into Binaryen's optimizations, which could introduce differences between local dev builds and remote release builds that would be hard to account for. Second, I wouldn't be surprised if there exist Wasm engines that can successfully allocate a 4GB array, so these optimizations would not be correct for modules that run on such engines.

Can we instead have a user-configurable size limit above which the interpreter will fail? IMO those failures should be graceful and may result in different optimization choices being made, but should not result in allocations being optimized out or any other visible side effects. Also, it seems that we should account for these traps in effects.h no matter what approach we take in the interpreter.

kripken:
Good point about determinism. I think that may be the better approach.

I disagree about `effects.h`, however. Adding it there would prevent the optimizer removing a dropped `array.new` for example. While that is *technically* an observable side effect, it is one we should ignore. That is, it's not like an `i32.load` which is a possible trap we really can't remove. A major difference between them is that program correctness cannot be lost by removing an allocation. With that said, I agree we don't have a fully precise distinction between the two, and just not mentioning it in `effects.h` works but is not elegant...

tlively:
Good point about program correctness. Allocation instructions have the obvious side effect of producing values, so the optimizer would only ever remove them if their values were not necessary. I also agree that simply not mentioning them in effects.h is less than elegant, but it's probably not worth spending extra energy on. How about we just add a TODO comment to effects.h about it?

kripken:
Thinking on this, I believe that it is indeed the right thing to not mark VM limitation traps as effects. I added a long comment now that explains the reasoning.

kripken:
And thinking on the determinism issue, I'm not sure how to fix it. We must catch `std::bad_alloc` as this PR does, to avoid an even worse outcome, so all we can do is add an artificial limit on top. But there is no natural way to do it - we may allocate in lots of different places, in different ways. Perhaps we should interpose between malloc? But we just want "user-driven" allocations to be limited, as otherwise a small artificial limit would hit us constantly (even a large one would, eventually). But marking "user-driven" allocations would require a very large refactoring.

tlively:
What about a similar mechanism to the `WebLimitations` we have now? We can just make it a validation error to have enormous structs. Large allocations for arrays should be limited to the interpreter where we can dynamically enforce the limit unless I'm missing something.

kripken:
We can add an arbitrary validation limit, yes. If so we could coordinate with VMs like on the existing limits, good point.

I'm not sure what you mean by "dynamically enforce the limit"? The validator can rule out structs past a certain size, but if you have a struct half that size it will validate - are you saying the interpreter should check for two allocations of that size? If so that seems possible, but it's not going to be precise (there are constant overheads per struct, etc.). That might be good enough for most things, but catching `std::bad_alloc` still seems like a good practice since otherwise we have no guarantees.

tlively:
Oh, I was just thinking about arrays with that comment. If an array would be allocated with a size greater than the limit, the interpreter would signal a trap. Having a specified limit instead of depending on a `std::bad_alloc` being thrown makes it deterministic at least.

If an allocation failure ever occurs (for example when allocating a couple very large structs or a bajillion reasonably-sized structs, I don't think there's a good reason to try to recover from that. The user should just buy more memory at that point ;) For comparison, we would encounter this type of problem already if there were too many tables or memories.

src/ir/effects.h
```diff
@@ -73,9 +72,23 @@ class EffectAnalyzer {
...
+  // official semantics, and they can occur anywhere: *any* instruction could in
+  // theory be implemented by a VM call (as will be the case when running in an
+  // interpreter), and such a call could run out of stack or memory in
+  // principle. To put it another way, an i32 division by zero is the program
+  // doing something bad that causes a trap, but the VM running out of memory is
+  // the VM doing something bad - and therefore the VM behaving in a way that is
+  // not according to the wasm semantics - and we do not model such things. Note
+  // that as a result we do not mark things like GC allocation instructions as
+  // having side effects, which makes it possible to eliminate an allocation
+  // whose result is not captured, which is beneficial.
```
tlively, line 91:
```suggestion
  // having side effects, which would make it possible to eliminate an allocation
  // whose result is not captured, which would be beneficial.
```

kripken:
Are you suggesting a specific arbitrary limit for array sizes, then, and nothing else? I'm not sure I see the value of special-casing things that way.

tlively:
No, I would expect the size limit to be the same for arrays and for structs. But you can't tell how big arrays will be statically, so the limit would have to be dynamically enforced in the interpreter. For structs, you statically know their size so you can enforce the limit at validation time.

kripken:
Then I have the same question, but slightly expanded. What is the benefit of adding a specific check, like of struct and array size, when there are many other possible OOM causes? I'm not sure what whack-a-mole buys here, but maybe I'm missing something...

tlively:
Yeah that's fair. I was assuming that the interpreter was much more likely than other parts of the code base to encounter large allocations and was therefore worth updating to deterministically avoid large allocations. I was also assuming that large struct and array allocations would be much more common than other large allocations. But perhaps that would not be the case.  If we can handle common cases deterministically, then we should, but if we can't or if there are no common cases, then it's probably not worth doing anything special.

kripken:
Returning to this, and thinking on it again, I am still not happy with an arbitrary limit, but the current PR increases the usage of exceptions in our codebase, which we'd like to reduce for other reasons. So I think I've changed over to the side of a fixed limit. I did my best to find a non-arbitrary one, as a too-small limit will end up annoying eventually (unless it is ecosystem-wide like WebLimitations). PTAL at the updated PR.

tlively (approved):
Did you want to put a similar check in for `struct.new` as well?

kripken:
I expect this is just a problem for arrays, as I think a struct type with that many fields would hit another limit in the type section anyhow (also, we don't create such types in the fuzzer, and we'll have a limit when we do I assume).

Issue #3497, opened 2021-01-15
How to implement delegate's target


aheejin:
The following is the new instruction `delegate`'s semantics, taken from my LLVM CL's comment:
```
Linearizing the control flow by placing TRY / END_TRY markers can create
mismatches in unwind destinations for throwing instructions, such as calls.

We use the 'delegate' instruction to fix the unwind mismatches. 'delegate'
instruction delegates an exception to an outer 'catch'. It can target not
only 'catch' but all block-like structures including another 'delegate',
but with slightly different semantics than branches. When it targets a
'catch', it will delegate the exception to that catch. It is being
discussed how to define the semantics when 'delegate''s target is a non-try
block: it will either be a validation failure or it will target the next
outer try-catch. But anyway our LLVM backend currently does not generate
such code. The example below illustrates where the 'delegate' instruction
in the middle will delegate the exception to, depending on the value of N.
try
  try
    block
      try
        try
          call @foo
        delegate N    ;; Where will this delegate to?
      catch           ;; N == 0
      end
    end               ;; N == 1 (invalid; will not be generated)
  delegate            ;; N == 2
catch                 ;; N == 3
end
                      ;; N == 4 (to caller)
```

So `delegate`'s immediate field can target all control flow structures including `block`s and `loop`s, like `br`s. But if the target is `block` or `loop`, it will be a validation failure. When it targets a `try`, it is really targetting not `try` but its `catch`.

Then here comes the catch: Currently `if` and `try` do not take label names; they use wrapping blocks or inner blocks to achieve the same semantics, mostly because there are too many places to handle if we add a new control flow structure that can have a label. But if we want to give `delegate` a label, it should target a `try` (more precisely its `catch`), and if it targets a `block`, it will be a validation failure. But I'm not sure if we want to give `try` a label and add its handling in a bunch of places, which is exactly what we wanted to avoid when we implemented `if` and `try` this way.

We are already doing some ugly things to implement try-catch. 
1. We create an inner block within `catch`, but we assume that block is removed when we print the text/binary. The reason it should be removed is catch body should start with a `pop` instruction (or a value on the stack in binary), and `block` should not interfere in between. This is easily done when there's no branch that targets the inner block.
2. When there is `br 0` within a `catch`, we wrap the whole try-catch with a `block` and make the `br` target it, not the inner block within `catch`, for the reason I described in 1. (#3129)

[...]

PR #3561, opened 2021-02-10
[EH] Support reading/writing of delegate

aheejin:
This adds support for reading/writing of the new `delegate` instruction
in the folded wast format, the stack IR format, the poppy IR format, and
the binary format in Binaryen. We don't have a formal spec written down
yet, but please refer to WebAssembly/exception-handling#137 and
WebAssembly/exception-handling#146 for the informal semantics. In the
current version of spec `delegate` is basically a rethrow, but with
branch-like immediate argument so that it can bypass other
catches/delegates in between.

`delegate` is not represented as a new `Expression`, but it is rather
an option within a `Try` class, like `catch`/`catch_all`.

One special thing about `delegate` is, even though it is written
_within_ a `try` in the folded wat format, like
```wasm
(try
  (do
    ...
  )
  (delegate $l)
)
```
In the unfolded wat format or in the binary format, `delegate` serves as
a scope end instruction so there is no separate `end`:
```wasm
try
  ...
delegate $l
```

`delegate` semantically targets an outer `catch` or `delegate`, but we
write `delegate` target as a `try` label because we only give labels to
block-like scoping expressions. So far we have not given `Try` a label
and used inner blocks or a wrapping block in case a branch targets the
`try`. But in case of `delegate`, it can syntactically only target `try`
and if it targets blocks or loops it is a validation failure.

So after discussions in #3497, we give `Try` a label but this label can
only be targeted by `delegate`s. Unfortunately this makes parsing and
writing of `Try` expression somewhat complicated. Also there is one
special case; if the immediate argument of `try` is the same as the
depth of control flow stack, this means the 'delegate' delegates to the
caller. To handle this case this adds a fake label
`DELEGATE_CALLER_TARGET`, and when writing it back to the wast format
writes it as an immediate value, unlike other cases in which we write
labels.

This uses `DELEGATE_FIELD_SCOPE_NAME_DEF/USE` to represent `try`'s label
and `delegate`'s target. There are many cases that `try` and
`delegate`'s labels need to be treated in the same way as block and
branch labels, such as for hashing or comparing. But there are routines
in which we automatically assume all label uses are branches. I thought
about adding a new kind of defines such as
`DELEGATE_FIELD_TRY_NAME_DEF/USE`, but I think it will also involve some
duplication of existing routines or classes. So at the moment this PR
chooses to use the existing `DELEGATE_FIELD_SCOPE_NAME_DEF/USE` for
`try` and `delegate` labels and makes only necessary amount of changes
in branch-utils. We can revisit this decision later if necessary.

Many of changes to the existing test cases are because now all `try`s
are automatically assigned a label. They will be removed in
`RemoveUnusedNames` pass in the same way as block labels if not targeted
by any delegates.

This only supports reading and writing and has not been tested against
any optimization passes yet.

---

Original unfolded wat file to generate test/try-delegate.wasm:
```wasm
(module
  (event $e)
  (func
    try
      try
      delegate 0
    catch $e
    end)

  (func
    try
      try
      catch $e
        i32.const 0
        drop
        try
        delegate 1
      end
    catch $e
    end
  )
)
```


test/try-delegate.wasm.fromBinary
```diff
@@ -0,0 +1,23 @@
+(module
```
aheejin, line 1: The original try-delegate.wasm file is this:
```wasm
try
  try 
  delegate 0
catch 0
  try
  delegate 0
end
```
This is to test whether targeting the depth 0 in a `try` body or `catch` body works, which is the innermost enclosing `try`. When parsing we create a block within each of `try` and `catch` body, but we shouldn't make those `delegate`s target those blocks because it will be a validation failure, so we have to remap those `delegate`s to the correct `try`.

test/exception-handling.wast
```diff
@@ -102,6 +102,35 @@
...
+            (throw $e-i32 (i32.const 0))
+          )
+          (catch $e-i32
+            (drop (pop i32))
+          )
+          (catch_all)
+        )
+      )
+    )
+
```
aheejin, line 133: This is not related to `delegate` but I just found out we don't have any tests that test nested try-catch, so I added one here

src/wasm/wasm-validator.cpp
```diff
@@ -276,11 +277,52 @@ struct FunctionValidator : public WalkerPass<PostWalker<FunctionValidator>> {
...
+  //   (do ... )
+  //   (catch $e
+  //     (try
+  //       (do ...)
+  //       (delegate $l0) ;; validation failure
+  //     )
+  //   )
+  // )
+  // Unlike branches, if delegate's target 'catch' is located above the
+  // delegate, it is a validation failure.
```
kripken, line 299: To be sure I understand, is this because it could be an infinite loop otherwise?

src/wasm/wasm-validator.cpp
```diff
@@ -2059,6 +2111,21 @@ void FunctionValidator::visitTry(Try* curr) {
...
+  shouldBeFalse(curr->isCatch() && curr->isDelegate(),
+                curr,
+                "try cannot have both catch and delegate at the same time");
+  shouldBeTrue(curr->isCatch() || curr->isDelegate(),
+               curr,
+               "try should have either catches or a delegate");
+
+  // (try $l (delegate $l)) is invalid. So we pop the current try's name before
+  // checking the delegate target.
+  delegateTargetNames.erase(curr->name);
```
kripken, line 2124: is this needed? I think it will have already been erased in `visitPreCatch`.

src/wasm-s-parser.h
```diff
@@ -242,7 +242,7 @@ class SExpressionWasmBuilder {
       i++;
     }
   }
-  Name getLabel(Element& s);
+  Name getLabel(Element& s, bool isBranch = true);
```
kripken, line 245: How about an enum to avoid the bool, like
```
enum class LabelType {
  Break
  Delegate
}
```
?

src/wasm/wasm-s-parser.cpp
```diff
@@ -1991,6 +1996,14 @@ Expression* SExpressionWasmBuilder::makeRefEq(Element& s) {
 // Any number of catch blocks can exist, including none. Zero or one catch_all
 // block can exist, and if it does, it should be at the end. There should be at
 // least one catch or catch_all body per try.
+//
+// try-delegate is written in the forded format as
```
kripken, line 2000:
```suggestion
// try-delegate is written in the folded format as
```

src/passes/Print.cpp
```diff
@@ -1955,6 +1959,8 @@ struct PrintSExpression : public OverriddenVisitor<PrintSExpression> {
   Function::DebugLocation lastPrintedLocation;
   bool debugInfo;
 
+  int controlFlowDepth = 0;
```
kripken, line 1962: please add a comment saying what this is used for

kripken (commented):
Nice! This is simpler and cleaner than I'd thought could be done, given the complexity.

Basically lgtm, aside from the comments.

aheejin:
Sorry I discovered there's a bug in binary parsing algorithm, in that it can't handle some cases. I'll probably fix that tomorrow

aheejin:
> Sorry I discovered there's a bug in binary parsing algorithm, in that it can't handle some cases. I'll probably fix that tomorrow

Oh no it seems to be fine after all...

src/wasm/wasm-validator.cpp
aheejin: That can be one reason, but I haven't thought about that actually. The motivation (for the spec) was just it was unnatural both intuitively and for the VM implementation to delegate upwards; it looks like we are branching upwards.

src/wasm/wasm-validator.cpp
aheejin: I think you're right. Removed.

src/wasm-s-parser.h
aheejin: Done

src/passes/Print.cpp
aheejin: Done

kripken (approved):

src/wasm/wasm-binary.cpp
```diff
@@ -5839,16 +5829,91 @@ void WasmBinaryBuilder::visitTryOrTryInBlock(Expression*& out) {
...
+  //   (try
+  //     ...
+  //     (catch $e
+  //       (block   ;; Now this can be deleted when writing binary
+  //         ...
+  //         (br $label0)
+  //       )
+  //     )
+  //   )
+  // )
```
aheejin, line 5909: This comment block was just moved

PR #3563, opened 2021-02-12
Print the features section in a comment

kripken:
This can help in debugging issues with the features section.

As it happens we already had a wasm using the features section
in the test suite, so no addition was needed here...

### DIFF
--- a/src/wasm-features.h
+++ b/src/wasm-features.h
@@ -75,6 +75,21 @@ struct FeatureSet {
     }
   }
 
+  std::string toString() {
+    std::string ret;
+    uint32_t x = 1;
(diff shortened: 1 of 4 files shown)


src/wasm-features.h
```diff
@@ -75,6 +75,21 @@ struct FeatureSet {
     }
   }
 
+  std::string toString() {
+    std::string ret;
+    uint32_t x = 1;
+    while (x != Feature::All + 1) {
```
tlively, line 81:
```suggestion
    while (x & Feature::All) {
```

tlively (approved):

PR #3564, opened 2021-02-12
Allow specifying additional features past the features section

kripken:
Perhaps I'm missing something, but is there a reason not to allow this?
That is, if a wasm says "simd", it seems ok to let the user specify simd
as well as more features, and the the optimizer can perhaps do something
with them.


tlively (approved):
Yes, this seems useful and unproblematic to me 👍

PR #3565, opened 2021-02-12
[EH] Update C and binaryen.js API for delegate

aheejin:
This updates C and binaryen.js API to match the new `Try` structure to
support `delegate`, added in #3561. Now `try` can take a name (which can
be null) like a block, and also has an additional `delegateTarget` field
argument which should only be used for try-delegate and otherwise null.

This also adds several more variant of `makeTry` methods in
wasm-builder. Some are for making try-delegate and some are for
try-catch(_all).

### DIFF
--- a/src/wasm-builder.h
+++ b/src/wasm-builder.h
@@ -637,81 +637,64 @@ public:
     ret->finalize();
     return ret;
   }
-  Try* makeTry(Expression* body,
-               const std::vector<Name>& catchEvents,
-               const std::vector<Expression*>& catchBodies) {
-    auto* ret = wasm.allocator.alloc<Try>();
-    ret->body = body;
-    ret->catchEvents.set(catchEvents);
-    ret->catchBodies.set(catchBodies);
-    ret->finalize();
-    return ret;
-  }
+private:
   Try* makeTry(Name name,
                Expression* body,
                const std::vector<Name>& catchEvents,
-               const std::vector<Expression*>& catchBodies) {
+               const std::vector<Expression*>& catchBodies,
+               Name delegateTarget,
+               Type type,
+               bool hasType) { // differentiate whether a type was passed in
     auto* ret = wasm.allocator.alloc<Try>();
     ret->name = name;
     ret->body = body;
     ret->catchEvents.set(catchEvents);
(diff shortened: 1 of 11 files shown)


src/wasm-builder.h
```diff
@@ -658,6 +670,49 @@ class Builder {
...
+    return ret;
+  }
+  Try* makeTry(Name name, Expression* body, Name delegateTarget, Type type) {
+    auto* ret = wasm.allocator.alloc<Try>();
+    ret->name = name;
+    ret->body = body;
+    ret->delegateTarget = delegateTarget;
+    ret->finalize(type);
+    return ret;
+  }
```
kripken, line 715: This ends up with duplication in all the variants of `makeTry`. How about one main method that is called by the others? Then each variant can be a single line.
```cpp
private:

Try* makeTry(Name name,
             Expression* body,
             const std::vector<Name>& catchEvents,
             const std::vector<Expression*>& catchBodies,
             Name delegateTarget,
             Type type,
             bool hasType) { // differentiate whether a type was passed in
  auto* ret = wasm.allocator.alloc<Try>();
  ret->name = name;
  ret->body = body;
  ret->catchEvents.set(catchEvents);
  ret->catchBodies.set(catchBodies);
  if (hasType) {
    ret->finalize(type);
  } else {
    ret->finalize();
  }
  return ret;
}

public:

Try* makeTry(Expression* body,
             const std::vector<Name>& catchEvents,
             const std::vector<Expression*>& catchBodies) {
  return makeTry(Name(), body, catchEvents, catchBodies, Name(), Type::none, false);
}

Try* makeTry(Expression* body,
             const std::vector<Name>& catchEvents,
             const std::vector<Expression*>& catchBodies,
             Type type) {
  return makeTry(Name(), body, catchEvents, catchBodies, Name(), type, true);
}

[..]

Try* makeTry(Name name, Expression* body, Name delegateTarget, Type type) {
  return makeTry(name, body, {}, {}, delegateTarget, type, true);
}
```

tlively (approved):
LGTM with @kripken's comment as well.

src/wasm-builder.h
aheejin: Thanks! Done.

PR #3568, opened 2021-02-14
[EH] Make rethrow's target a try label

aheejin:
I was previously mistaken about `rethrow`'s argument rule and thought
it only counted `catch`'s depth. But it turns out it follows the same
rule `delegate`'s label: the immediate argument follows the same rule as
when computing branch labels, but it only can target `try` labels
(semantically it targets that `try`'s corresponding `catch`); otherwise
it will be a validation failure. Unlike `delegate`, `rethrow`'s label
denotes not where to rethrow, but which exception to rethrow. For
example,
```wasm
try $l0
catch ($l0)
  try $l1
  catch ($l1)
   rethrow $l0 ;; rethrow the exception caught by 'catch ($l0)'
  end
end
```

Refer to this comment for the more detailed informal semantics:
https://github.com/WebAssembly/exception-handling/issues/146#issuecomment-777714491

---

This also reverts some of `delegateTarget` -> `exceptionTarget` changes
done in #3562 in the validator. Label validation rules apply differently
for `delegate` and `rethrow` for try-catch. For example, this is valid:
```wasm
try $l0
  try
  delegate $l0
catch ($l0)
end
```
But this is NOT valid:
```wasm
try $l0
catch ($l0)
  try
  delegate $l0
end
```
So `try`'s label should be used within try-catch range (not catch-end
range) for `delegate`s.

But for the `rethrow` the rule is different. For example, this is valid:
```wasm
try $l0
catch ($l0)
  rethrow $l0
end
```
But this is NOT valid:
```wasm
try $l0
  rethrow $l0
catch ($l0)
end
```
So the `try`'s label should be used within catch-end range instead.

### DIFF
--- a/src/wasm/wasm-s-parser.cpp
+++ b/src/wasm/wasm-s-parser.cpp
@@ -2067,37 +2067,14 @@ Expression* SExpressionWasmBuilder::makeTry(Element& s) {
     // We create a different name for the wrapping block, because try's name can
     // be used by internal delegates
     block->name = nameMapper.pushLabelName(sName);
-    // For simplicity, try's name canonly be targeted by delegates. Make the
-    // branches target the new wrapping block instead.
+    // For simplicity, try's name can only be targeted by delegates and
+    // rethrows. Make the branches target the new wrapping block instead.
     BranchUtils::replaceBranchTargets(ret, ret->name, block->name);
     block->list.push_back(ret);
     nameMapper.popLabelName(block->name);
     block->finalize(type);
     return block;
   }
-  return ret;
-}
-
-Expression*
-SExpressionWasmBuilder::makeTryOrCatchBody(Element& s, Type type, bool isTry) {
-  if (isTry && !elementStartsWith(s, "do")) {
-    throw ParseException("invalid try do clause", s.line, s.col);
-  }
-  if (!isTry && !elementStartsWith(s, "catch") &&
-      !elementStartsWith(s, "catch_all")) {
-    throw ParseException("invalid catch clause", s.line, s.col);
-  }
-  if (s.size() == 1) { // (do) / (catch) / (catch_all) without instructions
-    return makeNop();
-  }
-  auto ret = allocator.alloc<Block>();
-  for (size_t i = 1; i < s.size(); i++) {
-    ret->list.push_back(parseExpression(s[i]));
-  }
-  if (ret->list.size() == 1) {
-    return ret->list[0];
-  }
-  ret->finalize(type);
   return ret;
 }
 
@@ -2118,7 +2095,7 @@ Expression* SExpressionWasmBuilder::makeThrow(Element& s) {
 
 Expression* SExpressionWasmBuilder::makeRethrow(Element& s) {
   auto ret = allocator.alloc<Rethrow>();
-  ret->depth = atoi(s[1]->str().c_str());
+  ret->target = getLabel(*s[1], LabelType::Exception);
   ret->finalize();
   return ret;
 }
--- a/src/wasm/wasm-validator.cpp
+++ b/src/wasm/wasm-validator.cpp
@@ -2541,6 +2541,9 @@ void FunctionValidator::visitFunction(Function* curr) {
       "function result must match, if function has returns");
   }
 
+  assert(breakInfos.empty());
+  assert(delegateTargetNames.empty());
+  assert(rethrowTargetNames.empty());
   returnTypes.clear();
   labelNames.clear();
   // validate optional local names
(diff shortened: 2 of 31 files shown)


src/wasm/wasm-s-parser.cpp
```diff
@@ -2078,29 +2078,6 @@ Expression* SExpressionWasmBuilder::makeTry(Element& s) {
...
-  for (size_t i = 1; i < s.size(); i++) {
-    ret->list.push_back(parseExpression(s[i]));
-  }
-  if (ret->list.size() == 1) {
-    return ret->list[0];
-  }
-  ret->finalize(type);
-  return ret;
-}
-
```
aheejin, line 2103: This method was already unused; it became unused in #3487 but wasn't deleted there.

test/passes/dce_all-features.wast
```diff
@@ -790,17 +791,16 @@
...
-              (rethrow 0)
-            )
+    (try $l0
+      (do)
+      (catch $e
+        (drop
+          ;; This i32.add will be dce'd
+          (i32.add
+            (i32.const 0)
+            (rethrow $l0)
```
aheejin, line 801: This function had to change because now `rethrow` should be within a `catch`.

src/wasm/wasm-validator.cpp
```diff
@@ -2073,14 +2076,20 @@ void FunctionValidator::visitRefEq(RefEq* curr) {
     "ref.eq's right argument should be a subtype of eqref");
 }
 
-void FunctionValidator::noteException(Name name, Expression* curr) {
+void FunctionValidator::noteDelegate(Name name, Expression* curr) {
   if (name != DELEGATE_CALLER_TARGET) {
-    shouldBeTrue(exceptionTargetNames.find(name) != exceptionTargetNames.end(),
+    shouldBeTrue(delegateTargetNames.find(name) != delegateTargetNames.end(),
```
tlively, line 2081: This could also be `delegateTargetNames.count(nam) != 0`, which is slightly shorter.

src/wasm/wasm-validator.cpp
```diff
@@ -2533,9 +2543,12 @@ void FunctionValidator::visitFunction(Function* curr) {
 
   shouldBeTrue(
     breakInfos.empty(), curr->body, "all named break targets must exist");
-  shouldBeTrue(exceptionTargetNames.empty(),
+  shouldBeTrue(delegateTargetNames.empty(),
                curr->body,
                "all named delegate targets must exist");
+  shouldBeTrue(rethrowTargetNames.empty(),
+               curr->body,
+               "all named rethrow targets must exist");
```
tlively, line 2551: Couldn't these only fail if there was a bug in the validator itself? If so, should these be asserts instead?

tlively (approved):
LGTM besides those few comments.

kripken (approved):

src/wasm/wasm-validator.cpp
aheejin: Done

src/wasm/wasm-validator.cpp
aheejin: Isn't that the same for all checking points within the validator? We already have similar lines:
https://github.com/WebAssembly/binaryen/blob/1d3f578d5dc26fc8ea83ea851ac0f1100a3cbad6/src/wasm/wasm-validator.cpp#L2534-L2538

tlively: In other words, is it possible to construct an invalid Binaryen IR module that would trigger this error? (I would ask the same questions of any of the other checks we have)

aheejin: Yeah I think so. Some flawed pass can assign a `block` label to `delegate`, or even a random string (unlikely), but anyway I don't think it's impossible.

tlively: But that would be caught by https://github.com/WebAssembly/binaryen/pull/3568/files#diff-2149be365b66a82038f1d3fa8f9fb1b4dcd40ab535c986f4bf0a4c37e669a1ceR2079-R2092, right? These checks are just checking that the target names were properly removed from their sets at the end of the corresponding scopes AFAICT, and it shouldn't be possible to create an invalid Binaryen module in which a try scope or catch scope never ends.

aheejin: Hmm, I think you're right. Then maybe I'll remove all those checks..?

aheejin: Removed.

aheejin: Oh no, you suggested to make them assertions... Will change them to assertions.

tlively (approved):

PR #3570, opened 2021-02-16
Fix removal of em_js strings

sbc100:


tlively (approved):
Oops! That looks like it was probably annoying to track down.

PR #3572, opened 2021-02-17
[reference-types] Support passive elem segments

martianboy:
Passive element segments do not belong to any table, so the link between
Table and elem needs to be weaker; i.e. an elem may have a table in case
of active segments, or simply be a collection of function references in 
case of passive/declarative segments.

This PR takes Table::Segment out and turns it into a first class module
element just like tables and functions. It also implements early support
for parsing, printing, encoding and decoding passive/declarative elem
segments.


martianboy:
@tlively @kripken @aheejin This is an early wip, and is still buggy/broken. I just wanted to get early comments on the architectural changes made in `wasm.h` and see if there are reservations/concerns about them. In particular, I haven't made up my mind about keeping `Table.segments` to reference their active elem segments. Appreciate your thoughts and comments.

tlively:
Thanks @martianboy! I will try to review this in the next few days, but from the direction you describe in the PR description sounds good to me 👍

martianboy:
@tlively This should be ready for a review! I'm just going over the changes myself to see if I haven't missed anything. I might be missing a few tests somewhere in case of modified functionality.

src/wasm.h
```diff
@@ -1720,33 +1722,41 @@ class Export {
   ExternalKind kind;
 };
 
+struct ElemSegment : Named {
+  enum Status { ACTIVE = 0, PASSIVE = 1, DECLARATIVE = 2 };
```
tlively, line 1726: Our convention has been to call this kind of enum `Kind` and to use use `PascalCase` rather than all-caps for the values.

Declarative segments are just for predeclaring what functions can be the operands of `ref.func`, right? Since they don't add any semantic value of their own, it might be simpler to not store them at all in the IR and to recreate them as part of Module writing. If we can avoid storing declarative segments, then we could just use the presence of `Name table` to differentiate between active and passive segments, so we wouldn't need this enum at all.

tlively (commented):
I haven't finished reading everything yet, but here are some comments on the important bits to start out.

src/wasm.h
```diff
@@ -1720,33 +1722,41 @@ class Export {
...
-    Segment(Expression* offset) : offset(offset) {}
-    Segment(Expression* offset, std::vector<Name>& init) : offset(offset) {
-      data.swap(init);
-    }
-  };
-
   Address initial = 0;
   Address max = kMaxSize;
-  std::vector<Segment> segments;
+  std::vector<ElemSegment*> segments;
```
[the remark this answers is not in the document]
tlively: If we got rid of these vectors, we could have some helper iterators for iterating over all the segments of a particular table to help keep other code simple.

src/wasm.h
martianboy: Hey I like that idea a lot! But wouldn't that mean we have to loop over all `ref.func` expressions to reconstruct the declarative element segment? It's either that, or keeping track of all `ref.func` operands while constructing/modifying the IR, which I suspect is even more complicated.

src/binaryen-c.cpp
```diff
@@ -3360,11 +3337,13 @@ BinaryenTableRef BinaryenAddTable(BinaryenModuleRef module,
...
-  Table::Segment segment((Expression*)offset);
+  auto elem = std::make_unique<ElemSegment>(name, (Expression*)offset);
+  elem->setName(Name::fromInt(((Module*)module)->elemSegments.size()), false);
   for (BinaryenIndex i = 0; i < numFuncNames; i++) {
-    segment.data.push_back(funcNames[i]);
+    elem->data.push_back(funcNames[i]);
   }
-  table->segments.push_back(segment);
+  table->segments.push_back(elem.get());
+  ((Module*)module)->addElemSegment(std::move(elem));
```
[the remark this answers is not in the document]
martianboy: Sure, I'm going to dedicate a separate PR to the public APIs as they are lacking in many areas related to tables. I'll fix this one here though. 👍🏼

src/wasm.h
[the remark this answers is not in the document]
martianboy: This was my original idea to drop `table.segments` completely, but then I thought of the various passes and decided it might be safer to keep it first and then see if it can be dropped after. I'll go over the changes again and check. 👍🏼

src/wasm.h
tlively: Yes, we would have to run an analysis pass over the module to find all the operands of `ref.func` instructions. We already do something similar to collect all the types that need to be emitted into the type section, so you could use something similar. Check out the implementation and uses of `ModuleUtils::collectHeapTypes`.

martianboy: Yeah great, I've already refactored the changes to remove `Table.segments` and also tried to eliminate all use cases of this `Status` enum. Reconstructing the declarative segment should be easy now.

martianboy:
@tlively Regarding `module-splitting.cpp`, I was wondering how the presence of passive elements and table instructions might affect the `TableSlotManager`. Right now it uses an existing table and have some restrictions with elements that have a global offset. But table instructions allow arbitrary modifications on the contents of a table at runtime, so an existing table would probably not be a safe choice anymore. Wouldn't it be simpler if we created a separate table for splitting purposes? We could check for existence of passive elements to do that, so we don't create unnecessary tables in simpler cases.

tlively:
Yes, when reference-types is enabled, it would be great to generate a fresh table for module splitting. No need to make functional changes like that in this PR, though. The module-splitting code is extremely experimental, so it's ok if it makes unsafe assumptions for now.

martianboy:
@tlively One thing I noticed while implementing `Module::collectReferencedFunctions`, was that with the naive approach of looping over all `ref.func`s, in a lot of cases we're creating a declarative elem segment without a good reason. If I'm not mistaken, the reason for the introduction of declarative segments (which was [discussed here](https://github.com/WebAssembly/reference-types/issues/31)) was to declare functions that are implicitly exported via another exported function that returns a funcref, or a `table.set` instruction on imported/exported tables.

So it appears that we can improve the naive approach a bit further by looping over exported functions that return a `funcref` (directly, or indirectly as part of a tuple or struct, right?), walk the function body and extract all `ref.func`ed functions to put in the declared segment. But it also sounds more complex.

What do you think? Can you think of a simpler approach that might have similar benefits?

tlively:
If I understand correctly,`ref.fun`c is only valid if the target function appears in some table, whether or not the `funcref` actually escapes the module via exported function or some other mechanism. To ensure that we emit valid modules, we then have to ensure that we have a declarative segment containing all function operands to `ref.func` _that are not already in some other segment_.

martianboy:
Oh interesting, the spec is a bit vague/hard to understand. Thanks for the explanation. In that case, I'll just have to exclude functions that are already in another elem segment.

martianboy:
Hey @dcodeIO, can you please take a look at the API changes in this PR and let me know what you think? I'm not very happy with it, having to pass a ref in some apis and an index in others. But I wanted to hear your thoughts first.

martianboy:
@tlively Sorry for the delay here, had a surgery last week. I think this PR is ready for a review aside from the missing JS APIs which I'll be adding later today. Sorry for adding more to the scope with the new C APIs, but a PR got merged in that prevented this PR to get away without at least a few meaningful APIs for elems.

By the way, something that came up during the API changes was that since we don't store the declarative segment in the IR, we can't offer an API to manually manipulate such a segment. One way would be to offer an API to generate the declarative segment (just like what is done automatically when writing or printing). WDYT? cc @dcodeIO

src/binaryen-c.h
```diff
@@ -2051,22 +2051,15 @@ BINARYEN_API void BinaryenRemoveEvent(BinaryenModuleRef module,
...
 
-// TODO: Add support for multiple segments in BinaryenSetFunctionTable.
-BINARYEN_API void BinaryenSetFunctionTable(BinaryenModuleRef module,
-                                           BinaryenIndex initial,
-                                           BinaryenIndex maximum,
-                                           const char** funcNames,
-                                           BinaryenIndex numFuncNames,
-                                           BinaryenExpressionRef offset);
-BINARYEN_API int BinaryenIsFunctionTableImported(BinaryenModuleRef module);
+BINARYEN_API int BinaryenIsTableImported(BinaryenModuleRef module);
```
dcodeIO, line 2054: Wondering whether APIs assuming that there is a single function table in a module are still needed, once there can be multiple tables. My expectation would be that one can add multiple tables, and table operations, like checking or setting whether a table is imported, adding segments, walking segments etc. would then operate upon a `BinaryenTableRef`. Or, to obtain all segments in a module, one would walk over all tables, and then over each segment in each table.

CHANGELOG.md
```diff
@@ -42,6 +42,8 @@ Current Trunk
   - `module.getTableByIndex`
   - `module.getNumTables`
   - `binaryen.getTableInfo`
+- `BinaryenSetFunctionTable` and `module.setFunctionTable` have been removed
+  in favor of `BinaryenAddTable` and `module.addTable` respectively.
```
dcodeIO, line 46: This looks like a good change to me. I recently ran into issues with `BinaryenSetFunctionTable` yielding an "already exists" kind of error when a table had already been imported, and instead operating on a per-table basis seems like a desirable change.

src/binaryen-c.h
martianboy: You're right about this API, it should take a `BinaryenTableRef` for sure. Just note that since we also have passive elem segments that do not belong to a table, obtaining all segments by walking over all tables is not possible.

dcodeIO: I see, hmm. That would actually make in favor of representing segments the binary format way, as a list of segments belonging to a table or not, with some passive, plus perhaps a convenience API to walk all segments attached to a specific table, if that is possible. Hmmhmhm.

martianboy: That's right, that's how Binaryen will represent elem segments internally with this PR. I'm just not sure what can be done for iterating over the segments that belong to a specific table. The problem is that tables don't have references to their segments in the IR, so we have to iterate over all elem segments in the module. One thought was to return all elem refs for that table in an array, but I don't think any other API behaves like that.

dcodeIO: Agree, I guess the convenience API isn't strictly necessary in this PR, as long as iterating segments is generally possible. Can still look into that later, once we have a better idea what API users would actually need.

src/binaryen-c.cpp
```diff
@@ -3399,59 +3365,82 @@ BinaryenTableRef BinaryenGetTableByIndex(BinaryenModuleRef module,
   }
   return tables[index].get();
 }
-
-int BinaryenIsFunctionTableImported(BinaryenModuleRef module) {
-  if (((Module*)module)->tables.size() > 0) {
-    return ((Module*)module)->tables[0]->imported();
+BinaryenElemSegmentRef
+BinaryenAddActiveElemSegment(BinaryenModuleRef module,
```
dcodeIO, line 3369: What do you think of spelling out "Elem" as "Element" here and in similar places? Would feel a bit more consistent to me, since we also do not shorten "Function" to "Func" or similar in function names. Not feeling strongly, though.

martianboy: Hmm so `BinaryenAddActiveElementSegment` or `BinaryenAddActiveElement`? Making it consistent would require changing the name across the Binaryen IR as well. 🤔 I don't feel strongly either, it just makes statements longer, specially inside Binaryen.

martianboy: By the way, for this specific API, I can also provide a simpler version that can be used for both active and passive segments. So if null is provided for table name, I'll make it a passive segment otherwise it'll be active. My initial idea was that making it explicit might help with readability.

martianboy:
C & JS APIs are done now. I will add more APIs (specially setters) in a subsequent PR.

src/binaryen-c.cpp
tlively: I would support changing "Elem" to "Element" everywhere, but I also don't feel strongly about it as long we use either one of those consistently.

martianboy: FYI, I changed Elem to Element everywhere. Doesn't look bad.

tlively (commented):
I'm not quite done looking through this, but to prevent unnecessary delays, here are my comments so far :)

src/ir/module-splitting.cpp
```diff
@@ -87,20 +87,18 @@ namespace ModuleSplitting {
...
-      Index offset = 0;
-      if (auto* c = segment.offset->dynCast<Const>()) {
-        offset = c->value.geti32();
-      } else if (auto* g = segment.offset->dynCast<GlobalGet>()) {
-        base = g->name;
-      }
-      for (size_t i = 0; i < segment.data.size(); ++i) {
-        f(table->name, base, offset + i, segment.data[i]);
-      }
+  ModuleUtils::iterActiveElementSegments(module, [&](ElementSegment* elem) {
```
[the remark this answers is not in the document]
martianboy: Done.

src/ir/module-utils.h
```diff
@@ -420,6 +451,28 @@ template<typename T> struct CallGraphPropertyAnalysis {
   }
 };
 
+inline void collectReferencedFunctions(Module* wasm, std::set<Name>& funcs) {
+  struct RefFuncCollector : public WalkerPass<PostWalker<RefFuncCollector>> {
```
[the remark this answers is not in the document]
martianboy: Took it out per our conversation.

src/ir/table-utils.h
```diff
@@ -29,32 +30,36 @@ struct FlatTable {
...
+    ModuleUtils::iterTableSegments(wasm, table.name, [&](ElementSegment* elem) {
+      auto offset = elem->offset;
       if (!offset->is<Const>()) {
         // TODO: handle some non-constant segments
         valid = false;
         return;
       }
       Index start = offset->cast<Const>()->value.geti32();
-      Index end = start + segment.data.size();
+      Index end = start + elem->data.size();
```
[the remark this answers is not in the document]
martianboy: sure, you're right. done.

src/ir/table-utils.h
```diff
@@ -65,10 +70,10 @@ inline Table::Segment& getSingletonSegment(Table& table, Module& wasm) {
 // module has a single table segment, and that the dylink section indicates
 // we can validly append to that segment, see the check below.
 inline Index append(Table& table, Name name, Module& wasm) {
-  auto& segment = getSingletonSegment(table, wasm);
-  auto tableIndex = segment.data.size();
+  auto segment = getSingletonSegment(table, wasm);
```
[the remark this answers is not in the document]
martianboy: Thanks for clarifying that, I was wondering if there's a preferred style I should follow. Done.

src/js/binaryen.js-post.js
```diff
@@ -3001,12 +3008,36 @@ Module['getGlobalInfo'] = function(global) {
...
+        // for (let j = 0; j !== segmentLength; ++j) {
+        //   var ptr = Module['_BinaryenElementSegmentGetData'](module, i, j);
+        //   names[j] = UTF8ToString(ptr);
+        // }
+    //     arr.push({
+    //       'offset': Module['_BinaryenElementSegmentGetOffset'](module, i),
+    //       'names': names
+    //     });
+    //   }
+    //   return arr;
```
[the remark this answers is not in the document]
martianboy: Sorry, this was left from refactoring the js api. `getTableInfo` does not have access to the module, so it cannot find segments.

src/passes/FuncCastEmulation.cpp
```diff
@@ -173,17 +173,16 @@ struct FuncCastEmulation : public Pass {
...
-          auto iter = funcThunks.find(name);
-          if (iter == funcThunks.end()) {
-            auto thunk = makeThunk(name, module, numParams);
-            funcThunks[name] = thunk;
-            name = thunk;
-          } else {
-            name = iter->second;
-          }
+    for (auto& elem : module->elementSegments) {
+      // FIXME: should this check active elems only?
```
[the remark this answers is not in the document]
martianboy: Thanks, removed.

src/passes/LegalizeJSInterface.cpp
```diff
@@ -97,12 +97,11 @@ struct LegalizeJSInterface : public Pass {
...
         // from JS is legal for JS. Our stub makes it look like a native wasm
         // function.
-        for (auto& table : module->tables) {
-          for (auto& segment : table->segments) {
-            for (auto& name : segment.data) {
-              if (name == im->name) {
-                name = funcName;
-              }
+        for (auto& segment : module->elementSegments) {
+          // FIXME: Exclude declarative segments?
```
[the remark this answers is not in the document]
martianboy: This comment was made by me before deciding not to represent declared segments in the IR. It's now removed.

src/passes/Print.cpp
```diff
@@ -2676,36 +2676,75 @@ struct PrintSExpression : public UnifiedExpressionVisitor<PrintSExpression> {
...
+    o << '(';
+    printMedium(o, "elem");
+    if (curr->hasExplicitName) {
+      o << ' ';
+      printName(curr->name, o);
+    }
+
+    if (curr->table.is()) {
       // TODO(reference-types): check for old-style based on the complete spec
       if (currModule->tables.size() > 1) {
```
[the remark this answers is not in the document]
martianboy: It actually is, since elem segments always have a table name.

third_party/llvm-project/include/llvm/BinaryFormat/Wasm.h
```diff
@@ -147,7 +147,7 @@ struct WasmDataSegment {
   uint32_t Comdat; // from the "comdat info" section
 };
 
-struct WasmElemSegment {
+struct WasmElementSegment {
```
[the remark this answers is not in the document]
martianboy: My bad, replaced all! :) Fixed.

src/tools/fuzzing.h
```diff
@@ -532,22 +539,23 @@ class TranslateToFuzzReader {
...
-        if (auto* offset = segment.offset->dynCast<GlobalGet>()) {
-          if (!wasm.getGlobal(offset->name)->imported()) {
-            // TODO: the segments must not overlap...
-            segment.offset =
-              builder.makeConst(Literal::makeFromInt32(0, Type::i32));
+      ModuleUtils::iterTableSegments(
+        wasm, table->name, [&](ElementSegment* elem) {
+          // If the offset is a global that was imported (which is ok) but no
+          // longer is (not ok) we need to change that.
+          if (auto* offset = elem->offset->dynCast<GlobalGet>()) {
```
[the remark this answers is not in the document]
martianboy: Done.

src/passes/RemoveUnusedModuleElements.cpp
```diff
@@ -201,20 +214,24 @@ struct RemoveUnusedModuleElements : public Pass {
...
-    }
+    module->removeElementSegments([&](ElementSegment* curr) {
+      return curr->data.empty() ||
+             analyzer.reachable.count(ModuleElement(
+               ModuleElementKind::ElementSegment, curr->name)) == 0;
+    });
+    std::unordered_set<Name> nonemptyTables;
+    ModuleUtils::iterActiveElementSegments(*module, [&](ElementSegment* elem) {
+      if (nonemptyTables.count(elem->table) == 0) {
+        nonemptyTables.insert(elem->table);
```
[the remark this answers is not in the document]
martianboy: I added a comment above to make it more clear, but since we are removing empty element segments above this, all tables that have remaining segments need to be kept. Removed the redundant `if`.

martianboy:
@tlively This is now rebased on the current `main` branch.

test/binaryen.js/kitchen-sink.js
```diff
@@ -676,13 +676,14 @@ function test_core() {
...
-
+  
   module.removeTable("t1");
   assert(module.getNumTables() === 0);
-
-  // Legacy
-  module.setFunctionTable(1, 0xffffffff, [ binaryen.getFunctionInfo(sinker).name ]);
+  
+  module.addTable("0", 1, 0xffffffff);
+  module.addActiveElementSegment("0", "0", [ binaryen.getFunctionInfo(sinker).name ]);
```
tlively, line 684: It would be good to use different names for the table and segment so they can be verified to end up in the right place in the test output.

src/wasm/wasm-binary.cpp
```diff
@@ -557,40 +554,50 @@ void WasmBinaryWriter::writeTableElements() {
     return;
   }
 
-  BYN_TRACE("== writeTableElements\n");
+  std::set<Name> referencedFuncs;
```
tlively, line 557: Is this still used?

src/wasm/wasm-binary.cpp
```diff
@@ -557,40 +554,50 @@ void WasmBinaryWriter::writeTableElements() {
...
-                       (isDeclarative ? BinaryConsts::IsDeclarative : 0)
-                   : (hasTableIndex ? BinaryConsts::HasIndex : 0)) |
-        (usesExpressions ? BinaryConsts::UsesExpressions : 0);
-
-      o << U32LEB(flags);
+  for (auto& segment : wasm->elementSegments) {
+    Index tableIdx = 0;
+
+    bool isPassive = segment->table.isNull();
+    bool usesExpressions = false;
```
tlively, line 567: It looks like this is never changed from `false`?

src/wasm/wasm-binary.cpp
```diff
@@ -2679,64 +2712,72 @@ void WasmBinaryBuilder::readFunctionTableDeclaration() {
...
+  BYN_TRACE("== readElementSegments\n");
   auto numSegments = getU32LEB();
   if (numSegments >= Table::kMaxSize) {
     throwError("Too many segments");
   }
   for (size_t i = 0; i < numSegments; i++) {
     auto flags = getU32LEB();
     bool isPassive = (flags & BinaryConsts::IsPassive) != 0;
-    bool hasTableIdx = (flags & BinaryConsts::HasIndex) != 0;
+    bool hasTableIdx = !isPassive & ((flags & BinaryConsts::HasIndex) != 0);
```
tlively, line 2724:
```suggestion
    bool hasTableIdx = !isPassive && ((flags & BinaryConsts::HasIndex) != 0);
```

src/wasm/wasm-binary.cpp
```diff
@@ -2679,64 +2712,72 @@ void WasmBinaryBuilder::readFunctionTableDeclaration() {
...
+    bool isDeclarative =
+      isPassive & ((flags & BinaryConsts::IsDeclarative) != 0);
```
tlively, line 2726:
```suggestion
      isPassive && ((flags & BinaryConsts::IsDeclarative) != 0);
```

tlively (commented):
Nice! I finally got through everything here and it's looking pretty good :)

src/passes/Print.cpp
```diff
@@ -2665,36 +2665,62 @@ struct PrintSExpression : public UnifiedExpressionVisitor<PrintSExpression> {
...
 
       if (currModule->tables.size() > 1) {
-        o << " func";
+        o << ' ';
+        TypeNamePrinter(o, currModule).print(HeapType::func);
       }
-
-      for (auto name : segment.data) {
+    } else {
+      if (isDeclarative) {
```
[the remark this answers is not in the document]
martianboy: Well that would involve making an `ElementSegment` for the declared functions, which I suspect isn't worth it, although it will make the code simpler.

src/wasm/wasm-s-parser.cpp
```diff
@@ -3243,63 +3253,79 @@ void SExpressionWasmBuilder::parseTable(Element& s, bool preParseImport) {
...
-    if (elementStartsWith(inner, TABLE)) {
-      oldStyle = false;
-      Name tableName = getTableName(*inner[1]);
-      table = wasm.getTable(tableName);
-    } else {
-      if (elementStartsWith(inner, "offset")) {
-        offset = parseExpression(inner[1]);
+    // Old style table index (elem 0 (i32.const 0) ...)
+    if (s[i]->isStr() && s[i]->str() == "0") {
+      table = wasm.tables.front().get();
```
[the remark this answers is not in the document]
martianboy: Ohh, nice catch! I guess I wanted to defer throwing the exception until after the while loop, but I missed the `.get()` part. How about we drop support for that weird syntax altogether? It's doesn't seem to be valid as per reference-types spec?

src/wasm/wasm-binary.cpp
martianboy: Correct. I just wanted to this part to follow the complete spec, and then finish it in the next PR when I'll be changing element segments to have expressions instead of just function names. I can remove this for now if you want.

src/wasm/wasm-binary.cpp
martianboy: Thanks!

src/wasm/wasm-s-parser.cpp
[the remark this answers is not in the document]
tlively: That seems fine to me 👍

src/wasm/wasm-binary.cpp
tlively: Would be good to either remove or add a TODO comment explaining the future intentions.

src/passes/Print.cpp
```diff
@@ -2665,36 +2665,62 @@ struct PrintSExpression : public UnifiedExpressionVisitor<PrintSExpression> {
...
+      *currModule, curr->name, [&](ElementSegment* segment) {
+        printElementSegment(segment);
+      });
+  }
+  void visitElementSegment(ElementSegment* curr) {
+    if (curr->table.is()) {
+      return;
+    }
+
+    printElementSegment(curr);
```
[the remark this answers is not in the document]
martianboy: Done

src/passes/Print.cpp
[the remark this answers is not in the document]
tlively: Ok, assuming that we don't want to make an artificial `ElementSegment` just for printing the declarative segments, it looks like `isDeclarative` should not be necessary in this function.

src/wasm/wasm-binary.cpp
martianboy: Thanks removed

src/wasm/wasm-binary.cpp
martianboy: Done

src/wasm/wasm-s-parser.cpp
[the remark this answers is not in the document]
martianboy: Actually changed the codes a bit more, you may wanna take another look. Got rid of the while loop and took a more linear approach.

src/passes/Print.cpp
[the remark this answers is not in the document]
martianboy: Right, it's now removed.

martianboy:
@tlively I think I've addressed all of your comments, thanks for the full review! :) You may still wanna take a second look to my changes in `wasm-s-parser.cpp`.

tlively (approved):
Looks good! Thanks again for your work on this :)

PR #3574, opened 2021-02-17
[EH] Change catch_all's opcode

aheejin:
We decided to change `catch_all`'s opcode from 0x05, which is the same
as `else`, to 0x19, to avoid some complicated handling in the tools.

See: https://github.com/WebAssembly/exception-handling/issues/147

lso this contains the original cpp file used to generate
dwarf_with_exceptions.wasm; instructions to generate the wasm from that
cpp file are in the comments.


kripken (approved):
I think the dwarf changes are mostly because there is an extra line early in the file - all the line numbers are pushed down by 1. If it's easy to do then removing that line and rebuilding the file could make the diff here almost go away. But if it's not trivial then I wouldn't worry about it.

Also may be worth checking in that file as a `.cpp` alongside the wasm.

aheejin:
I removed the extra line, but this time I didn't try to 'clean up' demangled function names, like, `_Z3foov` -> `foo` or something, to enhance the regenerability from the cpp file. (I only removed my personal IDs like 'aheejin' from the debug info) Does this look better now? I can't really tell...

Yeah I'll also upload the cpp file.

aheejin:
I don't understand the error for emscripten CI faling. Does any one have a clue?

kripken (approved):
The `$foo => $foo\28\29` change is from C => C++ I think? Not sure how it compiled as C before, maybe some extern "C", but looks better as it is now.

kripken:
The emscripten errors are on missing symbols for `___cxa_is_pointer_type` and `___cxa_can_catch`. My first guess is that may be related to emscripten changes in deps_info that sam just did (and reached tot builds which affected this). I'll try to reproduce locally.

aheejin:
> The `$foo => $foo\28\29` change is from C => C++ I think? Not sure how it compiled as C before, maybe some extern "C", but looks better as it is now.

As I said, I manually modified C file to "clean up" dmangled names 😂 That's the reason how it was compiled as just `foo`. But now I uploaded the original file, I left function names as they are to enhance reproducibility from the original cpp file to the final wasm file. (I still removed my personal info though)

aheejin:
Oh I meant manually modified "ll" files.... (fixed the comment)

kripken:
Oh, heh, I missed that part about the manual cleanup... :smile: anyhow, looks good as it is.

Testing locally, I can't reproduce the emscripten error, even with this PR applied. Maybe it was previous tot breakage? Anyhow, unrelated to this PR.

aheejin:
Will land the PR; the CI failure doesn't seem to be related.

PR #3575, opened 2021-02-17
[EH] Remove dependency of reference types from EH

aheejin:
The new spec does not have `exnref` so EH does not have dependency of
the reference types proposal anymore.

exception_handling_target_feature.wasm's contents are the same except
previously its target features section contained both reference-types
and exception-handling but now it only has exception-handling.


kripken (approved):

tlively (approved):

aheejin:
Will land the PR; the CI failure doesn't seem to be related.

PR #3578, opened 2021-02-18
Fix TypeBuilder canonicalization

tlively:
When types or heap types were used multiple times in a TypeBuilder instance, it
was possible for the canonicalization algorithm to canonicalize a parent type
before canonicalizing all of its component child types, leaking the temporary
types into globally interned types. This bug led to incorrect canonicalization
results and use-after free bugs.

The cause of the bug was that types were canonicalized in the reverse of the
order that they were visited in, but children were visited after the first
occurrence of their parents, not necessarily after the last occurrence of their
parents. One fix could have been to remove the logic that prevented types from
being visited multiple times so that children would always be visited after
their parents. That simple fix, however, would not scale gracefully to handle
recursive types because it would require some way to detect recursions without
accidentally reintroducing these bugs.

This PR implements a more robust solution: topologically sorting the traversed
types to ensure that children are canonicalized before their parents. This
solution will be trivial to adapt for recursive types because recursive types
are trivial to detect from the reachability graph used to perform the
topological sort.

This supersedes the incomplete fix in #3571.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -1138,7 +1138,7 @@ struct Canonicalizer {
 
   // Maps Type and HeapType IDs to the IDs of their ancestor Types and HeapTypes
   // in the type graph. Only considers compound Types and HeapTypes.
-  std::unordered_map<TypeID, std::set<TypeID>> ancestors;
+  std::unordered_map<TypeID, std::unordered_set<TypeID>> ancestors;
 
   // Maps each temporary Type and HeapType to the locations where they will have
   // to be replaced with canonical Types and HeapTypes.
@@ -1327,7 +1327,7 @@ std::vector<Canonicalizer::Item> Canonicalizer::getOrderedItems() {
 
   // Collect the items to be sorted, including all the result HeapTypes and any
   // type that participates in the ancestor relation.
-  std::set<TypeID> allIDs;
+  std::unordered_set<TypeID> allIDs;
   for (auto ht : results) {
     allIDs.insert(ht.getID());
   }
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -1308,25 +1308,29 @@ std::vector<Canonicalizer::Item> Canonicalizer::getOrderedItems() {
   std::vector<TypeID> sorted;
   std::unordered_set<TypeID> seen;
 
+  // Topologically sort so that all parents are pushed before their children.
+  // This sort will be reversed later to have children before their parents.
   std::function<void(TypeID)> visit = [&](TypeID i) {
     if (seen.count(i)) {
       return;
     }
-    // Visit all children before the current type.
-    auto childrenIt = ancestors.find(i);
-    if (childrenIt != ancestors.end()) {
-      for (auto child : childrenIt->second) {
-        visit(child);
+    // Push ancestors of the current type before pushing the current type.
+    auto it = ancestors.find(i);
+    if (it != ancestors.end()) {
+      for (auto ancestor : it->second) {
+        visit(ancestor);
       }
     }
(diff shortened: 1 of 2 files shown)


src/wasm/wasm-type.cpp
```diff
@@ -1277,27 +1280,86 @@ void Canonicalizer::scanType(Type* type) {
...
+      succs.insert(nextAncestors.begin(), nextAncestors.end());
+      if (succs.size() != oldSize) {
         changed = true;
       }
     }
   } while (changed);
 }
 
+std::vector<Canonicalizer::Item> Canonicalizer::getOrderedItems() {
+  // Topological sort the types and heaptypes so that all children are
```
kripken, line 1305:
```suggestion
  // Topologically sort the Types and HeapTypes so that all children are
```

src/wasm/wasm-type.cpp
```diff
@@ -1277,27 +1280,86 @@ void Canonicalizer::scanType(Type* type) {
...
+
+  std::function<void(TypeID)> visit = [&](TypeID i) {
+    if (seen.count(i)) {
+      return;
+    }
+    // Visit all children before the current type.
+    auto childrenIt = ancestors.find(i);
+    if (childrenIt != ancestors.end()) {
+      for (auto child : childrenIt->second) {
+        visit(child);
```
kripken, line 1319: I'm missing something. `ancestors` is defined as
```cpp
  // Maps Type and HeapType IDs to the IDs of their ancestor Types and HeapTypes
  // in the type graph. Only considers compound Types and HeapTypes.
```
From this description I expect `ancestors[id]` to be a list of the ancestors of `id`. But the loop here instead calls those items "children" - ?

tlively: Oh yeah, I got myself confused about the correct direction of the relation when I was writing this and I guess I forgot to update this code. I'll fix this. `ancestors` and your expectation are correct.

tlively: @kripken this should be fixed now.

src/wasm/wasm-type.cpp
```diff
@@ -1277,27 +1280,93 @@ void Canonicalizer::scanType(Type* type) {
...
+        visit(ancestor);
+      }
+    }
+    seen.insert(i);
+    sorted.push_back(i);
+  };
+
+  // Collect the items to be sorted, including all the result HeapTypes and any
+  // type that participates in the ancestor relation.
+  std::set<TypeID> allIDs;
```
kripken, line 1330:
```suggestion
  std::unordered_set<TypeID> allIDs;
```
I doubt it matters for speed, but it may be clearer to read, as it indicates there is no sorting order in this data structure?

kripken (approved):

src/wasm/wasm-type.cpp
tlively: I intentionally used a `set` here so that the result of the topological would be deterministic. I'll add a comment.

kripken: Ah, then I missed something.

Do you mean it keeps the sort at each topological depth deterministic?

tlively: I meant for this `std::set` to keep the top-level driving of the sort deterministic and for the `std::set` in `ancestors` to keep each recursive step of the sort deterministic. But now that I think about it, the TypeIDs themselves aren't deterministic, so using a `std::set` here is not sufficient to make the sort deterministic. Furthermore, it shouldn't matter whether this sort is deterministic; that shouldn't be visible to the user or affect anything else. I'll switch to using `std::unordered_set` here and in `ancestors`.

kripken: Thanks! Sounds good.

PR #3580, opened 2021-02-18
Create a table if there is none in GenerateDynCalls::generateDynCallThunk (?)

kripken:
This was intended to fix [current LLVM=>emscripten autoroller breakage](https://logs.chromium.org/logs/emscripten-releases/buildbucket/cr-buildbucket.appspot.com/8854969628629209744/+/steps/Emscripten_testsuite__upstream__wasms_/0/stdout) due to https://github.com/llvm/llvm-project/commit/f48923e884611e6271a8da821a58aedd24d91cf7
```
commit f48923e884611e6271a8da821a58aedd24d91cf7 (HEAD)
Author: Andy Wingo <wingo@igalia.com>
Date:   Wed Feb 17 17:20:28 2021 +0100

    [WebAssembly][lld] --importTable flag only imports table if needed
    
    Before, --importTable forced the creation of an indirect function table,
    whether it was needed or not.  Now it only imports a table if needed.
    
    Differential Revision: https://reviews.llvm.org/D96872
```
However, this does **not** fix the issue. The assert does not occur, but instead the test fails at runtime with
```
RuntimeError: invalid index into function table
```
I have not followed the LLVM work here so I'm not sure if there is something wrong with that LLVM change or if we need more work on the binaryen side?

cc @wingo @tlively @sbc100

### DIFF
--- a/src/passes/GenerateDynCalls.cpp
+++ b/src/passes/GenerateDynCalls.cpp
@@ -127,9 +127,9 @@ void GenerateDynCalls::generateDynCallThunk(Signature sig) {
   for (const auto& param : sig.params) {
     args.push_back(builder.makeLocalGet(++i, param));
   }
-  // FIXME: Should the existence of a table be ensured here? i.e. create one if
-  // there is none?
-  assert(wasm->tables.size() > 0);
+  if (wasm->tables.empty()) {
+    wasm->addTable(Builder::makeTable(Name::fromInt(0)));
+  }
   f->body = builder.makeCallIndirect(wasm->tables[0]->name, fptr, args, sig);
 
   wasm->addFunction(std::move(f));
--- a/src/passes/GenerateDynCalls.cpp
+++ b/src/passes/GenerateDynCalls.cpp
@@ -128,6 +128,8 @@ void GenerateDynCalls::generateDynCallThunk(Signature sig) {
     args.push_back(builder.makeLocalGet(++i, param));
   }
   if (wasm->tables.empty()) {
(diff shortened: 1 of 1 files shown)


sbc100:
Thats odd, I thought I had run the entire emscripten test suite with that llvm change applied..I guess I did something wrong.

src/passes/GenerateDynCalls.cpp
```diff
@@ -127,9 +127,9 @@ void GenerateDynCalls::generateDynCallThunk(Signature sig) {
   for (const auto& param : sig.params) {
     args.push_back(builder.makeLocalGet(++i, param));
   }
-  // FIXME: Should the existence of a table be ensured here? i.e. create one if
-  // there is none?
-  assert(wasm->tables.size() > 0);
+  if (wasm->tables.empty()) {
+    wasm->addTable(Builder::makeTable(Name::fromInt(0)));
```
sbc100, line 131: Most likely you we want to create a table import here.. to mimic the older linker behaviour (which was to create a an unused import).

kripken: Ah, thanks, that makes sense. When I add an import with the exact same name as the backend would, it works.

kripken:
TBRing to fix breakage, and since I assume this is correct as I've implemented the review suggestion.

tlively (commented):
LGTM FWIW 👍

wingo:
lgtm fwiw as well; apologies for the clownshoes!

PR #3582, opened 2021-02-18
[Wasm Exceptions] Scan catch events in RemoveUnusedModuleElements

kripken:
There was an existing test for this, it turns out.

Also refactor away some annoying repeated code in that pass. `visitTry` is
the only actual change.

Noticed by the exceptions fuzzer https://github.com/emscripten-core/emscripten/pull/13485, e.g. on `wasm-opt -O1` on
```cpp
#include <stdio.h> // avoid iostream C++ code, just test libc++abi, not libc++
#include <stdint.h>

extern void refuel();
extern void checkRecursion();
extern bool getBoolean();

struct Class {
  Class();
  ~Class();
};

void func_0() {
  while (getBoolean()) {
  }
}

int main() {
  // func_0
  puts("calling func_0");
  refuel();
  try {
    func_0();
  } catch(...) {
    puts("main caught from func_0");
  }
  return 0;
}```

### DIFF
--- a/src/passes/RemoveUnusedModuleElements.cpp
+++ b/src/passes/RemoveUnusedModuleElements.cpp
@@ -85,32 +85,26 @@ struct ReachabilityAnalyzer : public PostWalker<ReachabilityAnalyzer> {
     }
   }
 
-  void visitCall(Call* curr) {
-    if (reachable.count(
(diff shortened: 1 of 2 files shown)


tlively (approved):

aheejin (approved):
Thanks! I forgot to do this...

PR #3583, opened 2021-02-18
[Wasm Exceptions] Fix RemoveUnusedNames on Try

kripken:
The `delegate` field of Try was not being scanned, so we could
remove a name that was used only by a delegate.

The bug was because `visitTry` overrides the generic visitor
`visitExpression`. So we need to call it manually. Sadly the code here
was pretty old (I probably wrote it back in 2015 or so) and it was
misleading, as it had unnecessary calls from the generic visitor
to `visitBlock, visitLoop`, which are not needed. This PR removes
them which is shorter and cleaner.

Also, we must handle the case of the delegate field being unset,
so check `name.is()`.

### DIFF
--- a/src/passes/RemoveUnusedNames.cpp
+++ b/src/passes/RemoveUnusedNames.cpp
@@ -37,16 +37,11 @@ struct RemoveUnusedNames
   std::map<Name, std::set<Expression*>> branchesSeen;
 
   void visitExpression(Expression* curr) {
-    if (auto* block = curr->dynCast<Block>()) {
-      visitBlock(block);
-      return;
-    }
-    if (auto* loop = curr->dynCast<Loop>()) {
-      visitLoop(loop);
-      return;
-    }
-    BranchUtils::operateOnScopeNameUses(
-      curr, [&](Name& name) { branchesSeen[name].insert(curr); });
+    BranchUtils::operateOnScopeNameUses(curr, [&](Name& name) {
+      if (name.is()) {
+        branchesSeen[name].insert(curr);
+      }
+    });
   }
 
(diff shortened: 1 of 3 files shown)


src/passes/RemoveUnusedNames.cpp
```diff
@@ -37,16 +37,11 @@ struct RemoveUnusedNames
...
 
   void visitExpression(Expression* curr) {
-    if (auto* block = curr->dynCast<Block>()) {
-      visitBlock(block);
-      return;
-    }
-    if (auto* loop = curr->dynCast<Loop>()) {
-      visitLoop(loop);
-      return;
-    }
```
aheejin, line 47: If this function does not call `visitBlock` and `visitLoop`, and this is a `UnifiedExpressionVisitor`, who calls them? Are they ever called?

aheejin (commented):
Thanks! FYI, the new impl hasn't been tested against optimization passes (especially `delegate`), so I guess there will be plenty of missing handling within the existing passes for that...

src/passes/RemoveUnusedNames.cpp
kripken: The mechanism of `UnifiedExpressionVisitor` is to define `visitFoo(Foo* curr) { visitExpression(curr); }` for each class. So they all redirect to the main `visitExpression` method. But it is possible to implement some and override that.

aheejin (approved):

kripken:
Btw, maybe I won't try to fuzz all the passes atm. But I was looking for non-optimizer fuzz bugs (bugs in basic binary reading and writing) and was using the optimizer to reduce them, which is why I ran into this...

PR #3584, opened 2021-02-18
Support type uses before definitions in text parser

tlively:
Traverses the module to find type definitions and uses a TypeBuilder to
construct the corresponding HeapTypes rather than constructing them directly.
This allows types to be used in the definitions of other types before they
themselves are defined, which is an important step toward supporting recursive
types. After this PR, no further text parsing changes will be necessary to
support recursive types.

Beyond allowing types to be used before their definitions, this PR also makes a
couple incidental changes to the parser's behavior. First, compound heaptypes
can now only be declared in `(type ...)` elements and cannot be declared inline
at their site of use. This reduces the flexibility of the parser, but is in line
with what the text format spec will probably look like eventually (see
WebAssembly/function-references#42).

The second change is that `(type ...)` elements are now all parsed before `(func
...)` elements rather than in text order with them, so the type indices will be
different and wasts using numeric type indices will be broken. Note however,
that we were already not completely spec compliant in this regard because we
parsed types defined by `(type...)` and `(func...)` elements before types
defined by the type uses of `call_indirect` instructions.

### DIFF
--- a/src/passes/Print.cpp
+++ b/src/passes/Print.cpp
@@ -175,7 +175,7 @@ static std::ostream& operator<<(std::ostream& o, const SExprType& sType) {
     o << '(';
     auto sep = "";
     for (const auto& t : type) {
-      o << sep << t;
+      o << sep << SExprType(t);
       sep = " ";
     }
     o << ')';
--- a/src/wasm-s-parser.h
+++ b/src/wasm-s-parser.h
@@ -70,6 +70,8 @@ public:
   List& list();
   Element* operator[](unsigned i);
   size_t size() { return list().size(); }
+  List::Iterator begin() { return list().begin(); }
+  List::Iterator end() { return list().end(); }
 
   // string methods
   cashew::IString str() const;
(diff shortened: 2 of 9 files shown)


tlively:
This currently includes an outdated version of #3578. I will rebase this once that lands, but I didn't want to delay posting this PR.

kripken:
Sorry, I'm not sure what the intended review plan is here. Do you want me to start reading this before the parent PR is ready for review? (still waiting there on some stuff I believe)

If so could this be relative to that branch, so that the full diff is not shown here?

tlively:
Unfortunately the branch for the other PR is my Binaryen fork, so there's no way to make a PR against it in this repo. The intention in posting it early was to give you the option to read it before the other one lands, but there's no rush. (Also the other one is ready for review again.)

src/passes/Print.cpp
```diff
@@ -175,7 +175,7 @@ static std::ostream& operator<<(std::ostream& o, const SExprType& sType) {
     o << '(';
     auto sep = "";
     for (const auto& t : type) {
-      o << sep << t;
+      o << sep << SExprType(t);
```
kripken, line 178: what does this fix?

kripken (approved):

src/passes/Print.cpp
tlively: Compound reference types in tuple locals were being fully printed rather than referring to a heap type by name. After this PR, inline heap types like that can no longer be parsed, so this change is necessary to use the type names instead.

PR #3585, opened 2021-02-18
[Wasm Exceptions] Handle delegation to the caller in RemoveUnusedNames

kripken:
Without this, the assertion just below fires, as we should erase each
branch target after we reach it, end end up with nothing left.

After this fix it looks like fuzzing can't find anything new after ~1000
iterations (100x more than before).


aheejin (approved):
Thanks!

PR #3588, opened 2021-02-19
Support type use before definition in binaries

tlively:
Update parsing of binary type sections to use TypeBuilder to support uses before
definitions. Now that both the binary and text parsers support out-of-order type
uses, this PR also relaxes the logic for emitting types to allow uses to be
emitted before definitions.


src/wasm-type.h
```diff
@@ -308,6 +308,9 @@ class HeapType {
     data,
   };
   static constexpr BasicHeapType _last_basic_type = data;
+  static constexpr uintptr_t _invalid_type = -1;
```
kripken, line 311: why not 0? (is `-1` guaranteed to segfault on use like 0 is?)

kripken (commented):
Nice!

lgtm aside from the one question.

src/wasm-type.h
kripken: Actually, is this needed? It would be better to avoid allowing an invalid HeapType at all. It seems like `getBasicHeapType` returns a boolean already, so we know if a value was found or not. Given that, the HeapType could be initialized to a valid value before the call? (There is a risk to doing that as well, but with a good comment there I think it's less risky than a general change to allow invalid HeapTypes to be declared as easily as `HeapType bad;`).

kripken:
Side note, @jakobkummerow and I happened to look at a wasm file with recursive types now (testcase for the names section), and I noticed that it crashes with this PR,
```
==580908== Stack overflow in thread #1: can't grow stack to 0x1ffe801000
==580908== 
==580908== Process terminating with default action of signal 11 (SIGSEGV)
==580908==  Access not within mapped region at address 0x1FFE801FF0
==580908== Stack overflow in thread #1: can't grow stack to 0x1ffe801000
==580908==    at 0x507BBEC: std::_Function_handler<void (unsigned long), wasm::(anonymous namespace)::Canonicalizer::getOrderedItems()::{lambda(unsigned long)#1}>::_M_invoke(std::_Any_data const&, unsigned long&&) (in /home/azakai/Dev/2-binaryen/lib/libbinaryen.so)
```
(Infinite recursion?)

That wasm has something like this:
```wat
// type 0: struct $StrA (field ($byte i8) ($word i16) ($pointer (ref $StrB)))
// type 1: struct $StrB (field ($next (ref null $StrA)))
```
[name-section.wasm.gz](https://github.com/WebAssembly/binaryen/files/6012602/name-section.wasm.gz)

tlively:
Yeah there are going to be lots of infinite recursions for infinite types right now, but I'm surprised that this didn't trigger an assertion in the TypeBuilder first.

src/wasm-type.h
tlively: Sure, I can have the default constructor just choose an arbitrary BasicHeapType. FWIW, address -1 should segfault on all reasonable platforms (as long as you're not in kernel mode or in WebAssembly or something 😉).

kripken (approved):

PR #3589, opened 2021-02-19
[Wasm GC] Add Names section support for field names

kripken:
Implementation of https://github.com/WebAssembly/gc/issues/193

Also fixes text format printing to use the names.

No extra tests needed, see the `.fromBinary` output which shows this working.

See the comment on the `const_cast` - maybe there's a better way
to do it? @tlively I remember we discussed the issue of returning the fields as
const or not a while ago.

cc @jakobkummerow


src/wasm/wasm-binary.cpp
```diff
@@ -2780,6 +2826,30 @@ void WasmBinaryBuilder::readNames(size_t payloadLen) {
...
+          for (size_t i = 0; i < numFields; i++) {
+            auto index = getU32LEB();
+            if (index >= fields.size()) {
+              std::cerr
+                << "warning: invalid field index in name field section\n";
+            } else {
+              // getStruct() intentionally returns a constant result. However,
+              // here we just adjust the name (which is not part of the hashing
+              // etc.) which is ok to do.
+              auto* field = const_cast<Field*>(&fields[index]);
```
tlively, line 2847: A few issues come to mind: First, if someone uses Binaryen as a library and parses multiple modules in parallel, there will be a data race here. Second, any use of Binaryen that parses multiple modules (not necessarily in parallel) will only end up with the names from one of the modules. Those are perhaps esoteric problems, but what if we kept the field names in a side table owned by the `Module` instead of making them part of the globally interned types? That would solve both of these problems and would also scale to preserving the names of HeapTypes as well, which I assume we will want to do as we parse and print wasts with complex types.

kripken: Yes, I think you're right. We keep local names in a map on the side in Functions, so this would be consistent with that. I'll look into a refactoring for it.

kripken:
This can now be done after all the refactorings and preparations landed, updated, PTAL.

New description:

```
Adds support for GC struct fields in the binary format, implementing
WebAssembly/gc#193

No extra tests needed, see the .fromBinary output which shows this working.

This also has a minor fix in the s-parser, we should not always add a name
to the map of index=>name - only if it exists. Without that fix, the binary
emitter would write out null strings.
```

src/wasm/wasm-binary.cpp
```diff
@@ -795,6 +795,33 @@ void WasmBinaryWriter::writeNames() {
   // TODO: label, type, and element names
   // see: https://github.com/WebAssembly/extended-name-section
 
+  // GC field names
+  if (wasm->features.hasGC()) {
+    std::vector<HeapType> namedTypes;
+    for (auto& type : types) {
+      if (type.isStruct() && wasm->typeNames.count(type) &&
+          !wasm->typeNames.at(type).fieldNames.empty()) {
```
tlively, line 803:
```suggestion
          !wasm->typeNames[type].fieldNames.empty()) {
```
I don't think there's any reason to use `.at()` instead of `[]` here, since we know this will never fail. Same below.

src/wasm/wasm-binary.cpp
```diff
@@ -795,6 +795,33 @@ void WasmBinaryWriter::writeNames() {
...
+        namedTypes.push_back(type);
+      }
+    }
+    if (!namedTypes.empty()) {
+      auto substart =
+        startSubsection(BinaryConsts::UserSections::Subsection::NameField);
+      o << U32LEB(namedTypes.size());
+      for (Index i = 0; i < namedTypes.size(); i++) {
+        auto type = namedTypes[i];
+        o << U32LEB(i);
```
tlively, line 813: IIUC, this emits the index into just the types with named fields, but it should emit the index into all the defined types.

src/wasm/wasm-binary.cpp
```diff
@@ -795,6 +795,33 @@ void WasmBinaryWriter::writeNames() {
   // TODO: label, type, and element names
   // see: https://github.com/WebAssembly/extended-name-section
 
+  // GC field names
+  if (wasm->features.hasGC()) {
+    std::vector<HeapType> namedTypes;
```
tlively, line 800: Maybe `typesWithNamedFields`? That's kind of verbose, though :/

src/wasm/wasm-binary.cpp
```diff
@@ -795,6 +795,33 @@ void WasmBinaryWriter::writeNames() {
...
+      }
+    }
+    if (!namedTypes.empty()) {
+      auto substart =
+        startSubsection(BinaryConsts::UserSections::Subsection::NameField);
+      o << U32LEB(namedTypes.size());
+      for (Index i = 0; i < namedTypes.size(); i++) {
+        auto type = namedTypes[i];
+        o << U32LEB(i);
+        auto& fieldNames = wasm->typeNames.at(type).fieldNames;
```
tlively, line 814: I would find this more readable with the type spelled out here instead of `auto`. That way it would be easier to figure out what is going on in the loop below. Just from the `fieldNames` variable name, I wouldn't have expected it to be a collection of pairs.

src/wasm/wasm-binary.cpp
kripken: I think `.at` is useful here because it cannot modify the underlying data structure. That is, `X[Y]` will potentially create an element.

While not strictly needed here, I think it improves readability. It would also allow the code to compile if the data structure were made `const` at some point.

src/wasm/wasm-binary.cpp
kripken: Good catch!

src/wasm/wasm-binary.cpp
kripken: Good point. How about `relevantTypes`?

src/wasm/wasm-binary.cpp
kripken: Fair point, added.

src/wasm/wasm-binary.cpp
kripken: Updated. This fixed something in the test suite, that I missed while scanning the output... I verified it carefully now.

src/wasm/wasm-binary.cpp
tlively: We know that `[]` will not perform any modifications in this case because we've already verified that `type` is present in the collection with `count`. Also, `[]` still works on const collections; it just returns a const reference instead of a mutable reference.

Anyway, I don't feel strongly about this and I agree that `.at` is generally better than `[]`. I was just surprised because generally we use `[]` over `.at` in similar situations. It would be nice if we could be more consistent in our usage. Maybe clang-tidy can help?

tlively (approved):

src/wasm/wasm-binary.cpp
kripken: I agree we should be more consistent. I can't remember offhand where this pattern happens, and it's not easy to grep for, though, so yeah, if tidy can help that could be great.

> Also, [] still works on const collections;

The problem is that if the map itself is const, we error on `[]` as it may insert a new element. Concrete example:
```cpp
#include <map>

int main() {
  const std::map<int, int> set;
  set[10];
}
```
```
$ g++ a.cpp
a.cpp: In function ‘int main()’:
a.cpp:5:9: error: passing ‘const std::map<int, int>’ as ‘this’ argument discards qualifiers [-fpermissive]
   set[10];
```

tlively: Oh yeah, I was wrong about `[]` working when `const`. That's very sad.

PR #3590, opened 2021-02-19
wasm-reduce: Do not verify binary roundtripping when not using binary files

kripken:
A reduction on text files may be done on text because binary writing is
where the bug is.

Diff is 99% whitespace.

### DIFF
--- a/src/tools/wasm-reduce.cpp
+++ b/src/tools/wasm-reduce.cpp
@@ -1249,7 +1249,7 @@ int main(int argc, const char* argv[]) {
 
   if (binary) {
     std::cerr << "|checking that command has expected behavior on "
-                " canonicalized (read-written) binary\n";
+                 " canonicalized (read-written) binary\n";
     {
       // read and write it
       auto cmd = Path::getBinaryenBinaryTool("wasm-opt") + " " + input +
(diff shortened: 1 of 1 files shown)


src/tools/wasm-reduce.cpp
```diff
@@ -1247,24 +1247,26 @@ int main(int argc, const char* argv[]) {
...
+  if (binary) {
+    std::cerr << "|checking that command has expected behavior on "
+                 " canonicalized (read-written) binary\n";
+    {
+      // read and write it
+      auto cmd = Path::getBinaryenBinaryTool("wasm-opt") + " " + input +
+                 " -o " + test + " " + extraFlags;
+      if (!binary) {
+        cmd += " -S ";
+      }
```
aheejin, line 1259:
```suggestion
```

aheejin (approved):
Don't try roundtripping to binary sounds right to me when not using binary, but I'm not sure "binary writing is where the bug is"..?

kripken:
Sorry, by `"binary writing is where the bug is"` I meant that I had a bug in binary writing. I ran the reducer on text to get a testcase for it, so the testcase was "write the binary and check for a crash". That's when I noticed the issue here.

kripken:
Actually I changed my mind here. I think it is good to verify that reading and writing the input works even with not reducing on binary. But the comments were confusing, which is why I was mixed up before. I fixed those now.

PR #3591, opened 2021-02-19
Properly use text format type names in printing

kripken:
This adds a TypeNames entry to modules, which can store names for types. So
far this PR uses that to store type names from text format. Future PRs will add
support for field names and for the binary format.

(Field names are added to wasm.h here to see if we agree on this direction.)

Most of the work here is threading a module through the various functions in
`Print.cpp`. This keeps the module optional, so that we can still print an
expression independently of a module, which has always been the case, and
which I think we should keep (but, if a module was mandatory perhaps this
would be a little simpler, and could be refactored into a form that depends on
that).

99% of this diff are test updates, since almost all our tests use the text
format, and many of them specify a type name but we used to ignore it.

This is a step towards a proper solution for https://github.com/WebAssembly/binaryen/pull/3589


src/passes/Print.cpp
```diff
@@ -169,13 +182,16 @@ struct SExprType {
...
+                                    const SExprType& sType,
+                                    Module* wasm = nullptr) {
   Type type = sType.type;
   if (type.isTuple()) {
     o << '(';
     auto sep = "";
     for (const auto& t : type) {
-      o << sep << SExprType(t);
+      o << sep;
+      printSExprType(o, t, wasm);
```
tlively, line 194: I do kind of like having these custom printers as subexpressions rather than as their own statements, even when they're not implemented via wrapper types. Being uniform about this would be a bigger separate change, though.

```suggestion
      o << sep << printSExprType(o, t, wasm);
```

tlively (commented):
I like this direction a lot, and I think preserving the original names makes the tests much easier to read.

I wonder if making all the print functions methods of a new `Printer` class that holds their shared context would help cut down on all the threading through of the `wasm` context or if that would just cause a bigger mess.

src/passes/Print.cpp
kripken: Interesting. I think that could work, but we'd need to remove the `o` parameter and have it return the result, I think (to avoid nondeterminism in what is called first if there are multiple calls on a single line).

I also agree that maybe a single big Printer class might help here. I'm not sure though.

For both, I think I'd prefer to leave them to future investigation/refactoring.

Separately, I think this PR should be complete now, after I figured out a nondeterminism bug.

src/wasm/wasm-s-parser.cpp
```diff
@@ -845,7 +845,15 @@ void SExpressionWasmBuilder::preParseHeapTypes(Element& module) {
...
-    wasm.typeNames[types[pair.second]].name = pair.first;
+    auto name = pair.first;
+    auto type = types[pair.second];
+    // A type may appear in the type section more than once, but we canonicalize
+    // types internally, so there will be a single name chosen for that type. Do
+    // so determistically.
+    if (wasm.typeNames.count(type) && wasm.typeNames[type].name.str < name) {
+      continue;
+    }
+    wasm.typeNames[type].name = name;
```
tlively, line 856: Why take the "smallest" name for a type rather than e.g. the first or last name that appears in the type section?

test/lit/forward-declared-types.wast
```diff
@@ -2,10 +2,10 @@
 
 ;; RUN: wasm-opt %s -all -S -o - | filecheck %s
 
-;; CHECK: (type $none_=>_none (func))
-;; CHECK: (type $none_=>_ref?|{ref?|[rtt_2_$none_=>_none]|_ref?|none_->_none|}| (func (result (ref null ${ref?|[rtt_2_$none_=>_none]|_ref?|none_->_none|}))))
-;; CHECK: (type ${ref?|[rtt_2_$none_=>_none]|_ref?|none_->_none|} (struct (field (ref null $[rtt_2_$none_=>_none])) (field (ref null $none_=>_none))))
-;; CHECK: (type $[rtt_2_$none_=>_none] (array (rtt 2 $none_=>_none)))
+;; CHECK: (type $func (func))
+;; CHECK: (type $none_=>_ref?|{ref?|[rtt_2_$func]|_ref?|none_->_none|}| (func (result (ref null $struct))))
```
tlively, line 6: Would it be possible to change the type name generation so that this becomes `(type $none_=>_ref?|$struct| ...)`? That would be much more readable.

src/wasm/wasm-s-parser.cpp
kripken: No super-strong reasons, but: The smallest leads to the smallest output, which can be more readable. It also has the benefit that reordering the wast does not alter the result, which is intuitive since in the wast they are named and the order shouldn't matter in principle.

test/lit/forward-declared-types.wast
kripken: Great point! Definitely much better that way. Fixed.

tlively (approved):
🎉

PR #3592, opened 2021-02-22
Fix global destructor ordering problem

aardappel:
On Windows/VS the maps in this code caused a double-delete of a Literal.
Given that order of destruction is unspecified, I decided to make them locals, which fixed it.
Not sure if there is still a latent ordering bug, but not having these as globals seems an improvement regardless.

### DIFF
--- a/src/tools/wasm-shell.cpp
+++ b/src/tools/wasm-shell.cpp
@@ -42,13 +42,6 @@ Name ASSERT_UNLINKABLE("assert_unlinkable");
 Name INVOKE("invoke");
 Name GET("get");
 
-// Modules named in the file
-
-std::map<Name, std::unique_ptr<Module>> modules;
-std::map<Name, std::unique_ptr<SExpressionWasmBuilder>> builders;
-std::map<Name, std::unique_ptr<ShellExternalInterface>> interfaces;
-std::map<Name, std::unique_ptr<ModuleInstance>> instances;
-
 //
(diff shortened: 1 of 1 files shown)


kripken (approved):
Nice!

tlively:
Does ASan tell you where the two deletes occurred? I don't see how destructor ordering could cause a `Literal` to be deleted twice, so I suspect a different underlying bug here.

aardappel:
@tlively this happens in visual studio, haven't tried to use Asan with that.
Looks like they just added it :) https://devblogs.microsoft.com/cppblog/addresssanitizer-asan-for-windows-with-msvc/

VS has double free debugging tools, but that only shows where something is allocated, not where it was freed the first time.

kripken:
@tlively The error @aardappel mentioned on chat was something like `"~Literal on unhandled type: ". Maybe the global type storage was deleted before something that used types?

If so, maybe that could be adjusted to happen last, or not at all?

tlively:
Oh yeah, that sounds very likely. @aardappel is there a good way to lower the priority of a global destructor that MSVC would respect? If not, we could probably figure out how to entirely avoid cleaning up the globally interned types and how to inform ASan that those leaks are not errors.

aardappel:
I don't think you can specify the order (across translation units), no. Presumably there are tricks to force an ordering explicitly, but I think the best solution is to simply not have them at all.

PR #3594, opened 2021-02-23
[Wasm Exceptions] Fix cfg-traversal on linking the basic block after a call

kripken:
This was an unfortunate case of the order of execution of call
arguments. `link(self->currBasicBlock, self->startBasicBlock())` would
run the call first, which sets `currBasicBlock`, so we'd end up with
the same value for both parameters.

Without this fix, the testcase would `drop` the result of the call,
as it thought it had no uses.

Also improve debug logging here a tiny bit.

Found by https://github.com/emscripten-core/emscripten/pull/13485


aheejin (approved):
Thank you!

Issue #3573, opened 2021-02-17
Globals, table & memory metadata not exposed to JS


brion:
I'm working on an experimental async wasm interpreter, piggybacking off binaryen.js to parse the binary and let me walk the AST. For the most part this is working well so far, but there's a few things that are missing from the JS API:

* Globals can be retrieved only via a reference from an export, or via a get/set expression. There's no way to walk the list of all globals, which I would need to initialize values without walking the entire codebase first.
* Memory segment info can be read, but not info about the memory itself such as its initial and maximum sizes, whether it's shared, or whether it's imported.
* Table imports are visible, but the initial and maximum sizes are also not exposed.

I'll try to whip up a provisional patch adding these (until then I can work around it by making certain assumptions that won't hold with all modules).

kripken:
A patch would be great!

Note that in the long term I hope to automate our JS API bindings. All the necessary info should be in declarative form here:

https://github.com/WebAssembly/binaryen/blob/main/src/wasm-delegations-fields.h

A patch to add specific things would be fine, but if you add more than one or two things it might actually be faster to write the autogenerator.

brion:
I'll take a peek, thanks!

brion:
@kripken
> Note that in the long term I hope to automate our JS API bindings. All the necessary info should be in declarative form here:
> 
> https://github.com/WebAssembly/binaryen/blob/main/src/wasm-delegations-fields.h
> 
> A patch to add specific things would be fine, but if you add more than one or two things it might actually be faster to write the autogenerator.

It looks like those delegation macros would allow replacing the implementation of `Module.getExpressionInfo` without having to manually add every new expression id. However I think this is separate from other metadata like looking at the attached globals, tables, and memories. I think the implementation of one could proceed separately from the implementation of the other and they would not interfere with each other.

kripken:
@brion Oh, yes, that's right. The proposed autogeneration stuff would be for expressions, but module level things are separate, at least for now.

brion:
Cool, well I might or might not get to that too but that'll be for another day. :D

brion:
@kripken while I'm thinking on it though one more thing -- the TypeScript annotations in the binaryen.js node package are nice as they help with code completion in Visual Studio Code; it'd be nice to figure a way to start generating them as well. These need to be static information, not dynamic code, so the delegate callbacks won't work unless code is run on a list of all possible expressions, which perhaps could be rigged up but feels like an extra step.

[...]

PR #3595, opened 2021-02-23
Add C and JS API functions for accessing memory info (#3573)

brion:
This allows identifying whether a memory is present on the module,
getting its initial and, if present, maximum sizes, and whether it's
shared as well as its import module and base name if imported.

Will require a matching update to index.d.ts in binaryen.js for
type hinting when used via npm.

### DIFF
--- a/src/binaryen-c.cpp
+++ b/src/binaryen-c.cpp
@@ -3511,6 +3511,37 @@ uint32_t BinaryenGetMemorySegmentByteOffset(BinaryenModuleRef module,
   Fatal() << "non-constant offsets aren't supported yet";
   return 0;
 }
+int BinaryenHasMemory(BinaryenModuleRef module) {
+  return ((Module*)module)->memory.exists;
+}
+size_t BinaryenMemoryGetInitial(BinaryenModuleRef module) {
+  return ((Module*)module)->memory.initial;
+}
+int BinaryenMemoryHasMax(BinaryenModuleRef module) {
+  return ((Module*)module)->memory.hasMax();
+}
+size_t BinaryenMemoryGetMax(BinaryenModuleRef module) {
+  return ((Module*)module)->memory.max;
+}
+const char* BinaryenMemoryImportGetModule(BinaryenModuleRef module) {
+  auto& memory = ((Module*)module)->memory;
+  if (memory.imported()) {
+    return memory.module.c_str();
+  } else {
+    return "";
+  }
+}
+const char* BinaryenMemoryImportGetBase(BinaryenModuleRef module) {
+  auto& memory = ((Module*)module)->memory;
+  if (memory.imported()) {
+    return memory.base.c_str();
+  } else {
+    return "";
+  }
+}
+int BinaryenMemoryGetShared(BinaryenModuleRef module) {
+  return ((Module*)module)->memory.shared;
+}
 size_t BinaryenGetMemorySegmentByteLength(BinaryenModuleRef module,
                                           BinaryenIndex id) {
   const auto& segments = ((Module*)module)->memory.segments;
--- a/src/binaryen-c.cpp
+++ b/src/binaryen-c.cpp
@@ -3514,13 +3514,13 @@ uint32_t BinaryenGetMemorySegmentByteOffset(BinaryenModuleRef module,
 int BinaryenHasMemory(BinaryenModuleRef module) {
   return ((Module*)module)->memory.exists;
 }
-size_t BinaryenMemoryGetInitial(BinaryenModuleRef module) {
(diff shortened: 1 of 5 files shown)


kripken (approved):
Code looks good, thanks!

For a test, please add to `test/binaryen.js/` (maybe in kitchen-sink or expressions?)

@dcodeIO , thoughts?

src/binaryen-c.cpp
```diff
@@ -3511,6 +3511,37 @@ uint32_t BinaryenGetMemorySegmentByteOffset(BinaryenModuleRef module,
...
+}
+const char* BinaryenMemoryImportGetBase(BinaryenModuleRef module) {
+  auto& memory = ((Module*)module)->memory;
+  if (memory.imported()) {
+    return memory.base.c_str();
+  } else {
+    return "";
+  }
+}
+int BinaryenMemoryGetShared(BinaryenModuleRef module) {
```
dcodeIO, line 3542: Here I'm not sure. Type upon construction is `uint8_t`, and looks like a bool, so perhaps
```suggestion
int BinaryenMemoryIsShared(BinaryenModuleRef module) {
```
? Hmm

src/binaryen-c.h
```diff
@@ -2097,6 +2097,15 @@ BINARYEN_API void BinaryenSetMemory(BinaryenModuleRef module,
                                     BinaryenIndex numSegments,
                                     uint8_t shared);
 
+BINARYEN_API int BinaryenHasMemory(BinaryenModuleRef module);
+BINARYEN_API size_t BinaryenMemoryGetInitial(BinaryenModuleRef module);
```
dcodeIO, line 2101:
```suggestion
BINARYEN_API BinaryenIndex BinaryenMemoryGetInitial(BinaryenModuleRef module);
```

src/binaryen-c.h
```diff
@@ -2097,6 +2097,15 @@ BINARYEN_API void BinaryenSetMemory(BinaryenModuleRef module,
...
+BINARYEN_API int BinaryenMemoryHasMax(BinaryenModuleRef module);
+BINARYEN_API size_t BinaryenMemoryGetMax(BinaryenModuleRef module);
```
dcodeIO, line 2103:
```suggestion
BINARYEN_API BinaryenIndex BinaryenMemoryGetMax(BinaryenModuleRef module);
```

src/binaryen-c.h
brion: I seem to remember initially trying that to fit the other functions but I got a compiler error saying it didn't match the size_t in the .cpp file (which also matched what I was trying to emulate, as I recall, but which compiled without error which confused me)

src/binaryen-c.cpp
brion: Should I use `uint8_t` for consistency with construction of this property or `int` for consistency with all the other booleans?

src/binaryen-c.cpp
```diff
@@ -3511,6 +3511,37 @@ uint32_t BinaryenGetMemorySegmentByteOffset(BinaryenModuleRef module,
   Fatal() << "non-constant offsets aren't supported yet";
   return 0;
 }
+int BinaryenHasMemory(BinaryenModuleRef module) {
+  return ((Module*)module)->memory.exists;
+}
+size_t BinaryenMemoryGetInitial(BinaryenModuleRef module) {
```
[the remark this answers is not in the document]
brion: Oh a further note -- this could have issues I think on 64-bit memories, as these values are Address instances and may be 64-bit on a 64-bit memory.

src/binaryen-c.h
brion: Looks like I did something wrong earlier. :D Changed them to BinaryenIndex and it's looking happy this time. :D

src/binaryen-c.cpp
brion: (I'll stick with int for now on the return, but going ahead and changing from `Get` to `Is`)

brion:
I'm making the suggested changes and adding a test case, will update shortly

src/binaryen-c.cpp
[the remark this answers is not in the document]
kripken: We may at some point need to make this 64-bit, but for now this can use `uint32_t` which is consistent with other places an Address is used (like a Load offset). Eventually for wasm64 we'll need to do a pass on all those.

kripken:
Looks like this got forgotten - if it was ready for a final review, apologies if I missed it.

#4682 has just been opened which rebases it, so hopefully the code can land shortly.

kripken:
Landed in #4682

PR #3599, opened 2021-02-23
[Wasm Exceptions] Fix StackIR writing of nested delegates

kripken:
We were missing a pop of catchIndexStack at a Delegate. It ends the scope,
so it should do that, like TryEnd does.

Found by https://github.com/emscripten-core/emscripten/pull/13485 on `-O2`.


aheejin (approved):
Oops I missed that. Thank you!

PR #3600, opened 2021-02-23
[Wasm GC] Move struct field names to their proper place

kripken:
#3591 adds type and field names to the Module object, and used that
for the type but not the fields. This uses it for the fields as well, and removes
the "name" field from the Field objects itself, completing the refactoring.

After this, binary format support can be added as a proper replacement for
#3589

edit: a big part of this is just whitespace

### DIFF
--- a/src/wasm/wasm-s-parser.cpp
+++ b/src/wasm/wasm-s-parser.cpp
@@ -782,7 +782,7 @@ void SExpressionWasmBuilder::preParseHeapTypes(Element& module) {
                      builder.getTempTupleType(results));
   };
 
-  // Mapes type indexes to a mapping of field index => name. We store the data
+  // Maps type indexes to a mapping of field index => name. We store the data
   // here while parsing as types have not been created yet.
   std::unordered_map<size_t, std::unordered_map<Index, Name>> fieldNames;
 
--- a/src/wasm/wasm-s-parser.cpp
+++ b/src/wasm/wasm-s-parser.cpp
@@ -782,7 +782,12 @@ void SExpressionWasmBuilder::preParseHeapTypes(Element& module) {
                      builder.getTempTupleType(results));
   };
 
-  auto parseField = [&](Element* elem) {
+  // Mapes type indexes to a mapping of field index => name. We store the data
+  // here while parsing as types have not been created yet.
+  std::unordered_map<size_t, std::unordered_map<Index, Name>> fieldNames;
+
+  // Parses a field, and optionally notes the name with a type index.
+  auto parseField = [&](Element* elem, size_t typeIndex = -1, size_t fieldIndex = -1) {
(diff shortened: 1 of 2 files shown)


src/wasm/wasm-s-parser.cpp
```diff
@@ -846,14 +853,22 @@ void SExpressionWasmBuilder::preParseHeapTypes(Element& module) {
...
       continue;
     }
-    wasm.typeNames[type].name = name;
+    auto& currTypeNames = wasm.typeNames[type];
+    currTypeNames.name = name;
+    if (type.isStruct()) {
+      auto& currFieldNames = fieldNames[index];
+      for (auto& pair : currFieldNames) {
+        currTypeNames.fieldNames[pair.first] = pair.second;
+      }
```
tlively, line 870: Can this just be something like `currTypeNames.fieldNames = fieldNames[index]`?

kripken: Yes, I think you're right. We don't update on top of anything existing, and I doubt we'll need to, so we can just replace. Done.

src/wasm/wasm-s-parser.cpp
```diff
@@ -782,43 +782,49 @@ void SExpressionWasmBuilder::preParseHeapTypes(Element& module) {
...
-    if (elementStartsWith(elem, FIELD)) {
-      if (elem->size() == 3) {
-        name = (*elem)[1]->str();
+  // Mapes type indexes to a mapping of field index => name. We store the data
+  // here while parsing as types have not been created yet.
+  std::unordered_map<size_t, std::unordered_map<Index, Name>> fieldNames;
+
+  // Parses a field, and optionally notes the name with a type index.
+  auto parseField =
+    [&](Element* elem, size_t typeIndex = -1, size_t fieldIndex = -1) {
```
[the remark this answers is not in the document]
kripken: Ah, good idea, that is indeed nicer! Done.

src/wasm/wasm-s-parser.cpp
```diff
@@ -803,28 +807,29 @@ void SExpressionWasmBuilder::preParseHeapTypes(Element& module) {
...
+    for (Index i = 1; i < elem.size(); i++) {
+      fields.emplace_back(parseField(elem[i], fieldNames[typeIndex][i - 1]));
     }
     return Struct(fields);
   };
 
   auto parseArrayDef = [&](Element& elem) {
-    return Array(parseField(elem[1]));
+    Name dummy;
+    return Array(parseField(elem[1], dummy));
```
tlively, line 832:
```suggestion
    Name unused;
    return Array(parseField(elem[1], unused));
```

tlively: This would be more precise.

tlively (approved):

PR #3601, opened 2021-02-24
Fix hashing of a use of a name without the context/target for it

kripken:
Before this we would assert on hashing e.g. `(br $x)` by itself, without the
context so we recognized the name `$x`. Somehow that was not an issue
until delegate, we just happened to not hash such things. I believe I remember
that @aheejin noticed this issue before, but given we didn't have a testcase,
we deferred fixing it - now is the time, I guess, as with delegate it is easy to
get e.g. CodeFolding to hash a Try with a delegate.

Issue found by https://github.com/emscripten-core/emscripten/pull/13485

### DIFF
--- a/src/ir/ExpressionAnalyzer.cpp
+++ b/src/ir/ExpressionAnalyzer.cpp
@@ -345,7 +345,8 @@ size_t ExpressionAnalyzer::hash(Expression* curr) {
       // But if the name is not known to us, hash the absolute one.
       if (!internalNames.count(curr)) {
         rehash(digest, 1);
-        rehash(digest, curr);
+        // Perform the same hashing as a generic name.
+        visitNonScopeName(curr);
         return;
       }
       rehash(digest, 2);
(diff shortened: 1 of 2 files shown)


src/ir/ExpressionAnalyzer.cpp
```diff
@@ -331,15 +331,28 @@ size_t ExpressionAnalyzer::hash(Expression* curr) {
...
+      // But if the name is not known to us, hash the absolute one.
+      if (!internalNames.count(curr)) {
+        rehash(digest, 1);
+        // Perform the same hashing as a generic name.
+        visitNonScopeName(curr);
+        return;
       }
+      rehash(digest, 2);
+      static_assert(sizeof(Index) == sizeof(int32_t),
+                    "wasm64 will need changes here");
```
tlively, line 354: Why is this?

tlively (approved):

src/ir/ExpressionAnalyzer.cpp
kripken: The data structure here is `std::map<Name, Index> internalNames;` so I think the idea in the assertion is that wasm64 might need something bigger than an Index if it has over `2^32` names? I agree it should be clarified.

aheejin (approved):
Thank you!

PR #3602, opened 2021-02-24
Support building recursive types

tlively:
Updates TypeBuilder to support recursive types. Recursive types are particularly
problematic because under the current scheme it is necessary to canonicalize the
uses of a type's immediate children before canonicalizing the type itself to
avoid leaking non-canonical, temporary types out of the TypeBuilder and into the
global type stores. In the case of recursive types, it is not possible to do
this because of their cyclic nature. In principle this could be overcome by
hashing recursive types based on their structure rather than their contents, but
that would be complicated. Instead, this PR takes the shortcut of not
canonicalizing self-referential HeapTypes at all, but rather moving them out of the
TypeBuilder and into the global type store without changing their addresses or
needing to update any of their use sites. This breaks all cycles and makes it
possible to canonicalize the other types correctly.

Note that this PR _only_ adds support for building recursive types. Doing almost
anything with the types, such as printing, comparing, or emitting them will
certainly lead to infinite recursions. A follow up PR will update all these
operations to work correctly with recursive types.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -1396,7 +1396,10 @@ std::vector<Canonicalizer::Item> Canonicalizer::getOrderedItems() {
   // Collect the remaining types that will be sorted.
   std::unordered_set<TypeID> toSort;
   for (auto& entry : builder.impl->entries) {
-    toSort.insert(entry.get().getID());
+    TypeID id = entry.get().getID();
+    if (!selfReferentialHeapTypes.count(id)) {
+      toSort.insert(id);
+    }
   }
   for (auto& kv : childrenDAG) {
     toSort.insert(kv.first);
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -1380,6 +1380,10 @@ std::vector<Canonicalizer::Item> Canonicalizer::getOrderedItems() {
   // consistent, but this is good enough for getting prototype toolchains up and
   // running.
 
+  // TODO: None of this is particularly optimized. Benchmark to see if this is a
+  // significant bottleneck and investigate using better data structures and
+  // algorithms.
+
   // Remove self-referential HeapTypes to cut cycles.
   auto childrenDAG = children;
   for (TypeID id : selfReferentialHeapTypes) {
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -285,10 +285,6 @@ template<typename Info> struct Store {
     typeIDs[info] = id;
     return typename Info::type_t(id);
   }
-
-  void insertWithoutCanonicalizing(std::unique_ptr<Info>&& info) {
-    constructedTypes.emplace_back(std::move(info));
-  }
 };
 
 struct TypeStore : Store<TypeInfo> {
@@ -1218,6 +1214,10 @@ struct Canonicalizer {
   void canonicalize(T* type, std::unordered_map<T, T>& canonicals);
 };
 
+// Used to extend the lifetime of self-referential HeapTypes so they don't need
+// to be canonicalized.
+std::vector<std::unique_ptr<HeapTypeInfo>> noncanonicalHeapTypes;
+
 // Traverse the type graph rooted at the initialized HeapTypeInfos in reverse
 // postorder, replacing in place all Types and HeapTypes backed by the
 // TypeBuilder's Stores with equivalent globally canonicalized Types and
@@ -1271,7 +1271,7 @@ Canonicalizer::Canonicalizer(TypeBuilder& builder) : builder(builder) {
   // the TypeBuilder without their IDs changing.
   for (auto& entry : builder.impl->entries) {
     if (selfReferentialHeapTypes.count(entry.get().getID())) {
-      globalHeapTypeStore.insertWithoutCanonicalizing(std::move(entry.info));
+      noncanonicalHeapTypes.emplace_back(std::move(entry.info));
     }
   }
 }
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -1395,6 +1395,9 @@ std::vector<Canonicalizer::Item> Canonicalizer::getOrderedItems() {
 
   // Collect the remaining types that will be sorted.
   std::unordered_set<TypeID> toSort;
+  for (auto& entry : builder.impl->entries) {
+    toSort.insert(entry.get().getID());
+  }
   for (auto& kv : childrenDAG) {
     toSort.insert(kv.first);
     for (auto& child : kv.second) {
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -285,6 +285,10 @@ template<typename Info> struct Store {
     typeIDs[info] = id;
     return typename Info::type_t(id);
   }
+
+  void insertWithoutCanonicalizing(std::unique_ptr<Info>&& info) {
+    constructedTypes.emplace_back(std::move(info));
+  }
 };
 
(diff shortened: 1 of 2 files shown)


tlively:
This doesn't actually work because it doesn't canonicalize recusive Types before "non-recursive" HeapTypes that contain them, where "non-recursive" means the HeapType doesn't contain itself (but it's still an infinite tree because it contains a "recursive" type).

So I don't forget, I need to:

 - Do a topological sort on the child-parent graph (before calculating its fixed point) with the self-reaching HeapTypes removed.
 - Set `recursive = true` on all infinite types, not just the self-reaching ones.

src/wasm/wasm-type.cpp
```diff
@@ -1239,17 +1283,12 @@ Canonicalizer::Canonicalizer(TypeBuilder& builder) : builder(builder) {
...
-      WASM_UNREACHABLE("TODO: support recursive types");
-    }
-  }
+  findRecursiveTypes();
 
-  // Visit the types and heap types in reverse postorder, replacing them with
-  // their canonicalized versions.
+  // Visit the use sites of Types and HeapTypes, replacing them with their
+  // canonicalized versions in an order that guarantees no temporary types will
+  // be leaked into the global stores..
```
kripken, line 1291:
```suggestion
  // be leaked into the global stores.
```

src/wasm/wasm-type.cpp
```diff
@@ -1260,6 +1299,15 @@ Canonicalizer::Canonicalizer(TypeBuilder& builder) : builder(builder) {
         break;
     }
   }
+
+  // Now that all uses of Types and non-recursive HeapTypes have been
+  // canonicalized, move recursive HeapTypes to the global store so that they
+  // will outlive the TypeBuilder without their IDs changing.
+  for (auto& entry : builder.impl->entries) {
+    if (entry.info->recursive) {
+      globalHeapTypeStore.insertWithoutCanonicalizing(std::move(entry.info));
```
kripken, line 1308: Can we do the same for all types? I understand this is needed for recursive types, but if it's possible for non-recursive ones as well that could be simpler.

src/wasm/wasm-type.cpp
```diff
@@ -1340,10 +1388,29 @@ void Canonicalizer::makeAncestorsFixedPoint() {
   } while (changed);
 }
 
+void Canonicalizer::findRecursiveTypes() {
+  for (auto& kv : ancestors) {
+    TypeID curr = kv.first;
+    if (kv.second.count(kv.first) != 0) {
```
kripken, line 1394:
```suggestion
    if (kv.second.count(curr) != 0) {
```
Also capturing `kv.second` to a local would give it a name which could be more readable I think.

src/wasm/wasm-type.cpp
tlively: If we end up partially or wholly switching to nominal types, it would make sense to move ownership of the nominal types from the canonicalizing global stores to the individual `Module`s. But for structural types, deduplication is a matter of correctness. If the proposal keeps fully structural typing, we will have to implement canonicalization of recursive types at some point as well.

kripken: Oh, I see, I didn't get that before. So this does not canonicalize recursive types despite it being technically incorrect, but it may be good enough for now? (Why is it good enough for now?)

src/wasm/wasm-type.cpp
```diff
@@ -285,6 +289,10 @@ template<typename Info> struct Store {
     typeIDs[info] = id;
     return typename Info::type_t(id);
   }
+
+  void insertWithoutCanonicalizing(std::unique_ptr<Info>&& info) {
```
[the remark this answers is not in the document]
tlively: `canonicalize` does not take ownership of anything because it copies the type info object, so I think it's current name describes its intended interface. I got rid of this new method, though.

src/wasm/wasm-type.cpp
tlively: This should be good enough for now because producers will generally emit one type declaration per source type and not deduplicate or mix the two. Because producers don’t treat the types as interchangeable, it’s ok that Binaryen doesn’t either. Another piece is that for now neither Binaryen nor the producers are dealing with systems of multiple modules. If they were, it would be important for Binaryen to recognize when the same type is declared in multiple modules, which would require properly canonicalizing recursive types.

src/wasm/wasm-type.cpp
```diff
@@ -1319,85 +1333,102 @@ void Canonicalizer::scanType(Type* type) {
...
-      succs.insert(nextAncestors.begin(), nextAncestors.end());
+      succs.insert(newSuccs.begin(), newSuccs.end());
       if (succs.size() != oldSize) {
         changed = true;
       }
     }
   } while (changed);
+
+  // Find HeapTypes that reach themselves
+  for (auto& entry : builder.impl->entries) {
```
kripken, line 1358: Can this loop be on `fixedPoint`? Basically the set we care about is `{ x \in children(x) | x \in AllThings }` so I'm not sure why this starts with `builder.impl->entries`.

src/wasm/wasm-type.cpp
```diff
@@ -1319,85 +1333,102 @@ void Canonicalizer::scanType(Type* type) {
...
+  // running.
+
+  // Remove self-referential HeapTypes to cut cycles.
+  auto childrenDAG = children;
+  for (TypeID id : selfReferentialHeapTypes) {
+    childrenDAG.erase(id);
+    for (auto& kv : childrenDAG) {
+      kv.second.erase(id);
+    }
+  }
```
kripken, line 1390: I don't know if this matters much, but for performance I suspect the reversed loops might be faster. I am assuming that most types are not self-referential - does that sound reasonable?

If so, then cache locality would benefit from having the inner loop be on the smaller thing, `selfReferentialHeapTypes`. But then you'd need two loops to avoid iterator invalidation I guess,
```cpp
for (auto& kv : childrenDAG) {
  kv.second.erase(selfReferentialHeapTypes.begin(), selfReferentialHeapTypes.end());
}
childrenDAG.erase(selfReferentialHeapTypes.begin(), selfReferentialHeapTypes.end());
```

kripken (commented):
Nice!

lgtm after we figure out those comments.

src/wasm/wasm-type.cpp
tlively: If we looped over `fixedPoint`, we would have to first traverse it to collect a set of all TypeIDs it references and at some point we would also have to check whether each TypeID corresponded to a HeapType. In contrast, the HeapTypes correspond 1:1 to builder entries, so it takes less code find them that way.

kripken: I see, sounds good.

src/wasm/wasm-type.cpp
tlively: IIUC, the variant of the `erase` method that takes iterators only accepts iterators into the collection it is being called on. So I don't think it's valid to pass iterators of `selfReferentialHeapTypes` to the `erase` method of the sets in `childrenDAG`. I could still invert the loops, but then there would be more loops overall :( Overall I prefer the current version.

```
  for (TypeID id : selfReferentialHeapTypes) {
    childrenDAG.erase(id);
  }
  for (auto& kv : childrenDAG) {
    for (TypeID id : selfReferentialHeapTypes) {
      kv.second.erase(id);
    }
  }
```

kripken: Oh right, I mixed up that API...

Given it's more code it's not worth doing this possible optimization now, I agree. Maybe a TODO to investigate? Then if we profile in the future and see this function is slow, we'd have a hint.

kripken (approved):

PR #3604, opened 2021-02-24
Slightly improve new table C APIs

dcodeIO:
Uses `BinaryenIndex` instead of `int` to mirror parameter types in table construction, and adds setters for `name`, `initial` and `max`. Plus descriptions and tests.


martianboy:
Thanks for this. I'm still working to implement the  tables part of reference types proposal, so more changes to the C APIs will be coming.

tlively (approved):
LGTM! Thanks for the fix ups :)

PR #3608, opened 2021-02-25
[Wasm GC] Print Struct field names

kripken:


src/passes/Print.cpp
```diff
@@ -2838,12 +2869,18 @@ struct PrintSExpression : public OverriddenVisitor<PrintSExpression> {
     handleFieldBody(curr.element);
     o << ')';
   }
-  void handleStruct(const Struct& curr) {
+  void handleStruct(HeapType type) {
+    const auto& fields = type.getStruct().fields;
```
tlively, line 2873: Would it be possible to have this continue taking a `const Struct& curr` and have it implicitly reconstruct the HeapType for the call to `processFieldName` below? That will be less efficient, but I think it will be clearer, and I don't think this code is performance-sensitive.

kripken: I don't think that's possible. `Struct`s do not have a reference to their corresponding `HeapType`.

kripken: (But I agree, that could have been clearer.)

tlively: Thanks to global canonicalization, you will get the original HeapType back when you call the `HeapType::HeapType(const Struct&)` constructor. It's also intentionally not `explicit`, so you don't actually have to write any code to use it.

kripken: Oh, I hadn't even considered that... yes, it could work, though it seems roundabout. It is cleaner on the one hand, but will probably stick out to people reading the full logic, I worry. But it may be better than the alternative, done.

tlively (approved):
Looks good!

PR #3609, opened 2021-02-25
Refactor name processing (escaping/deduplication) to a shared place. NFC

kripken:
(not 100% NFC since it also fixes a bug by moving a line out of
a loop)


tlively (approved):
Nice 👍

PR #3614, opened 2021-02-25
Adds BinaryenI31GetGetIsSigned to API for consistency

### LINKED ISSUE #3613: API function name `BinaryenI31GetIsSigned` is not consistent
The API function name `BinaryenI31GetIsSigned` is not consistent and should probably be instead `BinaryenI31GetGetIsSigned`.

https://github.com/WebAssembly/binaryen/blob/d1d631d5f354ef3470c15003118d1c750b557fe2/src/binaryen-c.h#L1901

At this point I am not sure what can be done since it's part of the API header and I am not sure of the policies to change this. Either 

1. it's left as it is, 
2. we deprecate this function and add the correct name for it, 
3. or we just change it. 

The first option is not great due to consistency of the API naming. The last will break those relying on the new name. I will propose a PR for alternative 2. However, if the maintainers prefer another alternative altogether, feel free to close the PR.

pmatos:
Marks BinaryenI31GetIsSigned as deprecated.

Fixes #3613

### DIFF
--- a/src/binaryen-c.cpp
+++ b/src/binaryen-c.cpp
@@ -3075,10 +3075,13 @@ void BinaryenI31GetSetI31(BinaryenExpressionRef expr,
   assert(i31Expr);
   static_cast<I31Get*>(expression)->i31 = (Expression*)i31Expr;
 }
-int BinaryenI31GetIsSigned(BinaryenExpressionRef expr) {
+int BinaryenI31GetGetIsSigned(BinaryenExpressionRef expr) {
   auto* expression = (Expression*)expr;
   assert(expression->is<I31Get>());
   return static_cast<I31Get*>(expression)->signed_;
+}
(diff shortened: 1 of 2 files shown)


dcodeIO:
Makes me wonder: Isn't "Binaryen I31Get IsSigned" good already? Convention so far has been to use `Is` for bools, and `Get` otherwise if I'm not mistaken.

tlively:
Yes, it looks like we do have a convention of omitting the `Get` when we have `Is` instead, so I don't think this is necessary.

pmatos:
Ah, yes, after a closer look and comparison with for example,`BinaryenLoad*Signed`, I can see we use `Is` instead of `Get` and for the `Set` we omit the `Is`. :)

@dcodeIO  @tlively thanks for the comments.

PR #3615, opened 2021-02-25
Support Type names in the Names section

kripken:


tlively (approved):
This doesn't handle field names yet, right? Do you plan a separate subsection for those?

kripken:
Yes, that will be in the next PR.

PR #3616, opened 2021-02-25
Remove .o and .obj from gitignore

sbc100:
Cmake puts all its object files inside CMakeFiles so they
are already handled by the CMakeFiles entry.

Having `*.o` in this file means that `git status` and `git clean`
don't help to find stray `.o` files that one might have made
inadvertently in the tree.

I personally often create small source files and object files
at the top level when working on a project and I expect them
to be cleanup by `git clean`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,8 +22,6 @@ CMakeFiles
 /bin/
 /lib/
 /config.h
-*.o
-*.obj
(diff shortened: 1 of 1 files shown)


tlively (approved):
Sounds useful to me!

PR #3617, opened 2021-02-25
[Wasm GC] Fix the order of operands in array.new and struct.new

kripken:


test/let.txt
```diff
@@ -0,0 +1,42 @@
+;; source code for let.wasm, which was created by wasp
```
tlively, line 1: Is this file just a drive-by fix because it was previously missing?

tlively (approved):

test/let.txt
kripken: Yes. (It was useful to debug this.)

PR #3618, opened 2021-02-25
Simplify printing code

kripken:
We can just iterate on children using the standard order as in
delegates.h, as used by the binary format as well. The only
exceptions are control flow instructions which need some
special handling.

### DIFF
--- a/src/passes/Print.cpp
+++ b/src/passes/Print.cpp
@@ -18,6 +18,7 @@
 // Print out text in s-expression format
 //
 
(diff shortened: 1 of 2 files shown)


tlively (approved):
Strong LGTM! 🎉 🎉 🎉

kripken:
Heh, tests fail because the tests will only pass after https://github.com/WebAssembly/binaryen/pull/3617 lands - proving that will avoid such bugs going forward.

PR #3619, opened 2021-02-26
Update clang-tidy-diff.sh for GH Actions

tlively:


scripts/clang-tidy-diff.sh
```diff
@@ -27,11 +21,11 @@ if [ ! -e "$CLANG_TIDY_DIFF" ]; then
   echo "Failed to find clang-tidy-diff.py ($CLANG_TIDY_DIFF)"
   exit 1
 fi
-TIDY_MSG=$(git diff -U0 $BRANCH... | $CLANG_TIDY_DIFF -quiet -p1 2> /dev/null)
+TIDY_MSG=$(git diff -U0 $GITHUB_BASE_REF $GITHUB_HEAD_REF | $CLANG_TIDY_DIFF -quiet -p1 2> /dev/null)
```
sbc100, line 24: I think the `...` is probably still a good thing just to be precise here.   We only care about change that on the RHS here.   I know its unlikely that the BASE could have change that are not in HEAD, but in theory its possible.

scripts/clang-tidy-diff.sh
```diff
@@ -4,17 +4,11 @@ set -o errexit
...
+if [ -z "$GITHUB_BASE_REF" ]; then
+  echo "Skipping since not running on GitHub Actions"
   exit 0
 fi
 
-if [ -n "$TRAVIS_BRANCH" ]; then
-  BRANCH=$TRAVIS_BRANCH
-else
-  BRANCH=origin/main
-fi
```
sbc100, line 16: I'd love to be able to run this script locally too.    And we have the head and base default to `HEAD` and `{upstream}` ?   I'm not sure if github actions is setup such that these will also just work there too, but maybe?

sbc100: Ideally I think the diff should look something like this `git diff "@{upstream}..."`     (using the magic upstream thing and the triple dots)

sbc100: On travis this didn't work IIRC because the upstream wasn't set correctly... but maybe it is on github?   Either way that is what we want the default / local version to be I believe.

tlively:
@sbc100 This is the first thing that I've actually been able to get working. It seems that GH Actions doesn't actually check out a real branch by default? WDYT?

scripts/clang-tidy-diff.sh
```diff
@@ -27,11 +14,11 @@ if [ ! -e "$CLANG_TIDY_DIFF" ]; then
   echo "Failed to find clang-tidy-diff.py ($CLANG_TIDY_DIFF)"
   exit 1
 fi
-TIDY_MSG=$(git diff -U0 $BRANCH... | $CLANG_TIDY_DIFF -quiet -p1 2> /dev/null)
+TIDY_MSG=$(git diff -U0 origin/$GITHUB_BASE_REF | $CLANG_TIDY_DIFF -quiet -p1 2> /dev/null)
```
sbc100, line 17: Can we not keep $BRANCH (perhaps call it $BASE if you like) and have it default to something that works when run locally?

tlively: Ah, good idea 👍

tlively:
Ok, this seems to work as desired now, so I'm going to revert the test error.

scripts/clang-tidy-diff.sh
```diff
@@ -2,25 +2,18 @@
...
 CLANG_TIDY=$(which clang-tidy)
 if [ ! -e "$CLANG_TIDY" ]; then
   echo "Failed to find clang-tidy ($CLANG_TIDY)"
   exit 1
 fi
 
+if [ -n "$GITHUB_BASE_REF" ]; then
+  BRANCH="origin/$GITHUB_BASE_REF"
+else
+  BRANCH=origin/main
```
sbc100, line 14: Goes \"@{upstream}"\ work here?

sbc100 (approved):
lgtm with comment/question.

scripts/clang-tidy-diff.sh
sbc100: Also diff would be smaller/more readable if you didn't move this block down.

aheejin (approved):
LGTM % @sbc100's comments too. Thanks!

PR #3621, opened 2021-02-26
Use Tarjan's SCC alg. to find self-referential types

tlively:
Previously we computed the fixed point of the parent-child relation to identify
self-referential HeapTypes in the TypeBuilder canonicalizer. That algorithm was
O(|V|^3) in the worst case and took over five seconds to find the
self-referential HeapTypes in an example program with just 1134 HeapTypes,
probably due to high allocation traffic from the std::unordered_map and
std::unordered_sets used to implement the parent-child graph's adjacency list.

This PR replaces that algorithm with Tarjan's strongly connected component
algorithm, which runs in O(|V|+|E|) and finds the self-referential HeapTypes in
the mentioned example program in under 30 ms. All strongly connected components
of more than one element in the HeapType parent-child graph correspond to sets
of mutually recursive HeapTypes that are therefore self-referential. The only
other self-referential HeapTypes are those that are not mutually recursive with
any other HeapTypes, but these are trivial to find because they must be their
own direct children.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -1448,10 +1448,10 @@ void Canonicalizer::findSelfReferentialHeapTypes() {
   // Use Tarjan's strongly connected components algorithm on the parent-child
   // graph to find self-referential types in O(|V|+|E|) time. Each HeapType in a
   // strongly connected component with multiple elements must be
-  // self-referential because it is mutually recursive with all other
-  // HeapTypes in that strongly connected component. HeapTypes in strongly
-  // connected components of size one may also be self-referential, but it is
-  // trivial to find these because they must be their own direct children. See
+  // self-referential because it is mutually recursive with all other HeapTypes
+  // in that strongly connected component. HeapTypes in strongly connected
+  // components of size one may also be self-referential, but it is trivial to
+  // find these because they must be their own direct children. See
   // https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm.
 
   // Get the HeapType children of a HeapType, skipping all intermediate Types.
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -1301,7 +1301,7 @@ struct Canonicalizer {
   std::unordered_map<HeapType, std::vector<HeapType*>> heapTypeLocations;
 
   // These heap types will not participate in canonicalization.
-  std::unordered_set<TypeID> selfReferentialHeapTypes;
+  std::unordered_set<HeapType> selfReferentialHeapTypes;
 
   // Maps Types and HeapTypes backed by the TypeBuilder's Stores to globally
   // canonical Types and HeapTypes.
@@ -1381,7 +1381,7 @@ Canonicalizer::Canonicalizer(TypeBuilder& builder) : builder(builder) {
   // self-referential HeapTypes to the global store so that they will outlive
   // the TypeBuilder without their IDs changing.
   for (auto& entry : builder.impl->entries) {
-    if (selfReferentialHeapTypes.count(entry.get().getID())) {
+    if (selfReferentialHeapTypes.count(entry.get())) {
       noncanonicalHeapTypes.emplace_back(std::move(entry.info));
     }
   }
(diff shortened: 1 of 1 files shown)


src/wasm/wasm-type.cpp
```diff
@@ -1445,33 +1445,92 @@ void Canonicalizer::scanType(Type* type) {
...
-    for (auto& entry : fixedPoint) {
-      auto& succs = entry.second;
-      std::unordered_set<TypeID> newSuccs;
-      for (auto& other : succs) {
-        auto& otherSuccs = fixedPoint[other];
-        newSuccs.insert(otherSuccs.begin(), otherSuccs.end());
+  // Use Tarjan's strongly connected components algorithm on the parent-child
+  // graph to find self-referential types in O(|V|+|E|) time. All HeapTypes in a
+  // strongly connected component with multiple elements must be
+  // self-referential because it is mutually recursive with all other HeapTypes
```
kripken, line 1451:
```suggestion
  // self-referential because they are mutually recursive with all other HeapTypes
```

src/wasm/wasm-type.cpp
```diff
@@ -1445,33 +1445,92 @@ void Canonicalizer::scanType(Type* type) {
...
+  // in its strongly connected component. HeapTypes in strongly connected
```
kripken, line 1452:
```suggestion
  // in that strongly connected component. HeapTypes in strongly connected
```

src/wasm/wasm-type.cpp
```diff
@@ -1445,33 +1445,92 @@ void Canonicalizer::scanType(Type* type) {
...
+    size_t index = indices.size();
+    indices[curr] = index;
+    minReachable[curr] = index;
+    stack.push_back(curr);
+    stackElems.insert(curr);
+
+    for (HeapType child : getChildren(curr)) {
+      if (!indices.count(child)) {
+        // Child has not been visited yet; visit it.
+        visit(child);
```
kripken, line 1497: We may need to un-recurse this at some point. But two stacks is pretty messy so hopefully we don't...

src/wasm/wasm-type.cpp
```diff
@@ -1445,33 +1445,92 @@ void Canonicalizer::scanType(Type* type) {
...
+        minReachable[curr] = std::min(minReachable[curr], minReachable[child]);
+      } else if (stackElems.count(child)) {
+        // Child was already visited but not committed to a strongly connected
+        // component.
+        minReachable[curr] = std::min(minReachable[curr], indices[child]);
```
kripken, line 1502: This implements the original algorithm but as pointed out in the wikipedia page (the comment with "`that is deliberate and from the original paper`"), that is confusing and unnecessary. The entire code could be
```cpp
if (!indices.count(child)) {
  // Visit the child for the first time, and lower down update minReachable
  // afterwards.
  visit(child);
} else if (!stackElems.count(child)) {
  // The child is not on the stack, so skip the update of minReachable
  // afterwards.
  continue;
}
minReachable[curr] = std::min(minReachable[curr], minReachable[child]);
```
This is shorter and also properly preserves the invariant.

Alternatively, you can replace `indices[child]` with `minReachable[child]` in your code (which would make the two update lines identical, and which my version deduplicates). I don't feel strongly between the two.

(The difference is not observable in the algorithm, so both are correct, of course, as the crucial thing is that minReachable is lower than the current node, and not by how much.)

kripken (commented):
Nice work!

On a minor note I am confused why this can be done on `HeapType`s. Are those not still being constructed here?

src/wasm/wasm-type.cpp
tlively: That comment in the Wikipedia pseudocode is calling the line out because it is suprising but necessary. Later in the article it says,

> Note that v.lowlink := min(v.lowlink, w.index) is the correct way to update v.lowlink if w is on stack. Because w is on the stack already, (v, w) is a back-edge in the DFS tree and therefore w is not in the subtree of v. Because v.lowlink takes into account nodes reachable only through the nodes in the subtree of v we must stop at w and use w.index instead of w.lowlink

tlively: To be fair, I don't quite understand why this is necessary, but I'd rather not diverge from the Wikipedia explanation without deep understanding and confidence.

kripken: Oh, yes - I missed that. Yes, I was wrong, it would be incorrect to mark a lower index in that case. (It would still work I think, but the invariant would be wrong: `v` cannot necessarily reach everything `w` reaches, if `w` was before it on the stack.)

Sgtm to keep this part as is.

tlively:
> Nice work!
> 
> On a minor note I am confused why this can be done on `HeapType`s. Are those not still being constructed here?

Thanks! You're right that the canonical HeapTypes are still being constructed; the HeapTypes involved here are the TypeBuilder's temporary, non-canonical HeapTypes.

kripken (approved):

PR #3623, opened 2021-02-26
[Wasm GC] Add test/spec/br_on_null.wast and validation fixes for it

kripken:
This adds `ValidationBuilder` which can allow sharing of builder code that also
validates, between the text and binary parsers. In general we share that code in
the validator, but the validator can only run once IR exists, and in some cases we
can't even emit valid IR structure at all.


tlively (commented):
Does the new test file exercise any of the new functionality?

kripken:
The new test file exercises all the changes here. Without them it will not pass.

src/wasm-builder.h
```diff
@@ -1116,6 +1117,65 @@ class Builder {
...
+        // create a br_on_cast in such a case, as we'd have no idea what it
+        // casts to.
+        return makeSequence(makeDrop(ref), rtt);
+      }
+    }
+    if (op == BrOnNull) {
+      if (!ref->type.isRef() && ref->type != Type::unreachable) {
+        throw ParseException("Invalid ref for br_on_null", line, col);
+      }
+    }
```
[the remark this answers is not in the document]
kripken: The text and binary parsers want to call `finalize()` after creating their nodes. But that method assumes the IR is valid enough to have a heap type there,

https://github.com/WebAssembly/binaryen/blob/1ba4d259616a71a6a80662f1dff95fc125ecac2e/src/wasm/wasm.cpp#L945-L949

So this check needs to happen in the building of the IR, not the validator.

src/wasm/wasm-s-parser.cpp
```diff
@@ -2380,17 +2370,8 @@ Expression* SExpressionWasmBuilder::makeBrOn(Element& s, BrOnOp op) {
...
-    if (rtt->type == Type::unreachable) {
-      // An unreachable rtt is not supported: the text format does not provide
-      // the type, so if it's unreachable we should not even create a br_on_cast
-      // in such a case, as we'd have no idea what it casts to.
-      return builder.makeSequence(builder.makeDrop(ref), rtt);
-    }
-  }
-  return builder.makeBrOn(op, name, ref, rtt);
+  return ValidatingBuilder(wasm, s.line, s.col)
+    .validateAndMakeBrOn(op, name, ref, rtt);
```
tlively, line 2374: The nullptr rtt is being dereferenced by the builder when op is BrOnCast.

tlively:
Ohhh, I was looking for assertions about invalid modules, but this is valid WebAssembly that cannot be turned directly into valid Binaryen IR. Got it 👍

src/wasm/wasm-s-parser.cpp
kripken: Thanks, oops... fixed.

src/wasm/wasm.cpp
```diff
@@ -1361,4 +1361,90 @@ void Module::updateMaps() {
 
 void Module::clearDebugInfo() { debugInfoFileNames.clear(); }
 
+// UniqueNameMapper
```
tlively, line 1364: IMO it would make sense to introduce a new .cpp file to correspond to parsing.h. If that were a separate PR, it would make sense to move the rest of the implementation code in parsing.h into the new .cpp file as well.

tlively (commented):
Besides the parsing.h code movement comment, LGTM!

src/wasm/wasm.cpp
kripken: Fair enough, that is worth the work. Done in #3635.

tlively (approved):
LGTM rebased on #3635 👍

PR #3624, opened 2021-02-27
Support printing recursive types

tlively:
Also fixes a few locations in Print.cpp where types were being printed directly
rather than going through the s-expression type printer and removes vestigial
wrapper types that were no longer used.


kripken (approved):
Very nice!

PR #3626, opened 2021-02-27
Consider self-referential HeapTypes to be canonical

tlively:
This PR makes the TypeBuilder move self-referential HeapTypes to global HeapType
store so that they are considered canonical. This means that when a HeapType is
constructed with an identical definition, it will be equivalent to the original
HeapType constructed by the TypeBuilder, but it does _not_ mean that
self-referential HeapTypes are deduplicated.

This fixes a bug in which two versions of each self-referential function
signature were emitted. Before this PR, the global HeapType store was not aware
of the original self-referential HeapTypes. When the function signatures were
used to construct HeapTypes during type collection, new HeapTypes were allocated
and emitted in addition to the original HeapTypes. Now the global HeapType store
returns the original HeapTypes, so the extra HeapType is never allocated.

@kripken you were totally right when you said you expected to see a move here!

### DIFF
--- a/test/lit/recursive-types.wast
+++ b/test/lit/recursive-types.wast
@@ -2,20 +2,12 @@
 
 ;; RUN: wasm-opt %s -all -S -o - | filecheck %s
 
-;; TODO: Fix the bug where self-referential function signatures are emitted
-;; twice. This happens because collectHeapTypes has to reconstruct a HeapType
-;; from each function's signature, but self-referential HeapTypes aren't
-;; canonicalized when they are parsed so collectHeapTypes ends up with a
-;; separate, unfolded version of the type.
-
 ;; TODO: Fix the bug where structurally identical types are given the same
 ;; generated name, making the wast invalid due to duplicate names.
 
 ;; CHECK:      (module
 ;; CHECK-NEXT:   (type $ref?|...0|_=>_ref?|...0| (func (param (ref null $ref?|...0|_=>_ref?|...0|)) (result (ref null $ref?|...0|_=>_ref?|...0|))))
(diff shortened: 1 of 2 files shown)


tlively:
Automerge enabled, FYI.

test/lit/recursive-types.wast
```diff
@@ -2,20 +2,12 @@
...
-;; from each function's signature, but self-referential HeapTypes aren't
-;; canonicalized when they are parsed so collectHeapTypes ends up with a
-;; separate, unfolded version of the type.
-
 ;; TODO: Fix the bug where structurally identical types are given the same
 ;; generated name, making the wast invalid due to duplicate names.
 
 ;; CHECK:      (module
 ;; CHECK-NEXT:   (type $ref?|...0|_=>_ref?|...0| (func (param (ref null $ref?|...0|_=>_ref?|...0|)) (result (ref null $ref?|...0|_=>_ref?|...0|))))
 ;; CHECK-NEXT:   (type $ref?|...0|_=>_ref?|...0| (func (param (ref null $ref?|...0|_=>_ref?|...0|)) (result (ref null $ref?|...0|_=>_ref?|...0|))))
```
kripken, line 10: Unrelated to this PR, this output looks like it emits two types with the exact same name, which should be invalid - that is, I think this file would error on round-tripping through the text format?

(I guess #3627 could fix that?)

kripken (approved):

test/lit/recursive-types.wast
tlively: Yep, correct on both counts 👍

PR #3632, opened 2021-02-28
[Wasm Exceptions] Properly ensure unique try labels after an inlining

kripken:
The old code here just referred to Block and Loop. Refactor it to use the
generic helper code that also handles Try.

Also add validation of Try names in the validator.

The testcase here would have `$label` appear twice before this fix. After
the fix there is `$label0` for one of them.


aheejin (approved):
Thank you!

PR #3633, opened 2021-03-01
[Wasm Exceptions] Fix/work around delegate 0 issues in Inlining pass

kripken:
1. Ignore the fake delegate target in the unique name mapper. The mapper
  is run after inlining, so this fixes inlining into a function that has a delegate 0.
2. Do not inline a function with delegate 0. We should support this eventually,
  but for now I think this is good enough.

(Btw what are the proper semantics of an inlined delegate 0? Should it delegate
to a new try at the inlining site, and do a rethrow from there, something like
that?)

After this Inlining should be safe to run on exceptions code.

### DIFF
--- a/src/parsing.h
+++ b/src/parsing.h
@@ -321,6 +321,11 @@ struct UniqueNameMapper {
   }
 
   Name sourceToUnique(Name sName) {
+    // DELEGATE_CALLER_TARGET is a fake target used to denote delegating to the
+    // caller. We do not need to modify it, as it has no definitions, only uses.
+    if (sName == DELEGATE_CALLER_TARGET) {
+      return DELEGATE_CALLER_TARGET;
+    }
     if (labelMappings.find(sName) == labelMappings.end()) {
       throw ParseException("bad label in sourceToUnique");
     }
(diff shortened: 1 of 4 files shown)


aheejin (approved):
Thanks! 0 does not mean any special thing. It can be any number. We just write it as a number when it targets the fake label `$DELEGATE_CALLER_TARGET`, because we can't write that label in wast file.

```wasm
(try
  (do)
  (try
    (do)
    (delegate 1)
  )
  (catch_all)
)
```
So in this example it is `delegate 1` that targets the fake caller label. (0 targets the outer try, but it would have been written as a label.) So if the immediate can be written as a try label within the function, we do that for readability, and if it targets the caller, we write that as an immediate number as we write it in binary.

I think it makes sense to disable inlining functions with try-delegates into another function, because we have to figure out what the next level `try` in that inlined-into function, which I think is possible but not urgent at the moment.

I'm not sure if there is any issue (other than the bug you fix) about inlining another function into a function that has try-delegates, unless that inlining changes the block depth that wraps try-delegates, which can change the meaning of try-delegate. Can the inlining change that?

kripken:
> 0 does not mean any special thing.

Oh sorry, I had that wrong. Yes, I should have written "delegate to the caller" in all the places I wrote "delegate 0"...

> Can the inlining change that?

I don't think so, that should be safe. Inlining just adds new code to replace a call, and for now we'll disallow any delegates in that new code. So the outside code should not notice any change.

PR #3637, opened 2021-03-02
wasm-autogen tool

kripken:
This implements one possible way of autogenerating C API code. In the future the JS API could also do this.

The approach taken here is:

* A C++ tool, `wasm-autogen` is created. It uses `wasm-delegates-fields` to emit `.h` and `.cpp` code for the C API.
* Those files are checked in as `*.autogen.h / *.autogen.cpp`. You can see them in this PR, implementing the `Struct` operations. (Note: that is after clang-format has been run on them.)
* Then the C API code is just normal code, and could be tested etc. (Note: I didn't quite finish writing a test for this in this PR, let me know if that would help the discussion.)

I am actually not that happy with this. I feel like a Python script could be simpler and cleaner. One issue is the hacks - both macros and templates - in order to do what amounts to string processing. Look at the code and tell me what you think... Another issue is that iteration on this tool is slightly annoying - imagine that you are in an intermediate step where `wasm-autogen` emits bad C API code. Then building `wasm-autogen` will fail (to avoid that, we'd need to get `wasm-autogen` to not depend on `libbinaryen.so` which includes the C API, perhaps splitting it up or such).

Instead, I suspect it would be nicer to parse `wasm-delegates-fields` from Python and do things there. We had a debate about this and decided to try to do more in C++; I think the above paragraph provides some new reasons to avoid C++ there.

cc @aardappel @tlively @aheejin @dschuff @dcodeIO 

Side issue, this reorders one wasm-builder method for consistency, as found by starting to autogenerate things.


src/tools/wasm-autogen.cpp
```diff
@@ -0,0 +1,219 @@
...
+using namespace wasm;
+
+namespace {
+
+// Emit a declaration. If "startOfImpl" is true then we emit the beginning of
+// the implementation, which is identical to the declaration except it has no
+// ";" at the end.
+template<typename T> void autogenOneCAPIDecl(bool startOfImpl = false) {
+  MixedArena allocator;
+  T curr(allocator);
```
kripken, line 33: This creates an instance of the class as a hack, so that we can use the delegates code on it.

sbc100:
nit: `autogen` is a pretty generic term.   If there is a tool it should probably include the term "API" or interface somewhere .. something like `-api-gen` perhaps?

dcodeIO:
Looks great! That'll certainly help a lot with all the boilerplate. I don't have a strong opinion on whether to use C++ or Python for it. As long as it is easy to maintain and makes our lives easier in the long run, that's fine for me. Regarding Python: One needs Python anyway to do anything of significance with the code base, like updating or running tests. Seems rare to generate something new and then don't test it.

tlively:
I came in expecting to strongly prefer a Python approach, but actually this looks pretty reasonable to me. It's certainly not beautiful code, but I don't think the Python version would be either, especially since it would have to parse the delegation header as well. I would be happy to be proven wrong about that, though!

src/tools/wasm-autogen.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+  std::cout << '\n';
+
+#define DELEGATE_ID curr._id
+
+#define DELEGATE_START(id)                                                     \
+  std::cout << "BINARYEN_API BinaryenExpressionRef\n"                          \
+            << "Binaryen" << #id << "(BinaryenModuleRef module";
+
+#define DELEGATE_FIELD_CHILD(id, name)                                         \
```
aardappel, line 43: why are these macros? looks like some inline functions would not be that much more verbose?

aardappel (commented):
Agree with @tlively that this looks like an uncomplicated amount of C++, and that having it in C++ has maintenance advantages (as discussed before). But I guess a lot of that is a cultural issue.

src/tools/wasm-autogen.cpp
tlively: This lets them all refer to their shared context, which is useful because theses are usually used to generate fragments of a single larger function.

kripken:
Ok, sounds like this is the direction most people prefer then, sgtm.

Updated with code to emit basically all the missing GC instructions. Testing is not yet there.

kripken:
Added a simple test to verify at least some of the autogenerated API works. (It's hard to add much without more work on creating GC types in the C API.) I think this is ready for review now.

src/tools/wasm-api-gen.cpp
```diff
@@ -0,0 +1,233 @@
...
+#define DELEGATE_START(id) className = #id;
+
+#define DELEGATE_FIELD_CHILD(id, name) params.push_back(#name);
+
+#define DELEGATE_FIELD_INT(id, name)                                           \
+  if (hasOp<T>()) {                                                            \
+    params.push_back(#id "Op(" #name ")");                                     \
+  } else {                                                                     \
+    params.push_back(#name);                                                   \
+  }
```
tlively, line 132: Some SIMD instructions have both op and other int operands, so this won't scale. Perhaps Op operands should get their own delegate to differentiate them?

tlively (commented):
Do we plan to replace the existing hand-written code with generated code eventually?

src/tools/wasm-api-gen.cpp
kripken: Good point. Yeah, we'll need to be more careful as we roll this out to replace the handwritten ones, which I think we should do eventually.

A new delegate is one option. Another might be to use the order, as the `op` is always first, so we could do that if-else only on the first item in the list.

kripken:
Hmm, I guess maybe we could avoid the `.h` file being checked in... I actually didn't even consider that at all until you mentioned it!

It feels natural to me to create a simple static `.h` file for C API users. They just want to see the C API, and not care about how it was created. That is, the difference is that this output will be seen by users of binaryen and not just the binaryen developers.

dcodeIO:
👍  Also helps when interested in the Git history of the C-API, e.g. to see what has been added or changed recently.

PR #3639, opened 2021-03-02
[Wasm Exceptions] Fix a handwritten exceptions test

kripken:
`catch_all` should not have a pop.


aheejin (approved):
Thanks! By the way it looks we need to update the output too.

PR #3640, opened 2021-03-02
Make the reduction script more robust, and document text reduction

kripken:
The check for a valid wasm file must be different if the wasm has
a feature section or not, so just try both ways, with `--detect-features`
and `--all-features`. If the wasm is valid, at least one will work.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -1207,9 +1207,10 @@ At least one of the next two values should be 0:
 The following value should be 1:
   1
 
-(If it does not, then one possible issue is that -ttf fails to write a valid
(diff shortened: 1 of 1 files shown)


scripts/fuzz_opt.py
```diff
@@ -1195,10 +1199,17 @@ def randomize_opt_flags():
...
 
-should be 0: 0
-should be 1: 1
+At least one of the next two values should be 0:
+  0
+  1
+The following value should be 1:
+  1
+
+(If it does not, then one possible issue is that -ttf fails to write a valid
```
tlively, line 1210: I don't remember what -ttf stands for. Could we use the full length option name here instead?

tlively (approved):

PR #3641, opened 2021-03-02
[Wasm GC] ref.cast should have zero immediates in current spec and prototype v3

kripken:


kripken:
The same is true of `ref.test`, fixing here as well.

tlively (approved):

Issue #3459, opened 2020-12-21
Release 98 lacks Linux build


kripken:
The others are there, but oddly linux is not,

https://github.com/WebAssembly/binaryen/releases/tag/version_98

@sbc100 did we change something there?

Nothing looks obviously wrong in the scripts, and I'm not sure where I would look for a log with an error on it for a failed build...

sbc100:
The failed build shows up here: https://github.com/WebAssembly/binaryen/actions/runs/304698147

sbc100:
"Error: The operation was canceled."

sbc100:
Sadly I don't see any way to "retry" the github action.   Maybe lets just try tagging 99 now?

sbditto85:
when I try to build based off of the tag "version_98" in my linux container I get a compiler error. Not sure if that is just my setup though.

sbc100:
Can you share the error you get?  I imagine its unrelated as our CI is green for that revision.

kripken:
Yes, let's try tagging now to see if it works. I can do it, unless there's a reason not to atm? (roller is green, so I'm not aware of any issues)

sbc100:
Sounds good!  Please go ahead.

kripken:
https://github.com/WebAssembly/binaryen/releases/tag/version_99 , let's see how it goes...

kripken:
The linux build failed again. https://github.com/WebAssembly/binaryen/pull/3475 *may* fix it. That is definitely a fix for a problem that would have that symptom. But the testcase that fails should not hit it, and does not locally, so I am not sure. But let's see after it lands.

(Is there an easy way to run the release action in a test mode for debugging?)

sbc100:
> (Is there an easy way to run the release action in a test mode for debugging?)

Not that I know of... if you figure out a why I would be very interested to know about it.  Kind of annoying that you can re-run actions or trigger them manually :(   (I mean you can probably do those things but I have no idea how to).

TriplEight:
I can't see the Linux bin in 99 either.

sbc100:
Indeed it looks like the alpine builder is failing during testing for some reason: https://github.com/WebAssembly/binaryen/runs/1669944081?check_suite_focus=true

I guess we should add alpine to the normal CI avoid the difference here.

TriplEight:
right, or release with ubuntu

sbc100:
[...]

PR #3642, opened 2021-03-02
ci: Build alpine linux to match what the release builder does

sbc100:
Also fix the underlying issue which was that `lit` was not
being correctly installed via pip3.

See #3459

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,6 +148,42 @@ jobs:
     - name: test
       run: python check.py --binaryen-bin=out/bin
 
+  # Build with gcc 6.3 and run tests on Alpine Linux (inside chroot).
+  # Note: Alpine uses musl libc.
+  # Keep in sync with build_release.yml
+  build-alpine:
+    name: alpine
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/setup-python@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -148,6 +148,42 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
+    - uses: actions/checkout@v1
+    - name: start docker
+      run: |
+        docker run -w /src -dit --name alpine -v $PWD:/src node:lts-alpine
+        echo 'docker exec alpine "$@";' > ./alpine.sh
```
tlively, line 165: Is this the only thing in alpine.sh? Why not write the "docker exec..." command out everywhere instead? That's slightly more duplication, but also seems more straightforward. My first thought when I saw a reference to alpine.sh was that it was a file I would be able to find in the Binaryen source tree.

kripken (approved):

.github/workflows/ci.yml
sbc100: I didn't write this code, I'm just copying it from build_release.yml .. if we want to make that change its can be done as a followup I think,

PR #3643, opened 2021-03-02
Remove PostAssemblyScript passes

dcodeIO:
AS does not use these passes anymore since switching to tracing GC, and recent Binaryen versions won't work with old versions of the compiler anymore anyway, so there is no need to keep these around.


README.md
```diff
@@ -26,7 +26,7 @@ effective**:
 
 Compilers using Binaryen include:
 
- * [`AssemblyScript`](https://github.com/AssemblyScript/assemblyscript) which compiles a subset of TypeScript to WebAssembly
+ * [`AssemblyScript`](https://github.com/AssemblyScript/assemblyscript) which compiles a strict variant of TypeScript to WebAssembly
```
tlively, line 29: Is there a difference between a "variant" and a "strict variant?"

tlively (approved):

README.md
dcodeIO: Hmm, good question. I find it hard to convey what AS is while still keeping the explanation short. What this tries to say is that AS is a variant (not purely a subset, not really a superset) of TS, but in a way that it is certainly stricter than TS, as a hint that just compiling idiomatic TS with it is not a realistic expectation at this point. Does that make sense? :)

tlively: Yes, thanks. I was wondering if there was some technical distinction here like between "subset" and "strict subset," but it sounds like this is meant to more informally describe the variant. Perhaps using "stricter" instead of "strict" would help clarify? Either way, LGTM :)

dcodeIO: Now that you mention it, yeah, that may indeed be a bit confusing. "Stricter" sounds good :)

PR #3644, opened 2021-03-02
[Wasm GC] Allow subtyping in arguments to struct.get etc.

### LINKED ISSUE #3636: Error on implicit upcast
If I give Binaryen this code, containing an implicit upcast from a (structural) subtype:
```wat
(module
    (type $A (struct (field (mut i32))))
    (type $B (struct (field (mut i32)) (field (mut i64))))
    (func $f (param $0 (ref $B))
        (struct.set $A 0 (local.get $0) (i32.const 5))
    )
)
```
I get this error message:
```
[parse exception: bad heap type: expected (struct (field (mut i32))) but found (ref null (struct (field (mut i32) (mut i64)))) (at 5:8)]
Fatal: error parsing wasm
```
It looks like the structural subtype check is not done properly for the validation here.

Even if I pass `--no-validation`, I still get the error.

kripken:
Note that Binaryen "canonicalizes" the type, so in the test output here
we end up with `$grandchild` twice. This is a consequence of us not
storing the heap type as an extra field. I can't think of a downside to
this canonicalization, aside from losing perfect roundtripping, but I think
that's a worthwhile tradeoff for efficiency as we've been thinking so far?

Fixes #3636


src/wasm/wasm-s-parser.cpp
```diff
@@ -3451,7 +3451,7 @@ void SExpressionWasmBuilder::validateHeapTypeUsingChild(Expression* child,
     return;
   }
   if ((!child->type.isRef() && !child->type.isRtt()) ||
-      child->type.getHeapType() != heapType) {
+      !HeapType::isSubType(child->type.getHeapType(), heapType)) {
```
tlively, line 3454: Could this function be deduplicated to parsing.h? The `Element` argument could become an optional pointer.

kripken: It almost can, but the mechanisms they use to throw are also different, which is why this wasn't deduplicated before.

A larger refactoring to unify how they throw might make sense I think.

tlively: ahhh, I missed that difference. Refactoring at some point makes sense to me, but definitely not urgent 👍

tlively (approved):

PR #3647, opened 2021-03-03
[Wasm GC] Parse text field names even of types that end up canonicalized together

kripken:
Names of structurally identical types end up "collapsed" together after the
types are canonicalized, but with this PR we can properly read content that
has structurally identical types with different names.


tlively (approved):
Thanks!

Issue #3499, opened 2021-01-19
Names section unaware of local reordering during binary writing


kripken:
```wat
(module
  (func $test
    (local $local_eqref eqref)
    (local $local_anyref anyref)
    (drop (local.get $local_eqref))
    (drop (local.get $local_anyref))
  )
)
```
```wat
$ wasm-opt a.wat --print -all -o b.wasm -g
(module
 (type $none_=>_none (func))
 (func $test
  (local $local_eqref eqref)
  (local $local_anyref anyref)
  (drop
   (local.get $local_eqref)
  )
  (drop
   (local.get $local_anyref)
  )
 )
)
$ wasm-dis b.wasm
(module
 (type $none_=>_none (func))
 (func $test
  (local $local_eqref anyref)
  (local $local_anyref eqref)
  (drop
   (local.get $local_anyref)
  )
  (drop
   (local.get $local_eqref)
  )
 )
)
```
Note how
```wat
  (local $local_eqref anyref)
  (local $local_anyref eqref)
```
is wrong. The problem is purely in the names section, it turns out, as the `local.gets` are valid (if they weren't the fuzzer would have found this too).

What seems to be going on is that when we write the binary, `BinaryInstWriter::mapLocalsAndEmitHeader` remaps the locals. That is more efficient (as we can put all locals of a type together). However, `WasmBinaryWriter::writeNames` is unaware of that mapping, so we write out a names section with wrong indices.

We already disable reordering when DWARF is enabled, so it may make sense to do it when the names section is present. However, this bug has been present for years, and not noticed, so it may not be urgent to fix, and worth waiting for a proper solution, which would probably be to rewrite the writing code to remember the mapping and use it in the names section.

Issue #3648, opened 2021-03-03
Strange issue with std::map<Type, ..>


kripken:
In `wasm-stack` we have `std::map<Type, size_t> numLocalsByType;`. On the testcase below something quite odd happens, on this code:
```cpp
  o << U32LEB(numLocalsByType.size());
  for (auto& typeCount : numLocalsByType) {
    o << U32LEB(typeCount.second);
    parent.writeType(typeCount.first);
  }
```
That writes the number of types on the first line, and then the type count and ID for each one. But while `.size()` reports 3, the loop only goes through 2 iterations.

This is not iterator invalidation. Removing the inner lines in the loop have no effect. (Also ubsan and asan find nothing.)

There is basically a divergence between the reported size of the container and the observed contents when iterating. Is there any situation in which that is expected to happen?

My only theory is some weirdness related to canonicalization of Types, that is, somehow 2 of the 3 keys in the map are seen as identical to each other, and we just iterate on those...? (but that doesn't seem to make sense) cc @tlively 

Testcase:
```wat
(module
 (type $a (func (param (ref null $b))))
 (type $b (struct (field (ref null $i))))
 (type $c (struct (field (ref null $l))))
 (type $d (func (param (ref null $b))))
 (type $e (func (result (ref null $g))))
 (type $f (func (result (ref null $c))))
 (type $g (struct (field (ref null $j))))
 (type $h (struct (field (ref null $k))))
 (type $i (struct (field (mut (ref null $a)))))
 (type $j (struct (field (mut (ref null $a))) (field (mut (ref null $a)))))
 (type $k (struct (field (mut (ref null $a))) (field (mut (ref null $a))) (field (mut (ref null $e)))))
 (type $l (struct (field (mut (ref null $a))) (field (mut (ref null $d))) (field (mut (ref null $f)))))

 (func $foo
  (local $1 (ref null $h))
  (local $2 (ref null $b)) ;; moving this line one down or up avoids the bug
  (local $3 (ref null $c))
 )
)
```
`wasm-opt a.wat  -all --roundtrip` will throw an exception when trying to read after writing an in-memory binary. (can also write to a file and read from there)

kripken:
(For posterity, the issue is that recursive types are not fully ordered, and `std::map` requires that, and apparently can have surprising behavior when it is not present...)

PR #3649, opened 2021-03-04
Fix binary writing of local name indexes

kripken:
When writing a binary, we take the local indexes in the IR and turn
them into the format in the binary, which clumps them by type. When
writing the names section we should be aware of that ordering, but
we never were, as noticed in https://github.com/WebAssembly/binaryen/issues/3499

This fixes that by saving the mapping of locals when we are emitting
the name section, then using it when emitting the local names.

This also fixes the order of the types themselves as part of the
refactoring. We used to depend on the ordering of types to decide
which to emit first, but that isn't good for at least two reasons. First,
it hits https://github.com/WebAssembly/binaryen/issues/3648 - that order is not fully
defined for recursive types. Also, it's not good for code size - we've
ordered the locals in a way we think is best already (ReorderLocals pass).
This PR makes us pick an order of types based on that, as much as
possible, that is, when we see a type for the first time we append it to
a list whose order we use.

Test changes: Some are just because we use a different order than
before, as in `atomics64`. But some are actual fixes, e.g. in `heap-types`
where we now have ` (local $tv (ref null $vector))` which is indeed
right - `v` there is for `vector`, and likewise `m` for `matrix` etc. - we
just had wrong names before. Another example, we now have
`(local $local_externref externref)` whereas before the name was
`funcref`, and which was wrong... seems like the incorrectness was
more common on reference types and GC types, which is why this was
not noticed before.

Fixes https://github.com/WebAssembly/binaryen/issues/3499

Makes part of https://github.com/WebAssembly/binaryen/issues/3648 moot.

### DIFF
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -338,14 +338,14 @@ void WasmBinaryWriter::writeFunctions() {
       StackIRToBinaryWriter writer(*this, o, func);
       writer.write();
       if (debugInfo) {
-        funcMappedLocals[func->name].swap(writer.getMappedLocals());
+        funcMappedLocals[func->name] = std::move(writer.getMappedLocals());
       }
     } else {
       BYN_TRACE("write Binaryen IR\n");
       BinaryenIRToBinaryWriter writer(*this, o, func, sourceMap, DWARF);
       writer.write();
       if (debugInfo) {
-        funcMappedLocals[func->name].swap(writer.getMappedLocals());
+        funcMappedLocals[func->name] = std::move(writer.getMappedLocals());
       }
     }
     size_t size = o.size() - start;
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -689,7 +689,8 @@ void WasmBinaryWriter::writeNames() {
           auto indexInFunc = indexedLocal.first;
           auto name = indexedLocal.second;
           // TODO: handle multivalue
-          auto indexInBinary = funcMappedLocals.at(func->name)[std::pair(indexInFunc, 0)];
+          auto indexInBinary =
+            funcMappedLocals.at(func->name)[std::pair(indexInFunc, 0)];
           o << U32LEB(localIndexInFunc);
           writeEscapedName(name.str);
         }
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -335,10 +335,18 @@ void WasmBinaryWriter::writeFunctions() {
     // Emit Stack IR if present, and if we can
     if (func->stackIR && !sourceMap && !DWARF) {
       BYN_TRACE("write Stack IR\n");
-      StackIRToBinaryWriter(*this, o, func).write();
+      StackIRToBinaryWriter writer(*this, o, func);
+      writer.write();
+      if (debugInfo) {
+        funcMappedLocals[func->name].swap(writer.getMappedLocals());
+      }
     } else {
       BYN_TRACE("write Binaryen IR\n");
-      BinaryenIRToBinaryWriter(*this, o, func, sourceMap, DWARF).write();
+      BinaryenIRToBinaryWriter writer(*this, o, func, sourceMap, DWARF);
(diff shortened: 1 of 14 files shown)


src/wasm/wasm-binary.cpp
```diff
@@ -335,10 +335,18 @@ void WasmBinaryWriter::writeFunctions() {
     // Emit Stack IR if present, and if we can
     if (func->stackIR && !sourceMap && !DWARF) {
       BYN_TRACE("write Stack IR\n");
-      StackIRToBinaryWriter(*this, o, func).write();
+      StackIRToBinaryWriter writer(*this, o, func);
+      writer.write();
+      if (debugInfo) {
+        funcMappedLocals[func->name].swap(writer.getMappedLocals());
```
tlively, line 341: Using `std::move` might be clearer here; `swap` makes it look like there might be something useful in the `funcMappedLocals` already. Same below.

src/wasm/wasm-binary.cpp
```diff
@@ -663,20 +671,28 @@ void WasmBinaryWriter::writeNames() {
...
         assert(localsWithNames.size());
-        o << U32LEB(indexedFunc.first);
+        o << U32LEB(index);
         o << U32LEB(localsWithNames.size());
         for (auto& indexedLocal : localsWithNames) {
-          o << U32LEB(indexedLocal.first);
-          writeEscapedName(indexedLocal.second.str);
+          auto indexInFunc = indexedLocal.first;
+          auto name = indexedLocal.second;
+          // TODO: handle multivalue
```
tlively, line 691: Do you plan to implement this soon or leave it for later?

tlively (commented):
Would you mind undoing the change from `std::map` to `std::unordered_map` in this PR? I would like to see the changes in the tests that are due just to the other changes.

tlively:
I find this code to be generally hard to understand (even though I wrote some of it myself) because of all the maps floating around as ambient state. Would it be a lot of extra work to split up the fixing of the names from the change to the local order? That might make it easier to review.

Also, I'm not sure new ordering is necessarily more principled than the old ordering. For example, there might be one extremely commonly accessed i32 local, but many more extremely rarely accessed i32 locals. It is possible that it might be better to split up the i32 locals into non-contiguous ranges. Assuming we don't want to split types into non-contiguous ranges of locals, though, it would be better to order types by the cumulative number of accesses to their locals rather than the maximum number of accesses to their locals.

src/wasm/wasm-binary.cpp
kripken: This is basically just writing out a TODO from earlier that wasn't spelled out - this never worked.

I don't have plans to work on it. I'm not sure offhand if anything can be done - the extra temp locals are not "normal", really, and may vanish in stacky form?

kripken:
> Would you mind undoing the change from std::map to std::unordered_map in this PR? I would like to see the changes in the tests that are due just to the other changes.

That part of the change has no effects. We stop iterating on the data structure, so I changed it to unordered to basically reflect that. Keeping it ordered would be confusing, in other words, but have no noticeable effect.

> I find this code to be generally hard to understand (even though I wrote some of it myself) because of all the maps floating around as ambient state. Would it be a lot of extra work to split up the fixing of the names from the change to the local order? That might make it easier to review.

I don't think it can be split up. Changing the local order would break the names section. That is, right now things "work" because we have glitches that happen to be coordinated in most cases. But fixing one will make the other much more noticeable. I think the only reasonable thing is to fix the order, and to avoid breaking things further, to fix the names section at the same time.

> Also, I'm not sure new ordering is necessarily more principled than the old ordering. For example, there might be one extremely commonly accessed i32 local, but many more extremely rarely accessed i32 locals. It is possible that it might be better to split up the i32 locals into non-contiguous ranges. Assuming we don't want to split types into non-contiguous ranges of locals, though, it would be better to order types by the cumulative number of accesses to their locals rather than the maximum number of accesses to their locals.

Definitely true, yes. This will be closer to the optimal order picked by ReorderLocals (in that before we were just "random"), but the problems you mention are possible, and we can consider possible solutions.

tlively (approved):

PR #3650, opened 2021-03-04
Fix TypeComparator to satisfy C++'s Compare requirement

### LINKED ISSUE #3648: Strange issue with std::map<Type, ..>

tlively:
The comparison implemented by TypeComparator was not previously antisymmetric
because it held that both A < B and B < A when A and B were structurally
identical but nominally distinct recursive types. This meant that it did not
satisfy the conditions of C++'s "Compare" requirement, which meant that std::set
did not operate correctly, as discovered in #3648. The PR fixes the problem by
having making A < B be false (and vice versa), making type comparisons properly
antisymmetric.

As a drive by, also switches to using std::stable_sort in collectHeapTypes to
make the order of the type section completely deterministic accross platforms.

Fixes #3648.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -100,8 +100,8 @@ struct HeapTypeInfo {
 // Helper for coinductively comparing Types and HeapTypes according to some
 // arbitrary notion of complexity.
 struct TypeComparator {
-  // Set of HeapTypes we are assuming satisfy the relation as long as we cannot
-  // prove otherwise.
+  // Set of HeapTypes we are assuming are equivalent as long as we cannot prove
+  // otherwise.
   std::unordered_set<std::pair<HeapType, HeapType>> seen;
   bool lessThan(Type a, Type b);
   bool lessThan(HeapType a, HeapType b);
@@ -118,8 +118,8 @@ struct TypeComparator {
 // Helper for coinductively checking whether a pair of Types or HeapTypes are in
 // a subtype relation.
 struct SubTyper {
-  // Set of HeapTypes we are assuming satisfy the relation as long as we cannot
-  // prove otherwise.
+  // Set of HeapTypes we are assuming are equivalent as long as we cannot prove
+  // otherwise.
   std::unordered_set<std::pair<HeapType, HeapType>> seen;
   bool isSubType(Type a, Type b);
   bool isSubType(HeapType a, HeapType b);
@@ -858,9 +858,9 @@ bool TypeComparator::lessThan(HeapType a, HeapType b) {
     return false;
   }
   if (seen.count({a, b})) {
-    // We weren't able to disprove that a < b since we last saw them, so the
-    // relation holds coinductively.
-    return true;
+    // We weren't able to disprove that a == b since we last saw them, so it
+    // holds coinductively. a < b is therefore false;
+    return false;
   }
   if (a.isBasic() && b.isBasic()) {
     return a.getBasic() < b.getBasic();
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -859,7 +859,7 @@ bool TypeComparator::lessThan(HeapType a, HeapType b) {
   }
   if (seen.count({a, b})) {
     // We weren't able to disprove that a == b since we last saw them, so it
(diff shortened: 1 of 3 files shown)


tlively:
After spending a few hours writing deterministic replacements for std::set and std::map and pulling out the use of type comparisons everywhere, I realized this was actually just a one line fix 🙃

src/wasm/wasm-type.cpp
```diff
@@ -858,9 +858,9 @@ bool TypeComparator::lessThan(HeapType a, HeapType b) {
     return false;
   }
   if (seen.count({a, b})) {
-    // We weren't able to disprove that a < b since we last saw them, so the
-    // relation holds coinductively.
-    return true;
+    // We weren't able to disprove that a == b since we last saw them, so it
+    // holds coinductively. a < b is therefore false;
+    return false;
```
kripken, line 863: It's not obvious to me why this is enough, but maybe I don't know the C++ spec well enough. Do you have a link?

My intuition is that before you could have `a < b, b < a, a != b`, while now you will have `a >= b, b >= a, a != b` which from a logical point of view is equivalent.

tlively: https://en.cppreference.com/w/cpp/named_req/Compare. The key point is that it _only_ requires < to create a weak ordering and it's ok if == is not consistent with that ordering.

src/wasm/wasm-type.cpp
```diff
@@ -858,9 +858,9 @@ bool TypeComparator::lessThan(HeapType a, HeapType b) {
     return false;
   }
   if (seen.count({a, b})) {
-    // We weren't able to disprove that a < b since we last saw them, so the
-    // relation holds coinductively.
-    return true;
+    // We weren't able to disprove that a == b since we last saw them, so it
+    // holds coinductively. a < b is therefore false;
```
kripken, line 862:
```suggestion
    // holds coinductively that a < b is therefore false.
```

src/wasm/wasm-type.cpp
kripken: I see, thanks! So yes, it must be a strict weak ordering, and so `!(a < b), !(b < a), a != b` is fine.

kripken (approved):

tlively:
I manually force-merged this because the alpine tests seemed to be hanging :/

PR #3652, opened 2021-03-04
Stop emitting features section in fuzzer

kripken:
I'm not sure what this was added - @tlively maybe you know?

It seems unneeded as we call all the commands with all the feature
args on the commandline anyhow.

Also, this caused an actual problem on a particular fuzz testcase I
found. Rather than figure out the details there, it seems simpler to not
depend on the features section while fuzzing?

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -857,7 +857,7 @@ def test_one(random_input, given_wasm):
     else:
         # emit the target features section so that reduction can work later,
         # without needing to specify the features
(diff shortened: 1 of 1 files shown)


tlively (approved):
Seems fine. I think this was a previous attempt to make reducing test cases easier, but now the reducer is more robust so this shouldn't be necessary. I'm concerned that it was causing problems, though.

kripken:
I assume the problems were that some part of the fuzzer added an extra flag or something like that. I think that shouldn't be a problem (more flags are ok, no loss in coverage).

PR #3653, opened 2021-03-04
Emit "elem declare" for functions that need it

kripken:
This adds support for reading `(elem declare func $foo ..` in the text and
binary formats. We can simply ignore it: we don't need to represent it in
IR, rather we find what needs to be declared when writing. That part takes
a little more work, for which this adds a shared helper function.

With this we should be emitting valid wasm binaries for more wasm GC
cases, and can parse wasm GC text testcases that use this.


tlively:
Note that this overlaps with #3572, but I think it still makes sense to merge this first while that PR goes through review. @martianboy, would you be able to split out the corresponding collection emission of declarative segments into its own PR as well? That way we can get ref.func working before we have full support for passive segments.

src/passes/Print.cpp
```diff
@@ -2832,6 +2833,11 @@ struct PrintSExpression : public UnifiedExpressionVisitor<PrintSExpression> {
       *curr, [&](Memory* memory) { visitMemory(memory); });
     ModuleUtils::iterDefinedTables(*curr,
                                    [&](Table* table) { visitTable(table); });
+    for (auto name : TableUtils::getFunctionsNeedingElemDeclare(*curr)) {
+      doIndent(o, indent);
+      printMedium(o, "(elem")
+        << " declare func $" << name << ')' << maybeNewLine;
+    }
```
tlively, line 2840: Instead of printing a separate declarative segment for each function, can we put all the functions into one declarative segment?

kripken:
Sorry, I didn't realize the overlap with that PR! I hadn't known that passive/active was related to declaration of items... I only noticed a  connection to passive segments in the binary format but I thought it was a minor optimization of the bitfield of flags... So the PR name didn't suggest to me an overlap.

I'm ok to do the order any way we prefer (if that one is close I can close this, too) - let me know what makes sense.

src/passes/Print.cpp
kripken: We can, yes. I was copying from a partner source file in which there was one each, but I'll do it that way.

First, though, what have we decided on the order? Do we need this PR?

martianboy:
Thanks for the heads up @tlively.

Sorry @kripken, you're right, it didn't have to be in the same PR but a misunderstanding of the spec made me do it in the same PR. Then, we decided to take an approach more or less similar to this PR.

Would you mind if I take a little time to review this PR? I fear it might be a bigger overlap.

kripken:
@martianboy Sounds good, yeah, take your time to read this PR and let me know what you think.

martianboy:
@kripken @tlively so this PR's approach is almost exactly the same as the other one, with a more elegant solution of course. I could use the same implementation in `table-utils.cpp` in my PR, as it was the only remaining comment so far. The rest of the changes here for writing, reading, parsing and printing such segments are identical in both. I'm also fine with merging this and resolving the conflicts later, shouldn't take more than half an hour.

tlively:
Thanks, @martianboy! Let's move forward this PR to get further GC testing unblocked, then. @kripken, besides that one comment, this LGTM.

src/passes/Print.cpp
kripken: Fixed, now this emits a single elem in the text format.

tlively (approved):

PR #3657, opened 2021-03-05
[effects] Record reads and writes of the GC heap

### LINKED ISSUE #3655: Invalid optimization, ignoring side effect on field value
With `-O1` or higher, Binaryen turns this:
```wat
(module
 (type $A (struct (field (mut i32)) (field (mut (ref null $B)))))
 (type $B (struct (field (mut i32)) (field (mut (ref null $C))) (field (mut (ref null $B)))))
 (type $C (struct (field (mut i32)) (field (mut i64))))
 (export "f" (func $f))
 (func $f (param $0 (ref null $A)) (result (ref null $C))
  (local $1 (ref null $C))
  (local.set $1
   (struct.get $B 1
    (struct.get $A 1
     (local.get $0)
    )
   )
  )
  (struct.set $A 1
   (local.get $0)
   (struct.get $B 2
    (struct.get $A 1
     (local.get $0)
    )
   )
  )
  (local.get $1)
 )
)
```
into this:
```wat
(module
 (type $B (struct (field (mut i32)) (field (mut (ref null $C))) (field (mut (ref null $B)))))
 (type $A (struct (field (mut i32)) (field (mut (ref null $B)))))
 (type $C (struct (field (mut i32)) (field (mut i64))))
 (type $ref?|$A|_=>_ref?|$C| (func (param (ref null $A)) (result (ref null $C))))
 (export "f" (func $f))
 (func $f (; has Stack IR ;) (param $0 (ref null $A)) (result (ref null $C))
  (struct.set $A 1
   (local.get $0)
   (struct.get $B 2
    (struct.get $A 1
     (local.get $0)
    )
   )
  )
  (struct.get $B 1
   (struct.get $A 1
    (local.get $0)
   )
  )
 )
)
```
Essentially, it is replacing
```
c = a.b.c
a.b = a.b.b
return c
```
by
```
a.b = a.b.b
return a.b.c
```

tlively:
Just as reads and writes to memory can interfere with each other, reads and
writes of GC objects can interfere with each other. This PR adds new `readsHeap`
and `writesHeap` fields to EffectAnalyzer to account for this interference. Note
that memory accesses can never alias with GC heap accesses, so they are
considered separately. Similarly, it would be possible to prove that different
GC heap accesses never interfere with each other based on the accessed types,
but that's left to future work.

Fixes #3655.


test/lit/gc-read-write-effects.wast
```diff
@@ -0,0 +1,73 @@
...
+   (local.get $0)
+   (struct.get $B 2
+    (struct.get $A 1
+     (local.get $0)
+    )
+   )
+  )
+  (local.get $1)
+ )
+)
```
kripken, line 73: This testcase can be simplified to
```wat
(module
 (type $A (struct (field (mut i32))))
 (func "test" (param $x (ref null $A)) (result i32)
  (local $y i32)
  (local.set $y
   (struct.get $A 0
    (local.get $x)
   )
  )
  (struct.set $A 0
   (local.get $x)
   (i32.const 10)
  )
  (local.get $y)
 )
)
```
And you just need to run `--simplify-locals` and not all of `-O1`.

kripken (commented):
Thanks!

Code lgtm, test lgtm with suggested simplifications.

test/lit/gc-read-write-effects.wast
```diff
@@ -0,0 +1,73 @@
...
+ ;; CHECK-NEXT:  (struct.set $A $b
+ ;; CHECK-NEXT:   (local.get $0)
+ ;; CHECK-NEXT:   (struct.get $B $b
+ ;; CHECK-NEXT:    (struct.get $A $b
+ ;; CHECK-NEXT:     (local.get $0)
+ ;; CHECK-NEXT:    )
+ ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT:  (local.get $1)
+ ;; CHECK-NEXT: )
```
kripken, line 53: side note, how do you write such an output? by hand, adding `;; CHECK-NEXT` etc., or with a tool?

tlively: For this file I just ran scripts/update_lit_checks.py! For other files where I am checking something besides function contents, I write them by hand.

tlively: Where "by hand" means liberal copy-paste and sometime emacs macros.

kripken: Got it, thanks!

kripken (approved):

PR #3658, opened 2021-03-05
[Wasm GC] Fix cost.h on array.new_default

kripken:
Trivial fix. Found by reducing a real-world codebase.

The test here is what actually crashed, a case of a pass that uses
cost analysis. With some extra work this could be a unit test perhaps
in C++. Would that be worth it?


tlively (approved):
Ehh, if our C++ unit tests had less overhead it might be worth it, but this seems fine to me.

PR #3659, opened 2021-03-05
[Wasm GC] Consider subtyping in least upper bounds

kripken:
Unless I am missing something, since we have structural subtyping
in wasm then when computing an LUB that must be taken into account.

Without this, the LUB of `{i32}, {i32, f64}` will be `dataref`. That breaks
on testcases like this when using SimplifyLocals:
```wat
(module
 (type $struct.A (struct (field i32)))
 (type $struct.B (struct (field i32) (field f64)))
 (func "test" (result (ref null $struct.A))
  (local $temp (ref null $struct.A))
  ;; this block will get a return value (replacing the sets to $temp). that
  ;; block's type must be valid. in the testcase here, B extends A, and so we
  ;; can use A, and do not need to go all the way up to the 'data' heap type.
  (block $return.label
   (if
    (i32.const 0)
    (block
     (local.set $temp
      (ref.null $struct.A)
     )
     (br $return.label)
    )
   )
   (local.set $temp
    (struct.new_default_with_rtt $struct.B
     (rtt.canon $struct.B)
    )
   )
  )
  (local.get $temp)
 )
)
```
Reduced from a real-world testcase.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -640,6 +640,12 @@ Type Type::getLeastUpperBound(Type a, Type b) {
   if (b == Type::unreachable) {
     return a;
   }
+  if (isSubType(a, b)) {
+    return b;
+  }
+  if (isSubType(b, a)) {
+    return a;
+  }
   if (a.size() != b.size()) {
     return Type::none; // a poison value that must not be consumed
   }
--- a/test/example/cpp-unit.cpp
+++ b/test/example/cpp-unit.cpp
@@ -569,11 +569,20 @@ void test_literals() {
   }
 }
 
+void test_LUB() {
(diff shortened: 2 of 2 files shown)


tlively:
This is not sufficient because it does not find the LUB of A and B when the LUB is not equal to A or B. We need code that specifically considers width and depth subtyping to calculate the LUB of structs. The code in this PR could potentially be the base case for avoiding infinite recursions on recursive types, but it would be more efficient to use a `seen` cache for detecting equality instead. I can work on this now if you'd like; just let me know.

kripken:
@tlively sounds like you know exactly what to do here (and that makes sense, although I didn't follow the cache part), so if you can do it that would be great!

PR #3660, opened 2021-03-05
Fixed .debug_loc parsing for wasm64 files

aardappel:
The address size was hard-coded to 4, it now gets this information from .debug_info.
This required changing the parsing order.
Also made failure to parse .debug_loc fail the program, as before this error was easy to ignore.

Note there is no test specific to the combination of dwarf + wasm64 in binaryen, as I am only just working on the tests in emscripten that uncovered this issue.

### DIFF
--- a/third_party/llvm-project/dwarf2yaml.cpp
+++ b/third_party/llvm-project/dwarf2yaml.cpp
@@ -116,7 +116,10 @@ void dumpDebugRanges(DWARFContext &DCtx, DWARFYAML::Data &Y) { // XXX BINARYEN
 }
 
 void dumpDebugLoc(DWARFContext &DCtx, DWARFYAML::Data &Y) { // XXX BINARYEN
-  uint8_t savedAddressByteSize = 4;
+  // This blindly grabs the first CU, which should be ok since they all have
+  // the same address size?
+  auto CU = DCtx.normal_units().begin()->get();
+  uint8_t savedAddressByteSize = CU->getFormParams().AddrSize;  // XXX BINARYEN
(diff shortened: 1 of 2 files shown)


third_party/llvm-project/dwarf2yaml.cpp
```diff
@@ -425,10 +428,12 @@ std::error_code dwarf2yaml(DWARFContext &DCtx, DWARFYAML::Data &Y) {
   dumpDebugAbbrev(DCtx, Y);
   dumpDebugStrings(DCtx, Y);
   dumpDebugARanges(DCtx, Y);
-  dumpDebugRanges(DCtx, Y); // XXX BINARYEN
-  dumpDebugLoc(DCtx, Y); // XXX BINARYEN
   dumpDebugPubSections(DCtx, Y);
   dumpDebugInfo(DCtx, Y);
   dumpDebugLines(DCtx, Y);
+  // These go last, because dumpDebugLoc relies on the address size being known
+  // from dumpDebugInfo.
```
kripken, line 435: the comment explains Loc, but why is Ranges also moved?

aardappel: I thought it made sense to keep em together, but I can move just loc..

aardappel:
Is this a known CI issue?
```
INFO: pip is looking at multiple versions of <Python from Requires-Python> to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement pyflakes<2.2.0,>=2.1.0 (from flake8)
ERROR: No matching distribution found for pyflakes<2.2.0,>=2.1.0
INFO: pip is looking at multiple versions of filecheck to determine which version is compatible with other requirements. This could take a while.
```

kripken (approved):

Issue #3656, opened 2021-03-05
Parse exception for valid heap type


manoskouk:
Hello,
I am trying to parse a .wasm file and I am getting the following message:
`[parse exception: invalid wasm heap type: -64]`
I believe this might be a large positive heap index which does not fit in one byte and is encoded as a LEB128 in multiple bytes, as required by spec: https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#heap-types-2.

kripken:
We use an S64LEB atm there,

https://github.com/WebAssembly/binaryen/blob/dfafe354a0ebb6250851399a36b8263dd3d43de3/src/wasm/wasm-binary.cpp#L1628-L1644

I think that is technically inaccurate as it should be S33, but there should be no noticeable difference aside from allowing higher (very unlikely) numbers. So this should be handling multi-byte indexes already, and I'm not sure what is wrong.

I also tested with a file with 5,000 heap types, using the below script, and I don't see an error. Binaryen can read the binaries it writes, and wasp can read them as well.

```py

print('(module')

TYPES = ['i32', 'i64', 'f32', 'f64', 'funcref', 'i31ref', 'externref', '(ref null 0)', '(ref null 1)']

i = 0

MAX = 5000

for a in TYPES:
  for b in TYPES:
    for c in TYPES:
      for d in TYPES:
        if i >= MAX:
          break
        fields = ['(field %s)' % t for t in (a, b, c, d)]
        print(' (type $struct%d (struct ' % i, ' '.join(fields), '))')
        i += 1

for a in range(i):
  print(' (func $func%d (param (ref null $struct%d)))' % (a, a))

print(')')
```

Is there a chance that wasm file is invalid?

manoskouk:
Looking up `getHeapType` I found this function:

https://github.com/WebAssembly/binaryen/blob/101aac786ccffe3714fdcf05437b2b740135b4de/src/wasm-builder.h#L914-L922

However, [rtts](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#reference-types-1) do not take heap types as inputs anymore, but rather type indices.

Maybe that could be at fault?

kripken:
Hmm, I'm not sure that is a problem. First, that code is for the IR itself, that's not binary reading or writing code. But, the binary reading code does write a heap type, that is true:

[...]

PR #3663, opened 2021-03-08
[Wasm GC] Properly handle "typeindex" in the binary format

kripken:
We handled them as S63 instead of U32. That should be fine, as all U32 values fit
in S63. But it is not strictly correct. The signed encoding may use an additional byte
which is unnecessary, and there is an actual correctness issue where a U32 may
be interpreted as a large negative S63 (because it sign extends a final bit that 
happens to be 1).

May fix https://github.com/WebAssembly/binaryen/issues/3656 but I  not have a testcase
yet on which to confirm (the error case above seems like the opposite of that issue).
cc @manoskouk

### DIFF
--- a/src/wasm-binary.h
+++ b/src/wasm-binary.h
@@ -1244,7 +1244,15 @@ public:
   Module* getModule() { return wasm; }
 
   void writeType(Type type);
+
+  // Gets an arbitrary heap type, which may be an indexed one or one of the
+  // basic types like funcref.
   void writeHeapType(HeapType type);
+  // Gets an indexed heap type. Note that this is encoded differently than a
+  // general heap type - a U32 LEB is used as negative values (for basic types)
+  // are not needed.
+  void writeIndexedHeapType(HeapType type);
+
   void writeField(const Field& field);
 
 private:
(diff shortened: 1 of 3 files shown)


manoskouk:
LGTM, thanks.

src/wasm-binary.h
```diff
@@ -1244,7 +1244,15 @@ class WasmBinaryWriter {
   Module* getModule() { return wasm; }
 
   void writeType(Type type);
+
+  // Gets an arbitrary heap type, which may be an indexed one or one of the
+  // basic types like funcref.
```
tlively, line 1249:
```suggestion
  // Writes an arbitrary heap type, which may be indexed or one of the
  // basic types like funcref.
```

src/wasm-binary.h
```diff
@@ -1244,7 +1244,15 @@ class WasmBinaryWriter {
...
   void writeHeapType(HeapType type);
+  // Gets an indexed heap type. Note that this is encoded differently than a
+  // general heap type - a U32 LEB is used as negative values (for basic types)
```
tlively, line 1252:
```suggestion
  // Writes an indexed heap type. Note that this is encoded differently than a
  // general heap type because it does not allow negative values for basic heap types.
```

PR #3666, opened 2021-03-09
[RT] Support expressions in element segments

martianboy:
This will add support for `ref.null t` as a valid element segment
item, by turning element segment data into a vector of `Expression*`
instead of just `Name`.

The abbreviated format of `(elem ... func $f $g...)` is kept in
both printing and binary emitting if all items are `ref.func`s. Public
APIs aren't updated in this PR.

### DIFF
--- a/src/ir/element-utils.h
+++ b/src/ir/element-utils.h
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2021 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef wasm_ir_element_h
+#define wasm_ir_element_h
+
+#include "wasm-builder.h"
+#include "wasm.h"
+
+namespace wasm {
+
+namespace ElementUtils {
+
+// iterate over functions referenced in an element segment
+template<typename T>
+inline void iterElementSegmentFunctionNames(ElementSegment* segment,
+                                            T visitor) {
+  // TODO(reference-types): return early if segment type is non-funcref
+  for (Index i = 0; i < segment->data.size(); i++) {
+    if (auto* get = segment->data[i]->dynCast<RefFunc>()) {
+      visitor(get->func, i);
+    }
+  }
+}
+
+// iterate over functions referenced in all element segments of a module
+template<typename T>
+inline void iterAllElementFunctionNames(const Module* wasm, T visitor) {
+  for (auto& segment : wasm->elementSegments) {
+    iterElementSegmentFunctionNames(
+      segment.get(), [&](Name& name, Index i) { visitor(name); });
+  }
+}
+
+} // namespace ElementUtils
+
+} // namespace wasm
+
+#endif // wasm_ir_element_h
--- a/src/ir/module-splitting.cpp
+++ b/src/ir/module-splitting.cpp
@@ -73,6 +73,7 @@
 //      complex code, so it is a good candidate for a follow up PR.
 
 #include "ir/module-splitting.h"
+#include "ir/element-utils.h"
 #include "ir/manipulation.h"
 #include "ir/module-utils.h"
 #include "ir/names.h"
@@ -95,10 +96,17 @@ template<class F> void forEachElement(Module& module, F f) {
     } else if (auto* g = segment->offset->dynCast<GlobalGet>()) {
       base = g->name;
     }
-    for (size_t i = 0; i < segment->data.size(); ++i) {
-      f(segment->table, base, offset + i, segment->data[i]);
-    }
+    ElementUtils::iterElementSegmentFunctionNames(
+      segment, [&](Name& entry, Index i) {
+        f(segment->table, base, offset + i, entry);
+      });
   });
+}
+
+static RefFunc* makeRefFunc(Module& wasm, Function* func) {
+  // FIXME: make the type NonNullable when we support it!
+  return Builder(wasm).makeRefFunc(func->name,
+                                   Type(HeapType(func->sig), Nullable));
 }
 
 struct TableSlotManager {
@@ -124,7 +132,7 @@ struct TableSlotManager {
   Table* makeTable();
 
   // Returns the table index for `func`, allocating a new index if necessary.
-  Slot getSlot(Name func);
+  Slot getSlot(RefFunc* entry);
   void addSlot(Name func, Slot slot);
 };
 
@@ -199,8 +207,8 @@ Table* TableSlotManager::makeTable() {
   return module.addTable(Builder::makeTable(Name::fromInt(0)));
 }
 
-TableSlotManager::Slot TableSlotManager::getSlot(Name func) {
-  auto slotIt = funcIndices.find(func);
+TableSlotManager::Slot TableSlotManager::getSlot(RefFunc* entry) {
+  auto slotIt = funcIndices.find(entry->func);
   if (slotIt != funcIndices.end()) {
     return slotIt->second;
   }
@@ -227,8 +235,10 @@ TableSlotManager::Slot TableSlotManager::getSlot(Name func) {
   Slot newSlot = {activeBase.tableName,
                   activeBase.global,
                   activeBase.index + Index(activeSegment->data.size())};
-  activeSegment->data.push_back(func);
-  addSlot(func, newSlot);
+
+  activeSegment->data.push_back(entry);
+
+  addSlot(entry->func, newSlot);
   if (activeTable->initial <= newSlot.index) {
     activeTable->initial = newSlot.index + 1;
   }
@@ -372,15 +382,15 @@ void ModuleSplitter::thunkExportedSecondaryFunctions() {
       // We've already created a thunk for this function
       continue;
     }
-    auto tableSlot = tableManager.getSlot(secondaryFunc);
     auto func = std::make_unique<Function>();
-
     func->name = secondaryFunc;
     func->sig = secondary.getFunction(secondaryFunc)->sig;
     std::vector<Expression*> args;
     for (size_t i = 0, size = func->sig.params.size(); i < size; ++i) {
       args.push_back(builder.makeLocalGet(i, func->sig.params[i]));
     }
+
+    auto tableSlot = tableManager.getSlot(makeRefFunc(primary, func.get()));
     func->body = builder.makeCallIndirect(
       tableSlot.tableName, tableSlot.makeExpr(primary), args, func->sig);
     primary.addFunction(std::move(func));
@@ -395,17 +405,21 @@ void ModuleSplitter::indirectCallsToSecondaryFunctions() {
     Builder builder;
     CallIndirector(ModuleSplitter& parent)
       : parent(parent), builder(parent.primary) {}
+    // Avoid visitRefFunc on element segment data
+    void walkElementSegment(ElementSegment* segment) {}
     void visitCall(Call* curr) {
       if (!parent.secondaryFuncs.count(curr->target)) {
         return;
       }
-      auto tableSlot = parent.tableManager.getSlot(curr->target);
-      replaceCurrent(builder.makeCallIndirect(
-        tableSlot.tableName,
-        tableSlot.makeExpr(parent.primary),
-        curr->operands,
-        parent.secondary.getFunction(curr->target)->sig,
-        curr->isReturn));
+      auto func = parent.secondary.getFunction(curr->target);
+      auto tableSlot =
+        parent.tableManager.getSlot(makeRefFunc(parent.primary, func));
+      replaceCurrent(
+        builder.makeCallIndirect(tableSlot.tableName,
+                                 tableSlot.makeExpr(parent.primary),
+                                 curr->operands,
+                                 func->sig,
+                                 curr->isReturn));
     }
     void visitRefFunc(RefFunc* curr) {
       assert(false && "TODO: handle ref.func as well");
@@ -454,14 +468,14 @@ void ModuleSplitter::setupTablePatching() {
     return;
   }
 
-  std::map<Index, Name> replacedElems;
+  std::map<Index, Function*> replacedElems;
   // Replace table references to secondary functions with an imported
   // placeholder that encodes the table index in its name:
   // `importNamespace`.`index`.
   forEachElement(primary, [&](Name, Name, Index index, Name& elem) {
     if (secondaryFuncs.count(elem)) {
-      replacedElems[index] = elem;
       auto* secondaryFunc = secondary.getFunction(elem);
+      replacedElems[index] = secondaryFunc;
       auto placeholder = std::make_unique<Function>();
       placeholder->module = config.placeholderNamespace;
       placeholder->base = std::to_string(index);
@@ -495,8 +509,8 @@ void ModuleSplitter::setupTablePatching() {
     // to be imported into the second module. TODO: use better strategies here,
     // such as using ref.func in the start function or standardizing addition in
     // initializer expressions.
-    const ElementSegment* primarySeg = tableManager.activeTableSegments.front();
-    std::vector<Name> secondaryElems;
+    ElementSegment* primarySeg = tableManager.activeTableSegments.front();
+    std::vector<Expression*> secondaryElems;
     secondaryElems.reserve(primarySeg->data.size());
 
     // Copy functions from the primary segment to the secondary segment,
@@ -507,33 +521,35 @@ void ModuleSplitter::setupTablePatching() {
          ++i) {
       if (replacement->first == i) {
         // primarySeg->data[i] is a placeholder, so use the secondary function.
-        secondaryElems.push_back(replacement->second);
+        secondaryElems.push_back(makeRefFunc(secondary, replacement->second));
         ++replacement;
-      } else {
-        exportImportFunction(primarySeg->data[i]);
-        secondaryElems.push_back(primarySeg->data[i]);
+      } else if (auto* get = primarySeg->data[i]->dynCast<RefFunc>()) {
+        exportImportFunction(get->func);
+        auto* copied =
+          ExpressionManipulator::copy(primarySeg->data[i], secondary);
+        secondaryElems.push_back(copied);
       }
     }
 
     auto offset = ExpressionManipulator::copy(primarySeg->offset, secondary);
-    auto secondaryElem = std::make_unique<ElementSegment>(
+    auto secondarySeg = std::make_unique<ElementSegment>(
       secondaryTable->name, offset, secondaryElems);
-    secondaryElem->setName(primarySeg->name, primarySeg->hasExplicitName);
-    secondary.addElementSegment(std::move(secondaryElem));
+    secondarySeg->setName(primarySeg->name, primarySeg->hasExplicitName);
+    secondary.addElementSegment(std::move(secondarySeg));
     return;
   }
 
   // Create active table segments in the secondary module to patch in the
   // original functions when it is instantiated.
   Index currBase = replacedElems.begin()->first;
-  std::vector<Name> currData;
+  std::vector<Expression*> currData;
   auto finishSegment = [&]() {
     auto* offset = Builder(secondary).makeConst(int32_t(currBase));
-    auto secondaryElem =
+    auto secondarySeg =
       std::make_unique<ElementSegment>(secondaryTable->name, offset, currData);
-    secondaryElem->setName(Name::fromInt(secondary.elementSegments.size()),
-                           false);
-    secondary.addElementSegment(std::move(secondaryElem));
+    secondarySeg->setName(Name::fromInt(secondary.elementSegments.size()),
+                          false);
+    secondary.addElementSegment(std::move(secondarySeg));
   };
   for (auto curr = replacedElems.begin(); curr != replacedElems.end(); ++curr) {
     if (curr->first != currBase + currData.size()) {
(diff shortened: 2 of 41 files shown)


martianboy:
@tlively reviews appreciated. :) This one should be relatively smaller in terms of significant changes across the code base.

tlively:
Sorry for the delay, @martianboy! I'm reviewing this now.

martianboy:
Yaay! Thanks @tlively. You may want to start with the core changes and also the `element-utils.h`, as much of the rest is just using the `ElementUtils::iter*` to get function names as before.

src/wasm/wasm-binary.cpp
```diff
@@ -2786,17 +2790,34 @@ void WasmBinaryBuilder::readElementSegments() {
...
 
-    size_t segmentIndex = functionTable.size();
-    auto& indexSegment = functionTable[segmentIndex];
+    auto& segmentData = elementSegments.back()->data;
     auto size = getU32LEB();
-    for (Index j = 0; j < size; j++) {
-      indexSegment.push_back(getU32LEB());
+    if (usesExpressions) {
+      for (Index j = 0; j < size; j++) {
+        segmentData.push_back(readExpression());
```
tlively, line 2810: Does this handle reading the `end`, too?

src/wasm/wasm-binary.cpp
```diff
@@ -2786,17 +2790,34 @@ void WasmBinaryBuilder::readElementSegments() {
...
+      }
+    } else {
+      for (Index j = 0; j < size; j++) {
+        Index index = getU32LEB();
+        auto sig = getSignatureByFunctionIndex(index);
+        auto* refFunc =
+          Builder(wasm).makeRefFunc("0", Type(HeapType(sig), Nullable));
```
tlively, line 2817: Is the "0" a placeholder name? If so, a comment would be useful.

src/wasm/wasm-s-parser.cpp
```diff
@@ -3291,11 +3291,17 @@ void SExpressionWasmBuilder::parseElem(Element& s, Table* table) {
...
+    auto makePassiveSegment = [&](bool usesExpressions) {
+      auto segment = std::make_unique<ElementSegment>();
+      segment->setName(name, hasExplicitName);
+      parseElemFinish(s, segment, i + 1, usesExpressions);
+    };
+    if (s[i]->str() == FUNC) {
+      return makePassiveSegment(false);
+    } else if (s[i]->str() == FUNCREF) {
+      return makePassiveSegment(true);
+    }
```
tlively, line 3304: It looks like this could be simplified by just calculating a bool `usesExpressions` first, then running what is now the body of the lambda.

src/wasm/wasm-s-parser.cpp
```diff
@@ -3345,13 +3352,40 @@ void SExpressionWasmBuilder::parseElem(Element& s, Table* table) {
...
+      auto& inner = *s[i];
+      if (elementStartsWith(inner, ITEM)) {
+        if (inner[1]->isList()) {
+          // (item (ref.func $f))
+          segment->data.push_back(parseExpression(inner[1]));
+        } else {
+          // (item ref.func $f)
+          inner.list().removeAt(0);
+          segment->data.push_back(parseExpression(inner));
+        }
```
tlively, line 3378: Are these really both allowed by the spec? That seems unnecessarily complicated :/

tlively (commented):
Looking good so far!

src/wasm/wasm-s-parser.cpp
martianboy: I totally agree, it doesn't make any sense. The whole `(item ...)` seems unnecessary. But as for this particular form, I've seen it only once in the spec tests here: https://github.com/WebAssembly/spec/blob/master/test/core/elem.wast#L11

I'm not even sure if it's really intended like that or it's just a typo!

src/wasm-interpreter.h
```diff
@@ -2309,21 +2309,27 @@ template<typename GlobalManager, typename SubType> class ModuleInstanceBase {
...
+        if (auto* get = segment->data[i]->dynCast<RefFunc>()) {
+          externalInterface->tableStore(
+            segment->table,
+            offset + i,
+            Literal::makeFunc(get->func, get->type));
+        } else if (segment->data[i]->is<RefNull>()) {
+          externalInterface->tableStore(
+            segment->table,
+            offset + i,
+            Literal::makeNull(segment->data[i]->type));
```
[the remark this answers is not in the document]
martianboy: It'd be much easier to do that once we have `table.init` instructions, in which case it could mirror `initializeMemoryContents`. I think I tried using `RuntimeExpressionRunner` for this first, but it made `ASan` unhappy. I'll give it another shot!

src/tools/wasm-reduce.cpp
```diff
@@ -789,46 +812,9 @@ struct Reducer
...
     std::cerr << "|    try to simplify elem segments\n";
-    Name first;
-    auto it =
-      std::find_if_not(module->elementSegments.begin(),
-                       module->elementSegments.end(),
-                       [&](auto& segment) { return segment->data.empty(); });
-
-    if (it != module->elementSegments.end()) {
-      first = it->get()->data[0];
-    }
```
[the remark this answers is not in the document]
martianboy: No, it was used to find the first function name used in a segment to use it as `zero`. Here I'm using `ref.null` as `zero` so this is no longer necessary.

src/wasm/wasm-s-parser.cpp
martianboy: Done. Although it required an additional `isPassive` variable as well. I added a comment for the `else` case.

src/wasm/wasm-binary.cpp
martianboy: Correct, it's a placeholder. I agree the `"0"` was misleading; replaced it with `Name::fromInt(index)`.

src/wasm/wasm-binary.cpp
martianboy: I believe so, yes. It calls `processExpressions()` which iteratively reads all expressions until `end` or `unreachable`.

src/wasm-interpreter.h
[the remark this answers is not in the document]
martianboy: Done. Seems to be working this time. :)

src/tools/wasm-reduce.cpp
```diff
@@ -789,46 +812,9 @@ struct Reducer
     return shrank;
   }
 
-  template<typename T, typename U>
-  void reduceByZeroing(T* segment, U zero, size_t bonus, bool shrank) {
```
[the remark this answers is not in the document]
martianboy: Done.

src/tools/wasm-reduce.cpp
```diff
@@ -752,7 +752,30 @@ struct Reducer
     }
     // the "opposite" of shrinking: copy a 'zero' element
     for (auto& segment : curr->segments) {
-      reduceByZeroing(&segment, 0, 2, shrank);
+      if (segment.data.empty()) {
+        return;
+      }
```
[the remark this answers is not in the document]
martianboy: This was actually wrong! It's now removed. Thanks.

src/tools/wasm-metadce.cpp
```diff
@@ -216,11 +216,13 @@ struct MetaDCEGraph {
     ModuleUtils::iterActiveElementSegments(wasm, [&](ElementSegment* segment) {
       // TODO: currently, all functions in the table are roots, but we
       //       should add an option to refine that
-      for (auto& name : segment->data) {
-        if (!wasm.getFunction(name)->imported()) {
-          roots.insert(functionToDCENode[name]);
-        } else {
-          roots.insert(importIdToDCENode[getFunctionImportId(name)]);
+      for (auto* curr : segment->data) {
+        if (auto* get = curr->dynCast<RefFunc>()) {
```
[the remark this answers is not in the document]
martianboy: Done.

test/multi-table.wast
```diff
@@ -9,7 +10,10 @@
   (elem (table $t2) (i32.const 0) func $f)
   (elem $activeNonZeroOffset (table $t2) (offset (i32.const 1)) func $f $g)
 
-  (elem $passive func $f $g)
+  (elem $e3-1 (table $t3) (i32.const 0) funcref (ref.func $f))
+  (elem $e3-2 (table $t3) (offset (i32.const 1)) funcref (item ref.func $f) (item (ref.func $g)))
```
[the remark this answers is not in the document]
martianboy: Yes it is. There was a `ref.null` on line 16. I added another one to `$e3-1` to test active segments too.

src/passes/GenerateDynCalls.cpp
```diff
@@ -57,8 +57,9 @@ struct GenerateDynCalls : public WalkerPass<PostWalker<GenerateDynCalls>> {
                            });
     if (it != segments.end()) {
       std::vector<Name> tableSegmentData;
-      for (const auto& indirectFunc : it->get()->data) {
-        generateDynCallThunk(wasm->getFunction(indirectFunc)->sig);
+      for (auto* entry : it->get()->data) {
```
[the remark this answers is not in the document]
martianboy: Done.

src/ir/table-utils.cpp
```diff
@@ -32,9 +33,8 @@ std::set<Name> getFunctionsNeedingElemDeclare(Module& wasm) {
 
   std::unordered_set<Name> tableNames;
   for (auto& segment : wasm.elementSegments) {
-    for (auto name : segment->data) {
-      tableNames.insert(name);
-    }
+    ElementUtils::iterElementSegmentFunctionNames(
+      segment.get(), [&](Name name, Index) { tableNames.insert(name); });
```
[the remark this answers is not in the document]
martianboy: Yes, done.

src/ir/element-utils.h
```diff
@@ -0,0 +1,53 @@
...
+namespace wasm {
+
+namespace ElementUtils {
+
+// iterate over functions referenced in an element segment
+template<typename T>
+inline void iterElementSegmentFunctionNames(ElementSegment* segment,
+                                            T visitor) {
+  // TODO(reference-types): return early if segment type is non-funcref
+
```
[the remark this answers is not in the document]
martianboy: Done.

martianboy:
Thanks for the comments, @tlively! I think I've addressed all of them.

src/wasm/wasm-s-parser.cpp
tlively: Looks like it's intentional: https://webassembly.github.io/reference-types/core/text/modules.html#text-elem. Probably better to just implement it like you've done here than to try to change it in the spec or to just not support it at all in Binaryen.

martianboy: Yeah, at least it doesn't come with a high implementation/complexity cost.

src/tools/fuzzing.h
```diff
@@ -725,7 +725,8 @@ class TranslateToFuzzReader {
     while (oneIn(3) && !finishedInput) {
       auto& randomElem =
         wasm.elementSegments[upTo(wasm.elementSegments.size())];
-      randomElem->data.push_back(func->name);
+      auto type = Type(HeapType(func->sig), Nullable);
```
kripken, line 728: FIXME here too,

src/tools/wasm-reduce.cpp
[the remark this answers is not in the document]
kripken: will that fail to validate if ReferenceTypes is disabled?

kripken (commented):
Very nice work!

Please fuzz this locally if you haven't already.

src/wasm/wasm-validator.cpp
```diff
@@ -2898,9 +2913,18 @@ static void validateTables(Module& module, ValidationInfo& info) {
...
+                    [](Expression* expr) {
+                      return expr->is<RefFunc>() ||
+                             (expr->is<RefNull>() &&
+                              Type::isSubType(expr->type, Type::funcref));
+                    }),
+        "elem",
+        "element segment items must be either ref.func or ref.null func");
     }
   }
 }
```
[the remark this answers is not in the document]
martianboy: Thanks for this, I was under the impression that the module will be walked during validation so expressions will be visited automatically. Now, I tried to use `FunctionValidator`, but looks like `ref.*` validation rules refuse to validate the module if reference-types is not enabled. I can validate existence of functions separately and not use `FunctionValidator` for now, until we come up with a better a way maybe?

src/ir/table-utils.h
```diff
@@ -84,7 +84,13 @@ inline Index append(Table& table, Name name, Module& wasm) {
     }
     wasm.dylinkSection->tableSize++;
   }
-  segment->data.push_back(name);
+
+  auto* func = wasm.getFunctionOrNull(name);
+  if (func == nullptr) {
+    Fatal() << "Cannot append non-existing function to a table.";
```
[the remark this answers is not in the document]
martianboy: I needed to safely get the function signature, otherwise it would crash in case of an undefined function before reaching the validator. Maybe there's a better way to do that?

src/tools/wasm-reduce.cpp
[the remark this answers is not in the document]
martianboy: Good point!! Thanks @kripken, so I guess we need this after all.

src/tools/fuzzing.h
martianboy: Done.

src/ir/module-splitting.cpp
```diff
@@ -95,12 +96,18 @@ template<class F> void forEachElement(Module& module, F f) {
...
+    ElementUtils::iterElementSegmentFunctionNames(
+      segment, [&](Name& entry, Index i) {
+        f(segment->table, base, offset + i, entry);
+      });
   });
 }
 
+inline RefFunc* makeRefFunc(Module& wasm, Function* func) {
+  return Builder(wasm).makeRefFunc(func->name,
+                                   Type(HeapType(func->sig), Nullable));
```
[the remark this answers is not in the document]
martianboy: Done.

src/ir/module-splitting.cpp
```diff
@@ -95,12 +96,18 @@ template<class F> void forEachElement(Module& module, F f) {
...
-      f(segment->table, base, offset + i, segment->data[i]);
-    }
+    ElementUtils::iterElementSegmentFunctionNames(
+      segment, [&](Name& entry, Index i) {
+        f(segment->table, base, offset + i, entry);
+      });
   });
 }
 
+inline RefFunc* makeRefFunc(Module& wasm, Function* func) {
```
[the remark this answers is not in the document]
martianboy: Done.

src/wasm/wasm-validator.cpp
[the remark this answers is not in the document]
martianboy: So I extracted the main validation logic from `visitRefFunc` and `visitRefNull` into `validateRefFunc` and `validateRefNull` to avoid duplication, and used those in `validateTables`. It works, but looks hacky.

src/tools/wasm-reduce.cpp
```diff
@@ -842,16 +838,36 @@ struct Reducer
...
+            return true;
+          } else if (first->is<RefNull>()) {
+            return false;
+          } else {
+            // Both are ref.func
+            auto* f = first->cast<RefFunc>();
+            auto* e = entry->cast<RefFunc>();
+            return f->func == e->func && f->type == e->type;
+          }
+        },
```
martianboy, line 868: @kripken Please take a look at this. I restored the logic for finding a `zero` as it was needed for non-reftypes modules, but the part where it needs to check if an element is equal to the `zero` has become bigger than it would seem necessary. Just want to make sure it's not doing unnecessary work.

martianboy:
@kripken Thanks for the review and comments! I fuzzed this using `./scripts/fuzz_opt.py`, it runs a few hundred iterations. I once saw an issue related to conflicting element segment names which was resolved in `main` so I rebased my branch. Another issue I just saw was with seed `13931693928481941102` exists on `main` as well. It's related to types.

A subsequent PR can add support for `ref.null` elements to the fuzzer.

test/spec/call_indirect_refnull.wast
```diff
@@ -0,0 +1,12 @@
...
+  (elem (table $t) (i32.const 0) funcref (ref.null func))
+
+  (func $call-refnull (export "call-refnull") (result f32)
+    (call_indirect (result f32) (i32.const 0))
+  )
+)
+(assert_trap
+  (invoke "call-refnull")
+  "uninitialized table element"
+)
```
martianboy, line 12: @kripken @tlively I added this to test `call_indirect` on `ref.null`. Is this a good way to do it, or should I go with another method?

src/ir/table-utils.h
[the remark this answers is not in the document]
kripken: I think the code is correct. Yes, you need the function to exist to know the signature. But the `Fatal()` should be an assert.

src/ir/table-utils.h
```diff
@@ -84,7 +84,14 @@ inline Index append(Table& table, Name name, Module& wasm) {
...
+  }
```
kripken, line 91:
```suggestion
  assert(func != nullptr && "Cannot append non-existing function to a table.");
```

src/wasm/wasm-validator.cpp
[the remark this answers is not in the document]
kripken: > looks like ref.* validation rules refuse to validate the module if reference-types is not enabled.

That is the bigger issue here, yes. This PR uses a reference type thing even in cases where reference types are not enabled. There is one example of that so far, using more specific types for `ref.func` even if typed function references are not enabled, but that one was carefully seen to be safe. This one looks like it would take a little more work.

But maybe not much more. Can we just check "if !getFunction then it's ok to use refFunc`? That is, allow it in global locations, but not in function bodies.

test/spec/call_indirect_refnull.wast
kripken: this test looks good to me.

src/tools/wasm-reduce.cpp
```diff
@@ -842,16 +838,36 @@ struct Reducer
...
     // try to reduce to first function. first, shrink segment elements.
     // while we are shrinking successfully, keep going exponentially.
     bool shrank = false;
     for (auto& segment : module->elementSegments) {
       shrank = shrinkByReduction(segment.get(), 100);
     }
     // the "opposite" of shrinking: copy a 'zero' element
+    if (first == nullptr) {
+      return;
+    }
```
kripken, line 850: this if could be on line 840

but i don't follow this, lower down it checks if `!first` which is no longer possible?

martianboy: The Alpine build was failing on this even though I had this check before entering the loop. (I suspect it had something to do with `first` being used in a lambda.)

src/wasm/wasm-validator.cpp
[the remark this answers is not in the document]
martianboy: That's what I was looking for! Thanks a lot for the direction. :) It's fixed now.

src/tools/wasm-reduce.cpp
```diff
@@ -851,7 +851,24 @@ struct Reducer
...
     for (auto& segment : module->elementSegments) {
-      reduceByZeroing(segment.get(), first, 100, shrank);
+      reduceByZeroing(
+        segment.get(),
+        first,
+        [&](Expression* entry) {
+          if (entry->is<RefNull>()) {
+            // we don't need to replace a ref.null
+            return true;
+          } else if (!first || first->is<RefNull>()) {
```
kripken, line 861: `first` cannot be null because of the check earlier.
```suggestion
          } else if (first->is<RefNull>()) {
```

src/tools/wasm-reduce.cpp
```diff
@@ -851,7 +851,24 @@ struct Reducer
...
+            return false;
+          } else {
+            // Both are ref.func
+            auto* f = first->cast<RefFunc>();
+            auto* e = entry->cast<RefFunc>();
+            return f->func == e->func && f->type == e->type;
```
kripken, line 867:
```suggestion
            return f->func == e->func;
```
The function ID is unique already, no need to check the type.

src/tools/wasm-reduce.cpp
kripken: I think it looks good with the suggested changes.

src/wasm/wasm-validator.cpp
```diff
@@ -1960,7 +1960,7 @@ void FunctionValidator::visitMemoryGrow(MemoryGrow* curr) {
...
 
 void FunctionValidator::visitRefNull(RefNull* curr) {
-  shouldBeTrue(getModule()->features.hasReferenceTypes(),
+  shouldBeTrue(!getFunction() || getModule()->features.hasReferenceTypes(),
```
kripken, line 1963:
```suggestion
  // If we are not in a function, this is a global location like a table. We allow
  // RefNull there as we represent tables that way regardless of what features
  // are enabled.
  shouldBeTrue(!getFunction() || getModule()->features.hasReferenceTypes(),
```
and similar down below

kripken (commented):
Ok, I think with these last comments this will look good from my perspective!

src/tools/wasm-reduce.cpp
martianboy: Good point, thanks! Updated.

src/wasm/wasm-validator.cpp
martianboy: Done.

src/tools/wasm-reduce.cpp
martianboy: Done. Alpine didn't complain this time. Looks good now.

kripken (approved):
lgtm!

For fuzzing, sadly it is a little hard on current `main` due to LUB fallout. But you can do some fuzzing and ignore anything with `Could not ensure supertype` in the error.

@tlively did you want to take another look?

tlively (approved):
Sorry for dropping the ball on this review, and thanks @kripken for jumping in :) I took another brief look and didn't see anything else to comment on, so lgtm as well.

tlively:
I'm doing a bit of fuzzing of this locally and I'll land it if nothing comes up :)

martianboy:
@tlively Thanks! I did fuzz this myself a few times, nothing came up really other than the LUB bug. Unfortunately it errors out within less than 50 iterations most of the time, although it once went on to 200 iterations.

PR #3667, opened 2021-03-09
[Wasm GC] Avoid extremely large outputs in TypeNamePrinter

kripken:
Also limit it to reasonable amounts of recursion.

### DIFF
--- a/src/passes/Print.cpp
+++ b/src/passes/Print.cpp
@@ -97,8 +97,9 @@ struct TypeNamePrinter {
   void print(const Array& array);
   void print(const Rtt& rtt);
 
-  // FIXME: This limit avoids infinite recursion and ensures reasonable-sized
-  // output, but we should have a better mechanism for this.
+  // FIXME: This hard limit on how many times we call print() avoids extremely
+  //        large outputs, which can be inconveniently large in some cases, but
+  //        we should have a better mechanism for this.
   static const size_t MaxPrints = 10;
 
   size_t prints = 0;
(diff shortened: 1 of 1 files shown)


src/passes/Print.cpp
```diff
@@ -96,9 +96,27 @@ struct TypeNamePrinter {
   void print(const Struct& struct_);
   void print(const Array& array);
   void print(const Rtt& rtt);
+
+  // FIXME: This limit avoids infinite recursion and ensures reasonable-sized
+  // output, but we should have a better mechanism for this.
```
tlively, line 101: We already have a mechanism for avoiding infinite recursions. If you want to avoid introducting a new `prints` variable here, you could reuse the existing `currHeapTypeDepth`, but perhaps what you have is better because it also avoids printing very wide structs fully.

```suggestion
  // FIXME: This limit ensures reasonable-sized
  // output, but we should have a better mechanism for this.
```

tlively (approved):

tlively:
Definitely change the commit name, though, since this doesn't affect infinite recursions.

src/passes/Print.cpp
kripken: Ok, it is indeed not infinitely recursing... just 1,336,335 times (!)

tlively: 🤯

kripken:
Alpine build seems stuck but I doubt it's this PR's fault, landing.

PR #3670, opened 2021-03-10
Use stdbool boolean instead of int to represent boolean values

### LINKED ISSUE #3664: Use `bool` in the C API
Assuming the C API supported C99 and above can/should we just use `bool` from `stdbool.h`?

See https://en.cppreference.com/w/c/types/boolean

pmatos:
Fixes #3664

### DIFF
--- a/test/example/c-api-kitchen-sink.c
+++ b/test/example/c-api-kitchen-sink.c
@@ -870,7 +870,7 @@ void test_core() {
   // Memory. One per module
 
   const char* segments[] = { "hello, world", "I am passive" };
-  int8_t segmentPassive[] = { 0, 1 };
+  bool segmentPassive[] = { false, true };
   BinaryenExpressionRef segmentOffsets[] = { BinaryenConst(module, BinaryenLiteralInt32(10)), NULL };
   BinaryenIndex segmentSizes[] = { 12, 12 };
   BinaryenSetMemory(module, 1, 256, "mem", segments, segmentPassive, segmentOffsets, segmentSizes, 2, 1);
@@ -1367,7 +1367,7 @@ void test_for_each() {
 
     const char* segments[] = { "hello, world", "segment data 2" };
     const uint32_t expected_offsets[] = { 10, 125 };
-    int8_t segmentPassive[] = { 0, 0 };
+    bool segmentPassive[] = { false, false };
     BinaryenIndex segmentSizes[] = { 12, 14 };
 
     BinaryenExpressionRef segmentOffsets[] = {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ Current Trunk
   `BinaryenElementSegmentGetLength`.
 - `BinaryenGetFunctionTableSegmentData` is replaced with
   `BinaryenElementSegmentGetData`.
+- Boolean values in the C API now should use `bool` instead of `int`.
 
 v100
 ----
--- a/src/binaryen-c.cpp
+++ b/src/binaryen-c.cpp
@@ -902,7 +902,7 @@ BinaryenExpressionRef BinaryenGlobalSet(BinaryenModuleRef module,
 }
 BinaryenExpressionRef BinaryenLoad(BinaryenModuleRef module,
                                    uint32_t bytes,
-                                   int8_t signed_,
+                                   bool signed_,
                                    uint32_t offset,
                                    uint32_t align,
                                    BinaryenType type,
@@ -1275,7 +1275,7 @@ BinaryenExpressionRef BinaryenI31New(BinaryenModuleRef module,
 
 BinaryenExpressionRef BinaryenI31Get(BinaryenModuleRef module,
                                      BinaryenExpressionRef i31,
-                                     int signed_) {
+                                     bool signed_) {
   return static_cast<Expression*>(
     Builder(*(Module*)module).makeI31Get((Expression*)i31, signed_ != 0));
 }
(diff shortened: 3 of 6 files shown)


test/example/c-api-kitchen-sink.c
```diff
@@ -870,7 +870,7 @@ void test_core() {
   // Memory. One per module
 
   const char* segments[] = { "hello, world", "I am passive" };
-  int8_t segmentPassive[] = { 0, 1 };
+  bool segmentPassive[] = { false, true };
```
kripken, line 873: are these changes necessary - do compilers not automatically convert these silently?

if so then it's a breaking change and so I think it's worth mentioning in the changelog.

kripken (commented):
Thanks! Lgtm aside from the question.

test/example/c-api-kitchen-sink.c
pmatos: They silently convert variables but not arrays, which is why this was failing: https://github.com/WebAssembly/binaryen/runs/2076841915?check_suite_focus=true#step:17:18642

Would a sentence like "Boolean values in the C API now should use `bool` instead of `int`." be enough, or do you want to list all the affected functions?

kripken: I think a short sentence is good enough, no need for a full list.

But this does make me worry. If arrays are common usage, then this could be annoying for users. I suspect it would be fine as most users actually use some kind of bindings to another language anyhow (not many write compilers in C). Curious to hear your thoughts and others'.

sbc100: I don't have see if errors or warnings when I do `bool foo[] = {0, 0, 1, 1};`... at last not from `clang -Wall`

pmatos: Actually if this is compiled in C, there's no problem. The problem occurs when you compile it in C++. You get a narrowing conversion warning like:
> warning: narrowing conversion of ‘x’ from ‘int8_t’ {aka ‘signed char’} to ‘bool’ [-Wnarrowing]

In any case, I don't totally understand why this could be annoying. You mean because people are used to writing 0 and 1 instead of false and true? Or because they have to type more characters? In any case, I don't personally find it annoying. Also as you say, most will use bindings but even if they didn't it is fine for me. Although maybe there are others who are more picky with this sort of thing. Lets wait for more opinions then.

sbc100: I don't see that warning with clang of gcc.. maybe a windows thing?

```
$ cat test.cpp 
#include <stdbool.h>

bool foo[] = {0, 0, 1, 1};
$ g++ -Wall test.cpp -c
$ clang++ -Wall test.cpp -c
```

sbc100: Regardless, seems like a pretty minimal breakage that would be easy for folks to fix (Also, its only a warning not an error)

tlively:
This strongly lgtm fwiw. `bool` makes much more sense than any kind of integer for these uses and this change does not seem more disruptive than other API changes we've been making recently (e.g. changing how tables and segments work).

test/example/c-api-kitchen-sink.c
pmatos: @sbc100 So, assigning `bool foo[] = {0, 0, 1, 1};` is fine. The conversion is done per value, which is ok since you can do `bool x = 1;`. The problem is for example with this:
```
bool foo(uint8_t x) {
    bool arr[] = { x, x};
    return arr[0];
}
```

pmatos: I sort of wandered away from the initial problem. I am sorry for this. The initial problem is not even C++ related, because the API tests are compiled with gcc. The initial problem is this: I changed the header file to accept `bool *` for passive segments field. But the test is calling it with an `int8_t *`. So the following example comes close to understanding the compilation issue:
```
#include <stdbool.h>
#include <stdint.h>

void FunctionExpectsBoolArray(bool*);

void foo(void) {
    int8_t arr[] = { 0, 1 };
    FunctionExpectsBoolArray(arr);
}
```

If you try with `gcc`:
```
gcc -Wall -c ~/tmp/bool4.c 
/home/pmatos/tmp/bool4.c: In function ‘foo’:
/home/pmatos/tmp/bool4.c:8:30: warning: passing argument 1 of ‘FunctionExpectsBoolArray’ from incompatible pointer type [-Wincompatible-pointer-types]
    8 |     FunctionExpectsBoolArray(arr);
      |                              ^~~
      |                              |
      |                              int8_t * {aka signed char *}
In file included from /home/pmatos/tmp/bool4.c:1:
/home/pmatos/tmp/bool4.c:4:31: note: expected ‘_Bool *’ but argument is of type ‘int8_t *’ {aka ‘signed char *’}
    4 | void FunctionExpectsBoolArray(bool*);
      |                               ^
```

Replying to @sbc100 , if you change the declaration to `bool arr[] = { 0, 1 };` it works, as you rightly pointed out.

pmatos:
@kripken I have added a line to the changelog fwiw.

kripken (approved):
I agree, this is worth doing even if there is a small chance of it being noticed by a user. The changlog mention mitigates that so lgtm!

pmatos:
Thanks @kripken , I guess we are just missing a lgtm comment from @sbc100 . :)

sbc100 (approved):
lgtm!  (I general, no need to wait for more than one lg)

pmatos:
Thanks @sbc100 . 
I don't have permissions to merge. Can someone please do it? Thanks.

PR #3672, opened 2021-03-10
[Wasm GC] Fix RTT type parsing

kripken:
RTT types have type indexes, not heap types (there is no `(rtt funcref)`).

Noticed by @manoskouk - thanks!

Fixes https://github.com/WebAssembly/binaryen/issues/3656 after which we can parse
that testcase - but not validate it yet, there is another issue remaining with typing.

### DIFF
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -1648,11 +1648,11 @@ Type WasmBinaryBuilder::getType(int initial) {
       return Type(getHeapType(), Nullable);
     case BinaryConsts::EncodedType::rtt_n: {
       auto depth = getU32LEB();
-      auto heapType = getHeapType();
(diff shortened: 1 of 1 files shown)


aheejin (approved):
If the spec says so this looks right, but I'm wondering why the spec prevents creating an RTT from basic heap types..?

kripken:
Good question, I have not followed the spec side closely enough. I'd guess the assumption is that dynamically casting to e.g. `funcref` is not needed as it can be cast to statically instead (this does seem to rule out a dynamic cast from `externref` to `funcref`, however...). Probably @tlively knows?

tlively:
You can cast from externref to funcref by using the `ref.as_func` instruction, since it accepts any anyref value, including any `externref` value. `ref.as_func` and the similar casting instructions make it so you would never need to use an rtt to cast to any of the basic heap types.

PR #3673, opened 2021-03-10
Make unreachable a subtype of everything

tlively:
Since in principle an unreachable expression can be used in any position. An
exception to this rule is in OptimizeInstructions, which avoids replacing
concrete expressions with unreachable expressions so that it doesn't need to
refinalize any expressions. Notably, Type::getLeastUpperBound was already
treating unreachable as the bottom type.

### DIFF
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -863,13 +863,14 @@ struct OptimizeInstructions
           // sides are identical, fold
           // if we can replace the if with one arm, and no side effects in the
           // condition, do that
-          auto needCondition = effects(iff->condition).hasSideEffects();
-          auto isSubType = Type::isSubType(iff->ifTrue->type, iff->type);
-          if (isSubType && !needCondition) {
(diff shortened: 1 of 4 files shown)


src/passes/OptimizeInstructions.cpp
```diff
@@ -863,13 +863,14 @@ struct OptimizeInstructions
...
+            iff->type.isConcrete() && iff->ifTrue->type == Type::unreachable;
+          if (!wouldBecomeUnreachable && !needCondition) {
             return iff->ifTrue;
           } else {
             Builder builder(*getModule());
-            if (isSubType) {
+            if (!wouldBecomeUnreachable) {
               return builder.makeSequence(builder.makeDrop(iff->condition),
                                           iff->ifTrue);
             } else {
```
kripken, line 876: the comment on the line below, could be made more specific, to replace "not" with "unreachable".

or the comment can be removed - the assertion says it all.

(I can't add a github suggestion because it's out of the range of the diff...)

kripken (approved):
Nice!

PR #3674, opened 2021-03-10
Fix fuzz issue with ElementSegment name collisions

kripken:
Now that they have names they can collide. All the fuzzer wants is
to ensure there is a segment to add to, so if there is one, do not
add another.


src/tools/fuzzing.h
```diff
@@ -425,15 +425,19 @@ class TranslateToFuzzReader {
...
       table->initial = table->max = 0;
 
       auto segment = std::make_unique<ElementSegment>(
         table->name, builder.makeConst(int32_t(0)));
       segment->setName(Name::fromInt(0), false);
       wasm.addElementSegment(std::move(segment));
-    } else {
+      return;
+    }
+    if (wasm.tables.empty()) {
```
tlively, line 440: In principle the segment names could still clash because there might be passive segments but no table (unless that's a validation error? I'm not sure...).

tlively (commented):
What are we even using the segments for here? Is it just to make sure that indirect calls work? If so, why does the table have size 0? Do we depend on there being a segment "0" because we reference it later? If so, we should ensure that there is a properly named segment if there are pre-existing segments.

kripken:
Yes, I think this is all just to add some of the functions to a table segment, so that call_indirect has something to call and we can test it.

The table starts at size 0 and then in `finalizeTable` we increase that.

Meanwhile I see we add to a random elementSegment, so there is nothing special about 0 I guess.

src/tools/fuzzing.h
kripken: Oh, good point, I'm not sure... for now though this seems to not be an issue in the fuzzer, it ran for quite some time with this.

martianboy: We don't have a validation error in this case right now. So the name collision is still a possibility with pre-existing passive segments.

kripken: Ok, thanks, updated to more carefully add a name, and also add the table only when needed

src/tools/fuzzing.h
```diff
@@ -427,25 +427,17 @@ class TranslateToFuzzReader {
...
       auto table = builder.makeTable(
         Names::getValidTableName(wasm, "fuzzing_table"), 0, 0);
       table->hasExplicitName = true;
+      wasm.addTable(std::move(table));
+    }
+    if (wasm.elementSegments.empty()) {
       auto segment = std::make_unique<ElementSegment>(
-        table->name, builder.makeConst(int32_t(0)));
-      segment->setName(Name::fromInt(0), false);
+        wasm.tables[0]->name, builder.makeConst(int32_t(0)));
```
tlively, line 438: Might be worth a TODO here about picking random tables eventually.

tlively (approved):
Looks good, thanks!

PR #3675, opened 2021-03-10
Fix fuzzer to generate fresh ElementSegment names

tlively:


kripken (commented):
This races with #3674 it seems...

This PR always adds a segment; the other adds one only if needed. Regardless they append names to segment 0 - so this appends to a pre-existing segment 0 if there was one. A risk with this PR is that appending to segment 0 may cause a validation error if there is a segment 1 at a fixed address that ends up overlapping (is that a validation error? if not, it could be confusing)

tlively:
Oops, sorry for the race. I'll close this one. FWIW, overlapping segments is not a validation error.

PR #3676, opened 2021-03-11
Revert "Make unreachable a subtype of everything"

tlively:
Reverts WebAssembly/binaryen#3673, which accidentally weakened validation by allowing unreachable nodes in places they were previously disallowed.


tlively:
Reverts WebAssembly/binaryen#3673, which accidentally weakened validation by allowing unreachable nodes in places they were previously disallowed.

kripken (approved):
Revert sgtm if there is a problem.

What specifically was the regression?

tlively:
I thought that #3673 had introduced a bug exposed by #3669, but actually this was just a bug in my local work on #3669.

PR #3677, opened 2021-03-11
[Wasm GC] Add more passing spec tests.

kripken:
The remaining tests use stacky IR or things like `table.set` which we don't
support yet.


martianboy:
@kripken I was wondering whether we have a process for adding/updating spec tests, or we just copy & paste the tests from the proposal repository whenever support for a feature is added?

kripken:
@martianboy We don't have a very good process for this, sadly. Binaryen's text format is not compatible with the spec's and so we can't simply copy them over. So we manually add tests when they make sense, and sometimes need to modify them a little.

Overall this isn't high priority as we can get very good testing in other ways (in GC I've been focused on real-world code and fuzzing) but it's useful when it's not hard to do.

In the long term there are some ideas about getting Binaryen full support for the spec text format, or for using the binary format to work around that (i.e., run spec tests by converting them to binary using another tool). If one of those happens we can improve this and hopefully get to the point of just pulling the upstream test suite as is.

martianboy:
Thanks for the insightful answer, @kripken. In that case, I can add (adapt) spec tests for elem and table once the remaining reference-types work is done.

tlively (approved):

PR #3680, opened 2021-03-12
[Wasm GC] Optimize struct stores like stores to memory, ignore unneeded bits

kripken:
When storing to an i8, we can ignore any higher bits, etc.

Adds a `getByteSize` utility to `Field` to make this convenient.

(diff is mostly whitespace)


src/passes/OptimizeInstructions.cpp
```diff
@@ -933,45 +933,51 @@ struct OptimizeInstructions
...
+        c->value = c->value.and_(
+          Literal::makeFromInt32(Bits::lowBitMask(bytes * 8), value->type));
       }
     }
     // stores of fewer bits truncates anyhow
-    if (auto* binary = curr->value->dynCast<Binary>()) {
+    if (auto* binary = value->dynCast<Binary>()) {
       if (binary->op == AndInt32) {
         if (auto* right = binary->right->dynCast<Const>()) {
           if (right->type == Type::i32) {
```
tlively, line 965: This section could use a matcher if you wanted.

tlively (approved):
Nice!

src/passes/OptimizeInstructions.cpp
kripken: This could be refactored to be shorter, good point. I'll leave that for a separate PR though, I tried to modify the existing code as little as possible.

PR #3681, opened 2021-03-12
[Wasm GC] Provide the local type in SimplifyLocals return value optimizations

kripken:
When we add a return value to an iff the we can no longer just call `finalize()`
and let it compute an LUB after #3669, instead we must provide that type. This is
more efficient anyhow, so we should have been doing it all along.

However @tlively , the fuzzer is finding more regressions and some look more
complex. For example this fails on `--remove-unused-brs`:
```wat
(module
 (func $foo (result i32)
  (unreachable)
 )
 (func $bar (result f64)
  (unreachable)
 )
 (func $0 (param $0 i32) (result funcref)
  (if
   (local.get $0)
   (return
    (ref.func $foo)
   )
   (return
    (ref.func $bar)
   )
  )
 )
)
```
The issue is that the returns are removed, letting the value flow out, and then we need to provide a type for the if. That is less trivial to fix than this PR as the if had no previous type, and the pass just calls ReFinalize which is unaware of the previously-existing returns. I'm not sure offhand how best to fix this, but it looks like there is some fallout from the LUB removal.


tlively:
Yeah, that pattern will need `LUBs` in `--remove-unused-brs`. It's precisely the pattern we were hoping not to see :(

kripken:
I think we might be able to avoid LUBs there, but it is not going to be easy in that pass. I hope it's the only pass, though, but am not sure. Anyhow, let's maybe chat about this later today?

A separate matter: this PR fails CI on something else. It claims that `(ref null func)` is not a supertype of `(ref null (func (result i32)))`. The reason it checks that is that we had `funcref` which turned into `(ref null func)` due to not allowing nullability. Now that subtype is a subtype of `funcref` but not a "nullable funcref", which seems wrong..? But what is wrong is that `funcref` means "any function" while `(ref null func)` means a function with no params or return. This seems like a mixup due to typed function references...

src/passes/SimplifyLocals.cpp
```diff
@@ -754,13 +760,14 @@ struct SimplifyLocals
...
     ifTrueBlock->finalize();
     assert(ifTrueBlock->type != Type::none);
+    auto localType = this->getFunction()->getLocalType(set->index);
     // Update the ifFalse side.
-    iff->ifFalse = builder.makeLocalGet(
-      set->index, this->getFunction()->getLocalType(set->index));
-    iff->finalize(); // update type
+    iff->ifFalse = builder.makeLocalGet(set->index, localType);
+    // Update the type of the if to that of the local.
+    iff->finalize(localType);
```
tlively, line 767: Is this one always precise? If not, it would be good to add a TODO like on the previous one.

tlively (approved):

tlively:
> A separate matter: this PR fails CI on something else. It claims that `(ref null func)` is not a supertype of `(ref null (func (result i32)))`. The reason it checks that is that we had `funcref` which turned into `(ref null func)` due to not allowing nullability. Now that subtype is a subtype of `funcref` but not a "nullable funcref", which seems wrong..? But what is wrong is that `funcref` means "any function" while `(ref null func)` means a function with no params or return. This seems like a mixup due to typed function references...

`(ref null func)` is identical to `funcref` both in Binaryen and in the spec, so something else doesn't add up here. It should certainly be true that `(ref null func) [i.e funcref] <: (ref null (func (result i32)))`, though.

kripken:
In principle this may be slightly more efficient, but as we have proper LUBs again, this is not needed, and doesn't seem worth the churn. Let's keep this heavily-fuzzed code as it is.

Issue #3682, opened 2021-03-12
Recursive GC types are not a partial order


kripken:
Tracking down the `std::map<Type, ...` issue, it looks like the title is the culprit. That is, there are recursive types for which `a < b, b < a`.

Testcase:
```wat
(module
 (type $A (func))
 (type $D (func (param i32)))
 (type $B (struct (field (ref $G))))
 (type $F (struct (field (ref $A))))
 (type $G (struct (field (ref $D))))
 (type $E (struct (field (ref $F)) (field i32)))
 (type $H (array (ref $B)))
 (func $foo
  (local $1 (ref $E))
  (local $2 (ref $B))
  (local $5 i32)
  (drop
   (ref.cast
    (array.get $H
     (ref.null $H)
     (local.get $5)
    )
    (rtt.canon $B)
   )
  )
 )
)
```

STR: apply this patch:
```diff
diff --git a/src/wasm/wasm-stack.cpp b/src/wasm/wasm-stack.cpp
index 2c8bc051f..e353ec6a0 100644
--- a/src/wasm/wasm-stack.cpp
+++ b/src/wasm/wasm-stack.cpp
@@ -2170,6 +2170,19 @@ void BinaryInstWriter::mapLocalsAndEmitHeader() {
     }
   }
   countScratchLocals();
+
+  for (auto& localType : localTypes) {
+    std::cout << "type: " << localType << '\n';
+  }
+  std::cout << "\n< matrix:\n";
+  for (auto& localType1 : localTypes) {
+    std::cout << "  ";
+    for (auto& localType2 : localTypes) {
+      std::cout << (localType1 < localType2) << ' ';
+    }
+    std::cout << '\n';
+  }
+
   std::map<Type, size_t> currLocalsByType;
   for (Index i = func->getVarIndexBase(); i < func->getNumLocals(); i++) {
     Index j = 0;
```
Then run
```
$ bin/wasm-opt w.wat -all -o a.wasm
warning: no passes specified, not doing any work
type: (ref null (struct (field (ref null (struct (field (ref null (func))))) i32)))
type: (ref null (struct (field (ref null (struct (field (ref null (func (param i32)))))))))
type: i32

< matrix:
  0 1 0 
  1 0 0 
  1 1 0 
```
The matrix shows that types 0 and 1 are mutually `<`.

cc @tlively

tlively:
Thanks for investigating this further. I will figure out what is wrong with `<`.

PR #3683, opened 2021-03-12
[Wasm GC] Avoid std::map<Type,..> in mapLocals

kripken:
Works around https://github.com/WebAssembly/binaryen/issues/3682 With this, I can roundtrip
a large real-world testcase.

There are other cases of `std::map<Type,..>` in the codebase, but none on
a critical path, so this PR only changes that one place.


tlively (approved):
Is this a determinism hazard or is this a fine long-term change? If it's a determinism hazard that should be reverted eventually, a comment would be good.

kripken:
I think it's fine long-term. We just don't need ordering here. So it's actually clearer written as unordered to annotate that.

PR #3689, opened 2021-03-15
Skip function bodies in wasm-emscripten-finalize when we don't need them

kripken:
After @sbc100 's work on `EM_ASM` and `EM_JS` they are now parsed from
the wasm using exports etc. and so we no longer need to parse function bodies.
As a result if we are not emitting a wasm from `wasm-emscripten-finalize` then all we are
doing is scanning global structures like imports and exports and emitting metadata
about them. And indeed we do not need to emit a wasm in some cases, specifically
when not optimizing and when using `WASM_BIGINT` (to avoid needing to
legalize).

We had considering skipping wasm-emscripten-finalize entirely in that situation,
and instead to parse the metadata from the wasm in python on the emscripten
side. However @sbc100 had the brilliant idea today to just skip function bodies.
That is very simple to do - no need to write another parser for wasm, and also
look at how simple this PR is - and also it will be faster to run
`wasm-emscripten-finalize` in this mode than to run python. (With the only
downside that the bytes of the wasm are loaded even if they aren't parsed; but
almost certainly they are in the disk cache anyhow.)

This PR implements that idea: when `wasm-emscripten-finalize` knows it will
not write a wasm output, it notes "skip function bodies". The binary reader then
skips the bodies and places `unreachable`s there instead (so that the wasm still
validates).

There are no new tests here because this can't be tested - by design it is an
unobservable optimization. (If we could notice the bodies have been skipped,
we would not have skipped them.) This is also why no changes are needed on
the emscripten side to benefit from this speedup. Basically when binaryen sees
it will not need X, it skips parsing of X automatically.

Benchmarking speed, it is as fast as you'd expect: the `wasm-emscripten-finalize`
step is 15x faster on SQLite (1MB of wasm) and almost 50x faster on the biggest
wasm I have on my drive (40MB of LLVM).

Tested manually and also on `wasm0 wasm2 other` on emscripten.

cc @pfaffe @bmeurer - you might like this!

### DIFF
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -2137,7 +2137,17 @@ void WasmBinaryBuilder::readFunctions() {
       assert(controlFlowStack.empty());
       assert(letStack.empty());
       assert(depth == 0);
-      func->body = getBlockOrSingleton(func->sig.results);
+      if (!skipFunctionBodies) {
+        func->body = getBlockOrSingleton(func->sig.results);
+      } else {
+        // When skipping the function body we need to put something valid in
+        // their place so we validate. An unreachable is always acceptable
+        // there.
+        func->body = Builder(wasm).makeUnreachable();
+
+        // Skip reading the contents.
+        pos = endOfFunction;
+      }
       assert(depth == 0);
       assert(breakStack.empty());
       assert(breakTargetNames.empty());
--- a/src/tools/fuzzing.h
+++ b/src/tools/fuzzing.h
@@ -27,6 +27,7 @@ high chance for set at start of loop
 
 // TODO Generate exception handling instructions
 
+#include "ir/branch-utils.h"
 #include "ir/memory-utils.h"
 #include <ir/find_all.h>
 #include <ir/literal-utils.h>
@@ -865,39 +866,20 @@ private:
 
       void visitExpression(Expression* curr) {
         // Note all scope names, and fix up all uses.
-
-#define DELEGATE_ID curr->_id
-
-#define DELEGATE_START(id)                                                     \
-  auto* cast = curr->cast<id>();                                               \
-  WASM_UNUSED(cast);
-
-#define DELEGATE_GET_FIELD(id, name) cast->name
-
-#define DELEGATE_FIELD_SCOPE_NAME_DEF(id, name)                                \
-  if (cast->name.is()) {                                                       \
-    if (seen.count(cast->name)) {                                              \
-      replace();                                                               \
-    } else {                                                                   \
-      seen.insert(cast->name);                                                 \
-    }                                                                          \
-  }
-
-#define DELEGATE_FIELD_SCOPE_NAME_USE(id, name) replaceIfInvalid(cast->name);
-
-#define DELEGATE_FIELD_CHILD(id, name)
-#define DELEGATE_FIELD_OPTIONAL_CHILD(id, name)
-#define DELEGATE_FIELD_INT(id, name)
-#define DELEGATE_FIELD_INT_ARRAY(id, name)
-#define DELEGATE_FIELD_LITERAL(id, name)
-#define DELEGATE_FIELD_NAME(id, name)
-#define DELEGATE_FIELD_NAME_VECTOR(id, name)
-#define DELEGATE_FIELD_SCOPE_NAME_USE_VECTOR(id, name)
-#define DELEGATE_FIELD_SIGNATURE(id, name)
-#define DELEGATE_FIELD_TYPE(id, name)
-#define DELEGATE_FIELD_ADDRESS(id, name)
-
-#include "wasm-delegations-fields.h"
+        BranchUtils::operateOnScopeNameDefs(curr, [&](Name& name) {
+          if (name.is()) {
+            if (seen.count(name)) {
+              replace();
+            } else {
+              seen.insert(name);
+            }
+          }
+        });
+        BranchUtils::operateOnScopeNameUses(curr, [&](Name& name) {
+          if (name.is()) {
+            replaceIfInvalid(name);
+          }
+        });
       }
 
       bool replaceIfInvalid(Name target) {
--- a/src/tools/wasm-emscripten-finalize.cpp
+++ b/src/tools/wasm-emscripten-finalize.cpp
@@ -203,9 +203,21 @@ int main(int argc, const char* argv[]) {
     Fatal() << "Need to specify an infile\n";
   }
 
+  // We will write the modified wasm if the user asked us to, either by
+  // specifying an output file or requesting text output (which goes to stdout
+  // by default).
+  auto writeOutput = outfile.size() > 0 || !emitBinary;
+
   Module wasm;
   ModuleReader reader;
   reader.setDWARF(DWARF);
+  if (!writeOutput) {
+    // If we are not writing the output then all we are doing is simple parsing
+    // of metadata from global parts of the wasm such as imports and exports. In
+    // that case, it is unnecessary to parse function contents which are the
+    // great bulk of the work, and we can skip all that.
+    reader.setSkipFunctionBodies(true);
+  }
   try {
     reader.read(infile, wasm, inputSourceMapFilename);
   } catch (ParseException& p) {
@@ -303,9 +315,7 @@ int main(int argc, const char* argv[]) {
   BYN_TRACE_WITH_TYPE("emscripten-dump", "Module after:\n");
   BYN_DEBUG_WITH_TYPE("emscripten-dump", std::cerr << wasm << '\n');
 
-  // Write the modified wasm if the user asked us to, either by specifying an
-  // output file, or requesting text output (which goes to stdout by default).
-  if (outfile.size() > 0 || !emitBinary) {
+  if (writeOutput) {
     Output output(outfile, emitBinary ? Flags::Binary : Flags::Text);
     ModuleWriter writer;
     writer.setDebugInfo(debugInfo);
--- a/src/wasm-binary.h
+++ b/src/wasm-binary.h
@@ -1307,6 +1307,7 @@ class WasmBinaryBuilder {
   std::istream* sourceMap;
   std::pair<uint32_t, Function::DebugLocation> nextDebugLocation;
   bool DWARF = false;
+  bool skipFunctionBodies = false;
 
   size_t pos = 0;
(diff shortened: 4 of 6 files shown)


src/wasm/wasm-binary.cpp
```diff
@@ -2137,7 +2137,17 @@ void WasmBinaryBuilder::readFunctions() {
...
       assert(letStack.empty());
       assert(depth == 0);
-      func->body = getBlockOrSingleton(func->sig.results);
+      if (!skipFunctionBodies) {
+        func->body = getBlockOrSingleton(func->sig.results);
+      } else {
+        // When skipping the function body we need to put something valid in
+        // their place so we validate. An unreachable is always acceptable
+        // there.
+        func->body = Builder(wasm).makeUnreachable();
```
sbc100, line 2146: Can we mark the function in such as way that if anyone tries to walk it we will assert?

sbc100 (commented):
In theory we could skip the entire code section, not just the bodies?

We don't actually need the module to a validate at all here,  right? 

Can we save even more time by skipping any other validation steps?

sbc100 (approved):
Thanks for doing this!

src/wasm/wasm-binary.cpp
kripken: There isn't a mechanism for that - we'd need to add a new property on functions. Or a new instruction "Unwalkable" perhaps? I'm not sure it's worth either though, as an error here would not be a subtle bug, it would break very loudly.

kripken:
> In theory we could skip the entire code section, not just the bodies?

Yes. But that would require a larger refactoring, as we do still need to create the functions, set up their global state, etc. (which can be checked by the validator). The current PR just skips the body and leaves the rest of the code as is.

Maybe we could also skip the local types, but I doubt that's a significant amount of work, and it would complicate the PR.

> We don't actually need the module to a validate at all here, right?

I think it's better to check it validates. There might be a validation error in an import for example that we don't want to miss because of this optimization.

sbc100:
> > In theory we could skip the entire code section, not just the bodies?
> 
> Yes. But that would require a larger refactoring, as we do still need to create the functions, set up their global state, etc. (which can be checked by the validator). The current PR just skips the body and leaves the rest of the code as is.
> 
> Maybe we could also skip the local types, but I doubt that's a significant amount of work, and it would complicate the PR.
> 
> > We don't actually need the module to a validate at all here, right?
> 
> I think it's better to check it validates. There might be a validation error in an import for example that we don't want to miss because of this optimization.

I think if we want the `finalize` step to perform validation we should be explicit about that.    Perhaps folks have been implicitly assuming that finalize does validation but:

a) I don't think that should ever have been part of our contract
b) doesn't this change break that contract.. invalid wasm modules (at least in the code section) can now pass through uncecked?

src/wasm/wasm-binary.cpp
sbc100: I think it might break silently.   What if I write a new pass in `wasm-emscripten-finalize` that searches for a certain instruction.   With this change that pass will silently always fail, right?

kripken: It will silently always return "false" I suppose, but that's what I meant by "loudly". But I guess there could be a subtle new pass in which it's not as loud as it would currently be, fair point...

kripken:
I think it's a basic assumption that all binaryen tools validate their input - not for external reasons (like looking for LLVM bugs) but because it is unsafe (both to binaryen and to the output) to run optimizations on unvalidated input.

You're right that an invalid module might pass through, but I think that's intended. If we skip function bodies then we also skip validation of function bodies. And the worst we can do in that case is pass through a bad output from LLVM (which I don't think that is something binaryen should be in charge of detecting).

sbc100:
> I think it's a basic assumption that all binaryen tools validate their input - not for external reasons (like looking for LLVM bugs) but because it is unsafe (both to binaryen and to the output) to run optimizations on unvalidated input.
> 
> You're right that an invalid module might pass through, but I think that's intended. If we skip function bodies then we also skip validation of function bodies. And the worst we can do in that case is pass through a bad output from LLVM (which I don't think that is something binaryen should be in charge of detecting).

I agree.. and I think that is fine.  I don't think it should be up to finalize to do validation (necessarily)   .. and thats why I think doing even less validation would also be fine.  

But anyway, this change lgtm.

kripken:
@tlively It would be hard to get the module to validate in that case. Exports and imports can refer to functions, and the validator checks that.

pfaffe:
This change saves 50% of the time spent in wasm-emscripten-finalize! We're going from 11s to 5.5.

pfaffe:
That being said, I'm all for doing even less work! +1 to what Sam said, validation could be guarded behind some command line flag.

pfaffe:
I did some more measurements, and it turns out we're spending 65% of the finalize time in `readNames()` now. Are the name section contents required for the EM_ASM/JS lowering or can we skip that step as well?

sbc100:
> I did some more measurements, and it turns out we're spending 65% of the finalize time in `readNames()` now. Are the name section contents required for the EM_ASM/JS lowering or can we skip that step as well?

I think it can be skipped:  finalize should work exactly the same with or without a names section.  (and perhaps there is also some optimization that could happen there?)

pfaffe:
I didn't look too long at the profile, but there weren't any low hanging fruit that stood out. The string manipulation is what's consuming basicly all of the time, i.e., escaping, deduplicating, hashing. The best "easy" improvement I can see is switching to a different string implementation to speed up manipulation, similar to LLVM's Twines. But we still spend most of our time hashing strings. Above 45% of our time is spent in IString::set, for example.

Heres some profile data:
```
+   64.53%     0.00%  wasm-emscripten  wasm-emscripten-finalize  [.] wasm::WasmBinaryBuilder::readUserSection                                                
+   64.48%     1.40%  wasm-emscripten  wasm-emscripten-finalize  [.] wasm::WasmBinaryBuilder::readNames                                                      
+   46.90%    23.28%  wasm-emscripten  wasm-emscripten-finalize  [.] cashew::IString::set                                                                    
+   36.02%     7.42%  wasm-emscripten  wasm-emscripten-finalize  [.] wasm::WasmBinaryBuilder::escape                                                         
+   22.56%    11.97%  wasm-emscripten  wasm-emscripten-finalize  [.] std::_Hashtable<char const*, char const*, std::allocator<char const*>, std::__detail::_I
+   17.21%     0.11%  wasm-emscripten  wasm-emscripten-finalize  [.] wasm::WasmBinaryBuilder::getInlineString                                                
+   10.55%     0.46%  wasm-emscripten  wasm-emscripten-finalize  [.] std::_Hashtable<char const*, char const*, std::allocator<char const*>, std::__detail::_I
+    9.98%     9.98%  wasm-emscripten  wasm-emscripten-finalize  [.] std::_Hashtable<char const*, char const*, std::allocator<char const*>, std::__detail::_I
+    6.85%     6.85%  wasm-emscripten  wasm-emscripten-finalize  [.] std::unordered_set<char const*, cashew::IString::CStringHash, cashew::IString::CStringEq
```

kripken:
Interesting @pfaffe !

I think this is related to the previous name-related slowdown I haven't seen on my machine. Did we ever figure that out? If not I wonder if maybe my build of LLVM is old enough that the names section has a different structure perhaps. If you get a chance can you send me your wasm, or instructions to build it?

I think we may be able to skip the names section but it's not trivial. Probably should be left to a separate PR. The main issue is that names may matter when validating, for error messages. Perhaps when finalize does not modify the binary it is safe to not validate, as @sbc100 proposed before. That would require changes on both emscripten and binaryen.

About the string slowdown, that's odd. The cost seems to be the very initial reading and hashing of the names, and perhaps in hash table collisions but those should be very rare. My theory from earlier is that maybe you have more collisions on your machine due to a different C++ STL but I don't think we've ever checked that. I suppose we could store the hash on IStrings to speed up hash collisions, but it would double their size so I'm not sure if that would end up faster.

tlively:
If wasm-emscripten-finalize only needs to read some particular information out of the module, it doesn't seem like it's a great fit for Binaryen anymore. What if we made it a small separate tool based on Wasp's fast streaming binary parser, which is able to skip directly to the interesting sections? That seems cleaner to me than adding options to sort-of-but-not-completely skip things in Binaryen's parser.

Of course this is a non-starter if wasm-emscripten-finalize ever has to run any other passes. Is that still the case?

sbc100:
> If wasm-emscripten-finalize only needs to read some particular information out of the module, it doesn't seem like it's a great fit for Binaryen anymore. What if we made it a small separate tool based on Wasp's fast streaming binary parser, which is able to skip directly to the interesting sections? That seems cleaner to me than adding options to sort-of-but-not-completely skip things in Binaryen's parser.

If we go that path we might as will just write it directly in emscripten/python I think.  That was the original plan but this patch is a nice way to get a lot of wins without much work.

> 
> Of course this is a non-starter if wasm-emscripten-finalize ever has to run any other passes. Is that still the case?

IIUC the only remaining pass that needs to modify the binary is when we lack WASM_BIGINT we need to run the legalization pass.

kripken:
In addition to legalization I think we also modify the wasm in finalize for things like adding stack checks and moving the memory initialization to a side file.

(In theory those could all be moved to `wasm-opt` later - and we could rename finalize "wasm-metadata" perhaps - but that would force running `wasm-opt` in more places, so it just moves the tradeoff around.)

kripken:
For some reason CI doesn't want to run on this PR... merged in main to try to get it.

Any objections to landing this? I think it's a good change regardless of that we also want to try to skip names parsing separately.

sbc100:
lgtm

sbc100:
Looks like this somehow broke an emscripten browser test:

```
test_pthread_unistd_io_bigint (test_browser.browser) ... wasm-emscripten-finalize: /b/s/w/ir/cache/builder/emscripten-releases/binaryen/src/wasm/wasm-emscripten.cpp:276: std::vector<AsmConst> wasm::findEmAsmConsts(wasm::Module &, bool): Assertion `asmConsts.size()' failed.
emcc: error: '/root/emsdk/upstream/bin/wasm-emscripten-finalize --detect-features --minimize-wasm-changes --bigint --no-dyncalls --no-legalize-javascript-ffi test.wasm' failed (-6)
FAIL
```

kripken:
Revert in https://github.com/WebAssembly/binaryen/pull/3705 with more details from offline discussion.

PR #3692, opened 2021-03-15
Add missing fields in ModuleUtils::copyModule

kripken:
We were missing type names and the features section.


tlively (approved):
Can the binary test file not be text?

kripken:
The difficulty is we can't declare the features section in text. (Maybe we could start with text and add a flag to add the features section, but that feels less direct of a test to me.)

tlively:
> The difficulty is we can't declare the features section in text. (Maybe we could start with text and add a flag to add the features section, but that feels less direct of a test to me.)

Hmm, actually that sounds better to me. At least we would be able to see what the test input is.

PR #3693, opened 2021-03-15
Emit dollar signs when relevant while debugging s-expression elements

kripken:
This is just noticeable when debugging locally and doing a quick print to
stdout.


tlively (approved):

aheejin (approved):
It looks we need to update the test output too

PR #3694, opened 2021-03-15
Validate code in global data structures

kripken:
This validation is almost never needed, but it starts to get interesting with
GC, where a global initializer can be an `rtt.sub` which must be valid.

No tests here as testing requires a further GC fix in a later PR.

Maybe it's not worth validating offsets? Those are even more constrained
as to what they can contain than even global initializers. But it seems best
to validate them just to be sure, so we don't miss something in some
future wasm feature.

### DIFF
--- a/src/wasm/wasm-validator.cpp
+++ b/src/wasm/wasm-validator.cpp
@@ -209,6 +209,13 @@ struct FunctionValidator : public WalkerPass<PostWalker<FunctionValidator>> {
 
   FunctionValidator(ValidationInfo* info) : info(*info) {}
 
+  void validateInModule(Module& wasm, Expression* curr) {
+    auto* oldModule = getModule();
+    setModule(&wasm);
+    walk(curr);
+    setModule(oldModule);
+  }
+
   std::unordered_map<Name, std::unordered_set<Type>> breakTypes;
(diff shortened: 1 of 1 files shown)


tlively (approved):

src/wasm/wasm-validator.cpp
```diff
@@ -209,6 +209,13 @@ struct FunctionValidator : public WalkerPass<PostWalker<FunctionValidator>> {
 
   FunctionValidator(ValidationInfo* info) : info(*info) {}
 
+  void validateInModule(Module& wasm, Expression* curr) {
+    auto* oldModule = getModule();
+    setModule(&wasm);
+    walk(curr);
+    setModule(oldModule);
```
aheejin, line 216: Is it possible to call this function with a different module than its old module?

aheejin (approved):

src/wasm/wasm-validator.cpp
kripken: Not in the current code. But it seemed risky to me to leave the module modified, as then a future refactoring might run into surprising behavior.

But you made me think that it does seem slightly odd to even consider that possible. I pushed a small refactoring now to require a module on creation of a FunctionValidator, which will then never change.

PR #3695, opened 2021-03-16
[NFC] Record isTemp and isSelfReferential on TypeInfos

tlively:
`isTemp` is used in new assertions that guard against leaking temporary types
and heap types into the global stores and `isSelfReferential` replaces and
external set of self-referential types. Both fields will additionally be used in
upcoming PRs improving type canonicalization.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -33,6 +33,7 @@ namespace {
 
 struct TypeInfo {
   using type_t = Type;
+  bool isTemp = false;
   enum Kind {
     TupleKind,
     RefKind,
@@ -68,6 +69,8 @@ struct TypeInfo {
 
 struct HeapTypeInfo {
   using type_t = HeapType;
+  bool isTemp = false;
+  bool isSelfReferential = false;
   enum Kind {
     SignatureKind,
     StructKind,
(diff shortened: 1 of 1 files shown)


src/wasm/wasm-type.cpp
```diff
@@ -1211,7 +1272,8 @@ struct TypeBuilder::Impl {
       info = std::make_unique<HeapTypeInfo>(Signature());
     }
     void set(HeapTypeInfo&& hti) {
-      *info = hti;
+      *info = std::move(hti);
```
kripken, line 1275: is this just for clarity?

kripken (approved):

src/wasm/wasm-type.cpp
tlively: C++ is weird because even if you have an rvalue (&&) reference, you still need to use std::move in order to get move semantics, so this code was not working as intended before. (Rust does not have such pitfalls 😉) See https://stackoverflow.com/questions/31213539/c-stdmove-with-rvalue-reference-is-not-moving-contents.

kripken: Ah, I did not know that... well, `std::move` everywhere then... :shrug:

PR #3698, opened 2021-03-17
Do not read the Names section when debug info is disabled by wasm-emscripten-finalize

kripken:
In that tool we know that if debug info is not requested then we do not even
need to read it, as metadata does not depend on names. This saves loading
and interning all the names, which takes several seconds on massive inputs.

This is possible in principle in other tools, but this does not change anything
in them for now. (We do use names internally in some nontrivial ways without
opting in to it, so that would require further refactoring. But in finalize the
situation is simple and easy to handle.)

This not 100% unobservable. If validation fails then the validation error would
just contain the function index instead of the name from the Names section if
there is one. In that case the command can be re-run with `-g` to parse the
debug info. This seems like a reasonable tradeoff in that validation errors in
finalize - that is, malformed LLVM output - are very rare (in fact we have
considered removing this validation entirely), and also the tool can
be invoked with `-g` as mentioned earlier if one always wants the full error.
That is in fact what emscripten does atm, so to get a speedup from this we
need a change there as well.

### DIFF
--- a/src/tools/wasm-emscripten-finalize.cpp
+++ b/src/tools/wasm-emscripten-finalize.cpp
@@ -210,6 +210,7 @@ int main(int argc, const char* argv[]) {
 
   Module wasm;
   ModuleReader reader;
+  reader.setDebugInfo(debugInfo);
   reader.setDWARF(DWARF);
   if (!writeOutput) {
     // If we are not writing the output then all we are doing is simple parsing
--- a/src/tools/wasm-emscripten-finalize.cpp
+++ b/src/tools/wasm-emscripten-finalize.cpp
@@ -210,8 +210,10 @@ int main(int argc, const char* argv[]) {
 
   Module wasm;
   ModuleReader reader;
-  reader.setDebugInfo(debugInfo);
-  reader.setDWARF(DWARF);
+  // If we are not writing output then we definitely don't need to read debug
+  // info, as it does not affect the metadata we will emit.
+  reader.setDebugInfo(debugInfo && writeOutput);
+  reader.setDWARF(DWARF && writeOutput);
   if (!writeOutput) {
     // If we are not writing the output then all we are doing is simple parsing
     // of metadata from global parts of the wasm such as imports and exports. In
--- a/src/tools/wasm-emscripten-finalize.cpp
+++ b/src/tools/wasm-emscripten-finalize.cpp
@@ -211,8 +211,10 @@ int main(int argc, const char* argv[]) {
   Module wasm;
   ModuleReader reader;
   // If we are not writing output then we definitely don't need to read debug
-  // info, as it does not affect the metadata we will emit.
-  reader.setDebugInfo(debugInfo && writeOutput);
+  // info, as it does not affect the metadata we will emit. (However, if we
+  // emit output then definitely load the names section so that we roundtrip
+  // names properly.)
+  reader.setDebugInfo(writeOutput);
   reader.setDWARF(DWARF && writeOutput);
   if (!writeOutput) {
     // If we are not writing the output then all we are doing is simple parsing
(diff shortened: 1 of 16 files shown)


src/tools/wasm-emscripten-finalize.cpp
```diff
@@ -210,6 +210,7 @@ int main(int argc, const char* argv[]) {
 
   Module wasm;
   ModuleReader reader;
+  reader.setDebugInfo(debugInfo);
```
sbc100, line 213: Is this the right flag though?    What if I have a debug build that I want finalize to be fast on?   Wouldn't make more sense to set this to `!emitBinary` or maybe even better `!emitBinary && debugInfo`?

kripken: This line just applies debugInfo from the commandline into the binary reader, which seems straightforward? If debug info handling is not requested, we don't even need to load it.

I agree this could perhaps be tied into `emitBinary` as well (if you are not writing anything, you are not writing debug info as well), but that does not improve things and adds a little more complexity to the code.

sbc100: But isn't the debug build case one of the main cases that we want to speed up.   In that case we are passing `-g` and we also don't want to modify the binary.   That is the case we care about the most isn't it?

kripken: Yes, that is the case.

I am trying to keep this PR as simple as possible. The idea is when debug info is not used we don't need to load it - that seems straightforward?

There will be more complexity on the emscripten side, but we'll try to keep it simple there. But yes, it will consider whether it is writing a wasm from finalize when it decides to pass in `-g` or not.

sbc100: I think emscripten should still pass `-g` to signal that debug info should not be stripped.   

Whether we parse debug info here in binaryen should depend of whether we need to read/modify or write it, regardless of the `-g` flag.

Anyway, this change lgtm.

sbc100 (approved):

src/tools/wasm-emscripten-finalize.cpp
kripken: I see your point of view as well. But consider the downside of this PR, which is that validation errors lose names (and have indexes instead). With this PR as it is written, you can re-add `-g` to get the name. If on the other hand we do it your way, then you'd need to add a `-o unused` for that. That seems less clear to me. But, I don't feel strongly.

sbc100: I don't think we should be depending on finalize to do validation.. its historically been a side effect, but we recently removed all code validation anyway.

If we want it to do validation (I'm not sure we will) then we can add an explicit `--validate` flag to disable both this and the code section skipping.

sbc100: By default we want finalize to be fast and skip any steps that it can which is why I think `--validate` should be opt in.

sbc100: (Obviously validation is required or some of things that finalize does so we have no choice in that case.. but when we do have a choice don't we want to skip by default?)

kripken: > I don't think we should be depending on finalize to do validation.. its historically been a side effect, but we recently removed all code validation anyway.

Huh, that surprises me...

Indeed I don't see it in the code. But where did we do this? I don't see it in the `git log` for that file.

kripken: Separately from that you are 100% right that this needs to check if we are writing output - the `lld` tests here fail on that. So that can't be done purely on the emscripten side. Anyhow, not a big deal, pushed to here.

kripken:
Actually another reason this is better this way is that an emscripten change will not be needed. Emscripten already does not ask us to write a wasm when we don't need to. And then after this PR we will not read names in that case automatically.

sbc100 (commented):
Nice!

kripken:
@pfaffe @bmeurer - this implements the optimization we spoke about. It turns out that we do not need an emscripten change, so this is done.

PR #3699, opened 2021-03-17
Change upstream to origin/main in clang-tidy-diff.sh

aheejin:
We changed the upstream branch from `origin/main` to `@{upstream}` in
 #3619:
Context:
https://github.com/WebAssembly/binaryen/pull/3619#discussion_r583338724
https://github.com/WebAssembly/binaryen/pull/3619#discussion_r583717142

But I don't think we were actually able to test that part in that PR,
because PRs use `$GITHUB_BASE_REF` instead. And it looks `@{upstream}`
doesn't work on my local machine with branches. Is it supposed to work
locally?

Also, it looks all clang-tidy operations in the repo's commits are
failing, which I think uses the upstream thing. But strangely the step
is reported as success.
Failure link:
https://github.com/WebAssembly/binaryen/runs/2134499599#step:8:50
I'm not sure why, because clang-tidy-diff.sh is
supposed to fail whenever a command in there fails, because we have
`set -o errexit` at the top.

This anyway reverts `@{upstream}` to `origin/main`. This makes it work
locally. (There is still one constraint: You have to commit your changes
locally to a branch before you run this. We use
`git diff -U0 $BRANCH...`, and the three dots thing doesn't work with
uncommitted local changes.) I can't test the repo commit issue locally,
but let's hope this also fixes the commit issue.

### DIFF
--- a/scripts/clang-tidy-diff.sh
+++ b/scripts/clang-tidy-diff.sh
@@ -5,7 +5,7 @@ set -o errexit
 if [ -n "$GITHUB_BASE_REF" ]; then
   BRANCH="origin/$GITHUB_BASE_REF"
 else
-  BRANCH="@{upstream}"
+  BRANCH=origin/main
 fi
 
 CLANG_TIDY=$(which clang-tidy)


aheejin:
More precisely, when I use `@{upstream}` in my command bash shell, it works. But when it is in the script and I run the script, it does not work. `origin/main` works in both cases. Not sure why.

kripken (approved):
lgtm if it works...

sbc100:
It seems reasonable to me that `clang-tidy-diff.sh` would only work if you have an upstream configured.  Otherwise is doesn't know what you want to "diff" against.

Regarding why it doesn't work in the script.. that sounds very odd.

tlively:
@sbc100, since you were advocating for `@{upstream}`, do you have a strong opinion about this? This LGTM personally.

sbc100:
Can you run the script with `sh -x` to see what final git command looks like:

```
$ sh -x ./scripts/clang-tidy-diff.sh  
+ set -o errexit
+ '[' -n '' ']'
+ BRANCH='@{upstream}'
++ which clang-tidy
+ CLANG_TIDY=/usr/bin/clang-tidy
+ '[' '!' -e /usr/bin/clang-tidy ']'
++++ readlink -f /usr/bin/clang-tidy
+++ dirname /usr/lib/llvm-11/bin/clang-tidy
++ dirname /usr/lib/llvm-11/bin
+ CLANG_DIR=/usr/lib/llvm-11
+ CLANG_TIDY_DIFF=/usr/lib/llvm-11/share/clang/clang-tidy-diff.py
+ '[' '!' -e /usr/lib/llvm-11/share/clang/clang-tidy-diff.py ']'
++ git diff -U0 '@{upstream}...'
...
```

For me that command (`git diff -U0 '@{upstream}...'`) works fine.. both in the script and when I run it directly.  What do you see?

aheejin:
This is what I saw after
```
(I'm in branch main in the beginning)
$ git checkout -b error
Introduce a clang-tidy error in some cpp file
$ git commit -a -m 'error'
$ git push --set-upstream origin test (to create upstream. without this the script errors out)
```

```
$ sh -x ./scripts/clang-tidy-diff.sh  
+ set -o errexit
+ '[' -n '' ']'
+ BRANCH='@{upstream}'
++ which clang-tidy
+ CLANG_TIDY=/usr/local/google/home/aheejin/apps/clang+llvm/bin/clang-tidy
+ '[' '!' -e /usr/local/google/home/aheejin/apps/clang+llvm/bin/clang-tidy ']'
++++ readlink -f /usr/local/google/home/aheejin/apps/clang+llvm/bin/clang-tidy
+++ dirname /usr/local/google/home/aheejin/apps/llvm-project-11.0.1/install/bin/clang-tidy
++ dirname /usr/local/google/home/aheejin/apps/llvm-project-11.0.1/install/bin
+ CLANG_DIR=/usr/local/google/home/aheejin/apps/llvm-project-11.0.1/install
+ CLANG_TIDY_DIFF=/usr/local/google/home/aheejin/apps/llvm-project-11.0.1/install/share/clang/clang-tidy-diff.py
+ '[' '!' -e /usr/local/google/home/aheejin/apps/llvm-project-11.0.1/install/share/clang/clang-tidy-diff.py ']'
++ git diff -U0 '@{upstream}...'
++ /usr/local/google/home/aheejin/apps/llvm-project-11.0.1/install/share/clang/clang-tidy-diff.py -quiet -p1
+ TIDY_MSG='No relevant changes found.'
+ '[' -n 'No relevant changes found.' -a 'No relevant changes found.' '!=' 'No relevant changes found.' ']'
```

It does not error out, but the clang-tidy error I introduced is not detected, and it says just "No relevant changes found."

sbc100:
And what does `git diff -U0 '@{upstream}...'` display ?

I think probably the `git push --set-upstream origin test ` command you are using is setting your upstream to `origin/test` rather than `origin/main`.. which would lead to an empty diff from your upstream.

I normally run `git branch -o origin/main` after creating a new branch.   

(If I want to push the branch I just usel `git push`)

Maybe if you are not keep on keeping your upstream maintained in this way we can just add an argument to the script.. but IMHO `@{upstream}` is the logical default when doing this kind of operation.

aheejin:
I use my fork and create PRs from that fork's branch. And I think that workflow is the only one possible for people who don't have the write access to the repo. (I have the access but I'm just used to that workflow) In that workflow you need to make a tracking branch, which is different from your `origin/main`, for each of your local branch.

If many of us use this workflow, I guess it is reasonable to make it work for them too. Also always having to have the upstream tracking branch is one more constraint; I don't always set them until right before I upload a PR.

Do you have any suggestions on how to "add an argument and keep use `@{upstream}`"? But if that makes the script complicated, I think using `origin/main` can be simpler. Does `origin/main` work for your workflow too?

aheejin:
And sorry I think I was confused. `@{upstream}` doesn't work both in script and in bash shell for me.

sbc100:
> `git diff -U0 '@{upstream}...'` displays nothing in my example.

Right, so you need to run `git branch -u origin/main` and that should fix it right?

> 
> I use my fork and create PRs from that fork's branch. And I think that workflow is the only one possible for people who don't have the write access to the repo. (I have the access but I'm just used to that workflow) In that workflow you need to make a tracking branch, which is different from your `origin/main`, for each of your local branch.

But you can still use `origin/main` as your upstream can't you?   Logically that is your upstream right?  Its the place you want the PR to land once its done.    You might also want to set it to `upstream/main` if upstream is the actual target of the PR (as opposed to origin which is your fork?)

> 
> If many of us use this workflow, I guess it is reasonable to make it work for them too. Also always having to have the upstream tracking branch is one more constraint; I don't always set them until right before I upload a PR.

Ok perhaps we can detect the absence of an upstream being set.  That seems reasonable too.

aheejin:
For me the purpose of setting upstream is so that you can push conveniently without arguments. If you make a `feature` branch and set the upstream to `origin/main` or even `upstream/main`, and you push to your `origin/feature` using `git push origin HEAD:feature`, I'm not sure what the point of setting remote tracking branch is? Also I'm not sure that's the workflow many people are using.

Does the `origin/main` revert of this PR not work for you? If this works for all people I'm not sure if we need to detect if the upstream is set...

aheejin:
Chatted w/ @sbc100 offline. It looks this `origin/main` doesn't work for his workflow, and we also concluded the reason for https://github.com/WebAssembly/binaryen/runs/2134499599#step:8:50 is unrelated to this. We can make the script a bit more complicated to make this work for both workflows, but I guess that would better be done in a new PR. Closing.

Issue #3697, opened 2021-03-17
crash in wasm-opt at -O2


sbc100:
The wasm2.test_poppler test is failing consistenly with recent builds of wasm-opt:

https://app.circleci.com/pipelines/github/emscripten-core/emscripten/11775/workflows/4dd05dca-592a-47f3-9e74-671debdfd795/jobs/404629

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  _Map_base::at
em++: error: '/root/emsdk/upstream/bin/wasm-opt --post-emscripten --no-exit-runtime -O2 --low-memory-unused --zero-filled-memory --strip-producers pdfimages.wasm -o pdfimages.wasm -g --mvp-features' failed (-6)
make[2]: *** [pdfimages] Error 1
make[2]: *** Waiting for unfinished jobs....
```

I have not yet bisected to a particular binaryen commit but I suspect its something in the last 24 hours.

kripken:
I don't see this locally. Perhaps it is running out of memory. I don't think we've changed anything on the Binaryen side that could explain it. Perhaps an LLVM change is causing more code or locals to be emitted?

sbc100:
I can't repro it locally either.  Strange.  I wonder if its circleci change?

sbc100:
Actually I can repro locally if I do `./emsdk install tot` and use that.

sbc100:
Seems to be related to the binaryen binaryen built on the waterfall.

kripken:
Interesting, if you can reproduce this with the waterfall builds then we can bisect it at least.

Looks like the sheriff is @aheejin - do you have time? Or did you intend to @sbc100 ? If neither then I can find time.

kripken:
@aheejin Actually I just tried this again before I saw you self-assigned this and I can get this to reproduce locally. From the stack trace it is in debug info code, so likely my fault or wouter - I can verify shortly.

kripken:
This was actually LLVM's "fault" for emitting different DWARF now. So this was a crash in all opt levels actually, whenever we rewrote DWARF. Fix in https://github.com/WebAssembly/binaryen/pull/3700

sbc100:
Do you want to make a PR to re-enable the poppler test? (or I can if you like)

kripken:
I will. Still waiting on the binaryen roll.

PR #3700, opened 2021-03-18
Ignore missing CUs in DWARF rewriting

kripken:
A recent change in LLVM causes it to sometimes end up with a thing
with no parent. That is, a debug_line or a debug_loc that has no CU that
refers to it. This is perhaps LLVM DCEing CUs, or something else that
changed - not sure. But it seems like valid DWARF we should handle.

This PR handles that in our code. Two things broke here. First, locs must
be simply ignored when there is no CU. Second, lines are trickier as we
used to compute their position by scanning them, and that list contained
only ones with a CU. So we missed some and ended up with wrong offsets.
To make things simpler and more robust, just track the position of each
line table on itself.

Fixes https://github.com/WebAssembly/binaryen/issues/3697

### DIFF
--- a/third_party/llvm-project/include/llvm/ObjectYAML/DWARFYAML.h
+++ b/third_party/llvm-project/include/llvm/ObjectYAML/DWARFYAML.h
@@ -151,6 +151,7 @@ struct LineTableOpcode {
 };
 
 struct LineTable {
+  uint64_t Position; // XXX BINARYEN: the binary location in .debug_line
   InitialLength Length;
   uint16_t Version;
   uint64_t PrologueLength;
--- a/src/wasm/wasm-debug.cpp
+++ b/src/wasm/wasm-debug.cpp
@@ -624,12 +624,6 @@ struct LocationUpdater {
   }
 
   BinaryLocation getNewDebugLineLocation(BinaryLocation old) const {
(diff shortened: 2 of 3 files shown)


third_party/llvm-project/include/llvm/ObjectYAML/DWARFYAML.h
```diff
@@ -151,6 +151,7 @@ struct LineTableOpcode {
 };
 
 struct LineTable {
+  uint64_t Position; // XXX BINARYEN: the binary location in .debug_line
```
aardappel, line 154: This is set but not used anywhere in this PR?

aardappel (approved):

third_party/llvm-project/include/llvm/ObjectYAML/DWARFYAML.h
kripken: It's used here: https://github.com/WebAssembly/binaryen/pull/3700/files#diff-62256ba16b71cbddf493d990a92cc71e51a3f2ae0a57289ae85ae5d794db3d31R765

We set this during load. During rewrite, we read these values so we know the old position. (We also update it there, although we do not currently read at any later time - I added that to avoid future confusion, seems best to the keep the IR consistent.)

PR #3708, opened 2021-03-18
Perform clang-tidy/clang-format linting only on PRs

sbc100:
These steps diff between a PR and upstream which doesn't
really make sense for normal CI runs.


aheejin (approved):
Thanks!

PR #3711, opened 2021-03-19
Allow clang-format-diff/clang-tidy-diff to take branch as argument

sbc100:
Also update clang-format-diff.sh to match recent changes to branch
name and CI system (these changes mirror those already made to
clang-tidy-diff.sh).


aheejin (approved):
Thank you!

PR #3714, opened 2021-03-22
wasm-emscripten-finalize: Do not skip the start function body

kripken:
When we can skip function bodies, we still need to parse the start function
for the pthreads case, see details in the comments. This still gives us 99%
of the speedup as the start function is just 1 function and it's not that big,
so with this we return to full speed after the reversion in #3705

While this is a hack, as @tlively mentioned offline the entire idea of skipping
function bodies is a hack. But it's the good kind of hack IMO, giving a huge
benefit for users and without too burdensome a code complexity cost.

cc @pfaffe @bmeurer


tlively (approved):

kripken:
It turns out it's not easy to add a test. Somehow the behavior of finalize is different on a text input vs binary, and it doesn't hit the bug in the proper way. After spending some time I'm not sure it's worth investigating fully as it's purely a testing issue, and I'm not familiar enough with the passive segment code to quickly debug it. Instead, this adds a binary testcase, which fails before the fix. A downside is that the testcase is not tiny.

sbc100 (approved):
lgtm either way... did the test work out in the end?

kripken:
The binary testcase that is added works properly, yes. It is just annoying it isn't tiny.

sbc100:
> `clang-tidy` complains about the `cpp` source file I added for the testcase (it compiles with emcc and uses a header that exists only there, and not in this project). Should I make it ignore this file?

I guess so .. or just override it and land it as is.

tlively:
Yeah, that seems reasonable 👍

kripken:
Well, that commit to ignore a file in clang-tidy doesn't work... and reading around I can't find another way. If no one has any ideas I'll commit it as is, and the idea is that later PRs don't look at that diff, so they are not affected?

sbc100:
> Well, that commit to ignore a file in clang-tidy doesn't work... and reading around I can't find another way. If no one has any ideas I'll commit it as is, and the idea is that later PRs don't look at that diff, so they are not affected?

Yes, we are not 100% clang-tidy clean I believe which is why we only check deltas during PR submittions, so it should be fine to override clang tidy and force submit a PR where you disagree with it (.. or it doesn't work)

PR #3717, opened 2021-03-23
Equirecursive type canonicalization

tlively:
Use Hopcroft's DFA minimization algorithm to properly canonicalize and
deduplicate recursive types. Type canonicalization has two stages:

1. Structural canonicalization

   - The top-level structure of HeapTypes is used to split the declared HeapTypes
   into their initial partitions.

   - Hopcroft's algorithm refines the partitions such that all pairs of
   distinguishable types end up in different partitions.

   - A fresh HeapTypeInfo is created for each final partition. Each new
   HeapTypeInfo is linked to other new HeapTypeInfos to create a minimal type
   definition graph that defines the same types as the original graph.

2. Global canonicalization

   - Each new minimal HeapTypeInfo that does not have the same finite
   structure as an existing globally canonical HeapTypeInfo is moved to the
   global heap type store to become the new canonical HeapTypeInfo.

   - Each temporary Type referenced by the newly canonical HeapTypeInfos is
   replaced in-place with the equivalent canonical Type to avoid leaking
   temporary Types into the global stores.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -33,6 +33,8 @@ namespace {
 
 struct TypeInfo {
   using type_t = Type;
+  // Used in assertions to ensure that temporary types don't leak into the
+  // global store.
   bool isTemp = false;
   enum Kind {
     TupleKind,
@@ -69,8 +71,14 @@ struct TypeInfo {
 
 struct HeapTypeInfo {
   using type_t = HeapType;
+  // Used in assertions to ensure that temporary types don't leak into the
+  // global store.
   bool isTemp = false;
-  bool isSelfReferential = false;
+  // If `isFinalized`, then hashing and equality are performed on the finite
+  // structure of the type definition tree rooted at the HeapTypeInfo.
+  // Otherwise, the type definition tree is still being constructed so hashing
+  // and equality use pointer identity.
+  bool isFinalized = true;
   enum Kind {
     SignatureKind,
     StructKind,
@@ -155,18 +163,54 @@ struct TypePrinter {
   std::ostream& print(const Rtt& rtt);
 
 private:
-  template<typename T, typename F> std::ostream& printChild(T curr, F printer) {
-    auto it = depths.find(curr.getID());
-    if (it != depths.end()) {
-      assert(it->second <= currDepth);
-      size_t relativeDepth = currDepth - it->second;
-      return os << "..." << relativeDepth;
-    }
-    depths[curr.getID()] = ++currDepth;
-    printer();
-    depths.erase(curr.getID());
-    return os;
-  }
+  template<typename T, typename F> std::ostream& printChild(T curr, F printer);
+};
+
+// Helper for hashing the structures of TypeInfos and HeapTypeInfos. Uses de
+// Bruijn indices to represent back edges so that infos referring to different
+// type IDs but sharing a finite structure will compare and hash the same.
+struct TypeStructureHasher {
+  bool topLevelOnly;
+  size_t currDepth = 0;
+  std::unordered_map<HeapType, size_t> depths;
+
+  TypeStructureHasher(bool topLevelOnly = false) : topLevelOnly(topLevelOnly) {}
+
+  size_t hash(Type type);
+  size_t hash(HeapType heapType);
+  size_t hash(const TypeInfo& info);
+  size_t hash(const HeapTypeInfo& info);
+  size_t hash(const Tuple& tuple);
+  size_t hash(const Field& field);
+  size_t hash(const Signature& sig);
+  size_t hash(const Struct& struct_);
+  size_t hash(const Array& array);
+  size_t hash(const Rtt& rtt);
+};
+
+// Helper for comparing the structures of TypeInfos and HeapTypeInfos for
+// equality. Like TypeStructureHasher, uses de Bruijn indices to represent back
+// edges. Note that this does not test for coinductive equality of the infinite
+// expansion of the type structure, but rather tests for equality of the finite
+// structure. If TypeStructureEquator reports that two type structures are
+// equal, TypeStructureHasher should produce the same hash for them.
+struct TypeStructureEquator {
+  bool topLevelOnly;
+  size_t currDepth = 0;
+  std::unordered_map<HeapType, size_t> depthsA, depthsB;
+
+  TypeStructureEquator(bool topLevelOnly = false) : topLevelOnly(topLevelOnly) {}
+
+  bool eq(Type a, Type b);
+  bool eq(HeapType a, HeapType b);
+  bool eq(const TypeInfo& a, const TypeInfo& b);
+  bool eq(const HeapTypeInfo& a, const HeapTypeInfo& b);
+  bool eq(const Tuple& a, const Tuple& b);
+  bool eq(const Field& a, const Field& b);
+  bool eq(const Signature& a, const Signature& b);
+  bool eq(const Struct& a, const Struct& b);
+  bool eq(const Array& a, const Array& b);
+  bool eq(const Rtt& a, const Rtt& b);
 };
 
 } // anonymous namespace
@@ -176,12 +220,31 @@ namespace std {
 
 template<> class hash<wasm::TypeInfo> {
 public:
-  size_t operator()(const wasm::TypeInfo&) const;
+  size_t operator()(const wasm::TypeInfo& info) const {
+    return wasm::TypeStructureHasher().hash(info);
+  }
 };
 
 template<> class hash<wasm::HeapTypeInfo> {
 public:
-  size_t operator()(const wasm::HeapTypeInfo&) const;
+  size_t operator()(const wasm::HeapTypeInfo& info) const {
+    return wasm::TypeStructureHasher().hash(info);
+  }
+};
+
+template<typename T> class hash<reference_wrapper<const T>> {
+public:
+  size_t operator()(const reference_wrapper<const T>& ref) const {
+    return hash<T>{}(ref.get());
+  }
+};
+
+template<typename T> class equal_to<reference_wrapper<const T>> {
+public:
+  bool operator()(const reference_wrapper<const T>& a,
+                  const reference_wrapper<const T>& b) const {
+    return equal_to<T>{}(a.get(), b.get());
+  }
 };
 
 } // namespace std
@@ -197,6 +260,10 @@ TypeInfo* getTypeInfo(Type type) {
 HeapTypeInfo* getHeapTypeInfo(HeapType ht) {
   assert(ht.isCompound());
   return (HeapTypeInfo*)ht.getID();
+}
+
+HeapType asHeapType(std::unique_ptr<HeapTypeInfo>& info) {
+  return HeapType(uintptr_t(info.get()));
 }
 
 Type markTemp(Type type) {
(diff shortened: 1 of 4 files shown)


kripken:
Great to see this!

Before getting to the code here: Does this pass the fuzzer? If so then I think it may be worth landing first and reviewing later. I've investigated the fuzz bugs fallout from the LUB removal and not all of them are about subtyping (e.g. https://github.com/WebAssembly/binaryen/pull/3715 - it's hard to find how many more remain). That blocks an emscripten release and might affect `tot` users. So the current state of master is a little uncertain, and if landing this quickly helps there I think we should do that.

tlively:
Hmm, it's a little hard to tell about the fuzzer. When I run it, I get a bunch of errors about "Could not ensure supertype," which is fallout from LUB removal that is probably not related to this PR. I haven't seen any other errors, though, so at least this is more stable than the LUB removal. For example, my last run lasted 240 iterations before hitting the no-LUB bug. Does the fuzzer actually generate interesting (i.e. duplicate, recursive) type definitions, though?

How about we proceed as normal with review while I re-implement LUBbing on top of this, and when I finish that we can decide whether it makes sense to accelerate the landing of this PR and the LUB PR based on what the fuzzer thinks of the new LUBbing?

kripken:
Sounds good @tlively . I incorrectly thought this PR included LUB calculations.

The fuzzer won't generate complex types, correct.

src/wasm/wasm-type.cpp
```diff
@@ -260,24 +324,11 @@ TypeInfo::~TypeInfo() {
...
-    case TupleKind:
-      return tuple == other.tuple;
-    case RefKind:
-      return ref.heapType == other.ref.heapType &&
-             ref.nullable == other.ref.nullable;
-    case RttKind:
-      return rtt == other.rtt;
-  }
-  WASM_UNREACHABLE("unexpected kind");
+  return TypeStructureEquator().eq(*this, other);
```
kripken, line 327: Is the idea that it is ok to only compare the finite shape because this is after global canonicalization? (if so then a comment maybe could help)

src/wasm/wasm-type.cpp
```diff
@@ -1137,6 +1207,9 @@ std::ostream& TypePrinter::print(Type type) {
   }
 
   return printChild(type, [&]() {
+    if (isTemp(type)) {
```
kripken, line 1210: with these uses of `isTemp` I think we can remove the hacks on lines 278-281 and 287-290.

src/wasm/wasm-type.cpp
```diff
@@ -155,18 +163,55 @@ struct TypePrinter {
...
-  }
+  template<typename T, typename F> std::ostream& printChild(T curr, F printer);
+};
+
+// Helper for hashing the structures of TypeInfos and HeapTypeInfos. Uses de
+// Bruijn indices to represent back edges so that infos referring to different
+// type IDs but sharing a finite structure will compare and hash the same.
+struct TypeStructureHasher {
+  bool topLevelOnly;
+  size_t currDepth = 0;
```
kripken, line 174: It looks like currDepth is in the de Bruijn sense, and not in the tree sense, is that right? As there is no decrement for it. If so a comment could help as the same name is used in the printer where it is in the tree sense. Or maybe a different name? https://en.wikipedia.org/wiki/De_Bruijn_index doesn't use "depth", so perhaps `currIndex`?

The `erase` where a decrement would appear (line 1378) also implies this index in a particular subtree, is that right?

src/wasm/wasm-type.cpp
```diff
@@ -1270,6 +1346,222 @@ std::ostream& TypePrinter::print(const Rtt& rtt) {
...
+  }
+  return digest;
+}
+
+size_t TypeStructureHasher::hash(const Array& array) {
+  return hash(array.element);
+}
+
+size_t TypeStructureHasher::hash(const Rtt& rtt) {
+  size_t digest = rtt.depth;
```
kripken, line 1457:
```suggestion
  size_t digest = wasm::hash(rtt.depth);
```

kripken (commented):
(comments up to line 1644)

src/wasm/wasm-type.cpp
```diff
@@ -69,8 +71,14 @@ struct TypeInfo {
 
 struct HeapTypeInfo {
   using type_t = HeapType;
+  // Used in assertions to ensure that temporary types don't leak into the
+  // global store.
   bool isTemp = false;
-  bool isSelfReferential = false;
+  // If `isFinalized`, then hashing and equality are performed on the finite
+  // structure of the type definition tree rooted at the HeapTypeInfo.
```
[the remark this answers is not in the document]
tlively: No, it's after the user has stopped adding new types to the TypeBuilder.

src/wasm/wasm-type.cpp
tlively: Yeah, that was a bug I think. But this should be fixed by the latest change.

src/wasm/wasm-type.cpp
tlively: Not after global canonicalization, but after shape canonicalization. Will add a comment.

src/wasm/wasm-type.cpp
tlively: Oh hm, I thought I had already done that. Thanks for the catch!

src/wasm/wasm-type.cpp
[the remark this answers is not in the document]
kripken: I see - that wasn't obvious to me then. Perhaps `stillBuilding` or `stillAdding` would be clearer?

src/wasm/wasm-type.cpp
```diff
@@ -1347,13 +1635,304 @@ Type TypeBuilder::getTempRttType(size_t i, uint32_t depth) {
...
+        TypeSet& distinguished = partitions[distinguishedIndex];
+        TypeSet intersection = getIntersection(distinguished, currPreds);
+        if (intersection.empty()) {
+          continue;
+        }
+        TypeSet difference = getDifference(distinguished, currPreds);
+        if (difference.empty()) {
+          continue;
+        }
+        // We can split the partition! Replace it with the intersection and add
```
kripken, line 1754: this `!` sounds too happy to me... we just found more work we have to reluctantly do! :wink:

src/wasm/wasm-type.cpp
```diff
@@ -1347,13 +1635,304 @@ Type TypeBuilder::getTempRttType(size_t i, uint32_t depth) {
...
+  // HeapTypeInfos rather than the original HeapTypeInfos. This newly formed
+  // graph will have a shape coinductively equivalent to the original graph's
+  // shape, but each type definition will be minimal and distinct.
+  for (auto& partition : partitions) {
+    const auto& representative = *getHeapTypeInfo(*partition.begin());
+    infos.push_back(std::make_unique<HeapTypeInfo>(representative));
+    infos.back()->isTemp = true;
+  }
+  for (auto& info : infos) {
+    for (auto* child : getChildren(asHeapType(info))) {
```
kripken, line 1829: I don't understand this loop. Each input HeapType is in one of the `partitions`, represented by one of the `infos` (at the same index). Can't we traverse all the HeapTypes by traversing the partitions and then the partition's contents? Traversing the children I suppose will cover the same things in the end, but it seems like more work?

(then `getChildren` would only be used in the predecessor calculation, which is what I'd expect is the main place that actually looks at the shape details)

src/wasm/wasm-type.cpp
```diff
@@ -1436,42 +1989,46 @@ Canonicalizer::Canonicalizer(TypeBuilder& builder) : builder(builder) {
...
-      case Item::HeapTypeKind:
-        canonicalize(it.heapType, canonicalHeapTypes);
-        break;
+  // Canonicalize HeapTypes at all their use sites. HeapTypes for which there
+  // was not already a globally canonical version are moved to the global store
+  // to become the canonical version. These new canonical HeapTypes still
+  // contain references to temporary Types owned by the TypeBuilder, so we must
+  // subsequently replace those references with references to canonical Types.
+  // Canonicalize non-tuple types before tuple types to avoid canonicalizing a
+  // tuple that still contains non-canonical Types.
```
kripken, line 1998: Why are tuples special? Structs also contain a list of children just like a Tuple so the difficulty seems the same?

src/wasm/wasm-type.cpp
tlively: Hahaha should I make it `// We must bid this partition farewell 😢`

src/wasm/wasm-type.cpp
tlively: We don't just need the HeapTypes; we need the HeapType _locations_ where they are reachable from the other HeapTypes so we can replace their uses in-place with uses of canonical HeapTypes.

src/wasm/wasm-type.cpp
tlively: Tuples are the only types that directly contain other types. Structs are HeapTypes that contain types, but we've already canonicalized HeapTypes.

src/wasm/wasm-type.cpp
kripken: I see, thanks. Makes sense. I guess that's already explained in the comment above but I missed it.

src/wasm/wasm-type.cpp
kripken: Oh - that is rather surprising that such a difference causes a need for special handling here, but it seems unavoidable.

src/wasm/wasm-type.cpp
```diff
@@ -155,18 +163,57 @@ struct TypePrinter {
...
+  size_t hash(const Rtt& rtt);
+};
+
+// Helper for comparing the shapes of TypeInfos and HeapTypeInfos for equality.
+// Like FiniteShapeHasher, keeps track of previously seen HeapTypes. Note that
+// this does not test for coinductive equality of the infinite expansion of the
+// type tree, but rather tests for equality of the finite shape of the graph. If
+// FiniteShapeEquator reports that two type shapes are equal, FiniteShapeHasher
+// should produce the same hash for them.
+struct FiniteShapeEquator {
```
kripken, line 199: We have `Comparer` and `Comparator` in other places, but not sure those are better than `Equator`. But it would be good to standardize on the best at some point.

kripken (approved):
lgtm. I have no confidence in my ability to find all possible bugs in this code :smile:  but everything makes sense - nice work! - and more testing over time should iron out issues.

In addition to other review comments I'd suggest checking this doesn't at least regress us on fuzzability before landing.

src/wasm/wasm-type.cpp
tlively: This used to be handled by performing a topological sort on all the types and heap types, but now that none of the heap types need to participate in that topological sort, the sort seemed like overkill compared to just deferring the canonicalization of tuples.

src/wasm/wasm-type.cpp
tlively: Yeah, makes sense. In this file we also have a `TypeComparator`, but that implements `<` rather than `==`, so I thought a different name would be good for making that difference clear.

tlively:
Thanks for the review! My fuzzing last night didn't turn up any issues beyond the LUB removal fallout, so I'll land this now. The new LUB implementation should be ready soon.

PR #3718, opened 2021-03-23
Reducer: skip more functions when failing to remove them

kripken:
This avoids an annoying case where in each iteration we try to remove
every function one by one and keep failing. Instead, we'll skip large
numbers of them when the factor is large at least.

Also shorten some unnecessary logging.

### DIFF
--- a/src/tools/wasm-reduce.cpp
+++ b/src/tools/wasm-reduce.cpp
@@ -303,8 +303,7 @@ struct Reducer
             // see if it is still has the property we are preserving
             if (ProgramResult(command) == expected) {
               std::cerr << "|    command \"" << currCommand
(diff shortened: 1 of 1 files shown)


tlively (approved):
So if I understand correctly, this will only do anything different when the factor gets above 100. Does the factor ever come back down? Also, would it be good to have an element of randomness here so it's not the same functions being skipped every time?

kripken:
The factor only goes down. Yes, as you said under 100 this PR has no change, so we still test every function when we are close to finishing the reduction. Only in early stages with a large factor do we skip over.

A pseudo-random skip would make sense in principle, yes. Currently the factor is kind of like that - we almost never keep the same factor, we always decrease it by at least a *little*. So it should not be too repetitive here I hope...

PR #3719, opened 2021-03-23
 [Wasm GC] Optimize br_on_*

kripken:
The type may prove the value is not null, and may also show it to be
of the type we are casting to. In that case, we can simplify things.

This kind of thing can happen after inlining for example.

### DIFF
--- a/test/passes/remove-unused-brs_all-features.wast
+++ b/test/passes/remove-unused-brs_all-features.wast
@@ -1,6 +1,7 @@
 (module
  (type $vector (array (mut i32)))
  (type $struct (struct (field (ref null $vector))))
+ (import "out" "log" (func $log (param i32)))
  (func $foo (result (ref null $struct))
   (if (result (ref null $struct))
    (i32.const 1)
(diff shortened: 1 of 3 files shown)


test/passes/remove-unused-brs_all-features.wast
```diff
@@ -53,4 +54,58 @@
...
  )
+
+ (func $br_on-to-br (param $func (ref func))
+  (call $log (i32.const 0))
+  (block $null
+   ;; a non-null reference is not null, and the br is never taken
+   (drop
+    (br_on_null $null (ref.func $br_on-to-br))
+   )
+   (call $log (i32.const 1))
```
aheejin, line 65: Is this `call $log` for debugging or they have a purpose here?

aheejin (approved):

test/passes/remove-unused-brs_all-features.wast
kripken: Mainly just to make it easy to debug and read the testcase, to easily see what matches what.

PR #3720, opened 2021-03-24
[Wasm GC] Handle non-nullable locals in Flatten pass

kripken:
That pass adds lots of new locals, and we need to handle non-nullable ones.

### DIFF
--- a/test/passes/flatten_all-features.txt
+++ b/test/passes/flatten_all-features.txt
@@ -2443,3 +2443,20 @@
   (unreachable)
  )
 )
+(module
+ (type $none_=>_none (func))
+ (type $none_=>_funcref (func (result funcref)))
+ (func $0 (result funcref)
+  (local $0 (ref null $none_=>_none))
+  (local.set $0
+   (ref.as_non_null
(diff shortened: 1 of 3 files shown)


test/passes/flatten_all-features.txt
```diff
@@ -2443,3 +2443,20 @@
   (unreachable)
  )
 )
+(module
+ (type $none_=>_none (func))
+ (type $none_=>_funcref (func (result funcref)))
+ (func $0 (result funcref)
+  (local $0 (ref null $none_=>_none))
+  (local.set $0
+   (ref.as_non_null
```
aheejin, line 2452: I wonder why was this `ref.as_non_null` generated? I thought they are generated only when we do `local.get`.

kripken: I think this `ref.as_non_null` was already present in the input. It's only the later one that is added.

aheejin: Oh right, sorry...

aheejin (approved):

PR #3721, opened 2021-03-24
[Wasm GC] Fix canMakeZero on tuples

kripken:
We can't make a zero tuple if any field cannot be a zero.

### DIFF
--- a/src/ir/literal-utils.h
+++ b/src/ir/literal-utils.h
@@ -32,9 +32,22 @@ inline Expression* makeFromInt32(int32_t x, Type type, Module& wasm) {
 }
 
 inline bool canMakeZero(Type type) {
-  // We can make a "zero" - a 0, or a null, or a trivial rtt, etc. - for pretty
-  // much anything except a non-nullable reference type.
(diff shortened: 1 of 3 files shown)


src/ir/literal-utils.h
```diff
@@ -32,9 +32,22 @@ inline Expression* makeFromInt32(int32_t x, Type type, Module& wasm) {
...
+      if (!canMakeZero(t)) {
+        return false;
+      }
+    }
+  }
+  if (type.isFunction()) {
+    // Making a "zero" function is possible, but requires global knowledge - a
+    // function, and of the right type - so skip that.
+    return false;
+  }
```
tlively, line 49: Returning ref.null would be a reasonable default for nullable function references, right?

aheejin (approved):

src/ir/literal-utils.h
kripken: Oh, right... clearly I should not have written this PR that late in the day :rofl: not sure how I forgot about `ref.null`.

That change is just not needed. This PR now only adds tuple checking.

PR #3724, opened 2021-03-24
[Wasm GC] Optimize BrOn* of the wrong kind

kripken:
#3719 optimized the case where the kind is what we want, like `br_on_func` of
a function. This handles the opposite case, where we know the kind is wrong, and
so the break is not taken.

This seems equally useful for "polymorphic" code that does a bunch of checks
and routes to different code for each case, as after inlining or other optimizations
we may see which paths are taken and which are not.

Also refactor the checking code to a shared location as RefIs/As will also want to
use it.


src/passes/RemoveUnusedBrs.cpp
```diff
@@ -386,51 +387,38 @@ struct RemoveUnusedBrs : public WalkerPass<PostWalker<RemoveUnusedBrs>> {
...
 
-    // If the type provides enough information we may be able to know if this
-    // br is taken or not. If so, the br_on* may be unneeded. First, check for a
-    // possible null which would prevent such an optimization.
+    // First, check for a possible null which would prevent all other
+    // optimizations.
+    // (Note: if the spec had BrOnNonNull, instead of BrOnNull, then we could
+    // replace a br_on_func whose input is (ref null func) with br_on_non_null,
+    // as only the null check would be needed. But as things are, we cannot do
+    // such a thing.)
```
aheejin, line 395: Aren't both of them (`br_on_func` and hypothetical `br_on_non_null`) conditional branches anyway? Is there a benefit in converting one to the other?

aheejin (approved):

src/passes/RemoveUnusedBrs.cpp
kripken: Both are conditional branches, yes, but in some VMs a null check might not need to load the value (it can just see the pointer is null). That might not be most VMs, though... but even in other VMs a null check is more common, and so perhaps more likely to be folded away in the optimizing compiler (if there's another null check before it anyhow).

A definite advantage though is that GC operations have a byte prefix. `br_on_null` and a hypothetical `br_on_non_null` could be one byte, while `br_on_func` is two.

PR #3725, opened 2021-03-24
[Wasm GC] Optimize RefIs and RefAs when the type lets us

kripken:
This is similar to the optimization of BrOn in #3719 and #3724. When the
type tells us the kind of input we have, we can tell at compile time what
result we'll get, like `ref.is_func` of something with type `(ref func)` will
always return 1, etc.

Those previous PRs optimized things that look obvious in terms of what
Binaryen should be doing, but this PR is a little less clear to me (especially
after @aheejin 's comments in #3724 that made me realize this is less
trivial than it seems), so cc'ing some VM people for thoughts:

@jakobkummerow @manoskouk @lars-t-hansen @eqrion

The main open question is the second TODO in the code, which is
basically: Is `ref.is_func` as fast as `ref.is_null`? If the latter is a null
check then maybe not. And what about `ref.is_func` vs the combination
of `ref.is_null` + `i32.eqz`?


src/passes/OptimizeInstructions.cpp
```diff
@@ -1005,6 +1006,98 @@ struct OptimizeInstructions
...
 
+  void visitRefIs(RefIs* curr) {
+    if (curr->type == Type::unreachable) {
+      return;
+    }
+
+    // Optimizating RefIs is not that obvious, since even if we know the result
+    // evaluates to 0 or 1 then the replacement may not actually save code size,
+    // since RefIsNull is a single byte (the others are 2), while adding a Const
+    // of 0 would be two bytes. Other factors are that we can remove the input
```
aheejin, line 1017: Does this include `drop` in the calculation too? For example, in `RefIsNull` case, we save one byte by removing `ref.is_null`, add one byte by `drop`, and add two more bytes for `i32.const x`, so isn't that net +2 bytes gain? (+1 byte gain for other `ref.is_***` instructions)

src/passes/OptimizeInstructions.cpp
```diff
@@ -1005,6 +1006,98 @@ struct OptimizeInstructions
...
+
+    if (result == GCTypeUtils::Success) {
+      // We know the kind is correct, so all that is left is a check for
+      // non-nullability, which we do lower down.
+      curr->op = RefAsNonNull;
+    } else if (result == GCTypeUtils::Failure) {
+      // This is the wrong kind, so it will trap. The binaryen optimizer does
+      // not differentiate traps, so we can perform a replacement here. We
+      // replace 2 bytes of ref.as_* with one byte of unreachable and one of a
+      // drop, and the drop can be optimized out later, so this saves both work
```
aheejin, line 1088: > and the drop can be optimized out later

Isn't this the case only when the value does not have a side effect?

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -90,4 +90,331 @@
...
+   (drop
+    (ref.is_func (local.get $func))
+   )
+   (drop
+    (ref.is_data (local.get $data))
+   )
+   (drop
+    (ref.is_i31 (local.get $i31))
+   )
+  )
```
aheejin, line 196:
```suggestion
  )
  
```

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -90,4 +90,331 @@
...
+   (drop
+    (ref.as_func (local.get $func))
+   )
+   (drop
+    (ref.as_data (local.get $data))
+   )
+   (drop
+    (ref.as_i31 (local.get $i31))
+   )
+  )
```
aheejin, line 355:
```suggestion
  )
  
```

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -90,4 +90,331 @@
...
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (block (result i32)
+  ;; CHECK-NEXT:    (drop
+  ;; CHECK-NEXT:     (local.get $i31)
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:    (i32.const 1)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $unneeded_is
```
aheejin, line 130: Existing tests in this file uses 2-space indent, so better be consistent? (I know 1-space and 2-space indents are mixed in many of our tests, but I guess it'd be good to be consistent for new tests)

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -90,4 +90,331 @@
...
+  ;; CHECK-NEXT:     (ref.as_non_null
+  ;; CHECK-NEXT:      (local.get $func)
+  ;; CHECK-NEXT:     )
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:    (i32.const 0)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $unneeded_is_bad_kinds
+   (param $func (ref null func))
```
aheejin, line 256:
```suggestion
   (param $func (ref func))
```
Make one of the params non-nullable to test both nullable and non-nullable cases?

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -90,4 +90,331 @@
...
+  ;; CHECK-NEXT:    (drop
+  ;; CHECK-NEXT:     (local.get $func)
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:    (unreachable)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $unneeded_as_bad_kinds
+   (param $func (ref null func))
+   (param $data (ref null data))
```
aheejin, line 386:
```suggestion
   (param $data (ref data))
```
Make one of the params non-nullable to test both nullable and non-nullable cases?

aheejin (commented):
By the way did the VM people respond to your questions on the efficiency of the instructions?

askeksa-google:
In TurboFan at least, any outermost `i32.eqz` in the condition of an `if` or `br_if` are folded into the branch condition, so they are essentially free.

src/passes/OptimizeInstructions.cpp
kripken: I should have mentioned the drop too, yes. In most cases the input value has no side effects so both it and the drop for it can be removed.

src/passes/OptimizeInstructions.cpp
kripken: Correct, I'll fix the comment.

test/lit/passes/optimize-instructions-gc.wast
kripken: Good point, I'll fix the whitespace.

test/lit/passes/optimize-instructions-gc.wast
kripken: That is tested lower down, we do `ref.as_non_null` on the nullable params to get non-nullable inputs. I'll add a comment to make it more obvious to the eye.

test/lit/passes/optimize-instructions-gc.wast
kripken: Good catch. I'll add three more testcases below to check those, and a comment.

kripken:
Thanks @askeksa-google ! Makes sense. And binaryen should remove an `i32.eqz` in those cases, so any left over would need to be actually executed. What's not obvious is how common that is, I guess.

@aheejin That's the only response so far from people familiar with VMs. People are busy I guess...

If we don't get more detailed responses I still think this is a good direction for now. But we should benchmark and evaluate these more later.

jakobkummerow:
> Is `ref.is_func` as fast as `ref.is_null`?

I would assume that `ref.is_null` is faster, though as always it's difficult to say by how much.

In our current implementation, a null check is a pointer comparison against a known constant (which in turn needs to be loaded or materialized, the cost of which depends on how long it's been since the last null check). A func check involves two chained loads (object -> shape descriptor -> type). In fact, our current implementation of `is_func` initially performs an `is_null` check as a fast path; however that's technically not necessary and we should probably measure at some point whether it makes more sense to keep or to drop that fast path.

kripken:
I see, thanks @jakobkummerow - very good to know the details here.

Sounds like this PR is in the right direction then. And we can evaluate more in detail as we get benchmarks.

PR #3726, opened 2021-03-24
[Wasm GC] Properly validate struct.new number of operands

kripken:
I almost fixed this recently, but forgot that a validation error is not an
early exit...

I wonder if maybe it should be. That would mean more usage of
exceptions in our codebase, and that we'd stop at the first validation
error in all cases, but it would be simpler...

### DIFF
--- a/test/spec/struct.wast
+++ b/test/spec/struct.wast
@@ -105,3 +105,29 @@
   )
   "invalid rtt"
 )
+
+(assert_invalid
+  (module
+    (type $vec (struct (field (ref func))))
+    (func $test
+      (drop
+        ;; too many arguments
+        (struct.new_with_rtt $vec (i32.const 1) (i32.const 2) (rtt.canon $vec))
+      )
+    )
+  )
+  "invalid numbe of arguments to struct.new"
+)
+
+(assert_invalid
+  (module
+    (type $vec (struct (field (ref func)) (field f64)))
+    (func $test
+      (drop
+        ;; too few arguments
+        (struct.new_with_rtt $vec (i32.const 1) (rtt.canon $vec))
+      )
+    )
+  )
+  "invalid numbe of arguments to struct.new"
+)
--- a/test/spec/struct.wast
+++ b/test/spec/struct.wast
@@ -108,7 +108,7 @@
 
 (assert_invalid
   (module
-    (type $vec (struct (field (ref func))))
+    (type $vec (struct (field i32)))
     (func $test
       (drop
         ;; too many arguments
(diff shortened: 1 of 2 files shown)


aheejin (approved):

test/spec/struct.wast
```diff
@@ -105,3 +105,29 @@
...
+  (module
+    (type $vec (struct (field (ref func))))
+    (func $test
+      (drop
+        ;; too many arguments
+        (struct.new_with_rtt $vec (i32.const 1) (i32.const 2) (rtt.canon $vec))
+      )
+    )
+  )
+  "invalid numbe of arguments to struct.new"
```
aheejin, line 119: numbe -> number.. but the message seems to be different anyway? The code says "struct.new must have the right number of operands". The same for the below.

test/spec/struct.wast
```diff
@@ -105,3 +105,29 @@
   )
   "invalid rtt"
 )
+
+(assert_invalid
+  (module
+    (type $vec (struct (field (ref func))))
```
aheejin, line 111: Would it be better to make this field `i32`, if we are using `i32`s for tests below? So that we can only test the number of fields, not the type of fields.

test/spec/struct.wast
```diff
@@ -105,3 +105,29 @@
...
+  "invalid numbe of arguments to struct.new"
+)
+
+(assert_invalid
+  (module
+    (type $vec (struct (field (ref func)) (field f64)))
+    (func $test
+      (drop
+        ;; too few arguments
+        (struct.new_with_rtt $vec (i32.const 1) (rtt.canon $vec))
```
aheejin, line 128: The number of fields seems to be the same? But the type looks different.

test/spec/struct.wast
kripken: Yeah, these text fields are not validated, just that there is an error. One issue is that the text usually comes from the spec tests and the binaryen text might differ (maybe not for a good reason?)

test/spec/struct.wast
kripken: Good point, will fix.

test/spec/struct.wast
kripken: The RTT field is not counted. So there is just one argument to the struct here.

tlively (approved):

test/spec/struct.wast
aheejin: Yeah I didn't count that either. But the `$vec` struct has one field and this has one field too. The test above has two fields and it has too many. And this has one field and this is too few?

aheejin: Oh sorry, this time `$vec` is different... I only looked at `$vec` in the first test. Sorry nevermind.

PR #3727, opened 2021-03-24
Validator: Pass the module along when printing errors, so type names are used

kripken:
For example, on this invalid wat:
```wat
(module
  (type $vec (struct (field i64)))
  (func $test
    (drop
      (struct.new_with_rtt $vec (i32.const 1) (rtt.canon $vec))
    )
  )
)
```
We used to print:
```
[wasm-validator error in function test] struct.new operand must have proper type, on 
(struct.new_with_rtt ${i64}
 (i32.const 1)
 (rtt.canon ${i64})
)
```
We will now print:
```
[wasm-validator error in function test] struct.new operand must have proper type, on 
(struct.new_with_rtt $vec
 (i32.const 1)
 (rtt.canon $vec)
)
```

Note that `$vec` is used. In real-world examples the autogenerated structural name
can be huge, which this avoids.

### DIFF
--- a/src/passes/Print.cpp
+++ b/src/passes/Print.cpp
@@ -31,7 +31,8 @@ namespace wasm {
 static std::ostream& printExpression(Expression* expression,
                                      std::ostream& o,
                                      bool minify = false,
-                                     bool full = false);
+                                     bool full = false,
+                                     Module* wasm = nullptr);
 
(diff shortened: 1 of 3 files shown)


aheejin (approved):

tlively (approved):
Nice solution!

PR #3728, opened 2021-03-24
Refactor TypeBuilder

tlively:
Makes TypeBuilders growable, adds a `getTempHeapType` method, allows the
`getTemp*Type` methods to take arbitrary temporary or canonical HeapTypes rather
than just an index, and allows BasicHeapTypes to be assigned to TypeBuilder
slots. All of these changes are necessary for the upcoming re-implementation of
equirecursive LUB calculation.

Also adds a new utility to TypeBuilder for using `operator[]` as an intuitive
and readable wrapper around the `getTempHeapType` and `setHeapType` methods.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -80,16 +80,19 @@ struct HeapTypeInfo {
   // TypeBuilder interface, so hashing and equality use pointer identity.
   bool isFinalized = true;
   enum Kind {
+    BasicKind,
     SignatureKind,
     StructKind,
     ArrayKind,
   } kind;
   union {
+    HeapType::BasicHeapType basic;
     Signature signature;
     Struct struct_;
     Array array;
   };
 
+  HeapTypeInfo(HeapType::BasicHeapType basic) : kind(BasicKind), basic(basic) {}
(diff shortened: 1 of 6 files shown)


kripken (approved):
The `[]` notation is indeed nicer...

The `BasicKind` addition is a little verbose but I can see why it's worthwhile.

tlively:
Hmm, the clang-tidy error here is just wrong afaict. @aheejin, @sbc100 is there a good way to make CI ignore it?

src/wasm/wasm-type.cpp
```diff
@@ -338,6 +344,9 @@ HeapTypeInfo::HeapTypeInfo(const HeapTypeInfo& other) {
 
 HeapTypeInfo::~HeapTypeInfo() {
   switch (kind) {
+    case BasicKind:
+      basic.HeapType::~BasicHeapType();
```
sbc100, line 348: Does it not want to you write `basic.~HeapType::BasicHeapType()`?   Would that work?

tlively: No, unfortunately that doesn't compile at all. I just ended up getting rid of this destructor call, since it wasn't doing anything anyway 🤷‍♂️

PR #3731, opened 2021-03-25
LUBs

tlively:
This is a partial revert of #3669, which removed the old implementation of
Type::getLeastUpperBound that did not correctly handle recursive types. The new
implementation in this PR uses a TypeBuilder to construct LUBs and for recursive
types, it returns a temporary HeapType that has not yet been fully constructed
to break what would otherwise be infinite recursions.


kripken:
Please add a test for `remove-unused-brs` with this:
```wat
 (func $selectify-no-type (param $x i32) (result funcref)
  ;; the type is not present on the if, which means we must get
  ;; it in a LUB calculation
  (if
   (local.get $x)
   (return
    (ref.func $none_=>_i32)
   )
   (return
    (ref.func $i32_=>_none)
   )
  )
 )
```
That would have caught the regression from earlier and shown us how tricky it is.

kripken:
The fuzzer itself hits a problem on these seeds: `18152240981383018426`, `16315457888497641454`. Those might be regressions in this PR as they are if-shape related and don't look familiar to me.

kripken:
Example of those:
```wat
[wasm-validator error in function func_152] returning if-else's true must have right type, on 
(if (result (ref null $none_=>_anyref))
 (i32.eqz
  (i32.trunc_sat_f64_u
   (f64.const -nan:0xfffffffea5e59)
  )
 )
 (ref.func $func_39)
 (block $label$5 (result (ref null $none_=>_funcref))
  (call $log-i32
   (ref.is_null
    (ref.null func)
   )
  )
  (ref.func $func_16)
 )
)
```

src/wasm-type.h
```diff
@@ -423,6 +416,7 @@ struct Field {
   } packedType; // applicable iff type=i32
   Mutability mutable_;
 
+  Field() = default;
```
kripken, line 419: why?

src/wasm/wasm-type.cpp
```diff
@@ -139,13 +139,34 @@ struct SubTyper {
...
+  TypeBuilder builder;
+  // The indices in `builder` at which the LUB of each pair of HeapTypes are
+  // being constructed.
+  std::unordered_map<std::pair<HeapType, HeapType>, size_t> indices;
+
+  bool hasLeastUpperBound(Type a, Type b);
+  Type getLeastUpperBound(Type a, Type b);
+
+  // Return true and set `out` to be the LUB iff a LUB was found. The HeapType
+  // overload is exceptional because it is infallible.
```
kripken, line 159: why is it infallible? that seems surprising to me, am I missing something?

kripken (commented):
Could use a few more tests perhaps, like

* `LUB(funcref, (ref specific func type)) = funcref`
* `LUB(struct(i32, i64), struct(i32)) = struct(i32)`

and that there is no useful LUB between arrays and structs, structs and funcs, etc.

src/wasm/wasm-type.cpp
tlively: `HeapType::any` is an upper bound of any two HeapTypes, so there is always a lowest upper bound.

tlively:
I verified that the failing fuzz seeds you shared now pass. The problem was that the TypeBounder and SubTyper disagreed on how signature subtyping worked. I had forgotten that V8 does not support any signature subtyping at all, not just no contravariance in parameters.

src/wasm/wasm-type.cpp
tlively: Similarly, I realized that the LUB of structs is similarly infallible because the empty struct is a supertype of all other structs.

src/wasm-type.h
tlively: So I can write this without having to assign some arbitrary value first, which would reduce readability:

```
    Field field;
    if (lub(a.fields[i], b.fields[i], field)) {
      ...
```

src/ir/stack-utils.cpp
```diff
@@ -164,6 +164,95 @@ bool StackSignature::isSubType(StackSignature a, StackSignature b) {
                     });
 }
 
+bool StackSignature::haveLeastUpperBound(StackSignature a, StackSignature b) {
```
tlively, line 167: Everything in this file and in the stack-utils tests is purely revert from the LUB removal and can be safely ignored.

src/wasm/wasm-type.cpp
```diff
@@ -139,13 +139,34 @@ struct SubTyper {
...
+  TypeBuilder builder;
+  // The indices in `builder` at which the LUB of each pair of HeapTypes are
+  // being constructed.
+  std::unordered_map<std::pair<HeapType, HeapType>, size_t> indices;
+
+  bool hasLeastUpperBound(Type a, Type b);
+  Type getLeastUpperBound(Type a, Type b);
+
+  // Return true and set `out` to be the LUB iff a LUB was found. The HeapType
+  // and Struct overloads are exceptional because they are infallible.
```
kripken, line 159: I'd expand this comment to add something like "because `anyref` is a supertype of all HeapTypes etc." as you mentioned

src/wasm-type.h
kripken: Makes sense, but it worries me a little. Does the C++ spec guarantee that the enums here will be initialized to their first value, or do we need an assignment for them?

src/wasm/wasm-type.cpp
```diff
@@ -1180,6 +1209,220 @@ bool SubTyper::isSubType(const Rtt& a, const Rtt& b) {
...
+  Type tempLUB;
+  if (!lub(a, b, tempLUB)) {
+    return Type::none;
+  }
+  // `tempLUB` is a temporary type owned by `builder`. Since TypeBuilder::build
+  // returns HeapTypes rather than Types, create a new HeapType definition meant
+  // only to get `tempLUB` canonicalized in a known location.
+  builder.grow(1);
+  builder[builder.size() - 1] = Array(Field(tempLUB, Mutable));
+  return builder.build().back().getArray().element.type;
```
kripken, line 1227: that this refers to `Array` seems odd even with the comment - maybe the comment could add `(the specific heap type does not matter; here we use an Array)`

src/wasm/wasm-type.cpp
```diff
@@ -1180,6 +1209,220 @@ bool SubTyper::isSubType(const Rtt& a, const Rtt& b) {
...
+  if (b == Type::unreachable) {
+    out = a;
+    return true;
+  }
+  if (a.isTuple() && b.isTuple()) {
+    Tuple tuple;
+    if (!lub(a.getTuple(), b.getTuple(), tuple)) {
+      return false;
+    }
+    out = builder.getTempTupleType(tuple);
```
kripken, line 1248: why is it ok to return temp types in all these places?

src/wasm-type.h
tlively: No, it leaves them uninitialized. I can add an explicit initializer so it at least has a valid value.

src/wasm/wasm-type.cpp
tlively: `lub` methods are only called by other `lub` methods, and by `TypeBounder::{get,has}LeastUpperBound`, none of which allow the temporary types to escape outside of the `TypeBounder` and its `TypeBuilder`.

tlively: I'll make the `lub` methods private to enforce that more.

src/wasm/wasm-type.cpp
```diff
@@ -1103,7 +1136,7 @@ bool SubTyper::isSubType(HeapType a, HeapType b) {
   }
   // Various things are subtypes of eq.
   if (b == HeapType::eq) {
-    return a == HeapType::i31 || a.isArray() || a.isStruct();
+    return a == HeapType::i31 || a == HeapType::data || a.isData();
```
kripken, line 1139:
```suggestion
    return a == HeapType::i31 || a.isData();
```

src/wasm/wasm-type.cpp
```diff
@@ -1180,6 +1213,221 @@ bool SubTyper::isSubType(const Rtt& a, const Rtt& b) {
...
+      case HeapType::any:
+        return HeapType::any;
+      case HeapType::eq:
+        if (b == HeapType::i31 || b.isData()) {
+          return HeapType::eq;
+        } else {
+          return HeapType::any;
+        }
+      case HeapType::i31:
+        if (b == HeapType::data || b.isData()) {
```
kripken, line 1299:
```suggestion
        if (b.isData()) {
```

src/wasm/wasm-type.cpp
```diff
@@ -1180,6 +1213,221 @@ bool SubTyper::isSubType(const Rtt& a, const Rtt& b) {
...
+  auto pair =
+    a.getID() < b.getID() ? std::make_pair(a, b) : std::make_pair(b, a);
+  size_t index = builder.size();
+  auto result = indices.insert({pair, index});
+  if (result.second) {
+    builder.grow(1);
+  } else {
+    // We've seen this pair before; stop recursing and do not allocate.
+    return builder[result.first->second];
+  }
```
kripken, line 1326:
```suggestion
  if (!result.second) {
    // We've seen this pair before; stop recursing and do not allocate.
    return builder[result.first->second];
  }
  builder.grow(1);
```

src/wasm/wasm-type.cpp
```diff
@@ -1180,6 +1213,221 @@ bool SubTyper::isSubType(const Rtt& a, const Rtt& b) {
...
+  // Allocate a new slot to construct the LUB of this pair if we have not
+  // already seen it before. Canonicalize the pair to have the element with the
+  // smaller ID first since order does not matter.
+  auto pair =
+    a.getID() < b.getID() ? std::make_pair(a, b) : std::make_pair(b, a);
+  size_t index = builder.size();
+  auto result = indices.insert({pair, index});
+  if (result.second) {
+    builder.grow(1);
```
kripken, line 1322: I'm confused. How can this not be a new pair every time? `index` is the size of the builder, which increases monotonically each time we arrive here.

kripken: Oh, right, I think I was confused by `insert` receiving a pair while the key is itself a pair...

kripken (approved):
Great!

kripken:
May be worth some fuzzing on this before landing, for the last changes (which I didn't check earlier in my large fuzzing).

tlively:
The fuzzer still finds some errors, but I haven't seen any that look related to this PR.

PR #3732, opened 2021-03-25
Flat IR: Allow ref.as_non_null in nested positions

kripken:
We can't disallow it, as its result is non-null which we can't spill to a local.

This may cause issues eventually in the combination of GC + flatten, but
I don't expect it to. If it does we may need to revisit.


test/lit/passes/opt_flatten.wast
```diff
@@ -0,0 +1,43 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
+
+;; Optimize the code, flatten, and run a pass that will verify it is properly
+;; flat.
+;; RUN: wasm-opt %s -all -O1 --flatten --rereloop -S -o - | filecheck %s
```
tlively, line 5: What does the -O1 do for the test? Would it be possible to start out directly at `--flatten`?

tlively (approved):

tlively:
If this does cause issues for flatten, one thing we might want to investigate is whether Poppy IR could be a suitable replacement for Flat IR. It essentially does a similar flattening, but without introducing new locals, so it wouldn't have a problem with non-nullable types.

test/lit/passes/opt_flatten.wast
kripken: The -O1 is necessary. I did not investigate in depth why, but I suspect it is something we add during binary reading that opts remove.

PR #3733, opened 2021-03-25
Fix nondefaultability of Tuple types

kripken:
This looks like a pre-existing issue to https://github.com/WebAssembly/binaryen/pull/3731, but
the testcase only fails on that PR for a reason I did not investigate in
depth, so it should land first.

Also fix the check for nondefaultability in the fuzzer.

### DIFF
--- a/test/spec/tuples.wast
+++ b/test/spec/tuples.wast
@@ -1,0 +1,9 @@
+(assert_invalid
+  (module
+    (func $foo
+      (local $temp ((ref func) i32))
+    )
+  )
+  "var must be defaultable"
+)
+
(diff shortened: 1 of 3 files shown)


test/spec/tuples.wast
```diff
@@ -0,0 +1,9 @@
+(assert_invalid
+  (module
+    (func $foo
+      (local $temp ((ref func) i32))
+    )
+  )
+  "var must be defaultable"
```
tlively, line 7: This would also be a good test to put in `test/lit/validation` if you want. (But LGTM either way.)

tlively (approved):

test/spec/tuples.wast
kripken: Interesting, thanks, I'll look at that later.

PR #3734, opened 2021-03-25
Fix binary reading of tuples containing non-nullable types

kripken:
We must write them to a tuple with nullable types, then fix that up when
reading. This is similar to what we do in handleNonNullableLocals, except
that it operates on the entire tuple type, so it can't share that code.

This fixes a regression from #3710 that was harder to notice by the fuzzer
until now.


test/lit/passes/roundtrip.wast
```diff
@@ -0,0 +1,46 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
+;; RUN: wasm-opt %s -all --roundtrip -S -o - | filecheck %s
```
tlively, line 2: Wouldn't we see the same result without the `--roundtrip`?

tlively (approved):

test/lit/passes/roundtrip.wast
kripken: No, the problem is in the binary reading code. The testcase is in text, so roundtrip is necessary to write and then read.

kripken: (The text format allows tuples in a direct way and does not lower then during text reading like binary does.)

PR #3735, opened 2021-03-25
[Wasm GC] Add a Name-Types pass

kripken:
The pass gives a simple short name to each type, `$type$N`. This can be
useful in debugging, to avoid the autogenerated names which can be very
long.

@tlively this may overlap with your intended work on naming types, but
this was a quick hack that helped me with debugging today, so maybe
it's worth committing. Also perhaps we can improve this code and use
it in the binary loader, to implement what you had in mind?


aheejin (approved):

tlively (approved):
There's some overlap, but I think that's fine. This pass could be updated later to produce more informative names after the default printing changes to using barebones indices.

PR #3738, opened 2021-03-25
Fix fuzzer on creating a function with a heaptype of just 'func'

kripken:
Without this the fuzzer calls `getSignature` on `HeapType::func` which asserts.

### DIFF
--- a/src/tools/fuzzing.h
+++ b/src/tools/fuzzing.h
@@ -2115,8 +2115,8 @@ private:
       if (heapType.isSignature()) {
         sig = heapType.getSignature();
       } else {
-        assert(heapType == HeapType::func);
-        // The specific signature does not matter.
+        // This is something like funcref or externref, and so the specific
+        // signature does not matter.
         sig = Signature(Type::none, Type::none);
       }
       auto* func = wasm.addFunction(builder.makeFunction(
--- a/src/tools/fuzzing.h
+++ b/src/tools/fuzzing.h
@@ -2118,6 +2118,9 @@ private:
         // This is something like funcref or externref, and so the specific
         // signature does not matter.
         sig = Signature(Type::none, Type::none);
+        // When we set the type of the ref.func, it must be funcref and not
+        // externref or such.
+        type = Type::funcref;
       }
       auto* func = wasm.addFunction(builder.makeFunction(
         Names::getValidFunctionName(wasm, "ref_func_target"),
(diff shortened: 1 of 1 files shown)


src/tools/fuzzing.h
```diff
@@ -2110,9 +2110,18 @@ class TranslateToFuzzReader {
         return builder.makeRefNull(type);
       }
       // Last resort: create a function.
+      auto heapType = type.getHeapType();
```
aheejin, line 2113: So is it guaranteed that `type.isFunction()` by the time we reach here? Can it be data? (Not `dataref`, because it was excluded above, but other data, like array)

If it is guaranteed to be a function somehow, how can its heaptype not be a signature?

src/tools/fuzzing.h
```diff
@@ -2115,8 +2115,8 @@ class TranslateToFuzzReader {
       if (heapType.isSignature()) {
         sig = heapType.getSignature();
       } else {
-        assert(heapType == HeapType::func);
-        // The specific signature does not matter.
+        // This is something like funcref or externref, and so the specific
+        // signature does not matter.
```
aheejin, line 2119: - Isn't `funcref` handled at the very top?
- If this is `externref`, isn't returning of a function an error? It is not a subtype of `externref`..
- Can this also be other data that is not `dataref`?

src/tools/fuzzing.h
kripken: You're right, the code above would let an externref fall through... which I forgot is not a supertype of function.

Should be fixed now.

The one case where a function is not a signature is funcref. That is a function, but has no signature info.

aheejin: Thanks. This is unrelated, but the comment in line 2080 refering to `func` seems outdated. It should be `funcContext`, right?

aheejin (approved):

src/tools/fuzzing.h
kripken: Yes, good catch, I'll update that.

src/tools/fuzzing.h
kripken: Good point that funcref is handled above. It is a little weird to handle `(ref func)` the non-nullable version separately lower down. This is probably worth refactoring, but for now I think it's good enough.

About dataref, is, `isData` can return true for struct and array types.

aheejin: > About dataref, `isData` can return true for struct and array types also.

I can't see uses of `isData` in this function, no?

By the way I don't mean we should use it here. You added `!type.isFunction()` and I think that covers the data case, in which we can't do much anyway.

kripken: Oh, sorry, I had `isData()` in an earlier version. Then I switched it to that `!isFunction()` so it also covers `externref`. Yes, I think that's good enough for the data case for now.

PR #3744, opened 2021-03-29
Scan module-level code in necessary places

kripken:
Several old passes like DeadArgumentElimination and DuplicateFunctionElimination
need to look at all `ref.func`s, and they scanned functions for that, but that is not
enough as such an instruction might appear in a global initializer. To fix this, add a
`walkModuleCode` method.

`walkModuleCode` is useful when doing the pattern of creating a function-parallel
pass to scan functions quickly, but we also want to do the same scanning of code
at the module level. This allows doing so in a single line.

(It is also possible to just do `walk()` on the entire module, which will find all code,
but that is not function-parallel. Perhaps we should have a `walkParallel()` option
to simplify this further in a followup, and that would call `walkModuleCode` afterwards
etc.)

Also add some missing validation and comments in the validator about issues that
I noticed in relation to the new testcases here.

### DIFF
--- a/src/wasm/wasm-validator.cpp
+++ b/src/wasm/wasm-validator.cpp
@@ -2851,7 +2851,7 @@ static void validateTables(Module& module, ValidationInfo& info) {
       info.shouldBeTrue(
         table != nullptr, "elem", "elem segment must have a valid table name");
       info.shouldBeTrue(!!segment->offset,
-                        segment,
+                        "elem",
                         "table segment offset should have an offset");
       info.shouldBeEqual(segment->offset->type,
                          Type(Type::i32),
@@ -2865,7 +2865,7 @@ static void validateTables(Module& module, ValidationInfo& info) {
       validator.validate(segment->offset);
     } else {
       info.shouldBeTrue(!segment->offset,
-                        segment,
+                        "elem",
                         "non-table segment offset should have no offset");
     }
     // Avoid double checking items
--- a/src/wasm/wasm-validator.cpp
+++ b/src/wasm/wasm-validator.cpp
@@ -2003,8 +2003,8 @@ void FunctionValidator::visitRefFunc(RefFunc* curr) {
   //   curr->type.getHeapType().getSignature()
   // That is blocked on having the ability to create signature types in the C
   // API (for now those users create the type with funcref). This also needs to
-  // be fixed in LegalizeJSInterface and other places that update function
-  // types.
+  // be fixed in LegalizeJSInterface and FuncCastEmulation and other places that
+  // update function types.
   // TODO: check for non-nullability
 }
 
--- a/src/wasm/wasm-validator.cpp
+++ b/src/wasm/wasm-validator.cpp
@@ -1995,19 +1995,8 @@ void FunctionValidator::visitRefFunc(RefFunc* curr) {
   }
   auto* func = getModule()->getFunctionOrNull(curr->func);
   shouldBeTrue(!!func, curr, "function argument of ref.func must exist");
-  if (shouldBeTrue(
-        curr->type.isFunction(), curr, "ref.func must have a function type")) {
-    // TODO: verify it also has a typed function references type,
-    //                    curr->type.getHeapType().isSignature(),
-    // That is blocked on having the ability to create signature types in the C
-    // API. For now those users create the type with funcref.
-    if (curr->type.getHeapType().isSignature()) {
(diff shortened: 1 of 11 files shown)


src/wasm/wasm-validator.cpp
```diff
@@ -2853,6 +2863,10 @@ static void validateTables(Module& module, ValidationInfo& info) {
                         segment->offset,
                         "table segment offset should be reasonable");
       validator.validate(segment->offset);
+    } else {
```
aheejin, line 2866: Maybe I'm not up-to-date with multiple tables stuff... In which case an element segment does not have a table? And why that shouldn't have an offset?

aheejin (approved):

src/wasm/wasm-validator.cpp
kripken: I think it might be the difference between active and passive segments, but I don't remember if that's the term for tables (and not just memories). @tlively and @martianboy would know better.

martianboy: Right, it's the same for memories and tables. Passive segments don't belong to a table on instantiation and hence have no offset either. They can be copied to a table using `table.init` instruction which is defined in [reference-types proposal](https://webassembly.github.io/reference-types/core/syntax/instructions.html#syntax-instr-table).

aheejin: Thanks. Do we have passive `elem` segment or `table.init` implemented in binaryen now? I searched with `passive` in test/ directory but that only turns up passive data segments.

martianboy: @aheejin table instructions aren't implemented yet, but we do have passive element segment support. You can take a look at this test: `test/multi-table.wast`.

aheejin: Thanks. So memory passive elements have `passive` keyword but passive data segments don't?

martianboy: No, there is no `passive` keyword. If a data segment has an offset then it's active, otherwise it's considered passive.

aheejin: Where do we have those text format definitions? The spec for [element segments](https://webassembly.github.io/reference-types/core/text/modules.html#element-segments) and [data segments](https://webassembly.github.io/reference-types/core/text/modules.html#data-segments) both have 'passive mode', but neither says it should be written in the text format.

martianboy: Not sure if I understand the question, are you asking whether there should be a keyword specifier? As per spec, if offset and an optional table index are present then it's active, if not it's passive. If `declare` is present in case of a passive element segment, it's considered a declarative segment.

aheejin: No what I'm asking is, 

> No, there is no `passive` keyword. If a data segment has an offset then it's active, otherwise it's considered passive.

You said there is a `passive` keyword in passive element segments but not in passive data segments. But the spec for two segments I linked above are identical.

martianboy: Oh sorry I think you misunderstood me. I was saying there is no `passive` keyword in either case. The spec is very similar in both cases.

aheejin: Hmm I see. I was confused because there are binaryen tests with `(data passive ...)` syntax, like this. https://github.com/WebAssembly/binaryen/blob/9b38e4394034d6f25f54fc471bb9ac8812815331/test/spec/bulk-memory.wast#L4

The `data passive` syntax appears in the [explainer](https://github.com/WebAssembly/bulk-memory-operations/blob/master/proposals/bulk-memory-operations/Overview.md#passive-segment-initialization-example) of the bulk memory proposal, but I guess that's not part of the final spec? The [spec repo](https://github.com/WebAssembly/spec) does not contain tests with that syntax either. I guess we can remove it? cc @tlively

martianboy: You're right, the parser seems to be outdated for data segments. I've also been keeping the "old-syntax" for tables and element segments, and I don't know if it's necessary.

kripken: In general I don't think we need to support text files outside of our test suite for stuff like this that is not in the upstream spec. Removing old syntax forms and updating our tests sounds good.

martianboy: I made two PRs (#3753 and #3757) to remove old table & elem syntax as well as the `passive` keyword for data segments. @tlively @aheejin

PR #3745, opened 2021-03-29
[Wasm GC] Fix SSA pass on non-nullable parameters

kripken:
If such a parameter is written to then we create a new local for each
such write, and must handle non-nullability of those new locals.


aheejin (approved):
Hmm, are there more passes that create locals? 
```
$ grep addVar * -R | sed -e 's/:.*//' | uniq
emscripten-optimizer/optimizer.h
ir/localize.h
passes/OptimizeAddedConstants.cpp
passes/MemoryPacking.cpp
passes/AvoidReinterprets.cpp
passes/ConstHoisting.cpp
passes/Inlining.cpp
passes/RemoveUnusedBrs.cpp
passes/Flatten.cpp
passes/AlignmentLowering.cpp
passes/I64ToI32Lowering.cpp
passes/ReReloop.cpp
passes/StackCheck.cpp
passes/SSAify.cpp
passes/LegalizeJSInterface.cpp
passes/DeadArgumentElimination.cpp
passes/Poppify.cpp
passes/Asyncify.cpp
wasm/wasm-binary.cpp
wasm2js.h
wasm-builder.h
```

Should we do something similar to these? (I guess some of them have been already handled though) Would it be better to give them some property like `modifyLocals` and run `handleNonNullableLocals` automatically for them?

kripken:
Some of those may need fixes, we'd need to check them each in detail. At least some don't (`I64ToI32Lowering`, `OptimizeAddedConstants`, `StackCheck`, I remember that they use integers).

That's an interesting idea to add a property of a pass for this. That could save code if we have many more. I hope there aren't, though, but I could be wrong... worth thinking about as a possible refactoring.

PR #3747, opened 2021-03-29
Fuzzing: Minor execution-results.h fixes

kripken:
* Log out an i64 as two i32s. That keeps the output consistent regardless of
  whether we legalize.
* Do not print a reference result. The printed value cannot be compared, as
  `funcref(10)` (where 10 is the index of the function) is not guaranteed to be
  the same after opts.
* Trap when trying to call an export with a nondefaultable type (instead of
  asserting when trying to create zeros for it).

### DIFF
--- a/src/tools/execution-results.h
+++ b/src/tools/execution-results.h
@@ -223,7 +223,8 @@ struct ExecutionResults {
       for (const auto& param : func->sig.params) {
         // zeros in arguments TODO: more?
         if (!param.isDefaultable()) {
-          std::cout << "[trap fuzzer can only send nullable parameters to exports]\n";
+          std::cout
+            << "[trap fuzzer can only send defaultable parameters to exports]\n";
         }
         arguments.push_back(Literal::makeZero(param));
       }
--- a/src/tools/execution-results.h
+++ b/src/tools/execution-results.h
@@ -44,8 +44,20 @@ struct LoggingExternalInterface : public ShellExternalInterface {
       std::cout << "[LoggingExternalInterface logging";
       loggings.push_back(Literal()); // buffer with a None between calls
       for (auto argument : arguments) {
-        std::cout << ' ' << argument;
-        loggings.push_back(argument);
+        if (argument.type == Type::i64) {
+          // To avoid JS legalization changing logging results, treat a logging
+          // of an i64 as two i32s (which is what legalization would turn us
+          // into).
+          auto low = Literal(int32_t(argument.getInteger()));
+          auto high = Literal(int32_t(argument.getInteger() >> int32_t(32)));
+          std::cout << ' ' << low;
+          loggings.push_back(low);
+          std::cout << ' ' << high;
+          loggings.push_back(high);
+        } else {
+          std::cout << ' ' << argument;
+          loggings.push_back(argument);
+        }
       }
       std::cout << "]\n";
       return {};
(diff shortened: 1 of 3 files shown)


src/tools/execution-results.h
```diff
@@ -97,8 +109,17 @@ struct ExecutionResults {
...
+            auto resultType = func->sig.results;
+            if (resultType.isRef()) {
+              // Don't print reference values, as funcref(N) contains an index
+              // for example, which is not guaranteed to remain identical after
+              // optimizations.
+              std::cout << resultType;
+            } else {
+              std::cout << ret;
+            }
+            std::cout << '\n';
```
aheejin, line 122:
```suggestion
              std::cout << resultType << '\n';
            } else {
              std::cout << ret << '\n';
            }
```

src/tools/execution-results.h
```diff
@@ -97,8 +109,17 @@ struct ExecutionResults {
...
+            std::cout << "[fuzz-exec] note result: " << exp->name << " => ";
+            auto resultType = func->sig.results;
+            if (resultType.isRef()) {
+              // Don't print reference values, as funcref(N) contains an index
+              // for example, which is not guaranteed to remain identical after
+              // optimizations.
+              std::cout << resultType;
+            } else {
+              std::cout << ret;
+            }
```
aheejin, line 121: Nit: We [used](https://github.com/WebAssembly/binaryen/pull/2451/files#diff-d2128595c03cac3cc010b8bbcf17973cc3583244dc6953a2755040825deb7082R73-R82) to do this here, which excluded reference types from both printing and comapring, but #3384 moved it to `areEqual`, which I think only handled comparison and not printing. Can we do that in the one place possibly?

src/tools/execution-results.h
```diff
@@ -202,6 +223,9 @@ struct ExecutionResults {
       // call the method
       for (const auto& param : func->sig.params) {
         // zeros in arguments TODO: more?
+        if (!param.isDefaultable()) {
+          std::cout << "[trap fuzzer cannot provide a non-null export param'\n";
```
aheejin, line 227: What is `export param`? You mean a param to an export function?

src/tools/execution-results.h
kripken: Hmm, good question but I don't see how to merge the two. The previous code handled comparing one run to another run inside the fuzzer. This code handles printing them out which allows comparing one run here to the results from another VM run in another process. I think we need both. And to unify them we'd need to have something like a result being not a Literal but either a Literal or a Type, and to vary over both...?

src/tools/execution-results.h
kripken: Yes, sorry, I did mean a param to an export. I'll clarify.

tlively (approved):

src/tools/execution-results.h
aheejin: I might be mistaken, but can't we execute this part only when the type is not a reference type? Which means, 1. We don't print it 2. We don't assign it to `ret`. Because we don't assign it to `ret`, we don't need to compare it later.
https://github.com/WebAssembly/binaryen/blob/cdac6b123b1231859dc0dfbcb88dfabe005855e3/src/tools/execution-results.h#L97-L102

kripken: That could work, but a downside is that then we wouldn't compare it at all - that is, we wouldn't even compare the types?

aheejin: Hmm, you're right.

aheejin (approved):

PR #3748, opened 2021-03-29
[Wasm GC] Optimize RefCast + ignore-implicit-traps

kripken:
If we are ignoring implicit traps, and if the cast is from a subtype to a supertype,
then we ignore the possible RTT-related inconsistency and can just drop the
cast.

See https://github.com/WebAssembly/binaryen/issues/3636

cc @askeksa-google


askeksa-google (approved):
Looks good.

test/lit/passes/optimize-instructions-gc-iit.wast
```diff
@@ -0,0 +1,144 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
+;; RUN: wasm-opt %s --optimize-instructions --ignore-implicit-traps --enable-reference-types --enable-gc -S -o - \
+;; RUN:   | filecheck %s
+
+(module
+  (type $A (struct (field i32)))
+  (type $B (struct (field i32) (field f64)))
+  (type $C (struct (field i64) (field f32)))
```
tlively, line 8: How about `parent`, `child`, and `other` rather than `A`, `B`, and `C`?

test/lit/passes/optimize-instructions-gc-iit.wast
```diff
@@ -0,0 +1,144 @@
...
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $ref-cast-iit-bad
+    (param $a (ref $A))
+    (param $b (ref $B))
+    (param $c (ref $C))
+
+    (param $a-rtt (rtt $A))
+    (param $b-rtt (rtt $B))
+    (param $c-rtt (rtt $C))
```
tlively, line 114: It looks like this only uses `$a` and `$a-rtt`. Removing the extra params would make the test output shorter.

tlively (approved):

PR #3749, opened 2021-03-29
Fix LegalizeJSInterface with RefFuncs

kripken:
This code used to remove functions it no longer thinks are needed. That is,
if it adds a legalized version of an import, it would remove the illegal
one which is no longer needed. To avoid removing an illegal import that
is still used it checked for `ref.func` appearances.

But this was bad in two ways:

1. We need to legalize the `ref.func`s too. We can't call an illegal import
  in any way, not using a direct call, indirect call, or call by reference of
  a `ref.func`.
2. It's silly to remove unneeded functions here. We have a pass for that.

This removes the removal of functions, and adds proper updating of
`ref.call`s, which means to call the stub function that looks like the
original import, but that calls the legalized one and connects things
up properly, exactly the same way as other calls.


src/passes/LegalizeJSInterface.cpp
```diff
@@ -165,14 +132,33 @@ struct LegalizeJSInterface : public Pass {
...
+        void visitRefFunc(RefFunc* curr) {
+          auto iter = illegalImportsToLegal->find(curr->func);
+          if (iter == illegalImportsToLegal->end()) {
+            return;
+          }
+
+          auto newName = iter->second;
+          if (newName == getFunction()->name) {
+            // inside the stub function itself, is the one safe place to do the
+            // call
```
aheejin, line 151: Why? I guess this function will be deleted anyway so doesn't matter, but is there a reason to make this an exception?

test/passes/legalize-js-interface_all-features.txt
```diff
@@ -2,16 +2,18 @@
...
+ (import "env" "imported" (func $imported (result i64)))
+ (import "env" "other" (func $other (param i32 i64 i64)))
  (import "env" "ref-func-arg" (func $ref-func-arg (result i64)))
  (import "env" "setTempRet0" (func $setTempRet0 (param i32)))
  (import "env" "getTempRet0" (func $getTempRet0 (result i32)))
  (import "env" "imported" (func $legalimport$imported (result i32)))
  (import "env" "other" (func $legalimport$other (param i32 i32 i32 i32 i32)))
  (import "env" "ref-func-arg" (func $legalimport$ref-func-arg (result i32)))
- (elem declare func $ref-func-arg)
+ (elem declare func $legalfunc$ref-func-arg)
```
aheejin, line 16: What causes this change? And it is unrelated, but what is this `declare` keyword? (I checked reference types, TFR, and GC proposals but couldn't find it..)

aheejin (commented):
Maybe I don't have the full understanding on this pass, so some questions:
- If you replace `call` instructions to legalized functions, where do we take care of the parents of those `call` instructions that will use the return values? So if a `call` that used to return one `i64` now returns two `i32`s, where do we change the instructions that use these values?
- Can we have a test case? (There are some test changes but I'm not sure that will represent this case?)

test/passes/legalize-js-interface_all-features.txt
martianboy: @aheejin A declared element segment is used as a forward declaration hint to engines so that they know what functions can be referenced using `ref.func`. It's completely unrelated to tables and other element segment types, and since they are not available at runtime, we don't represent them in the IR. Instead, we ignore this segment on parsing and regenerate it on writing or printing. Only functions that haven't been referenced in other element segments will be included in the declared segment, and only one segment of this type is emitted.

aheejin: I see. Thanks for the explanation!

tlively (approved):

test/passes/legalize-js-interface_all-features.txt
kripken: About the change: the change is because the `ref.func` lower down has been replaced to refer to another function. This has to correspond to that.

kripken:
@aheejin 

1. The idea here is that if we had an illegal import that returns and i64, we create a "thunk" that calls a legal import and returns and i64. So all we need to do is call the thunk instead of the import in all places, as it looks exactly like the illegal import used to.
2. The existing testcase here should be enough. It has a `ref.func` that is updated. (The title of the PR was RefCall by mistake, which might have been confusing but the problem is about RefFuncs - once we update them, later RefCalls will work fine, and don't need testing.)

src/passes/LegalizeJSInterface.cpp
kripken: Good catch! I think this code is pointless... No one will call the illegal function after this transform anyhow. I copied it from above as Call has the same problem to solve, but it's not needed in both places... I'll remove it.

PR #3750, opened 2021-03-30
SetGlobals pass

kripken:
This allows changing a global's value on the commandline in an easy way.

In some toolchains this is useful as the build can contain a global that
indicates something like a logging level, which this can customize.

### DIFF
--- a/test/passes/set-globals.wast
+++ b/test/passes/set-globals.wast
@@ -1,0 +1,7 @@
+(module
+ ;; an imported i32
+ (import "env" "in" (global $foo i32))
+
+ ;; a defined i64
+ (global $bar i64 (i64.const 1234))
+)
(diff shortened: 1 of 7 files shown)


test/passes/set-globals.wast
```diff
@@ -0,0 +1,7 @@
+(module
+ ;; an imported i32
+ (import "env" "in" (global $foo i32))
+
+ ;; a defined i64
+ (global $bar i64 (i64.const 1234))
```
aheejin, line 6: Since it also support creating a global, we can have a test case for it too I think?

aheejin (approved):
Why do we not support floats and doubles?

test/passes/set-globals.wast
kripken: Oh, oops, that was a bug. I intended it to skip missing globals. Fixing now.

(Adding globals is possible, but we'd need to be given the type too, and not sure there is a use case.)

kripken:
Float and double parsing is trickier, so I avoided it for now, as the use cases that were described to me don't need them.

If we do need them then maybe this should be generalized to use the parser's code, so the value would be wat like `(f32.const ...)` etc. But that would take some refactoring.

PR #3751, opened 2021-03-30
Fix DeadArgumentElimination pass on non-nullable locals

kripken:
In this case, there is a natural place to fix things up right after
removing a parameter (which is where a local gets added). Doing it
there avoids doing work on all functions unnecessarily.

Note that we could do something even simpler here than calling
the generic code: the parameter was not used, so the new local
is not used, and we could just change the type of the local or not
add it at all. Those would be slightly more code though, and add
complexity to the parameter removal method itself.

### DIFF
--- a/src/passes/DeadArgumentElimination.cpp
+++ b/src/passes/DeadArgumentElimination.cpp
@@ -41,6 +41,7 @@
 #include "ir/effects.h"
 #include "ir/element-utils.h"
 #include "ir/module-utils.h"
+#include "ir/type-updating.h"
 #include "pass.h"
 #include "passes/opt-utils.h"
 #include "support/sorted_vector.h"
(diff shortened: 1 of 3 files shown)


aheejin (approved):
This LGTM, but questions on your PR description:
Do we always create a local? Even if the argument is completely unused? I guess that's harmless because there should be a pass that removes unused locals, right?

And this is a separate question about the pass.. Why can't we reuse the same index and decrement all indices after the removed index by 1 here? https://github.com/WebAssembly/binaryen/blob/9b38e4394034d6f25f54fc471bb9ac8812815331/src/passes/DeadArgumentElimination.cpp#L443-L467

kripken:
Good point. Re-reading the old pass code, my theory is that it was meant to handle a more general case  where the parameter might be used (or it prepares for handling that in the future - which maybe we should). In that case we need to create a local with the value.

Yes, later passes remove unneeded locals so there is no noticeable harm from this atm.

I think we can't reuse the index because parameters come first. If we have `paramA paramB varA varB` and we remove `paramA` then we can only replace it with something at index 1 or above, which is where vars go, after the remaining `paramB`.

PR #3752, opened 2021-03-30
Scan all module-level code in collectHeapTypes()

kripken:
The key fix here is to call `walkModuleCode` so that we pick up on
types that appear only in the table and nowhere else. The rest of
the patch refactors the code so that that is practical.

@tlively this bug may be interfering with your investigation of that
other bug with heap type roundtripping.

Side question: can two wasm modules appear in a single lit test file?


tlively:
No, you can't generally have two modules in a lit test right now because our tools cannot consume text files with multiple modules. It would be possible to fix this via a wrapper script that split up the input and invoked the tool multiple times automatically. We already have a `not` wrapper script that works similarly.

kripken:
I see, thanks @tlively 

If we end up with many small files maybe such a script could be useful. For now it doesn't seem that bad to have small separate files.

PR #3753, opened 2021-03-30
Remove old syntax from table and elem in parser

martianboy:
We've been keeping old syntax in the text format parser although they've
been removed from the parser and hardly any test case relies on them.
This PR will remove old syntax support for tables and element segments
and simplify the corresponding parser functions. A few test files were
affected by this that are updated.


martianboy:
@tlively @aheejin @kripken Reviews appreciated here, too. (Alpine build step seems to be stuck.)

src/wasm/wasm-s-parser.cpp
```diff
@@ -3209,13 +3202,37 @@ void SExpressionWasmBuilder::parseTable(Element& s, bool preParseImport) {
...
   }
-  if (i == s.size()) {
-    wasm.addTable(std::move(table));
-    return;
+
+  bool hasExplicitLimit = false;
+
+  if (s[i]->isStr()) {
+    char c = s[i]->c_str()[0];
+    if ('0' <= c && c <= '9') {
```
tlively, line 3210: You can use `wasm::isInteger` to check for numbers here and below.

tlively (approved):

src/wasm/wasm-s-parser.cpp
martianboy: I'm not sure if I understand the point here, `wasm::isInteger` seems to check that a `double` is actually an integer and also not infinite.

tlively: Sorry, I misremembered the function name. I meant `String::isNumber` :)

martianboy: Oh nice! Thanks. Fixed now. By the way, do we parse hex numbers in general? Is there a utility for detecting and parsing hex the way it's defined and used in the spec? e.g. `0x1000_0000`

tlively: No, not that I know of.

PR #3757, opened 2021-03-30
Remove passive keyword from data segment parser

### LINKED ISSUE #2339: Remove "passive" from WAT for passive data segments
`wasm-opt -ttf --enable-bulk-memory --emit-text` can generate passive data segments like `(data passive "hi")` but this syntax is outdated. Instead it should be plain old `(data "hi")` -- I guess with the passivity inferred now? The removal issue doesn't provide context.

Originally incorrectly filed against WABT: https://github.com/WebAssembly/wabt/issues/1152

See https://github.com/WebAssembly/bulk-memory-operations/pull/84 for removal of "passive" in the WAT.

cc @tlively

martianboy:
The passive keyword has been removed from spec's text format, and now
any data segment that doesn't have an offset is considered as passive.
This PR remove that from both parser and the Print pass, plus all tests
that used that syntax.

Fixes #2339


src/wasm/wasm-s-parser.cpp
```diff
@@ -2858,27 +2858,32 @@ void SExpressionWasmBuilder::parseData(Element& s) {
...
   }
-  if (s[i]->isStr()) {
-    // data is passive or named
-    if (s[i]->str() == PASSIVE) {
-      isPassive = true;
+
+  if (s[i]->isList()) {
+    // Optional (memory <memoryidx>)
+    if (elementStartsWith(s[i], MEMORY)) {
+      i += 1;
```
tlively, line 2873: It would be good to add a TODO here about supporting multiple memories propoerly.

aheejin:
Oops, we raced... #3758

martianboy:
Oh sorry @aheejin! I went for it since I had just refactored `parseElem` which is very similar to `parseData`.

src/wasm/wasm-s-parser.cpp
martianboy: Done. Thanks!

PR #3758, opened 2021-03-30
Remove 'passive' keyword from data segments

### LINKED ISSUE #2339: Remove "passive" from WAT for passive data segments

aheejin:
This syntax is outdated and does not exist in the formal spec.
Fixes #2339.


aheejin:
Closing in favor of #3757.

Issue #3759, opened 2021-03-30
Flatten pass can make invalid modules on tuple types


martianboy:
This came up while fuzzing, and I traced it using `git bisect` to PR #3710 where non-nullable locals are handled by `handleNonNullableLocals`. It doesn't seem to handle tuple types that contain a non-nullable ref, resulting in a validation error after Flatten pass.

    seed: 17382324165229480416

Full validation result:

```[PassRunner]   (validating)
[wasm-validator error in function 3] unexpected false: vars must be defaultable, on 
(v128 (ref (func (param f32))) v128 i64)
[wasm-validator error in function 3] unexpected false: vars must be defaultable, on 
(v128 (ref (func (param f32))) v128 i64)
[wasm-validator error in function 3] unexpected false: vars must be defaultable, on 
(v128 (ref (func (param f32))) v128 i64)
[wasm-validator error in function 3] unexpected false: vars must be defaultable, on 
(v128 (ref (func (param f32))) v128 i64)
Fatal: Last pass (flatten) broke validation. Run with BINARYEN_PASS_DEBUG=2 in the env to see the earlier state, or 3 to dump byn-* files for each pass
```

Also `wasm-reducer` crashes with the following message when reducing the original file:

    Missing type: (func (param externref f32) (result f32))
    wasm-reduce: ../src/wasm/wasm-binary.cpp:529: uint32_t wasm::WasmBinaryWriter::getTypeIndex(wasm::HeapType) const: Assertion `0' failed.

kripken:
Thanks, https://github.com/WebAssembly/binaryen/pull/3760 will work around this in the fuzzer for now. Let's keep this open for a proper solution.

PR #3760, opened 2021-03-30
Avoid flatten + multivalue + reference types in the fuzzer

kripken:
The problem is that a tuple with a non-nullable element cannot be stored
to a local. We'd need to split up the tuple, but that raises questions about
what should be allowed in flat IR (we'd need to allow nested tuple ops
in more places). That combination doesn't seem urgent, so add a clear
error for now, and avoid it in the fuzzer.

Avoids #3759 in the fuzzer

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -988,6 +988,9 @@ def randomize_opt_flags():
             if '--disable-exception-handling' not in FEATURE_OPTS:
                 print('avoiding --flatten due to exception catching which does not support it yet')
                 continue
+            if '--disable-multivalue' not in FEATURE_OPTS and '--disable-reference-types' not in FEATURE_OPTS:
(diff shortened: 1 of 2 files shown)


kripken:
Landing quickly as this is inconvenient for people fuzzing right now, but PTAL when you can.

tlively (commented):
Looks good. This may not be urgent, but it definitely needs to be fixed, right? Do you expect this to be resolved by the current discussions around non-nullable types or will we need to look at other options like replacing flatten with poppify?

kripken:
It would be good to fix this eventually, but until we get a toolchain that uses both multivalue and GC, it's not really needed. (And I'm not sure if there is such a toolchain, but presumably eventually there will be...)

If we have to we can fix this in flatten with some work, by allowing tuple operations in nested places. But yes, poppify might be a more natural fit. There are a bunch of other stuff to think about there (like how current flat IR consumers would use poppy IR).

PR #3761, opened 2021-03-31
Fix type canonicalization bugs

### LINKED ISSUE #3736: Regression from equirecursive type canonicalization
Bisected to #3717

Testcase:
```wat
(module
 (type $type$0 (struct (field (ref $type$1))))
 (type $type$2 (func (param (ref $type$0)) (result (ref $type$2))))
 (type $type$1 (struct (field (mut (ref $type$2)))))
 (type $type$3 (struct (field (ref $type$1)) (field (ref $type$0))))

 (global $global$0 (rtt 0 $type$0) (rtt.canon $type$0))
 (global $global$1 (rtt 1 $type$3) (rtt.sub $type$3
  (global.get $global$0)
 ))
)
```
STR:
```
bin/wasm-opt a.wat -all --print --roundtrip
```
Note that writing to a file, then running another command to read from it works. It is only roundtrip, or the reducer, that read and write in the same process, that hit this. Perhaps some interaction with the existing types in the global type store?

tlively:
When canonical heap types were already present in the global store, for example
during the --roundtrip pass, type canonicalization was not working correctly.
The issue was that the GlobalCanonicalizer was replacing temporary HeapTypes
with their canonical equivalents one type at a time, but the act of replacing a
temporary HeapType use with a canonical HeapType use could change the shape of
later HeapTypes, preventing them from being correctly matched with their
canonical counterparts. This PR fixes that problem by computing all the
temporary-to-canonical heap type replacements before executing them.

To avoid a similar problem when canonicalizing Types, one solution would have
been to pre-calculate the replacements before executing them just like with the
HeapTypes, but that would have required either complex bookkeeping or moving
temporary Types into the global store when they are first canonicalized. That
would have been complicated because unlike for temporary HeapTypeInfos, the
unique_pointer to temporary TypeInfos is not readily available. This PR instead
switches back to using pointer-identity based equality and hashing for
TypeInfos, which works because we only ever canonicalize Types with canonical
children. This change should be a nice performance improvement as well.

Another bug this PR fixes is that shape hashing and equality considered
BasicKind HeapTypes to be different from their corresponding BasicHeapTypes,
which meant that canonicalization could produce different types for the same
type definition depending on whether the definition used a TypeBuilder or not.
The fix is to pre-canonicalize BasicHeapTypes (and Types that have them as
children) during shape hashing and equality. The same mechanism is also used to
simplify Store's canonicalization.

Fixes #3736.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -71,6 +71,10 @@ struct TypeInfo {
 
   bool isNullable() const { return kind == RefKind && ref.nullable; }
 
+  // If this TypeInfo represents a Type that can be represented more simply, set
+  // `out` to be that simpler Type and return true. For example, this handles
+  // canonicalizing the TypeInfo representing (ref null any) into the BasicType
+  // anyref. It also handles eliminating singleton tuple types.
   bool getCanonical(Type& out) const;
 
   bool operator==(const TypeInfo& other) const;
@@ -114,6 +118,9 @@ struct HeapTypeInfo {
   constexpr bool isArray() const { return kind == ArrayKind; }
   constexpr bool isData() const { return isStruct() || isArray(); }
 
+  // If this HeapTypeInfo represents a HeapType that can be represented more
+  // simply, set `out` to be that simpler HeapType and return true. This handles
+  // turning BasicKind HeapTypes into their corresponding BasicHeapTypes.
   bool getCanonical(HeapType& out) const;
 
   HeapTypeInfo& operator=(const HeapTypeInfo& other);
@@ -318,6 +325,9 @@ bool isTemp(HeapType type) {
   return !type.isBasic() && getHeapTypeInfo(type)->isTemp;
 }
 
+// Given a Type that may or may not be backed by the simplest possible
+// representation, return the equivalent type that is definitely backed by the
+// simplest possible representation.
 Type asCanonical(Type type) {
   if (!type.isBasic()) {
     getTypeInfo(type)->getCanonical(type);
@@ -325,6 +335,9 @@ Type asCanonical(Type type) {
   return type;
 }
 
+// Given a HeapType that may or may not be backed by the simplest possible
+// representation, return the equivalent type that is definitely backed by the
+// simplest possible representation.
 HeapType asCanonical(HeapType type) {
   if (!type.isBasic()) {
     getHeapTypeInfo(type)->getCanonical(type);
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -27,6 +27,12 @@
 #include "wasm-features.h"
 #include "wasm-type.h"
 
+#define TRACE_CANONICALIZATION 0
+
+#if TRACE_CANONICALIZATION
+#include <iostream>
+#endif
+
 namespace wasm {
 
 namespace {
@@ -64,6 +70,8 @@ struct TypeInfo {
   constexpr bool isRtt() const { return kind == RttKind; }
 
   bool isNullable() const { return kind == RefKind && ref.nullable; }
+
+  bool getCanonical(Type& out) const;
 
   bool operator==(const TypeInfo& other) const;
   bool operator!=(const TypeInfo& other) const { return !(*this == other); }
@@ -105,6 +113,8 @@ struct HeapTypeInfo {
   constexpr bool isStruct() const { return kind == StructKind; }
   constexpr bool isArray() const { return kind == ArrayKind; }
   constexpr bool isData() const { return isStruct() || isArray(); }
+
+  bool getCanonical(HeapType& out) const;
 
   HeapTypeInfo& operator=(const HeapTypeInfo& other);
   bool operator==(const HeapTypeInfo& other) const;
@@ -251,9 +261,7 @@ namespace std {
 
 template<> class hash<wasm::TypeInfo> {
 public:
-  size_t operator()(const wasm::TypeInfo& info) const {
-    return wasm::FiniteShapeHasher().hash(info);
-  }
+  size_t operator()(const wasm::TypeInfo& info) const;
 };
 
 template<> class hash<wasm::HeapTypeInfo> {
@@ -310,6 +318,20 @@ bool isTemp(HeapType type) {
   return !type.isBasic() && getHeapTypeInfo(type)->isTemp;
 }
 
+Type asCanonical(Type type) {
+  if (!type.isBasic()) {
+    getTypeInfo(type)->getCanonical(type);
+  }
+  return type;
+}
+
+HeapType asCanonical(HeapType type) {
+  if (!type.isBasic()) {
+    getHeapTypeInfo(type)->getCanonical(type);
+  }
+  return type;
+}
+
 TypeInfo::TypeInfo(const TypeInfo& other) {
   kind = other.kind;
   switch (kind) {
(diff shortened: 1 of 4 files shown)


kripken:
> the act of replacing a temporary HeapType use with a canonical HeapType use could change the shape of later HeapTypes, preventing them from being correctly matched with their canonical counterparts. 

Can you maybe give a concrete example of this? Sorry, I'm having a hard time getting it.

tlively:
Sure, here's a concrete example with canonical mutually recursive types `A` and `B` and equivalent temporary types `A'` and `B'` that are being canonicalized.

```
Global store: A <=> B
TypeBuilder: A' <=> B'
```

If we replace `A'` with `A` before we replace `B'` with `B`, we will end up with this:

```
Global store: A <=> B
TypeBuilder: B' -> A <=> B
```

What happened here was that `B'` was updated to refer to `A` rather than `A'`, but `A` still refers to `B`, not `B'`. Now `B'`'s shape no longer matches `B`'s shape, so it will not be canonicalized to `B`.

src/wasm/wasm-type.cpp
```diff
@@ -65,6 +71,8 @@ struct TypeInfo {
 
   bool isNullable() const { return kind == RefKind && ref.nullable; }
 
+  bool getCanonical(Type& out) const;
```
kripken, line 74: the meaning of the bool result isn't immediately obvious by the name - worth a comment i think

src/wasm/wasm-type.cpp
```diff
@@ -310,6 +318,20 @@ bool isTemp(HeapType type) {
   return !type.isBasic() && getHeapTypeInfo(type)->isTemp;
 }
 
+Type asCanonical(Type type) {
```
kripken, line 321: this also seems worth a comment. trying to figure out from the code (and from getCanonical at the same time), I'm still unsure.

src/wasm/wasm-type.cpp
```diff
@@ -341,11 +363,67 @@ TypeInfo::~TypeInfo() {
...
 
+bool TypeInfo::getCanonical(Type& out) const {
+  if (isTuple()) {
+    if (tuple.types.size() == 0) {
+      out = Type::none;
+      return true;
+    }
+    if (tuple.types.size() == 1) {
+      out = tuple.types[0];
+      return true;
```
kripken, line 374: still trying to reason about what this method does, I got stuck here. It looks like if a tuple of size 1 arrives here then we return that single value, and `true`. Why is the "canonical" value of the tuple that type? Maybe this is using "canonical" in a way I don't follow.

kripken (commented):
Thanks for the example!

Now I think I have enough to try to read the code, but I got stuck pretty quickly...

src/wasm/wasm-type.cpp
tlively: This code is almost exactly what `TypeStore::canonicalize` used to do, if that helps.

tlively:
Added comments—let me know if they help!

src/wasm/wasm-type.cpp
tlively: For the tuple case specifically: we don't allow singleton tuple types, so if you try to construct one, you should get the single type back instead of a fresh type.

src/wasm/wasm-type.cpp
kripken: Does it do anything else than those two examples?

What I'm looking for is the purpose of the method. "Simpler" in what sense? (`none` is simpler than anything :wink: ) And "respresented" in what sense? (if we can turn a singleton tuple into its sole value, that implies the tuple itself doesn't matter, but it's hard for me to guess at a generalized rule for that.)

Also, it sounds like this is unrelated to canonicalization or recursive types? If so, could it be `getSimpler`?

src/wasm/wasm-type.cpp
```diff
@@ -65,6 +71,12 @@ struct TypeInfo {
 
   bool isNullable() const { return kind == RefKind && ref.nullable; }
 
+  // If this TypeInfo represents a Type that can be represented more simply, set
+  // `out` to be that simpler Type and return true. For example, this handles
+  // canonicalizing the TypeInfo representing (ref null any) into the BasicType
+  // anyref. It also handles eliminating singleton tuple types.
+  bool getCanonical(Type& out) const;
```
kripken, line 78: Could this perhaps return `none` instead of false? The out param + bool adds some complexity here. But I guess that's not possible if you want this to turn `none` into something or something into `none` - maybe not though?

tlively: Ideally it return a `std::optional<Type>`, but until we have C++17 that seems better. I think we could use `Type::none` in principle, but the fact that I had to think through that for a bit make me think that it would be better not to do so.

src/wasm/wasm-type.cpp
tlively: Simpler as in how having Type::funcref is simpler than having a RefKind TypeInfo pointing to a BasicKind HeapTypeInfo that in turn contains HeapType::any. This is crucial for canonicalization and used to be part of `Store<TypeInfo>::canonicalize`. The reason this PR splits them out into separate methods is that the same logic needs to be applied in `FiniteShapeHasher` and `FiniteShapeEquator` (to fix canonicalization bugs) and it's much safer to use the same implementation in all locations than to duplicate the logic.

kripken: I see, thanks.

src/wasm/wasm-type.cpp
kripken: Should we make it an error to try to construct a tuple of size 1 then?

src/wasm/wasm-type.cpp
```diff
@@ -384,6 +475,14 @@ HeapTypeInfo::~HeapTypeInfo() {
   WASM_UNREACHABLE("unexpected kind");
 }
 
+bool HeapTypeInfo::getCanonical(HeapType& out) const {
+  if (isFinalized && kind == BasicKind) {
```
kripken, line 479: why does it matter if it's finalized or not?

src/wasm/wasm-type.cpp
tlively: I don't think so. Code that currently constructs types that may or may not be tuples can currently treat both cases uniformly, but if we made it an error to even try to construct a tuple of size 1 (or zero), that code would have to start dealing with the new discontinuity explicitly.

src/wasm/wasm-type.cpp
tlively: Imagine I have a temporary heap type backed by a HeapTypeInfo in a TypeBuilder. It may currently be set to a BasicHeapType, but in the future I may change it to be some other complex HeapType instead. It would be bad if temporary Types that refer to that TypeBuilder slot were prematurely canonicalized on creation to refer directly to a BasicHeapType instead because they would not observe the later changes to the temporary (mutable) HeapType.

kripken: I may have forgotten something. What is the use case for having a type in a TypeBuilder change? I thought the usage is to define the N types, each once, and then the builder handles canonicalization etc., but none of that could replace a basic heap type with a complex one?

tlively: There is no concrete use case specifically for replacing a basic heap type with a complex one, but we do depend more generally on the type being able to be set more than once because the TypeBuilder entries are initialized to arbitrary types and then set again by the user later. So far we have not enforced any additional constraints on how it can be mutated. IMO, the simplest mental model for the TypeBuilder is that it is entirely mutable with no constraints.

kripken: I see, that makes sense. It does seem hard to think of an API that would not allow user changes.

kripken (approved):

tlively:
Thanks, @kripken!

PR #3763, opened 2021-03-31
[RT] Add type to tables and element segments

martianboy:
This will add support for other reference types to tables and elems,
which was introduced in reference-types proposal and extended in typed
function references proposal. Non-nullable types are not supported yet,
since the spec only mentions the idea of using an initializer, but
it has remained a TODO item to this date.
Spec tests have also been partially updated to reflect the new changes
from upstream.


martianboy:
This includes and is based on the changes from #3753, so I marked this a draft until that one is merged.

@tlively @kripken @aheejin It's possible that I have some misunderstandings regarding the spec, specially when it comes to the new types introduced in GC proposal. Please take a look and let me know if something is incorrect. Thanks!

src/ir/module-utils.h
```diff
@@ -510,6 +513,12 @@ inline void collectHeapTypes(Module& wasm,
   for (auto& curr : wasm.events) {
     counts.note(curr->sig);
   }
+  for (auto& curr : wasm.tables) {
+    counts.maybeNote(curr->type);
+  }
+  for (auto& curr : wasm.elementSegments) {
+    counts.maybeNote(curr->type);
+  }
```
martianboy, line 521: @tlively is this enough for making sure table & elem types are counted correctly everywhere, or have I missed something?

martianboy:
This is now ready for review, I've rebased it on top of the current main after #3753 was merged.

src/wasm/wasm-validator.cpp
```diff
@@ -2845,23 +2852,66 @@ static void validateTables(Module& module, ValidationInfo& info) {
...
+                          table->type == Type::externref,
+                        "table",
+                        "Only funcref and externref are valid for table type.");
+    }
+  }
+
   for (auto& segment : module.elementSegments) {
+    info.shouldBeTrue(segment->type != Type::externref,
+                      "elem",
+                      "element segment type cannot be externref");
```
tlively, line 2880: This is surprising!

tlively (commented):
Nice work! I only have a few questions :)

src/wasm/wasm-validator.cpp
```diff
@@ -797,6 +797,10 @@ void FunctionValidator::visitCallIndirect(CallIndirect* curr) {
   if (curr->target->type != Type::unreachable) {
     auto* table = getModule()->getTableOrNull(curr->table);
     shouldBeTrue(!!table, curr, "call-indirect table must exist");
+    shouldBeSubType(table->type,
+                    Type::funcref,
+                    curr,
+                    "call-indirect table type must be a funcref.");
```
kripken, line 803: Could we allow `table->type.isFunction()` here? More specific function types are ok unless I'm missing something.

src/wasm/wasm-validator.cpp
```diff
@@ -2829,6 +2833,9 @@ static void validateTables(Module& module, ValidationInfo& info) {
                       "--enable-reference-types)");
     if (!module.tables.empty()) {
       auto& table = module.tables.front();
+      info.shouldBeTrue(table->type == Type::funcref,
+                        "table",
+                        "Only funcref is valid for table type.");
```
kripken, line 2838:
```suggestion
                        "Only funcref is valid for table type (when reference types are disabled).");
```

src/wasm/wasm-validator.cpp
```diff
@@ -2845,23 +2852,66 @@ static void validateTables(Module& module, ValidationInfo& info) {
...
+    if (!module.features.hasTypedFunctionReferences()) {
+      info.shouldBeTrue(table->type == Type::funcref ||
+                          table->type == Type::externref,
+                        "table",
+                        "Only funcref and externref are valid for table type.");
+    }
+  }
+
   for (auto& segment : module.elementSegments) {
+    info.shouldBeTrue(segment->type != Type::externref,
```
kripken, line 2878: we have `shouldBeUnequal`

src/passes/Print.cpp
```diff
@@ -2711,9 +2712,9 @@ struct PrintSExpression : public UnifiedExpressionVisitor<PrintSExpression> {
       });
     auto printElemType = [&]() {
       if (allElementsRefFunc) {
-        TypeNamePrinter(o, currModule).print(HeapType::func);
+        o << "func";
       } else {
-        TypeNamePrinter(o, currModule).print(Type::funcref);
+        printType(o, curr->type, currModule);
```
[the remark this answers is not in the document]
martianboy: This should be a value type. My understanding is that the only heap type allowed here is `func` which is used in the abbreviated form that doesn't use expressions. In all other cases, it should be a value type and items should be expressions.

src/wasm/wasm-binary.cpp
```diff
@@ -2765,11 +2761,11 @@ void WasmBinaryBuilder::readTableDeclarations() {
   auto numTables = getU32LEB();
 
   for (size_t i = 0; i < numTables; i++) {
-    auto elemType = getS32LEB();
-    if (elemType != BinaryConsts::EncodedType::funcref) {
-      throwError("Non-funcref tables not yet supported");
+    auto elemType = getType();
+    if (!elemType.isRef()) {
```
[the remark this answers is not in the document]
martianboy: In MVP all tables should be `funcref`, right? We defer checking features until validation.

src/wasm/wasm-s-parser.cpp
```diff
@@ -3205,7 +3205,7 @@ void SExpressionWasmBuilder::parseTable(Element& s, bool preParseImport) {
       table->module = inner[1]->str();
       table->base = inner[2]->str();
       i++;
-    } else {
+    } else if (!elementStartsWith(inner, REF)) {
```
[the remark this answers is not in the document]
martianboy: No, we are here after we saw `s[i]` to be a list. In that case, we still don't want to parse the type yet, just making sure it passes over the parse exception.

test/spec/elem_reftypes.wast
```diff
@@ -0,0 +1,261 @@
...
+(module
+  (table 20 funcref)
+  (elem (i32.const 20))
+)
+
+;; (module
+;;   (import "spectest" "table" (table 0 funcref))
+;;   (func $f)
+;;   (elem (i32.const 0) $f)
+;; )
```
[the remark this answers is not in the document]
martianboy: I explained the reasons above, but sure I'll add a comment. Thanks!

src/wasm/wasm-validator.cpp
martianboy: Again, this is only my understanding of the spec but haven't seen the spec text mention it anywhere, nor any spec tests that expects this to be invalid. But, here's my reasoning:

Element segments are only used to initialize tables and cannot be imported from outside. Given the fact that they cannot include function call expressions, there's no way for them to hold `externref` values. The only way for a wasm module to change items on an `externref` table is via the table instructions like `table.set` and `table.copy` that are allowed at runtime.

src/wasm/wasm-validator.cpp
martianboy: You're right, I had totally missed this. I'll add a test case for it as well. Thanks!

src/wasm/wasm-validator.cpp
```diff
@@ -2874,7 +2924,12 @@ static void validateTables(Module& module, ValidationInfo& info) {
         info.shouldBeTrue(
           expr->is<RefFunc>() || expr->is<RefNull>(),
```
martianboy, line 2875: One missing part seems to be here, where we validate allowed instructions for elem expressions. My understanding is that we should allow things like `struct.new` and `array.new`. Is that correct? @kripken @tlively

src/wasm/wasm-validator.cpp
kripken: Oh, I missed that it checked for a subtype... yeah, it's correct, but would be clearer the other way, I agree

src/wasm/wasm-validator.cpp
kripken: I think the answer is no. Only rtts and global.get: https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#constant-expressions

martianboy: @kripken hmm thanks! So far I've been avoiding RTTs to be honest :) so my knowledge of them in GC is limited, and I will dig deeper for sure. But just to be clear, that would mean I can't put an array on an element segment and initialize a table with it, right? E.g.

```
(module
  (type $vec (array f32))
  (table 1 dataref)
  (elem (i32.const 0) dataref (array.new_default $vec (i32.const 3) (rtt.canon $vec)))
)
```

martianboy: So it appears that there's an open issue for extending the list of constant instructions, WebAssembly/gc#189. That answers my questions. I'll make some changes accordingly.

martianboy: @kripken GC's MVP document [says](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md#reference-types)

    (rtt <n>? $t) == (ref (rtt <n>? $t))

But we don't seem to parse `(ref (rtt <n>? $t))` at all, right? And we don't consider `(rtt <n>? $t)` to be a reference type? So basically I need to check that the table type is either `.isRef()` or `.isRtt()`.

kripken: Making RTTs into reference types is a very recent change to the GC spec, that I don't think anyone has implemented. I don't even think it has been discussed at length in the CG actually.

For now, binaryen implements the [GC prototype 3 doc](https://docs.google.com/document/d/1DklC3qVuOdLHSXB5UXghM_syCh-4cMinQ50ICiXnK3Q/edit) which is a snapshot of the spec from earlier (snapshots allow us to interop with compilers and VMs, so we all do the same thing, ignoring the daily churn in the spec). So for us it is correct to not parse `(ref (rtt ..))`, for now.

martianboy: Thanks for the explanation. In that case, I will skip RTT altogether since we don't consider them as reference types and even then, they don't seem to be nullable in the current version of the spec.

src/wasm/wasm-validator.cpp
```diff
@@ -797,10 +797,9 @@ void FunctionValidator::visitCallIndirect(CallIndirect* curr) {
   if (curr->target->type != Type::unreachable) {
     auto* table = getModule()->getTableOrNull(curr->table);
     shouldBeTrue(!!table, curr, "call-indirect table must exist");
-    shouldBeSubType(table->type,
-                    Type::funcref,
-                    curr,
-                    "call-indirect table type must be a funcref.");
+    shouldBeTrue(table->type.isFunction(),
+                 curr,
+                 "call-indirect table type must be a funcref.");
```
kripken, line 802:
```suggestion
                 "call-indirect table must be of function type.");
```

martianboy: Done.

src/wasm/wasm-validator.cpp
martianboy: Done.

src/wasm/wasm-validator.cpp
martianboy: Done.

src/wasm/wasm-validator.cpp
martianboy: It's now using `->isFunction()`.

src/wasm/wasm-validator.cpp
martianboy: This is now changed to only allow function types for element segments. I added a comment block explaining the reason.

test/spec/elem_reftypes.wast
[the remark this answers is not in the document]
martianboy: I added a comment explaining the reason.

kripken (approved):
Nice work! No more comments from me.

tlively:
I fuzzed this a bit and it looks like the fuzzer is producing invalid modules for seed 7344750093000028152. Can you reproduce that? (Besides fuzzing this LGTM, too!)

martianboy:
@tlively thanks I will look into it!

martianboy:
@tlively Thanks for the catch! The fuzzer wasn't correctly handling modules that only have element segments with more specific function types. I fixed it by making sure at least one `funcref` element segment and table exist. See my last commit. On my local machine, the fuzzer can now run a few hundred iterations without a problem.

tlively:
Excellent, thanks @martianboy!

PR #3764, opened 2021-03-31
Fix an iterator invalidation error in the fuzzer

kripken:


aheejin (approved):

src/tools/fuzzing.h
```diff
@@ -934,8 +934,14 @@ class TranslateToFuzzReader {
...
     auto chance = upTo(RESOLUTION + 1);
+    // Do not iterate on wasm.functions itself, as we may add to it as we go
+    // through the functions (make() can add new functions to implement a
+    // RefFunc).
+    std::vector<Function*> functions;
     for (auto& ref : wasm.functions) {
-      auto* func = ref.get();
+      functions.push_back(ref.get());
+    }
+    for (auto* func : functions) {
```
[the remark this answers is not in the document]
kripken: Oh, good point, thanks! Done.

Also I think we can just iterate with a dynamic check on `size()`. It's not bad to iterate on the new functions, in fact we might as well, to avoid bugs in the future.

src/tools/fuzzing.h
```diff
@@ -934,8 +934,13 @@ class TranslateToFuzzReader {
     // Pick a chance to fuzz the contents of a function.
     const int RESOLUTION = 10;
     auto chance = upTo(RESOLUTION + 1);
-    for (auto& ref : wasm.functions) {
-      auto* func = ref.get();
+    // Do not iterate on wasm.functions itself, as we may add to it as we go
+    // through the functions - make() can add new functions to implement a
+    // RefFunc. Looping on an index avoids an iterator invalidation, and also
```
tlively, line 939: This first sentence no longer applies, since we're now iterating on wasm.functions itself but just avoiding using iterators.

kripken: Good point, that's not clear enough. I meant "not iterating on `wasm.functions`" as "not doing `for (x : wasm.functions)` directly, as opposed to an index. Pushed a better wording (I hope).

tlively (approved):

PR #3765, opened 2021-03-31
[Wasm GC] Local-Subtyping pass

kripken:
If a local is say `anyref`, but all values assigned to it are something
more specific like `funcref`, then we can make the type of the local
more specific. In principle that might allow further optimizations, as
the `local.get`s of that local will have a more specific type that their
users can see, like this:
```wat
(import .. (func $get-funcref (result funcref)))

(func $foo (result i32)
  (local $x anyref)
  (local.set $x (call $get-funcref))
  (ref.is_func (local.get $x))
)
=>
(func $foo (result i32)
  (local $x funcref) ;; updated to a subtype of the original
  (local.set $x (call $get-funcref))
  (ref.is_func (local.get $x)) ;; this can now be optimized to "1"
)
```
A possible downside is that using more specific types may not end
up allowing optimizations but may end up increasing the size of
the binary (say, replacing lots of `anyref` with various specific
types that compress more poorly; also, for recursive types the LUB
may be a unique type appearing nowhere else in the wasm as
pointed out by @tlively ).

So it is not clear if this is worth it or not. A global version of this pass
that specializes types of functions is also worth looking into, with
similar issues (although there we can also add non-nullability, which
may make it more beneficial). Marking as draft for now until we get
more real-world code to test on, and to get feedback.


src/passes/LocalSubtyping.cpp
```diff
@@ -0,0 +1,129 @@
...
+        // Remove non-nullability, which can't be stored in a local.
+        if (newType.isRef() && !newType.isNullable()) {
+          newType = Type(newType.getHeapType(), Nullable);
+          // Note that the old type must have been nullable as well, as non-
+          // nullable types cannot be locals, which means that we will not have
+          // to do any extra work to handle non-nullability if we update the
+          // type: we are just updating the heap type, and leaving the type
+          // nullable as it was.
+          assert(oldType.isRef() && oldType.isNullable());
+        }
```
tlively, line 101: This extra code could be avoided by including the original type of the local in the LUB calculation. All upper bounds of a nullable type are nullable. If we ever support non-nullable locals, the current code would be better, though.

tlively (commented):
Neat! One interesting thing to think about is that using LUBs would never increase code size under a nominal type system because the LUBs would necessarily have to already be defined in the module.

src/passes/LocalSubtyping.cpp
kripken: Wouldn't that defeat the optimization? If the local type is `anyref` and the assigned value is `funcref`, the LUB of those two is `anyref`, so we wouldn't change anything (even though we could).

Or do you mean to include "nullability" in the LUB but without the local's HeapType somehow?

tlively: Oh oops, you're totally right 👍 I was confused :)

kripken:
As our current focus is on speed and not size, I think this makes sense to land. I updated and polished it up now.

kripken:
Testing this on Dart, I see some benefits. This allows the removal of 3.5% of local.sets that otherwise would be present. Not sure if that translates into a speedup or not, but it seems good. (The overall code size reduction is very small, 0.2%, but at least it is an improvement and not a regression.)

tlively (approved):
There are cases where it would be beneficial to do this in a loop with the parameter refinement, for example when the parameter locals are part of a chain. I don't know how common they would be in practice, though.

kripken:
> There are cases where it would be beneficial to do this in a loop with the parameter refinement, for example when the parameter locals are part of a chain. I don't know how common they would be in practice, though.

True, yeah. Given they are in separate passes, it's not easy to loop on just them, but `--converge` would handle this.

PR #3766, opened 2021-03-31
Catch bad tuple.extract index in parser

### LINKED ISSUE #3703: Tuple-related segfault
Testcase:
```wat
(module
 (type $none_=>_none (func))
 (type $funcref_funcref_v128_=>_v128 (func (param funcref funcref v128) (result v128)))
 (type $eqref_f32_=>_externref_v128_f32_eqref_funcref_funcref (func (param eqref f32) (result externref v128 f32 eqref funcref funcref)))
 (type $none_=>_externref_v128_f32_eqref_ref?|none_->_none|_funcref (func (result externref v128 f32 eqref (ref null $none_=>_none) funcref)))
 (global $global$4 (mut i32) (i32.const 10))
 (func $0 (param $0 funcref) (param $1 funcref) (param $2 v128) (result v128)
  (unreachable)
 )
 (func $1 (param $0 eqref) (param $1 f32) (result externref v128 f32 eqref funcref)
  (local $2 (externref v128 f32 eqref funcref))
  (local $3 funcref)
  (tuple.make
   (block (result externref)
    (local.set $3
     (tuple.extract 5
      (block (result externref v128 f32 eqref (ref null $none_=>_none))
       (global.set $global$4
        (i32.sub
         (global.get $global$4)
         (i32.const 1)
        )
       )
       (local.tee $2
        (tuple.make
         (ref.null extern)
         (v128.const i32x4 0x00000000 0x00000000 0x00000000 0x00000000)
         (f32.const 0)
         (ref.null eq)
         (ref.null $none_=>_none)
        )
       )
      )
     )
    )
    (ref.null extern)
   )
   (v128.const i32x4 0x00000000 0x00000000 0x00000000 0x00000000)
   (f32.const 0)
   (ref.null eq)
   (ref.null func)
  )
 )
)
```
STR: `wasm-opt -all a.wat`

tlively:
Previously an out-of-bounds index would result in an out-of-bounds read during
finalization of the tuple.extract expression.

Fixes #3703.

### DIFF
--- a/src/wasm/wasm-s-parser.cpp
+++ b/src/wasm/wasm-s-parser.cpp
@@ -2527,6 +2527,10 @@ Expression* SExpressionWasmBuilder::makeTupleExtract(Element& s) {
   auto ret = allocator.alloc<TupleExtract>();
   ret->index = atoi(s[1]->str().c_str());
   ret->tuple = parseExpression(s[2]);
+  if (ret->tuple->type != Type::unreachable &&
+      ret->index >= ret->tuple->type.size()) {
+    throw ParseException("Bad index on tuple.extract", s[1]->line, s[1]->col);
+  }
   ret->finalize();
   return ret;
 }
--- a/src/wasm/wasm.cpp
+++ b/src/wasm/wasm.cpp
@@ -872,6 +872,7 @@ void TupleExtract::finalize() {
   if (tuple->type == Type::unreachable) {
     type = Type::unreachable;
   } else {
+    assert(index < tuple->type.size());
     type = tuple->type[index];
   }
 }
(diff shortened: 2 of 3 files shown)


kripken (commented):
Could be better to add this check to `ValidatingBuilder` and use that here, and then it could be used in the binary parser as well.

tlively:
The binary parser never has to deal with this problem because tuple.extract never appears in binaries. I'll add an assertion in `TupleExtract::finalize`, though, so if some pass creates a tuple.extract with a bad index, we will still catch it.

kripken (approved):
Oh, right...

PR #3769, opened 2021-04-01
Improve a test from binary to text

kripken:
Instead of specifying the features in the wasm, enable them using
flags. This allows the testcase to be in text.

Suggested in https://github.com/WebAssembly/binaryen/pull/3692#issuecomment-800727474


tlively (approved):
Nice! This would also make a good lit test, where the input, output, and flags could all be in one file.

kripken:
I tried to make a lit test but I couldn't figure out how? The auto-update script only updates function results, and this test checks module-level things (the type section at the top, and the features printed at the bottom). So nothing of what we need to check is added as a check it seems.

tlively:
You would have to hand-write this one, but that's generally ok. (Alternatively, the script could be changed to handle more than just the common case of checking function contents.)

kripken:
I see, thanks.

I think I'll wait on a script to auto-update those parts too. Without such a script we could end up with a lot of work some day if we need to update tests by hand.

PR #3770, opened 2021-04-02
Reorder global definitions in Print pass

martianboy:
This is needed to make sure globals are printed before element segments,
where `global.get` can appear both as offset and an item expression.


martianboy:
I needed this for #3763.

tlively (approved):

PR #3773, opened 2021-04-05
Wasm2JS Fuzzing: Print types when we have nothing better

kripken:
This matches #3747 which makes us not log out reference values, instead
we print just their types.

This also prints a type for non-reference things, replacing a previous
exception, which affects things like SIMD and BigInts, but those trap
anyhow at the JS boundary I believe (or did that change for SIMD?).
Anyhow, printing the type won't give a false "ok" when comparing wasm2js
output to the interpreter, assuming the interpreter prints out a value and
not just a type (which is the case). We could try to do better, but this
code is on the JS side, where we don't have the type - just a string
representation of it, which we'd need to parse etc.


aheejin (approved):
> Anyhow, printing the type won't give a false "ok" when comparing wasm2js
output to the interpreter, 

The current code just throws. Does that give a false "ok"?

kripken:
Throwing would end up printing that the call to the export trapped, which normally would not look like a false "ok" (unless if the export actually was supposed to trap). So I think this change should be strictly better.

PR #3774, opened 2021-04-05
Wasm2JS fuzzing: Emit null for reference type params instead of 0

kripken:
VMs will not convert a 0 or `undefined` from JS into a wasm null reference - it must be `null`.

### DIFF
--- a/src/tools/js-wrapper.h
+++ b/src/tools/js-wrapper.h
@@ -103,7 +103,7 @@ static std::string generateJSWrapper(Module& wasm) {
     }
     ret += std::string("instance.exports.") + exp->name.str + "(";
     bool first = true;
-    for (const auto param : func->sig.params) {
+    for (auto param : func->sig.params) {
       // zeros in arguments TODO more?
       if (first) {
         first = false;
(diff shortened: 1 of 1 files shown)


aheejin (approved):
So wasmjs supports reference types? And not GC I guess?

kripken:
Oh, sorry, I mislabeled these PRs. It's not wasm2js - which does not support reference types of GC yet - but running the wasm in a JS VM. We do support GC there.

src/tools/js-wrapper.h
```diff
@@ -103,16 +103,20 @@ static std::string generateJSWrapper(Module& wasm) {
...
       // zeros in arguments TODO more?
       if (first) {
         first = false;
       } else {
         ret += ", ";
       }
-      ret += "0";
-      if (param == Type::i64) {
-        ret += ", 0";
+      if (param.isRef()) {
```
aheejin, line 113: If we support GC, can this be data or something else too?

kripken: `isRef()` returns true for a reference to GC data, so this should cover that. (Or did I misunderstand your question?)

aheejin: Ah you're right. Then is this possible to be an RTT?

kripken: Oh right, it could be... in that case we can't do anything (no way to create an RTT in JS), and the JS side will trap. When I get to adding RTTs as parameters in the fuzzer I'll have to handle that somehow...

Issue #3767, opened 2021-03-31
Type of br_if should be the target's type, not the value


kripken:
```wat
(block $target (result ..type..)
  (br_if $target
    (..value..)
    (..condition..)
  )
)
```
The type of the `br_if` should be the type of `$target`, not of `..value..`. Binaryen gets this wrong while v8, wabt, and wasp get it right.

In practice the return value of `br_if` is not that useful - I've only ever seen it dropped, except in fuzzer-generated code. That's probably why this never came up. (The fuzzer found it now.)

For that reason I'm not sure it's worth fixing. Another reason is that this is parallel to https://github.com/WebAssembly/gc/issues/201 where there is discussion on making `local.tee`'s type that of the value and not the local, which is basically the same issue. If we make that change then Binaryen is already doing the right thing.

aheejin:
I agree; I found this problem in Jan 2020 and tried to fix it by giving a type every time we make `br_if`, but it was a lot harder than giving a type `local.tee`... Making what we're doing correct sounds good to me.

tlively:
We now work around this in the binary writer by inserting casts after branches as necessary, so I'll go ahead and close this. We may need to revisit this strategy in the future, for example if we introduce new references that cannot be cast (continuation references are possibly already in this category), but we can discuss that in follow-on issues.

PR #3775, opened 2021-04-06
Fuzzer: Ignore our current bug with the type of br_if

kripken:
We give `br_if` a too specific type: https://github.com/WebAssembly/binaryen/issues/3767

This is only noticeable with GC, and in rare cases where the type of `br_if`
is actually used - which realistically it never is, so really just fuzzer testcases.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -351,7 +351,12 @@ def fix_spec_output(out):
 def run_vm(cmd):
     # ignore some vm assertions, if bugs have already been filed
     known_issues = [
-        'local count too large',    # ignore this; can be caused by flatten, ssa, etc. passes
+        # can be caused by flatten, ssa, etc. passes
+        'local count too large',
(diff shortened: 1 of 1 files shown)


tlively (commented):
Would it be better to just relax our validation rules? That way at least we could have tests demonstrating our implemented behavior.

kripken:
The validation error comes from v8 in this case. We emit code that is valid if the `br_if`'s type is more specific, then `v8` (or any other compliant VM) sees a too-general type.

Getting us to emit the proper general type is not trivial (it's not present on the `br_if` - it's only known by the block we branch to). But also as mentioned in the linked issue, we probably want to keep the specific type (as with `tee`)....

tlively:
Oh, I see, we would have to insert extra downcasts to make V8 happy, but that's not worth it considering that V8 will probably be updated to accept what we are emitting. Makes sense 👍

tlively (approved):

PR #3779, opened 2021-04-07
Write names subsections by increasing id

sbc100:
See https://webassembly.github.io/spec/core/appendix/custom.html

"Each subsection may occur at most once, and in order of increasing id."


tlively (approved):
Did this cause issues somewhere?

sbc100:
> Did this cause issues somewhere?

Yes, when writing files in binaryen and then reading them wabt (which checks for this ordering).  This doesn't normally happen in emscripten, but I do it sometimes.

PR #3787, opened 2021-04-08
Fuzzer: Do not assume table 0 is funcref, as initial content may include a non-funcref table

kripken:
Noticed on fuzz seed: 13022862020752824190

cc @tlively @martianboy - was there already a mechanism for this that I missed perhaps?

Also includes test suite changes that clang-format found.


tlively (approved):
Looks good! I wonder why the format issue wasn't caught by CI in #3783, though 🤔

martianboy:
@kripken No specific mechanism, no. I saw this bug with call_indirect a couple of days ago. The approach is simple enough, although I think you need to make sure the element segment, which is randomly selected here, also has a compatible type. (Element segments can only have function types, but that could still be incompatible with the desired type.)

martianboy:
@tlively I don't think GCC tests (test/example) are covered by the CI linter.

martianboy:
Right, we do ensure that a `funcref` element segment exists in `setupTables`, we just don't select the correct segment in `makeCallIndirect`.

kripken:
> we just don't select the correct segment in makeCallIndirect.

makeCallIndirect doesn't get a segment, though, just a table? Do you mean that if the contents of the segment have the wrong type it can throw? That's ok, and it's good the fuzzer generates testcases that cover that. (By default, without initial content, we should not be throwing most of the time, so we should cover both possibilities.)

kripken:
(anyhow, landing this for now to fix that seed, we can add further improvements separately)

PR #3790, opened 2021-04-08
RefFunc: Validate that the type is non-nullable, and avoid possible bugs in the builder

kripken:
The builder can receive a HeapType so that callers don't need to set non-nullability
themselves.

Not NFC as some of the callers were in fact still making it nullable.


tlively (approved):
This is a good idea. Would it makes sense to go even further and just look up the signature given the function name, or are there use cases where ref.func needs to be built while the referenced function is not owned by the module?

kripken:
Yes, it's not possible in general to go further. You can construct function A with a reference to B before B has been created (only its name is known). And overall we don't require a full module context when creating function-level content.

PR #3791, opened 2021-04-09
wasm-split: Update dylink section when growing table

sbc100:


sbc100:
Do you think I can/should write a test for this?

tlively (approved):
Do we have any sort of text format output for the dylink section? If so, it would be nice to see that it is updated. Otherwise, it's probably not worth having a binary test.

PR #3793, opened 2021-04-10
Remove Type ordering

tlively:
As found in #3682, the current implementation of type ordering is not correct,
and although the immediate issue would be easy to fix, I don't think the current
intended comparison algorithm is correct in the first place. Rather than try to
switch to using a correct algorithm (which I am not sure I know how to
implement, although I have an idea) this PR removes Type ordering entirely. In
places that used Type ordering with std::set or std::map because they require
deterministic iteration order, this PR uses new set and map wrappers to provide
deterministic insertion-order iteration.


src/support/insertion_order.h
```diff
@@ -0,0 +1,133 @@
...
+  std::vector<Key> vec;
+
+public:
+  // Do not allow modification of items.
+  using iterator = typename std::vector<Key>::const_iterator;
+
+  iterator begin() { return vec.cbegin(); }
+  iterator end() { return vec.cend(); }
+
+  bool empty() const { return vec.size() == 0; }
```
kripken, line 48:
```suggestion
  bool empty() const { return vec.empty(); }
```

src/support/insertion_order.h
```diff
@@ -0,0 +1,133 @@
...
+  std::vector<std::pair<const Key, T>> vec;
+
+public:
+  using value_type = std::pair<const Key, T>;
+  using iterator = typename std::vector<value_type>::iterator;
+
+  iterator begin() { return vec.begin(); }
+  iterator end() { return vec.end(); }
+
+  bool empty() const { return vec.size() == 0; }
```
kripken, line 92:
```suggestion
  bool empty() const { return vec.empty(); }
```

src/wasm-stack.h
```diff
@@ -141,7 +141,7 @@ class BinaryInstWriter : public OverriddenVisitor<BinaryInstWriter> {
 
   // Keeps track of the binary index of the scratch locals used to lower
   // tuple.extract.
-  std::map<Type, Index> scratchLocals;
+  std::unordered_map<Type, Index> scratchLocals;
```
kripken, line 144: this must be sorted, we iterate on it and it affects the order of `localTypes` which is reflected in the binary IIANM

kripken (commented):
Nice!

For fuzzing this, I'd up the frequency of the `CheckDeterminism` tool from 0.1 to 1.

tlively:
Good idea with the fuzzing. I quickly found that there was indeed a determinism issue. The parallel function analysis used a map that was deterministically ordered by non-deterministic function addresses, which ultimately caused the insertion order in `collectHeapTypes` to be non-deterministic. I fixed this by switching to an ordered_map in the parallel function analysis, which required using an ordered_map in a few other places as well to get the types to match.

src/passes/Asyncify.cpp
```diff
@@ -380,8 +381,8 @@ class FakeGlobalHelper {
   }
 
 private:
-  std::map<Type, Name> map;
-  std::map<Name, Type> rev;
+  std::unordered_map<Type, Name> map;
+  std::unordered_map<Name, Type> rev;
```
kripken, line 385: How about for changes like this, we create a new `uniterable_map` (or better name), which is `unordered_map` without the ability to iterate? That would avoid the possibility of nondeterminism, which otherwise can be very annoying to find later if we miss anything in this PR.

In other words, the PR could change a `map` into either `uniterable_map` (if we never iterate on it) or `ordered_map` (if we do). Then we should have 0 risk.

src/ir/module-utils.h
```diff
@@ -377,7 +378,7 @@ template<typename T> struct CallGraphPropertyAnalysis {
     bool hasNonDirectCall = false;
   };
 
-  typedef std::map<Function*, T> Map;
+  typedef ordered_map<Function*, T> Map;
```
kripken, line 381: `CallGraphPropertyAnalysis` is used widely so the extra overhead here worries me a little. Worth measuring the memory and speed impact of this PR.

src/passes/Asyncify.cpp
tlively: If we had an `uniterable_map` available in a header, I have a hard time imagining that I would use it beyond this single PR. The tough part about getting the determinism correct here is that now getting deterministic iteration depends on having deterministic insertions, whereas previously the insertions could be non-deterministic and the iteration would still be deterministic due to sorting. An `uniterable_map` would not help solve that problem, unfortunately.

kripken: I see, very good point...

Yes, maybe not that generally useful then. Still I think it's worthwhile, but can discuss separately if it doesn't help you here.

tlively:
I plan to update this so that not all parallel analyses are affected; only the ones that matter.

kripken:
Btw, I noticed that we already have code for insert-ordered data structures, `InsertOrderedSet` and `InsertOrderedMap` in `src/cfg/Relooper.h`. If you want I can look into refactoring the relooper code for that out (although I think it might be trivial).

tlively:
Oh, good to know. I'll take a look at that refactoring.

src/support/insert_ordered.h
```diff
@@ -126,13 +140,26 @@ template<typename Key, typename T> struct InsertOrderedMap {
...
-    // TODO, watch out for iterators.
-    WASM_UNREACHABLE("unimp");
+    if (this != &other) {
+      this->~InsertOrderedMap();
+      new (this) InsertOrderedMap<Key, T>(other);
+    }
+    return *this;
+  }
+  InsertOrderedMap(InsertOrderedMap&& other)
+    : Map(std::move(other.Map)), List(std::move(other.List)) {}
```
kripken, line 156: Are move semantics guaranteed to get this right? I feel like I've read that they have, but I've never depended on that as much as this line seems to...

test/passes/const-hoisting.txt
```diff
@@ -9,22 +9,22 @@
   (local $5 i32)
   (block
    (local.set $0
-    (i32.const -1048577)
+    (i32.const 8192)
```
kripken, line 12: This does not look right, but reading more carefully, it actually is...

kripken (approved):
Nice!

lgtm, but I'd recommend fuzzing heavily before landing (and with CheckDeterminism as before)

src/support/insert_ordered.h
tlively: Good question. I couldn't find any specific spec guarantees, so I removed this new constructor and assignment operator. They were left over from shots in the dark trying to fix a bug that turned out to be unrelated.

PR #3800, opened 2021-04-12
[Wasm GC] Optimize away unnecessary non-null assertions

kripken:
`ref.as_non_null` is not needed if the value flows into a place that traps
on null anyhow. We replace a trap on one instruction with a trap on
another, but we allow such things (and even changing trap types, which
does not happen here).


tlively (approved):
Nice!

test/example/c-api-kitchen-sink.c
```diff
@@ -654,15 +654,27 @@ void test_core() {
...
+    BinaryenSIMDLoad(module,
+                     BinaryenLoadExtSVec32x2ToVecI64x2(),
+                     0,
+                     8,
+                     makeInt32(module, 128)),
+    BinaryenSIMDLoad(module,
+                     BinaryenLoadExtUVec32x2ToVecI64x2(),
+                     0,
+                     8,
+                     makeInt32(module, 128)),
```
aheejin, line 677: I guess this is not related to this PR? (If we want to clang-format this test file, which we didn't end up doing when we mass-clang-formatted the whole codebase, I am in favor, but we can do that on the whole test files in a separate PR.)

kripken: I think the issue is that on CI we don't check for format on test files, so it allows PRs to land with issues (this fix is for a recent landing). Then when we run clang-format locally it gets fixed up, so I saw it now. But I can remove it from here if you prefer.

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -386,29 +386,23 @@
   ;; CHECK-NEXT:  (drop
   ;; CHECK-NEXT:   (block
   ;; CHECK-NEXT:    (drop
-  ;; CHECK-NEXT:     (ref.as_non_null
-  ;; CHECK-NEXT:      (local.get $data)
-  ;; CHECK-NEXT:     )
+  ;; CHECK-NEXT:     (local.get $data)
```
aheejin, line 389: Why is this optimized? It doesn't have instructions like `struct.get`, no?

src/passes/OptimizeInstructions.cpp
```diff
@@ -1110,6 +1135,8 @@ struct OptimizeInstructions
       return;
     }
 
+    skipNonNullCast(curr->value);
```
aheejin, line 1138: Why do we skip nonnull cast here?

test/example/c-api-kitchen-sink.c
aheejin: Unless not fixing this makes the CI fail (it doesn't right?), I think it'd be good to update the whole test files in a separate PR. I was in favor of it when we mass-formated the whole codebase, but I remember you somehow didn't prefer it.

kripken: It's just that one test file I think - everything else has the proper format. That is, when I run clang-format locally that is the only thing that it decides to change... But, removed from here, I agree it's a distraction.

src/passes/OptimizeInstructions.cpp
kripken: All the RefAs operations will trap on non null anyhow (including RefAsNonNull, funnily...).

test/lit/passes/optimize-instructions-gc.wast
kripken: This began as `(ref.as_func (ref.as_non_null (local.get $data)))`, and the `ref.as_func` will trap on null anyhow, so the `ref.as_non_null` in the middle can be removed.

(And after that, we see that the `ref.as` will always fail, so it drops the input and adds an unreachable.).

aheejin (approved):

PR #3801, opened 2021-04-12
Fuzzer: Distinguish traps from host limitations

kripken:
Host limitations are arbitrary and can be modified by optimizations, so
ignore them. For example, if the optimizer removes allocations then a
host limit on an allocation error may vanish. Or, an optimization that
removes recursion and replaces it with a loop may avoid a host limit
on call depth (that is not done currently, but might some day).

This removes a class of annoying false positives in the fuzzer.


tlively (approved):
Nice!

PR #3802, opened 2021-04-12
Reducer: Replace entire function bodies with either unreachable or nop

kripken:
Previously we just used unreachable. This also tries nop when it is possible,
and sometimes that is better (if the code is called, a nop may be less intrusive
of a change).


tlively (approved):
Does this make a noticeable difference?

kripken:
Probably rare, but I had a case where it helped (the testcase trapped, and replacing the body with an unreachable got confused with that - the reduction script needs to be more precise to avoid that, but this change made that unnecessary).

PR #3803, opened 2021-04-13
[Wasm GC] Full precompute support for GC

kripken:
The precompute pass ignored all reference types, but that was overly
pessimistic: we can precompute some of them, namely a null and a
reference to a function are fully precomputable, etc.

To allow that to work, add missing integration in `getFallthrough` as well.

With this, we can precompute quite a lot of field accesses in the existing
`-Oz` testcase, as can be seen from the output. That testcase runs
`--fuzz-exec` so it prints out all those logged values, proving they have
not changed.


test/lit/passes/precompute-gc.wast
```diff
@@ -0,0 +1,35 @@
...
+ ;; CHECK-NEXT:    )
+ ;; CHECK-NEXT:    (ref.null func)
+ ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT:  (i32.const 1)
+ ;; CHECK-NEXT: )
+ (func $test-fallthrough (result i32)
+  (local $x funcref)
+  (local.set $x
+   ;; the fallthrough value should be used (which is possible as with names
```
tlively, line 21: This comment looks cut off.

tlively (approved):

test/lit/passes/precompute-gc.wast
```diff
@@ -0,0 +1,35 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
+;; RUN: wasm-opt %s --remove-unused-names --precompute-propagate -all -S -o - \
```
tlively, line 2: Why `--remove-unused-names`?

test/lit/passes/precompute-gc.wast
kripken: Oh, heh, that's what was explaining `--remove-unused-names`, fixing...

tlively (approved):
Makes sense!

kripken:
Fuzzed for a few hours, looks ok...

PR #3805, opened 2021-04-14
[Wasm GC] Fix precomputing

kripken:
#3803 was incorrect: the optimizations there are only valid if the referred
data is immutable (otherwise we'd need escape analysis to make sure it is
not modified elsewhere).

The bug happened because we missed a call to `canEmitConstantFor` in
the propagation code, and reading the test outputs I missed that it was
invalid.

Add a bunch more tests that clarify the issues and will prevent regressions.


kripken:
Actually this is still wrong, let me get back to this with a proper fix.

test/lit/passes/precompute-gc.wast
```diff
@@ -34,4 +38,267 @@
...
+   (i32.add
+    (struct.get $struct 0
+     (local.get $x)
+    )
+    (i32.const 1)
+   )
+  )
+ )
+ ;; --fuzz-exec verifies the output of this function, checking that the change
+ ;; makde in modify-gc-heap is not ignored
```
tlively, line 177:
```suggestion
 ;; made in modify-gc-heap is not ignored.
```

Also, the update script has a bug in which it fails to recognize functions that aren't given a `$name`. Are you exploiting that intentionally here to skip the output for this function?

test/lit/passes/precompute-gc.wast
```diff
@@ -34,4 +38,267 @@
...
+  )
+ )
+ ;; CHECK:      (func $load-from-struct-bad-arrive (param $x (ref null $struct))
+ ;; CHECK-NEXT:  (call $log
+ ;; CHECK-NEXT:   (struct.get $struct 0
+ ;; CHECK-NEXT:    (local.get $x)
+ ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT: )
+ (func $load-from-struct-bad-arrive (param $x (ref null $struct))
```
tlively, line 200: Is this just testing that no precomputation happens when the struct is a parameter?

tlively:
Oops! Ignore the review, then :)

test/lit/passes/precompute-gc.wast
```diff
@@ -34,4 +38,267 @@
...
+ (func $load-from-struct
+  (local $x (ref null $struct))
+  (local.set $x
+   (struct.new_with_rtt $struct
+    (i32.const 1)
+    (rtt.canon $struct)
+   )
+  )
+  ;; we don't precompute these, as we don't know if the GC data was modified
+  ;; elsewhere (we'd need immutability or escape analysis)
```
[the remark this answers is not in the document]
kripken: "elsewhere" was intended to mean by a called function, for example. so on the same thread, but outside the current function - which means precompute knows nothing about it.

test/lit/passes/precompute-gc.wast
```diff
@@ -34,4 +38,267 @@
...
+  (call $log
+   (struct.get $struct 0 (local.get $x))
+  )
```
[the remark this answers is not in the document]
kripken: the logs were just meant to avoid the optimizer removing too much (a drop would be removed)

test/lit/passes/precompute-gc.wast
kripken: yes

test/lit/passes/precompute-gc.wast
kripken: Ah, I didn't know that, thanks... I'll fix this for the fixed PR.

Issue #3806, opened 2021-04-14
Release build "wasm-opt" crashes on Android


truboxl:
1.
I am trying to package emscripten for Termux on Android. I ran some core tests provided in emscripten. Some failed while some errored as expected. But the tests result were not repeatable. They often errored out on `wasm-opt`. I ran 3 tests here.
https://gist.github.com/truboxl/02f56ff555998174a7c47268f07192dc

Next, I compiled a small hello world program `emcc hello.c -o hello.html -O2` and also failed with `wasm-opt`. I rebuild binaryen with `RelWithDebInfo` configuration and got gdb backtrace below:
```
Starting program: /data/data/com.termux/files/home/emsdk/binaryen/main_64bit_binaryen/bin/wasm-opt --strip-dwarf --post-emscripten --no-exit-runtime -O2 --low-memory-unused --zero-filled-memory --strip-debug --strip-producers hello.wasm -o hello.wasm --mvp-features
[New LWP 27113]
[New LWP 27114]
[New LWP 27115]
[New LWP 27116]
[New LWP 27117]
[New LWP 27118]
[New LWP 27119]
[New LWP 27120]

Thread 7 "wasm-opt" received signal SIGSEGV, Segmentation fault.
[Switching to LWP 27118]
MixedArena::allocSpace (this=<optimized out>, size=32, align=8)
    at /data/data/com.termux/files/home/emsdk/binaryen/main/src/mixed_arena.h:93
93	      while (myId != curr->threadId) {


#0  MixedArena::allocSpace (this=<optimized out>, size=32, align=8)
    at /data/data/com.termux/files/home/emsdk/binaryen/main/src/mixed_arena.h:93
#1  0x0000007fb641d108 in MixedArena::alloc<wasm::Unary> (this=0x7fffffdfe8)
    at /data/data/com.termux/files/home/emsdk/binaryen/main/src/mixed_arena.h:145
#2  wasm::Builder::makeUnary (this=<optimized out>, op=wasm::EqZInt32, value=<optimized out>)
    at /data/data/com.termux/files/home/emsdk/binaryen/main/src/wasm-builder.h:550
#3  wasm::RemoveUnusedBrs::doWalkFunction(wasm::Function*)::FinalOptimizer::restructureIf(wasm::Block*) (this=this@entry=0x7fb0b21820, curr=curr@entry=0x7fb6ae3e50)
[...]

PR #3807, opened 2021-04-14
Fix lock-free MixedArena allocation

tlively:
To reduce allocator contention, MixedArena transparently forwards allocations to a
thread-local arena by traversing a linked list of arenas, looking for the one
corresponding to the current thread. If the search reaches the end of the linked
list, it allocates a new arena for the current thread and atomically appends it
to the end of the list using a compare-and-swap operation.

The problem was that the previous code used `compare_exchange_weak`, which is
allowed to spuriously fail, i.e. it is allowed to behave as though the original
value is not the same as the expected value, even when it is. In this case, that
means that it might fail to append the new allocation to the list even if the
`next` pointer is still null, which results in a subsequent null pointer
dereference. The fix is to use `compare_exchange_strong`, which is not allowed
to spuriously fail in this way.

Reported in #3806.

### DIFF
--- a/src/mixed_arena.h
+++ b/src/mixed_arena.h
@@ -104,7 +104,7 @@ struct MixedArena {
         if (!allocated) {
           allocated = new MixedArena(); // has our thread id
         }
-        if (curr->next.compare_exchange_weak(seen, allocated)) {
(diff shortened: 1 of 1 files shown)


kripken (approved):
threading is hard...

PR #3808, opened 2021-04-14
[Wasm GC] Do not inline a function with an RTT parameter

kripken:
Inlined parameters become locals, and rtts cannot be handled as locals, unlike
non-nullable values which we can at least fix up. So do not inline functions with
rtt params.

### DIFF
--- a/src/ir/type-updating.h
+++ b/src/ir/type-updating.h
@@ -307,10 +307,15 @@ struct TypeUpdater
 
 namespace TypeUpdating {
 
+// Checks whether a type is valid as a local, or whether
+// handleNonDefaultableLocals() can handle it if not.
+bool canHandleAsLocal(Type type);
+
 // Finds non-nullable locals, which are currently not supported, and handles
 // them. Atm this turns them into nullable ones, and adds ref.as_non_null on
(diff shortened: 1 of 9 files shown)


src/ir/type-updating.h
```diff
@@ -307,10 +307,15 @@ struct TypeUpdater
...
 
+// Checks whether a type is valid as a local, or whether
+// handleNonDefaultableLocals() can handle it if not.
+bool canHandleAsLocal(Type type);
+
 // Finds non-nullable locals, which are currently not supported, and handles
 // them. Atm this turns them into nullable ones, and adds ref.as_non_null on
 // their uses (which keeps the type of the users identical).
-void handleNonNullableLocals(Function* func, Module& wasm);
+// This may also handle other types of nondefaultable locals in the future.
```
aheejin, line 317: Which types do you mean?

aheejin (approved):

tlively (approved):

src/ir/type-updating.h
kripken: If we end up adding `let` or such then we can use that to handle rtts, maybe.

PR #3809, opened 2021-04-14
Fuzzer: Do not fuzz multivalue testcases in initial contents

kripken:
There is a conflict between multivalue and GC, see the details in the
comment. There isn't a good way to get the fuzzer to avoid the combination
of them, and GC is more urgent, so disable multivalue in that area for now.

(This does not disable all multivalue fuzzing - the fuzzer can still emit stuff.
This just disables initial content from test suite having multivalue, which
is enough for now, until the fuzzer can emit more GC things, and then we'll
need to do more.)

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -214,7 +214,7 @@ def pick_initial_contents():
         '--disable-memory64',
         # avoid multivalue for now due to bad interactions with gc rtts in
         # stacky code. for example, this fails to roundtrip as the tuple code
-        # ends up creating stack binary code that needs to spill rtts to locals,
+        # ends up creating stacky binary code that needs to spill rtts to locals,
         # which is not allowed:
         #
         # (module
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -212,6 +212,27 @@ def pick_initial_contents():
         '--disable-exception-handling',
         # has not been fuzzed in general yet
         '--disable-memory64',
+        # avoid multivalue for now due to bad interactions with gc rtts in
+        # stacky code. for example, this fails to roundtrip as the tuple code
+        # ends up creating stack binary code that needs to spill rtts to locals,
+        # which is not allowed:
+        #
+        # (module
+        #  (type $other (struct))
+        #  (func $foo (result (rtt $other))
+        #   (select
+        #    (rtt.canon $other)
+        #    (rtt.canon $other)
+        #    (tuple.extract 1
+        #     (tuple.make
+        #      (i32.const 0)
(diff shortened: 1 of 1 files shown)


scripts/fuzz_opt.py
```diff
@@ -212,6 +212,27 @@ def pick_initial_contents():
         '--disable-exception-handling',
         # has not been fuzzed in general yet
         '--disable-memory64',
+        # avoid multivalue for now due to bad interactions with gc rtts in
+        # stacky code. for example, this fails to roundtrip as the tuple code
+        # ends up creating stack binary code that needs to spill rtts to locals,
```
aheejin, line 217:
```suggestion
        # ends up creating stacky binary code that needs to spill rtts to locals,
```

aheejin (approved):

scripts/fuzz_opt.py
```diff
@@ -212,6 +212,27 @@ def pick_initial_contents():
...
+        #    (rtt.canon $other)
+        #    (tuple.extract 1
+        #     (tuple.make
+        #      (i32.const 0)
+        #      (i32.const 0)
+        #     )
+        #    )
+        #   )
+        #  )
+        # )
```
tlively, line 234: Will the rtts need to be spilled to locals in this example? That isn't obvious to me. Perhaps an example doing a `tuple.extract 0` on a tuple containing rtts would be clearer.

tlively (approved):

scripts/fuzz_opt.py
kripken: I think the tuple issue you mention is caused at the writing stage and not the reading?

Here is the roundtripped (invalid) code here:

```wat
(module
 (type ${} (struct ))
 (type $none_=>_rtt_{} (func (result (rtt ${}))))
 (func $0 (result (rtt ${}))
  (local $0 i32)
  (local $1 i32)
  (local $2 (rtt 0 ${}))
  (select
   (rtt.canon ${})
   (block (result (rtt 0 ${}))
    (local.set $2
     (rtt.canon ${})
    )
    (drop
     (block (result i32)
      (local.set $1
       (i32.const 0)
      )
      (local.set $0
       (i32.const 0)
      )
      (local.get $1)
     )
    )
    (local.get $2)
   )
   (local.get $0)
  )
 )
)
```
The stackiness is generating during reading here: https://github.com/WebAssembly/binaryen/blob/b2c63a9665a9758c50eac60af605f0399f66580f/src/wasm/wasm-binary.cpp#L2570-L2599

tlively: I see, I hadn't thought through the full round trip.

tlively (approved):

PR #3814, opened 2021-04-15
[Wasm GC] Optimize reference identity checks

kripken:
* Note that `ref.cast` has a fallthrough value.
* Optimize `ref.eq` on identical inputs (see notes there).
* To make that practical, add an `equalAndRemovable` helper, and use that
  in a few more places.


src/passes/OptimizeInstructions.cpp
```diff
@@ -1170,6 +1177,39 @@ struct OptimizeInstructions
   // Information about our locals
   std::vector<LocalInfo> localInfo;
 
+  // Check if two inputs to an instruction are equal, and can be removed. This
+  // is simpler than the general question of equality because we know they are
+  // inputs to the same instruction, and so they execute one after the other,
+  // with nothing else in the middle. So all we need to check is that they are
```
aheejin, line 1183: Not sure if I understand... How do we guarantee that `left` and `right` are the inputs to the same instruction? And why do they need to e inputs from the same instruction and execute one after the other in the first place?

src/ir/properties.h
```diff
@@ -259,6 +259,8 @@ inline Expression* getFallthrough(Expression* curr,
     if (!EffectAnalyzer(passOptions, features, tryy->body).throws) {
       return getFallthrough(tryy->body, passOptions, features);
     }
+  } else if (auto* as = curr->dynCast<RefCast>()) {
+    return getFallthrough(as->ref, passOptions, features);
```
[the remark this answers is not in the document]
kripken: The actual data referred to is identical, and the reference to it is the same pointer - only the type changes. ref.cast checks if it has the right type using the RTT, and traps if not, but if the RTT is suitable then the same value is returned,

https://github.com/WebAssembly/binaryen/blob/bd2e8661a31aa02f701e31110108a5f5c194afed/src/wasm-interpreter.h#L1452-L1455

src/passes/OptimizeInstructions.cpp
kripken: We don't guarantee that here, it's a precondition for calling this method.

Basically this is called from simple peephole optimization rules, checking if two inputs to an instruction are identical, and if so, it can be simplified in various ways.

Does that make sense? If not I can try to rephrase the comment?

src/ir/properties.h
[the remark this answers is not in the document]
tlively: Should there be a check for whether we are ignoring implicit traps here?

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -533,4 +533,158 @@
...
+          (call $nothing)
+          (local.get $x)
+        )
+        (block (result eqref)
+          (call $nothing)
+          (local.get $x)
+        )
+      )
+    )
+    ;; allocation prevents optimization
```
tlively, line 653: It might be worth a test that dropped (or otherwise side-effect free) allocations do not prevent the optimization, since the fallthrough is done before the check for allocation instructions.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1634,10 +1674,8 @@ struct OptimizeInstructions
     if (auto* left = binary->left->dynCast<Binary>()) {
       if (auto* right = binary->right->dynCast<Binary>()) {
         if (left->op != right->op &&
-            ExpressionAnalyzer::equal(left->left, right->left) &&
-            ExpressionAnalyzer::equal(left->right, right->right) &&
-            !effects(left->left).hasSideEffects() &&
-            !effects(left->right).hasSideEffects()) {
+            equalAndRemovable(left->left, right->left) &&
+            equalAndRemovable(left->right, right->right)) {
```
tlively, line 1678: This part is a functional change because it now checks only the fallthrough values, right? Does this interact with GC at all?

test/lit/passes/optimize-instructions.wast
```diff
@@ -7811,6 +7811,44 @@
...
+  ;; CHECK-NEXT:    (if (result i32)
+  ;; CHECK-NEXT:     (local.get $x)
+  ;; CHECK-NEXT:     (local.get $x)
+  ;; CHECK-NEXT:     (return)
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:    (local.get $x)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $xor-of-identical-after-fallthrough (param $x i32)
```
tlively, line 7829: This doesn't look like it has anything to do with GC. Is this connected to the `combineOr` change somehow?

src/passes/OptimizeInstructions.cpp
aheejin: I see, but I'm still not sure why it should be a precondition that they should be inputs to the same instruction and there shouldn't be any instructions in between. I get they are currently used only in those context though.

test/lit/passes/optimize-instructions.wast
aheejin: +1 on splitting non-GC changes into a separate PR

src/passes/OptimizeInstructions.cpp
```diff
@@ -1170,6 +1177,39 @@ struct OptimizeInstructions
...
+    }
+    // Reference equality also needs to check for allocation, which is *not* a
+    // side effect, but which results in different references:
+    // repeatedly calling (struct.new $foo)'s output will return different
+    // results (while i32.const etc. of course does not).
+    if (left->type.isRef()) {
+      if (FindAll<StructNew>(left).has() || FindAll<ArrayNew>(left).has()) {
+        return false;
+      }
+    }
```
aheejin, line 1209: I think @tlively also pointed this out in the tests, but this doesn't seem to be able to rule out this case:
```wast
(ref.eq
  (block (result i32)
    (struct.new_default_with_rtt $struct
      (rtt.canon $struct)
    )
    (i32.const 0)
  )
  (block (result i32)
    (struct.new_default_with_rtt $struct
      (rtt.canon $struct)
    )
    (i32.const 0)
  )
)
```
This should be reduced to `(i32.const 1)` but it cannot because they contain allocations.

test/lit/passes/optimize-instructions-gc.wast
kripken: Good idea, adding.

kripken (commented):
Ok, feedback should be addressed. Many new comments, and all non-GC code removed.

aheejin (approved):

PR #3817, opened 2021-04-18
Run spec test all at once after binary transform

martianboy:
#3792 added support for module linking and `(register` command to
`wasm-shell`, but forgot about three problems:

1. Splitting spec tests prevents linking test modules together.
2. Registered modules may still be used in assertions or an `invoke`
3. Modules may re-export imported objects

This PR appends transformed modules after binary checks to a `spec.wast`
file, plus assertion tests and register commands. Then runs `wasm-shell`
on the whole file. It also keeps both the module name and its registered
name available in `wasm-shell` for use in shell commands and linked
modules. Furthermore, it correctly finds the module where an object is
defined even if it is imported and re-exported several times.

The updated version of `imports.wast` spec test is enabled to verify the
fixes.

### DIFF
--- a/scripts/test/support.py
+++ b/scripts/test/support.py
@@ -145,7 +145,7 @@ def split_wast(wastFile):
             ret += [(chunk, [])]
         elif chunk.startswith('(assert_invalid'):
             continue
-        elif chunk.startswith(('(assert', '(invoke')):
+        elif chunk.startswith(('(assert', '(invoke', '(register')):
             # ret may be empty if there are some asserts before the first
             # module. in that case these are asserts *without* a module, which
             # are valid (they may check something that doesn't refer to a module
--- a/src/wasm-interpreter.h
+++ b/src/wasm-interpreter.h
@@ -2422,23 +2422,24 @@ private:
   protected:
     // Returns the instance that defines the memory used by this one.
     SubType* getMemoryInstance() {
-      if (instance.wasm.memory.imported()) {
-        return instance.linkedInstances.at(instance.wasm.memory.module).get();
-      } else {
-        return static_cast<SubType*>(&instance);
+      SubType* inst = static_cast<SubType*>(&instance);
+      while (inst->wasm.memory.imported()) {
+        inst = inst->linkedInstances.at(inst->wasm.memory.module).get();
       }
+      return inst;
     }
 
     // Returns a reference to the current value of a potentially imported global
     Literals& getGlobal(Name name) {
-      auto* global = instance.wasm.getGlobal(name);
-      if (global->imported()) {
-        auto inst = instance.linkedInstances.at(global->module);
+      SubType* inst = static_cast<SubType*>(&instance);
+      auto* global = inst->wasm.getGlobal(name);
+      while (global->imported()) {
+        inst = inst->linkedInstances.at(global->module).get();
         Export* globalExport = inst->wasm.getExport(global->base);
-        return inst->globals[globalExport->value];
-      } else {
-        return instance.globals[name];
+        global = inst->wasm.getGlobal(globalExport->value);
       }
+
+      return inst->globals[global->name];
     }
 
   public:
--- a/src/wasm-interpreter.h
+++ b/src/wasm-interpreter.h
@@ -2422,7 +2422,7 @@ private:
   protected:
     // Returns the instance that defines the memory used by this one.
     SubType* getMemoryInstance() {
-      SubType* inst = static_cast<SubType*>(&instance);
+      auto* inst = instance.self();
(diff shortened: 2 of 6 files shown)


scripts/test/support.py
```diff
@@ -145,7 +145,7 @@ def to_end(j):
             ret += [(chunk, [])]
         elif chunk.startswith('(assert_invalid'):
             continue
-        elif chunk.startswith(('(assert', '(invoke')):
+        elif chunk.startswith(('(assert', '(invoke', '(register')):
```
kripken, line 148: This seems to assume that the format of spec tests is "a module, then some more registered modules for that module, plus asserts/invokes". Is that true? It seems like in theory there could be a few modules up front that are registered and then used by multiple modules later for example.

martianboy: The syntax for `register` is simply `(register "a")`, so it always expects at least a module first. And it's often the case that a module is registered first then is used in subsequent modules. So in this line, I simply wanted to include any `(register` commands, so that we can rebuild the spec test with roughly the same structure. It should not assume any specific structure on spec tests more than it was doing before. I may be missing the point, though...

kripken: Yes, it doesn't assume any more structure, I am just worried it is not enough for future spec tests that need more structure. Specifically, this ends up putting each registered module into one "chunk", but it may be used by multiple ones?

Anyhow, that doesn't seem urgent to figure out in this PR, and maybe my worry is not relevant...

kripken:
I think this PR is good to land if it is a useful incremental step. Does it allow running more spec tests than before? If so is it possible to add one or more so that this is tested?

scripts/test/support.py
martianboy: I may have misunderstood your comment, but just for the sake of clarification, note that this PR won't run spec tests in chunks anymore. It only run `wasm-opt`, `wasm-as` and `wasm-dis` on chunks, then puts the whole thing together again and run `wasm-shell` once on the resulting `spec.wast`, so that all registered modules are available once again.

martianboy:
@kripken Yes, without this PR it's not possible to run spec tests that have `(register`. I was actually going to enable `table_init.wast` test in my other branch and ran into this problem that spec running spec tests in chunks will not allow linking registered modules.

I will try to find a spec test that has `register` but doesn't require significant changes.

martianboy:
@kripken The only spec test I could find that didn't require unrelated changes was `imports.wast`. While running `wasm-shell` on it, I noticed a couple of issues that I hadn't noticed in #3792:

1. Registered modules may still be referenced in an assertion by name. This wasn't supported by `wasm-shell`.
2. Imported objects may have been re-exported imports, but the interpreter wasn't looking beyond the first level.

I fixed both issues here, and with those the spec tests now cover `imports.wast` as well. It includes various combinations of `register` and assertions using registered modules directly or indirectly.

Although there are now more than a single issue are fixed by this PR, all three issues are directly related to #3792 and only together enable covering `imports.wast`. But still, I can split the changes in separate PRs if you think that's better. I will update the PR description after that.

scripts/test/support.py
kripken: Thanks! That's what I was missing. I didn't see the higher-up running of all the spec test at once...

src/wasm-interpreter.h
```diff
@@ -2422,23 +2422,24 @@ template<typename GlobalManager, typename SubType> class ModuleInstanceBase {
   protected:
     // Returns the instance that defines the memory used by this one.
     SubType* getMemoryInstance() {
-      if (instance.wasm.memory.imported()) {
-        return instance.linkedInstances.at(instance.wasm.memory.module).get();
-      } else {
-        return static_cast<SubType*>(&instance);
+      SubType* inst = static_cast<SubType*>(&instance);
```
kripken, line 2425: Why is this static cast needed?

kripken (commented):
Seems ok to me to have this all in one PR, as it's low-risk and it allows proper testing.

src/wasm-interpreter.h
martianboy: `instance` is of type `wasm::ModuleInstanceBase<GlobalManager, SubType>*`, while `linkedInstances` have `SubType*` so it needs casting. But I didn't need to cast directly, as we have a `wasm::ModuleInstanceBase<GlobalManager, SubType>::self()` method that returns does it for us. So I changed this part to use `instance.self()` instead.

martianboy:
Thanks @kripken. I updated the PR description to reflect the new fixes as well.

kripken (approved):
Great, thanks!

PR #3818, opened 2021-04-18
Fix element segment ordering in Print

martianboy:
We used to print active element segments right after corresponding
tables, and passive segments came after those. We didn't print internal
segment names, and empty segments weren't being printed at all. This
meant that there was no way for instructions to refer to those table
segments after round tripping.

This will fix those issues by printing segments in the order they were
defined, including segment names when necessary and not omitting
empty segments anymore.


martianboy:
This came up when testing `elem.drop` in the other branch with optimization passes.

martianboy:
@kripken @tlively PTAL

src/passes/Print.cpp
```diff
@@ -2666,7 +2651,7 @@ struct PrintSExpression : public UnifiedExpressionVisitor<PrintSExpression> {
     doIndent(o, indent);
     o << '(';
     printMedium(o, "elem");
-    if (curr->hasExplicitName) {
+    if (curr->hasExplicitName || currModule->elementSegments.size() > 1) {
```
kripken, line 2654: How about a comment like:
```suggestion
    // If there is no explicit name, and there are multiple segments, use our
    // internal names to differentiate them.
    if (curr->hasExplicitName || currModule->elementSegments.size() > 1) {
```

kripken (approved):
Nice!

tlively:
Why do segments without explicit names need to be differentiated via their internal names? Wouldn't they be referred to by index? I can see that printing segments out of order or skipping empty segments would cause problems with that, and based on the PR description I expected that this PR would fix those issues, but it looks like it works around them by printing internal names instead. Would it be better to fix the printing so we wouldn't have to use internal names?

martianboy:
@tlively The problem arises when a passive segment is referenced by its index in `table.init` and then an optimization pass invalidates that index. See the following:

```
(module
  (table 0 funcref)
  (elem funcref)
  (elem funcref)

  (func $init
    (table.init 1 (i32.const 0) (i32.const 0) (i32.const 0))
  )
)
```

`RemoveUnusedModuleElements` will remove the first segment and leaves the second one that is used in `table.init`, but the index remains unchanged. Now we have a problem. `table.init` is still referring to the segment `$1`. But without printing the segment's internal name, on a second load Binaryen will generate `$0` instead which is now different from its original one.

I understand that this problem is not specific to this case and could happen on any object that can be referenced by index. So perhaps there is a mechanism to fix it that I have missed?

src/passes/Print.cpp
martianboy: Done, thanks!

tlively:
The `Expression` for `table.init` should refer to the segment by name rather than by index so the module remains valid after `RemoveUnusedModuleElements` runs. (MemoryInit still uses an index because memory segments are not yet top-level named Module items like ElementSegments are now.) In general it is also fine for internal names to change when a module is round tripped through either the text or binary formats; I don't think we do anything to avoid that in other cases either.

martianboy:
@tlively `TableInit` does refer to the segment by name, but how can a named reference remain valid without printing the name of the referenced object? Maybe I'm still missing your point, but let me clarify by using my earlier example:

```wat
(module
  (table 0 funcref)
  (elem funcref)
  (elem funcref)

  (func $init
    (table.init 1 (i32.const 0) (i32.const 0) (i32.const 0))
  )
)
```

Without printing internal names, this will be reduced to the following after the optimization passes:

```wat
(module
  (table $0 0 funcref)
  (elem funcref)

  (func $init
    (table.init $1 (i32.const 0) (i32.const 0) (i32.const 0))
  )
)
```

But this is not a valid module, since there is no element segment with an internal name `$1`.

tlively:
Ah, I was thinking that we would assign indices to each element segment before printing and actually print the indices themselves (without `$`) rather than names, but actually just printing the internal names is much simpler. This LGTM, and thanks for bearing with my confusion :)

martianboy:
@tlively oh absolutely, I really appreciate all the great reviews. :)

tlively:
I guess the part of the PR description that still confuses me is "This caused internal names of segments to change after print," since printing shouldn't change the internal names and it is not usually a problem if round-tripping through the text format changes the internal names. I guess the real issue is that skipping empty segments and skipping printing segment names meant that there was no way for instructions to refer to to table segments, given that we aren't calculating indices for them either. Could you clarify that somehow for the commit message?

martianboy:
@tlively Ah now I see where the confusion came from! Sorry, by "after print" I meant after round tripping! Is it better now?

tlively:
Perfect, thanks!

PR #3819, opened 2021-04-19
Implement missing if restructuring

kripken:
The existing restructuring code could turn a block+br_if into an if in
simple cases, but it had some TODOs that I noticed were helpful on
GC benchmarks.

One limitation was that we need to reorder the condition and the value,
```wat
(block
  (br_if
    (value)
    (condition)
  )
  (...)
)
=>
(if
  (condition)
  (value)
  (...)
)
```
The old code checked for side effects in the condition. But it is ok for it
to have side effects if they can be reordered with the value (for example,
if the value is a constant then it definitely does not care about side effects
in the condition).

The other missing TODO is to use a select when we can't use an if:
```wat
(block
  (br_if
    (value)
    (condition)
  )
  (...)
)
=>
(select
  (value)
  (...)
  (condition)
)
```
In this case we do not reorder the condition and the value, but we do
reorder the condition with the rest of the block.

### DIFF
--- a/src/passes/RemoveUnusedBrs.cpp
+++ b/src/passes/RemoveUnusedBrs.cpp
@@ -1011,7 +1011,9 @@ struct RemoveUnusedBrs : public WalkerPass<PostWalker<RemoveUnusedBrs>> {
                   list[0] = &nop;
                   auto canReorder = EffectAnalyzer::canReorder(
                     passOptions, features, br->condition, curr);
-                  auto hasSideEffects = EffectAnalyzer(passOptions, features, curr).hasSideEffects();
+                  auto hasSideEffects =
+                    EffectAnalyzer(passOptions, features, curr)
+                      .hasSideEffects();
                   list[0] = old;
                   if (canReorder && !hasSideEffects) {
                     ExpressionManipulator::nop(list[0]);
--- a/src/passes/RemoveUnusedBrs.cpp
+++ b/src/passes/RemoveUnusedBrs.cpp
@@ -915,7 +915,7 @@ struct RemoveUnusedBrs : public WalkerPass<PostWalker<RemoveUnusedBrs>> {
 
       // Restructuring of ifs: if we have
       //   (block $x
-      //     (br_if $x (cond))
+      //     (drop (br_if $x (cond)))
       //     .., no other references to $x
       //   )
       // then we can turn that into (if (!cond) ..).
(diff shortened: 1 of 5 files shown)


test/lit/passes/remove-unused-brs.wast
```diff
@@ -26,4 +26,261 @@
...
+            (call $nothing)
+            (i32.const 100)
+          )
+          (local.get $x)
+        )
+      )
+      ;; and we can't do a select because of effects here
+      (call $nothing)
+      ;; the condition cannot be reordered with this
+      (i32.const 100)
```
aheejin, line 248: Isn't the instruction the condition cannot be reordered with not this `(i32.const 100)` but `(call $nothing)`?

aheejin (approved):
Are `restructure-br_if-value-effectful-corner-case-N` tests different any meaningful way? They look mostly the same or test the same things to me that there's one effect that prevents it from being converted to `if` and the other effect that prevents it from being converted to `select`. Maybe I'm mistaken.

By the way this lit tests make reviewing really easier; thanks @tlively. I guess I should consider transfer EH tests to lit tests too.

src/passes/RemoveUnusedBrs.cpp
```diff
@@ -956,30 +956,73 @@ struct RemoveUnusedBrs : public WalkerPass<PostWalker<RemoveUnusedBrs>> {
...
+                    ExpressionManipulator::nop(list[0]);
+                    replaceCurrent(
+                      builder.makeIf(br->condition, br->value, curr));
+                  }
+                } else {
+                  // The value has side effects, so it must always execute. We
+                  // may still be able to optimize this, however, by using a
+                  // select:
+                  //   (block $x
+                  //     (br_if $x (value) (cond))
```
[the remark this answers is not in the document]
kripken: Oh, good catch, yes - there should be a drop in the docs here. I'll fix it.

test/lit/passes/remove-unused-brs.wast
kripken: Right, thanks, this was a copy-paste that I forgot to remove, will fix...

kripken:
Oh, I guess the code doesn't check for the block's final element specifically. So maybe the testing is a little more detailed than necessary. Still it seems good to have (maybe future bugs could be prevented).

PR #3820, opened 2021-04-19
[Wasm GC] Reorder ref.as_non_null with tee and cast

kripken:
In both cases doing the `ref.as_non_null` last is beneficial as we have
optimizations that can remove it based on where it is consumed.


src/passes/OptimizeInstructions.cpp
```diff
@@ -1063,6 +1084,36 @@ struct OptimizeInstructions
...
+    // and so reordering does not change whether a trap happens (and reordering
+    // traps is allowed), and does not change the value flowing out at the end.
+    // It is better to have the ref.as_non_null on the outside since it allows
+    // outer instructions to potentially optimize it away (should we find
+    // optimizations that can fold away a ref.cast on an outer instruction, that
+    // might motivate changing this).
+    //
+    // Note that other ref.as* methods, like ref.as_func, are not obviously
+    // worth reordering with ref.cast. For example, the type of ref.as_data is
+    // (ref data), which is less specific than what ref.cast would have.
```
aheejin, line 1107: Why? If we reorder `ref.data` so that the sequence is `(ref.as_non_null (ref.data ..))`, is `ref.as_non_null` cannot be optimized away as in the case of `ref.cast`?

aheejin (approved):

src/passes/OptimizeInstructions.cpp
kripken: Oh, true, but that's with `ref.as_non_null`. In that case we would already optimize out a `ref.as_non_null` that is either before or after a `ref.as_data` (before, then we know the `ref.as_data` will trap anyhow; after, then the type is non-nullable).

The code here is for `ref.cast, where the downside is that `(ref.as_data (ref.cast` would be of type `dataref`, but the opposite order would have a more specific type (from the RTT of the cast).

aheejin: Oh I see, I was confused. Thanks.

PR #3821, opened 2021-04-19
[Wasm GC] Fix the type of ref.as_func, which should be non-nullable

kripken:
It's easy to make a mistake like this because the spec varies on whether
a basic type is nullable or not (funcref is nullable, dataref and i31ref are not)...

It's not easy to test this - perhaps a contrived test could depend on it in
an optimization, but I'm not sure if it's worth it.

### DIFF
--- a/src/wasm/wasm.cpp
+++ b/src/wasm/wasm.cpp
@@ -1028,7 +1028,7 @@ void RefAs::finalize() {
       type = Type(value->type.getHeapType(), NonNullable);
       break;
     case RefAsFunc:
-      type = Type::funcref;
+      type = Type(HeapType::func, NonNullable);
       break;
     case RefAsData:
       type = Type::dataref;


src/wasm/wasm.cpp
```diff
@@ -1028,7 +1028,7 @@ void RefAs::finalize() {
       type = Type(value->type.getHeapType(), NonNullable);
       break;
     case RefAsFunc:
-      type = Type::funcref;
+      type = Type(HeapType::func, NonNullable);
```
tlively, line 1031: `dataref` and `i31ref` are currently specified to be non-nullable, but maybe we should expand them similarly here to be robust against future changes and to make this more obviously correct and consistent to readers who may not remember which abbreviations are nullable.

tlively (approved):

src/wasm/wasm.cpp
tlively: Yes, I think it makes sense to revisit that suggestion, since operating on `funcref` and friends instead of the underlying heap type is almost never what we want. There might be a performance impact, but we have lower hanging fruit when it comes to GC type system performance ;)

kripken: Yes, that's the key thing, I think - we almost never want to work on `funcref` etc. They are mainly a risk for confusion atm...

Ok, let's land this for now and keep thinking about this stuff.

PR #3822, opened 2021-04-19
Optimize if/select with one arm an EqZ and another a 0 or a 1

kripken:
```
(select
  (i32.eqz (X))
  (i32.const 0|1)
  (Y)
)

=>

(i32.eqz
  (select
    (X)
    (i32.const 1|0)
    (Y)
  )
)
```
This is beneficial as the eqz may be folded into something on the outside. I
see this pattern in real-world code, both a GC benchmark (which is why I
noticed it) and it shrinks code size by tiny amounts on the emscripten
benchmark suite as well.


tlively (approved):
Nice!

PR #3825, opened 2021-04-20
Minor wasm-split improvements

tlively:
- Support functions appearing more than once in the table. It turns out we
   were assuming and asserting that functions would appear at most once, but we
   weren't making use of that assumption in any way.

 - Make TableSlotManager::getSlot take a function Name rather than a RefFunc
   expression to avoid allocating and leaking unnecessary expressions.

 - Add and use a Builder interface for building TableElementSegments to make
   them more similar to other module-level items.

cc @martianboy, comments welcome!

Possibly fixes https://github.com/emscripten-core/emscripten/issues/13928, but I'll check to make sure tomorrow.

### DIFF
--- a/src/ir/module-splitting.cpp
+++ b/src/ir/module-splitting.cpp
@@ -54,8 +54,8 @@
 // ref.func instructions, they will have to be modified to use a similar layer
 // of indirection.
 //
-// The code as currently written makes a few assumptions about the module that
-// is being split:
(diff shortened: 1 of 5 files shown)


martianboy:
Nice! Looks good to me.

sbc100 (approved):

kripken (approved):

PR #3826, opened 2021-04-20
Refactor Child*Iterator for simplicity and to allow modifications. NFC

kripken:
Instead of creating a class and doing a traversal, simply use
wasm-delegates-fields to get the children directly. This is simpler and
faster.

This requires a new way to override the behavior by ChildValueIterator.
I ended up using CRTP, but maybe there is a fancier way?

This stores pointers to the children instead of the children. This will
allow modifying the children using these classes (which is the reason
I started this refactoring - next PR will use it). This is also slightly
faster in theory as we avoid doing loads from memory to find the
children (which maybe we never get to doing if the iteration is stopped
early).

Not the goal here, but this speeds up roundtripping by 12%.

This is NFC as nothing uses the new ability to modify things yet.

### DIFF
--- a/src/ir/iteration.h
+++ b/src/ir/iteration.h
@@ -59,13 +59,16 @@ template<class Specific> class AbstractChildIterator {
     Expression*& operator*() {
       assert(index < parent.children.size());
 
-      // The children are in the order of appearance in the IR, which needs to
-      // be reversed for the order of execution.
+      // The vector of children is in reverse order, as that is how
+      // wasm-delegations-fields works. To get the order of execution, reverse
+      // things.
       return *parent.children[parent.children.size() - 1 - index];
     }
   };
 
 public:
+  // The vector of children in the order emitted by wasm-delegations-fields
+  // (which is in reverse execution order).
   SmallVector<Expression**, 4> children;
 
   AbstractChildIterator(Expression* parent) {
--- a/src/ir/iteration.h
+++ b/src/ir/iteration.h
@@ -56,7 +56,13 @@ template<class Specific> class AbstractChildIterator {
 
     void operator++() { index++; }
 
-    Expression*& operator*() { return *parent.children[index]; }
+    Expression*& operator*() {
+      assert(index < parent.children.size());
+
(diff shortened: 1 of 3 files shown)


src/ir/iteration.h
```diff
@@ -55,59 +56,88 @@ template<template<class, class> class Scanner> class AbstractChildIterator {
...
+#define DELEGATE_FIELD_NAME(id, name)
+#define DELEGATE_FIELD_NAME_VECTOR(id, name)
+#define DELEGATE_FIELD_SCOPE_NAME_DEF(id, name)
+#define DELEGATE_FIELD_SCOPE_NAME_USE(id, name)
+#define DELEGATE_FIELD_SCOPE_NAME_USE_VECTOR(id, name)
+#define DELEGATE_FIELD_SIGNATURE(id, name)
+#define DELEGATE_FIELD_TYPE(id, name)
+#define DELEGATE_FIELD_ADDRESS(id, name)
+
+#include "wasm-delegations-fields.h"
```
tlively, line 104: I think it would be good to reverse the children vector at the end because most uses of this end up iterating on `children` directly, iirc. In fact, perhaps we could get rid of `operator*` and `operator++` entirely, since it's not clear to me how those would effectively be used. What would the end condition for the iteration be?

tlively (commented):
Wow, that's a nice speed improvement!

src/ir/iteration.h
kripken: Someone might get the iterator for the end, and start iterating at the start, doing `++` until they reach the end? But I do think we do not have such users atm in the codebase.

Reversing the vector seems slightly risky to me - someone might not iterate through them all (even when iterating in a loop, they can break) so we'd be doing a linear pass (of memory stores, not free) unnecessarily. The subtraction seems cheaper. (Reading the size is not free though.)

tlively: Aha, I had forgotten that in fact this is normally used like `for (auto it : ChildIterator(...)) `. LGTM!

tlively (approved):

PR #3828, opened 2021-04-21
OptimizeInstructions: Move identical unary code out of if/select arms

kripken:
```wat
(select
  (foo
    (X)
  )
  (foo
    (Y)
  )
  (condition)
)
=>
(foo
  (select
    (X)
    (Y)
    (condition)
  )
)
```

To make this simpler, refactor `optimizeTernary` to be templated.


test/lit/passes/optimize-instructions.wast
```diff
@@ -11818,4 +11818,199 @@
...
+  ;; CHECK-NEXT:      (local.get $y)
+  ;; CHECK-NEXT:      (local.get $z)
+  ;; CHECK-NEXT:     )
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $ternary-identical-arms-type-change (param $x f64) (param $y f64) (param $z i32)
+    (drop
+      (select
```
tlively, line 11876: If this were an `if` instead of a `select`, then the type change would be visible in the test output.

test/lit/passes/optimize-instructions.wast
```diff
@@ -11818,4 +11818,199 @@
...
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:    (i32.add
+  ;; CHECK-NEXT:     (local.get $y)
+  ;; CHECK-NEXT:     (local.get $y)
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:    (local.get $z)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $ternary-identical-arms-but-binary (param $x i32) (param $y i32) (param $z i32)
```
tlively, line 12000: Once we have Poppy IR in place to optimize multivalue code, it might actually be worth allowing this to work with higher arity expressions. That would be neat!

tlively (approved):
Very nice! These tests are excellent, too. I like the approach of using very small functions with descriptive names. I left a couple comments, but LGTM if you don't think they should be actionable right now.

src/passes/OptimizeInstructions.cpp
```diff
@@ -2833,12 +2832,78 @@ struct OptimizeInstructions
...
+        // TODO: handle none as well as concrete types (can pull a drop out, for
+        //       example)
+        // TODO: consider the case with more children than 1
+        if (curr->ifTrue->type.isConcrete() &&
+            !Properties::isControlFlowStructure(curr->ifTrue) &&
+            ExpressionAnalyzer::shallowEqual(curr->ifTrue, curr->ifFalse)) {
+          ChildIterator ifTrueChildren(curr->ifTrue);
+          if (ifTrueChildren.children.size() == 1) {
+            // If the expression we are about to move outside has side effects,
+            // we cannot replace two instances with one (and there might be
```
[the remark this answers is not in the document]
kripken: Yes, I think that's possible. I'll add some more cases in a followup.

src/passes/OptimizeInstructions.cpp
```diff
@@ -2833,12 +2832,78 @@ struct OptimizeInstructions
...
+      //    (select
+      //      (X)
+      //      (Y)
+      //      (Z)
+      //    )
+      //  )
+      //
+      // Continue doing this while we can, noting the chain of moved expressions
+      // as we go, then do a single replaceCurrent() at the end.
+      SmallVector<Expression*, 1> chain;
```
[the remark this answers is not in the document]
kripken: I did consider that, but it seemed less simple to me. I suppose it would be more efficient in memory, but it's extremely rare that there is more than 1 item here I suspect...

test/lit/passes/optimize-instructions.wast
kripken: Good point! Fixing.

test/lit/passes/optimize-instructions.wast
kripken: Yeah, I think that might work well together, good point...

kripken:
(also the test was wrong, i mixed up `demote` with `trunc`... only the former actually changes the type. fixed!)

PR #3831, opened 2021-04-21
wasm-metadce: Keep symbols alive if there is any reference to corresponding GOT.mem or GOT.func import

sbc100:
This prevents the DCE of used-via-address-only symbols in emscripten's `MAIN_MODULE=2`
use case which we are starting to use and recommend a lot more.

Part of the fix for https://github.com/emscripten-core/emscripten/issues/13786

### DIFF
--- a/src/tools/wasm-metadce.cpp
+++ b/src/tools/wasm-metadce.cpp
@@ -26,6 +26,7 @@
 
 #include <memory>
 
+#include "asmjs/shared-constants.h"
 #include "ir/element-utils.h"
 #include "ir/module-utils.h"
 #include "pass.h"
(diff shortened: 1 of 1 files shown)


src/wasm.h
```diff
@@ -32,6 +32,7 @@
 #include <string>
 #include <vector>
 
+#include "asmjs/shared-constants.h"
```
kripken, line 35: why is this needed? if the metadce cpp change needs it, include it there maybe?

kripken (commented):
Change looks good aside from the include question.

src/wasm.h
sbc100: Sorry, not sure how that happened!

kripken (approved):

PR #3832, opened 2021-04-21
Create a shasum file for the Windows tarball

SpecLad:
CMake has a `sha256sum` subcommand that can be used to calculate checksums portably. Use it in the release workflow to create checksums on Windows.

For consistency, use it on macOS and Linux too.


kripken (approved):

sbc100 (commented):
sweet.  good to know about that!

PR #3833, opened 2021-04-21
Generalize moving of identical code from if/select arms

kripken:
* Effects are fine in the moved code, if we are doing so on an `if`
  (which runs just one arm anyhow).
* Allow `unreachable`, which lets us hoist `return`s for example.
* Allow `none,` which lets us hoist `drop` and `call` for example. For
  this we also need to be careful with subtyping, as at least `drop`
  is polymorphic, so the child types may not have an LUB (see
  example in code).

Adds a small ShallowEffectAnalyzer child of EffectAnalyzer that
calls `visit` to just do a shallow analysis (instead of `walk` which
walks the children).

### DIFF
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -2859,7 +2859,7 @@ private:
         // Ignore control flow structures (which are handled in MergeBlocks).
         if (!Properties::isControlFlowStructure(curr->ifTrue) &&
             ExpressionAnalyzer::shallowEqual(curr->ifTrue, curr->ifFalse)) {
-          // TODO: consider the case with more children than 1
+          // TODO: consider the case with more than one child.
           ChildIterator ifTrueChildren(curr->ifTrue);
           if (ifTrueChildren.children.size() == 1) {
             ChildIterator ifFalseChildren(curr->ifFalse);
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -2870,8 +2870,9 @@ private:
           if (ifTrueChildren.children.size() == 1) {
             // If the expression we are about to move outside has side effects,
             // then we cannot do so in general with a select: we'd be reducing
-            // the amount of the effects as well as moving them.
-            // TODO: handle certain side effects when possible
+            // the amount of the effects as well as moving them. For an if,
+            // the side effects execute once, so there is no problem.
+            // TODO: handle certain side effects when possible in select
             if (std::is_same<T, If>::value ||
                 !ShallowEffectAnalyzer(getPassOptions(),
                                        getModule()->features,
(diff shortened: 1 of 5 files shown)


src/passes/OptimizeInstructions.cpp
```diff
@@ -2860,27 +2856,42 @@ struct OptimizeInstructions
...
       while (1) {
-        // TODO: handle none as well as concrete types (can pull a drop out, for
-        //       example)
-        // TODO: consider the case with more children than 1
-        if (curr->ifTrue->type.isConcrete() &&
-            !Properties::isControlFlowStructure(curr->ifTrue) &&
+        // Ignore control flow structures (which are handled in MergeBlocks).
+        if (!Properties::isControlFlowStructure(curr->ifTrue) &&
             ExpressionAnalyzer::shallowEqual(curr->ifTrue, curr->ifFalse)) {
+          // TODO: consider the case with more children than 1
```
tlively, line 2862:
```suggestion
          // TODO: consider the case with more than one child.
```

test/lit/passes/optimize-instructions.wast
```diff
@@ -12013,4 +12024,108 @@
...
+  ;; CHECK-NEXT:     (local.get $y)
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $ternary-identical-arms-br_if-same (param $x i32) (param $y i32) (param $z i32)
+    (block $block
+      (if
+        (local.get $z)
+        ;; two br_ifs with different targets are shallowly identical
```
tlively, line 12042:
```suggestion
        ;; two br_ifs with the same target are shallowly identical
```

test/lit/passes/optimize-instructions.wast
```diff
@@ -12013,4 +12024,108 @@
       )
     )
   )
+  ;; CHECK:      (func $ternary-identical-arms-br_if-same (param $x i32) (param $y i32) (param $z i32)
+  ;; CHECK-NEXT:  (block $block
+  ;; CHECK-NEXT:   (br_if $block
+  ;; CHECK-NEXT:    (if (result i32)
+  ;; CHECK-NEXT:     (local.get $z)
+  ;; CHECK-NEXT:     (local.get $x)
+  ;; CHECK-NEXT:     (local.get $y)
```
tlively, line 12033: It looks like we could further simplify `if`s into `select`s whenever the only side effects are reading.

tlively (approved):
Nice 👍

test/lit/passes/optimize-instructions.wast
kripken: I believe we already do that. A read is not a side effect, and RemoveUnusedBrs has a `selectify` method to turn relevant ifs into selects. It has some heuristics based on how expensive executing both sides,

https://github.com/WebAssembly/binaryen/blob/cb776da297a845e15171405b7f7518619122d7aa/src/passes/RemoveUnusedBrs.cpp#L1057-L1063

PR #3834, opened 2021-04-21
[Wasm GC] Skip DeadArgumentElimination of an RTT parameter

kripken:
We could more carefully see when a local is not needed there, but atm
we always add one, and that doesn't work for something nondefaultable.

### DIFF
--- a/src/passes/DeadArgumentElimination.cpp
+++ b/src/passes/DeadArgumentElimination.cpp
@@ -383,8 +383,8 @@ struct DAE : public Pass {
           // replace the parameter with a local).
           // TODO: if there are no references at all, we can avoid creating a
           //       local
-          bool typeIsValid = TypeUpdating::canHandleAsLocal(
-            func->getLocalType(i));
+          bool typeIsValid =
+            TypeUpdating::canHandleAsLocal(func->getLocalType(i));
           if (callParamAreValid && typeIsValid) {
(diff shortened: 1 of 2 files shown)


src/passes/DeadArgumentElimination.cpp
```diff
@@ -373,13 +373,19 @@ struct DAE : public Pass {
           // Great, it's not used. Check if none of the calls has a param with
           // side effects, as that would prevent us removing them (flattening
           // should have been done earlier).
-          bool canRemove =
+          bool callParamAreValid =
```
tlively, line 376:
```suggestion
          bool callParamsAreValid =
```

tlively (approved):

src/passes/DeadArgumentElimination.cpp
```diff
@@ -373,13 +373,19 @@ struct DAE : public Pass {
...
                 .hasSideEffects();
             });
-          if (canRemove) {
+          // The type must be valid for us to handle as a local (since we
+          // replace the parameter with a local).
+          // TODO: if there are no references at all, we can avoid creating a
+          //       local
+          bool typeIsValid =
+            TypeUpdating::canHandleAsLocal(func->getLocalType(i));
+          if (callParamAreValid && typeIsValid) {
```
kripken, line 388:
```suggestion
          if (callParamsAreValid && typeIsValid) {
```

PR #3836, opened 2021-04-22
Run lit tests with pass-debug as well, to check for more errors

kripken:
99% of the diff is whitespace

### DIFF
--- a/check.py
+++ b/check.py
@@ -331,15 +331,21 @@ def run_unittest():
 
 
 def run_lit():
-    lit_script = os.path.join(shared.options.binaryen_bin, 'binaryen-lit')
-    lit_tests = os.path.join(shared.options.binaryen_root, 'test', 'lit')
-    # lit expects to be run as its own executable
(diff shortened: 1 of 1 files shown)


check.py
```diff
@@ -331,15 +331,21 @@ def run_unittest():
...
+        lit_tests = os.path.join(shared.options.binaryen_root, 'test', 'lit')
+        # lit expects to be run as its own executable
+        cmd = [sys.executable, lit_script, lit_tests, '-vv']
+        result = subprocess.run(cmd)
+        if result.returncode != 0:
+            shared.num_failures += 1
+        if shared.options.abort_on_first_failure and shared.num_failures:
+            raise Exception("lit test failed")
+
+    run()
```
tlively, line 345: Running them just once with pass-debug enabled is probably fine.

tlively (approved):

check.py
kripken: Makes sense, done.

PR #3840, opened 2021-04-23
OptimizeInstructions: Do not change unreachability in if/select changes

kripken:
For example,

```wat
(if (result i32)
  (local.get $x)
  (return
    (local.get $y)
  )
  (return
    (local.get $z)
  )
)
```

If we move the returns outside we'd become unreachable, but we should
not make such type changes in this pass (they are handled by DCE and
Vacuum).

(found by the fuzzer)

### DIFF
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -2888,9 +2888,8 @@ private:
             ChildIterator ifFalseChildren(curr->ifFalse);
             auto* ifTrueChild = *ifTrueChildren.begin();
             auto* ifFalseChild = *ifFalseChildren.begin();
-            auto newCurrType =
-              Type::getLeastUpperBound(ifTrueChild->type, ifFalseChild->type);
-            bool validTypes = newCurrType != Type::none;
+            bool validTypes =
+              Type::hasLeastUpperBound(ifTrueChild->type, ifFalseChild->type);
 
             // In addition, after we move code outside of curr then we need to
             // not change unreachability - if we did, we'd need to propagate
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -2909,7 +2909,7 @@ private:
             //  )
             assert(curr->ifTrue->type == curr->ifFalse->type);
             auto newOuterType = curr->ifTrue->type;
-            if ((newOuterType == Type::unreachable) ^
+            if ((newOuterType == Type::unreachable) !=
                 (curr->type == Type::unreachable)) {
               validTypes = false;
             }
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -2874,9 +2874,10 @@ private:
           // TODO: consider the case with more than one child.
           ChildIterator ifTrueChildren(curr->ifTrue);
           if (ifTrueChildren.children.size() == 1) {
-            ChildIterator ifFalseChildren(curr->ifFalse);
             // ifTrue and ifFalse's children will become the direct children of
-            // curr, and so there must be an LUB for curr to have a proper type.
+            // curr, and so there must be an LUB for curr to have a proper new
+            // type after the transformation.
+            //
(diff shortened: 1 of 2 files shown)


src/passes/OptimizeInstructions.cpp
```diff
@@ -2874,20 +2874,47 @@ struct OptimizeInstructions
...
             //    (drop (f64.const 2.0))
             //  )
+            ChildIterator ifFalseChildren(curr->ifFalse);
             auto* ifTrueChild = *ifTrueChildren.begin();
             auto* ifFalseChild = *ifFalseChildren.begin();
-            bool validTypes =
-              Type::hasLeastUpperBound(ifTrueChild->type, ifFalseChild->type);
+            auto newCurrType =
+              Type::getLeastUpperBound(ifTrueChild->type, ifFalseChild->type);
+            bool validTypes = newCurrType != Type::none;
```
tlively, line 2893: Why switch to `getLeastUpperBound` here? `hasLeastUpperBound` is more efficient when the result is not used.

test/lit/passes/optimize-instructions.wast
```diff
@@ -12212,4 +12212,27 @@
...
+  ;; CHECK-NEXT:   (local.get $x)
+  ;; CHECK-NEXT:   (return
+  ;; CHECK-NEXT:    (local.get $y)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:   (return
+  ;; CHECK-NEXT:    (local.get $z)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $if-dont-change-to-unreachable (param $x i32) (param $y i32) (param $z i32) (result i32)
```
tlively, line 12226: In principle this would be fine for `select`s, right? But we choose not to handle them here since DCE will take care of that. Can we add a test showing that case as well?

src/passes/OptimizeInstructions.cpp
kripken: Oh, I think it was clearer to have `newCurrType` to match `newOuterType`. But yeah, not worth it I guess...

test/lit/passes/optimize-instructions.wast
kripken: No, select would work here actually, and we do handle it (no reason not to - it would be more code to avoid). There is already a testcase showing it works for if (when the if has unreachable type), I'll add one for select.

kripken: Oh, actually it doesn't work for select as we execute both sides, and a return has side effects. So for now it's ignored for that reason. Testcase added.

test/lit/passes/optimize-instructions.wast
```diff
@@ -12188,6 +12188,33 @@
...
+  ;; CHECK-NEXT:    (return
+  ;; CHECK-NEXT:     (local.get $y)
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:    (local.get $z)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $ternary-identical-arms-return-select (param $x i32) (param $y i32) (param $z i32) (result i32)
+    (block $block
+      ;; we cannot optimize a select currently as the return has side effects
```
tlively, line 12206: What if it's a select of two blocks that end in unreachable? Not sure whether that's worth testing, though. It's definitely getting into the weeds :)

tlively (approved):

test/lit/passes/optimize-instructions.wast
kripken: We don't do this optimization if the children of the select are control flow structures like blocks (we'd need to handle branches etc. to the block, and other stuff). There is a test for that:

https://github.com/WebAssembly/binaryen/blob/375be0495e5e4e80cdd389bfe70bb16c1cbeccf8/test/lit/passes/optimize-instructions.wast#L12049-L12079

PR #3842, opened 2021-04-23
Followups to #3841

kripken:
If we don't think that preventing copies in assignment makes sense by
itself (since we allow them on construction) then I think we can just
remove the restriction and also the implicit copy constructor.


tlively (commented):
A little belated, but lgtm 😅

PR #3844, opened 2021-04-28
Generic type traversal and fix a LUB bug

### LINKED ISSUE #3843: ReFinalize generates new types, not synchronized with the global store
Testcase (reduced from 800K...):
```wat
(module
 (type $A (struct (field (ref null $C))))
 (type $B (struct (field (ref null $D))))
 (type $C (struct (field (mut (ref $A)))))
 (type $D (struct (field (mut (ref $A))) (field (mut (ref $A)))))

 (func $foo (param $a (ref null $A)) (result (ref null $A))
  ;; the select should have type $A
  (select (result (ref null $A))
   ;; one arm has type $A
   (local.get $a)
   ;; one arm has type $B (a subtype of $A)
   (ref.null $B)
   (i32.const 0)
  )
 )
)
```
STR: `wasm-opt w.wat -all --print --precompute  --print --roundtrip`

The first print is
```wat
(module
 (type $A (struct (field (ref null $C))))
 (type $ref?|$A|_=>_ref?|$A| (func (param (ref null $A)) (result (ref null $A))))
 (type $B (struct (field (ref null $D))))
 (type $C (struct (field (mut (ref $A)))))
 (type $D (struct (field (mut (ref $A))) (field (mut (ref $A)))))
 (func $foo (param $a (ref null $A)) (result (ref null $A))
  (select (result (ref null $A))
   (local.get $a)
   (ref.null $B)
   (i32.const 0)
  )
 )
)
```
Note that the select has the right type. The we precompute and print again, and now get this:
```wat
(module
 (type $A (struct (field (ref null $C))))
 (type $ref?|$A|_=>_ref?|$A| (func (param (ref null $A)) (result (ref null $A))))
 (type $B (struct (field (ref null $D))))
 (type $C (struct (field (mut (ref $A)))))
 (type $D (struct (field (mut (ref $A))) (field (mut (ref $A)))))
 (func $foo (param $a (ref null $A)) (result (ref null $A))
  (select (result (ref null ${ref?|$C|}))
   (local.get $a)
   (ref.null $B)
   (i32.const 0)
  )
 )
)
```
Note how the type went from `$A` to `${ref?|$C|}` - a new type (as can be seen from its lack of a name), structurally identical to `$A`, but different.

Precompute is not to blame here: nothing is precomputed. But Precompute runs ReFinalize which recomputes LUBs.

Finally, the roundtrip fails on
```
Missing type: (struct (field (ref null (struct (field (mut (ref (struct (field ...3)))))))))
wasm-opt: src/wasm/wasm-binary.cpp:529: uint32_t wasm::WasmBinaryWriter::getTypeIndex(wasm::HeapType) const: Assertion `0' failed.
Aborted
```
That happens in writing out the select, so it looks like the type confusion from before leads to us not having that new type in our list, and we don't know how to write it out.

cc @tlively

tlively:
Fixes #3843.

The issue was that during LUB type building, Hopcroft's algorithm was only
running on the temporary HeapTypes in the TypeBuilder and not considering the
globally canonical HeapTypes that were reachable from the temporary HeapTypes.
That meant that temporary HeapTypes that referred to and were equirecursively
equivalent to the globally canonical types were not properly minimized and could
not be matched to the corresponding globally canonical HeapTypes.

The fix is to run Hopcroft's algorithm on the complete HeapType graph, not just
the root HeapTypes. Since there were already multiple implementations of type
graph traversal, this PR consolidates them into a generic type traversal
utility. Although this creates more boilerplate, it also reduces code
duplication and will be easier to maintain and reuse.

Now that Hopcroft's algorithm partitions can contain globally canonical
HeapTypes, this PR also updates the `translateToTypes` step of shape
canonicalization to reuse the globally canonical types unchanged, since they
must already be minimal. Without this change, `translateToTypes` could end up
incorrectly inserting temporary HeapTypes into the globally canonical type
graph. Unfortunately, this change complicates the interface presented by
`ShapeCanonicalizer` because it no longer owns the HeapTypeInfos backing all of
the minimized types. Fixing this is left as future work.


kripken:
> The fix is to run Hopcroft's algorithm on the complete HeapType graph

I may not understand the issue here yet. Why not fix this by canonicalizing the output of the LUB calculation? (That is the only output of the calculation, and either it exists or does not exist already in the global store?)

tlively:
The output of the LUB calculation is already the result of global canonicalization. The output was wrong because the canonicalization itself was wrong.

kripken (approved):
> The output of the LUB calculation is already the result of global canonicalization.

Ah yes, I forgot about the type builder work that goes on in the middle of TypeBounder, it's can't all be done at the end...

This PR looks good to me. I'd considered asking to split the refactoring out, but I think it's nonintrusive and it does make the overall code simpler and smaller even with the actual bugfixes here. The hackish parts here aren't that bad either given they are short, and sounds like you have a plan to address them?

tlively:
Yes, I think the hackish parts will become a nonissue once we improve canonicalization to happen one strongly connected component at a time, which will require merging global canonicalization into shape canonicalization to some extent. Even with this change, global canonicalization becomes a single function, so the only reason not to invoke it directly from the shape canonicalizer is to keep this diff simpler and smaller.

kripken:
This hit a TSan error when it landed,

https://github.com/WebAssembly/binaryen/runs/2464701903

I've not seen that before, and the stack traces look relevant (e.g. `#0 wasm::TypeBuilder::build()`) so I think it might be a real issue introduced or uncovered here.

tlively:
Oh interesting, I'll take a look!

PR #3845, opened 2021-04-28
Save the exit block, if any, in CFGTraversal. NFC

kripken:
This will be used (and tested) in dead store elimination.

### DIFF
--- a/src/cfg/cfg-traversal.h
+++ b/src/cfg/cfg-traversal.h
@@ -48,9 +48,18 @@ struct CFGWalker : public ControlFlowWalker<SubType, VisitorType> {
   // The entry block at the function's start. This always exists.
   BasicBlock* entry;
 
-  // The exit block at the end where control flow leaves the function. This may
-  // not exist if nothing flows out directly (for example, if the function ends
-  // in a return or an unreachable).
+  // The exit block at the end, where control flow reaches the end of the
+  // function. If the function has a control flow path that flows out (that is,
+  // exits the function without a return or an unreachable or such), then that
+  // control flow path ends in this block. In particular, if the function flows
+  // out a value (as opposed to only returning values using an explicit return)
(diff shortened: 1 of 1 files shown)


src/cfg/cfg-traversal.h
```diff
@@ -45,7 +45,13 @@ struct CFGWalker : public ControlFlowWalker<SubType, VisitorType> {
     std::vector<BasicBlock*> out, in;
   };
 
-  BasicBlock* entry; // the entry block
+  // The entry block at the function's start. This always exists.
+  BasicBlock* entry;
+
+  // The exit block at the end where control flow leaves the function. This may
```
tlively, line 51: Maybe "reaches the end" rather than "leaves" to make it clearer that `return` doesn't count.

tlively (approved):

src/cfg/cfg-traversal.h
```diff
@@ -45,7 +45,13 @@ struct CFGWalker : public ControlFlowWalker<SubType, VisitorType> {
...
+  // not exist if nothing flows out directly (for example, if the function ends
+  // in a return or an unreachable).
+  BasicBlock* exit;
```
aheejin, line 54: - If a function ends in a return, is that not considered "flowing out directly"?
- What if there are multiple exit points?

src/cfg/cfg-traversal.h
kripken: Ah, that is better, thanks.

src/cfg/cfg-traversal.h
kripken: I improved the comment now, PTAL. I agree it was not clear enough before. It should address your questions here I hope.

aheejin (approved):
Thanks for the detailed comment!

PR #3847, opened 2021-04-28
UniqueDeferredQueue improvements

kripken:
Add `clear()`.

Add `UniqueNonrepeatingDeferredQueue` which also has the property
that it never repeats values in the output.

Also add unit tests.

This will be used in dead store elimination.

### DIFF
--- a/src/support/unique_deferring_queue.h
+++ b/src/support/unique_deferring_queue.h
@@ -57,6 +57,32 @@ template<typename T> struct UniqueDeferredQueue {
       // skip this one, keep going
     }
   }
+
+  void clear() {
+    while (!data.empty()) {
+      data.pop();
+    }
+    count.clear();
+  }
+};
+
+// As UniqueDeferredQueue, but once an item has been processed through the queue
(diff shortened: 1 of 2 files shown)


src/support/unique_deferring_queue.h
```diff
@@ -57,6 +57,32 @@ template<typename T> struct UniqueDeferredQueue {
       // skip this one, keep going
     }
   }
+
+  void clear() {
+    while (!data.empty()) {
+      data.pop();
+    }
```
aheejin, line 64: How about
```suggestion
    std::queue<int> empty;
    std::swap(data, empty);
```

aheejin (approved):

tlively (approved):
Neat! I look forward to seeing how this will be used.

src/support/unique_deferring_queue.h
kripken: That is nicer, thanks. Weird std::queue doesn't have a `clear()` method...

PR #3848, opened 2021-04-28
Add LocalGraph::equivalent

kripken:
This compares two `local.get`s and checks whether we are sure they are
equivalent, that is, they contain the same value.

This does not solve the general problem, but uses the existing info to get
a positive answer for the common case where two gets only receive values
by a single set, like
```wat
(local.set $x ..)
(a use.. (local.get $x))
(another use.. (local.get $x))
```
If they only receive values from the same single set, then we know it must
dominate them. The only risk is that the set is "in between" the gets, that is,
that the set occurs after one get and before the other. That can happen in
a loop in theory,
```wat
(loop $loop
  (use (local.get $x))
  (local.set $x ..some new value each iteration..)
  (use (local.get $x))
  (br_if $loop ..)
)
```
Both of those gets receive a value from the set, and they may be different
values, from different loop iterations. But as mentioned in the source code,
this is not a problem since wasm always has a zero-initialization value, and
so the first local.get in that loop would have another set from which it can
receive a value, the function entry. (The only way to avoid that is for this
entire code to be unreachable, in which case nothing matters.)

This will be useful in dead store elimination, which has to use this to reason
about references and pointers in order to be able to do anything useful with
GC and memory.

### DIFF
--- a/src/ir/LocalGraph.cpp
+++ b/src/ir/LocalGraph.cpp
@@ -252,33 +252,33 @@ bool LocalGraph::equivalent(LocalGet* a, LocalGet* b) {
   // there is a single set: if the set did not dominate one of the gets then
   // there would definitely be another set for that get, the zero initialization
   // at the function entry, if nothing else.)
-  if (aSets.size() == 1 && bSets.size() == 1) {
-    auto* aSet = *aSets.begin();
-    auto* bSet = *bSets.begin();
-    if (aSet == bSet) {
-      if (!aSet) {
-        // They are both nullptr, indicating the implicit value for a parameter
-        // or the zero for a local.
-        if (func->isParam(a->index)) {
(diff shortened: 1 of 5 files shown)


aheejin (approved):

src/ir/LocalGraph.cpp
```diff
@@ -244,6 +244,43 @@ LocalGraph::LocalGraph(Function* func) {
...
+          return func->getLocalType(a->index) == func->getLocalType(b->index);
+        }
+      } else {
+        // They are both the same actual set.
+        return true;
+      }
+    }
+    return false;
+  }
+  return false;
```
tlively, line 278: This all would be easier to read as a sequence of early returns rather than as nested `if`s, IMO.

tlively (commented):
Looks good besides that one comment!

PR #3849, opened 2021-04-29
wasm-reduce: Always decrease the factor

kripken:
When things go well, the reducer shrinks the factor by 50% or more, but
when things are slow it kept the factor unchanged. That is annoying in
some cases where you really have no benefit from reduction until the
factor gets small. So this at least reduces it by 10% in each iteration.

On the large wasm GC testcase I recently reduced this helped a lot. I
don't have a guarantee it will help elsewhere, though... but seems low risk.

### DIFF
--- a/src/tools/wasm-reduce.cpp
+++ b/src/tools/wasm-reduce.cpp
@@ -1359,13 +1359,18 @@ int main(int argc, const char* argv[]) {
     }
     lastPostPassesSize = newSize;
 
-    // if destructive reductions lead to useful proportionate pass reductions,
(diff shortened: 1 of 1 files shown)


aheejin (approved):
Maybe add some comments?

src/tools/wasm-reduce.cpp
```diff
@@ -1366,6 +1366,7 @@ int main(int argc, const char* argv[]) {
     if (passProgress >= 4 * lastDestructiveReductions) {
       // don't change
```
tlively, line 1367: This comment in particular looks stale.

PR #3850, opened 2021-04-29
[Wasm GC] Add a print limit in TypePrinter

kripken:
This is similar to the limit in `TypeNamePrinter` in `Print.cpp`. This limit
affects the printed type when debugging with `std::cout << type` etc.,
which just prints the structure and not the name.

This made debugging a recent wasm GC testcase much easier.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -209,6 +209,22 @@ struct TypePrinter {
 
 private:
   template<typename T, typename F> std::ostream& printChild(T curr, F printer);
+
+  // FIXME: This hard limit on how many times we call print() avoids extremely
+  //        large outputs, which can be inconveniently large in some cases, but
+  //        we should have a better mechanism for this.
+  static const size_t MaxPrints = 100;
+
+  size_t prints = 0;
+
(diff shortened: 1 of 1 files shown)


src/wasm/wasm-type.cpp
```diff
@@ -1502,6 +1518,9 @@ bool TypeBounder::lub(const Rtt& a, const Rtt& b, Rtt& out) {
 
 template<typename T, typename F>
 std::ostream& TypePrinter::printChild(T curr, F printer) {
+  if (exceededLimit()) {
+    return os << "..!";
```
aheejin, line 1522: How is this going to be printed? `exceededLimit` also prints `?`, so..

tlively: Oh yeah, we don't need both prints. I like this one better FWIW. Otherwise LGTM.

tlively (approved):

src/wasm/wasm-type.cpp
kripken: Oh, good point... fixed.

tlively (approved):

aheejin (approved):

PR #3852, opened 2021-04-29
Fix data races in type canonicalization

tlively:
Fix two potential sources of data races identified with the help of thread
sanitizer.

First, keep a lock on the global HeapType store as long as it can reach
temporary types to ensure that no other threads observe the temporary types, for
example if another thread concurrently constructs a new HeapType with the same
shape as one being canonicalized here. This cannot happen with Types because
they are hashed in the global store by pointer identity, which has not yet
escaped the builder, rather than shape.

Second, in the shape canonicalizer, do not replace children of the new, minimal
HeapTypeInfos if they are already canonical. Even though these writes are always
no-ops, they still race because they are visible to other threads via canonical
Types.

Fix two sources of potential data races. First, acquire locks on the global Type and HeapType stores during global canonicalization to prevent other threads from observing temporary types. This could happen, for example,

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -500,7 +500,7 @@ bool HeapTypeInfo::operator==(const HeapTypeInfo& other) const {
 }
 
 template<typename Info> struct Store {
-  std::mutex mutex;
+  std::recursive_mutex mutex;
 
   // Track unique_ptrs for constructed types to avoid leaks.
   std::vector<std::unique_ptr<Info>> constructedTypes;
@@ -554,7 +554,7 @@ typename Info::type_t Store<Info>::canonicalize(const Info& info) {
   if (info.getCanonical(canonical)) {
     return canonical;
   }
-  std::lock_guard<std::mutex> lock(mutex);
+  std::lock_guard<std::recursive_mutex> lock(mutex);
   auto indexIt = typeIDs.find(std::cref(info));
   if (indexIt != typeIDs.end()) {
     return typename Info::type_t(indexIt->second);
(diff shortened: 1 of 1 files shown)


kripken:
>  Even though these writes are always no-ops, they still race because they are visible to other threads via canonical
Types.

How can no-op writes be visible? Do you mean something other than "write the same value that is already there" by "no-op"?

tlively:
Well TSan doesn't know that the value will always be the same, so at the very least that change silences a TSan warning. I suspect that it is technically still UB even if the written value is always the same as the value that's already there, but I don't think it could ever cause problems in practice given that this is immutable data that lives for the lifetime of the program.

tlively:
As for how they are visible, here's the setup: The temporary HeapType contains a canonical Type that itself contains a canonical HeapType. That same canonical Type might be contained in other canonical HeapTypes that are being read by other threads, so the child HeapType is visible to other threads, even while it is a (indirect) child of the temporary HeapType.

kripken (approved):
Interesting, thanks.

Makes sense. I also wonder if this might fix possible tearing issues. But just to avoid UB and TSan false positives is worth it.

Did you measure speed?

tlively:
No, but I could imagine this might be bad for speed because it introduces a critical section into LUB calculation. Is there a particular workload you would suggest testing? I also added a tiny bit to avoid going through canonicalization when it's not necessary (but it will usually be necessary).

kripken:
About benchmarking: We may need to wait for more real-world code, or fuzzer support for adding new types (but not sure it's worth doing that before wasm GC typing settles down). But I'd think something like running ReFinalize on many cores on many functions that have lots of LUB calculations in them (lots of ifs/selects on reference types).

tlively:
Thankfully this change will not affect LUBs on non-GC modules because they don't use types in subtyping hierarchies, so e.g. Emscripten should not be affected. Also, future work to perform canonicalization one strongly connected component at a time should be able to eliminate having temporary types reachable from global stores entirely and allow us to remove this extra locking.

PR #3855, opened 2021-05-03
Add an ExpressionReplacer utility

kripken:
This will be useful in dead store elimination.

### DIFF
--- a/src/ir/replacer.h
+++ b/src/ir/replacer.h
@@ -22,7 +22,19 @@
 
 namespace wasm {
 
-// A map of all replacements to perform, and a walk to perform them.
+// A map of replacements to perform on expressions. After filling the map,
+// simply do a walk(..) and they will be replaced.
+//
+// This is useful for a general replacement of expressions. Some passes can
+// store pointers to expressions and update only those, but then there is the
+// nesting problem,
+//
+//  (foo
+//   (bar
+//    (baz
+//
+// If we replace bar with something then the pointer to baz may change. Doing
+// the replacements in the order of a walk, as is done here, is always safe.
 struct ExpressionReplacer
   : PostWalker<ExpressionReplacer,
                UnifiedExpressionVisitor<ExpressionReplacer>> {
--- a/src/ir/replacer.h
+++ b/src/ir/replacer.h
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2021 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


src/ir/replacer.h
```diff
@@ -0,0 +1,53 @@
...
+//  (foo
+//   (bar
+//    (baz
+//
+// If we replace bar with something then the pointer to baz may change. Doing
+// the replacements in the order of a walk, as is done here, is always safe.
+struct ExpressionReplacer
+  : PostWalker<ExpressionReplacer,
+               UnifiedExpressionVisitor<ExpressionReplacer>> {
+  std::unordered_map<Expression*, Expression*> replacements;
```
tlively, line 41: It might make sense for this to be a const reference to a map, since there should never be a reason to copy the map into the replacer object.

tlively (commented):
It would be good to add a constructor and possibly a static function/method that handles instantiating the `ExpressionReplacer` and conducting the walk. It would also be good to add some basic testing.

src/ir/replacer.h
```diff
@@ -0,0 +1,53 @@
...
+//
+// This is useful for a general replacement of expressions. Some passes can
+// store pointers to expressions and update only those, but then there is the
+// nesting problem,
+//
+//  (foo
+//   (bar
+//    (baz
+//
+// If we replace bar with something then the pointer to baz may change. Doing
```
aheejin, line 36: I'm not sure if I understand. If we replace `bar` with something new, which doesn't point `baz`, why does the pointer to `baz` matter?

src/ir/replacer.h
kripken: Actually the use I was envisioning here was to not even need to write the type of the map twice (which creating it outside and then passing it in would require). That is, as the comment says,
```
+// A map of replacements to perform on expressions. After filling the map,
+// simply do a walk(..) and they will be replaced.
```
IOW the idea is to create an instance, fill the map on it, and call walk. Something like:
```
ExpressionReplacer replacer
replacer.replacements[foo] = bar
replacer.walk()
```

kripken: If you prefer I can leave this in the dead store elimination PR, which will show such a use?

src/ir/replacer.h
kripken: Say we want to replace baz with baz2. Then we need to make the pointer to baz point to baz2.

if we have `bar->baz` (bar points to baz as its single child), and we want to have `bar2->baz2`, then if we replace bar first we end up with `bar2->baz` (a new object bar2 that also points to baz), and then we need to modify bar2's pointer to its child, not bar. (so ifwe saved a pointer to `&bar->baz` , that would be wrong)

PR #3859, opened 2021-05-05
Refactor getFallthrough() to use a loop instead of recursion. NFC

kripken:
A new `getImmediateFallthrough` is called in the loop.

Aside from this being more efficient than recursion,  the new method will
be used in escape analysis.


aheejin (approved):
Nice!

tlively (approved):

PR #3860, opened 2021-05-05
BranchUtils additions

kripken:
Add `operateOnScopeNameUsesAndSentValues` which is like the
existing utilities, but provides the child expression that is sent as a
value on the branch.

Also provide a `BranchTargets` utility that maps target names to the
control flow structure for them.

Both will be used and tested in escape analysis.

### DIFF
--- a/src/ir/branch-utils.h
+++ b/src/ir/branch-utils.h
@@ -96,9 +96,9 @@ void operateOnScopeNameUsesAndSentValues(Expression* expr, T func) {
     // There isn't a delegate mechanism for getting a sent value, so do a direct
     // if-else chain. This will need to be updated with new br variants.
     if (auto* br = expr->dynCast<Break>()) {
-      func(name, br->value ? br->value : nullptr);
+      func(name, br->value);
     } else if (auto* sw = expr->dynCast<Switch>()) {
-      func(name, sw->value ? sw->value : nullptr);
+      func(name, sw->value);
     } else if (auto* br = expr->dynCast<BrOn>()) {
       func(name, br->ref);
     } else {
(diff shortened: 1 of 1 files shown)


aheejin (approved):

src/ir/branch-utils.h
```diff
@@ -88,6 +88,25 @@ void operateOnScopeNameUsesAndSentTypes(Expression* expr, T func) {
...
 
+// Similar to operateOnScopeNameUses, but also passes in the expression that is
+// sent if the branch is taken. nullptr is given if there is no value.
+template<typename T>
+void operateOnScopeNameUsesAndSentValues(Expression* expr, T func) {
+  operateOnScopeNameUses(expr, [&](Name& name) {
+    // There isn't a delegate mechanism for getting a sent value, so do a direct
+    // if-else chain. This will need to be updated with new br variants.
+    if (auto* br = expr->dynCast<Break>()) {
+      func(name, br->value ? br->value : nullptr);
```
tlively, line 99: I don't think the ternary expression adds anything here or below.

src/ir/branch-utils.h
```diff
@@ -333,6 +352,27 @@ class BranchSeekerCache {
   }
 };
 
+// Provides a map of branch target names to the expression that corresponds to
+// them.
```
tlively, line 356: This is clearer because the previous wording might have described the target block expressions as well.

```suggestion
// Provides a map of branch target names to the expressions that branch to
// them.
```

tlively (approved):

src/ir/branch-utils.h
kripken: Oh good point, fixed...

PR #3862, opened 2021-05-06
Fix typo in function name: BinayenElementSegmentIsPassive

pmatos:
Becomes BinaryenElementSegmentIsPassive


martianboy:
Thanks for the catch. Can you please make sure it's used correctly in the JS APIs as well? (If it's not currently used there then all is good, too.)

tlively (approved):
Nice catch! I don't see any other uses of these misspellings, so I'll go ahead and land this.

PR #3863, opened 2021-05-06
Fix interpreting of a ref.cast of a function that is not on the module

kripken:
Binaryen allows optimizing functions in function-parallel passes while the
module is still being built, that is, while not all the other functions have
even been added to the module yet. Since the removal of asm2wasm that
has not been heavily tested, but the fuzzer found a closely related bug:
in passes like `inlining-optimizing`, that inline and then optimize the
functions we inlined into, the mechanism for optimizing only the relevant
functions is to create a module with only some of them. (We only want to
optimize the relevant ones, that we inlined into, because this happens
after the main optimization pipeline - we don't want to re-optimize all the
functions if we just inlined into one of them.)

The specific bug here is that `ref.cast` of a funcref looked up the target
function on the module (in order to get its signature, to see if the cast
has the right RTT for it). The fix is to return a nonconstant flow in that
case, as it is something we cannot precompute. (This does mean we
may miss some optimization opportunities, but as in the case of where
we optimize functions before the module is fully built up, we do still
get 99% of function-local optimizations that way, and a subsequent
round of full optimizations can be done later if necessary.)


tlively (approved):
LGTM! It might also be a good idea to add a way to run passes on a subset of functions without creating a new module to avoid this problem. Perhaps we could have a wrapper pass that is parameterized with a normal pass to run and a list of functions to run on?

kripken:
Interesting, yeah, that might be worth doing.

OTOH the current approach does force some testing of optimizing without a full module... so maybe the question is do we still care about that. It was a big speedup for asm2wasm, but with wasm-opt it's kind of hard to implement. Maybe it's simpler to just remove.

PR #3864, opened 2021-05-06
Fuzzer: Ignore things we should ignore even if the process succeeded

kripken:
(can ignore whitespace to remove most of the diff)

We only ignored known issues if the process failed. However, some things
do not bring the process down, but are still necessary to ignore, which this
fixes.

Another approach might be to make all the things we need to ignore fail
the entire process. However, that could be annoying for other debugging:
we don't want a host limit on say hitting a VM limit on recursion to bring
down the entire process, as those limits manifest as traps, and we can
still run after them (and do need to test that). The specific host limit that
made me fix this was the trap on OOM when trying to allocate an array
of size 4GB.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -370,25 +370,31 @@ def fix_spec_output(out):
 
 
 def run_vm(cmd):
-    # ignore some types of errors
-    known_issues = [
-        # can be caused by flatten, ssa, etc. passes
-        'local count too large',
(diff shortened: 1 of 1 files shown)


tlively (approved):
I suspect this could be refactored to avoid running the workload twice, but I'm not sure it would be worth it.

kripken:
I considered that too, yeah. As this only happens on an error, it's rare. And the code is shorter this way I think.

PR #3865, opened 2021-05-06
[Memory64] further memory limit fixes

aardappel:
That were somehow missed.. triggered by emscripten tests


aardappel:
WiP.. not sure what tests would be required.

src/wasm/wasm-binary.cpp
```diff
@@ -107,10 +107,18 @@ void WasmBinaryWriter::writeResizableLimits(
...
-    o << U32LEB(maximum);
+  if (is64) {
+    o << U64LEB(initial);
+    if (hasMaximum) {
+      o << U64LEB(maximum);
+    }
+  } else {
+    o << U32LEB(initial);
+    if (hasMaximum) {
+      o << U32LEB(maximum);
```
kripken, line 118: If this is common perhaps we could add a `UMemoryLEB()` etc., but if it's just here then maybe not worth it.

src/passes/Memory64Lowering.cpp
```diff
@@ -98,6 +98,9 @@ struct Memory64Lowering : public WalkerPass<PostWalker<Memory64Lowering>> {
     }
     // This is visited last.
     memory->indexType = Type::i32;
+    if (memory->hasMax() && memory->max > Memory::kMaxSize32) {
+      memory->max = Memory::kMaxSize32;
+    }
   }
```
kripken, line 104: This one seems like it could be tested in a new small module in `./test/passes/memory64-lowering_enable-memory64_enable-bulk-memory.wast` ?

src/wasm/wasm-binary.cpp
```diff
@@ -107,10 +107,18 @@ void WasmBinaryWriter::writeResizableLimits(
...
+    }
```
kripken, line 119: To test this one, can use `--roundtrip` in the test's passes, and use an offset too large to fit in 32bit, perhaps.

src/wasm/wasm-binary.cpp
aardappel: searching for `U64LEB` this is so far the only location that is conditional on Memory64.

aardappel:
Added tests..

aardappel:
Depends on https://github.com/WebAssembly/binaryen/pull/3974

kripken (approved):

PR #3870, opened 2021-05-10
Fix link to Contributing.md

pmatos:
The `design` repo is still using `master` for it's default branch.


kripken:
Thanks!

Alternatively maybe we can just update the design repo, and not do anything here... https://github.com/WebAssembly/design/issues/1417

kripken:
The design repo's default branch has moved, so our current link has become the correct one :partying_face: 

And so I think we can close this. Thanks for noticing the issue @pmatos !

PR #3871, opened 2021-05-10
Fuzz with maximal inlining some of the time

kripken:


scripts/fuzz_opt.py
```diff
@@ -1055,6 +1055,14 @@ def randomize_opt_flags():
...
     if random.random() < 0.05:
         ret += ['--converge']
+    # possibly inline all the things as much as possible. inlining that much may
+    # be realistic in some cases (on GC benchmarks it is very helpful), but
+    # also, inlining so much allows other optimizations to kick in, which
+    # increases coverage
+    # (the specific number here doesn't matter, but it is far higher than the
+    # wasm limitation on function body size which is 128K)
+    if random.random() < 0.5:
+        ret += ['-fimfs=99999999']
```
aheejin, line 1065: How much inlining does this cause? Does this try to inline everything into everything? If so and if we do that 50% of the time, would it be making the runs too long?

kripken: In general it will try a lot more, but we still won't inline things that don't seem worth it, like things with loops, and recursion. So it's not as extreme as it could be.

Good question about the time... I did some measurements now, and I see a slowdown of around 2%. Seems not bad.

aheejin: Sounds good!

aheejin (approved):

PR #3874, opened 2021-05-10
[Wasm GC] Fix StructSet::finalize on an unreachable value

kripken:
Also improve the testing: put each unreachable expression in its
own function (as otherwise everything after the first is dead).

### DIFF
--- a/src/wasm/wasm.cpp
+++ b/src/wasm/wasm.cpp
@@ -977,7 +977,7 @@ void StructGet::finalize() {
 }
 
 void StructSet::finalize() {
-  if (ref->type == Type::unreachable) {
+  if (ref->type == Type::unreachable || value->type == Type::unreachable) {
     type = Type::unreachable;
   } else {
     type = Type::none;
--- a/test/heap-types.wast
+++ b/test/heap-types.wast
@@ -258,12 +258,18 @@
       )
     )
   )
-  (func $unreachables
+  (func $unreachables-1
     (drop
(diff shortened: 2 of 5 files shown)


src/passes/Print.cpp
```diff
@@ -2346,6 +2346,33 @@ struct PrintSExpression : public UnifiedExpressionVisitor<PrintSExpression> {
...
+    }
+    visitExpression(curr);
+  }
+  void visitStructGet(StructGet* curr) {
+    if (curr->ref->type == Type::unreachable) {
+      printUnreachableReplacement(curr);
+      return;
+    }
+    visitExpression(curr);
+  }
```
aheejin, line 2375: Aren't these handled by `PrintExpressionContents::visitStructGet` and `PrintExpressionContents::visitStructSet`? The call to `visitExpression` will call them I think..

aheejin (approved):

src/passes/Print.cpp
kripken: We have to override the default `visitExpression` here, in the case of unreachable children, so that we can fully handle that case, using `printUnreachableReplacement` to print the children with drops, and enclose them with a block. Otherwise we don't know how to print the Struct operation without an RTT.

Print `PrintExpressionContents` does something similar, but it doesn't print children, it just just prints the contents of the current node. So the code there is not enough, without the new code here we can't print the children properly.

PR #3881, opened 2021-05-12
[Wasm GC] Heap2Local: Handle branches

kripken:
If we branch to a block, and there are no other branches or a final
value on the block either, then there is no mixing, and we may be able
to optimize the allocation. Before this PR, all branches stopped us.

To do this, add some helpers in BranchUtils.

The main flow logic in Heap2Local used to stop when we reached a
child for the second time. With branches, however, a child can flow both to
its immediate parent, and to branch targets, and so the proper thing to look
at is when we reach a parent for the second time (which would definitely
indicate mixing).

Tests are added for the new functionality. Note that some existing tests
already covered some things we should not optimize, and so no tests
were needed for them. The existing ones are: `$get-through-block`,
`$branch-to-block`.


test/lit/passes/heap2local.wast
```diff
@@ -239,11 +239,13 @@
 
   ;; CHECK:      (func $ignore-unreachable
   ;; CHECK-NEXT:  (drop
-  ;; CHECK-NEXT:   (block
-  ;; CHECK-NEXT:    (struct.new_with_rtt $struct.A
-  ;; CHECK-NEXT:     (i32.const 2)
-  ;; CHECK-NEXT:     (unreachable)
-  ;; CHECK-NEXT:     (rtt.canon $struct.A)
+  ;; CHECK-NEXT:   (block ;; (replaces something unreachable we can't emit)
+  ;; CHECK-NEXT:    (drop
```
kripken, line 243: this diff fixes a merge conflict on `main` due to recent landings.

src/passes/Heap2Local.cpp
```diff
@@ -444,10 +447,10 @@ struct Heap2LocalOptimizer {
       // look at something that another allocation reached, which would be in a
       // different call to this function and use a different queue (any overlap
       // between calls would prove non-exclusivity).
-      if (seen.count(child)) {
+      if (seen.count(parent)) {
```
aheejin, line 450: Is it not possible that
```
  GP
 /  \
P1   P2
 \  /
   C
```
C flows the same value to P1 and P2 so even though GP are visited twice by two different passes but they end up in the same value?

> what this check does is verify we don't look at something that another allocation reached, which would be in a different call to this function and use a different queue (any overlap between calls would prove non-exclusivity).

Not sure about this either... What do you mean by "another allocation"? Aren't we dealing with a single allocation in this function?

test/lit/passes/heap2local.wast
```diff
@@ -1530,6 +1532,161 @@
...
+          ;; A branch to the block of our allocation.
+          (br_if $block
+            (local.get $0)
+            (i32.const 0)
+          )
+        )
+        (drop
+          ;; Another branch, causing mixing that prevents optimizations.
+          (br_if $block
+            (ref.null $struct.A)
```
aheejin, line 1638: What if this is `(local.get $0)`? We don't check for merging right?

test/lit/passes/heap2local.wast
```diff
@@ -1530,6 +1532,161 @@
...
+  (func $br_if_flow (result f64)
+    (local $0 (ref null $struct.A))
+    (local.set $0
+      (struct.new_default_with_rtt $struct.A
+        (rtt.canon $struct.A)
+      )
+    )
+    (struct.get $struct.A 1
+      (block $block (result (ref null $struct.A))
+        ;; The branch here is optimizable in principle. However, the br_if also
```
aheejin, line 1677: Nit: `in principle` -> `if were not for the call_ref` or something..? "In principle" sounds little confusing to me.

src/passes/Heap2Local.cpp
kripken: Yes, a diamond like that is possible, but not handled by this pass yet. There is a TODO on (new) lines 634-5.

We deal with a single allocation in this function, yes, but we call it multiple times on different allocations (that's why the comment mentions "different call" - maybe not clear enough?). After we have seen one node be used exclusively by allocation A, it is not possible for it to be used exclusively by allocation B, so we stop.

test/lit/passes/heap2local.wast
kripken: Correct, that would not work yet, until we support merges. (I don't plan to support them soon though, as it's not a common pattern I hope, at least after other optimizations run.)

kripken: (added a test now)

src/passes/Heap2Local.cpp
aheejin: Ah, I didn't realize `seen` is not a local variable to this function but a member variable. Thanks.

aheejin (approved):

PR #3882, opened 2021-05-12
Add pass argument sugar to commandline

kripken:
We have `--pass-arg` that allows sending an argument to a pass, like this:
```
wasm-opt --do-stuff --pass-arg=do-stuff@FUNCTION_NAME
```
With this PR that is equivalent to this:
```
wasm-opt --do-stuff=FUNCTION_NAME
```
That is,one can just give an argument to a pass on the commandline.

This fixes the Optional mode in `command-line.h/cpp`. That was not
actually used anywhere before this PR.

Also rename `--extract-function`'s pass argument to match it. That is, the usage
used to be
```
wasm-opt --extract-function --pass-arg=extract@FUNCTION_NAME
```
Note how the pass name differed from the pass-arg name. This changes it to
match. This is a breaking change, but I doubt this is used enough to justify
any deprecation / backwards compatibility effort, and any usage is almost
certainly manual, and with PR writing it manually becomes easier as one
can do
```
wasm-opt --extract-function=FUNCTION_NAME
```

The existing test for that is kept (&renamed), and a new test added to test the
new notation.

This is a step towards unifying the symbol map functionality between
wasm-as and wasm-opt (later PRs will turn the symbol mapping pass into
a pass that receives an argument).


src/tools/optimization-options.h
```diff
@@ -202,12 +202,28 @@ struct OptimizationOptions : public ToolOptions {
...
+                  //
+                  //   --foo --pass-arg=foo@ARG
+                  Options::Arguments::Optional,
+                  [this, p](Options*, const std::string& arg) {
+                    if (!arg.empty()) {
+                      if (passOptions.arguments.count(p)) {
+                        Fatal()
+                          << "Cannot pass multiple pass arguments to " << p;
+                      }
+                      passOptions.arguments[p] = arg;
```
tlively, line 223: I've been a big fan of `insert` to reduce map lookups recently, but if you think the other way is more readable, that's probably fine, too.

```suggestion
                      if (!passOptions.arguments.insert({p, arg}).second) {
                        Fatal()
                          << "Cannot pass multiple pass arguments to " << p;
                      }
```

tlively (approved):
LGTM modulo optional comment. Also, replacing these tests with a lit test might be nice because you could have just a single file with multiple RUN lines rather than having to duplicate files.

src/tools/optimization-options.h
kripken: I'd like `insert` more if the `.second` were `.inserted`... as things stand it requires mental work when I read such a line. Just me?

kripken:
Can a lit test do this? That is, write to a temp file, then match the contents there?

src/tools/optimization-options.h
tlively: Yes, totally agree that it's a terrible API. My hope is that if we use it more, we won't need to think about it as much when we see it. I don't think it's important in this particular case, though.

tlively:
It can write to a temp file, but is that even necessary? I image you could do something like this:

```
;; RUN: wasm-opt %s -S -o - --extract-function --pass-arg=extract-function@foo | filecheck %s
;; RUN: wasm-opt %s -S -o - --extract-function=foo | filecheck %s
```

src/tools/optimization-options.h
kripken: Fair point. Maybe we should use it more. Or maybe we could wrap around it with something clearer. Let's leave that separately though.

kripken:
Oh thanks, now I see what you mean about lit. Yes, I suppose this could use it - not sure what I was confused about with a temp file there, it seems doable. But let me first move the existing test to lit in another PR, as I'd rather not do both at once here.

kripken:
Oh, actually, wait - the test file here uses multiple modules. Until we support that in lit, I think it makes sense to leave it as it is.

So let's land this as is for now. But I agree moving more to lit where possible is good.

PR #3885, opened 2021-05-13
Support --symbolmap and --symbolmap=FOO in wasm-opt

kripken:
wasm-as supports `--symbolmap=FOO` as an argument. We got a request to
support the same in wasm-opt. wasm-opt does have `--print-function-map` which
does the same, but as a pass. To unify them, use the new pass arg sugar from
#3882 which allows us to add a `--symbolmap` pass whose argument can be
set as `--symbolmap=FOO`. That perfectly matches the wasm-as notation.

For now, keep the old `--print-function-map` notation as well, to not break
emscripten. After we remove it there we can remove it here.


src/passes/PrintFunctionMap.cpp
```diff
@@ -33,9 +34,14 @@ struct PrintFunctionMap : public Pass {
   bool modifiesBinaryenIR() override { return false; }
 
   void run(PassRunner* runner, Module* module) override {
+    // If an argument is provided, write to that file; otherwise write to
+    // stdout.
+    auto outFile = runner->options.getArgumentOrDefault("symbolmap", "");
```
aheejin, line 39: Is this default argument empty string means `cout`? [This](https://github.com/WebAssembly/binaryen/blob/f59174c1cf526d7e9fe5d4ba74ab0370f40a1856/src/support/file.cpp#L91-L92) says `-` is `cout`, so...

aheejin (approved):

src/passes/PrintFunctionMap.cpp
kripken: Interesting, I guess both work then. [Here](https://github.com/WebAssembly/binaryen/blob/f59174c1cf526d7e9fe5d4ba74ab0370f40a1856/src/support/file.h#L53) is a comment documenting the empty file... and the code indeed supports both. We should unify the comments I guess.

aheejin (approved):

PR #3887, opened 2021-05-14
Do not attempt to preserve DWARF if a previous pass removes it

kripken:
If we run a pass that removes DWARF followed by one that could destroy it, then
there is no possible problem - there is nothing left to destroy. We can run the later
pass with no issues (and no warnings).

Fixes https://github.com/emscripten-core/emscripten/issues/14161

### DIFF
--- a/src/passes/pass.cpp
+++ b/src/passes/pass.cpp
@@ -811,4 +811,25 @@ int PassRunner::getPassDebug() {
   return passDebug;
 }
 
+bool PassRunner::passRemovesDebugInfo(const std::string& name) {
+  return pass == "strip" || pass == "strip-debug" || pass == "strip-dwarf";
+}
+
+bool PassRunner::shouldPreserveDWARF() {
+  // Check if the debugging subsystem wants to preserve DWARF.
+  if (!Debug::shouldPreserveDWARF(options, *wasm)) {
+    return false;
+  }
+
+  // We may need DWARF. Check if one of our previous passes would remove it
+  // anyhow, in which case, there is nothing to preserve.
+  for (auto& pass : passes) {
+    if (passRemovesDebugInfo(pass->name)) {
+      return false;
+    }
+  }
+
+  return true;
+}
+
 } // namespace wasm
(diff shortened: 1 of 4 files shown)


src/passes/pass.cpp
```diff
@@ -811,4 +810,25 @@ int PassRunner::getPassDebug() {
...
+
+  // We may need DWARF. Check if one of our previous passes would remove it
+  // anyhow, in which case, there is nothing to preserve.
+  for (auto& pass : passes) {
+    if (passRemovesDebugInfo(pass->name)) {
+      return false;
+    }
+  }
+
+  return true;
```
sbc100, line 831: These lines look like a duplicate of `willRemoveDebugInfo`,  worth merging?

src/passes/pass.cpp
```diff
@@ -811,4 +810,25 @@ int PassRunner::getPassDebug() {
...
+    return false;
+  }
+
+  // We may need DWARF. Check if one of our previous passes would remove it
+  // anyhow, in which case, there is nothing to preserve.
+  for (auto& pass : passes) {
+    if (passRemovesDebugInfo(pass->name)) {
+      return false;
+    }
+  }
```
sbc100, line 829: Is this called a lot?  Is it worth caching the result?

sbc100 (approved):
Great!

src/passes/pass.cpp
kripken: I don't think it's called much, as it's limited by the number of passes. But I suppose it's worth caching, even though it adds some complexity. I added that now, and also some assertions to avoid possible issues.

src/passes/pass.cpp
kripken: I'm not sure that's feasible. Especially after adding the caching, they look even less similar. One is to scan pass names 100% in advance, and the other is to check as we go while adding passes.

src/passes/pass.cpp
sbc100: No worries.  Looks good as is then.

PR #3888, opened 2021-05-16
[NFC] Move InsertOrdered{Set,Map} into a new header

tlively:
Move the `InsertOrderedSet` and `InsertOrderedMap` implementations out of
Relooper.h and into a new insert_ordered.h so they can be used more widely. Only
changes the implementation code to use unordered_maps.

### DIFF
--- a/src/support/insert_ordered.h
+++ b/src/support/insert_ordered.h
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2021 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


src/support/insert_ordered.h
```diff
@@ -0,0 +1,128 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// like std::set, except that begin() -> end() iterates in the
+// order that elements were added to the set (not in the order
+// of operator<(T, T))
+template<typename T> struct InsertOrderedSet {
+  std::map<T, typename std::list<T>::iterator> Map;
```
kripken, line 21: I think we should change the two `std::maps` to unordered (so that we can use them safely with Types). That would keep this NFC (as the order isn't depended on), but that can be later too, as you prefer.

kripken (approved):

PR #3893, opened 2021-05-17
[Wasm GC] Heap2Local: Replace the allocation with null

kripken:
Previously we would try to stop using the allocation as much as possible,
for example not writing it to locals any more, and leaving it to other passes
to actually remove it (and remove gets of those locals etc.). This seemed
simpler and more modular, but does not actually work in some cases as
the fuzzer has found. Specifically, if we stop writing our allocation to
locals, then if we do a `(ref.as_non_null (local.get ..))` of that, then we
will trap on the null present in the local.

Instead, this changes our rewriting to do slightly more work, but it is
simpler in the end. We replace the allocation with a null, and replace
all the places that use it accordingly, for example, updating types to be
nullable, and removing RefAsNonNulls, etc. This literally gets rid of the
allocation and all the places it flows to (leaving less for other passes to
do later).

### DIFF
--- a/src/passes/Heap2Local.cpp
+++ b/src/passes/Heap2Local.cpp
@@ -376,8 +376,8 @@ struct Heap2LocalOptimizer {
       // Replace the allocation with a null reference. This changes the type
       // from non-nullable to nullable, but as we optimize away the code that
       // the allocation reaches, we will handle that.
-      contents.push_back(builder.makeRefNull(
-        Type(allocation->type.getHeapType(), Nullable)));
+      contents.push_back(
+        builder.makeRefNull(Type(allocation->type.getHeapType(), Nullable)));
       replaceCurrent(builder.makeBlock(contents));
     }
 
--- a/src/passes/Heap2Local.cpp
+++ b/src/passes/Heap2Local.cpp
@@ -301,6 +301,16 @@ struct Heap2LocalOptimizer {
       }
     }
 
+    void visitBreak(Break* curr) {
+      if (!reached.count(curr)) {
+        return;
+      }
+
+      // Breaks that our allocation flows through may change type, as we now
+      // have a nullable type there.
+      curr->finalize();
+    }
+
     void visitStructNew(StructNew* curr) {
       if (curr != allocation) {
         return;
--- a/src/passes/Heap2Local.cpp
+++ b/src/passes/Heap2Local.cpp
@@ -356,21 +356,12 @@ struct Heap2LocalOptimizer {
             builder.makeLocalGet(tempIndexes[i], fields[i].type)));
         }
 
-        // Read the values in the allocation (we don't need to, as the
-        // allocation is not used after our optimization, but we need something
-        // with the right type anyhow).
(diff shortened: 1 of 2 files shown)


src/passes/Heap2Local.cpp
```diff
@@ -265,6 +265,31 @@ struct Heap2LocalOptimizer {
...
+    // effects.
+
+    void visitBlock(Block* curr) {
+      if (!reached.count(curr)) {
+        return;
+      }
+
+      // Our allocation passes through this block. We must turn its type into a
+      // nullable one, because we will remove things like RefAsNonNull of it,
+      // which means we may no longer have a non-nullable value as out input,
```
tlively, line 284:
```suggestion
      // which means we may no longer have a non-nullable value as our input,
```

src/passes/Heap2Local.cpp
```diff
@@ -265,6 +265,31 @@ struct Heap2LocalOptimizer {
...
+      // and we could fail to validate. It is safe to make this change in terms
+      // of our parent, since we know very specifically that only safe things
+      // will end up using our value, like a StructGet or a Drop, which do not
+      // care about non-nullability.
```
tlively, line 288: This is going to be slightly tricky to scale to multivalue tuples. Do we have tests that demonstrate what happens when values flow through tuples today?

tlively (approved):
What pass will clean up the unused null values? I wonder if it is worth adding another run line to the lit test with output demonstrating that everything we expect to get cleaned up by the full optimization pipeline is actually cleaned up.

src/passes/Heap2Local.cpp
kripken: Currently flowing into a tuple will stop this optimization. We must explicitly handle everything we flow into, and anything not declared blocks (is considered to maybe escape).

tlively: Yes, I was thinking that it would be useful to have tests for that so it is easy to see the behavior difference if we ever improve the pass to handle tuples in the future.

kripken:
> What pass will clean up the unused null values? I wonder if it is worth adding another run line to the lit test with output demonstrating that everything we expect to get cleaned up by the full optimization pipeline is actually cleaned up.

SimplifyLocals, MergeBlocks, and Vacuum should do it.

But it's not practical to run them on these inputs - they will optimize a lot more out than we want, as we have things like a drop of a struct get. We'd need to write testcases intended for this, if we want that, I think.

Btw, how would I do it? Just adding another run line seems to overwrite the former, and I don't see examples of multiple run lines in our tests.

tlively:
You would have to add a `--check-prefix FOO` argument to the `filecheck` command to tell it to check against `;; FOO`, `;; FOO-NEXT`, etc. The update_lit_checks.py script should be able to handle this.

kripken:
Oh right, thanks @tlively 

Testing with that now, yeah, as I expected the other opts turn almost all the function bodies into a single `nop`, so this isn't useful given how the tests are structured atm. And we have enough coverge of the other opts I think. But I like that idea for future testing.

PR #3894, opened 2021-05-19
[wasm-split] Add a --symbolmap option

tlively:
The new option emits a symbol map file for each of the split modules. The file
names are created by appending ".symbolmap" to each of the Wasm output file
names.

### DIFF
--- a/test/lit/wasm-split/symbolmap.wast
+++ b/test/lit/wasm-split/symbolmap.wast
@@ -1,11 +1,15 @@
 ;; RUN: wasm-split %s --keep-funcs=bar -o1 %t.1.wasm -o2 %t.2.wasm --symbolmap
-;; RUN: filecheck %s --check-prefix PRIMARY < %t.1.wasm.symbolmap
-;; RUN: filecheck %s --check-prefix SECONDARY < %t.2.wasm.symbolmap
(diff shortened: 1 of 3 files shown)


test/lit/wasm-split/symbolmap.wast
```diff
@@ -0,0 +1,24 @@
+;; RUN: wasm-split %s --keep-funcs=bar -o1 %t.1.wasm -o2 %t.2.wasm --symbolmap
```
kripken, line 1: what is `%t` here? a temp filename?

kripken (approved):
How about `.symbols` for the file suffixes? (that's what emscripten does)

OTOH using `symbolmap` for the option is consistent with the other tools, and there is value in keeping the name identical.

Optimally we'd receive a name here and not use a default suffix at all, but as you need two names, I guess this is ok as is.

test/lit/wasm-split/symbolmap.wast
tlively: Exactly. The full list of available substitutions is here: https://llvm.org/docs/CommandGuide/lit.html#substitutions.

tlively:
Yeah, having to prompt for two separate file names felt burdensome. I'll switch the suffixes to ".symbols" for consistency 👍

PR #3901, opened 2021-05-21
[wasm-split] Add an option to emit only the module names

tlively:
Even when other names are stripped, it can be useful for wasm-split to preserve
the module name so that the split modules can be differentiated in stack traces.
Adding this option to wasm-split requires adding similar options to ModuleWriter
and WasmBinaryWriter.

### DIFF
--- a/src/wasm-io.h
+++ b/src/wasm-io.h
@@ -82,7 +82,10 @@ private:
 
 class ModuleWriter : public ModuleIOBase {
   bool binary = true;
+
+  // TODO: Remove `emitModuleName`. See the comment in wasm-binary.h
   bool emitModuleName = false;
+
   std::string symbolMap;
   std::string sourceMapFilename;
   std::string sourceMapUrl;
--- a/src/wasm-io.h
+++ b/src/wasm-io.h
@@ -29,7 +29,7 @@ namespace wasm {
 
 class ModuleIOBase {
 protected:
-  bool debugInfo = true;
+  bool debugInfo;
 
 public:
   // Whether we support debug info (the names section).
--- a/src/wasm-io.h
+++ b/src/wasm-io.h
@@ -29,7 +29,7 @@ namespace wasm {
 
 class ModuleIOBase {
 protected:
-  bool debugInfo;
+  bool debugInfo = true;
 
 public:
   // Whether we support debug info (the names section).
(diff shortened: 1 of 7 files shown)


kripken:
1. Do browsers not use the wasm binary name in stack traces? (If not, what do they do?)
2. When debugging, why not use the entire name section? I'm unclear on whether this new option would be used in production or not. Is this proposing a new mode in between a debug build and a release build?

tlively:
> 1. Do browsers not use the wasm binary name in stack traces? (If not, what do they do?)

AFAIK, they do when the binary is fetched from a URL via e.g. `instantiateStreaming`. However, if the module is compiled from a byte array, it is not automatically given a name, so this can help in that case.

> 2. When debugging, why not use the entire name section? I'm unclear on whether this new option would be used in production or not. Is this proposing a new mode in between a debug build and a release build?

I believe this would be used in production with the other names stripped. In a situation with multiple Wasm modules and multiple source maps, the module name is enough to figure out which source map to use to recreate a stack trace entry.

See discussion on https://github.com/emscripten-core/emscripten/issues/14133.

kripken:
Ok, thanks, I'll comment there.

kripken:
Sorry, I'd written a comment, but apparently not hit submit...

On the Web an Object URL should be a solution for this. Each wasm could be loaded from a different Object URL, which would differentiate them in stack traces. This works already in firefox, but not chrome,

https://bugs.chromium.org/p/v8/issues/detail?id=11808

If we don't want to wait for that, I'm not opposed to landing this with a TODO to remove it eventually.

tlively:
I took a look and it looks like Node does not support the `Blob` API. Does that affect our ability to remove this hack in the future? Is there a better solution that would work for Node as well?

kripken:
Hmm, indeed Node doesn't seem to have ObjectURL support. But I guess it's not really needed there since Node has synchronous filesystem access etc. So we could always ensure a unique URL that way, with a little more work. Maybe not worth it though.

src/wasm-io.h
```diff
@@ -29,7 +29,7 @@ namespace wasm {
 
 class ModuleIOBase {
 protected:
-  bool debugInfo;
+  bool debugInfo = true;
```
kripken, line 32: This default should never have been used, if it wasn't set before. Why add one?

If we need one, how about `false` which is safer? (missing debug info is obvious; extra debug info can subtly bloat)

tlively: It looked dangerous to not have a default, but I can remove this. I don't think it's necessary for anything else in the PR.

kripken (approved):
lgtm, seems reasonable for now.

tlively:
Thanks!

PR #3904, opened 2021-05-24
Add SIMDLoadStoreLane get/setters to C/JS API

dcodeIO:
Adds the respective getters, setters and wrapper class for `SIMDLoadStoreLane`. With this, the C- and JS-APIs should be ready for SIMD :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ test/lit/lit.site.cfg.py
 
 # files commonly related to building out-of-tree
 /build/
+/embuild/
 
 # macOS
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -36,7 +36,6 @@ test/lit/lit.site.cfg.py
 
 # files commonly related to building out-of-tree
 /build/
-/embuild/
 
 # macOS
 .DS_Store
--- a/src/js/binaryen.js-post.js
+++ b/src/js/binaryen.js-post.js
@@ -82,6 +82,7 @@ function initializeConstants() {
     'SIMDTernary',
     'SIMDShift',
     'SIMDLoad',
+    'SIMDLoadStoreLane',
     'MemoryInit',
     'DataDrop',
     'MemoryCopy',
@@ -2998,6 +2999,17 @@ Module['getExpressionInfo'] = function(expr) {
         'align': Module['_BinaryenSIMDLoadGetAlign'](expr),
         'ptr': Module['_BinaryenSIMDLoadGetPtr'](expr)
       };
+    case Module['SIMDLoadStoreLaneId']:
+      return {
+        'id': id,
+        'type': type,
+        'op': Module['_BinaryenSIMDLoadStoreLaneGetOp'](expr),
+        'offset': Module['_BinaryenSIMDLoadStoreLaneGetOffset'](expr),
+        'align': Module['_BinaryenSIMDLoadStoreLaneGetAlign'](expr),
+        'index': Module['_BinaryenSIMDLoadStoreLaneGetIndex'](expr),
+        'ptr': Module['_BinaryenSIMDLoadStoreLaneGetPtr'](expr),
+        'vec': Module['_BinaryenSIMDLoadStoreLaneGetVec'](expr)
+      };
     case Module['MemoryInitId']:
       return {
         'id': id,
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -36,6 +36,7 @@ test/lit/lit.site.cfg.py
 
 # files commonly related to building out-of-tree
 /build/
+/embuild/
```
tlively, line 39: Is this a standard name emitted by some tool?

test/binaryen.js/expressions.js
```diff
@@ -1207,6 +1207,68 @@ console.log("# SIMDLoad");
...
+
+  theSIMDLoadStoreLane.op = op = binaryen.Operations.Load16LaneVec128;
+  assert(theSIMDLoadStoreLane.op === op);
+  theSIMDLoadStoreLane.offset = offset = 32;
+  assert(theSIMDLoadStoreLane.offset === offset);
+  theSIMDLoadStoreLane.align = align = 4;
+  assert(theSIMDLoadStoreLane.align === align);
+  theSIMDLoadStoreLane.index = index = 2;
+  assert(theSIMDLoadStoreLane.index === index);
+  theSIMDLoadStoreLane.ptr = ptr = module.i32.const(2);
```
tlively, line 1240: How about using a different number for the index and the pointer arguments to make sure the test shows that they end up in the right places?

test/binaryen.js/expressions.js
```diff
@@ -1207,6 +1207,68 @@ console.log("# SIMDLoad");
...
+  assert(theSIMDLoadStoreLane.ptr === ptr);
+  assert(theSIMDLoadStoreLane.vec === vec);
+  assert(theSIMDLoadStoreLane.type === binaryen.v128);
+  assert(theSIMDLoadStoreLane.store === false);
+
+  theSIMDLoadStoreLane.op = op = binaryen.Operations.Load16LaneVec128;
+  assert(theSIMDLoadStoreLane.op === op);
+  theSIMDLoadStoreLane.offset = offset = 32;
+  assert(theSIMDLoadStoreLane.offset === offset);
+  theSIMDLoadStoreLane.align = align = 4;
```
tlively, line 1236: I don't think this alignment is valid. IIRC, the spec only allows allows alignments that are the same size or smaller than the access size. In case we start checking this at some point in the future, it would be better to use a valid alignment here.

src/js/binaryen.js-post.js
```diff
@@ -4243,6 +4255,48 @@ Module['SIMDLoad'] = makeExpressionWrapper({
...
+  },
+  'getVec'(expr) {
+    return Module['_BinaryenSIMDLoadStoreLaneGetVec'](expr);
+  },
+  'setVec'(expr, ptrExpr) {
+    Module['_BinaryenSIMDLoadStoreLaneSetVec'](expr, ptrExpr);
+  },
+  'isStore'(expr) {
+    return Boolean(Module['_BinaryenSIMDLoadStoreLaneIsStore'](expr));
+  }
```
tlively, line 4297: it doesn't look like this one is exercised by the JS tests.

.gitignore
dcodeIO: That's something I made up, as I typically have to compile both wasm-opt and binaryen.js to run the tests respectively update the tests, so I am using two folders for the different builds, and `embuild` seemed appropriate for the Emscripten one :)

src/js/binaryen.js-post.js
dcodeIO: It is tested by accessing the `.store` property, which is generated from the `.isStore` function.

test/binaryen.js/expressions.js
dcodeIO: It kinda does that, by initializing with `index = 1` and `ptr = i32.const(1)`, testing that, and then updating to `index = 2` and `ptr = i32.const(2)` before testing again here. I may be missing something though :)

tlively: Ah yes, fair enough. I was only thinking about the values that get printed.

.gitignore
dcodeIO: Removed this one again, as I could ofc just use two subfolders if I weren't so lazy heh

tlively: Thanks. I don't see any reason not to add this to the .gitignore since we already have `/build/` in there, which I think is in a similar situation.

tlively (approved):

PR #3905, opened 2021-05-26
[wasm-split] Minimize names of newly created exports 

tlively:
wasm-split would previously use internal function names to create the external
names of the functions that are newly exported from the primary module to be
imported into the secondary module. When the input module contains full function
names (as is commonly the case when emitting symbol maps), this caused the
function names to be preserved as the export names, even when names are
otherwise being stripped. To save on code size and properly anonymize functions,
generate minimal export names when debuginfo is disabled instead.


tlively:
@kripken, it looks like the alpine builder just timed out, so this should be ready for review.

src/ir/module-splitting.cpp
```diff
@@ -336,6 +340,34 @@ ModuleSplitter::initExportedPrimaryFuncs(const Module& primary) {
   return functionExportNames;
 }
 
+std::string ModuleSplitter::generateMinimizedExportName() {
+  // Minimized export names don't need to be human readable at all, so use as
+  // much of the space of single-byte UTF-8 unicode code points as possible.
+  // Binaryen does not actually handle unicode well, though, so only use code
+  // points that correspond to easily printable glyphs: 0x20-0x7E.
+  static std::array<uint8_t, 95> chars = ([]() {
```
kripken, line 348: I think it would be good to emit JS-friendly identifiers. That would leave us the flexibility to integrate the modules using JS as needed. This would have a cost (avoid special characters and reserved words in JS), but I think it would be small. And then we could reuse the code from `src/passes/MinifyImportsAndExports.cpp` (`MinifiedNames`) which would be simpler, too?

tlively: I agree that it would be good to reuse existing name minification, but the code from `MinifyImportsAndExports.cpp` would require serious refactoring to be able to be used here, where not all imports/exports should be minimized and the minimized names should be generated on the fly rather than all at the same time. I would be happy to do that refactoring, but I'm not sure the benefit of emitting JS-friendly identifiers is worth it. These minimized exports are only exported at all to make them available across the module split; they should not be used, inspected, or referenced in any way by JS. I am comfortable requiring any low-level use case that requires hacking around that abstraction to deal with arbitrary non-JS-friendly names.

kripken: I think we can move just `MinifiedNames` out of there, and use it here? (I mean literally just that one class) I agree most of the logic can't be shared, but just the minified names generator code (not what decides which imports and exports to be minified).

tlively: Ah, I hadn't read carefully enough to see that `MinifiedNames` was easily separable. Refactoring done in #3909. Will rebase this PR on top of that once that lands.

test/lit/wasm-split/minimized-exports.wast
```diff
@@ -0,0 +1,38 @@
+;; RUN: wasm-split %s --keep-funcs=foo,bar --export-prefix='%' -o1 %t.1.wasm -o2 %t.2.wasm
+;; RUN: wasm-dis %t.1.wasm | filecheck %s --check-prefix PRIMARY
+;; RUN: wasm-dis %t.2.wasm | filecheck %s --check-prefix SECONDARY
+
+;; PRIMARY:      (module
+;; PRIMARY-NEXT:   (type $none_=>_none (func))
+;; PRIMARY-NEXT:   (export "%a" (func $1))
+;; PRIMARY-NEXT:   (export "%b" (func $0))
```
kripken, line 8: I'm missing something... now that this uses the shared minified names generator, that should be emitting valid JS names, but `%a` isn't such?

tlively: This is using the JS-friendly minified names with a `%` prefix manually specified on the command line.

tlively: In particular, the `%` prefix is what the Emscripten integration currently uses to differentiate wasm-split exports from "real" exports, so this models how wasm-split would actually be invoked in practice.

kripken: Oh, I see, I wasn't aware of how it worked on the emscripten side. So it's intentionally not a JS-friendly prefix. (That removes some of the benefit of sharing the JS-friendly minification code here, I suppose, but it's still simpler I think.)

kripken (approved):

PR #3909, opened 2021-05-27
[NFC] Factor out and simplify minified name generation

tlively:
Simplifies the public API to not unnecessarily take an index and simplifies the
implementation to use a single integer as state rather than a vector of indices.

### DIFF
--- a/src/ir/names.h
+++ b/src/ir/names.h
@@ -88,15 +88,6 @@ inline Name getValidElementSegmentName(Module& module, Name root) {
 }
 
 class MinifiedNameGenerator {
-  // Reserved words we must not emit.
-  static std::unordered_set<std::string> reserved;
-
-  // Possible initial letters.
-  static std::string validInitialChars;
-
-  // Possible later letters.
-  static std::string validLaterChars;
-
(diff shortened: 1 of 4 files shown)


tlively:
The lack of changes to test/passes/minify-imports-and-exports_all-features.wast shows that this is a non-functional change.

src/ir/names.h
```diff
@@ -88,6 +87,23 @@ inline Name getValidElementSegmentName(Module& module, Name root) {
...
 
+class MinifiedNameGenerator {
+  // Reserved words we must not emit.
+  static std::unordered_set<std::string> reserved;
+
+  // Possible initial letters.
+  static std::string validInitialChars;
+
+  // Possible later letters.
+  static std::string validLaterChars;
```
kripken, line 98: Do the statics need to be in the header? They could be only in the cpp I think, which seems better to me.

kripken (approved):
lgtm % comment

tlively:
Not sure why macos is listed as "skipped" and alpine is taking so long as usual, but I'll land this manually because it otherwise looks good.

PR #3911, opened 2021-05-27
[Wasm GC] Add experimental array.copy

kripken:
Spec for it is here:

https://docs.google.com/document/d/1DklC3qVuOdLHSXB5UXghM_syCh-4cMinQ50ICiXnK3Q/edit#

Also reorder some things in wasm.h that were not in the canonical order (that has
no effect, but it is confusing to read).


tlively (approved):
LGTM!

kripken:
The clang-tidy error here makes no sense to me... the change here is trivial. It seems tidy dislikes the existing code before this PR, but the errors look like weird false positives. Like it is actually compiling the file instead of analyzing it.

I'll ignore the tidy errors to land, but maybe someone that understands tidy better wants to take a look at that?

PR #3913, opened 2021-05-27
[Wasm GC] Add negated BrOn* operations

kripken:
They are basically the flip versions. The only interesting part in the impl is that their
returned typed and sent types are different.

Spec: https://docs.google.com/document/d/1DklC3qVuOdLHSXB5UXghM_syCh-4cMinQ50ICiXnK3Q/edit


kripken:
Also added BrOnNonNull, which was missing in the doc, but has just been added.

tlively (approved):
LGTM besides previous comments.

kripken:
Thanks for the feedback! Updated.

PR #3916, opened 2021-05-30
[wasm-split] Make option validation declarative

tlively:
In anticipation of adding a third wasm-split mode, merge-profiles, in addition
to the existing split and instrument modes, refactor wasm-split's option
validation to let the valid modes be declared for each option. This approach is
more scalable and robust than the ad-hoc validation we had previously.

### DIFF
--- a/test/lit/wasm-split/help.test
+++ b/test/lit/wasm-split/help.test
@@ -4,7 +4,8 @@ CHECK:      wasm-split INFILE
 CHECK-NEXT:
 CHECK-NEXT: Split a module into a primary module and a secondary module, or instrument a
 CHECK-NEXT: module to gather a profile that can inform future splitting, or manage such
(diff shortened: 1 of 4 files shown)


test/lit/wasm-split/help.test
```diff
@@ -0,0 +1,67 @@
+;; RUN: wasm-split --help | filecheck %s
+
+CHECK:      wasm-split INFILE
+CHECK-NEXT:
+CHECK-NEXT: Split a module into a primary module and a secondary module, or instrument a
+CHECK-NEXT: module to gather a profile that can inform future splitting, or manage such
+CHECK-NEXT: profiles.
```
kripken, line 7: the text here could perhaps mention what the `[split]` etc stuff means?

kripken (approved):
lgtm % comments

Issue #3912, opened 2021-05-27
[wasm-split] using multiple profiles at once


tlively:
Right now wasm-split's instrumentation produces one profile per run and the splitting logic consumes one profile. That means that it only natively supports splitting based on a single workload. In practice, it is more useful to be able to run multiple workloads and use them all to inform splitting.

I was thinking that the simplest solution would be to allow profiles to be concatenated to each other to form a larger file and to have wasm-split accept and use such concatenated profiles.

An alternative solution that would keep profile files smaller would be to add an option to wasm-split to combine two (or multiple) profiles into a single profile. This would be more complicated because it would require wasm-split to do more things, but it wouldn't need separate user-provided concatenation logic.

Any thoughts about what would be the best design and semantics here?

cc @sbc100, @awtcode, @sps-gold, @Mintyboi, @waterlike86

sps-gold:
I think rather than concatenating profiles into large file, it will be better to merge all profile data into one table. This will keep output profile file size constant, and we not loose any info because all we need during split is just to know if the function should belong to main or deferred module.

I also prefer alternate solution as if we extend wasm-split to use for merging profiles, say like
`wasm-split --merge_profiles output_profile <list of input profiles>` and/or
`wasm-split --merge_profiles main_profile profile_to_merge`
then we can reuse existing read/write profile logic and no maintenance of separate code base be required. 

PS: It might also be good to log some information around merge. Like if profile_to_merge is a subset of main_profile in example above then we can drop the run for profile_to_merge as it does not contribute any extras.

tlively:
This was implemented in #3917.

PR #3917, opened 2021-05-31
 [wasm-split] Add a merge-profiles mode

tlively:
Given a list of profiles for the same module, --merge-profiles produces a single
combined profile the contains the minimum timestamp among the original profiles
for each function. When verbose output is enabled, also emit a message for each
profile that could individually be removed without affecting the set of
functions in the combined profile, as suggested in #3912.

Marking this draft until #3916, which this includes, is merged. At that point I will
rebase this PR and mark it non-draft. Until then, just look at the diff for commits
after the first one.

### DIFF
--- a/src/tools/wasm-split.cpp
+++ b/src/tools/wasm-split.cpp
@@ -824,7 +824,7 @@ void mergeProfiles(const WasmSplitOptions& options) {
   // Read the initial profile. We will merge other profiles into this one.
   ProfileData data = readProfile(options.inputFiles[0]);
 
-  // In verbose mode, we want to find profiles that don't contributed to the
+  // In verbose mode, we want to find profiles that don't contribute to the
   // merged profile. To do that, keep track of how many profiles each function
   // appears in. If any profile contains only functions that appear in multiple
   // profiles, it could be dropped.
--- a/src/tools/wasm-split.cpp
+++ b/src/tools/wasm-split.cpp
@@ -26,6 +26,7 @@
 #include "support/path.h"
 #include "support/utilities.h"
 #include "tool-options.h"
+#include "wasm-binary.h"
 #include "wasm-builder.h"
 #include "wasm-io.h"
 #include "wasm-type.h"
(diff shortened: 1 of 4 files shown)


src/tools/wasm-split.cpp
```diff
@@ -777,6 +820,77 @@ void splitModule(Module& wasm, const WasmSplitOptions& options) {
   writeModule(*secondary, options.secondaryOutput, options);
 }
 
+void mergeProfiles(const WasmSplitOptions& options) {
+  // Read the initial profile. We will merge other profiles into this one.
+  ProfileData data = readProfile(options.inputFiles[0]);
+
+  // In verbose mode, we want to find profiles that don't contributed to the
```
kripken, line 827:
```suggestion
  // In verbose mode, we want to find profiles that don't contribute to the
```

src/tools/wasm-split.cpp
```diff
@@ -777,6 +820,77 @@ void splitModule(Module& wasm, const WasmSplitOptions& options) {
...
+  // merged profile. To do that, keep track of how many profiles each function
+  // appears in. If any profile contains only functions that appear in multiple
+  // profiles, it could be dropped.
```
kripken, line 830: But why do we want to drop it and/or report it is redundant? That's not really an error IIUC... maybe just one profile happened to be a "sub-profile" of another.

tlively: This is a feature request from https://github.com/WebAssembly/binaryen/issues/3912#issuecomment-850189858. This information is not an error, but it is useful for paring down the number of workloads used to create the combined profile. (Of course it would be more useful if we implemented a solution to the [set cover problem](https://en.wikipedia.org/wiki/Set_cover_problem), but this seems good enough for now.)

kripken: I am a little worried of feature creep here. But if it's requested then I'm not strongly opposed.

kripken (approved):

src/tools/wasm-split.cpp
tlively: I agree. It may make more sense to refactor this into a separate mode or flag in the future rather than using `verbose` as a catch-all flag for dumping information. I also think it will make sense to split wasm-split up into multiple files soon, which should help with the maintenance burden of all these random features.

PR #3919, opened 2021-06-03
Initial nominal typing support

tlively:
In nominal mode, HeapType constructors besides the Signature constructor always
produce fresh types distinct from any previously created types. The HeapType
constructor that takes a Signature maintains its previous behavior of
constructing a canonical representative of the given signature because it is
used frequently throughout the code base and never in a situation that would
benefit from creating a fresh type. It is left as future work to clean up this
discrepancy between the Signature HeapType constructor and other HeapType
constructors.

TypeBuilder skips shape canonicalization in nominal mode and always creates a
fresh type for each of its entries. It is left as future work to simplify the
global canonicalization step in nominal mode by disallowing the construction of
basic HeapTypes, which are the only HeapTypeBuilder entries that still require
canonicalization.

The new mode is not available in any of the Binaryen tools yet because it is
still missing critical functionality like the ability to declare subtyping
relations and correctly calculate LUBs. This functionality will be implemented
in future PRs.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -42,6 +42,9 @@
 
 namespace wasm {
 
+static TypeSystem typeSystem = TypeSystem::Equirecursive;
+void setTypeSystem(TypeSystem system) { typeSystem = system; }
+
 namespace {
 
 struct TypeInfo {
@@ -403,9 +406,7 @@ public:
 
 template<> class hash<wasm::HeapTypeInfo> {
 public:
-  size_t operator()(const wasm::HeapTypeInfo& info) const {
-    return wasm::FiniteShapeHasher().hash(info);
-  }
+  size_t operator()(const wasm::HeapTypeInfo& info) const;
 };
 
 template<typename T> class hash<reference_wrapper<const T>> {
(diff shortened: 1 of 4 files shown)


tlively:
Oops, I jumped the gun a bit here. There are still a couple issues to iron out.

src/wasm/wasm-type.cpp
```diff
@@ -689,40 +708,56 @@ template<typename Info> bool Store<Info>::isGlobalStore() {
...
   typename Info::type_t canonical;
   if (info.getCanonical(canonical)) {
     return canonical;
   }
   std::lock_guard<std::recursive_mutex> lock(mutex);
-  auto indexIt = typeIDs.find(std::cref(info));
-  if (indexIt != typeIDs.end()) {
-    return typename Info::type_t(indexIt->second);
+  if (std::is_same<Info, TypeInfo>::value ||
+      typeSystem == TypeSystem::Equirecursive) {
```
kripken, line 718: why is TypeInfo special here?

kripken (commented):
Sorry for the delay in review. Github does not send notifications when a PR changes from "draft" to "ready", it seems...

src/wasm/wasm-type.cpp
tlively: Using nominal types should only change the behavior of the HeapType store, not the Type store. It should not be possible, for example, to have two distinct nullable reference types to the same HeapType, so the Type store needs to keep canonicalizing ust as it does in the equirecursive system.

kripken: I see, thanks! That's what I was missing.

kripken (approved):
lgtm with maybe a comment there?

tlively:
Thanks, added comments. I'll make sure to add a GitHub comment when marking PRs ready for review in the future, too. It's too bad that GitHub doesn't send an email for that.

PR #3923, opened 2021-06-07
[EH] Convert EH tests into lit tests

aheejin:
This converts most EH tests in test/passes into test/lit/passes. Fixed
some files to follow 2-space indentation and improved some comments.


tlively:
Awesome! Besides the edits you mentioned, did you just use the update_lit_checks.py script to do this conversion?

aheejin:
Except for 2-3 tests yes, which required 1. stack IR printing or 2. module elements printing. They don't have the 
```
```
at the top.

I thought about not autogenerating the whole function in case it is slightly longer but only adding lines that are relevant, but it was not very easy because in case we want to check this pattern in a function:
```wasm
(drop
  (i32.add
    (local.get $x)
    (i32.const 3)
  )
)
```
 
Simply doing this doesn't work in a lot of times because the first `(drop` can be matched to another line that comes before the part we want to match, and all those `CHECK-NEXT` will fail. So I found, unlike LLVM in which we don't use autogeneration very often, using the autogeneration to generate `CHECK` lines was easier.
```
;; CHECK:     (drop
;; CHECK-NEXT:  (i32.add
;; CHECK-NEXT:    (local.get $x)
;; CHECK-NEXT:    (i32.const 3)
;; CHECK-NEXT:  )
;; CHECK-NEST: )
```

test/lit/passes/dce-eh.wast
```diff
@@ -0,0 +1,126 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
+;; RUN: wasm-opt %s --dce -all -S -o - | filecheck %s
+
+;; Exception handling instruction support
```
tlively, line 4:
```suggestion
```

This section header comment probably isn't necessary since the whole file is testing EH support now.

test/lit/passes/inlining-eh.wast
```diff
@@ -0,0 +1,99 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
+;; RUN: wasm-opt %s --inlining -all -S -o - | filecheck %s
+
+(module
+  ;; ---------------------------------------------------------------------------
+  (import "a" "b" (func $foo (result i32)))
+  (event $event$0 (attr 0) (param i32))
+  (func $callee-with-label
```
tlively, line 8: Why doesn't this function have check lines? If they were manually deleted, they will come back next time someone runs `./auto_update_tests.py`, so it would be better just to keep them in.

tlively (approved):

test/lit/passes/inlining-eh.wast
aheejin: This is strange. update_lit_checks.py actually doesn't update this. Also `(func $callee` in the same file. Not sure why.

tlively: Must be a bug in the regex used by the script. I wouldn't worry about it if those functions aren't critical to the test.

kripken (approved):

test/lit/passes/inlining-eh.wast
aheejin: This indeed can't exist because `$callee-with-label` is inlined! 🙃

tlively: ohhh, that makes sense. Sorry for the confusion!

PR #3927, opened 2021-06-08
Nominal subtyping

tlively:
Add methods to the TypeBuilder interface to declare subtyping relationships
between the built types. These relationships are validated and recorded globally
as part of type building. If the relationships are not valid, a fatal error is
produced. In the future, it would be better to report the error to the
TypeBuilder client code, but this behavior is sufficient for now. Also updates
SubTyper and TypeBounder to be aware of nominal mode so that subtyping and LUBs
are correctly calculated.

Tests of the failing behavior will be added in a future PR that exposes this
functionality to the command line, since the current `example` testing
infrastructure cannot handle testing fatal errors.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -1470,7 +1470,10 @@ HeapType TypeBounder::lub(HeapType a, HeapType b) {
   }
 
   if (typeSystem == TypeSystem::Nominal) {
-    // Walk up the subtype tree to find the LUB.
+    // Walk up the subtype tree to find the LUB. Ascend the tree from both `a`
+    // and `b` in lockstep. The first type we see for a second time must be the
+    // LUB because there are no cycles and the only way to encounter a type
+    // twice is for it to be on the path above both `a` and `b`.
     std::unordered_set<HeapTypeInfo*> seen;
     auto* currA = infoA;
     auto* currB = infoB;
@@ -2954,10 +2957,14 @@ std::vector<HeapType> buildEquirecursive(TypeBuilder& builder) {
   return heapTypes;
 }
 
-void validateSubTyping(const std::vector<HeapType>& heapTypes) {
-  // Ensure there are no cycles in the subtype graph.
+void validateNominalSubTyping(const std::vector<HeapType>& heapTypes) {
+  assert(typeSystem == TypeSystem::Nominal);
+
+  // Ensure there are no cycles in the subtype graph. This is the classic DFA
+  // algorithm for detecting cycles, but in the form of a simple loop because
+  // each node (type) has at most one child (supertype).
   std::unordered_set<HeapTypeInfo*> seen;
-  for (HeapType type : heapTypes) {
+  for (auto type : heapTypes) {
     std::unordered_set<HeapTypeInfo*> path;
     for (auto* curr = getHeapTypeInfo(type);
          seen.insert(curr).second && curr->supertype != nullptr;
@@ -3026,7 +3033,7 @@ std::vector<HeapType> buildNominal(TypeBuilder& builder) {
   auto afterMove = std::chrono::steady_clock::now();
 #endif
 
-  validateSubTyping(heapTypes);
+  validateNominalSubTyping(heapTypes);
 
 #if TIME_CANONICALIZATION
   auto end = std::chrono::steady_clock::now();
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -101,6 +101,8 @@ struct HeapTypeInfo {
   // Otherwise, the type definition tree is still being constructed via the
   // TypeBuilder interface, so hashing and equality use pointer identity.
   bool isFinalized = true;
+  // In nominal mode, the supertype of this HeapType, if it exists.
+  HeapTypeInfo* supertype = nullptr;
   enum Kind {
     BasicKind,
     SignatureKind,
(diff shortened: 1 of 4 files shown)


src/wasm/wasm-type.cpp
```diff
@@ -766,6 +769,11 @@ TypeID Store<Info>::doInsert(std::unique_ptr<Info>&& info) {
   return id;
 }
 
+// For nominal mode, explicitly track all the subtype relations. Each HeapType
+// can have only one immediate supertype. TODO:
```
kripken, line 773: what is the TODO?

src/wasm/wasm-type.cpp
```diff
@@ -766,6 +769,11 @@ TypeID Store<Info>::doInsert(std::unique_ptr<Info>&& info) {
...
+std::unordered_map<HeapType, HeapType> nominalSuperTypes;
+std::mutex nominalSuperTypesLock;
```
kripken, line 775: why is a separate lock needed for this? the class has a lock already, so i assume there is a specific design in mind here? comment would be good.

src/wasm/wasm-type.cpp
```diff
@@ -1283,9 +1291,24 @@ bool SubTyper::isSubType(HeapType a, HeapType b) {
   if (b == HeapType::func) {
     return a.isSignature();
   }
+  if (typeSystem == TypeSystem::Nominal) {
+    // Subtyping must be declared in a nominal system, not derived from
+    // structure, so we will not recurse.
+    std::lock_guard<std::mutex> lock(nominalSuperTypesLock);
```
kripken, line 1297: Taking a lock on each call to isSubType sounds worrying. Can the subtype relations be stored on the heap types perhaps?

src/wasm/wasm-type.cpp
tlively: Oops, I think this is stale.

src/wasm/wasm-type.cpp
tlively: This lock isn't part of a `Store`. It's a standalone lock that protects the single global `nominalSuperTypes`.

src/wasm/wasm-type.cpp
tlively: Yeah there are a few options here. The sets of supertypes or subtypes could perhaps be stored directly on the HeapTypeInfos or the relation could be cached in thread-local maps. I want to start with the easiest possible thing, though, which is why I just have a single global source of truth with a single course-grained lock and no caching anywhere.

kripken: I see. But storing the relationships on the HeapTypes isn't "caching" - that would be the single source of truth? To me it feels most natural actually, but maybe I don't understand the entire design here yet.

Do you want to defer this topic for later? I don't feel strongly about when we resolve it.

tlively: Ohh, I see what you mean! Yes, that is an excellent idea. I think it's worth updating the implementation now.

src/wasm/wasm-type.cpp
```diff
@@ -1416,44 +1436,68 @@ HeapType TypeBounder::lub(HeapType a, HeapType b) {
...
+      auto* nextA = currA->supertype;
+      auto* nextB = currB->supertype;
+      if (nextA == nullptr && nextB == nullptr) {
+        // Did not find a LUB in the subtype tree.
+        return getBasicLUB();
+      }
+      if (nextA) {
+        if (!seen.insert(nextA).second) {
+          return HeapType(uintptr_t(nextA));
         }
```
kripken, line 1489: Do I understand correctly that the algorithm here depends on the lack of recursion in the nominal type tree, and therefore a single `seen` is ok (and not one for each of them)? That is, if we find next**A** is already seen, then it must have been seen while traversing **B**, and vice versa?

Sounds correct, but maybe worth a comment?

src/wasm/wasm-type.cpp
```diff
@@ -2872,6 +2954,60 @@ std::vector<HeapType> buildEquirecursive(TypeBuilder& builder) {
   return heapTypes;
 }
 
+void validateSubTyping(const std::vector<HeapType>& heapTypes) {
+  // Ensure there are no cycles in the subtype graph.
+  std::unordered_set<HeapTypeInfo*> seen;
+  for (HeapType type : heapTypes) {
```
kripken, line 2960:
```suggestion
  for (auto type : heapTypes) {
```
(also below)

src/wasm/wasm-type.cpp
```diff
@@ -2872,6 +2954,60 @@ std::vector<HeapType> buildEquirecursive(TypeBuilder& builder) {
   return heapTypes;
 }
 
+void validateSubTyping(const std::vector<HeapType>& heapTypes) {
+  // Ensure there are no cycles in the subtype graph.
```
kripken, line 2958: maybe add an assert that the type system is nominal? (or, the name could include "nominal" in it perhaps)

kripken (commented):
Nice! lgtm aside from final comments/questions.

src/wasm/wasm-type.cpp
tlively: Yes, that understanding sounds correct. The only way to see a type twice is if it is on the paths above both `a` and `b` since there are no cycles. Adding a comment.

tlively:
Note that I turned auto-merge on, so disable it first if you want to LGTM with additional comments.

kripken (approved):
Comment looks good!

PR #3929, opened 2021-06-09
Store signatures as HeapTypes when parsing binaries

tlively:
When parsing func.ref instructions, we need to get the HeapType corresponding to
the referenced function's signature. Since constructing HeapTypes from
Signatures can be expensive under equirecursive typing, keep track of the
original function signature HeapTypes directly during parsing rather than
storing them as Signatures.

@kripken, what do you think about this as an alternative to #3925? IMO it would be good to avoid adding more state to binary parsing if possible. The possible downside of this approach is that it will not deduplicate signature HeapTypes in a nominal system. If that's something we want to do, we will have to either go with the caching approach or possibly perform the deduplication in `ModuleUtils::collectHeapTypes`. Without knowing more about the semantics of a nominally typed GC proposal, though, I don't know if we will even want to deduplicate signature types, so I'm inclined to not worry about that right now.

### DIFF
--- a/src/wasm-binary.h
+++ b/src/wasm-binary.h
@@ -1401,11 +1401,11 @@ public:
   // The signatures of each function, including imported functions, given in the
   // import and function sections. Store HeapTypes instead of Signatures because
   // reconstructing the HeapTypes from the Signatures is expensive.
-  std::vector<HeapType> functionSignatures;
+  std::vector<HeapType> functionTypes;
 
   void readFunctionSignatures();
-  HeapType getSignatureByFunctionIndex(Index index);
-  HeapType getSignatureByTypeIndex(Index index);
+  HeapType getTypeByFunctionIndex(Index index);
+  HeapType getSigTypeByIndex(Index index);
 
   size_t nextLabel;
 
--- a/src/wasm-binary.h
+++ b/src/wasm-binary.h
@@ -1404,8 +1404,10 @@ public:
   std::vector<HeapType> functionTypes;
 
   void readFunctionSignatures();
+  HeapType getTypeByIndex(Index index);
   HeapType getTypeByFunctionIndex(Index index);
-  HeapType getSigTypeByIndex(Index index);
+  Signature getSignatureByTypeIndex(Index index);
+  Signature getSignatureByFunctionIndex(Index index);
 
   size_t nextLabel;
 
--- a/src/wasm-binary.h
+++ b/src/wasm-binary.h
@@ -1398,12 +1398,14 @@ public:
                           Address defaultIfNoMax);
   void readImports();
 
-  // The signatures of each function, given in the function section
-  std::vector<Signature> functionSignatures;
+  // The signatures of each function, including imported functions, given in the
+  // import and function sections. Store HeapTypes instead of Signatures because
+  // reconstructing the HeapTypes from the Signatures is expensive.
+  std::vector<HeapType> functionSignatures;
 
   void readFunctionSignatures();
-  Signature getSignatureByFunctionIndex(Index index);
-  Signature getSignatureByTypeIndex(Index index);
+  HeapType getSignatureByFunctionIndex(Index index);
+  HeapType getSignatureByTypeIndex(Index index);
 
   size_t nextLabel;
 
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -1657,7 +1657,7 @@ Type WasmBinaryBuilder::getType(int initial) {
   // Single value types are negative; signature indices are non-negative
   if (initial >= 0) {
     // TODO: Handle block input types properly.
-    return getSignatureByTypeIndex(initial).results;
+    return getSignatureByTypeIndex(initial).getSignature().results;
   }
   Type type;
   if (getBasicType(initial, type)) {
@@ -1986,9 +1986,9 @@ void WasmBinaryBuilder::readImports() {
     switch (kind) {
       case ExternalKind::Function: {
         Name name(std::string("fimport$") + std::to_string(functionCounter++));
-        auto index = getU32LEB();
-        auto curr =
-          builder.makeFunction(name, getSignatureByTypeIndex(index), {});
+        auto type = getSignatureByTypeIndex(getU32LEB());
+        functionSignatures.push_back(type);
+        auto curr = builder.makeFunction(name, type.getSignature(), {});
         curr->module = module;
         curr->base = base;
         functionImports.push_back(curr.get());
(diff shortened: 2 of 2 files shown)


kripken:
First question, how does this benchmark against current trunk, and #3925 ? It's not obvious to me which of the two PRs would be faster. I lean towards the faster, as we will likely rework everything later anyhow when the nominal/structural stuff is decided, and for now saving significant amounts of time is useful.

tlively:
I'm getting these measurements:

main: 77.344s
#3925: 14.71s
this PR: 11.772s

src/wasm-binary.h
```diff
@@ -1398,12 +1398,14 @@ class WasmBinaryBuilder {
...
+  // The signatures of each function, including imported functions, given in the
+  // import and function sections. Store HeapTypes instead of Signatures because
+  // reconstructing the HeapTypes from the Signatures is expensive.
+  std::vector<HeapType> functionSignatures;
 
   void readFunctionSignatures();
-  Signature getSignatureByFunctionIndex(Index index);
-  Signature getSignatureByTypeIndex(Index index);
+  HeapType getSignatureByFunctionIndex(Index index);
+  HeapType getSignatureByTypeIndex(Index index);
```
kripken, line 1408: how about renaming these to `getType*`?

src/wasm-binary.h
```diff
@@ -1398,12 +1398,14 @@ class WasmBinaryBuilder {
                           Address defaultIfNoMax);
   void readImports();
 
-  // The signatures of each function, given in the function section
-  std::vector<Signature> functionSignatures;
+  // The signatures of each function, including imported functions, given in the
+  // import and function sections. Store HeapTypes instead of Signatures because
+  // reconstructing the HeapTypes from the Signatures is expensive.
+  std::vector<HeapType> functionSignatures;
```
kripken, line 1404: how about renaming this to `functionTypes`?

src/wasm/wasm-binary.cpp
```diff
@@ -1657,7 +1657,7 @@ Type WasmBinaryBuilder::getType(int initial) {
   // Single value types are negative; signature indices are non-negative
   if (initial >= 0) {
     // TODO: Handle block input types properly.
-    return getSignatureByTypeIndex(initial).results;
+    return getSignatureByTypeIndex(initial).getSignature().results;
```
kripken, line 1660: after the renaming this would no longer read as `getSignature...getSignature()`

src/wasm/wasm-binary.cpp
```diff
@@ -2109,13 +2104,13 @@ Signature WasmBinaryBuilder::getSignatureByTypeIndex(Index index) {
...
   }
-  return heapType.getSignature();
+  return heapType;
 }
 
 void WasmBinaryBuilder::readFunctions() {
   BYN_TRACE("== readFunctions\n");
   size_t total = getU32LEB();
-  if (total != functionSignatures.size()) {
+  if (total != functionSignatures.size() - functionImports.size()) {
```
kripken, line 2113: this change seems very surprising, as I thought the PR is an NFC change that just stores heap types instead of signatures, but keeps indexing all the same..?

src/wasm/wasm-binary.cpp
```diff
@@ -2129,7 +2124,7 @@ void WasmBinaryBuilder::readFunctions() {
 
     auto* func = new Function;
     func->name = Name::fromInt(i);
-    func->sig = functionSignatures[i];
+    func->sig = functionSignatures[functionImports.size() + i].getSignature();
```
kripken, line 2127: ditto

kripken (commented):
Numbers look great!

src/wasm/wasm-binary.cpp
tlively: The imported `Function`s don't contain their HeapTypes, so we need to change `functionSignatures` (soon to be renamed `functionTypes`) to include the HeapTypes of the imported functions as well to avoid having to resynthesize them later.

src/wasm-binary.h
tlively: Also note the expanded comment calling out that this now includes imported functions as well.

src/wasm-binary.h
```diff
@@ -1398,12 +1398,14 @@ class WasmBinaryBuilder {
...
+  // The signatures of each function, including imported functions, given in the
+  // import and function sections. Store HeapTypes instead of Signatures because
+  // reconstructing the HeapTypes from the Signatures is expensive.
+  std::vector<HeapType> functionTypes;
 
   void readFunctionSignatures();
-  Signature getSignatureByFunctionIndex(Index index);
-  Signature getSignatureByTypeIndex(Index index);
+  HeapType getTypeByFunctionIndex(Index index);
+  HeapType getSigTypeByIndex(Index index);
```
tlively, line 1408: I made this `getSigTypeByIndex` instead of just `getTypeByIndex` because it still checks whether the type is a signature type and fails if it is not.

kripken: How about keeping `getSignature*` then, as convenience methods, that do the `.getSignature()` for us? Shorter to write, and clearer, I think.

tlively: Sure, that makes sense.

src/wasm/wasm-binary.cpp
```diff
@@ -2129,7 +2139,7 @@ void WasmBinaryBuilder::readFunctions() {
 
     auto* func = new Function;
     func->name = Name::fromInt(i);
-    func->sig = functionSignatures[i];
+    func->sig = functionTypes[functionImports.size() + i].getSignature();
```
kripken, line 2142:
```suggestion
    func->sig = getSignatureByFunctionIndex(i);
```

src/wasm/wasm-binary.cpp
```diff
@@ -6043,14 +6053,11 @@ void WasmBinaryBuilder::visitRefIs(RefIs* curr, uint8_t code) {
 void WasmBinaryBuilder::visitRefFunc(RefFunc* curr) {
   BYN_TRACE("zz node: RefFunc\n");
   Index index = getU32LEB();
-  if (index >= functionImports.size() + functionSignatures.size()) {
-    throwError("ref.func: invalid call index");
-  }
```
kripken, line 6048: We still need this check, I think. The binary can contain an invalid index.

tlively: The call to `getTypeByFunctionIndex` below will check that the index is valid.

kripken: Ah, I see. That is slightly surprising, in particular, since `functionRefs[index]` seems like it could access memory out of bounds - if it was an array initialized to the number of functions. How about adding a comment like
```
// Note that we do not need to check that 'index' is in bounds, as that will be
// verified lower down. (Also, note that functionRefs[index] may write to an odd
// place in the functionRefs map if index is invalid, but that can cause no
// harm.)
```
?

kripken (approved):
lgtm with last comment. thanks!

PR #3931, opened 2021-06-11
[wasm-split] Add an option to emit a placeholder map

tlively:
The new instruction emits a file containing a map between placeholder index and
the name of the split out function that placeholder is replacing in the table.
This map is intended to be useful for debugging, as discussed in
https://github.com/emscripten-core/emscripten/issues/14330.

### DIFF
--- a/src/tools/wasm-split.cpp
+++ b/src/tools/wasm-split.cpp
@@ -813,7 +813,7 @@ void splitModule(const WasmSplitOptions& options) {
   }
   config.minimizeNewExportNames = !options.passOptions.debugInfo;
   auto splitResults = ModuleSplitting::splitFunctions(wasm, config);
-  std::unique_ptr<Module> secondary = std::move(splitResults.secondary);
+  auto& secondary = splitResults.secondary;
 
   adjustTableSize(wasm, options.initialTableSize);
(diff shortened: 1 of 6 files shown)


src/tools/wasm-split.cpp
```diff
@@ -795,8 +812,8 @@ void splitModule(const WasmSplitOptions& options) {
     config.newExportPrefix = options.exportPrefix;
   }
   config.minimizeNewExportNames = !options.passOptions.debugInfo;
-  std::unique_ptr<Module> secondary =
-    ModuleSplitting::splitFunctions(wasm, config);
+  auto splitResults = ModuleSplitting::splitFunctions(wasm, config);
+  std::unique_ptr<Module> secondary = std::move(splitResults.secondary);
```
kripken, line 816:
```suggestion
  auto& secondary = splitResults.secondary;
```

kripken: Seems simpler, but I don't feel strongly.

kripken (approved):
lgtm with a test for the new functionality (unless there's a reason not to?)

tlively:
Oh oops, I forgot to `git add` the test!

PR #3932, opened 2021-06-11
[Wasm GC] Add experimental support for non-nullable locals

kripken:
This adds a new feature flag, `GCNNLocals` that enables support for
non-nullable locals. No validation is applied to check that they are
actually assigned before their use yet - this just allows experimentation
to begin.

This feature is *not* enabled by default even with `-all`. If we enabled it,
then it would take effect in most of our tests and likely confuse current
users as well. Instead, the flag must be opted in explicitly using
`--enable-gc-nn-locals`. That is somewhat awkward, but as this is a
temporary feature flag it seems reasonable - that is, once the spec
stabilizes, we will remove the feature anyhow when we implement the
final status of non-nullability.


src/wasm-features.h
```diff
@@ -134,7 +139,16 @@ struct FeatureSet {
...
-  void setAll(bool v = true) { features = v ? All : MVP; }
+  void setGCNNLocals(bool v = true) { set(GCNNLocals, v); }
+  void setAll(bool v = true) {
+    features = v ? All : MVP;
+    // By default disable GCNNLocals, forcing the user to opt in to that feature
+    // explicitly. That is, wasm-opt -all will enable GC but *not* enable
+    // non-nullable locals. To get them, do wasm-opt -all --enable-gc-nn-locals
+    // FIXME: When the wasm spec stabilizies, this feature will go away, and
+    //        this awkwardness with it.
+    setGCNNLocals(false);
```
tlively, line 150: It might be simpler to just have `All` not include the bit for `GCNNLocals`. Then whether `GCNNLocals` is enabled would be completely independent of all other feature flags, which I think is a simpler mental model.

tlively (commented):
LGTM besides this one comment.

src/wasm-features.h
kripken: I considered that, but it is riskier and would require more work, because there are a few places that loop over `All` (for example, printing), and those would break.

Instead I think it is pretty clear mentally to only modify `setAll`. That is, this new feature is a normal feature, except for the one oddity of not being set by default in `setAll`.

tlively: What if instead of unconditionally unsetting the feature in `setAll`, we instead just left it unchanged? That would be the simpler mental model IMO without the implementation downsides you mentioned.

src/wasm-features.h
```diff
@@ -134,7 +139,16 @@ struct FeatureSet {
...
     set(TypedFunctionReferences, v);
   }
-  void setAll(bool v = true) { features = v ? All : MVP; }
+  void setGCNNLocals(bool v = true) { set(GCNNLocals, v); }
+  void setAll(bool v = true) {
+    features = v ? All : MVP;
+    // By default disable GCNNLocals, forcing the user to opt in to that feature
+    // explicitly. That is, wasm-opt -all will enable GC but *not* enable
+    // non-nullable locals. To get them, do wasm-opt -all --enable-gc-nn-locals
+    // FIXME: When the wasm spec stabilizies, this feature will go away, and
```
aheejin, line 148:
```suggestion
    // FIXME: When the wasm spec stabilizes, this feature will go away, and
```

src/wasm-features.h
aheejin: Does `--enable-gc-nn-locals -all` actually disable gc-nn-locals then?

kripken: @tlively 

> What if instead of unconditionally unsetting the feature in setAll, we instead just left it unchanged?

The downside there is that then `wasm-opt -all` would enable non-nullable locals. A specific consequence of that is that we would not run the code to add in `ref.as_non_null`s to fix up new locals that we add that are non-nullable (as many passes do). And then our output would not work in VMs by default. Basically all our current users would need to either start using the non-nullable experiment in v8 as well, or start passing `wasm-opt -all --disable-wasm-nn-locals`.

As this is an experiment, I think it makes sense instead to not break out existing users, and to make this opt-in.

@aheejin 

> Does --enable-gc-nn-locals -all actually disable gc-nn-locals then?

Oh - good point! They would run in order, and so we would actually disable it... I'll fix that.

src/wasm-features.h
```diff
@@ -134,7 +139,25 @@ struct FeatureSet {
...
+    //        or it will go away.
+    if (v) {
+      // Leave the old GCNNLocals value unmodified. This makes things like
+      // --enable-gc-nn-locals -all work (that is, if we enable the feature,
+      // then -all does not disable it; it simply does not enable it by itself).
+      auto oldGCNNLocals = hasGCNNLocals();
+      features = All;
+      setGCNNLocals(oldGCNNLocals);
+    } else {
+      features = MVP;
```
aheejin, line 158: What does `setAll(false)` mean? I couldn't find that usage anywhere. If that is the same as `--mvp-features`, `--enable-gc-nn-locals --mvp-features` does not enable gc-nn-locals and it is intended, right?

aheejin (approved):

src/wasm-features.h
kripken: Yes, I'm not sure either why `setAll()` takes a parameter - for symmetry maybe? But it does seem like `setAll(true)` is the only use case...

Yes, `--enable-X --mvp-features` would disable X, effectively. The new feature behaves the same as all the others (unless I have a bug here).

tlively: I'm surprised that the `false` case still unsets the new feature given that the `true` case does not set it. But if we have no users of the `false` case, perhaps the better solution would be to just remove the parameter.

src/wasm-features.h
tlively: > The downside there is that then wasm-opt -all would enable non-nullable locals.

I would say the root issue causing that is that the default feature set includes non-nullable locals.

kripken: Fair enough, but then we return to the previous point that if we remove it from `All` then we'd need to update code in every current place that loops over All, like printing of features, iteration on features, etc. That's a much more intrusive change.

If we had a long-term use for this, then I'd suggest a larger refactor like adding a new set of ExperimentalFeatures that is handled differently than the normal ones. But as this is pretty temporary, and this PR works properly, and it seems to have little risk, I'd prefer to keep it as is. If you feel strongly though I can look into that refactoring. And e.g. if we think we'll have new experimental features over time then it might pay off eventually.

kripken:
Removed the parameter to `setAll`, and renamed `makeMVP` to `setMVP`. Now we have `setAll / setMVP / setX` for all features consistently, and only `setX` take a boolean parameter.

tlively (approved):
Thanks, this lgtm 👍

Issue #3903, opened 2021-05-23
no 32bit binaries... fails to compile with latest mingw


MyTDT-Mysoft:
so... i'm trying to compile this as part of the requirement for emscripten/emsdk, but theres no luck... would you guys be kind and provide 32bit binaries as well?

kripken:
I don't have a win32 machine but I tried to use [zig cc](https://andrewkelley.me/post/zig-cc-powerful-drop-in-replacement-gcc-clang.html) to make builds for that. First time I'm trying it, and it looks almost magical in how easy it is to cross-compile (I built win32 from my linux64, only issues were minor cmake things), but without a machine to test on I can't be sure it worked. Though they seem right as best I can tell:
```
$ file wasm*
wasm-as:                  PE32 executable (console) Intel 80386, for MS Windows
wasm-ctor-eval:           PE32 executable (console) Intel 80386, for MS Windows
wasm-dis:                 PE32 executable (console) Intel 80386, for MS Windows
wasm-emscripten-finalize: PE32 executable (console) Intel 80386, for MS Windows
wasm-metadce:             PE32 executable (console) Intel 80386, for MS Windows
wasm-opt:                 PE32 executable (console) Intel 80386, for MS Windows
wasm-reduce:              PE32 executable (console) Intel 80386, for MS Windows
wasm-shell:               PE32 executable (console) Intel 80386, for MS Windows
wasm-split:               PE32 executable (console) Intel 80386, for MS Windows
```
I added them to this gist:

https://gist.github.com/kripken/3cd771d3ee05e7fbc4f7f1b251f8d95e

You can download the zip of that to get the files. Please let me know if they work!

PR #3935, opened 2021-06-16
Preserve function heap types during text parsing

tlively:
Previously, ref.func instructions would be assigned the canonical (i.e. first
parsed) heap type for the referenced function signature rather than the HeapType
actually specified in the type definition. In nominal mode, this could cause
validation failures because the types assigned to ref.func instructions would
not be correct.

Fix the problem by tracking function HeapTypes rather than function Signatures
when parsing the text format.

There can still be validation failures when round-tripping modules because
function HeapTypes are not properly preserved after parsing, but that will be
addressed in a follow-up PR.


src/wasm/wasm-s-parser.cpp
```diff
@@ -3507,7 +3512,9 @@ void SExpressionWasmBuilder::parseEvent(Element& s, bool preParseImport) {
   event->attribute = atoi(attrElem[1]->c_str());
 
   // Parse typeuse
-  i = parseTypeUse(s, i, event->sig);
+  HeapType eventType;
+  i = parseTypeUse(s, i, eventType);
+  event->sig = eventType.getSignature();
```
kripken, line 3517: This repeated pattern is slightly annoying, but I can't think of an obvious improvement...

kripken (approved):

src/wasm/wasm-s-parser.cpp
tlively: A (large) follow-up will change functions and events to store HeapTypes rather than signatures, so this particular pattern will go away.

tlively:
The CI errors here will be fixed by #3938.

tlively:
This is deterministically failing the Windows build, but I have no idea why that would be 🤯

kripken:
I think on windows that code (`0xc0000005`) means a segfault. Running a debug build in a debugger on a windows VM might give a quick answer.

(Or, if you're adventurous, you might be able to trigger the bug using `zig cc` to [cross-compile to windows](https://github.com/WebAssembly/binaryen/issues/3903#issuecomment-846599069) and then running the output in qemu...)

src/wasm/wasm-validator.cpp
```diff
@@ -2318,10 +2318,12 @@ void FunctionValidator::visitStructNew(StructNew* curr) {
...
-        shouldBeSubType(curr->operands[i]->type,
-                        fields[i].type,
-                        curr,
-                        "struct.new operand must have proper type");
+        shouldBeSubType(
+          curr->operands[i]->type,
+          fields[i].type,
+          curr,
+          ("struct.new operand " + std::to_string(i) + " must have proper type")
+            .c_str());
```
kripken, line 2326: I wonder if this is not undefined behavior. You are creating a temporary `std::string`, then getting a C string from it, but the `std::string` may not live long enough for the call to occur. Can try to revert this to see if it fixes the windows bot...

tlively:
Ahhhh it auto-merged all my debugging changes 😭

kripken:
What was the solution to the windows issue?

kripken:
This appears to break the binaryen roller:

https://logs.chromium.org/logs/emscripten-releases/buildbucket/cr-buildbucket.appspot.com/8843703665766206672/+/steps/Build_Binaryen/0/stdout

tlively:
This commit was the fix: https://github.com/WebAssembly/binaryen/pull/3935/commits/5d018f3b208e0cc3fb652deb8de53ee6f72f6e6f. I'm fairly certain this should be a no-op change, so I think we were just hitting a miscompilation before :/

kripken:
Oh wow... [it's never the compiler, until it is](https://screenrant.com/house-never-lupus-diagnosis-episode-flynn/).

kripken:
@tlively Looks like this causes breakage in the fuzzer, e.g., seed `2099213060994934502`.

Simplied STR: `wasm-opt a.wasm -all -o b.wasm` will segfault.

[a.wasm.gz](https://github.com/WebAssembly/binaryen/files/6718547/a.wasm.gz)

kripken:
Hmm, the segfault is not deterministic actually - let me run that bisection again, sorry for the noise...

kripken:
Ok, a proper bisection shows the culprit is actually #3933

Before the PR, we used to emit
```
[parse exception: bad type form -36 (at 0:95)]
```
on that wasm with that command. After that PR, we segfault.

Separately from that, we need to figure out why the fuzzer is emitting an invalid binary, but I hope that fixing the segfault might also help understand that issue as well.

kripken:
I redid the bisection. Indeed #3933 is the culprit - it is when we start to emit a bad binary in the fuzzer (using the seed mentioned earlier). This PR that landed later changes the behavior on the bad binary, but that's it.

PR #3936, opened 2021-06-16
[Wasm GC] rtt.fresh_sub

kripken:
This is the same as rtt.sub, but creates a "new" rtt each time. See

https://docs.google.com/document/d/1DklC3qVuOdLHSXB5UXghM_syCh-4cMinQ50ICiXnK3Q/edit#

The old Literal implementation of rtts becomes a little more complex here,
as it was designed for the original spec where only structure matters. It may
be worth a complete redesign there, but for now as the spec is in flux I think
the approach here is good enough.


kripken:
(clang-tidy lint error is the existing one from before... I would appreciate help figuring out how best to get that ignored...)

tlively (approved):

PR #3937, opened 2021-06-17
[EH] Replace event with tag

aheejin:
We recently decided to change 'event' to 'tag', and  'event section' to
'tag section', out of the rationale that the section contains a
generalized tag that references a type, which may be used for something
other than exceptions, and the name 'event' can be confusing in the web
context.

See
- https://github.com/WebAssembly/exception-handling/issues/159#issuecomment-857910130
- https://github.com/WebAssembly/exception-handling/pull/161


kripken (approved):

aheejin:
Don't know what's going on with clang-tidy... Anyone has a clue? https://github.com/WebAssembly/binaryen/pull/3937/checks?check_run_id=2844692867

kripken:
No idea... that started happening in every PR that modifies the delegations headers. It's a false positive, but I don't know how to ignore it...

tlively:
We explicitly `#error` out in that file if any of the `DELEGATE_FIELD_*` macros are not defined. In normal usage, we expect those macros to be defined by the including source file. Clang-tidy looks at the header in isolation, though, so it never has a chance to define the macros and our own errors are being triggered.

I haven't seen any configuration options in clang-tidy that could fix this, but if we wanted to get clang-tidy working correctly we could downgrade these `#error`s to `#warning`s. We could also do something more complicated like error out if any macros are missing only if at least one is supplied.

aheejin:
Uploaded #3941 to fix the clang-tidy issue. If that lands, I can rebase this PR onto that.

PR #3938, opened 2021-06-17
[wasm2js] Refactor assertion parsing

tlively:
Assertions were previously parsed by replacing "invoke" with "call" and using
the normal s-expr parser. The parseCall method of the s-expr parser uses the
call target to look up the correct signature on the module, but the invoke
targets in assertions use export names rather than internal function names, so
the signature lookups were inserting new bogus entries with default values.

This issue didn't seem to cause any big problems before, but #3935 turns it into
a hard error because the default `HeapType` does not have an associated
signature.

Fix the problem (at least in the common case of trivial arguments and expected
results) by manually construction a `Call` expression rather than depending on
the s-expr parser to construct it.

### DIFF
--- a/src/tools/wasm2js.cpp
+++ b/src/tools/wasm2js.cpp
@@ -839,8 +839,7 @@ void AssertionEmitter::emit() {
     Element& e = *root[i];
     if (e.isList() && e.size() >= 1 && e[0]->isStr() &&
         e[0]->str() == Name("module")) {
-      wasm.~Module();
-      new (&wasm) Module();
+      ModuleUtils::clearModule(wasm);
       std::stringstream funcNameS;
       funcNameS << ASM_FUNC.c_str() << i;
       std::stringstream moduleNameS;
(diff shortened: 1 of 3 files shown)


src/tools/wasm2js.cpp
```diff
@@ -812,17 +833,20 @@ void AssertionEmitter::emit() {
...
   Builder wasmBuilder(sexpBuilder.getModule());
   Name asmModule = std::string("ret") + ASM_FUNC.str;
+  // Track the last built module.
+  Module wasm;
   for (size_t i = 0; i < root.size(); ++i) {
     Element& e = *root[i];
     if (e.isList() && e.size() >= 1 && e[0]->isStr() &&
         e[0]->str() == Name("module")) {
+      wasm.~Module();
+      new (&wasm) Module();
```
kripken, line 843:
```suggestion
      ModuleUtils::clearModule(wasm);
```
(that literally does those two lines)

kripken (approved):

src/tools/wasm2js.cpp
tlively: Ah, I had thought we had a function for doing that, but I was looking in the wrong place. Thanks!

PR #3939, opened 2021-06-17
Add a extract-function-index pass

tlively:
This is a useful alternative to extract-function when you don't know the
function's name.

Also moves the extract-function tests to be lit tests and re-uses them as
extract-function-index tests.


src/passes/pass.cpp
```diff
@@ -129,8 +129,11 @@ void PassRegistry::registerPasses() {
   registerPass("extract-function",
                "leaves just one function (useful for debugging)",
                createExtractFunctionPass);
-  registerPass(
-    "flatten", "flattens out code, removing nesting", createFlattenPass);
+  registerPass("extract-function-index",
+               "leaves just one function selected by index",
+               createExtractFunctionIndexPass),
```
kripken, line 134:
```suggestion
               createExtractFunctionIndexPass);
```
then the indentation below would be fixed by clang-format

test/lit/passes/extract-function-table-unused.wast
```diff
@@ -0,0 +1,26 @@
...
+;; CHECK-NEXT:  (type $none_=>_none (func))
+;; CHECK-NEXT:  (import "env" "other" (func $other))
+;; CHECK-NEXT:  (export "foo" (func $foo))
+;; CHECK-NEXT:  (func $foo
+;; CHECK-NEXT:   (nop)
+;; CHECK-NEXT:  )
+;; CHECK-NEXT: )
+
+(module
+  ;; Use another function in the table, but the table is not used in the
```
kripken, line 16:
```suggestion
  ;; Use a function in the table, but the table is not used in the
```
A downside of separate lit files is that we can't have a "conversation" about their connections...

test/lit/passes/extract-function-table-used.wast
```diff
@@ -0,0 +1,32 @@
...
+;; CHECK-NEXT:  (export "foo" (func $foo))
+;; CHECK-NEXT:  (func $foo
+;; CHECK-NEXT:   (call_indirect (type $none)
+;; CHECK-NEXT:    (i32.const 10)
+;; CHECK-NEXT:   )
+;; CHECK-NEXT:  )
+;; CHECK-NEXT: )
+
+(module
+  ;; Use another function in the table, and the table *is* used. As a result,
```
kripken, line 20:
```suggestion
  ;; Use a function in the table, and the table *is* used. As a result,
```

kripken (approved):

tlively:
@kripken, I added a way to have multiple modules in a single lit test. PTAL!

test/lit/passes/extract-function.wast
```diff
@@ -0,0 +1,72 @@
...
+;; RUN: foreach %s %t wasm-opt --extract-function-index --pass-arg=extract-function-index@0 -S -o - | filecheck %s
+
+;; CHECK:      (module
+;; CHECK-NEXT:  (type $none_=>_none (func))
+;; CHECK-NEXT:  (import "env" "bar" (func $bar))
+;; CHECK-NEXT:  (export "foo" (func $foo))
+;; CHECK-NEXT:  (func $foo
+;; CHECK-NEXT:   (call $bar)
+;; CHECK-NEXT:  )
+;; CHECK-NEXT: )
```
kripken, line 13: These are manually-created, correct? I worry about us accumulating a lot of those and having a bad time some day when we need to update them. Adding `foreach` seems like it would increase the amount of such test code... How hard would it be to support auto updating in these eventually?

kripken (approved):

test/lit/passes/extract-function.wast
tlively: Yes, these are manually created for now. I think it would be reasonable to update the auto update script in the near future to handle multiple multiple modules in a file and also to give it an option to emit checks on a per-module rather than a per-function basis.

PR #3941, opened 2021-06-18
Rename wasm-delegations[-fields].h to def files (NFC)

aheejin:
These files are special in that they use define symbols that are not
defined within those files or other files included in those files; they
are supposed to be defined in source files that include these headers.
This has caused clang-tidy to fail every time these files have changed
because they are not compilable per se.

This PR solves the problem by changing their extension to `def`, which
is also used in LLVM codebase. LLVM has dozens of files like this whose
extension is `def`, which makes these not checked by clang-tidy.


tlively (approved):
Good thinking! This is a much better solution than what I was thinking of :)

kripken (approved):
Very nice!

aheejin:
@tlively Out of curiosity, what was the solution you were thinking of?

tlively:
The workarounds I mentioned in https://github.com/WebAssembly/binaryen/pull/3937#issuecomment-863240184. Downgrading the errors to warnings would have been an unfortunate loss of functionality and avoiding the errors in the special case where none of the macros are defined would have been unnecessarily complicated.

PR #3943, opened 2021-06-18
[Wasm GC] Fix RSE on non-nullable locals

### LINKED ISSUE #3942: Assert when enabling non-nullable locals
When optimizing this program (which doesn't even contain any non-nullable locals) with `-all --enable-gc-nn-locals -O2`:
```wat
(module
 (type $type$0 (struct (field (mut i32)) (field (mut i32)) (field (mut i64)) (field (mut (ref null $type$0)))))
 (type $none_=>_none (func))
 (type $type$22 (func (param (ref $type$0))))
 (export "f" (func $10))
 (func $10
  (local $0 (ref null $type$0))
  (call $11
   (ref.as_non_null
    (local.get $0)
   )
  )
 )
 (func $11 (param $0 (ref $type$0))
  (struct.set $type$0 2
   (local.get $0)
   (i64.const 0)
  )
  (struct.set $type$0 3
   (local.get $0)
   (ref.null $type$0)
  )
 )
)
```
Binaryen hits this assert:
```
wasm-opt: /usr/local/google/home/askesc/checkout/binaryen/src/literal.h:252: static wasm::Literal wasm::Literal::makeNull(wasm::Type): Assertion `type.isNullable()' failed.
```
The program as-is obviously traps on the first `ref.as_non_null`, but it's reduced from a bigger program (Dart MergeSort benchmark) which doesn't trap and hits the same assert.

kripken:
RedundantSetElimination checks if a local already has the default
value when we assign the default to it. For a non-nullable local, however,
there is no initial value - it cannot be used before it is assigned to. So
we just need to skip such locals, and not assume they contain a default
value we can compare against (we would assert on trying to create a
"zero" for such a non-nullable type).

Fixes #3942


tlively (approved):
Nice! For the test, I assume the local would be removed if its type were nullable?

kripken:
This pass actually removes redundant sets. So if there were a set it could remove, it would do so. Another pass (ReorderLocals) removes unused locals. Before the PR, that testcase would crash on the assert mentioned above.

PR #3945, opened 2021-06-18
[Wasm GC] Fix inlining + non-nullable locals

### LINKED ISSUE #3944: Assert when enabling non-nullable locals, take 2
The fix for https://github.com/WebAssembly/binaryen/issues/3942 still fails for [the original program](https://github.com/WebAssembly/binaryen/files/6679233/m13.wasm.gz). A different reduction that still hits the assert is:
```wat
(module
 (type $type$0 (struct (field (mut i32))))
 (type $type$1 (func))
 (type $type$2 (struct (field (mut i32)) (field (mut i64))))
 (type $type$3 (struct (field (mut i32)) (field (mut f64))))
 (type $type$4 (func (param (ref $type$0) (ref $type$0)) (result (ref $type$0))))
 (type $type$5 (func (param i64) (result i64)))
 (global $global$0 (rtt 0 $type$0) (rtt.canon $type$0))
 (global $global$23 (rtt 1 $type$0) (rtt.sub $type$0
  (global.get $global$0)
 ))
 (global $global$24 (rtt 2 $type$0) (rtt.sub $type$0
  (global.get $global$23)
 ))
 (global $global$25 (rtt 3 $type$3) (rtt.sub $type$3
  (global.get $global$24)
 ))
 (global $global$29 (rtt 2 $type$0) (rtt.sub $type$0
  (global.get $global$23)
 ))
 (global $global$30 (rtt 3 $type$2) (rtt.sub $type$2
  (global.get $global$29)
 ))
 (export "f" (func $4))
 (func $4
  (call $5)
 )
 (func $5
  (call $14)
 )
 (func $14
  (drop
   (ref.cast
    (call $15
     (struct.new_with_rtt $type$2
      (i32.const 30)
      (i64.const 87)
      (global.get $global$30)
     )
     (struct.new_with_rtt $type$2
      (i32.const 30)
      (i64.const 87)
      (global.get $global$30)
     )
    )
    (global.get $global$30)
   )
  )
 )
 (func $15 (param $0 (ref $type$0)) (param $1 (ref $type$0)) (result (ref $type$0))
  (if (result (ref $type$0))
   (i32.eq
    (struct.get $type$0 0
     (local.get $1)
    )
    (i32.const 30)
   )
   (struct.new_with_rtt $type$2
    (i32.const 30)
    (call $16
     (struct.get $type$2 1
      (ref.cast
       (local.get $0)
       (global.get $global$30)
      )
     )
    )
    (global.get $global$30)
   )
   (struct.new_with_rtt $type$3
    (i32.const 25)
    (f64.const 87)
    (global.get $global$25)
   )
  )
 )
 (func $16 (param $0 i64) (result i64)
  (i64.const 87)
 )
)
```

kripken:
We don't need to assign a zero value for such locals (and we can't, as no
default value exists for them).

Fixes #3944


src/passes/Inlining.cpp
```diff
@@ -292,9 +292,16 @@ doInlining(Module* module, Function* into, const InliningAction& action) {
   // Zero out the vars (as we may be in a loop, and may depend on their
   // zero-init value
   for (Index i = 0; i < from->vars.size(); i++) {
+    auto type = from->vars[i];
+    if (type.isRef() && !type.isNullable()) {
+      // Non-nullable locals do not need to be zeroed out. They have no zero
+      // value, and by definition should not be used before being read from, so
```
aheejin, line 298:
```suggestion
      // value, and by definition should not be used before being written to, so
```
Is this what you meant?

test/lit/passes/inlining-gc.wast
```diff
@@ -0,0 +1,45 @@
...
+ ;; CHECK:      (func $caller-n
+ ;; CHECK-NEXT:  (local $0 funcref)
+ ;; CHECK-NEXT:  (block $__inlined_func$target-n
+ ;; CHECK-NEXT:   (local.set $0
+ ;; CHECK-NEXT:    (ref.null func)
+ ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:   (nop)
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT: )
+ (func $caller-n
```
aheejin, line 14: Nit: How about `n`  -> `nullable` and `nn` -> `non-nullable` in function names? Not a big deal but will be slightly easier to read later.

aheejin (approved):

src/passes/Inlining.cpp
kripken: Yes, good catch.

test/lit/passes/inlining-gc.wast
kripken: Done.

PR #3946, opened 2021-06-18
Remove (attr 0) from tag text format

aheejin:
This attribute is always 0 and reserved for future use. In Binayren's
unofficial text format we were writing this field as `(attr 0)`, but we
have recently come to the conclusion that this is not necessary.

Relevant discussion:
https://github.com/WebAssembly/exception-handling/pull/160#discussion_r653254680


tlively (approved):
Perhaps you already have this planned for a follow-up PR, but it would also make sense to remove the attribute field from the `Event` class for similar reasons.

aheejin:
You mean, remove the field altogether and just write `0` in its place instead? That sounds OK, but not very sure it's worth the hassle of deleting the `enum` and such, in case we add another field to that `enum`. Usually a single module will have just a few (actually, mostly 1) tag anyway, so the memory saving will not be noticeable either.

tlively:
Yeah, agreed it probably doesn't make much of a difference. It would be slightly simpler overall to treat the 0 as purely an encoding detail, though.

aheejin:
Not very sure what "treat the 0 as purely an encoding detail" means. I thought we were talking about one of these:

```cpp
// src/wasm.h
enum WasmTagAttribute : uint8_t { WASM_TAG_ATTRIBUTE_EXCEPTION = 0x0 };

class Tag : public Importable {
public:
  uint8_t attribute = WASM_TAG_ATTRIBUTE_EXCEPTION;
  Signature sig;
};

// src/wasm/wasm-binary.cpp
o << uint8_t(tag->attribute);
```
vs.
```cpp
// src/wasm.h
class Tag : public Importable {
public:
  Signature sig;
};

// src/wasm/wasm-binary.cpp
o << uint8_t(tag->attribute);
```

Not sure why the latter treats 0 as purely an encoding detail while the former does not..?

By the way, another reason I prefer the status quo is it make it clear that what the reserved field is. If we write just `0` we will have to write some comments on what that 0 is. But I admit this is a pretty nitty point though.

tlively:
Yes, we’re thinking about the same potential change. By “pure encoding detail” I mean that only the binary parser and binary writer need to know that it exists at all. This is similar to Rossberg’s suggestion that it only be mentioned in the binary format part of the spec.

aheejin:
> Yes, we’re thinking about the same potential change. By “pure encoding detail” I mean that only the binary parser and binary writer need to know that it exists at all. This is similar to Rossberg’s suggestion that it only be mentioned in the binary format part of the spec.

Done in #3947.

PR #3947, opened 2021-06-20
[EH] Make tag's attribute encoding detail

aheejin:
This removes `attribute` field from `Tag` class, making the reserved and
unused field known only to binary encoder and decoder. This also removes
the `attribute` parameter from `makeTag` and `addTag` methods in
wasm-builder.h, C API, and Binaryen JS API.

Suggested in
https://github.com/WebAssembly/binaryen/pull/3946#pullrequestreview-687756523.


tlively (approved):
🎉 Thanks! I think this is a good simplification.

PR #3950, opened 2021-06-22
Revert accidental debugging changes from #3935

tlively:


tlively:
Landing TBR since this never should have been in the tree to begin with.

kripken:
Ah, maybe this will fix the roller breakage from the last PR, if that was from the debugging changes...

tlively:
Yes, I think this should fix that breakage.

PR #3952, opened 2021-06-24
Preserve Function HeapTypes

tlively:
When using nominal types, func.ref of two functions with identical signatures
but different HeapTypes will yield different types. To preserve these semantics,
Functions need to track their HeapTypes, not just their Signatures.

This PR replaces the Signature field in Function with a HeapType field and adds
new utility methods to make it almost as simple to update and query the function
HeapType as it was to update and query the Function Signature.


src/tools/wasm2js.cpp
```diff
@@ -658,7 +658,7 @@ Ref AssertionEmitter::emitAssertReturnFunc(Builder& wasmBuilder,
   std::unique_ptr<Function> testFunc(
     wasmBuilder.makeFunction(testFuncName,
                              std::vector<NameType>{},
-                             body->type,
+                             Signature(Type::none, body->type),
```
kripken, line 661: How did this work before?

src/wasm-builder.h
```diff
@@ -981,7 +984,7 @@ class Builder {
   }
 
   static void clearLocals(Function* func) {
-    func->sig.params = Type::none;
+    func->type = Signature(Type::none, func->getResults());
     func->vars.clear();
     clearLocalNames(func);
   }
```
kripken, line 990: huh, this is an odd API... removing params as it does.

It looks like there are no uses of it anywhere. I suggest we delete it.

src/wasm-interpreter.h
```diff
@@ -2425,6 +2425,7 @@ template<typename GlobalManager, typename SubType> class ModuleInstanceBase {
   void initializeTableContents() {
     ModuleUtils::iterActiveElementSegments(wasm, [&](ElementSegment* segment) {
       Function dummyFunc;
+      dummyFunc.type = Signature(Type::none, Type::none);
```
kripken, line 2428: Hmm, Signature having a default value but HeapType not does mean we may have bugs in places we missed adding such a thing... but seems not likely to be a problem.

src/wasm.h
```diff
@@ -1658,6 +1658,12 @@ class Function : public Importable {
     delimiterLocations;
   BinaryLocations::FunctionLocations funcLocation;
 
+  Signature getSig() { return type.getSignature(); }
+  Type getParams() { return getSig().params; }
+  Type getResults() { return getSig().results; }
+  void setParams(Type params) { type = Signature(params, getResults()); }
+  void setResults(Type results) { type = Signature(getParams(), results); }
```
kripken, line 1665: the set operations are significantly slower than before, correct? What about the others?

It may be worth doing some benchmarking on this before landing, just to be safe.

kripken (approved):
Nice work! New API looks like a good choice.

lgtm % comments.

src/tools/wasm2js.cpp
tlively: This builder API used to take only the function return type, `body->type`. Now it takes a HeapType implicitly constructed from the function signature.

src/wasm-interpreter.h
tlively: It's probably worth giving Function a default HeapType corresponding to the empty signature. I'll do that.

src/wasm.h
tlively: Yes, constructing or finding the new HeapType is much more expensive than what we previously did, which was to just set a field directly in `Function::sig`. I'm not too worried about that because changing the Function parameters or results is not done except in DAE in the common optimization pass.

`getParams` and `getResults` are also more expensive than before because they now have to chase a pointer (the HeapType ID) rather than directly read the field of the signature in the Function.

I'll benchmark this a bit and see what I find.

tlively: It doesn't look like this makes an appreciable performance difference when running `wasm-opt -O2` on the poppler binary.

kripken: Might be worth checking on a large GC testcase too?

tlively: Ok, looks like this might slow things down by 0.6% (I don't know how statistically significant that is). This seems acceptable to me.

kripken: sgtm, thanks for checking.

PR #3954, opened 2021-06-25
[Wasm GC] Fix LinearExecutionWalker

kripken:
The pass did not mention BrOn, which led to it doing incorrect work in
SimplifyLocals.

This PR also moves that code to a new place. The movement is
important because the actual fix is this:
```
      case Expression::Id::BrOnId: {
        self->pushTask(SubType::doVisitBrOn, currp);
        self->pushTask(SubType::doNoteNonLinear, currp);
        self->maybePushTask(SubType::scan, &curr->cast<BrOn>()->rtt);
        self->pushTask(SubType::scan, &curr->cast<BrOn>()->ref);
        break;
      }
      default: {
        // All relevant things should have been handled.
        assert(!Properties::isControlFlowStructure(curr));
        assert(!Properties::isBranch(curr));
```
Note the assertions at the end, that aim to prevent future issues. I
didn't want to include "properties.h" in this header (which is used by
almost everything else), and splitting out this rarely-used class seems
better anyhow.

The rest of the fix is to make SimplifyLocals not assume that things
are a Switch if they are not an if/block/etc. That is, we handle BrOn
together with Switch now, and anything else we might add later.

@askeksa-google with this I can compile your recent large benchmark
with `-Os`, in hopes of improving code size. That saves 9% (but it
might affect speed too).


aheejin:
Can we split this into two PRs so the first one is NFC and only moves code and the second one does the fix? I think the diff will be easier to view that way..

kripken:
Good idea, here is the NFC part: #3956

kripken:
Rebased after the NFC refactoring landed.

src/passes/SimplifyLocals.cpp
```diff
@@ -140,9 +140,9 @@ struct SimplifyLocals
     } else if (curr->is<If>()) {
       assert(!curr->cast<If>()
                 ->ifFalse); // if-elses are handled by doNoteIf* methods
-    } else if (curr->is<Switch>()) {
-      auto* sw = curr->cast<Switch>();
-      auto targets = BranchUtils::getUniqueTargets(sw);
+    } else {
+      // TODO support BrOn, Switch, etc. where possible
+      auto targets = BranchUtils::getUniqueTargets(curr);
```
aheejin, line 145: Is this currently supporting `BrOn` and `Switch` or not? The `TODO` sounds like it isn't, but then what is this code doing?

Also, is there any reason the support for `BrOn` or `Switch` should be different from that of `Break`, which is handled above, given that we are using generic methods like `getUniqueTargets`?

kripken: I'll clarify that better, good point. Basically, the code supports BrOn and Switch, but it ignores them. The TODO is to actually optimize them.

I think you're right that we could use generic methods to handle them. We'd also need to generically check for an existing return value (as the Break code does), as well as for the number of targets (Break can assume there is just one). For now though I'd prefer to not look into those things.

aheejin (approved):

PR #3955, opened 2021-06-25
[Wasm GC] Add support for non-nullable locals in binary reading

### LINKED ISSUE #3953: Non-nullable locals mode not producing non-nullable locals
I have been trying out the `--enable-gc-nn-locals` flag on some output [modules](https://github.com/WebAssembly/binaryen/files/6716379/MergeSort.zip) from `dart2wasm`. I was surprised to find that the optimized Binaryen output when using non-nullable locals was bigger than the one using only nullable locals, even though for the original modules, the one using non-nullable locals was smaller.

I looked at the output and noticed a few things:

- The optimized non-nullable version doesn't actually contain any non-nullable locals. This is the case even for a non-optimizing wasm-to-wat pass-through.
- The code contains many superfluous `ref.as_non_null`, i.e. on a non-nullable value or as input to something accepting a nullable value.

It looks like the code was processed with non-nullable locals in mind, but for the output, the locals somehow became nullable and all the `ref.as_non_null` were inserted to make the types match up.

Since at least some of these could potentially trap, it would seem that removing them is only valid with the `--ignore-implicit-traps` option. Thus, a fix for https://github.com/WebAssembly/binaryen/issues/3934 could be valuable for doing a fair comparison between having non-nullable locals and not.

kripken:
We only tested that feature on the text format. For binary support, the reader needs
to know that the feature is enabled.

@tlively This sadly gets back to the issue of when to apply the features. This applies
them before and after, which works, but I think we agreed before is not great. With
typed-function-references we found a workaround (always use the refined type), but
I can't think of such a thing here that is not annoying. Thoughts?

Fixes #3953

@askeksa-google This PR may be useful while you wait.


tlively:
@kripken, what would you think about adding a pass in the binary parser to seek out and parse the target features section before parsing anything else? We're already loading full binaries into memory before parsing them, so I don't think this would have much of a performance impact.

Somewhat separate from this particular PR but on the topic of applying features:

The current model is that the command line feature flags, if they're supplied, must explicitly enable a superset of the features enabled by the target features section. This is annoying because enabling one additional feature on the command line requires either providing all the flags for the features in the target features section as well or providing the strange `--detect-features` flag.

I would be open to simplifying this to a model where the command line feature flags are applied on top of and in addition to the target feature section, rather than having to explicitly be a superset. We could emit a warning whenever the flags would try to disable a feature present in the target features section.

kripken:
Pre-parsing the features section might not be that bad. Are you saying that if we pre-parse it, then we can set the features once in that pre-parsing, and not once before and once after?

Another option here is for `--enable-FOO` to do two things: set the feature `FOO` on the module, and also keep it around for aligning with the feature section later. Then we'd only call applyFeatures once. (but roundtrip would still do it twice)

The simplification you suggest in feature handling sgtm. Would that help here?

tlively:
> Pre-parsing the features section might not be that bad. Are you saying that if we pre-parse it, then we can set the features once in that pre-parsing, and not once before and once after?

Hmm, I was thinking that if we pre-parse the features section, then in this PR you wouldn't need to apply the command line features before the parse to have the correct features enabled. But that won't be sufficient if the target features section is absent (for example when parsing a text module) or if it is present but does not contain the relevant feature needed for the parse.

> Another option here is for `--enable-FOO` to do two things: set the feature `FOO` on the module, and also keep it around for aligning with the feature section later. Then we'd only call applyFeatures once. (but roundtrip would still do it twice)
> 
> The simplification you suggest in feature handling sgtm. Would that help here?

Actually, something slightly different would help here. We could apply the features from the command line just once _before_ parsing the module and let the target features section add additional features on top of that (probably before parsing the rest of the module if the enabled features can affect parsing). Then we would emit a warning whenever the target features section disables a feature that was enabled by the command line flags (which is the opposite of what I suggested in my previous comment).

kripken:
Interesting idea @tlively , that does sound like it could be nice and simple. I'll work on a patch to see.

kripken:
This PR has been rebased on top of #3960 and is now very simple!

tlively (approved):
🎉

PR #3956, opened 2021-06-25
Refactor LinearExecutionWalker to a separate file. NFC

kripken:
Separated from #3954


aheejin (approved):
LGTM, but what's the reason you split this part into another file?

tlively (approved):

kripken:
> LGTM, but what's the reason you split this part into another file?

1. It's only used in a few places, so faster for compile times to split it out.
2. It will help #3954 in which we want to add more to that header.

PR #3957, opened 2021-06-26
Generate FileCheck checks for all module items

tlively:
Instead of only generating checks for functions, generate checks for all named top-level module items, such as types, tags, tables, and memories. Because module items can be in different orders in the input and the output but FileCheck checks must follow the order of the output, we need to be slightly clever about when we emit the checks. Consider these types in the input file:

```
  (type $A (...))
  (type $B (...))
```

If their order is reversed in the output file, then the checks for $B need to be emitted before the checks for $A, so the resulting module will look like this:

```
  ;; CHECK: (type $B (...))
  ;; CHECK: (type $A (...))
  (type $A (...))
  (type $B (...))
```

Rather than this, which looks nicer but would be incorrect:

```
  ;; CHECK: (type $A (...))
  (type $A (...))
  ;; CHECK: (type $B (...))
  (type $B (...))
```


scripts/update_lit_checks.py
```diff
@@ -31,9 +31,15 @@
...
 NOTICE = (f';; NOTE: Assertions have been generated by {script_name} and ' +
           'should not be edited.')
+
+
 RUN_LINE_RE = re.compile(r'^\s*;;\s*RUN:\s*(.*)$')
 CHECK_PREFIX_RE = re.compile(r'.*--check-prefix[= ](\S+).*')
-FUNC_RE = re.compile(r'(^\s*)\(func \$(\S*).*$', re.MULTILINE)
+
+items = ['type', 'import', 'global', 'memory', 'data', 'table', 'elem', 'tag',
+         'export', 'start', 'func']
```
kripken, line 40: can we find all of these based on their indentation, as opposed to a hardcoded list?

test/lit/forward-declared-types.wast
```diff
@@ -1,17 +1,16 @@
...
 ;; RUN: wasm-opt %s -all -S -o - | filecheck %s
 
-;; CHECK: (type $func (func))
-;; CHECK: (type $none_=>_ref?|$struct| (func (result (ref null $struct))))
-;; CHECK: (type $struct (struct (field (ref $array)) (field (ref null $func))))
-;; CHECK: (type $array (array (rtt 2 $func)))
-
 (module
+  ;; CHECK:      (type $func (func))
+  ;; CHECK:      (type $struct (struct (field (ref $array)) (field (ref null $func))))
```
kripken, line 8: This reordering is definitely confusing at first, but seems unavoidable for the reasons you mentioned. Ideally filecheck could have an "ignore order" mode, but I guess this way we also check the order does not change.

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -612,8 +615,12 @@
     )
   )
 
+  ;; CHECK:      (func $nothing
+  ;; CHECK-NEXT:  (nop)
+  ;; CHECK-NEXT: )
   (func $nothing)
```
kripken, line 621: this also changes behavior on a single-line function. lgtm, but was that intentional?

kripken (approved):
I assume all the test changes here are the new automatic output?

test/lit/forward-declared-types.wast
tlively: Full-fledged FileCheck in LLVM actually does support ignoring order, but I'm not sure the python package does.

test/lit/passes/optimize-instructions-gc.wast
tlively: Yes, I intentionally updated the regex to handle this correctly.

scripts/update_lit_checks.py
tlively: In principle, yes, but I think it would be complex enough that it would be better to go with the list. This script is already more complex than I would like it to be :/

tlively:
> I assume all the test changes here are the new automatic output?

Yes, although I also updated the spacing on some tests to be more readable in the presence of the new checks.

PR #3958, opened 2021-06-29
Only set `supertype` if nominal typing is enabled

tlively:
The code for printing and emitting the experimental nominal type constructors
added in #3933 assumes that supertypes were only returned from `getSuperType`
when nominal typing was enabled. `getSuperType` in turn was assuming that the
supertype field would only be set if nominal typing was enabled, but this was
not the case. This bug caused use-after-free errors because equirecursive
canonicalization left the supertype field pointing to a temporary HeapTypeInfo
that would be freed at the end of parsing but then accessed during module
writing.

To fix the issue, only set `supertype` if nominal typing is enabled, as
originally intended.


tlively:
Thanks for the pointer earlier, @kripken! This turned out to be totally trivial after all.

kripken (approved):
Great!

PR #3959, opened 2021-06-29
Set an absolute maximum inlining limit

kripken:
We have seen some cases in both Chrome and Firefox where
extremely large modules cause overhead,

https://github.com/WebAssembly/binaryen/pull/3730#issuecomment-867939138 (and link therein)
https://github.com/emscripten-core/emscripten/issues/13899#issuecomment-825073344

There is no "right" value to use as a limit here, but pick an
arbitrary one that is very high. (This setting is verified to have
no effect on the emscripten benchmark suite.)

### DIFF
--- a/src/passes/Inlining.cpp
+++ b/src/passes/Inlining.cpp
@@ -420,6 +420,9 @@ struct Inlining : public Pass {
           continue;
         }
         Name inlinedName = inlinedFunction->name;
+        if (!canInlineInto(func->name, inlinedName)) {
+          continue;
+        }
 #ifdef INLINING_DEBUG
         std::cout << "inline " << inlinedName << " into " << func->name << '\n';
 #endif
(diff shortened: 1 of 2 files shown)


src/passes/Inlining.cpp
```diff
@@ -446,6 +449,20 @@ struct Inlining : public Pass {
...
+  bool canInlineInto(Name target, Name source) {
+    // We wish to avoid certain extremely-large sizes after inlining, as they
+    // may hit limits in VMs and/or slow down startup (measurements there
+    // indicate something like ~1 second to optimize a 100K function). See e.g.
+    // https://github.com/WebAssembly/binaryen/pull/3730#issuecomment-867939138
+    // https://github.com/emscripten-core/emscripten/issues/13899#issuecomment-825073344
+    auto combinedSize = infos[target].size + infos[source].size;
+    // Estimate the binary size from the number of instructions.
+    auto estimatedBinarySize = Measurer::BytesPerExpr * combinedSize;
+    const Index MaxCombinedBinarySize = 400 * 1024;
```
aheejin, line 463: Is this an arbitrary limit? Maybe good to add a single-line comment on what this is. If this is an arbitrary limit, stating that will help readers too, in case they think this number has a specific meaning.

aheejin (approved):

src/passes/Inlining.cpp
```diff
@@ -446,6 +449,20 @@ struct Inlining : public Pass {
     // return whether we did any work
     return inlinedUses.size() > 0;
   }
+
+  // Checks if we are allowed to inline source into target.
+  bool canInlineInto(Name target, Name source) {
```
aheejin, line 454: Nit: How about `isProfitableToInline` or something..? It is not that technically we cannot inline but we choose not to because of performance reasons..

tlively: Maybe `isUnderSizeLimit` to be precise?

kripken:
Thanks for the feedback, done.

PR #3961, opened 2021-07-01
Add option to add checks for all items

tlively:
Add an --all-items flag to update_lit_checks.py to emit checks for all module
items, not just those that match items in the input. Update two tests to use
generated input with the new flag.

Also, to improve readability, insert an empty line between consecutive checks
for different items.


kripken (approved):
Nice!

PR #3962, opened 2021-07-02
Support generating checks for multiple modules

tlively:
In conjunction with the `foreach` tool, allows autogenerating checks for lit
tests containing multiple modules. Supporting this will help automatically port
existing bespoke wast tests to be lit tests, since many of those tests contain
multiple modules per file.


tlively:
@kripken, If you'd like, I can try to split this into separate commits (or PRs) to separate the NFC refactoring and code movement and the functional part. Let me know if you think that's worth it.

aheejin:
Yeah it'd be nice to split this so seeing diff will be easier.

tlively:
The NFC components of this change are in https://github.com/WebAssembly/binaryen/pull/3964.

tlively:
Here's the diff from #3964 without whitespace changes: https://github.com/WebAssembly/binaryen/pull/3962/commits/b24fadbfe28b5ab8dcb28d8187d65e060371faca?w=1

tlively:
Rebased so this can be reviewed normally now. I recommend viewing the diff without whitespace changes.

test/lit/passes/extract-function.wast
```diff
@@ -1,71 +1,84 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
 ;; RUN: foreach %s %t wasm-opt --extract-function=foo -S -o - | filecheck %s
 ;; RUN: foreach %s %t wasm-opt --extract-function --pass-arg=extract-function@foo -S -o - | filecheck %s
 ;; RUN: foreach %s %t wasm-opt --extract-function-index=0 -S -o - | filecheck %s
 ;; RUN: foreach %s %t wasm-opt --extract-function-index --pass-arg=extract-function-index@0 -S -o - | filecheck %s
 
-;; CHECK:      (module
```
aheejin, line 6: Were we already supporting multiple modules within a file in lit tests? I thought that splitting and splicing was supported in our old tests but not here, but I might have missed a few PRs

aheejin (commented):
Just started looking at this but will be OOO for an hour or two for some errands so only one question..

test/lit/passes/extract-function.wast
tlively: Yes, but this is the only test that uses it so far. This test was introduced along with the `foreach` tool in https://github.com/WebAssembly/binaryen/pull/3939.

kripken (approved):

PR #3965, opened 2021-07-02
Lit tests for tool help messages

tlively:
Add list tests for the help messages of all tools, factoring out common options
into shared tests. This is slightly brittle because the text wrapping depends on
the length of the longest option, but that brittleness should be worth the
benefit of being able to see the actual help text in the tests.


aheejin (approved):
Nice!

kripken (approved):

PR #3963, opened 2021-07-02
Add a script for porting passes tests to lit

tlively:
And use it to port the very simple untee test.

PR #3966, opened 2021-07-03
Port most passes tests to lit

tlively:
Using the script from #3963. Be careful to exclude metrics and fuzz-exec tests,
which have output that is not captured by the check generator script.


aheejin:
image
😮🤯🤩

kripken:
It's hard to review this as the `/files/` github page is extremely slow. How about splitting this up into smaller pieces? That would be easier in other ways as well.

tlively:
Sure, splitting this up makes sense!

tlively:
I'll close this in favor of the split-up approach.

PR #3968, opened 2021-07-07
Filter more passes tests in porting script

tlively:
Filter out tests using --metrics, --emit-{js,spec}-wrapper, --fuzz-exec,
--print-stack-ir, --dwarfdump, etc. because they have ouput that will not be
captured in the automatically generated checks. Also filter based on the list of
passes rather than the test file in case the test uses a .passes file.

### DIFF
--- a/scripts/port_passes_tests_to_lit.py
+++ b/scripts/port_passes_tests_to_lit.py
@@ -40,14 +40,6 @@ def port_test(args, test):
         warn('Skipping because only .wast files are supported')
         return
 
-    if 'translate-to-fuzz' in test or 'dwarf' in test:
-        warn('Skipping due to Windows issues')
-        return
-
-    if 'noprint' in test:
-        warn('Skipping due to nonstandard output')
-        return
-
     dest = os.path.join(test_dir, 'lit', 'passes', name)
     if not args.force and os.path.exists(dest):
         warn('Skipping because destination file already exist')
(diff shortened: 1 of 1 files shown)


scripts/port_passes_tests_to_lit.py
```diff
@@ -59,6 +51,15 @@ def port_test(args, test):
         with open(passes_file) as f:
             joined_passes = f.read().strip()
 
+    if 'translate-to-fuzz' in joined_passes or 'dwarf' in joined_passes:
+        warn('Skipping due to Windows issues')
+        return
+
+    for bad in ['noprint', 'metrics', 'fuzz', 'print', 'emit', 'dump']:
```
kripken, line 58:
```suggestion
    for bad in ['noprint', 'metrics', 'fuzz-exec', '-fuzz-exec-before', 'print', 'emit', 'dump']:
```

kripken (commented):
Capturing `stdout` output would allow handling at least`metrics` and `fuzz-exec`, I believe, if we want that eventually.

scripts/port_passes_tests_to_lit.py
tlively: This is just a substring match on the string containing all the passes, so these are already filtered out by `fuzz`.

tlively:
> Capturing `stdout` output would allow handling at least`metrics` and `fuzz-exec`, I believe, if we want that eventually.

Yeah, I think this would make sense as a flag to `update_lit_checks.py` and possibly `port_passes_tests_to_lit.py` once the "easy" cases are done.

scripts/port_passes_tests_to_lit.py
kripken: Oh right, `joined_passes` is a string...

kripken (approved):

PR #3969, opened 2021-07-07
Port test/passes/O* to lit

tlively:


kripken (commented):
There are two tests named `O`, `O.wast` and `O.wasm` which are different. The latter appears to be deleted but not ported but maybe I am misreading this?

tlively:
Good catch! Yes, were skipping all binary tests, but then deleting everything that shared a base name with ported text tests. This is fixed now.

kripken (approved):

PR #3971, opened 2021-07-07
Port test/passes/flatten* to lit

tlively:


aheejin:
Maybe it's too late for this at this point given that other PRs have been merged, but I wonder moving all files to test/lit/passes directory in one PR and modifying them in place in another PR would allow us not losing the "moving" history of files.. Currently git recognize some files as moves but others as just new files.

tlively:
I think it would be a little strange to have tests in the `lit` directory that are not yet lit tests. Also, I don't know of a good way to have `lit` run some of the .wast files in a directory and have the custom check script run the other .wast files in the same directory. Is git detecting moved files valuable? Since the files keep their names, it should be easy to manually track files back through these commits.

aheejin:
> I think it would be a little strange to have tests in the `lit` directory that are not yet lit tests. Also, I don't know of a good way to have `lit` run some of the .wast files in a directory and have the custom check script run the other .wast files in the same directory.

If we are to go this way, I guess we have to ignore the CI for those PRs.

> Is git detecting moved files valuable? Since the files keep their names, it should be easy to manually track files back through these commits.

It's generally better to preserve git history, that's all. You can use `git log --follow` to follow the history of the file including moves.

kripken:
(Sorry, I wrote in the other PR first. As I said there, I personally don't see a need to do extra work for `--follow` to work, but if it would be useful to anyone else on the team then :+1: from me.)

aheejin (approved):
If both of you find it not very useful, then maybe it is not.. I'm OK either way.

tlively:
Thanks, @aheejin! If it were simpler to do the two phase approach I think it might be worth it, but given the extra complexity I don't think it's worth it, so I'll land these PRs without it.

PR #3974, opened 2021-07-08
Do not execute lit tests via an external shell

tlively:
We were previously configuring lit to execute test scripts via an external
shell, which caused lit to look for a bash installation on Windows. To make lit
tests work more "out of the box" on Windows, change the configuration to use
lit's built in shell capabilities instead.


aardappel (approved):
verified that this fixes lit tests on Windows

PR #3976, opened 2021-07-09
[Wasm GC] Add Wasm GC support to InstrumentMemory

kripken:
This adds calls to imports around every struct load and store, to
note their values, and also to arrays (where it also notes the
index).

This has been very useful in debugging LowerGC (lowering of Wasm
GC to wasm MVP).


src/passes/InstrumentMemory.cpp
```diff
@@ -138,20 +161,132 @@ struct InstrumentMemory : public WalkerPass<PostWalker<InstrumentMemory>> {
...
+      target = struct_get_val_i64;
+    } else if (curr->type == Type::f32) {
+      target = struct_get_val_f32;
+    } else if (curr->type == Type::f64) {
+      target = struct_get_val_f64;
+    } else {
+      return; // TODO: other types, unreachable, etc.
+    }
+    replaceCurrent(builder.makeCall(
+      target, {builder.makeConst(int32_t(id++)), curr}, curr->type));
```
aheejin, line 179: In `load_TYPE`/`struct_get_val_TYPE`/`array_get_val_TYPE` methods, the `value` parameter seems to be the address of the expression itself (`Load`/`StructGet`/`ArrayGet`). But In `store_TYPE`/`struct_set_TYPE`/`array_set_TYPE` methods, the `value` parameter corresponds to the value to store and there is no parameter for the expression itself (`Store`/`StructSet`/`ArraySet`).

- What is the use of logging the address of expression themselves?
- Why are they logged only for `load`/`struct_get`/`array_get`s but not for `store`/`struct_set`/`array_set`s?

src/passes/InstrumentMemory.cpp
```diff
@@ -70,6 +75,24 @@ static Name store_val_i32("store_val_i32");
...
+static Name array_get_val_i32("array_get_val_i32");
+static Name array_get_val_i64("array_get_val_i64");
+static Name array_get_val_f32("array_get_val_f32");
+static Name array_get_val_f64("array_get_val_f64");
+static Name array_set_val_i32("array_set_val_i32");
+static Name array_set_val_i64("array_set_val_i64");
+static Name array_set_val_f32("array_set_val_f32");
+static Name array_set_val_f64("array_set_val_f64");
+static Name array_get_index("array_get_index");
+static Name array_set_index("array_set_index");
```
aheejin, line 95: Why do we not have `get/set_index` for structs?

scripts/wasm2js.js
```diff
@@ -191,6 +191,80 @@ var asmLibraryArg = {
...
+  struct_get_val_i32: function(loc, value) {
+    console.log('struct_get_val_i32 ' + [loc, value]);
+    return value;
+  },
+  struct_get_val_i64: function(loc, value) {
+    console.log('struct_get_val_i64 ' + [loc, value]);
+    return value;
+  },
+  struct_get_val_f32: function(loc, value) {
+    console.log('loastruct_get_val_i32d_ptr ' + [loc, value]);
```
tlively, line 204: All these `i32d_ptr` should be `f32`, right?

src/passes/InstrumentMemory.cpp
kripken: > What is the use of logging the address of expression themselves?

Do you mean the address of the Binaryen IR instruction? I don't think that is logged. We pass `curr` because we have the instruction as a child of the new one.

Overall, in both load and store, the value is the value at that address (in the case of store, after the store). We also have the address, but that is separate. A load turns into this (from the existing test):
```wat
   (call $load_val_i32
    (i32.const 1)
    (i32.load8_s
     (call $load_ptr
      (i32.const 1)
      (i32.const 1)
      (i32.const 0)
      (i32.const 0)
     )
    )
   )
```
We call `load_ptr` on the pointer, to log it. Then we do the load (using a normal wasm load). Then we call `load_val` on the loaded value, to log that.

Store has the same: a call to `store_ptr` and `store_val` for the pointer and value.

(It may be easier to read the existing test output than the code.)

src/passes/InstrumentMemory.cpp
kripken: The struct index is constant in the wasm, so it seemed less useful - it's visible when reading the code.

scripts/wasm2js.js
kripken: Heh, good catch... looks like a very weird copy-paste error. Fixed.

tlively (approved):

src/passes/InstrumentMemory.cpp
aheejin: Ah I see, I was confused. Thanks.

aheejin (approved):

PR #3977, opened 2021-07-10
Improve NameTypes to keep reasonable names

kripken:
We only set a name now if there was no name, or the existing name
was really really long.


test/lit/passes/name-types.wast
```diff
@@ -1,12 +1,24 @@
-;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
 ;; RUN: wasm-opt %s -all --name-types -S -o - | filecheck %s
 
 (module
+ ;; An obnoxious name that will be renamed.
  (type $obnoxious-super-long-type-name_____________________________1 (struct))
- ;; CHECK:      (func $foo (param $x (ref $type$1))
+
```
kripken, line 7: @tlively the position of the comment right below this is odd. looks like the auto update script did not keep the comment belonging to the type attached to it?

aheejin (approved):

test/lit/passes/name-types.wast
tlively: For each matching item in the input and output, `--all-items` causes all the output items since the last match to be emitted as well. The update script does not know about comments, so it does not know to emit the comment after those extra items. One solution would be to teach the update script about comments and treat them specially. Another solution for this particular test would be to include globals or functions that reference the types and get the type names in the output that way rather than using `--all-items`.

tlively (approved):

test/lit/passes/name-types.wast
kripken: I see, thanks. I guess this is good enough for now. Not sure it's worth working around the issue in the test.

PR #3980, opened 2021-07-12
Fix features section handling in the fuzzer

kripken:
The features section is additive since #3960. For the fuzzer to know which features
are used, it therefore needs to also scan the features section. To do this,
run `--print-features` to get the total features used from both flags + the
features section.

A result of this is that we now have a list of enabled features instead of
"enable all, then disable". This is actually clearer I think, but it does require
inverting the logic in some places.

May help with #3979 , although I haven't checked yet (cc @MaxGraey ).

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -123,9 +123,8 @@ ALL_FEATURE_OPTS = ['--all-features', '-all', '--mvp-features', '-mvp']
 def update_feature_opts(wasm):
     global FEATURE_OPTS
     # we will re-compute the features; leave all other things as they are
-    EXTRA = [x for x in FEATURE_OPTS if not x.startswith('--enable') and \
-                                        not x.startswith('--disable') and \
-                                        not x in ALL_FEATURE_OPTS]
+    EXTRA = [x for x in FEATURE_OPTS if not x.startswith('--enable') and
+             not x.startswith('--disable') and x not in ALL_FEATURE_OPTS]
     FEATURE_OPTS = run([in_bin('wasm-opt'), wasm] + FEATURE_OPTS + ['--print-features']).strip().split('\n')
     # filter out '', which can happen if no features are enabled
     FEATURE_OPTS = [x for x in FEATURE_OPTS if x]
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -464,7 +464,7 @@ def run_d8_wasm(wasm, liftoff=True):
 
 
 def all_disallowed(features):
-    return not any([('--enable-' + x) in FEATURE_OPTS for x in features])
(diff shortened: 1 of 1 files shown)


scripts/fuzz_opt.py
```diff
@@ -117,6 +117,17 @@ def randomize_feature_opts():
     print('randomized feature opts:', ' '.join(FEATURE_OPTS))
 
 
+def update_feature_opts(wasm):
+    global FEATURE_OPTS
+    # we will re-compute the features; leave all other things as they are
+    EXTRA = [x for x in FEATURE_OPTS if not x.startswith('--enable') and not x.startswith('--disable') and x != '--all-features']
```
tlively, line 123: What is `EXTRA` for? Why is `--all-features` excluded but not `-all` or `--mvp-features` or `-mvp`?

scripts/fuzz_opt.py
```diff
@@ -563,7 +574,7 @@ def can_run(self, wasm):
                 if random.random() < 0.5:
                     return False
                 # wasm2c doesn't support most features
-                return all([x in FEATURE_OPTS for x in ['--disable-exception-handling', '--disable-simd', '--disable-threads', '--disable-bulk-memory', '--disable-nontrapping-float-to-int', '--disable-tail-call', '--disable-sign-ext', '--disable-reference-types', '--disable-multivalue', '--disable-gc']])
+                return not any([x in FEATURE_OPTS for x in ['--enable-exception-handling', '--enable-simd', '--enable-threads', '--enable-bulk-memory', '--enable-nontrapping-float-to-int', '--enable-tail-call', '--enable-sign-ext', '--enable-reference-types', '--enable-multivalue', '--enable-gc']])
```
tlively, line 577: It might be good to factor this pattern out into a helper function that takes a list of feature names.

scripts/fuzz_opt.py
kripken: Ah, I just handled the actually used ones. But good point, I generalized it now.

scripts/fuzz_opt.py
kripken: Done.

scripts/fuzz_opt.py
```diff
@@ -448,6 +463,10 @@ def run_d8_wasm(wasm, liftoff=True):
     return run_d8_js(in_binaryen('scripts', 'fuzz_shell.js'), [wasm], liftoff=liftoff)
 
 
+def all_disallowed(features):
+    return not any([('--enable-' + x) in FEATURE_OPTS for x in features])
```
tlively, line 467:
```suggestion
    return not any(('--enable-' + x) in FEATURE_OPTS for x in features)
```

I think this does the same thing without materializing the list (and is slightly easier to read).

tlively (approved):

Issue #3627, opened 2021-02-27
Printing indices instead of generated heap type names?


tlively:
As we move on from working exclusively with tiny GC test programs, I'm not sure the code for generating heap type names is pulling its own weight in terms of the value it brings against its complexity. For real programs, the heap types are so large and complex that the generated names fail to be useful for basic things like identifying at a glance whether two places use the same or different heap types. Printing simple indices into the type section, while providing less local information about type structure, would make for better readability overall IMO. This would also eliminate current bugs with type name clashes.

Would anyone object to me removing the name generation code?

tlively:
cc @kripken

tlively:
Thinking about this more, one problematic case would be when heap types need to be printed as part of individual expressions without the context of a full module. In that situation, printing type indices would be meaningless. I propose that we print the full type inline in those cases rather than trying to give it a name. Separately, we should add a mechanism for globally configuring the default print depth of expressions and types to avoid information overload in that and other situations. I got a an error message that printed a type with 1.8MB of text the other day and it was not too useful 😅

kripken:
I agree with the above. When printing the full module, we may as well index them. When there is no module (printing just an expression) then printing in the old way makes sense.

If it's not hard, then having both an index and a shortened full name might be nice, e.g.
```wat
(type $type42$ref?|i31|) ;; full text name
(type $type1337$ref?|ref...) ;; shortened when it would be too long
```

tlively:
I'm proposing that we not give types names at all (besides names present in the input, of course), so there wouldn't be a need for short or long names.

kripken:
Are you opposed to also adding a short name after the index? If so, why?

(Seems like the name could help readability in some cases at least. And having the index up front in the name would avoid current name issues.)

tlively:
I'm just not sure a short name with structural rather than semantic information would ever be useful for real world programs because there will be a very large number of types that all look very similar from a 10,000 foot view. It will simplify our code a lot if we can just print `123` instead of a name, even a name like `$123`, because then we won't need to worry about name clashes with user-defined names at all.

kripken:
Fair enough. I'm thinking not just of large real-world programs but also smaller things, like maybe a reduced testcase from one of them. But we can consider adding more than just indexes later if it seems useful, I agree your direction is good.

tlively:
This is done. There are still some issues about printing meaningless type indices in error messages, but those can be addressed separately.

PR #3986, opened 2021-07-14
Fuzzer: Use NameTypes in RoundTrip

kripken:
This works around the issue with wasm gc types sometimes getting
truncated (as the default names can be very long or even infinitely
recursive). If the truncation leads to name collision, the wast is not
valid.

Should we run name-types during load of a file using wasm GC? I
think you had some other idea here @tlively but I don't remember
what.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -886,7 +886,11 @@ class RoundtripText(TestCaseHandler):
     frequency = 0.05
 
     def handle(self, wasm):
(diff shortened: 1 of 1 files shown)


tlively:
Yes, I think running name-types during loading (maybe only with `-g`?) makes sense. I think the net result of that would be roughly the same as what I was thinking of doing in https://github.com/WebAssembly/binaryen/issues/3627.

tlively (approved):

PR #3987, opened 2021-07-14
Port test/passes/a* to lit

tlively:


tlively:
This is the first of 23 PRs to finish porting the passes tests in batches of no more than 10 tests. Since GitHub doesn't handle stacked PRs well, I'll submit them for review one at a time, but they're all numbered and present at https://github.com/tlively/binaryen/branches/all.

tlively:
Actually, the easiest way to see all the individual commits is by looking at the last commit: https://github.com/WebAssembly/binaryen/compare/main...tlively:port-tests-22-?expand=1

tlively:
Also, I have a local script to generate these commits, so it would be trivial to adjust the maximum batch size for whatever is easiest to review. Just let me know :)

kripken (approved):

kripken:
If 23 sequential PRs feels annoying to you (which I would totally understand!) I'm not opposed to a single big one. That one would just take longer to review, but at least it'd be done with...

tlively:
No, I think you're right that an incremental approach is better. Especially because getting these changes in is not urgent and some of the tests might be updated in the meantime.

PR #4002, opened 2021-07-17
Revert "Fix Precompute on SIMD"

kripken:
Reverts WebAssembly/binaryen#3983

Instead of applying the workaround to avoid precomputing SIMD in more places,
which prevents things we could optimize before, we should probably focus on
making the workaround not needed - that is, implement full SIMD support in the
intepreter (the reason for that PR), and the other TODO in the comment here,

```
    // Until engines implement v128.const and we have SIMD-aware optimizations
    // that can break large v128.const instructions into smaller consts and
    // splats, do not try to precompute v128 expressions.
```


tlively (approved):
Seems reasonable

PR #4007, opened 2021-07-20
[Wasm GC] Avoid dangerous StackIR opts on GC

kripken:
`removeUnneededBlocks()` can force us to use a local when we
load the emitted wasm, which can't work for something nondefaultable
like an RTT.

For now, just don't run that optimization if GC is enabled. Eventually,
perhaps we'll want to enable this optimization in a different way.


src/passes/StackIR.cpp
```diff
@@ -65,7 +69,39 @@ class StackIROptimizer {
...
+    //       (nop)
+    //       (local.get)
+    //     )
+    //     (i32)
+    //   )
+    //
+    // However, that is not valid as an rtt cannot be set to a local.
+    if (!features.hasGC()) {
+      removeUnneededBlocks();
+    }
```
tlively, line 104: Would it be feasible to fix this issue on a case by case basis by looking at what type would need to be assigned to the local?

tlively (approved):
LGTM if it's not worth doing a more fine-grained mitigation.

src/passes/StackIR.cpp
kripken: I think it's possible, but it's kind of annoying, since it's the type of the local "behind" us... that is, the type right before the block that we actually remove.

aheejin:
Sorry for the basic GC questions:
- I guess RTT is not nullable, right?
- Then RTT cannot be assigned to a local, even if for a non-nullable local? It looks RTT can be passed as parameters though.

kripken:
RTTs are pretty weird...

The spec implemented by v8 and binaryen has RTTs as special values, and not references, so nullability is not relevant for them. But they have no default value, so they are like non-nullable types. They are the only non-defaultable type that is not nullable.

Binaryen can handle non-nullable references by adding `ref.as_non_null` for them. We could also do something similar for RTTs perhaps.

Meanwhile the GC spec changed and it proposes that RTTs are in fact references. Until that stabilizes, probably we shouldn't do much in binaryen for it...

aheejin:
> RTTs are pretty weird...
> 
> The spec implemented by v8 and binaryen has RTTs as special values, and not references, so nullability is not relevant for them. But they have no default value, so they are like non-nullable types, and cannot be in locals.

Even if non-nullable local is enabled?

> They are the only non-defaultable type that is not nullable.

Aren't non-nullable references also non-defaultable and not nullable?

> Binaryen can handle non-nullable references by adding `ref.as_non_null` for them. We could also do something similar for RTTs perhaps.

Does `ref.as_non_null` take RTTs?

> Meanwhile the GC spec changed and it proposes that RTTs are in fact references.

🙃

kripken:
> Even if non-nullable local is enabled?

Oh, yes, if that is enabled then we do allow them.

> Aren't non-nullable references also non-defaultable and not nullable?

Sorry, I meant, RTTs are non-defaultable and also have no concept of nullability. Thanks to nullability we can work around a non-nullable local, by adding `ref.as_non_null` as mentioned earlier (we "force" it to be non-nullable every place we write and read it).

> Does ref.as_non_null take RTTs?

No. We'd need to do something weird like create a struct that contains a single field, an RTT, and create such a struct just for this purpose (and fill in `rtt.canon` for the "default" value). So it could be done, but as RTTs may become references in the new spec, maybe we don't need to...

src/passes/StackIR.cpp
```diff
@@ -65,7 +69,39 @@ class StackIROptimizer {
...
+    //       (local.set
+    //         (rtt)
+    //       )
+    //       (nop)
+    //       (local.get)
+    //     )
+    //     (i32)
+    //   )
+    //
+    // However, that is not valid as an rtt cannot be set to a local.
```
aheejin, line 101: Even without GC, is it beneficial to remove blocks that have stacky code, or multiple return values anyway, if we have to introduce `local.set` and `local.get`?

kripken: It is harmful if we roundtrip, yes. But it does emit a slightly smaller wasm in the middle, which can be shipped. It is a little awkward that binaryen has different behavior on stacky code than VMs. I wonder if we can fix that somehow, but I doubt it.

PR #4008, opened 2021-07-20
Port test/passes/d* to lit

tlively:


test/lit/passes/dce_all-features.wast
```diff
@@ -1,3 +1,7 @@
+;; NOTE: This test was ported using port_test.py and could be cleaned up.
+
+;; RUN: foreach %s %t wasm-opt --dce --all-features -S -o - | filecheck %s
+
 (module
   (memory 10)
   (type $ii (func (param i32 i32)))
```
kripken, line 7: Where is the output for this file? E.g. the first function `$call-me` has no `CHECK`s, nor the second.

tlively: Not sure why the checks were missing for this file. I just regenerated all these tests now and it looks fine.

test/lit/passes/duplicate-function-elimination_all-features.wast
```diff
@@ -0,0 +1,80 @@
...
+
+ ;; CHECK:      (func $foo (result i32)
+ ;; CHECK-NEXT:  (unreachable)
+ ;; CHECK-NEXT: )
+ (func $foo (result i32)
+  (unreachable)
+ )
+ (func $bar (result i32)
+  (unreachable)
+ )
```
kripken, line 69: The function `$bar` is removed by this optimization, but the CHECKs don't make that obvious. Even just the fact that it has no CHECKs isn't quite enough, as the function after it, "export", also has no CHECKs - its CHECKs actually appear after the module.

Could we emit a CHECK that says a function no longer exist?

Separately, adding a `$` name for the final function, and not just an export name, would help things like this.

tlively:
Ok I removed all duplicate elimination tests from this to sidestep the issues with them for now. Once all the straightforward tests are ported we can budget some time to do a good job of porting the rest.

kripken (approved):

PR #4019, opened 2021-07-23
[Wasm GC] Handle nondefaultable types in LocalSubtyping

kripken:
The pass handled non-nullability, but another case is a tuple with nullable
values in it that is assigned non-nullable values, and in general, other stuff
that is nondefaultable (but not non-nullable). Ignore those.

### DIFF
--- a/test/lit/passes/local-subtyping.wast
+++ b/test/lit/passes/local-subtyping.wast
@@ -214,7 +214,8 @@
   ;; CHECK-NEXT: )
   (func $nondefaultable
     (local $x (anyref anyref))
-    ;; A tuple
+    ;; This tuple is assigned non-nullable values, which means the subtype is
+    ;; nondefaultable, and we must not apply it.
     (local.set $x
       (tuple.make
         (ref.func $i32)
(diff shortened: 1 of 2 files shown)


test/lit/passes/local-subtyping.wast
```diff
@@ -202,4 +202,25 @@
...
+  ;; CHECK-NEXT:  (local $x (anyref anyref))
+  ;; CHECK-NEXT:  (local.set $x
+  ;; CHECK-NEXT:   (tuple.make
+  ;; CHECK-NEXT:    (ref.func $i32)
+  ;; CHECK-NEXT:    (ref.func $i32)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $nondefaultable
+    (local $x (anyref anyref))
```
tlively, line 216: It's a little unfortunate that we can't at least make this `(funcref funcref)`, which is more refined than `(anyref anyref)` but is still defaultable. Also, it's too bad that the non-nullable local support doesn't handle tuples, but I guess our tuple lowering still needs to be better for that to make any sense.

tlively (approved):

PR #4024, opened 2021-07-26
[Wasm GC] Handle uses of default values in LocalSubtyping

kripken:
If the default value is used, it is better to not change the local type. See
details in the comment.

### DIFF
--- a/src/passes/LocalSubtyping.cpp
+++ b/src/passes/LocalSubtyping.cpp
@@ -44,12 +44,9 @@ struct LocalSubtyping : public WalkerPass<PostWalker<LocalSubtyping>> {
     auto numLocals = func->getNumLocals();
 
     // Compute the local graph. We need to get the list of gets and sets for
-    // each local, so that we can do the analysis, and also we need to know
-    // when the default value of a local is used. If the default is actually
-    // used then we cannot change that type, as then we might end up with a null
-    // of a different type - while all nulls compare equally, it can be
-    // confusing (and errors in the fuzzer). Furthermore, with non-nullable
-    // locals we can get invalid code where if we change the local type to
+    // each local, so that we can do the analysis. For non-nullable locals, we
+    // also need to know when the default value of a local is used: if so then
+    // we cannot change that type, as if we change the local type to
     // non-nullable then we'd be accessing the default, which is not allowed.
     //
     // TODO: Optimize this, as LocalGraph computes more than we need, and on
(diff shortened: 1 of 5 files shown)


test/lit/passes/local-subtyping.wast
```diff
@@ -223,4 +223,30 @@
...
+    (local $x anyref)
+    (if
+      (local.get $i)
+      ;; The only set to this local uses a more specific type than anyref.
+      (local.set $x (ref.func $uses-default))
+    )
+    (drop
+      ;; This get may use the default value, and so we should not alter the
+      ;; type of the local.
+      (local.get $x)
```
tlively, line 249: We could still make the type more specific and leave it nullable, couldn't we? I would think that would be worth it.

src/passes/LocalSubtyping.cpp
```diff
@@ -35,29 +36,56 @@ struct LocalSubtyping : public WalkerPass<PostWalker<LocalSubtyping>> {
...
+    std::unordered_set<Index> usesDefault;
+
+    for (auto& kv : localGraph.getSetses) {
+      auto* get = kv.first;
+      auto& sets = kv.second;
+      auto index = get->index;
+      if (func->isVar(index) &&
+          std::any_of(sets.begin(), sets.end(), [&](LocalSet* set) {
+            return set == nullptr;
+          })) {
```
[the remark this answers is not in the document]
kripken: Yes, the "ssa" pass will separate disjoint uses of locals like that.

test/lit/passes/local-subtyping.wast
kripken: We could, yeah. It would require some fuzzer changes - as mentioned in the comment, to ignore type differences between nulls. I slightly worry that making that change might make us miss other bugs. But I'm not sure either way.

kripken: I pushed a change to allow default values where possible, making us only avoid making a non-nullable local where the default value is used, to show what that would look like.

src/passes/LocalSubtyping.cpp
```diff
@@ -35,29 +36,60 @@ struct LocalSubtyping : public WalkerPass<PostWalker<LocalSubtyping>> {
...
+        getsForLocal[get->index].push_back(get);
+      }
+    }
+
+    // Find which vars use the default value, if we allow non-nullable locals.
+    //
+    // If that feature is not enabled, then we can safely assume that the
+    // default is never used - the default would be a null value, and the type
+    // of the null does not really matter as all nulls compare equally, so we do
+    // not need to worry.
```
tlively, line 75: It's not like we're replacing it with some random other null, either, right? The new default null will always be a subtype of the original default null. This would even work in a hypothetical future where you could specify some other non-null default value for locals.

src/tools/execution-results.h
```diff
@@ -144,7 +144,9 @@ struct ExecutionResults {
   }
 
   bool areEqual(Literal a, Literal b) {
-    if (a.type != b.type) {
+    // We allow nulls to have different types (as they compare equal regardless)
+    // but anything else must have an identical type.
```
tlively, line 148: Should we be checking subtype relations rather than equality? For nulls specifically and for users of this function more generally.

tlively (commented):
This seems like a useful change to me.

src/tools/execution-results.h
kripken: Subtype checks would handle nulls properly, but on other values they might not give us what we want.

If the type values are references, we only compare their types (see comment below this code), and if we allow subtyping, we might have one value of `{}` and the other of `{i32}`, say - subtypes, but observably different, and the fuzzer should error.

tlively (approved):

PR #4025, opened 2021-07-26
[Wasm GC] DeadArgumentElimination: Do not refine return types of tail callees

kripken:
If there is a tail call, we can't change the return type of the function, as it
must match in the functions doing a tail call of it.


tlively (approved):

tlively:
I think this should be reverted after https://github.com/WebAssembly/binaryen/pull/4032. Sorry I didn't think about this sooner! Happy to chat more about this tomorrow, too.

kripken:
No worries, sounds good. After that lands I'll revert the code and update the tests here. And I'll convert the other PR to just a test.

PR #4029, opened 2021-07-27
[Wasm GC] DeadArgumentElimination: Do not refine return types of a function that tail calls

kripken:
Similar to #4025, but when there is a tail call in the function as opposed to
the function is tail-called from somewhere.

### DIFF
--- a/src/passes/DeadArgumentElimination.cpp
+++ b/src/passes/DeadArgumentElimination.cpp
@@ -663,6 +663,26 @@ private:
     }
     assert(refinedType != originalType);
 
+    // The body is unreachable, and there are no returns, so this function
+    // never exits, and there is nothing to optimize.
+    if (refinedType == Type::unreachable) {
+      return false;
+    }
+
+    // The returns allow us to refine the type. However, if there are tail calls
+    // in this function, they limit our ability to change the return type (as it
+    // must exactly match the type returned from the tail calls).
+    for (auto* call : FindAll<Call>(func->body).list) {
+      if (call->isReturn) {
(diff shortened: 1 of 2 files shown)


tlively:
Doesn't the tail call count as a return and get factored into the LUB, though?

kripken:
It doesn't, as we just look at `return`s and not `return_call*`s. And I believe we should not look at those things, as their presence requires us to return a particular type exactly and nothing more refined (unlike the usual subtyping rules).

tlively:
Do return_calls not obey subtyping? It seems like they should.

kripken:
The validator currently requires precise matching of the type. The spec appears to do the same,

https://github.com/WebAssembly/tail-call/blob/master/proposals/tail-call/Overview.md#validation

That does not surprise me, as it seems like it makes the VM's life a lot simpler when doing a tail to know that the types are identical - it can use the same stack frame without thinking about it. But I'm not an expert on that.

tlively:
The tail call proposal spec is not based on any of the other proposals that include subtyping. The point of subtyping is that subtypes can be used wherever supertypes are expected with no coersions, so engines wouldn't have to do anything to make this work safely. I'm working on a PR to update our validator now.

tlively:
Ok, after https://github.com/WebAssembly/binaryen/pull/4032 I believe the return_calls can be handled the same as normal returns. I would be happy to talk about this more tomorrow.

kripken:
I'll open a new PR for this after https://github.com/WebAssembly/binaryen/pull/4035 lands.

PR #4030, opened 2021-07-27
Parse and print pops of compound types

tlively:
This is necessary when using GC and EH together, for instance.


src/wasm/wasm-s-parser.cpp
```diff
@@ -2159,7 +2159,11 @@ Expression* SExpressionWasmBuilder::makePop(Element& s) {
   auto ret = allocator.alloc<Pop>();
   std::vector<Type> types;
   for (size_t i = 1; i < s.size(); ++i) {
-    types.push_back(stringToType(s[i]->str()));
+    if (s[i]->isStr()) {
+      types.push_back(stringToType(s[i]->str()));
+    } else {
+      types.push_back(elementToType(*s[i]));
+    }
```
aheejin, line 2166: Doesn't `elementToType` call `stringToType` within itself? https://github.com/WebAssembly/binaryen/blob/b3375bc19f67838d80eaa9af011952bfdaae243f/src/wasm/wasm-s-parser.cpp#L1167-L1170

aheejin (approved):
Thank you!

kripken (approved):
lgtm with @aheejin 's suggestion.

PR #4031, opened 2021-07-27
[Wasm GC] DeadArgumentElimination: Update tees after refining param types

kripken:


src/passes/LocalSubtyping.cpp
```diff
@@ -134,8 +134,9 @@ struct LocalSubtyping : public WalkerPass<PostWalker<LocalSubtyping>> {
           // NB: These tee updates will not be needed if the type of tees
           //     becomes that of their value, in the spec.
           for (auto* set : setsForLocal[i]) {
-            if (set->isTee() && set->type != Type::unreachable) {
```
tlively, line 137: Why this change?

tlively (approved):

src/passes/LocalSubtyping.cpp
kripken: Just for consistency with the similar code fragment in this PR.

It's cleaner to not think about unreachability in each place like this. Instead, calling the finalize method will handle unreachability and anything else relevant in a transparent way (although, there is nothing else relevant in this case, in the current IR).

PR #4032, opened 2021-07-28
Support subtyping in tail calls

tlively:
The tail call spec does not include subtyping because it is based on the
upstream spec, which does not contain subtyping. However, there is no reason
that subtyping shouldn't apply to tail calls like it does for any other call or
return. Update the validator to allow subtyping and avoid a related null pointer
dereference while we're at it.

Do not run the test in with --nominal because it is buggy in that mode.


test/lit/tail-call.wast
```diff
@@ -0,0 +1,78 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
+
+;; Check that tail calls are parsed, validated, and printed correclty
```
kripken, line 3:
```suggestion
;; Check that tail calls are parsed, validated, and printed correctly
```
(or was it an intentional joke...?)

kripken (approved):

kripken:
cc @fgmccabe - I'm curious to hear your thoughts and experience on this. My intuition is that it's scary to support subtyping in tail calls, as only when the types are identical then is it really obvious that it's fine to "just" reuse the stack frame. My sense is that this type of thing could be a source of exploits. But I don't have much VM experience here - is my intuition wrong?

test/lit/tail-call.wast
tlively: Haha no this was an unintentional joke.

fgmccabe:
Thomas is correct. The tail-call is actually return-call which gives an important clue here. When checking types, the return type of the callee should be compared against the return type of the current function. The typing of the arguments to return-call should be the same as for any call.

kripken:
Great, thanks @fgmccabe !

PR #4036, opened 2021-07-28
[Wasm GC] Refine return types of tail calling functions in DeadArgumentElimination

kripken:
To do this we need to look at tail calls and not just returns.


test/lit/passes/dae-gc.wast
```diff
@@ -594,4 +600,104 @@
...
+ ;; CHECK-NEXT:  (return_call $tail-callee)
+ ;; CHECK-NEXT: )
+ (func $tail-caller-no (result anyref)
+  ;; This function's return type cannot be refined because of another return
+  ;; whose type prevents it.
+  (if (i32.const 1)
+   (return (ref.null any))
+  )
+  (return_call $tail-callee)
+ )
```
tlively, line 636: For these *-no tests, it would be more interesting if the tail call type blocked the optimization, since we know from previous tests that normal returns can block the optimization.

tlively (approved):
Separately, it might be a good idea to move all the argument specialization tests into a separate file from the normal DAE tests to keep the test files smaller and more focused.

test/lit/passes/dae-gc.wast
kripken: Hmm, perhaps, but the test it is most directly compared to is the one right above it, so the smaller the diff between them the better I think.

I can add an additional flipped test, if you think that's worth it? As the logic for all returns is identical (they all call the same shared method) that feels a little unnecessary to me personally.

kripken:
Sounds good to move the tests to smaller files as a followup.

test/lit/passes/dae-gc.wast
tlively: Fair enough!

PR #4038, opened 2021-07-29
[Wasm GC] Handle call_ref in DeadArgument return refinement

kripken:
This should be the last call type...

If this pattern appears anywhere else (get all the return types) we should
have a helper for it, but it seems to only happen in this one pass.


tlively (approved):
Oops! I had completely forgotten that we already support `return_call_ref`. It is annoying that these expressions can't be both `Return` and `Call` expressions at the same time.

PR #4039, opened 2021-07-29
[Wasm GC] Optimize repeated identical ref.casts

kripken:


src/passes/OptimizeInstructions.cpp
```diff
@@ -1204,6 +1207,29 @@ struct OptimizeInstructions
...
+    while (!ref->is<RefCast>()) {
+      auto* last = ref;
+      ref = Properties::getImmediateFallthrough(ref, passOptions, features);
+      if (ref == last) {
+        break;
+      }
+    }
+    if (auto* child = ref->dynCast<RefCast>()) {
+      // Check if the casts are identical.
+      if (ExpressionAnalyzer::equal(curr->rtt, child->rtt) &&
```
tlively, line 1226: Would it be worth checking instead whether the current rtt is a sub-rtt of the child's rtt? Then we could skip the child cast and just keep the current cast. That optimization would be more general than this optimization where they have to be identical.

kripken: To find out if an RTT is a sub-RTT of another we need more than the static types, though. Even if two rtts have identical types, the cast may or may not fail, e.g. these two have identical type `(rtt 2 $C)` but they cannot be cast:
```wat
(rtt.sub $C (rtt.sub $B (rtt.canon $A)))
(rtt.sub $C (rtt.sub $B (rtt.canon $B)))
```

Unless I'm missing something, handling the general case doesn't seem practical. (And is one reason why RTTs make optimization hard at the wasm level IMO.)

tlively: Ugh right, that still always confuses me. This seems like a good reason to get rid of first-class RTTs for nominal types.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1204,6 +1207,29 @@ struct OptimizeInstructions
...
+    auto features = getModule()->features;
+
+    // Repeated identical ref.cast operations are unnecessary, if using the
+    // exact same rtt - the result will be the same. Find the immediate child
+    // cast, if there is one, and see if it is identical.
+    // TODO: Look even further through incompatible casts?
+    auto* ref = curr->ref;
+    while (!ref->is<RefCast>()) {
+      auto* last = ref;
+      ref = Properties::getImmediateFallthrough(ref, passOptions, features);
```
tlively, line 1219: Why `ImmediateFallthrough` and not just `Fallthrough`?

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -936,4 +935,126 @@
...
+  ;; CHECK-NEXT:   (ref.cast
+  ;; CHECK-NEXT:    (ref.cast
+  ;; CHECK-NEXT:     (local.get $x)
+  ;; CHECK-NEXT:     (rtt.canon $empty)
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:    (rtt.canon $struct)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $ref-cast-squared-no-different (param $x eqref)
```
tlively, line 1017: I would remove "no-" from these names; it makes it sound like the test will contain no difference or no effects, but actually you mean the opposite.

tlively (approved):

src/passes/OptimizeInstructions.cpp
kripken: Fallthrough will look through a cast! :smile: 

The meaning of fallthrough is that the value passes through, without being changed (unless we trap or branch). But here we want to ignore all other falling-through but via casting...

test/lit/passes/optimize-instructions-gc.wast
kripken: Good point, will update.

src/passes/OptimizeInstructions.cpp
tlively: Oh, that makes sense!

PR #4042, opened 2021-07-30
Properly print struct field names with --nominal

tlively:
The field name lookup was previously failing because the lookup used a fresh
struct type rather than using the intended struct type.


test/lit/nominal-named-field.wast
```diff
@@ -0,0 +1,31 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
+
+;; Regression test for a bug in which field names were not printed in nominal mode.
+
+;; RUN: wasm-opt %s -all --nominal -S -o - | filecheck %s
+;; RUN: wasm-opt %s -all --nominal --roundtrip -S -o - | filecheck %s
+
+(module
+
```
kripken, line 9: This line looks unnecessary (we don't have it in other tests).

kripken (commented):
lgtm % comment.

(this time I noticed automerge was on before clicking approve!)

kripken (approved):

PR #4043, opened 2021-07-31
Test GC lit tests with --nominal as well

tlively:
Add a new run line to every list test containing a struct type to run the test
again with nominal typing. In cases where tests do not use any struct subtyping,
this does not change the test output. In cases where struct subtyping is used, a
new check prefix is introduced to capture the difference that `(extends ...)`
clauses are emitted in nominal mode but not in equirecursive mode. There are no
other test differences.

Some tests are cleaned up along the way. Notably,
O_all-features{,-ignore-implicit-traps}.wast is consolidated to a single file.


src/ir/module-utils.h
```diff
@@ -569,8 +569,11 @@ inline void collectHeapTypes(Module& wasm,
     if (ht.getSuperType(super)) {
       if (!counts.count(super)) {
         newTypes.insert(super);
+        // We should unconditionally count supertypes, but while the type system
+        // is in flux, skip counting them to keep the type ordering in test
+        // outputs more similar.
```
kripken, line 574:
```suggestion
        // outputs more similar to structural typing. FIXME
```

kripken: (clever trick, btw!)

kripken:
> O_all-features{,-ignore-implicit-traps}.wast is consolidated to a single file.

I don't seem to see that file - maybe it wasn't `git add`ed?

src/ir/module-utils.h
tlively: Thanks!

tlively:
> > O_all-features{,-ignore-implicit-traps}.wast is consolidated to a single file.
> 
> I don't seem to see that file - maybe it wasn't `git add`ed?

Oops, that was indeed the problem. Thanks!

test/lit/passes/O_gc.wast
```diff
@@ -0,0 +1,43 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
+
+;; RUN: wasm-opt %s -O --all-features -S -o - | filecheck %s
+;; RUN: wasm-opt %s -O --all-features --nominal -S -o - | filecheck %s
+;; RUN: wasm-opt %s -O --all-features --ignore-implicit-traps -S -o - | filecheck %s --check-prefix=TRAPS
+;; RUN: wasm-opt %s -O --all-features --ignore-implicit-traps --nominal -S -o - | filecheck %s --check-prefix=TRAPS
```
kripken, line 6: How about IIT or IGNORE-TRAPS instead of TRAPS? TRAPS feels a little odd as it will *not* trap / assumes there are no traps. But I don't feel strongly, lgtm either way.

tlively: I was trying to match the number of characters in CHECK so that every lined up, but I agree that the language could be more clear. Will fix.

kripken: Oh, I see... how about `ITRAP`?

kripken (approved):

test/lit/passes/O_gc.wast
tlively: Looking at the new output, I actually like that the two output streams are differentiated by different indentation, so I'm happy the way it is.

PR #4045, opened 2021-08-02
Fix assert on access of empty vector

aardappel:
(on VS2019)
Triggered by:
wasm-emscripten-finalize --minimize-wasm-changes -g --bigint --no-dyncalls --dwarf test_asan_api.wasm -o test_asan_api.wasm --detect-features


kripken (approved):
Interesting, looks like this was fixed in upstream llvm since the code was copied to here (in the same way).

aardappel:
Ahh.. didn't even notice I was modifying LLVM code :)

PR #4046, opened 2021-08-02
Fix a bug in nominal LUB calculation

tlively:
The wrong variable was null checked, leading to segfaults.


tlively:
Will add a test momentarily.

kripken (approved):

PR #4047, opened 2021-08-02
[Wasm GC] Remove I31New from isSingleConstantExpression. NFC

kripken:
We have a not-very-clear distinction between

isSingleConstantExpression

and

canInitializeGlobal

To make it useful, the former should mean a constant expression:
something that is a simple constant like a number or a ref.func, that
will always be the same value at runtime. That means we should
move the I31 constructor from there to the second function. After that,
the I31 constructor is no longer in a weird place, and is alongside its
sibling constructors for other GC things.

This is NFC, but perhaps it also fixes a bug in a pass that uses
isSingleConstantExpression incorrectly. I was unable to find such
a bug, however. After this PR, I think we can use
isSingleConstantExpression in more places (like in an upcoming
GC type LTO pass I am working on, that wants to propagate
constant values).

Also add ArrayNew alongside StructNew and I31New.

### DIFF
--- a/src/ir/global-utils.h
+++ b/src/ir/global-utils.h
@@ -64,7 +64,7 @@ inline bool canInitializeGlobal(const Expression* curr) {
   }
   return Properties::isSingleConstantExpression(curr) ||
          curr->is<GlobalGet>() || curr->is<RttCanon>() || curr->is<RttSub>() ||
-         curr->is<StructNew>();
+         curr->is<StructNew>() || curr->is<ArrayNew>() || curr->is<I31New>();
 }
 
 } // namespace GlobalUtils
--- a/src/ir/global-utils.h
+++ b/src/ir/global-utils.h
@@ -63,9 +63,9 @@ inline bool canInitializeGlobal(Expression* curr) {
     }
     return true;
   }
-  if (Properties::isSingleConstantExpression(curr) ||
-      curr->is<GlobalGet>() || curr->is<RttCanon>() || curr->is<RttSub>() ||
-      curr->is<StructNew>() || curr->is<ArrayNew>() || curr->is<I31New>()) {
+  if (Properties::isSingleConstantExpression(curr) || curr->is<GlobalGet>() ||
+      curr->is<RttCanon>() || curr->is<RttSub>() || curr->is<StructNew>() ||
+      curr->is<ArrayNew>() || curr->is<I31New>()) {
     for (auto* child : ChildIterator(curr)) {
       if (!canInitializeGlobal(child)) {
         return false;
--- a/src/ir/global-utils.h
+++ b/src/ir/global-utils.h
@@ -20,6 +20,7 @@
 #include <algorithm>
 #include <vector>
 
+#include "ir/iteration.h"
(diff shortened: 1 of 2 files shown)


src/ir/global-utils.h
```diff
@@ -64,7 +64,7 @@ inline bool canInitializeGlobal(const Expression* curr) {
   }
   return Properties::isSingleConstantExpression(curr) ||
          curr->is<GlobalGet>() || curr->is<RttCanon>() || curr->is<RttSub>() ||
-         curr->is<StructNew>();
+         curr->is<StructNew>() || curr->is<ArrayNew>() || curr->is<I31New>();
```
tlively, line 67: Shouldn't we be walking or recursing into the operands to e.g. struct.new to make sure none of the operands are disallowed in global initializers? Previously we only allowed constant i31.new, but now we would allow an i31.new wtth a multiplication or similar as its operand.

kripken: Good point, this was inconsistent - we checked I31 children but not Struct ones. Fixed to check all children.

(Removed a `const*` to make this compile - keeping the const would require refactoring the Iterator class. Not sure that's worth it, given the const here doesn't give much benefit, not sure why it was there.)

src/ir/global-utils.h
```diff
@@ -62,9 +63,17 @@ inline bool canInitializeGlobal(const Expression* curr) {
...
     return true;
   }
-  return Properties::isSingleConstantExpression(curr) ||
-         curr->is<GlobalGet>() || curr->is<RttCanon>() || curr->is<RttSub>() ||
-         curr->is<StructNew>();
+  if (Properties::isSingleConstantExpression(curr) || curr->is<GlobalGet>() ||
+      curr->is<RttCanon>() || curr->is<RttSub>() || curr->is<StructNew>() ||
+      curr->is<ArrayNew>() || curr->is<I31New>()) {
+    for (auto* child : ChildIterator(curr)) {
+      if (!canInitializeGlobal(child)) {
```
tlively, line 70: Would it be worth making a one-off visitor here to avoid the recursion?

tlively (approved):
LGTM if you don't want to use a Visitor. The commit title and description should probably be updated, though.

src/ir/global-utils.h
kripken: Not sure what you mean by "one-off visitor"?

(If you mean to avoid recursion using an assumption about the depth, I think this can be pretty deep, e.g., struct.new with a field that contains another struct.new, repeated rtt.subs on each other, etc.)

tlively: Sorry, I meant `Walker` when I wrote `Visitor`. I was thinking that we could use a `PostWalker` subclass to avoid the current unbounded recursion.

kripken: Oh, I see. I agree that a Walker would be more efficient, but it would be more code and less clear IMO.

OTOH maybe that's a reason to try to make a more ergonomic walker for things like this... let me land this, and I'll think about the ergonomics. If I have a good idea there I can do a followup for this and other places in the code maybe.

PR #4051, opened 2021-08-03
[Wasm GC] Fix LocalSubtyping on unreachable sets with incompatible values

kripken:
We ignore sets in unreachable code, but their values may not be
compatible with a new type we specialize a local for. That is, the
validator cares about unreachable sets, while logically we don't need
to, and this pass doesn't. Fix up such unreachable sets at the end.

### DIFF
--- a/src/passes/LocalSubtyping.cpp
+++ b/src/passes/LocalSubtyping.cpp
@@ -27,8 +27,8 @@
 #include <ir/local-graph.h>
 #include <ir/utils.h>
 #include <pass.h>
-#include <wasm.h>
 #include <wasm-builder.h>
+#include <wasm.h>
 
 namespace wasm {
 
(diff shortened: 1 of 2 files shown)


src/passes/LocalSubtyping.cpp
```diff
@@ -178,10 +179,37 @@ struct LocalSubtyping : public WalkerPass<PostWalker<LocalSubtyping>> {
...
+        if (!Type::isSubType(set->value->type, newType)) {
+          // The type is incompatible. To fix this, replace
+          //
+          //  (set (bad-value))
+          //
+          // with
+          //
+          //  (set (block
+          //   (drop (bad-value))
+          //   (get)
```
tlively, line 203: Why even keep the `set` or even the operand? If this is unreachable code we can just replace the whole expression with `unreachable`. Is it to avoid having to refinalize? If so, how about we replace `(set (bad-value))` with `(nop)` and `(tee (bad-value)` with `(get)`?

Both in my suggestion and in the current code, can the `(get)` in the new pattern cause problems with non-nullable locals enabled?

tlively (commented):
So many edge cases!

src/passes/LocalSubtyping.cpp
kripken: We do still need to keep the value around, as it might have noticeable effects. In particular, a block is reachable if it has any branch to it, and that includes a branch in unreachable code (validation doesn't require first identifying unreachable code and then ignoring branches in it).

We could replace the set, but that would take a little more work (to get pointers to the sets), that I'm not sure is worth it.

It would be simpler to run DCE before this pass, or require DCE to have been run. But we don't have a current mechanism for such things, and it would make this pass less focused to add an internal DCE run before we do anything.

Good point about non-nullable locals and the get... depending on what the final validation rules look like there, this might be an issue. I think we can just put an unreachable there. As we already do a refinalize, that would propagate properly. Done.

aheejin (approved):

PR #4052, opened 2021-08-04
[Wasm GC] Constant Field Propagation

kripken:
A field in a struct is constant if we can see that in the entire program we
only ever write the same constant value to it. For example, imagine a
vtable type that we construct with the same funcrefs each time, then (if
we have no struct.sets, or if we did, and they had the same value), we
could replace a get with that constant value, since it cannot be anything
else:
```wat

(struct.new $T (i32.const 10) (rtt))

..no other conflicting values..

(struct.get $T 0) => (i32.const 10)
```

If the value is a function reference, then this may allow other passes
to turn what was a `call_ref` into a direct call and perhaps also get
inlined, effectively a form of devirtualization.

This only works in nominal typing, as we need to know the supertype
of each type. (It could work in theory in structural, but we'd need to do
hard work to find all the possible supertypes, and it would also
become far less effective.)

This deletes a trivial test for running `-O` on GC content. We have
many more tests for GC at this point, so that test is not needed, and
this PR also optimizes the code into something trivial and
uninteresting anyhow.


src/passes/ConstantFieldPropagation.cpp
```diff
@@ -0,0 +1,393 @@
...
+      return;
+    }
+
+    // We can do this! Replace the get with a throw on a null reference (as the
+    // get would have done so), plus the constant value. (Leave it to further
+    // optimizations to get rid of the ref.)
+    replaceCurrent(builder.makeSequence(
+      builder.makeDrop(builder.makeRefAs(RefAsNonNull, curr->ref)),
+      builder.makeConstantExpression(info.getConstantValue())));
+    changedTypes = true;
```
tlively, line 290: The struct field type and the actual value type are both known here, so we could be more precise and say the types have only changed if the types are different.

src/passes/ConstantFieldPropagation.cpp
```diff
@@ -0,0 +1,393 @@
...
+    // Combine the data from the functions.
+    StructValuesMap combinedInfos;
+    for (auto& kv : functionInfos) {
+      StructValuesMap& infos = kv.second;
+      for (auto& kv : infos) {
+        auto type = kv.first;
+        auto& info = kv.second;
+        auto& combinedInfo = combinedInfos[type];
+        for (Index i = 0; i < info.size(); i++) {
+          combinedInfo[i].combine(info[i]);
```
tlively, line 334: nit: the `combinedInfo` intermediate variable doesn't pull its weight IMO.
```suggestion
          combinedInfos[type][i].combine(info[i]);
```

src/passes/ConstantFieldPropagation.cpp
```diff
@@ -0,0 +1,393 @@
...
+    scanner.run(runner, module);
+    scanner.walkModuleCode(module);
+
+    // Combine the data from the functions.
+    StructValuesMap combinedInfos;
+    for (auto& kv : functionInfos) {
+      StructValuesMap& infos = kv.second;
+      for (auto& kv : infos) {
+        auto type = kv.first;
+        auto& info = kv.second;
```
tlively, line 331: It would be awfully nice to have C++17's destructuring assignment here :(

src/passes/ConstantFieldPropagation.cpp
```diff
@@ -0,0 +1,393 @@
...
+    //
+    //  (struct.set $B x (REF-2) (..value..))
+    //
+    // where $B is a subtype of $A, because at runtime the value REF-1 might not
+    // only be of type $A but also any subtype of $A. To make the lookup
+    // efficient when we see the get, we propagate information about sets to
+    // their supers, recursively, so long as the super still has that field.
+    // (Note that this also handles the case of REF-2 being a subtype of $B.)
+    //
+    // TODO: Are cycles possible with nominal supertypes?
```
tlively, line 357: No, cycles are disallowed. Would it be worth a TODO to visit the types in topological order to avoid repeated work?

tlively (commented):
I still have to take a look at the tests, but here are some comments on the code!

test/lit/passes/cfp.wast
```diff
@@ -0,0 +1,1350 @@
...
+;; an even field, but gets of the subtype and sub-sub-type can always be
+;; optimized, even the even ones, as a reference there cannot be to the first
+;; type.
+(module
+  ;; CHECK:      (type $struct3 (struct (field i32) (field i32) (field f64) (field f64) (field anyref) (field anyref)) (extends $struct2))
+  (type $struct3 (struct i32 i32 f64 f64 anyref anyref) (extends $struct2))
+
+  ;; CHECK:      (type $struct1 (struct (field i32) (field i32)))
+  (type $struct1 (struct i32 i32))
+  ;; CHECK:      (type $struct2 (struct (field i32) (field i32) (field f64) (field f64)) (extends $struct1))
```
tlively, line 904:
```suggestion

  ;; CHECK:      (type $struct2 (struct (field i32) (field i32) (field f64) (field f64)) (extends $struct1))
```

test/lit/passes/cfp.wast
```diff
@@ -0,0 +1,1350 @@
...
+    )
+    (drop
+      (struct.new_with_rtt $struct3
+        (i32.const 10)
+        (i32.const 999) ;; use a different value here
+        (f64.const 2.71828)
+        (f64.const 9.9999999)
+        (ref.null any)
+        (ref.as_non_null (ref.null any)) ;; use a non-constant value here, which
+                                         ;; can never be optimized.
```
tlively, line 947: This would have to be updated if we look at fallthrough values, I think.

test/lit/passes/cfp.wast
```diff
@@ -0,0 +1,1350 @@
...
+  )
+)
+
+;; Multi-level subtyping with conflicts. The even-numbered fields will get
+;; different values in the sub-most type. Create the top and bottom types, but
+;; not the middle one. As a result, gets of the first type cannot optimize
+;; an even field, but gets of the subtype and sub-sub-type can always be
+;; optimized, even the even ones, as a reference there cannot be to the first
+;; type.
+(module
```
tlively, line 898: Subtypes don't actually have to add more fields. This test might be simpler if all the structs had the same fields.

test/lit/passes/cfp.wast
```diff
@@ -0,0 +1,1350 @@
...
+        (ref.null $struct)
+      )
+    )
+    ;; Also test for multiple gets of the same field.
+    (drop
+      (struct.get $struct 4
+        (ref.null $struct)
+      )
+    )
+  )
```
tlively, line 1349: It would be good to have a struct.get of the rtt field as well; I don't think any of the other tests tried to get an rtt.

tlively:
This looks really nice! Do you have a sense of how much it optimizes j2cl output? Until V8 can parse the nominal output, you can test this by running the optimizer first in nominal mode to get this optimization then in equirecursive mode to strip the nominal extensions and emit code that V8 understands.

src/passes/ConstantFieldPropagation.cpp
```diff
@@ -0,0 +1,393 @@
...
+  }
+
+private:
+  FunctionStructValuesMap* functionInfos;
+
+  StructValuesMap& getInfos() { return (*functionInfos)[getFunction()]; }
+
+  // Note a value, checking whether it is a constant or not.
+  void noteExpression(Expression* expr, PossibleConstantValues& info) {
+    if (!Properties::isConstantExpression(expr)) {
```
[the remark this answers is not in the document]
kripken: Good point! Added.

src/passes/ConstantFieldPropagation.cpp
kripken: Fair point. I'll rename the constant and add a comment. (I'd rather not add more code, which would add complexity.)

src/passes/ConstantFieldPropagation.cpp
kripken: Even this, when no field is added?
```wat
(type $A (struct i32 (extends $B)))
(type $B (struct i32 (extends $A)))
```
If so, then different nominal types can never be equivalent, IIUC?

I'll add a TODO.

test/lit/passes/cfp.wast
kripken: The current test does also check for fields not existing in the super, however?

test/lit/passes/cfp.wast
kripken: Done, switched to a local.get.

test/lit/passes/cfp.wast
kripken: Not sure I follow - the rtt is not a field that can be read. Are you suggesting adding a new test where a field happens to be an rtt?

kripken (commented):
Thanks for the feedback!

kripken:
> Do you have a sense of how much it optimizes j2cl output?

I finally got that to build now... sadly it crashes, so there is a bug I need to figure out...

(Ignoring that, running `--cfp` on that binary removes around 10% of `struct.get`s, of which over 90% are turned into `ref.func`s. Seems promising, but until it's not buggy it's hard to be sure...)

test/lit/passes/cfp.wast
tlively: Oh, I confused myself and thought that the rtt operand to struct.new was also a field of the created struct. No need to change anything!

tlively:
Nice! This LGTM so far, but I assume you want to find that bug before landing.

kripken:
The bug is now fixed. The issue is that we need to propagate types both ways... given a `struct.get` and a `struct.set`, it is symmetrical to check whether they alias: if one is a subtype of the other, they can alias.

src/passes/ConstantFieldPropagation.cpp
```diff
@@ -348,17 +380,25 @@ struct ConstantFieldPropagation : public Pass {
...
-    // where $B is a subtype of $A, because at runtime the value REF-1 might not
-    // only be of type $A but also any subtype of $A. To make the lookup
-    // efficient when we see the get, we propagate information about sets to
-    // their supers, recursively, so long as the super still has that field.
-    // (Note that this also handles the case of REF-2 being a subtype of $B.)
+    // then
+    //
+    //  1. If $B is a subtype of $A, it is relevant: the actual value might be
+    //     of type $B (or even more specific), and cast to $A before it reaches
+    //     the get.
```
tlively, line 392:
```suggestion
    //     of type $B (or even more specific).
```
Even at the engine level, there's no cast when accessing a value that is dynamically a subtype of the statically accessed type.

kripken:
I measured some initial data on j2cl that looks promising. This pass removes 8.4% of total `struct.get`s, 95% of which are `ref.func`s. So, as expected, this appears to mostly apply to vtable-like things. Combined with other opts, this ends up removing over 2/3 of all `call_ref`s (turning them into direct calls), which shrinks code by 4% and makes it 18% faster. (This will depend a lot on inlining heuristics etc., though, so I wouldn't focus on the specific numbers too much.)

tlively:
Ugh I wrote a big second comment but it looks like it got lost when I finalized the review. Will recreate most of it...

src/passes/ConstantFieldPropagation.cpp
```diff
@@ -0,0 +1,450 @@
...
+    //  (struct.set $B x (REF-2) (..value..))
+    //
+    // then
+    //
+    //  1. If $B is a subtype of $A, it is relevant: the actual value might be
+    //     of type $B (or even more specific), and cast to $A before it reaches
+    //     the get.
+    //  2. If $B is a supertype of $A that still has the field x then it may
+    //     also be relevant: the actual value might be of type $A (or even more
+    //     specific), and cast to $B before it reaches the set.
```
tlively, line 395: I was confused by this at first, but I get it now. Here's how I would explain it:

Given (struct.get $A x ...) and (struct.set $B x ...)

 - If $A <: $B, then (struct.set $B x ...) can write a struct of type $A.
 - If $B <: $A, then (struct.get $A x ...) can read a struct of type $B.

In both cases, the struct.get can observe the struct.set, so struct.sets to all subtypes and supertype must agree for a get of a type to be optimized.

As you've noted, struct.new does not have this bidirectionality and could be modeled separately. This will be important for devirtualization because vtables are never written to after their creation and also because vtable subtypes will have different values than their supertypes.

src/passes/ConstantFieldPropagation.cpp
tlively: There's no cast in either case, so I think the mention of casts is part of what confused me at first.

src/passes/ConstantFieldPropagation.cpp
```diff
@@ -39,6 +40,36 @@ namespace wasm {
...
+private:
+  // Add a type to the graph.
+  void note(HeapType type) {
+    HeapType super;
+    if (type.getSuperType(super)) {
+      typeSubTypes[super].insert(type);
+    }
+  }
+
+private:
```
tlively, line 68:
```suggestion
```

tlively (approved):
LGTM with hat small change and the comments clarified. It would be good to model struct.new separately soon in a follow up.

src/passes/ConstantFieldPropagation.cpp
kripken: Yes, maybe "cast" is confusing. I meant that the value is passed into a location declared as a supertype (which is possible without casting).

src/passes/ConstantFieldPropagation.cpp
kripken: Good summary, updated the comment now.

PR #4062, opened 2021-08-06
Fix BrOn logic in RemoveUnusedBrs

kripken:
This only moves code around. `visitBrOn` was in the main part of the pass,
which was incorrect as it could interfere with other work being done there.
Specifically, we have a stack of Ifs there, and if we replace a BrOn with
an If, an assertion was hit.  To fix this, run it like `sinkBlocks()`, in a separate
interleaved phase.

This fixes a bug reported by @askeksa-google here:

https://github.com/WebAssembly/gc/issues/226#issuecomment-868739853

### DIFF
--- a/src/passes/RemoveUnusedBrs.cpp
+++ b/src/passes/RemoveUnusedBrs.cpp
@@ -703,7 +703,14 @@ struct RemoveUnusedBrs : public WalkerPass<PostWalker<RemoveUnusedBrs>> {
 
     optimizer.setModule(getModule());
     optimizer.doWalkFunction(func);
-    return optimizer.worked;
+
+    // If we removed any BrOn instructions, that might affect the reachability
+    // of the things they used to break to, so update types.
+    if (optimizer.worked) {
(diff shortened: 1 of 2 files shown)


aheejin:
Which assertion is hit? If RemoveUnusedBrs also optimizes `if`s, do we not optimize the new `if` (replaced with `br_on`)?

kripken:
There is some logic using an `ifStack` in this pass, and we hit an assertion on using it badly (popping from an empty stack). I think the issue is that we always call `visitAny` for all nodes, after their `visitFoo`. But if `visitBrOn` replaces the BrOn with an If, then we end up calling `visitAny` on that replacement If (since the visitor is called on the pointer to the pointer). That If was a child of the BrOn, and we've already visited it, so we try to pop the if stack again.

(Overall that logic is likely too complicated anyhow...)

kripken:
We should optimize the if. First, it was a child of the BrOn anyhow, so it's been processed. But also, we will do another full iteration through the pass, so we will see it at least once more.

aheejin (approved):

PR #4063, opened 2021-08-06
Refactor TypeUpdating non-nullability code. NFC

kripken:
This moves code out of the existing methods so that it can be
called directly from more places. A later PR will use the new
methods.

### DIFF
--- a/src/ir/type-updating.cpp
+++ b/src/ir/type-updating.cpp
@@ -51,13 +51,8 @@ void handleNonDefaultableLocals(Function* func, Module& wasm) {
       // We do not need to process params, which can legally be non-nullable.
       continue;
     }
-    auto type = func->getLocalType(get->index);
-    if (type.isNonNullable()) {
-      // The get should now return a nullable value, and a ref.as_non_null
-      // fixes that up.
-      get->type = Type(type.getHeapType(), Nullable);
-      *getp = builder.makeRefAs(RefAsNonNull, get);
-    }
+    get->type = func->getLocalType(get->index);
+    *getp = fixLocalGet(get, wasm);
   }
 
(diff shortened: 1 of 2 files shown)


src/ir/type-updating.cpp
```diff
@@ -51,13 +51,8 @@ void handleNonDefaultableLocals(Function* func, Module& wasm) {
...
     }
-    auto type = func->getLocalType(get->index);
-    if (type.isNonNullable()) {
-      // The get should now return a nullable value, and a ref.as_non_null
-      // fixes that up.
-      get->type = Type(type.getHeapType(), Nullable);
-      *getp = builder.makeRefAs(RefAsNonNull, get);
-    }
+    get->type = func->getLocalType(get->index);
+    *getp = fixLocalGet(get, wasm);
```
aheejin, line 55: Nit: It feels a little unintuitive to me that we assign `get->type` before we call `fixLocalGet` which will possibly modify/fix `local.get` itself. Here `get->type` is more like the result type the the return value of `fixLocalGet` should have.. How about passing it as an argument to `fixLocalGet`, like,
```cpp
Expression* fixLocalGet(LocalGet* get, Module& wasm, Type resultType);
```

aheejin (approved):

src/ir/type-updating.cpp
kripken: Ah, good point, that line does look unintuitive. I think in fact that `get->type = ` assignment is not needed. I removed it.

(I think the API does not need to change after that removal, but we can consider it later if needed.)

kripken: (the line is not needed since the get's type was already set to the local type - that line happens before we change anything)

PR #4064, opened 2021-08-06
[Wasm GC] Track struct.new and struct.set separately in ConstantFieldPropagation

kripken:
Previously we tracked them in the same way. That means that we did the same
when seeing if either a `struct.new` or a `struct.set` can write to the memory that
is read by a `struct.get`, where the rule is that if either type is a subtype of the
other then they might. But with `struct.new` we know the precise type, which
means we can do better.

Specifically, if we see a new of type B, then only a get of a supertype of B can
possibly read that data: it is not possible for our struct of type B to appear in
a location that requires a subtype of B. Conceptually:

```python
A = type struct
B = type extends A
C = type extends B

x = struct.new<B>

struct.get<A>(y) // x might appear here, as it can be assigned to a
                 // variable y of a supertype

struct.get<C>(y) // x cannot appear here
```

This allows more devirtualization. It is a followup for #4052 that implements
a TODO from there.

The diff without whitespace is simpler.

### DIFF
--- a/test/lit/passes/cfp.wast
+++ b/test/lit/passes/cfp.wast
@@ -553,9 +553,7 @@
 
 ;; Subtyping: Create a supertype and get a subtype. As we never create a
 ;;            subtype, the get must trap anyhow (the reference it receives can
-;;            only by null in this closed world). Since the only possible writes
-;;            to the field are of the value 10, we will optimize to that. (But
-;;            as it will trap anyhow, that doesn't really matter.)
+;;            only be null in this closed world).
 (module
   ;; CHECK:      (type $none_=>_none (func))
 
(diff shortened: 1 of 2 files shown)


test/lit/passes/cfp.wast
```diff
@@ -771,17 +824,15 @@
...
-;;            for the shared field, but get from the subtype. As the field is
-;;            shared between the types, we cannot optimize here: we model the
-;;            struct.new as a set, and if there is a set of the supertype, then
-;;            it might be to a reference of the subtype.
-;;            TODO: model new and set separately
+;;            for the shared field, but get from the subtype. The field is
+;;            shared between the types, but we only create the substruct with
+;;            one value, so we can optimize.
 (module
+  ;; CHECK:      (type $none_=>_none (func))
```
tlively, line 831: It would be good to reorder the types in the input to match the output both here and below.

tlively (approved):
Awesome! Did this make a big difference on top of the last PR?

test/lit/passes/cfp.wast
kripken: Ah, thanks, looks like I missed a few...

kripken:
This made a fairly small difference it seems. I don't remember the specific number, but it was less than 10%. Still, any devirtualization that ends up affecting an inner loop can be a big deal...

PR #4066, opened 2021-08-09
[Wasm GC] RefEq(x, null) => RefIsNull(x)

kripken:


src/passes/OptimizeInstructions.cpp
```diff
@@ -1117,6 +1117,18 @@ struct OptimizeInstructions
   }
 
   void visitRefEq(RefEq* curr) {
+    // Canonicalize nulls to make the next pattern simpler (and also it may
+    // compress better).
```
aheejin, line 1121: Not sure what this means... Aren't we converting this into `ref.is_null` after all? Not sure "next pattern simpler" and "compress better" means.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1117,6 +1117,18 @@ struct OptimizeInstructions
...
+    if (curr->left->is<RefNull>()) {
+      std::swap(curr->left, curr->right);
+    }
+
+    // RefEq of a value to Null can be replaced with RefIsNull.
+    if (curr->right->is<RefNull>()) {
+      replaceCurrent(Builder(*getModule()).makeRefIs(RefIsNull, curr->left));
+      return;
+    }
```
aheejin, line 1130: Is there a case when we are comparing to `ref.is_null`s so we can end up with `i32.const 1`? Should we move these after the `areConsecutiveInputsEqual` check below?

kripken: 1. Hmm, at the moment we don't optimize `ref.is_null` to 1 ever - but we should! I'll add that in a followup. In that case the two optimizations would work together in a single pass (a replacement will cause the rules to all be tried again).

2. I don't think the order of the cases matters here, as there is no overlap. I just put this first because it is faster to compute, for lack of a better reason.

src/passes/OptimizeInstructions.cpp
kripken: Good point, yeah, that wasn't very clear... I rewrote it. Also the compression part is actually inaccurate - if there is a null, we optimize it out anyhow, so we can't be helping compression... I removed that bit.

src/passes/OptimizeInstructions.cpp
kripken: (for 1, actually there is no need for a followup, as the Precompute pass will turn `ref.is_null(ref.null)` into 1 already)

aheejin:
Sorry these might be basic questions:
- Does `(ref.eq (ref.null eq) (ref.null data))` evaluate to true?
- When the local `$x`'s type is `eqref`, is `(ref.eq (ref.null data) (local.get $x))` equivalent to `(ref.is_null (local.get $x))`?

So the gist of the questions is, should we care about the heap type?

src/passes/OptimizeInstructions.cpp
aheejin: > 2. I don't think the order of the cases matters here, as there is no overlap. I just put this first because it is faster to compute, for lack of a better reason.

If the both sides are `ref.null`, aren't they considered as "ConsecutiveInputsEqual" and can be converted to `i32.const 1` here if we do `areConsecutiveInputsEqual` first? For example,

```wasm
      (ref.eq
        (ref.null eq)
        (ref.null eq)
      )
```
Currently this will be converted to `(ref.is_null (ref.null eq))`, but can't we make this just `(i32.const 1)`?

Maybe doesn't matter much if we do that anyway, but looks simple enought to be done.

kripken: Oh, you're right, these patterns do collide there...

While precompute will turn it into a constant later anyhow, it makes sense to do it more eagerly here since it's just a reordering.

kripken:
All nulls compare equal, no matter their type, so we don't need to think about the heap type here.

aheejin (approved):

PR #4068, opened 2021-08-09
Improve optimization of call_ref into direct calls

kripken:
First, move the tiny pattern of call-ref-of-ref-func from Directize
into OptimizeInstructions. This is important because Directize is
a global optimization pass - it looks at the table to see if a
CallIndirect can be turned into a direct call. We only run global
passes at the end of the pipeline, but we don't need any global
data for call-ref of a ref-func, and OptimizeInstructions is the
place for such patterns.

Second, extend that to also handle fallthrough values. This is
less simple, but as call_ref is so inefficient, it's worth doing all
we can.


test/lit/passes/optimize-instructions-call_ref.wast
```diff
@@ -0,0 +1,201 @@
...
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT: )
+ (func $fallthrough (param $x i32)
+  ;; This call_ref should become a direct call, even though it doesn't have a
+  ;; simple ref.func as the target - we need to look into the fallthrough, and
+  ;; handle things with locals.
+  (call_ref
+   ;; Write to $x before the block, and write to it in the block; we should not
+   ;; reorder these things as the side effects could alter what value appears
+   ;; in the get of $x.
```
aheejin, line 74: Is there any reordering optimization in this PR? I don't think there's one, so not sure why it is mentioned

test/lit/passes/optimize-instructions-call_ref.wast
```diff
@@ -0,0 +1,201 @@
...
+ ;; CHECK-NEXT:    (rtt.canon $none_=>_i32)
+ ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT: )
+ (func $fallthrough-bad-type (result i32)
+  ;; A fallthrough appears here, and we cast the function type to something else
+  ;; in a way that is bad: the actual target function has a different return
+  ;; type than the cast type. The cast will fail at runtime, and we should not
+  ;; emit non-validating code here, which would happen if we replace the
+  ;; call_ref that returns an i32 with a call that returns nothing.
```
aheejin, line 154:
```suggestion
  ;; call_ref that returns nothing with a call that returns an i32.
```

aheejin (approved):

test/passes/Oz_fuzz-exec_all-features.txt
```diff
@@ -351,7 +351,9 @@
   (call $log
    (i32.const 2)
   )
-  (call $a-void-func)
+  (call $log
+   (i32.const 1337)
+  )
```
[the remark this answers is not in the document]
kripken: This begins as
```wat
  (call_ref
   (ref.cast (ref.func $a-void-func) (rtt.canon $void_func))
  )
```
OptimizeInstructions can remove the cast, and then also turn the call_ref of a ref.func into a direct call. Then inlining brings that 1337 to here.

Previously, only Directize would turn this into a direct call - and that runs at the very end of the pipeline, after inlining. So inlining did not happen.

test/lit/passes/optimize-instructions-call_ref.wast
kripken: There isn't, but there is a risk of reordering if we naively moved the call target (the block) to before the first parameter (the tee). Instead, we append it after the second param (the get) which keeps the order. I'll clarify in the test.

test/lit/passes/optimize-instructions-call_ref.wast
```diff
@@ -0,0 +1,201 @@
...
+ (func $foo (param i32) (param i32)
+  (unreachable)
+ )
+ ;; CHECK:      (func $bar (param $x i32) (param $y i32)
+ ;; CHECK-NEXT:  (call $foo
+ ;; CHECK-NEXT:   (local.get $x)
+ ;; CHECK-NEXT:   (local.get $y)
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT: )
+ (func $bar (param $x i32) (param $y i32)
```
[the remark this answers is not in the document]
kripken: Done. This was moved to here from the existing Directize test and I didn't notice the name...

PR #4069, opened 2021-08-09
Fix OptimizeInstructions on folding of identical code with nominal typing

kripken:
```wat
(if (result i32)
 (local.get $x)
 (struct.get $B 1
  (ref.null $B)
 )
 (struct.get $C 1
  (ref.null $C)
 )
)
```

With structural typing it is safe to turn this into this:

```wat
(struct.get $A 1
 (if (result (ref $A))
  (local.get $x)
  (ref.null $B)
  (ref.null $C)
 )
)
```

Here `$A` is the LUB of the others. This works since `$A` must have
field 1 in it. But with nominal types it is possible that the LUB in fact
does not have that field, and we would not validate.

This actually seems like a more general issue that might happen with
other things, even though atm perhaps it can't. For simplicity, avoid this
pattern in both nominal and structural typing, to avoid making a
difference between them.

This happened on a 30MB wasm file that took some time to reduce and
investigate...


test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -2,10 +2,11 @@
 ;; RUN: wasm-opt %s --remove-unused-names --optimize-instructions --enable-reference-types --enable-gc -S -o - \
 ;; RUN:   | filecheck %s
 ;; RUN: wasm-opt %s --remove-unused-names --optimize-instructions --enable-reference-types --enable-gc --nominal -S -o - \
-;; RUN:   | filecheck %s
+;; RUN:   | filecheck %s --check-prefix NOMNL
```
aheejin, line 5: - Is this file autogenerated by update_lit_checks.py, as written at the top? Then why does it contain `(type ...)`s when the top comment line does not say `update_lit_checks.py --all-items`?
- Do we need separate check lines for structural and nominal types, given that we don't do any structural- or nominal- type dependent optimizations in OptimizeInstructions now? Those check lines look mostly the same, except some `(types)` in `NOMNL` lines have `extends` clause in them, which I don't think is relevant to OptimizeInstructions pass
- If we need both set of check lines after all, can we split adding the `NOMNL` check lines as a separate PR and rebase this PR on top of that?

tlively: The script emits checks for all items that have $names by default. The types being the only difference between nominal and structural typing is a common case, so it probably worth adding a flag to the update script to skip emitting checks for types or similar. A more complex solution that would be closer to what we might write by hand would be to have NOMNL and EQREC prefixes for types and a shared CHECK prefix for everything else, but that would be hard to add to the script.

kripken: I see, thanks.

Ok, for now it seems like doing what @aheejin suggested and first adding the new prefix to this test, then rebasing this on top, is the best until we figure out a more general script solution. I opened #4081

kripken: Rebased.

aheejin (approved):

PR #4070, opened 2021-08-09
SimplifyGlobals: Optimize away globals that are only read in order to write themselves

kripken:
If the only uses of a global are
```js
if (global == 0) {
  global = 1;
}
```
Then we do not need that global: while it has both reads and
writes, the value in the global does not cause anything observable.
It is read, but only to write to itself, and nothing else.

This happens in real-world code from j2cl quite a lot, as they have
an initialization pattern with globals, and in some cases we can
optimize away the work done in the initialization, leaving only the
globals in this pattern.


src/passes/SimplifyGlobals.cpp
```diff
@@ -250,21 +320,79 @@ struct SimplifyGlobals : public Pass {
...
+        // TODO: optimize the case of an imported but immutable global, etc.
+        continue;
+      }
+
+      // We only ever read-only-to-write if all of our reads are done in places
+      // we identified as read-only-to-write. That is, we have eliminated the
+      // possibility of any other uses. (Technically, each read-to-write
+      // location might have more than one read since we did not count them, but
+      // only verified there was one read or more; but this is good enough as
+      // the common case has exactly one.)
```
aheejin, line 353: Not sure if I understand... Can you elaborate?

> each read-to-write location might have more than one read since we did not count them,

Didn't we count them? Aren't we checking the number of reads is the same as the number of `readOnlyToWrite` right below?
```cpp
bool onlyReadOnlyToWrite = (info.read == info.readOnlyToWrite);
```

src/passes/SimplifyGlobals.cpp
```diff
@@ -250,21 +320,79 @@ struct SimplifyGlobals : public Pass {
...
+      //
+      // Note that there might be more writes, if there are additional writes
+      // besides those in the read-only-to-write locations. But we can ignore
+      // those, as whatever they write will not be read in order to do anything
+      // of value.
+      bool onlyReadOnlyToWrite = (info.read == info.readOnlyToWrite);
+      assert(!onlyReadOnlyToWrite || info.written >= info.readOnlyToWrite);
```
aheejin, line 360: Not sure if I understand the purpose of this `assert`... Isn't `info.written >= info.readOnlyToWrite` always true, no matter `onlyReadOnlyToWrite` > 0 or not, because `written` is a superset of `readOnlyToWrite`?

test/lit/passes/simplify-globals-read_only_to_write.wast
```diff
@@ -0,0 +1,240 @@
...
+      )
+    )
+  )
+  ;; CHECK:      (func $baz
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (i32.const 2)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $baz
+    ;; An addition set does not prevent this optimization: the value written
```
aheejin, line 58:
```suggestion
    ;; An additional set does not prevent this optimization: the value written
```

src/passes/SimplifyGlobals.cpp
```diff
@@ -50,10 +50,25 @@ namespace wasm {
...
+  // How many times the global is written and read.
+  std::atomic<Index> written{0};
+  std::atomic<Index> read{0};
+
+  // How many times the global is "read, but only to write", that is, is used in
+  // this pattern:
+  //
+  //   if (global == X) { global = Y }
+  //
+  // where X and Y have no side effects. If all we have are such reads only to
```
aheejin, line 66: Why is it required that X and Y have no side effects? We are not reordering anything, so as long as the concerned global is read only to write, can't other patterns with other orthogonal side effects be optimized too?

Or, is this pattern without side effects just an easy and common enough to detect and optimize away? In that case, maybe better to elaborate that side effects don't necessary prevent the optimization but they are not the common pattern or something...

aheejin (commented):
I think I understand the gist of the PR, but have some minor questions:

src/passes/SimplifyGlobals.cpp
```diff
@@ -65,9 +80,53 @@ struct GlobalUseScanner : public WalkerPass<PostWalker<GlobalUseScanner>> {
...
+
+  void visitIf(If* curr) {
+    // We are looking for
+    //
+    //   if (global == X) { global = Y }
+    //
+    // Ignore an if-else, which cannot be that.
+    if (curr->ifFalse) {
+      return;
+    }
```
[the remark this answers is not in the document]
kripken: The first can work. I haven't seen it in j2cl output, but it's easy to do, good idea.

The second is not valid, though. If it has some other side effect than globals, like say writing to a local, then it matters what value the global has. So we can't remove the global. For example, if might set the global from 0 to 1 the first time we get here, and all other times it increments some local.

src/passes/SimplifyGlobals.cpp
kripken: We might have something like this:
```js
if (global + global == 1) { global = 2 }
```
That has two reads of the global. If that is the entire program, then we'd count 2 reads and only 1 read-only-to-write. In that case we wouldn't optimize, although it happens to be safe.

We could count the number of gets inside each read-only-to-write, but I'm not sure it's worth the complexity. The j2cl pattern is pretty specific.

src/passes/SimplifyGlobals.cpp
kripken: Good point, yeah, I'll fix that.

src/passes/SimplifyGlobals.cpp
kripken:
```js
if (global == 0) {
  global = 1;
  print("first time");
}
```
That side effect will be noticed, as it's printed the first time we reach it. So we can't remove writes to the global, as then we'd end up with
```js
// The read turns into 0, the default value, which is never
// written so it is constant.
if (0 == 0) {
  // The write that was here was removed.
  print("first time");
}
```
That will never print.

src/passes/SimplifyGlobals.cpp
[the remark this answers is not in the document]
kripken: Actually, thinking more on the first, I'm not sure it's worth it. If we start with if-else, we may as well do if-else-if etc. nesting - but then it gets more complicated. I think maybe we can wait on that to see if there is a use case. For now this PR handles the j2cl output that exists.

kripken:
I added more comments that I hope clarify the above questions.

src/passes/SimplifyGlobals.cpp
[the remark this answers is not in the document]
aheejin: > The second is not valid, though. If it has some other side effect than globals, like say writing to a local, then it matters what value the global has. So we can't remove the global. For example, if might set the global from 0 to 1 the first time we get here, and all other times it increments some local.

Even if it has other side effects, if what it does with the specified global is only reading it, wouldn't it be possible to replace it with a constant, because it is guaranteed that there's no other reads or writes to the global?

But anyway I understand this pattern has not been observed a lot yet and you want to optimize the common case first. Just wondering why the side effects, if they are not about the global, matter.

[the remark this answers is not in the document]
aheejin: Oh sorry, nevermind... https://github.com/WebAssembly/binaryen/pull/4070#discussion_r686108438 explains that.

aheejin (approved):

PR #4076, opened 2021-08-11
Fix the Switch operand order in LinearExecutionWalker

kripken:
This caused no noticeable bugs, but it could in theory in new passes - in fact
in a pass I will open later this week it did.

Also fix the order in wasm.h. That part has no effect, but it is nice to be
consistent. After this PR, everything should match the single source of
truth which is wasm-delegations-fields.h (as that is used in printing, binary
reading/writing, etc., so it has to be correct). Also Switch now matches
the ordering in Break.


kripken:
Actually I may have this wrong... the fuzzer found an issue.

kripken:
Sorry for the noise - this PR is right, I was running the fuzzer wrong.

I now have a proper testcase for this, but it needs to wait for a new pass that will allow testing.

aheejin (approved):

PR #4079, opened 2021-08-12
LocalCSE rewrite

kripken:
Technically this is not a new pass, but it is a rewrite almost from scratch.

Local Common Subexpression Elimination looks for repeated patterns,
stuff like this:

```
x = (a + b) + c
y = a + b

=>

temp = a + b
x = temp + c
y = temp
```

The old pass worked on flat IR, which is inefficient, and was overly
complicated because of that. The new pass uses a new algorithm that
I think is pretty simple, see the detailed comment at the top.

This keeps the pass enabled only in `-O4`, like before - right after
flattening the IR. That is to make this as minimal a change as possible.
Followups will enable the pass in the main pipeline, that is, we will
finally be able to run it by default. (Note that to make the pass work
well after flatten, an extra simplify-locals is added - the old pass used
to do part of simplify-locals internally, which was one source of
complexity. Even so, some of the `-O4` tests have changes, due to
minor factors - they are just minor orderings etc., which can be
seen by inspecting the outputs before and after using e.g.
`--metrics`)

This plus some followup work leads to large wins on wasm GC output.
On j2cl there is a common pattern of repeated `struct.get`s, so common
that this pass removes 85% of all `struct.get`s, which makes the total
binary 15% smaller. However, on LLVM-emitted code the benefit is
minor, less than 1%.


test/lit/passes/local-cse.wast
```diff
@@ -0,0 +1,433 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
+;; NOTE: This test was ported using port_test.py and could be cleaned up.
```
aheejin, line 2: Was it? It looks like it's newly written.. Was this copied from the deleted tests?

test/lit/passes/local-cse_all-features.wast
```diff
@@ -0,0 +1,272 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
+;; NOTE: This test was ported using port_test.py and could be cleaned up.
```
aheejin, line 2: The same... Was this ported from somewhere else?

test/lit/passes/local-cse.wast
```diff
@@ -0,0 +1,433 @@
...
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (i32.load
+  ;; CHECK-NEXT:    (i32.const 10)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $loads
+    ;; The possible trap on loads prevents optimization.
+    ;; TODO: optimize that too
```
aheejin, line 282: How? Is this related to TrapsNeverHappen pass?

test/lit/passes/local-cse.wast
```diff
@@ -0,0 +1,433 @@
...
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (global.set $glob
+  ;; CHECK-NEXT:   (i32.const 200)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (global.get $glob)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $global
+    ;; We should optimize redundantglobal.get operations.
```
aheejin, line 423:
```suggestion
    ;; We should optimize redundant global.get operations.
```

test/lit/passes/local-cse_all-features.wast
```diff
@@ -0,0 +1,272 @@
...
+  ;; CHECK-NEXT:      )
+  ;; CHECK-NEXT:      (i32.const 16)
+  ;; CHECK-NEXT:     )
+  ;; CHECK-NEXT:     (i32.shr_s
+  ;; CHECK-NEXT:      (i32.shl
+  ;; CHECK-NEXT:       (local.get $1)
+  ;; CHECK-NEXT:       (i32.const 16)
+  ;; CHECK-NEXT:      )
+  ;; CHECK-NEXT:      (i32.const 16)
+  ;; CHECK-NEXT:     )
```
aheejin, line 230: These two look the same and could have been optimized?

aheejin (commented):
I'm not familiar with the old pass, but the new pass looks very nice and straightforward!

test/lit/passes/local-cse_all-features.wast
aheejin: Oh sorry, the local numbers are different. But then this test contains only one thing to optimize (as the `CHECK` lines show), while the test description says "containing multiple nested things that can be optimized". Is this what you intended?

MaxGraey:
I'm wondering how hard to make more sophisticated scanner which also handle this case?
```ts
x = a + (b + c)
y = a + b
```
=>
```ts
temp = a + b
x = temp + c
y = temp
```
where a, b and are integers and associative property is valid.

src/ir/properties.h
```diff
@@ -335,6 +335,48 @@ inline bool canEmitSelectWithArms(Expression* ifTrue, Expression* ifFalse) {
...
+//
+// Thus, loads are nondeterministic but not intrinsically so, while GC
+// allocations are actual examples of intrinsically nondeterministic
+// instructions. If wasm were to add "get current time" or "get a random number"
+// instructions then those would also be intrinsically nondeterministic.
+//
+//  * The word "can" appears in "can return different results" in the definition
+//    at the top because of NaN nondeterminism, which we ignore here. It is a
+//    valid wasm implementation to have deterministic NaN behavior, and we
+//    optimize under that assumption.
```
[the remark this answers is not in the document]
kripken: The goal here is to ignore NaN nondeterminism. Yeah, this text doesn't look clear enough, and you're right the word "can" is not relevant. I'll rewrite this.

src/ir/properties.h
```diff
@@ -335,6 +335,48 @@ inline bool canEmitSelectWithArms(Expression* ifTrue, Expression* ifFalse) {
...
+//  * Note that calls are ignored here. In theory this concept could be defined
+//    either way for them, but it is simpler to ignore them. We only consider
```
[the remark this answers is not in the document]
kripken: I'll clarify this too.

src/ir/properties.cpp
```diff
@@ -0,0 +1,42 @@
...
+    return false;
+  }
+
+  struct Scanner : public PostWalker<Scanner> {
+    bool deterministic = true;
+    void visitStructNew(StructNew* curr) { deterministic = false; }
+    void visitArrayNew(ArrayNew* curr) { deterministic = false; }
+  } scanner;
+  scanner.walk(curr);
+  return !scanner.deterministic;
```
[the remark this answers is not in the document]
kripken: That would be slightly shorter to write, but it would do two iterations over the expression, so it is less efficient. Sometimes that clarity matters, but I think in this tiny function it is clear enough anyhow, and this might be called a lot in theory.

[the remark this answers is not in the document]
kripken: Good point that that pattern appears in OptimizeInstructions! I had totally forgotten about that place... will fix.

src/passes/pass.cpp
```diff
@@ -423,7 +423,11 @@ void PassRunner::addDefaultFunctionOptimizationPasses() {
   // that depend on flat IR
   if (options.optimizeLevel >= 4) {
     addIfNoDWARFIssues("flatten");
+    // LocalCSE is particularly useful after flatten, but we must simplify
+    // locals a little first (as flatten adds many new and redundant ones).
```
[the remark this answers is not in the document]
kripken: This was referring to this comment in the pass itself:
```cpp
// This pass only finds entire expression trees, and not parts of them, so we
// will not optimize this:
//
//  (A (B (C (D1))))
//  (A (B (C (D2))))
//
// The innermost child is different, so the trees are not identical. However,
// running flatten before running this pass would allow those to be optimized as
// well.
```
I'll improve the comment here to refer to there, and explain why simplify-locals is needed (flatten adds new redundant locals, so things look different that are actually equal, until we simplify locals).

test/lit/passes/local-cse.wast
kripken: Yes, this is ported from the deleted ones. The filename changes (since no more flatten), and also I did a lot of cleanup and changes, so I guess git doesn't consider it a "diff" but a new file.

test/lit/passes/local-cse.wast
kripken: The idea here is that if the same expression appears twice, it is ok to fold it into a single appearance with CSE: if it traps, the first appearance will trap and the second is not reached anyhow.

test/lit/passes/local-cse_all-features.wast
kripken: The inputs to the xors are identical, but that includes their children being identical. That's what it means by "multiple nested things", but maybe that's not very clear... Anyhow, the test checks we only CSE the outer parent, and do not CSE children unnecessarily, and also do not CSE where the locals are different.

Maybe this test is slightly redundant with the above ones, but it's real-world code that is nice to see is optimized 100% right I think, and it existed before in the test suite, this is just a movement.

kripken:
@MaxGraey Canonicalization of symmetric binary operations in OptimizeInstructions should help here. But yes, in general what is needed is to do more work to handle such things, which most GVNs do I believe.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1470,30 +1470,21 @@ struct OptimizeInstructions
...
-    // repeatedly calling (struct.new $foo)'s output will return different
-    // results (while i32.const etc. of course does not).
-    // Note that allocations may have appeared in the original inputs to this
-    // function, and skipped when we focused on what falls through; since there
-    // are no side effects, any allocations there cannot reach the fallthrough.
-    if (left->type.isRef()) {
-      if (FindAll<StructNew>(left).has() || FindAll<ArrayNew>(left).has()) {
-        return false;
-      }
+    // To be equal, they must also be know to return the same result
```
aheejin, line 1484:
```suggestion
    // To be equal, they must also be known to return the same result
```

aheejin (approved):

src/ir/properties.h
```diff
@@ -335,6 +335,51 @@ inline bool canEmitSelectWithArms(Expression* ifTrue, Expression* ifFalse) {
...
+//    So NaN nondeterminism does not cause anything to be intrinsically
+//    nondeterministic.
+//  * Note that calls are ignored here. In theory this concept could be defined
+//    either way for them (that is, we could potentially define them as either
+//    intrinsically nondeterministic, or not, and each could make sense in its
+//    own way). It is simpler to ignore them here, which means we only consider
+//    the behavior of the expression provided here (including its chldren), and
+//    not external code.
+//
+bool isIntrinsicallyNondeterministic(Expression* curr, FeatureSet features);
```
tlively, line 381: Drive-by comment: Perhaps "generative" would be a better term here than "intrinsically nondeterministic," since these instructions always generate new values but aren't really nondeterministic in the normal sense - they generate deterministically fresh values with deterministic structure. This would also be shorter and allow the discussion of NaNs to be elided because floating point operations are clearly not generative.

src/passes/LocalCSE.cpp
```diff
@@ -17,257 +17,532 @@
...
+//
+// This pass only finds entire expression trees, and not parts of them, so we
+// will not optimize this:
+//
+//  (A (B (C (D1))))
+//  (A (B (C (D2))))
+//
+// The innermost child is different, so the trees are not identical. However,
+// running flatten before running this pass would allow those to be optimized as
+// well.
```
tlively, line 107: Wouldn't running flatten leave these expressions with different local.gets as their children, so they still couldn't be optimized? Unless I'm missing something (which is likely), I think this kind of optimization would require outlining.

src/ir/properties.h
kripken: I agree "generative" would capture the concept of allocation, but if wasm adds a "get current time" or "get a random number" methods then we'd need something more than just allocation here. I'm open to better terms than nondeterminism, though - I considered "nonconstant", "impure", but couldn't think of anything good...

src/passes/LocalCSE.cpp
kripken: Flatten would create different local.gets, correct - see pass.cpp, where we also run some local simplification. I'll mention that here too.

src/ir/properties.h
kripken: (I'll land this once tests pass, to unblock j2cl opts that depend on this, but let's keep discussing the terminology.)

tlively: I think "generative" would cover something like getting the current time, too; each time the instruction is executed it (potentially) generates a fresh value.

kripken: Interesting. Yes, I can see how that works here. And for random numbers, again, it is generating a new value...

I don't have a strong preference myself either way, I think. So happy to go with yours if you feel strongly, and depending on what @aheejin thinks?

tlively: I would say I feel medium-strongly about it :) Also would be happy to hear what @aheejin thinks.

aheejin: I think "generative" sounds good, but I'm OK with either. But even if we pick "generative", it'd be better to have clarifying comments on what "generative" means here and the difference between things like NaN and these.

PR #4080, opened 2021-08-13
Support nominal typing in wasm-reduce

kripken:
1. Use ToolOptions there, which adds `--nominal` support.
2. We must also pass `--nominal` to the sub-commands we run.


aheejin (approved):
LGTM with fixing the help text failure in CI

PR #4084, opened 2021-08-16
[Wasm GC] ConstantFieldPropagation: Ignore copies

kripken:
When looking for all values written to a field, we can ignore
values that are loaded from that same field, i.e., are copied from
something already present there. Such operations never introduce
new values.

This helps by a small but non-zero amount on j2cl.

### DIFF
--- a/src/passes/ConstantFieldPropagation.cpp
+++ b/src/passes/ConstantFieldPropagation.cpp
@@ -253,7 +253,8 @@ struct Scanner : public WalkerPass<PostWalker<Scanner>> {
     }
 
     // Note a write to this field of the struct.
-    noteExpression(curr->value, type.getHeapType(), curr->index, functionSetInfos);
+    noteExpression(
+      curr->value, type.getHeapType(), curr->index, functionSetInfos);
   }
 
 private:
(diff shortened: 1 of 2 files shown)


tlively (commented):
I'm not sure this is sound. In pseudocode:

```
struct $A i32
struct $B i32 extends $A

func
  $a = new $A 42
  $b = new $B 1337
  $a.0 = (($A)$b).0
  return $a.0
```

This will conclude that field 0 of $A is always 42, but actually in this instance it has been set to 1337 because the reference argument to the copy was dynamically a subtype.

kripken:
You're right that this not as simple as the description claims - I should extend that. But I believe it is safe. In your example, it isn't
```
  $a.0 = (($A)$b).0
```
that is critical, it is
```
  $b = new $B 1337
```
That line already prevents us from optimizing anyhow. Because then we cannot know what is in that field - we might point to an instance of the subtype. In full,
```
struct $A i32
struct $B i32 extends $A

func
  $a = new $A 42
  $b = new $B 1337
  $a = ... ;; assign something of type $A or $B here
  return $a.0
```

That is, if there is already a problem with a subtype, the subtype analysis will figure that out. The mere presence of a copy of the field to itself would not create a new problem. This is because the analysis is global on the types - we don't follow code paths of the life cycles of individual objects.

tlively (approved):
Ah, that makes sense. LGTM!

PR #4085, opened 2021-08-17
LocalCSE: ignore traps

kripken:
If we replace
```
A
A
A
```
with
```
(local.set A)
(local.get)
(local.get)
```
then it is ok for A to trap (so long as it does so deterministically), as if
it does trap then the first appearance will do so, and the others not
be reached anyhow.

This helps GC code as often there are repeated `struct.gets` and such that
may trap.

### DIFF
--- a/test/lit/passes/local-cse_all-features.wast
+++ b/test/lit/passes/local-cse_all-features.wast
@@ -60,17 +60,15 @@
   ;; CHECK:      (type $A (struct (field i32)))
   (type $A (struct (field i32)))
 
+  ;; CHECK:      (type $ref|$A|_=>_none (func (param (ref $A))))
+
   ;; CHECK:      (type $B (array (mut i32)))
   (type $B (array (mut i32)))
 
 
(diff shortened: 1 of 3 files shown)


tlively (approved):

test/lit/passes/local-cse_all-features.wast
```diff
@@ -87,7 +129,7 @@
     ;; Repeated loads from a struct can be optimized.
     ;;
     ;; Note that these struct.gets cannot trap as the reference is non-nullable,
-    ;; so there are no side effects here, and we can optimize.
+    ;; so this is "easier" than the previous testcase.
```
aheejin, line 132: Isn't it the same given that we ignore traps now? "Easier" sounds like somehow this one does something different..

aheejin (approved):

test/lit/passes/local-cse_all-features.wast
kripken: I'll clarify.

PR #4088, opened 2021-08-17
[Wasm GC] Effects: Differentiate Struct and Array types

kripken:
This allows common patterns in J2CL to be optimized, where we write
to various array indices and get the values or the reference from a
struct.

It would be nice to do even better here, and look at actually specific
types, but I think we should be careful to keep the runtime constant.
That seems hard to do if we accumulate a list of types and do
`Type::isSubType` on them etc. But maybe someone has a better
idea than this PR?

### DIFF
--- a/src/ir/effects.h
+++ b/src/ir/effects.h
@@ -72,10 +72,12 @@ public:
   std::set<Name> globalsWritten;
   bool readsMemory = false;
   bool writesMemory = false;
-  // TODO: Type-based alias analysis. For example, writes to Arrays never
-  // interfere with reads from Structs.
-  bool readsHeap = false;
-  bool writesHeap = false;
+  // TODO: More specific type-based alias analysis, and not just at the
+  //       struct/array level.
+  bool readsStruct = false;
+  bool writesStruct = false;
+  bool readsArray = false;
+  bool writesArray = false;
   // A trap, either from an unreachable instruction, or from an implicit trap
   // that we do not ignore (see below).
   //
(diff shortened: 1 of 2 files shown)


aheejin (approved):
This looks like a simple enough addition and wouldn't incur much more overhead to `Effects` computation, so LGTM.

About more fine-grained analysis, maybe we can consider different levels of effects computation, with default set to the current (basic) level, because most of optimization passes don't need that..? Or for efficacy we can maybe add a framework to cache effects in case the expression hasn't changed but we need the result again.

aheejin:
Also when we compute effects for an expression, we will traverse and recompute effects for all its children, right? Would there be a way to reuse children's effects when computing one? Surely it needs some updating methods in case its children changes though (like `ReFinalize`)

kripken:
@aheejin 

> maybe we can consider different levels of effects computation

Oh, good point - yeah, we have the pass options already in this class (I believe you did that work actually) so we could use them. That's worth exploring.

> Would there be a way to reuse children's effects when computing one?

It is possible to do that, by calling visit instead of walk, and on the parent doing `mergeIn(childEffects)` on each child. I believe a few passes do that.

tlively:
Would it be worth creating a new kind of `EffectsWalker` that efficiently tracks effects for the current expression? That seems less error prone than manually merging the effects as part of the pass implementation.

tlively (approved):
This LGTM as well. For more precise analysis, one possibility would be to maintain a small constant (but configurable?) number of root types. If the number of observed read or written types exceeds that small constant, we would fall back to the less granular analysis.

kripken:
> Would it be worth creating a new kind of EffectsWalker that efficiently tracks effects for the current expression?

Possibly, but it looks like the two places that use `mergeIn` actually do custom processing (along Block children) and don't even use a Walker. But maybe we could find a better helper for those two, although they are very stable at this point.

kripken:
Ok, let's land this and look into other improvements later. Good ideas here for options in that direction.

PR #4091, opened 2021-08-18
Optimize LocalCSE hash computations using a stack. NFC

kripken:
Before, we'd compute the hash of a child, then store that in a map, then
the parent would find the child's hash in the map using the pointer to the
child. But as we do a simple postorder walk, we can use a stack, and
avoid hashing the child pointers.

This makes it 10% faster or so.


aheejin (approved):
Nice!

PR #4095, opened 2021-08-19
Inlining: Allow inlining heavy functions that may skip their heavy work

kripken:
Normally we only inline "light" functions (using various notions of "light"). One
pattern that shows up in J2CL are methods like

```js
function throwIfNull(x) {
  if (!x) throw "Null pointer exception";
  return x;
}
```

We *do* want to inline such things: The heavy work (throwing, in this example)
may be skipped. If it is skipped, then all that is left is trivial, and so getting rid of
the call overhead will be noticeable compared to that trivial work.

This also adjusts the cost values of calls, which were ridiculously low for some
reason (likely that no pass cared about the cost of calls before).


test/lit/passes/inlining-optimizing_optimize-level=3.wast
```diff
@@ -3675,25 +3731,6 @@
    (local.get $0)
   )
  )
- ;; CHECK:      (func $___syscall_ret (param $0 i32) (result i32)
```
kripken, line 3678: This inlined&removed function happened to match this pattern, in the existing test suite.

src/passes/Inlining.cpp
```diff
@@ -158,10 +184,56 @@ struct FunctionInfoScanner
...
+    Index avoidableWork = 0;
+    Index unavoidableWork = 0;
+
+    auto process = [&](Expression* child) {
+      if (auto* iff = child->dynCast<If>()) {
+        // The condition cannot be avoided.
+        unavoidableWork += CostAnalyzer(iff->condition).cost;
+
+        // Only one arm will be taken.
+        Index ifTrue = CostAnalyzer(iff->ifTrue).cost;
```
tlively, line 205: It seems strange that we are using `Index` for costs. Can we use `uint32_t` or similar instead?

src/passes/Inlining.cpp
```diff
@@ -62,6 +64,22 @@ struct FunctionInfo {
...
+    // knowledge about the parameters passed to them - we should do more there
+    // TODO). However, some functions may skip their heavy work, such as
+    //
+    // function foo(x) {
+    //   if (!x) return;
+    //   heavyWork();
+    // }
+    //
+    // That function *is* worth inlining, as the heavy work is conditional, and
+    // if in practice it is avoided, then avoiding the call overhead there can
```
tlively, line 80: IIUC, there is some risk here for situations where the heavy work does happen in practice. Are we concerned about this at all?

src/passes/Inlining.cpp
```diff
@@ -158,10 +184,56 @@ struct FunctionInfoScanner
...
+    // Limit the unavoidable work to really trivial things, like say one or two
+    // eqz, ref.is_null, etc. operations.
+    const Index MaxUnavoidable = 2;
+
+    // To count as avoiding heavy work, the avoidable work must be pretty
+    // significant - something like a call, throw, etc. This should be something
+    // like an order of magnitude higher than MaxUnavoidable.
+    const Index MinAvoidable = 20;
+
+    return unavoidableWork <= MaxUnavoidable && avoidableWork >= MinAvoidable;
```
tlively, line 235: If the avoidable work is huge, it probably isn't worth the code size hit of inlining. Should we have an upper bound on the avoidable work as well?

src/passes/Inlining.cpp
kripken: Hmm, yeah, this is indeed odd as it isn't proportionate to an index in a wasm module. Let me do that as a followup.

src/passes/Inlining.cpp
kripken: Risk in terms of speed? That should be low, I think. The main risk here is code size cost, which we at least know up front. The potential speedup is something we hope for, but that's about it.

I think I think this is a plausible rule, and it does look for very very minimal unavoidable work - almost zero - so it should not apply excessively. But we could make it even more specific if we look for some kind of indication that the skippable code is unlikely. If we had branch-hinting that would be good; otherwise, perhaps throwing an exception is such a hint.

src/passes/Inlining.cpp
kripken: This does not supersede the usual code size limits. This just avoids us dismissing an opportunity just because of seeing a call or a loop or something else heavy.

I was thinking that's enough, but now that you mention it maybe it's not. We could make this rule only apply when not optimizing for size - how does that sound?

tlively: Ah that makes sense. Yes, using this only when not optimizing for size sounds right to me.

src/passes/Inlining.cpp
tlively: Right, I was thinking of code size risk. But you explained that code size is tracked separately, so this seems fine to me. I agree that there are good heuristics we could use to guess at branch probability like throwing or trapping. I would expect us to add those as part of some future work to support branch-hinting.

tlively (approved):

src/passes/Inlining.cpp
kripken: Actually, it turns out that was already the case - just the code was a little unclear. I reordered it and added more comments.

tlively (approved):

kripken:
Closing this - it is not enough, as we do not want to inline the entire function here - just the condition. I'll open a PR with "partial" inlining after current open PRs are done.

PR #4096, opened 2021-08-20
ConstantFieldPropagation: Infer specific types

kripken:
This aims to optimize the following patterns:
```cpp
Parent* x = new Child();
x.foo();

Parent* y = new Parent();
y.foo();
```

In both cases we can see that the local refers to either a Parent or a Child
instance, and so we do not need to assume that it is Parent or any subtype
as the local variable suggests. (And, in the first case, we should infer that
it is Child specifically and not any subtypes of Child if they exist.)

To achieve this, this does an analysis to look where values arrive from
via locals, and where we can trace things back to a `struct.new` we can
infer a specific type.


kripken:
(added a few lines to also support inferring from `global.get` where the global is immutable, which appears in priactice on J2CL)

kripken:
Let's pause this PR for now. The benefit on J2CL is minor - 7 gets in the huge entire program. We need to figure out why that is, which might be due the actual pattern being different than this. (This might be useful for other languages, of course, but until we see a benefit that justifies the complexity we should wait.)

PR #4097, opened 2021-08-20
OptimizeInstructions: Handle trivial ref.cast and ref.test

kripken:
If the types are completely incompatible, we know the cast will fail. However,
ref.cast does allow a null to pass through, which makes it a little more
complicated.


src/passes/OptimizeInstructions.cpp
```diff
@@ -1263,41 +1277,78 @@ struct OptimizeInstructions
...
+      // reference's type.
       if (HeapType::isSubType(curr->ref->type.getHeapType(),
-                              curr->rtt->type.getHeapType()) &&
-          canReorder(curr->ref, curr->rtt)) {
-        Builder builder(*getModule());
-        replaceCurrent(
-          builder.makeSequence(builder.makeDrop(curr->rtt), curr->ref));
+                              curr->rtt->type.getHeapType())) {
+        Reorderer reorderer(
+          curr->ref, curr->rtt, getFunction(), getModule(), passOptions);
```
aheejin, line 1349: Do we not need to check if they can be reordered effect-wise?

test/lit/passes/inlining-optimizing.wast
```diff
@@ -15,10 +14,7 @@
  ;; CHECK:      (func $1
  ;; CHECK-NEXT:  (drop
  ;; CHECK-NEXT:   (call_ref
- ;; CHECK-NEXT:    (ref.cast
- ;; CHECK-NEXT:     (ref.func $0)
- ;; CHECK-NEXT:     (rtt.canon $none_=>_i32)
- ;; CHECK-NEXT:    )
+ ;; CHECK-NEXT:    (unreachable)
```
aheejin, line 17: Do we need to change something in the comment below?

test/lit/passes/optimize-instructions-call_ref.wast
```diff
@@ -143,9 +143,14 @@
...
- ;; CHECK-NEXT:    (ref.func $return-nothing)
- ;; CHECK-NEXT:    (rtt.canon $none_=>_i32)
+ ;; CHECK-NEXT:   (block
+ ;; CHECK-NEXT:    (drop
+ ;; CHECK-NEXT:     (ref.func $return-nothing)
+ ;; CHECK-NEXT:    )
+ ;; CHECK-NEXT:    (drop
+ ;; CHECK-NEXT:     (rtt.canon $none_=>_i32)
+ ;; CHECK-NEXT:    )
+ ;; CHECK-NEXT:    (unreachable)
```
aheejin, line 153: Here too; does the comment need changing?

aheejin (commented):
Haven't finished reading the tests; some questions:

src/ir/reorderer.h
```diff
@@ -0,0 +1,69 @@
...
+namespace wasm {
+
+//
+// Reorder two concrete expressions, using a local if necessary. Given
+//
+//  (first, second)
+//
+// we generate something so that it is valid to write
+//
+//  (Reorderer.second, Reorderer.first)
```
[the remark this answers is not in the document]
kripken: Interesting, it's the opposite for me :smile: 

I think about it as "maybe the Reorderer will do nothing, and so `Reorderer.second = second`, which means the user wants to put that first when they reorder".

How about `(Reorderer.reorderedFirst, Reorderer.reorderedSecond)`? But that is still ambiguous...

src/passes/OptimizeInstructions.cpp
```diff
@@ -1263,41 +1277,78 @@ struct OptimizeInstructions
...
+    // would change the type, which can happen in e.g.
+    //   (ref.cast (ref.as_non_null (.. (ref.null)
+    if (fallthrough->is<RefNull>()) {
+      Expression* rep =
+        builder.makeBlock({builder.makeDrop(curr->ref),
+                           builder.makeDrop(curr->rtt),
+                           builder.makeRefNull(curr->rtt->type.getHeapType())});
+      if (curr->ref->type.isNonNullable()) {
+        // Avoid a type change by forcing to be non-nullable. In practice, this
+        // would have trapped before we get here, so this is just for
```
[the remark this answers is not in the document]
kripken: We receive a null value from a place that has a non-nullable type. Somehow the null falls through there, which I think is only possible using `ref.as_non_null` which will trap on a null.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1263,41 +1277,78 @@ struct OptimizeInstructions
...
+
+    // For the cast to be able to succeed, the value being cast must be a
+    // subtype of the desired type, as RTT subtyping is a subset of static
+    // subtyping. For example, trying to cast an array to a struct would be
+    // incompatible.
+    if (!canBeCastTo(curr->ref->type.getHeapType(),
+                     curr->rtt->type.getHeapType())) {
+      // This cast cannot succeed. It will either trap if the input is not a
+      // null, or it will return null if it is (and we have already handled the
+      // case of null before).
```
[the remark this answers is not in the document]
kripken: Good point, yeah, this is unclear. I'll rephrase.

src/passes/OptimizeInstructions.cpp
kripken: That is what the Reorderer does - it potentially adds more code to handle that problem (it adds a local and saves the value here, keeping their order identical).

test/lit/passes/optimize-instructions-call_ref.wast
kripken: Sounds good, thanks, I'm updating these comments too.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1263,41 +1277,78 @@ struct OptimizeInstructions
...
-      // We also do not do this if the arguments cannot be reordered. If we
-      // can't do that then we need to add a drop, at minimum (which may still
-      // be worthwhile, but depends on other optimizations kicking in, so it's
-      // not clearly worthwhile).
+      // Aside from the issue of type incompatibility as mentioned above, the
+      // cast can trap if the types *are* compatible but it happens to be the
+      // case at runtime that the value is not of the desired subtype. If we
+      // do not consider such traps possible, we can ignore that. Note, though,
+      // that we cannot do this if we cannot replace the current type with the
+      // reference's type.
```
aheejin, line 1345: Are traps possible even if `ref`'s heap type is a subtype of `rtt`'s heap type?

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -14,22 +14,27 @@
     (field $i64 (mut i64))
   ))
 
+  ;; CHECK:      (type $array (array (mut i8)))
+
+  ;; CHECK:      (type $B (struct (field i32) (field i32) (field f32)))
+
```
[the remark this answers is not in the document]
kripken: I think this is just because more uses were added, so it alters the order in which the types are emitted (more frequently used ones appear first).

src/passes/OptimizeInstructions.cpp
```diff
@@ -212,20 +213,33 @@ struct OptimizeInstructions
...
+      ReFinalize().walkFunctionInModule(func, getModule());
+    }
+
+    // Final optimizations.
     {
       FinalOptimizer optimizer(getPassOptions());
       optimizer.walkFunction(func);
     }
+
+    TypeUpdating::handleNonDefaultableLocals(func, *getModule());
```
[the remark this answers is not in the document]
kripken: Good point, adding a comment.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1263,41 +1277,78 @@ struct OptimizeInstructions
...
     }
 
+    Builder builder(*getModule());
     auto passOptions = getPassOptions();
 
+    auto fallthrough = Properties::getFallthrough(
+      curr->ref, getPassOptions(), getModule()->features);
+
+    // If the value is a null, it will just flow through, and we do not need the
+    // cast. However, if that would change the type, then things are less
```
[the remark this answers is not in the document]
kripken: The code here is actually careful about the heap type, I'll improve the comment though.

Basically, we use the heap type of the RTT to create the null - that is the type on the outside.

src/passes/OptimizeInstructions.cpp
kripken: Yes, this is an unfortunate property of rtts. E.g. `(rtt.canon $x) != (rtt.sub $x (rtt.canon $x))`.

This is bad for optimization, so we are hoping to prototype a more static approach soon, that would be as you describe basically.

kripken:
I think your intuition about the Reorderer is right, the more I think about it. I rewrote it to a method `getResultOfFirst` that I think is unambiguous, PTAL.

src/passes/OptimizeInstructions.cpp
aheejin: > `(rtt.canon $x) != (rtt.sub $x (rtt.canon $x))`

😨

aheejin (approved):

PR #4100, opened 2021-08-23
Dominator Tree

kripken:
Add a class to compute the dominator tree for a CFG consisting
of a list of basic blocks assumed to be in reverse postorder.

This will be useful once cfg-walker emits blocks in reverse-postorder
(which it almost does, another PR will handle that). Then we can write
optimization passes that use block dominance.


src/cfg/domtree.h
```diff
@@ -0,0 +1,157 @@
...
+// dominator) in the tree, that is,
+//
+//  parents[0] = a nonsense value, as the entry node has no immediate dominator
+//  parents[1] = the index of the immediate dominator of CFG.blocks[1]
+//  parents[2] = the index of the immediate dominator of CFG.blocks[2]
+//  etc.
+//
+// The BasicBlock type is assumed to have a ".in" property which declares a
+// vector of pointers to the incoming blocks, that is, the predecessors.
+template<typename BasicBlock> struct DomTree {
```
aheejin, line 45: Why is this a template class? Do you plan to use this on other than `BasicBlock`? Oh, I noticed you defined another `BasicBlock` class in test/domtree.cpp, but is that the reason? That test file can include a Binaryen header that defines `BasicBlock` I guess, and if this is not a template class, we can move the algorithm to a cpp file.

src/cfg/domtree.h
```diff
@@ -0,0 +1,157 @@
...
+      // of them.
+      auto& preds = blocks[index]->in;
+      Index newParent = nonsense;
+      for (auto* pred : preds) {
+        auto predIndex = blockIndices[pred];
+        if (parents[predIndex] == nonsense) {
+          // This pred has yet to be processed; we'll get to it in a later
+          // cycle.
+          continue;
+        }
```
aheejin, line 114: This only happens for loop backedges, which should be dominated by the current block (loop header) in reducible CFGs, and thus don't affect the idom computation. I guess this can also happen for unreachable blocks but I'm not sure that can be created from a Binaryen AST? Anyway they don't affect idom computation either. Then do we need to "get to it in a later cycle"?

(I think, if we don't consider unreachable blocks, we can even simplify the loop to
```cpp
if (predIndex > index) {
  continue;
}
```
)

Also, do we need the outer `while (changed)` in that sense? If we don't consider irreducible CFGs, it looks the process can be completed with a single traversal of `blocks` without the `while` loop. It wouldn't be able to handle the irreducible CFG in your last example, but we don't need to do that for Binaryen anyway. If we have this `while (changed)`, every `DomTree` construction will traverse the `blocks` twice instead of once, even for reducible CFGs, because `changed` will be set to true for every CFG in its first traversal.

src/cfg/domtree.h
```diff
@@ -0,0 +1,157 @@
...
+          }
+          while (right > left) {
+            right = parents[right];
+          }
+        }
+        newParent = left;
+      }
+
+      // We may have found a new value here.
+      if (newParent != parents[index]) {
```
aheejin, line 140: If we remove the outer `while`, we don't need this `if` either, because we only traverse each BB once.

test/example/domtree.cpp
```diff
@@ -0,0 +1,180 @@
...
+  // entry and a non-connected (unreachable) block.
+  {
+    CFG cfg;
+    auto* entry = cfg.add();
+    auto* a = cfg.add();
+
+    DomTree<BasicBlock> domTree(cfg);
+    assert(domTree.parents.size() == 2);
+    assert(domTree.parents[0] == Index(-1));
+    assert(domTree.parents[0] == Index(-1)); // unreachables have no parent.
```
aheejin, line 65:
```suggestion
    assert(domTree.parents[1] == Index(-1)); // unreachables have no parent.
```

src/cfg/domtree.h
kripken: This is because CFGWalker is a template itself, and the BasicBlock type inside it depends on the template parameter. So there isn't a single type here.

There might be a clever C++ way to solve this, like moving BasicBlock out of CFGWalker and subclassing it inside to add the data with the templated type, but I can't see a simple way to do that which works...

src/cfg/domtree.h
```diff
@@ -0,0 +1,157 @@
...
+//
+//  parents[0] = a nonsense value, as the entry node has no immediate dominator
+//  parents[1] = the index of the immediate dominator of CFG.blocks[1]
+//  parents[2] = the index of the immediate dominator of CFG.blocks[2]
+//  etc.
+//
+// The BasicBlock type is assumed to have a ".in" property which declares a
+// vector of pointers to the incoming blocks, that is, the predecessors.
+template<typename BasicBlock> struct DomTree {
+  std::vector<Index> parents;
```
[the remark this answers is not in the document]
kripken: Makes sense, renaming to `iDoms`.

src/cfg/domtree.h
kripken: Good points! Yes, we may as well simplify things given reducibility. I was focused on implementing the algorithm as is, but we really don't need that.

src/cfg/domtree.h
kripken: True. I'll update the comment, but I think there is still a use for the if, to be able to assert that no changes happen after the first pass (see updated code).

src/cfg/domtree.h
aheejin: Why should `BasicBlock` be inside `CFGWalker`?

src/cfg/domtree.h
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    return changed;
+  };
+
+  processBlocks();
+
+  // We must have finished all the work in a single traversal, since our input
+  // is reducible.
+  assert(!processBlocks());
```
aheejin, line 169: Do we ship release builds with assertions or without? If we ship with them, isn't this gonna run the process twice unnecessarily?

src/cfg/domtree.h
```diff
@@ -0,0 +1,178 @@
...
+        if (predIndex > index) {
+          continue;
+        }
+
+        // All of our predecessors will have been processed before us, except
+        // if they are unreachable from the entry, in which case, we can ignore
+        // them.
+        if (iDoms[predIndex] == nonsense) {
+          continue;
+        }
```
aheejin, line 125: Are unreachable blocks possible to be generated from Binaryen AST?

src/cfg/domtree.h
kripken: Currently CFGWalker creates and owns a list of BasicBlocks. Subclasses of CFGWalker add more fields to BasicBlock as needed. Maybe this could all be restructured, but it would probably need more dynamic checks I think (right now it uses CRTP).

src/cfg/domtree.h
kripken: By default we build with assertions, but I think @dschuff was looking into disabling that in LTO builds in the emsdk, or maybe we already do?

src/cfg/domtree.h
kripken: Yes, e.g.
```wat
(unreachable)
(nop)
(if
 ..
..
)
(nop)
```
The basic block after the unreachable has no predecessor. And the diamond pattern after it are also effectively unreachable, and they are the reason for that commit + test.

CFGWalker atm does not try to avoid creating unreachable blocks (I guess it could, but it would need to look at all predecessors etc.).

src/cfg/domtree.h
aheejin: Maybe we can do
```cpp
BYN_DEBUG(!processBlocks());
```
using [BYN_DEBUG](https://github.com/WebAssembly/binaryen/blob/a2323f2cfd90089c54100ab98c439b9438cc4dc1/src/support/debug.h#L58)? It only runs with debug mode. Disabling assertions altogether may have some consequences, but running the whole process once again may be too high a cost here..

kripken: Interesting idea. However, I think we rarely build with `BYN_DEBUG` though - we don't test it by default anywhere, do we? And the overhead here is quite small, as this will run in 2 passes instead of 1, but the original algorithm would take potentially many more passes, but is still considered one of the fastest in this area. That feels like a reasonable amount of overhead to me in a build with assertions.

aheejin (approved):

PR #4101, opened 2021-08-23
Ensure cfg-traversal emits blocks in reverse postorder, refactoring try. NFC

kripken:
Reverse postorder basically just means that a block's immediate dominator
must precede it in the list. That is useful because then algorithms that look at
dominance can simply process the list in order, and the immediate dominator
will have already been seen before each block.

Another way to put it is that in reverse postorder a block's dominators
appear before it in the list, as do all non-loop predecessors. At least in
reducible graphs that is the case, and our IR, like wasm, is reducible.

It is pretty natural to emit reverse postorder on wasm given the
reducibility: simply process the wasm in postorder, and make sure to
create new basic blocks only when reaching their code - that is, do not
create them "ahead of time". We were doing that in a single place, for
try-catch, so this PR refactors that. Specifically it makes us create the
basic blocks for catches right when we reach them, and not earlier.
So the data structure that used to store them becomes a list of things
to connect to them.

This is useful for #4100 , see more details there.

### DIFF
--- a/src/cfg/cfg-traversal.h
+++ b/src/cfg/cfg-traversal.h
@@ -96,14 +96,14 @@ struct CFGWalker : public ControlFlowWalker<SubType, VisitorType> {
 
   // stack of the last blocks of try bodies
   std::vector<BasicBlock*> tryStack;
-  // Stack of the blocks that can reach the first blocks of catches that
-  // throwing instructions should unwind to at any moment. That is, the topmost
-  // item in this vector relates to the current try-catch scope, and the vector
-  // there is a list of the items that can reach catch blocks (each item is
-  // assumed to be able to reach any of the catches, although that could be
-  // improved perhaps).
-  std::vector<std::vector<BasicBlock*>> unwindCatchStack;
-  // stack of 'Try' expressions corresponding to unwindCatchStack.
+  // Stack of the blocks that contain a throwing instruction, and therefore they
+  // can reach the first blocks of catches that throwing instructions should
+  // unwind to at any moment. That is, the topmost item in this vector relates
+  // to the current try-catch scope, and the vector there is a list of the items
+  // that can reach catch blocks (each item is assumed to be able to reach any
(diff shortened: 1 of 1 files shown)


src/cfg/cfg-traversal.h
```diff
@@ -279,19 +288,32 @@ struct CFGWalker : public ControlFlowWalker<SubType, VisitorType> {
...
   }
 
   static void doStartCatches(SubType* self, Expression** currp) {
     self->tryStack.push_back(self->currBasicBlock); // last block of try body
-    self->processCatchStack.push_back(self->unwindCatchStack.back());
+
+    // Now that we are starting the catches, create the basic blocks that they
+    // begin with.
+    auto* last = self->currBasicBlock;
+    auto* tryy = (*currp)->dynCast<Try>();
```
aheejin, line 301:
```suggestion
    auto* tryy = (*currp)->cast<Try>();
```

src/cfg/cfg-traversal.h
```diff
@@ -88,9 +96,12 @@ struct CFGWalker : public ControlFlowWalker<SubType, VisitorType> {
...
   std::vector<BasicBlock*> tryStack;
-  // stack of the first blocks of catches that throwing instructions should
-  // unwind to at any moment. Because there can be multiple catch blocks for a
-  // try, we maintain a vector of first blocks of catches.
+  // Stack of the blocks that can reach the first blocks of catches that
+  // throwing instructions should unwind to at any moment. That is, the topmost
+  // item in this vector relates to the current try-catch scope, and the vector
+  // there is a list of the items that can reach catch blocks (each item is
+  // assumed to be able to reach any of the catches, although that could be
+  // improved perhaps).
```
aheejin, line 104: How about changing the name to `unwindingInstsStack` or `throwingInstsStack` or something? This doesn't contain catch blocks anymore.

aheejin (approved):

PR #4109, opened 2021-08-27
Support specialized function types in element segments

kripken:
Before this, the element segments would be printed as having type
`funcref`, and then if their table had a specialized type, the element
type would not be a subtype of the table and validation would fail.

To fix this I had to rework several things, and I am not that familiar
with this code. I suspect I am missing something regarding the
`usesExpressions` logic - is that a workaround for a limitation of
ours that can be removed (as I did here), or not?


src/passes/Print.cpp
```diff
@@ -2740,7 +2740,15 @@ struct PrintSExpression : public UnifiedExpressionVisitor<PrintSExpression> {
...
-      if (allElementsRefFunc) {
+      // If all elements are ref.func, then we may be a simple MVP table. We
+      // populate such tables with ref.func internally, but we print them as
+      // having type "func" and not a specialized type. However, if we have a
+      // table with a specialized type then we must emit that type (as our type
+      // must be a subtype of that type).
+      bool hasTableOfSpecializedType =
+        curr->table.is() &&
+        currModule->getTable(curr->table)->type != Type::funcref;
+      if (allElementsRefFunc && !hasTableOfSpecializedType) {
```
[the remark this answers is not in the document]
kripken: By "this logic" do you mean `hasTableOfSpecializedType`, or `allElementsRefFunc && !hasTableOfSpecializedType`?

kripken:
I'm not that familiar with the spec here, sorry - which spec added the new style? And is "uses expressions" the terminology used there? (Searching for that fails, but I assume "expressions" is specific to binaryen - what is the spec terminology, do you know?)

src/passes/Print.cpp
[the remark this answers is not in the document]
tlively: I mean `allElementsRefFunc && !hasTableOfSpecializedType`.

tlively:
The relevant section of the spec is https://webassembly.github.io/spec/core/binary/modules.html#binary-elemsec, but it's a little hard to read. It was added as part of reference-types, but I'm not seeing a better exposition in that repo.

Note that some elem declarations use `y∗:vec(funcidx)` and some use `el∗:vec(expr)`. In the note below the grammar this is pointed out as "bit 2 indicates the use of element type and element expressions instead of element kind and element indices." MVP-style segments always use the former format and the new-style format is the latter.

In binaryen we represent segments uniformly as lists of expressions, but if all the expressions are ref.func and there is no more specific segment type, then we should emit the segment using the MVP-style format because it is more compact.

kripken:
Thanks!

kripken:
Ok, pushed an update with what I think is correct.

This uncovered a test issue, for which I had to add `BinaryenUpdateMaps` to the C API. This was already broken before - we have to update maps after renaming a module element, which that test does - but the test didn't actually use the maps. This new code however does. I can open a separate PR with that change, of course, but first I wanted to see that it looks good as a whole here both in review and tests.

tlively (commented):
The code structure looks a lot better now, thanks!

Would it be reasonable to add to the existing multi-table.wast test rather than introduce a new one?

src/wasm-binary.h
```diff
@@ -346,9 +346,10 @@ enum SegmentFlag {
   // Bit 1 if active: 0 = index 0, 1 = index given
   HasIndex = 1 << 1,
   // Table element segments only:
-  // Bit 2: 0 = elemType is funcref and vector of func indexes given
+  // Bit 2: 0 = elemType is funcref and vector of func indexes given (that is,
+  //            this is MVP)
   //        1 = elemType is given and vector of ref expressions is given
-  UsesExpressions = 1 << 2
+  PostMVP = 1 << 2
```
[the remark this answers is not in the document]
kripken: Ah, thanks! This is an area of the spec I really should know more about.

kripken:
Sounds good, test moved to add to the existing one, that is indeed better.

src/wasm-binary.h
```diff
@@ -346,9 +346,10 @@ enum SegmentFlag {
   // Bit 1 if active: 0 = index 0, 1 = index given
   HasIndex = 1 << 1,
   // Table element segments only:
-  // Bit 2: 0 = elemType is funcref and vector of func indexes given
+  // Bit 2: 0 = elemType is funcref and vector of func indexes given (that is,
+  //            this is MVP)
```
tlively, line 350: I would undo this change because some non-MVP segments also set this bit to 0.

src/wasm-binary.h
```diff
@@ -346,9 +346,10 @@ enum SegmentFlag {
...
   //        1 = elemType is given and vector of ref expressions is given
-  UsesExpressions = 1 << 2
+  usesExpressions = 1 << 2
```
tlively, line 352:
```suggestion
  UsesExpressions = 1 << 2
```

The rest of the enum values are PascalCase, so we should match them (or change them all) for consistency.

src/wasm/wasm-binary.cpp
```diff
@@ -2864,12 +2861,12 @@ void WasmBinaryBuilder::readElementSegments() {
       if (usesExpressions) {
         segment->type = getType();
         if (!segment->type.isFunction()) {
-          throwError("Invalid type for an element segment");
+          throwError("Invalid type for a post-MVP element segment");
```
tlively, line 2864: and also invalid for MVP element segments!

src/wasm/wasm-binary.cpp
```diff
@@ -2864,12 +2861,12 @@ void WasmBinaryBuilder::readElementSegments() {
...
         }
       } else {
         auto elemKind = getU32LEB();
         if (elemKind != 0x0) {
-          throwError("Only funcref elem kinds are valid.");
+          throwError("Only funcref elem kinds are valid in the MVP.");
```
tlively, line 2869: Maybe "The only allowed elemKind is funcref"? This field also appears in non-MVP segment formats.

test/multi-table.wast
```diff
@@ -25,7 +28,11 @@
   (elem $empty func)
   (elem $declarative declare func $h)
 
+  ;; This elem will be emitted as usesExpressions because of the type of the
+  ;; table.
+  (elem $especial (table $tspecial) (i32.const 0) (ref null $none_=>_none) $f $h)
```
tlively, line 33: Hmm, I don't think we should be able to parse this format that uses function names and a specific type. The only allowable use of function indices or names is with "func" and it's always short for "funcref"

https://webassembly.github.io/spec/core/text/modules.html#:~:text=also%2C%20the%20element%20list%20may%20be%20written%20as%20just%20a%20sequence%20of%20function%20indices%3A

kripken: Hmm, you might be right, based on that spec fragment. However, the reference spec interpreter supports a bunch of extra sugar - is all that documented in the spec?

Separately, I'm not sure there is harm in reading this, since it actually simplifies our code to do so (a single loop) and we aren't really in the business of parsing the spec format with 100% precision anyhow.

kripken:
(All comments aside from the last should be addressed.)

tlively (approved):
Looks good!

kripken:
I agree in the long term it would be good to fix this in the spec - to precisely define all the relevant text sugar that is allowed. And maybe a binaryen text parser rewrite/replacement could make us properly conform - that might be a requirement for such a project, actually.

PR #4110, opened 2021-08-27
Add a test for too many locals in Asyncify

kripken:
Followup to #4108


dschuff (approved):
yeah this seems fine to me too.

PR #4111, opened 2021-08-27
Optimize away dominated calls to functions that run only once

kripken:
Some functions run only once with this pattern:

```js
function foo() {
  if (foo$ran) return;
  foo$ran = 1;
  ...
}
```

If that global is not ever set to 0, then the function's payload (after the
initial if and return) will never execute more than once. That means we
can optimize away dominated calls:

```js
foo();
foo(); // we can remove this
```

To do this, we find which globals are "once", which means they can
fit in that pattern, as they are never set to 0. If a function looks like the
above pattern, and it's global is "once", then the function is "once" as
well, and we can perform this optimization.

This removes over 8% of static calls in j2cl.


src/passes/OnceReduction.cpp
```diff
@@ -0,0 +1,414 @@
...
+
+    for (Index i = 0; i < basicBlocks.size(); i++) {
+      auto* block = basicBlocks[i].get();
+
+      // Note that we take a reference here, which is how the data we accumulate
+      // ends up stored. The blocks we dominate will see it later.
+      auto& onceGlobalsWritten = onceGlobalsWrittenVec[i];
+
+      // Note information from our immediate dominator.
+      // TODO: we could also intersect information from all of our preds.
```
tlively, line 244: Repeatedly taking the intersection of all the preds until a fixed point is reached is how a classic dataflow analysis works, right? Are there other places where we have to do similar analyses? I wonder if it makes sense to have a generic utility for it.

src/passes/OnceReduction.cpp
```diff
@@ -0,0 +1,414 @@
...
+      // For a global to possibly be "once", it must be initialized to a
+      // constant. As we scan code we will turn this into false if we see
+      // anything that proves the global is not "once".
+      //   * Note that we don't check that the constant is zero - that is fine
+      //     for us to optimize, though it does indicate that the once function
+      //     will never ever run, which we could optimize further. TODO
+      //   * TODO: non-integer types?
+      optInfo.onceGlobals[global->name] = global->type.isInteger() &&
+                                          !global->imported() &&
+                                          global->init->is<Const>();
```
tlively, line 336: Can't we further limit this to globals that are initialized to zero?

src/passes/OnceReduction.cpp
```diff
@@ -0,0 +1,414 @@
...
+      if (onceGlobal.is() && !optInfo.onceGlobals[onceGlobal]) {
+        onceGlobal = Name();
+      }
+    }
+
+    // Optimize using what we found. Keep iterating while we find things to
+    // optimize, which we estimate using a counter of the total number of once
+    // globals set by functions: as that increases, it means we are propagating
+    // useful information.
+    // TODO: limit # of iterations?
```
tlively, line 360: It might be faster to first propagate the onceGlobalsSetInFuncs through the call graph (which doesn't require scanning through the blocks or expressions of a function) and only then start modifying functions. I think separating these concerns would make the code easier to read as well.

tlively (commented):
I've read through the code but not yet through the tests.

src/cfg/domtree.h
```diff
@@ -47,6 +47,10 @@ namespace wasm {
 template<typename BasicBlock> struct DomTree {
   std::vector<Index> iDoms;
 
+  // Use a nonsense value to indicate what has yet to be initialized or what is
+  // irrelevant.
+  enum { nonsense = Index(-1) };
```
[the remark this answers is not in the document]
kripken: Sadly the const version runs into C++ issues:
```
/usr/bin/ld: src/passes/CMakeFiles/passes.dir/OnceReduction.cpp.o: warning: relocation against `_ZN4wasm7DomTreeINS_9CFGWalkerINS_12_GLOBAL__N_19OptimizerENS_7VisitorIS3_vEENS2_9BlockInfoEE10BasicBlockEE8nonsenseE' in read-only section `.text'
/usr/bin/ld: src/passes/CMakeFiles/passes.dir/OnceReduction.cpp.o: relocation R_X86_64_PC32 against undefined symbol `_ZN4wasm7DomTreeINS_9CFGWalkerINS_12_GLOBAL__N_19OptimizerENS_7VisitorIS3_vEENS2_9BlockInfoEE10BasicBlockEE8nonsenseE' can not be used when making a shared object; recompile with -fPIC
```
That is with this diff + a file that contains the actual def:
```diff
diff --git a/src/cfg/domtree.h b/src/cfg/domtree.h
index 5753a3114..a6f436eef 100644
--- a/src/cfg/domtree.h
+++ b/src/cfg/domtree.h
@@ -49,7 +49,7 @@ template<typename BasicBlock> struct DomTree {
 
   // Use a nonsense value to indicate what has yet to be initialized or what is
   // irrelevant.
-  enum { nonsense = Index(-1) };
+  static const Index nonsense;
 
   DomTree(std::vector<std::unique_ptr<BasicBlock>>& blocks);
 };
```

This is beyond my C++ arcana, but I suspect it's something to do with it being a template...

Anyhow, I kind of prefer the enum. There is no reason to actually store this in static memory, it's just a definition.

src/passes/OnceReduction.cpp
kripken: We might want that eventually, yeah. We don't atm have other users of precisely that (we have iterations to fixed points, but this isn't an interation, it's one pass; and liveness might be similar but it does a union and not an intersection IIRC).

src/cfg/domtree.h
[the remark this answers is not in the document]
tlively: Oh yikes. Should we be compiling with -fPIC as suggested in the error message anyway? Anyway this change seems fine for working around that error.

test/lit/passes/once-reduction.wast
```diff
@@ -0,0 +1,1352 @@
...
+      (global.get $once)
+      (return)
+    )
+    (global.set $once (i32.const 42))
+  )
+
+  ;; CHECK:      (func $caller
+  ;; CHECK-NEXT:  (call $once)
+  ;; CHECK-NEXT:  (nop)
+  ;; CHECK-NEXT:  (nop)
```
tlively, line 497: Interesting that we can optimize this `set` out as well, even if it has some special other value. I guess this is because the global is not exposed outside the module and we never read it anywhere except to check whether it is zero or nonzero?

I didn't see any tests for imported or exported `once` globals; those would be good to add. Also, it seems that we could optimize all these different non-zero constants to just `1` for a tiny code size savings (but I guess that's probably not worth it).

test/lit/passes/once-reduction.wast
```diff
@@ -0,0 +1,1352 @@
...
+  ;; CHECK-NEXT:   (call $once)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $caller
+    (drop (call $once))
+    (drop (call $once))
+  )
+)
+
+;; Corner case: "Once" function body is not a block.
```
tlively, line 746: It might also be interesting to test `func $once (if !once ...), which we could recognize as a once function but do not (probably because other optimizations will transform it into the form we do recognize?)

tlively (commented):
Tests look great!

src/passes/OnceReduction.cpp
```diff
@@ -0,0 +1,414 @@
...
+    if (numBlocks == 0) {
+      return;
+    }
+
+    // Each index in this vector is the set of "once" globals written to in the
+    // basic block with the same index.
+    std::vector<std::unordered_set<Name>> onceGlobalsWrittenVec;
+    onceGlobalsWrittenVec.resize(numBlocks);
+
+    for (Index i = 0; i < basicBlocks.size(); i++) {
```
aheejin, line 253: In case we don't switch the above to `basicBlocks.empty()`
```suggestion
    for (Index i = 0; i < numBlocks; i++) {
```

src/passes/OnceReduction.cpp
```diff
@@ -0,0 +1,414 @@
...
+    // appears in block A, then we do not need to make any calls in any blocks
+    // dominated by A.
+    DomTree<Parent::BasicBlock> domTree(basicBlocks);
+
+    // Perform the work by going through the blocks in reverse postorder and
+    // filling out which "once" globals have been written to.
+    auto numBlocks = basicBlocks.size();
+    if (numBlocks == 0) {
+      return;
+    }
```
aheejin, line 246: Maybe we can check this at the top of the function? We can just use `basicBlocks.empty()` I guess

src/passes/OnceReduction.cpp
```diff
@@ -0,0 +1,414 @@
...
+      // Note that we take a reference here, which is how the data we accumulate
+      // ends up stored. The blocks we dominate will see it later.
+      auto& onceGlobalsWritten = onceGlobalsWrittenVec[i];
+
+      // Note information from our immediate dominator.
+      // TODO: we could also intersect information from all of our preds.
+      auto parent = domTree.iDoms[i];
+      if (parent == domTree.nonsense) {
+        // This is either the entry node (which we need to process), or an
+        // unreachable block (which we do not - leave that to DCE).
```
aheejin, line 265: Is anything missing after `which we do not`..? Like, 'remove', or, 'deal with', or something like that?

src/passes/OnceReduction.cpp
kripken: Sounds good, adding a comment.

test/lit/passes/once-reduction.wast
kripken: Yes, a "once" global is effectively a boolean. So the specific non-zero value does not matter. Good point we could turn them into 1, I'll add a TODO.

Good catch about tests! An import test was indeed missing, and exports were not even checked for. Fixed.

test/lit/passes/once-reduction.wast
kripken: Yes, atm I do not see that form appear anywhere. But I added a TODO now.

src/passes/OnceReduction.cpp
```diff
@@ -339,6 +356,14 @@ struct OnceReduction : public Pass {
       // Fill in the map so that it can be operated on in parallel.
       optInfo.onceFuncs[func->name] = Name();
     }
+    for (auto& ex : module->exports) {
+      if (ex->kind == ExternalKind::Global) {
+        // An exported global cannot be "once" since the outside may read and
+        // write to it in ways we are unaware.
+        // TODO: This could perhaps only be on mutable ones.
```
aheejin, line 363: Why don't we check that now, given that it should be simple..?

src/passes/OnceReduction.cpp
```diff
@@ -0,0 +1,439 @@
...
+// we see a call, we can infer things), and when it is finishes with a function
+// it has learned which "once" globals it must set, and it then writes out
+// newOnceGlobalsSetInFuncs with that result. Later iterations will then use
+// those values in place of onceGlobalsSetInFuncs, which propagates things to
+// callers. This in effect mixes local optimization with the global
+// propagation - as we need to run the full local optimization in order to infer
+// the new values for onceGlobalsSetInFuncs, that is unavoidable (in principle,
+// we could also do a full propagation to a fixed point in between running
+// local optimization, but that would require more code - it might be more
+// efficient, though).
```
aheejin, line 208:
```suggestion
// we see a call, we can infer things), and when it finishes with a function it
// has learned which "once" globals it must set, and it then writes out
// newOnceGlobalsSetInFuncs with that result. Later iterations will then use
// those values in place of onceGlobalsSetInFuncs, which propagate things to
// callers. This in effect mixes local optimization with the global propagation
// - as we need to run the full local optimization in order to infer the new
// values for onceGlobalsSetInFuncs, that is unavoidable (in principle, we could
// also do a full propagation to a fixed point in between running local
// optimization, but that would require more code - it might be more efficient,
// though).
```
(I only deleted the stray `is` and the rest is 80-col wrapping)

> as we need to run the full local optimization in order to infer the new values for onceGlobalsSetInFuncs, that is unavoidable

The 'optimization' is the part we convert operations with nops, and I don't think we have to run that to infer the new values..? (I've seen the comments and I understand you want to mix them; here I'm mostly confused about the comments)

> it might be more efficient, though.

Why?

aheejin (commented):
Haven't read the test yet but some questions:

tlively (approved):
LGTM :tada:

test/lit/passes/once-reduction.wast
```diff
@@ -0,0 +1,1425 @@
...
+    ;; A short function with a single call.
+    (call $once)
+  )
+
+  ;; CHECK:      (func $caller-empty
+  ;; CHECK-NEXT:  (nop)
+  ;; CHECK-NEXT: )
+  (func $caller-empty
+    ;; A tiny function with nothing at all.
+  )
```
aheejin, line 192: What is this function for? It doesn't contain `once` and it is not called from anywhere...

test/lit/passes/once-reduction.wast
```diff
@@ -0,0 +1,1425 @@
...
+  ;; CHECK-NEXT:  (if
+  ;; CHECK-NEXT:   (global.get $once)
+  ;; CHECK-NEXT:   (return)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (global.set $once
+  ;; CHECK-NEXT:   (i32.const 1)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (call $foo)
+  ;; CHECK-NEXT: )
+  (func $once
```
aheejin, line 258:
```suggestion
  (func $not-once
```
Because it is not once, maybe rename it to something else..? (Applies for all not-once corner cases in the file)

Edit: It looks like there are too many places to fix and you use 'once' in the comments as well... So maybe nevermind.

src/passes/OnceReduction.cpp
aheejin: Why does this have to be a constant? We don't care whether the value is zero or nonzero anyway.

src/passes/OnceReduction.cpp
```diff
@@ -0,0 +1,414 @@
...
+  bool isFunctionParallel() override { return true; }
+
+  Scanner(OptInfo& optInfo) : optInfo(optInfo) {}
+
+  Scanner* create() override { return new Scanner(optInfo); }
+
+  // All the globals we read from. Any read of a global prevents us from
+  // optimizing, unless it is the single read at the top of an "only" function
+  // (as other reads might be used to check for the value of the global in
+  // complex ways that we do not want to try to reason about).
```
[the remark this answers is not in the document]
kripken: Other reads may be unrelated to the first call of the "once" function, like this which could appear in some random function:
```js
if (once == 123) { do something }
```
We will be trying to remove not just calls to the "once" function but also other writes to that global. And other writes could be noticed by such a read.

src/passes/OnceReduction.cpp
```diff
@@ -0,0 +1,414 @@
...
+      auto global = getOnceGlobal(curr->body);
+      if (global.is()) {
+        // This is a "once" function, as best we can tell for now. Further
+        // information may cause a problem, say, if the global is used in a bad
+        // way in another function, so we may undo this.
+        optInfo.onceFuncs.at(curr->name) = global;
+
+        // We can ignore the get in the "once" pattern at the top of the
+        // function.
+        readGlobals[global]--;
```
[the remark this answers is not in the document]
kripken: Yes, `Scanner` is a function-parallel pass, so a new instance of it is created for each function.

src/passes/OnceReduction.cpp
```diff
@@ -0,0 +1,414 @@
...
+        optInfo.onceGlobals.at(global) = false;
+      }
+    }
+  }
+
+  // Check if a function body is in the "once" pattern. Return the name of the
+  // global if so, or an empty name otherwise.
+  //
+  // TODO: If the "once" function is inlined, this pattern can show up in random
+  //       places, and we can look for it there as well.
```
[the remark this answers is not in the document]
kripken: We couldn't know if something was caused by inlining, that's true. But if we recognized the more general pattern then we would handle inlining, and other things too. I'll clarify the comment.

src/passes/OnceReduction.cpp
```diff
@@ -0,0 +1,414 @@
...
+      return Name();
+    }
+    if (!iff->ifTrue->is<Return>() || iff->ifFalse) {
+      return Name();
+    }
+    auto* set = list[1]->dynCast<GlobalSet>();
+
+    // Note that we have already checked the set's value earlier, but we do need
+    // it to not be unreachable (so it is actually set).
+    if (!set || set->name != get->name || set->type == Type::unreachable) {
```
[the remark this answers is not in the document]
kripken: Hmm, yes, this comment is not that clear - above we can ignore unreachable sets, but here we require reachability. I'll rephrase, thanks.

src/passes/OnceReduction.cpp
kripken: Good idea, that would be nicer, fixed.

src/passes/OnceReduction.cpp
kripken: Good point, this was a little unclear. Rewriting.

test/lit/passes/once-reduction.wast
kripken: It just tests that we do not crash when processing such a tiny function. (We used to at some point in testing...) I'll add a comment.

src/passes/OnceReduction.cpp
kripken: Hmm, good point! The initial value may as well be anything. Fixed.

src/passes/OnceReduction.cpp
kripken: It is simple to check, but really only mutable globals are relevant here - a mutable one would never fit the pattern we look for. I think such globals would be optimized by other passes already. I'll improve the comment here.

src/passes/OnceReduction.cpp
kripken: In addition to converting global sets and calls into nops, we also note which globals are set (by global sets and calls). That information is what we propagate to blocks dominated by us, but also to functions that call us.

It might be more efficient to propagate fully before doing more local work because the local work will go through all the basic blocks in each function, it's not just doing an operation on the graph of functions.

aheejin: > It might be more efficient to propagate fully before doing more local work because the local work will go through all the basic blocks in each function, it's not just doing an operation on the graph of functions.

When @tlively suggested to split those two, you said it's good to keep the status quo because this is more efficient.. (I think you said it's free) And the comment here says otherwise.

aheejin (approved):

src/passes/OnceReduction.cpp
kripken: Ah, good point, I was maybe inconsistent there. Looks like the original context was a response to @tlively saying this:

> It might be faster to first propagate the onceGlobalsSetInFuncs through the call graph (which doesn't require scanning through the blocks or expressions of a function) and only then start modifying functions.

Doing a first propagation through the call graph wouldn't speed things up, I don't think, as I said above, because to get the full info to propagate we have to scan into functions. We can learn new things from both operations, so we must cycle on both.

Separating a call graph propagation out *might* make sense. It would be more efficient if there is a deep call graph path that we need to get to - we'd get there immediately, instead of each step going through the function contents. I suspect that might be more common than the opposite case, where it would be slower - where we must do full function-level work at each step. That's why I wrote it might be more efficient to add more code for call graph propagation.

As it's not clear which is faster (or if it matters), I think the way with less code is best, as is implemented atm.

PR #4113, opened 2021-08-29
Remove out-of-date multivalue info in README

tlively:
These paragraphs talked about push and pop, which are no longer relevant. More up-to-date information is already included above in the "Tree structure" section.


kripken (commented):
Pop at least is still relevant, isn't it? We use it in a catch block, for example. I think we should still mention that somewhere (looks like with this removal, no mentions would be left).

I think there is value to the 1-3% figure mentioned here, as a rationale for why the entire IR is not multivalue, and why our level of support is limited to tuple types. How about adding a mention of that higher up alongside tree structure etc., while removing this text?

tlively:
Good idea, done.

kripken (approved):

PR #4114, opened 2021-08-30
Use ModuleReader::readStdin for file "-"

### LINKED ISSUE #4105: `wasm-opt` can’t read from stdin/write to stdout
I’d like to use `wasm-opt` as part of a pipe setup, but it seems that `wasm-opt` does not support the typical `-` notation. I’d like this invocation to read a wasm binary from stdin and write the optimized wasm module to stdout:

```
$ cat mymodule.wasm | wasm-opt -o - - > output.wasm
```

I stumbled over this when compiling to wasm via `clang`, which will automatically invoke `wasm-opt`:
```
$ echo "int main() {return 0;}" | clang --target=wasm32 -nostdlib -o - -x c++ -Wl,--no-entry -
```

cc @sbc100

tlively:
After #4106 we already treat the input file "-" as a shorthand for reading from
stdin at the file.cpp level. However, the "-" input was still treated as a
normal file name at the wasm-io.cpp file and as a result was always treated as
text input. This commit updates wasm-io.cpp to use the stdin code path
supporting both binary and text input for "-".

Fixes #4105 (again).


src/wasm/wasm-io.cpp
```diff
@@ -89,7 +89,7 @@ void ModuleReader::read(std::string filename,
                         Module& wasm,
                         std::string sourceMapFilename) {
   // empty filename means read from stdin
```
kripken, line 91: please update the comment as well.

kripken (approved):

PR #4115, opened 2021-08-30
Add a Module parameter to EffectAnalyzer. NFC

kripken:
Knowing the module will allow us to do more analysis in the
effect analyzer. For now, this just refactors the code to allow
providing a module instead of features, and to infer the
features from the module. This actually shortens the code in
most places which is nice (just pass `module` instead of
`module->features`).

This modifies basically all callers to use the new module
form, except for the fallthrough logic. That would require
some more refactoring, so to keep this PR reasonably small
that is not yet done.

### DIFF
--- a/src/ir/effects.h
+++ b/src/ir/effects.h
@@ -31,21 +31,28 @@ public:
                  Expression* ast = nullptr)
     : ignoreImplicitTraps(passOptions.ignoreImplicitTraps),
       trapsNeverHappen(passOptions.trapsNeverHappen),
-      debugInfo(passOptions.debugInfo), features(features) {
+      debugInfo(passOptions.debugInfo), module(nullptr), features(features) {
     if (ast) {
       walk(ast);
     }
   }
 
   EffectAnalyzer(const PassOptions& passOptions,
-                 Module& wasm,
+                 Module& module,
                  Expression* ast = nullptr)
-    : wasm(&wasm), EffectAnalyzer(passOptions, wasm.features, ast) {}
+    : ignoreImplicitTraps(passOptions.ignoreImplicitTraps),
+      trapsNeverHappen(passOptions.trapsNeverHappen),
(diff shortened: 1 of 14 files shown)


src/ir/effects.h
```diff
@@ -680,6 +693,15 @@ class EffectAnalyzer {
     return !aEffects.invalidates(bEffects);
   }
 
+  static bool canReorder(const PassOptions& passOptions,
+                         Module& module,
+                         Expression* a,
+                         Expression* b) {
+    EffectAnalyzer aEffects(passOptions, module, a);
+    EffectAnalyzer bEffects(passOptions, module, b);
+    return !aEffects.invalidates(bEffects);
```
aheejin, line 702: Are we planning to remove the `canReorder` above that takes `features`? If we plan to keep both `canReorder`, we can make one call the other, but if we plan to remove the former the current temporary state looks fine. (The same with the constructor.)

aheejin (approved):

src/ir/effects.h
kripken: Yes, I think it would be good to remove the older form in both cases. I am not 100% sure we can do so, as in general we do allow optimizing code without a module being defined - but it's possible we don't run into that in any caller.

Regardless I actually couldn't find a nice way to make the constructors call each other. The problem is that the original one (with features) needs to set the module field to nullptr, and that happens after the second constructor gets to its ` : module(module)` initialization - so the order is wrong. I considered putting the second constructor's initialization of module in the `{ }` code after the field initializations, but it seemed less nice. As the body is so short, the current form is the best I could think of... but open to ideas, maybe I missed something.

I also couldn't find a way to make `canReorder` call each other. The constructors they call are different, so those lines must be different, which just leaves one identical line.

kripken: Ok, I verified that we can in fact remove the old forms entirely, so this issue will go away. I'll land this and open a followup PR to finish the refactoring.

PR #4118, opened 2021-09-01
Fix the effects of array.copy

kripken:
This appeared to be a regression from #4117, however this
was always a bug, and that PR just exposed it. That is,
somehow we forgot to indicate the effects of ArrayCopy, and
after that PR we'd vacuum it out incorrectly.


tlively (approved):
Oops!

PR #4120, opened 2021-09-03
[wasm-split] Add an option for recording profile data in memory

tlively:
To avoid requiring a static memory allocation, wasm-split's instrumentation
defaults to recording profile data in Wasm globals. This causes problems for
multithreaded applications because the globals are thread-local, but it is not
always feasible to arrange for a separate profile to be dumped on each thread.
To simplify the profiling of such multithreaded applications, add a new
instrumentation mode that stores the profiling data in shared memory instead of
in globals. This allows a single profile to be written that correctly reflects
the called functions on all threads.

This new mode is not on by default because it requires users to ensure that the
program will not trample the in-memory profiling data. The data is stored
beginning at address zero and occupies one byte per declared function in the
instrumented module. Emscripten can be told to leave this memory free using the
GLOBAL_BASE option.

### DIFF
--- a/src/tools/wasm-split/instrumenter.cpp
+++ b/src/tools/wasm-split/instrumenter.cpp
@@ -22,9 +22,8 @@
 
 namespace wasm {
 
-Instrumenter::Instrumenter(const std::string& profileExport,
-                           uint64_t moduleHash)
-  : profileExport(profileExport), moduleHash(moduleHash) {}
+Instrumenter::Instrumenter(const WasmSplitOptions& options, uint64_t moduleHash)
+  : options(options), moduleHash(moduleHash) {}
 
 void Instrumenter::run(PassRunner* runner, Module* wasm) {
   this->runner = runner;
@@ -35,6 +34,10 @@ void Instrumenter::run(PassRunner* runner, Module* wasm) {
 }
 
 void Instrumenter::addGlobals() {
+  if (options.storageKind != WasmSplitOptions::StorageKind::InGlobals) {
+    // Don't need globals
+    return;
+  }
   // Create fresh global names (over-reserves, but that's ok)
   counterGlobal = Names::getValidGlobalName(*wasm, "monotonic_counter");
   functionGlobals.reserve(wasm->functions.size());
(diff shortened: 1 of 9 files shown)


src/tools/wasm-split/instrumenter.cpp
```diff
@@ -142,18 +171,76 @@ void Instrumenter::addProfileExport() {
...
+                    MulInt32, getFuncIdx(), builder.makeConst(uint32_t(4)))),
+                builder.makeAtomicLoad(1, 0, getFuncIdx(), Type::i32),
+                Type::i32),
+              builder.makeLocalSet(
+                funcIdxVar,
+                builder.makeBinary(
+                  AddInt32, getFuncIdx(), builder.makeConst(uint32_t(1)))),
+              builder.makeBreak("l")))));
+      break;
+    }
```
kripken, line 243: Is `writeData` needed for InMemory - I think it can just be read from where it is? Unless I guess you want a uniform interface.

test/lit/wasm-split/instrument-in-memory.wast
```diff
@@ -0,0 +1,88 @@
...
+
+;; Check that a memory has been added
+;; CHECK: (memory $0 1 1)
+
+;; And the profiling function exported
+;; CHECK: (export "__write_profile" (func $__write_profile))
+
+;; Check that the function instrumentation is correct
+
+;; CHECK:      (func $bar
```
kripken, line 26: why does this show up outside the module and not alongside the definition of `bar`?

test/lit/wasm-split/instrument-in-memory.wast
```diff
@@ -0,0 +1,88 @@
...
+;; CHECK-NEXT:       (i32.add
+;; CHECK-NEXT:        (local.get $addr)
+;; CHECK-NEXT:        (i32.mul
+;; CHECK-NEXT:         (local.get $funcIdx)
+;; CHECK-NEXT:         (i32.const 4)
+;; CHECK-NEXT:        )
+;; CHECK-NEXT:       )
+;; CHECK-NEXT:       (i32.atomic.load8_u
+;; CHECK-NEXT:        (local.get $funcIdx)
+;; CHECK-NEXT:       )
```
kripken, line 74: This atomic load could be non-atomic, I think, unless the profile might still be written at this time? But maybe this is better for clarity actually.

tlively: Yes, the in-memory data could be written by other threads at the same time a profile is being written.

src/tools/wasm-split/instrumenter.cpp
tlively: Yes, I definitely want to give a uniform interface. Also, the in-memory representation (1 byte per function) is not the same as the profile representation (hash + 4 bytes per function), and I'd also rather keep a uniform profile format for now as well.

test/lit/wasm-split/instrument-in-memory.wast
tlively: I generally hand-write the wasm-split tests rather than generate their output with a script, although I suppose this particular one could be generated now that we have `--all-items`.

kripken (approved):
I see, makes sense!

PR #4121, opened 2021-09-03
wasm-split: Export the memory if it is not already

kripken:


src/tools/wasm-split/instrumenter.cpp
```diff
@@ -179,7 +179,18 @@ void Instrumenter::addProfileExport() {
...
+  bool memoryExported = false;
+  for (auto& ex : wasm->exports) {
+    if (ex->kind == ExternalKind::Memory) {
+      memoryExported = true;
+      break;
+    }
+  }
+  if (!memoryExported) {
+    wasm->addExport(Builder::makeExport(
+      "profile-memory", wasm->memory.name, ExternalKind::Memory));
```
tlively, line 192: It would be good to use `Names::getValidExportName` here. Also, what do you think about just calling the export "memory" by default? That seems simpler, and there probably won't already be a "memory" export for that to clash with.

tlively (approved):
Thanks!

src/tools/wasm-split/instrumenter.cpp
kripken: Updated to use `Names`, good point.

I actually picked `profile-memory` intentionally because it seems more future-proof. Once we have multiple memories, we'd want to use a separate memory here I think, so being clear on the purpose of the memory seems good.

tlively (approved):
LGTM, thanks!

PR #4126, opened 2021-09-07
Add an Intrinsics mechanism, and a call.if.used intrinsic

kripken:
An "intrinsic" is modeled as a call to an import. We could also add new
IR things for them, but that would take more work and lead to less
clear errors in other tools if they try to read a binary using such a
nonstandard extension. Curious if there are other ideas here though!

A first intrinsic is added here, `call.if.used`. This is basically the same
as `call_ref` except that the optimizer is free to remove the call if the
result is not used. That is, if `foo()` returns an int, then we normally
cannot optimize way `foo();` (a call whose result is not used). But
if we write `call.if.used(foo);` then the optimizer can remove that.

A lowering pass for intrinsics is provided. Rather than automatically
lower them to normal wasm at the end of optimizations, the user must
call that pass explicitly. A typical workflow might be
```
-O --intrinsic-lowering -O
```
That optimizes with the intrinsic present - perhaps removing calls
thanks to it - then lowers it into normal wasm - it turns into a `call_ref` -
and then optimizes further, which would turns the `call_ref` into a
direct call, potentially inline, etc.

The purpose of `call.if.used` is to allow us to optimize away calls to
things that have unnecessary side effects. For example, if we have a
method `getSingletonFoo()` that returns a `Foo` object in j2cl then
we might have code like this:
```js
function bar() {
  var foo = getSingletonFoo();
  if (0) foo.something();
}
```
After local optimizations we end up with
```js
function bar() {
  getSingletonFoo();
}
```
Imagine that `getSingletonFoo()` creates a `Foo` on the first call,
so it has side effects. But, the wasm producer knows that it is ok
to not run those side effects if the value is not used - the object
will be created later (or never), and that is ok in terms of the
language semantics.


test/lit/passes/intrinsic-lowering.wast
```diff
@@ -0,0 +1,42 @@
...
+;; RUN: wasm-opt %s --intrinsic-lowering -all -S -o - | filecheck %s
+
+(module
+  ;; call.if.used with no params.
+  ;; CHECK:      (import "binaryen-intrinsics" "call.if.used" (func $ciu-v (param funcref) (result i32)))
+  (import "binaryen-intrinsics" "call.if.used" (func $ciu-v (param funcref) (result i32)))
+
+  ;; call.if.used with some params.
+  ;; CHECK:      (import "binaryen-intrinsics" "call.if.used" (func $ciu-dif (param f64 i32 funcref) (result f32)))
+  (import "binaryen-intrinsics" "call.if.used" (func $ciu-dif (param f64) (param i32) (param funcref) (result f32)))
```
aheejin, line 11: Are we able to import the same module/base combination multiple times with different signatures..? Didn't know that...

test/spec/binaryen-intrinsics.wast
```diff
@@ -0,0 +1,22 @@
+;; Test `call` operator
```
aheejin, line 1:
```suggestion
;; Test `call.if.used` intrinsic
```
?

test/lit/passes/vacuum-intrinsics.wast
```diff
@@ -0,0 +1,248 @@
...
+  ;; CHECK:      (func $unused-fj-side-effects
+  ;; CHECK-NEXT:  (local $f32 f32)
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (local.tee $f32
+  ;; CHECK-NEXT:    (f32.const 2.718280076980591)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (ref.func $fj)
+  ;; CHECK-NEXT:  )
```
aheejin, line 59: When is it the case that not calling the function is fine but its cannot be omitted? In general, is there any side effect of `ref.func` other than making it "referenced" so we can't remove them or something?

aheejin (commented):
- Are toolchains responsible for generating calls to `call.if.used`?
- Would something like `ignore.side.effects` that can wrap any expression more versatile? Or can we add an attribute to a function, not a call, given that this is likely a function-level characteristic?
- What are other intrinsic candidates you have in mind?

kripken:
@aheejin 

> Are toolchains responsible for generating calls to call.if.used?

Yes.

> Would something like ignore.side.effects that can wrap any expression be more versatile?

@tlively had the same idea offline, coincidentially :smile: Yeah, I think you are both probably right, that is actually better than this. I tried to do the "minimal" thing here, but side effects is the real issue.

Specifically, "no side effects" captures the key issue of removing a call if the result is not used, and also it allows more things like reordering with "call if used" does not.

> Or can we add an attribute to a function, not a call, given that this is likely a function-level characteristic?

That's also an option. It would take more work, though, to define that in the IR and to define a binary and text format for it, which is why I prefer a per-call method.

There may not be a reason for the extra flexibility that per-call allows, but given it doesn't disallow anything, and is easier to emit and to process, I think at this stage in experimentation it's best. However, as the wasm spec's code annotations framework matures, perhaps we should end up using it somehow, and yes, maybe we should switch to per-function at that stage.

> What are other intrinsic candidates do you have in mind?

Another idea that's been mentioned by some wasm GC producers is pure/const, but that may be close enough to "no side effects" that we don't need it, not sure...

README.md
```diff
@@ -141,6 +141,45 @@ Notes when working with Binaryen IR:
...
+
+Implementing them that way allows them to be read and written by other tools,
+and it avoids confusing errors on a binary format error that could happen in
+those tools if we had a custom binary format extension.
+
+An intrinsic method may be optimized away by the optimizer. If it is not, it
+must be **lowered** before shipping the wasm, as otherwise it will look like a
+call to an import that does not exist (and VMs will show an error on not having
+a proper value for that import). That final lowering is *not* done
+automatically - a user of intrinsics must run the pass for that explicitly. Note
```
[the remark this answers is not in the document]
kripken: I'll add some more text for this. The key issue is that it's not clear when is the right time to do it automatically - we need to know when the code is meant for "shipping". That is, if the user is just optimizing binaryen IR locally (maybe experimenting, or maybe doing fuzzing/reducing etc.), then we don't want to lower, but if it's the final optimization before shipping, then we do.

test/lit/passes/intrinsic-lowering.wast
kripken: Yes, it's a little odd, but it's useful in rare cases like when importing something from JS that we can pick different types for.

test/lit/passes/vacuum-intrinsics.wast
kripken: (I can't parse your first sentence - what does "its" refer to?)

`ref.func` doesn't have any side effects, but currently if any of the params has a side effect, we drop them all and leave it for later passes to remove.

aheejin: its -> it
Yeah I was wondering if we can remove the `ref.func` too. But if the later pass does that that should be fine.

aheejin (approved):
I see, thanks for the explanation!

kripken:
I pushed an update to replace this with `call.without.effects` as suggested. Turns out that's even simpler to implement, so my earlier guess was wrong that it would be more work...

src/ir/intrinsics.h
```diff
@@ -0,0 +1,95 @@
...
+  //   (import "binaryen-intrinsics" "call.without.effects"
+  //     (func (..params..) (param $target funcref) (..results..)))
+  //
+  // call.if.used can take any parameters, and in addition a funcref, and return
+  // any result.
+  //
+  // Precise semantics:
+  //
+  //  * The optimizer will assume this instruction has no side effects.
+  //  * Final lowering turns a call.if.used into a call of the given function
```
tlively, line 52:
```suggestion
  //  * Final lowering turns a call.without.effects into a call of the given function
```

src/ir/intrinsics.h
```diff
@@ -0,0 +1,95 @@
...
+public:
+  Intrinsics(Module& module) : module(module) {}
+
+  //
+  // Check if an instruction is the call.without.effects intrinsic.
+  //
+  //   (import "binaryen-intrinsics" "call.without.effects"
+  //     (func (..params..) (param $target funcref) (..results..)))
+  //
+  // call.if.used can take any parameters, and in addition a funcref, and return
```
tlively, line 46:
```suggestion
  // call.without.effects can take any parameters, and in addition a funcref, and return
```

tlively:
> > Would something like ignore.side.effects that can wrap any expression be more versatile?
> 
> @tlively had the same idea offline, coincidentially  Yeah, I think you are both probably right, that is actually better than this. I tried to do the "minimal" thing here, but side effects is the real issue.

@aheejin's suggested `ignore.side.effects` is actually more general than what I suggested, which IIRC was essentially the current `call.without.effects`. I think this PR would be simpler if it implemented @aheejin's suggestion, i.e. an intrinsic that could wrap any expression (including any kind of call). Specifically, the lowering would be simpler because the intrinsic would simply be replaced with its single operand and validation would be simplified because it would take the type of its single operand. Function references would not be necessary at all.

kripken:
Oh, sorry, I did not understand that specific proposal, then.

There are composition issues with that idea, if I understand it correctly. E.g.
```wat
(local.set $x
  (ignore.side.effects
    (call $foo)))
```
would not be equivalent to
```wat
(local.set $temp
  (call $foo))
(local.set $x
  (ignore.side.effects
    (local.get $temp)))
```
(flatten converts from the former to the latter, and  simplify-locals in reverse).

And also things like
```wat
(ignore.side.effects
  (call $foo))

?

(ignore.side.effects
  (block (result ..)
    (call $foo)))
```
Does it check effects on the block, or on all nested children? If just the block, then that's wrong; if all nested children, then we can hit a simplify-locals issue like from earlier:
```wat
(local.set $x
  (call $bar))
(ignore.side.effects
  (call $foo
    (local.get $x)))

?

(ignore.side.effects
  (call $foo
    (call $bar)))
```

There isn't a good way to force two expressions to remain "attached", that is, stay as the direct child of the other (and also no way to prevent adding more nested children in addition). More generally, the "meaning" of a node should not depend on "context" around it (with the exceptions of breaks).

tlively:
Oof, that's unfortunate. Thanks for the clear examples.

tlively (approved):

src/ir/intrinsics.h
```diff
@@ -0,0 +1,95 @@
...
+  //
+  //  (local.set $x (call $call.without.effects (ref.func $get-something)))
+  //     =>
+  //  (local.set $x (call $get-something))
+  //
+  // Later passes will then turn that into a direct call and further optimize
+  // things.
+  //
+  bool isCallWithoutEffects(Function* func);
+  Call* isCallWithoutEffects(Expression* curr);
```
aheejin, line 90: Should this be a class with non-static member functions? These functions don't seem to refer to the `module` member...

src/passes/Intrinsics.cpp
```diff
@@ -0,0 +1,61 @@
...
+      // forgets to optimize.
+      Builder builder(*getModule());
+      if (auto* refFunc = target->dynCast<RefFunc>()) {
+        auto name = refFunc->func;
+        auto* func = getModule()->getFunction(name);
+        if (!HeapType::isSubType(func->type, refFunc->type.getHeapType())) {
+          Fatal() << "call.without.effects is not of a subtype of the called "
+                     "function ("
+                  << name << ")\n";
+        }
```
aheejin, line 46: Isn't this a part of the validation of `RefFunc`, separate from `call.without.effects`, which should have been done in the validator already?

src/passes/Intrinsics.cpp
```diff
@@ -0,0 +1,61 @@
...
+        replaceCurrent(builder.makeCall(name, operands, curr->type));
+      } else {
+        if (target->type.isBasic()) {
+          Fatal() << "Cannot emit call_ref for call.without.effects without "
+                     "a specialized function type";
+        }
```
aheejin, line 52: Shouldn't this be in the validation of `call.without.effects` instead?

test/lit/passes/vacuum-intrinsics.wast
```diff
@@ -0,0 +1,233 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
+;; RUN: wasm-opt %s --vacuum -all -S -o - | filecheck %s
```
aheejin, line 2: Now we don't have any changes in Vacuum.cpp, then where do we do these optimizations?

src/ir/intrinsics.h
kripken: It's used here: https://github.com/WebAssembly/binaryen/pull/4126/files#diff-4e0ac873d5e741220507e824e0b69ef79057f2ff1de3117afeb9ceb84cd4f714R33 to get the Function from the module using its name.

src/passes/Intrinsics.cpp
kripken: Good catch! I wanted to check that the type of the `call.without.effects` is the right one. It should be identical to the function reference's, ignoring the last parameter (which is the function reference), but this code does not do that...

Actually checking this would be a bunch more code, so I think maybe I'll just remove it, and leave it to the validator. I'd hoped for a clearer error message here, but let's see if it's a problem in practice.

test/lit/passes/vacuum-intrinsics.wast
kripken: Just by seeing that the call has no side effects, vacuum is able to remove more things. It will remove a drop of anything without side effects, for example.

aheejin (approved):

PR #4131, opened 2021-09-08
[Wasm GC] Properly error on invalid struct/array.get packing

### LINKED ISSUE #4129: struct.get is silently turned into struct.get_u for packed fields
It would be more correct for this to be a validation failure.

kripken:
Fixes #4129


src/wasm-s-parser.h
```diff
@@ -279,12 +279,17 @@ class SExpressionWasmBuilder {
   Expression* makeRttCanon(Element& s);
   Expression* makeRttSub(Element& s);
   Expression* makeRttFreshSub(Element& s);
+
+  // A field may be packed, in which case it is signed or unsigned, or it may
+  // not be packed.
+  enum class Packed { Signed, Unsigned, Non };
```
tlively, line 285: Maybe `Extension` would be more descriptive than `Packed` here? If so, I would also switch `Non` to `None`.

src/wasm/wasm-s-parser.cpp
```diff
@@ -2680,12 +2689,21 @@ Expression* SExpressionWasmBuilder::makeArrayNew(Element& s, bool default_) {
...
   auto heapType = parseHeapType(*s[1]);
+  auto element = heapType.getArray().element;
+  if (element.isPacked() && packed == Packed::Non) {
+    throw ParseException(
+      "packed element must be read as packed", s.line, s.col);
+  }
+  if (!element.isPacked() && packed != Packed::Non) {
+    throw ParseException(
+      "non-packed element must not be read as packed", s.line, s.col);
+  }
```
tlively, line 2702: This check could be factored out into a shared function.

tlively (approved):
Nice! LGTM with two optional changes.

src/wasm-s-parser.h
kripken: Hmm, I actually had to read your sentence twice as my brain thought you meant something to do with a suffix (like a file extension). But fair point that extending is the actual operation here. Anyhow, I think I prefer it as it is, but I don't feel strongly.

src/wasm/wasm-s-parser.cpp
kripken: Fair enough. The error text is different, though, which will make that slightly less nice, but I agree it's probably better overall.

src/wasm-s-parser.h
kripken: Perhaps "PackedExtend"? Seems verbose though.

tlively: Maybe `Extend` would be better than `Extension`? It's too bad there's not a better word describing both sext and zext.

src/wasm/wasm-s-parser.cpp
```diff
@@ -2643,20 +2643,28 @@ Index SExpressionWasmBuilder::getStructIndex(Element& type, Element& field) {
   return atoi(field.c_str());
 }
 
+static void verifyPackedRead(const Field& field,
+                             SExpressionWasmBuilder::Packed packed,
+                             Element& s,
+                             std::string what) {
```
tlively, line 2649: The `what` could probably just be "field" in both cases, since the line and column will identify the precise culprit.

src/wasm-s-parser.h
tlively: Or maybe `PackedRead` or `PackedReadKind`? It's not really the packed field that is signed, but the read operation itself.

src/wasm/wasm-s-parser.cpp
kripken: Hmm, but Array has `element` which I've assumed was intentional?

Regardless, for the error message I think we want to say if it's a struct or an array, so the text should differ I think.

src/wasm-s-parser.h
kripken: Hmm, I'm not that happy with any of these names I guess. Maybe the name could refer to "signedness" more generally, as in "this is signed / this is unsigned / signedness does not matter here" - ?

tlively: `Signedness` sgtm and could probably replace some bools elsewhere in our code base.

kripken: Good point, yes - I think that is the right direction. It feels correct to replace the various `bool signed_` with something like that. PR updated, and if that makes sense further refactoring could be done later.

src/wasm-type.h
```diff
@@ -58,6 +58,15 @@ struct Rtt;
 enum Nullability { NonNullable, Nullable };
 enum Mutability { Immutable, Mutable };
 
+enum Signedness {
+  // The value has signedness, and is signed, like e.g. int32_t.
+  Signed,
+  // The value has signedness, and is unsigned, like e.g. uint32_t.
+  Unsigned,
+  // The value has no concept of signedness, like e.g. float or raw bits.
+  Nonsigned
```
tlively, line 67: I would prefer `None` here to visually differentiate more from the other two options.

tlively (approved):

src/wasm-type.h
kripken: I see your point, but it would have to be an enum class, then, unlike the others, to force `Signedness::None`. But ` Signedness::Signed` seems burdensome, hmm...

Perhaps `Sign::Signed/Unsigned/None`? Also kind of weird though...

tlively: Oh, good point about the enum class being unwieldy. How about `NoSign`, then?

kripken: `NoSign` sounds good!

PR #4134, opened 2021-09-09
Make static buffers in numToString thread local

tlively:
Validation is performed on multiple threads at once and when there are multiple
validation failures, those threads can all end up in `numToString` at the same
time as they construct their respective error messages. Previously the threads
would race on their access to the snprintf buffers, sometimes leading to
segfaults. Fix the data races by making the buffers thread local.

### DIFF
--- a/src/emscripten-optimizer/simple_ast.h
+++ b/src/emscripten-optimizer/simple_ast.h
@@ -947,8 +947,9 @@ struct JSPrinter {
     bool integer = wasm::isInteger(d);
 #define BUFFERSIZE 1000
     // f is normal, e is scientific for float, x for integer
-    char full_storage_f[BUFFERSIZE];
-    char full_storage_e[BUFFERSIZE];
+    // These need to be thread-local because they are returned.
+    thread_local char full_storage_f[BUFFERSIZE];
+    thread_local char full_storage_e[BUFFERSIZE];
     // full has one more char, for a possible '-'
     char* storage_f = full_storage_f + 1;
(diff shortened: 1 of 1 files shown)


kripken (commented):
Good find!

Is it worth doing `thread_local` here? How about just removing the `static`, which would make these normal stack allocations? That seems simpler and it might actually be faster.

tlively:
Sure, I guess a couple kilobytes on the stack shouldn't matter.

src/emscripten-optimizer/simple_ast.h
```diff
@@ -947,17 +947,18 @@ struct JSPrinter {
     bool integer = wasm::isInteger(d);
 #define BUFFERSIZE 1000
     // f is normal, e is scientific for float, x for integer
-    static char full_storage_f[BUFFERSIZE], full_storage_e[BUFFERSIZE];
+    char full_storage_f[BUFFERSIZE] = {};
+    char full_storage_e[BUFFERSIZE] = {};
```
kripken, line 951: does `= {}` initialize it to 0? I don't think we need that here.

tlively: Removed.

kripken (approved):

tlively:
Oh, it looks like the `thread_local` on the buffers might be important after all, since this function returns pointers to the buffers and the buffers need to live longer than the function call.

kripken:
Oh, good point, yeah... Indeed that is necessary. lgtm with that.

PR #4138, opened 2021-09-09
[Wasm GC] ArrayInit support

kripken:
`array.init` is like `array.new_with_rtt` except that it takes
as arguments the values to initialize the array with (as opposed to
a size and an optional initial value).

Spec: https://docs.google.com/document/d/1afthjsL_B9UaMqCA5ekgVmOm75BVFu6duHNsN9-gnXw/edit#

I opened a comment there about the name, as in that doc the
text is `array.init` (without `_with_rtt`). edit: for now, let's copy
that document - so `array.init`.


src/wasm-binary.h
```diff
@@ -1056,6 +1056,7 @@ enum ASTNodes {
   ArraySet = 0x16,
   ArrayLen = 0x17,
   ArrayCopy = 0x18,
+  ArrayInitWithRtt = 0x19,
```
tlively, line 1059: Should this be just `ArrayInit`?

src/wasm/wasm-validator.cpp
```diff
@@ -2416,6 +2417,31 @@ void FunctionValidator::visitArrayNew(ArrayNew* curr) {
...
+    return;
+  }
+  auto heapType = curr->rtt->type.getHeapType();
+  if (!shouldBeTrue(
+        heapType.isArray(), curr, "array.init heap type must be array")) {
+    return;
+  }
+  const auto& element = heapType.getArray().element;
+  for (auto* value : curr->values) {
+    shouldBeSubType(value->type,
```
tlively, line 2438: Does array.init support packed data? If so, does this condition work for them? I guess it would since `i32 <: i32`, right?

tlively (approved):

src/wasm/wasm-validator.cpp
kripken: Yes, and yes, it should just work. Added a test to be sure.

src/wasm-binary.h
kripken: Good point, thanks, fixed.

PR #4142, opened 2021-09-10
[NFC] Add a comment in MergeBlocks about subtyping

kripken:
Followup to #4137 , I looked into the idea @tlively had but I
realized it is actually not necessary for a non-obvious reason.
This PR adds a comment.


tlively (approved):
Makes sense, thanks!

PR #4144, opened 2021-09-10
RemoveUnusedBrs::tablify() improvements: handle EqZ and tee

kripken:
`tablify()` attempts to turns a sequence of `br_if`s into a single
`br_table`. This PR adds some flexibility to the specific pattern it
looks for, specifically:

* Accept `i32.eqz` as a comparison to zero, and not just to look
  for `i32.eq` against a constant.
* Allow the first condition to be a `tee`. If it is, compare later
  conditions to `local.get` of that local.

This will allow more `br_table`s to be emitted in j2cl output.

### DIFF
--- a/src/passes/RemoveUnusedBrs.cpp
+++ b/src/passes/RemoveUnusedBrs.cpp
@@ -1316,7 +1316,15 @@ struct RemoveUnusedBrs : public WalkerPass<PostWalker<RemoveUnusedBrs>> {
             // dce will clean it up
             return nullptr;
           }
-          auto* binary = br->condition->dynCast<Binary>();
+          auto* condition = br->condition;
+          // Also support eqz, which is the same as == 0.
+          if (auto* unary = condition->dynCast<Unary>()) {
+            if (unary->op == EqZInt32) {
+              return br;
+            }
+            return nullptr;
+          }
+          auto* binary = condition->dynCast<Binary>();
           if (!binary) {
             return nullptr;
           }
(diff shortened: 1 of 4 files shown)


src/passes/RemoveUnusedBrs.cpp
```diff
@@ -1360,6 +1381,14 @@ struct RemoveUnusedBrs : public WalkerPass<PostWalker<RemoveUnusedBrs>> {
             start++;
             continue;
           }
+          // If the first condition value is a tee, that is ok, so long as the
+          // others afterwards are gets of the value that is tee'd.
```
tlively, line 1385: Where is the check that the following tees are of the same value?

test/passes/remove-unused-brs_enable-multivalue.txt
```diff
@@ -1669,6 +1669,109 @@
   )
   (unreachable)
  )
+ (func $br-to-table-initial-tee (param $a i32)
```
tlively, line 1672: How would you feel about porting this test to lit as a preliminary step?

kripken: Hmm, I think the remaining unported tests had reasons they were not ported yet? Do you remember what reason this particular file had? I don't think I see anything obvious.

But if the test can be ported automatically then I'd rather do that later as I've already done the work to add a test in the current form.

src/passes/RemoveUnusedBrs.cpp
kripken: We only allow the first item in a future `br_table` to be a tee - the other parts of the future `br_table` must be gets of the tee'd value.

That is enforced by only having tee logic up here (before the loop that finds the rest of the items), and by setting the condition that the others will be compared to to a get of the proper local.

(Separately from this PR, this pattern also has no "holes" in the middle, so nothing gets a chance to write to the local in between.)

test/passes/remove-unused-brs_enable-multivalue.txt
tlively: No, there are still plenty of tests to port automatically. I'll get that PR stream restarted. Porting the test before hand would make reviewing it marginally easier, but porting it later sounds fine to me.

tlively (approved):

PR #4154, opened 2021-09-14
OptimizeInstructions: Select => and/or in more cases

kripken:
```
x ? 0 : y  ==>   z & y   where z = !x

x ? y : 1  ==>   z | y   where z = !x
```

Only do this when we have `z = !x`, that is, we can invert `x` without adding
an actual `eqz`. (Adding an `eqz` would turn this from one instruction into
two, which means more work, which could be slower.)


src/passes/OptimizeInstructions.cpp
```diff
@@ -1835,17 +1835,46 @@ struct OptimizeInstructions
...
+      // TODO: Perhaps also do this even if we can't remove the eqz, if we are
+      //       optimizing for size? We should measure the performance aspect
+      //       first though, to see if we need that.
+      //
+      // (Note that aside from a relational binary which we can flip, we could
+      // also check for an eqz unary, but a select with an eqz condition would
+      // have already been removed in the cases relevant to us, since the true
+      // and false arms can be flipped given that one of them is a constant.)
+      if (auto* binary = curr->condition->dynCast<Binary>()) {
+        if (binary->isRelational()) {
```
MaxGraey, line 1865: Also note `invertBinaryOp` can be done only for integer relational binary operations. So this line required also check `binary->left->type.isInteger()`

src/passes/OptimizeInstructions.cpp
```diff
@@ -1835,17 +1835,46 @@ struct OptimizeInstructions
...
+      // Some operations require us to add an eqz, like
+      // x ? 0 : y   ==>   !x & y
+      // We replace the select with an and, which does not change code size, and
+      // then reduce code size by removing the number (2 bytes) and adding an
+      // eqz (one byte). However, the eqz adds work, so it is not obvious that
+      // this is beneficial to do - only do it when we know we can fold away the
+      // eqz.
+      // TODO: Perhaps also do this even if we can't remove the eqz, if we are
+      //       optimizing for size? We should measure the performance aspect
+      //       first though, to see if we need that.
```
MaxGraey, line 1858: I guess instead `eqz` it could be `x ^ 1` which may be a faster but bigger in term of code size

src/passes/OptimizeInstructions.cpp
```diff
@@ -1835,17 +1835,46 @@ struct OptimizeInstructions
...
+      //
+      // (Note that aside from a relational binary which we can flip, we could
+      // also check for an eqz unary, but a select with an eqz condition would
+      // have already been removed in the cases relevant to us, since the true
+      // and false arms can be flipped given that one of them is a constant.)
+      if (auto* binary = curr->condition->dynCast<Binary>()) {
+        if (binary->isRelational()) {
+          // x ? 0 : y   ==>   !x & y
+          if (matches(curr->ifTrue, ival(0))) {
+            binary->op = invertBinaryOp(binary->op);
```
MaxGraey, line 1868: Also note `invertBinaryOp` should check for `InvalidBinary` which can be appear for some relational floating point ops like `<`, `>` `>=`, `<=` due to NaN behavior

kripken: True, but we have checked above that all the children of the select are `i32` here, so that should not be a problem.

src/passes/OptimizeInstructions.cpp
kripken: Hmm, I'd expect the VM to optimize that, if it were faster. Otherwise every eqz should be replaced with that larger pattern.

MaxGraey: replacing `eqz(x)` to `x ^ 1` possible only if `getMaxBits(x) == 1` so it's not always possible and I guess not always clear for VMs

src/passes/OptimizeInstructions.cpp
MaxGraey: But it's not relate to `ifFalse` or `ifTrue`. `condition` also have `i32` result while binary operation of `condition` could be floating point. I mean this case:
```wat
x > 2.0 ? 1 : y < 0.0
```
where `x` and `y` are `f64` but all expressions results are `i32`.

src/passes/OptimizeInstructions.cpp
kripken: Ah, good point. Might be worth looking at then.

src/passes/OptimizeInstructions.cpp
kripken: I see, thanks. Ok, then this PR needs more work then.

MaxGraey: if just need to add one more simple check:
```cpp
BinaryOp op  = invertBinaryOp(binary->op);
if (op != InvalidBinary) {
  binary->op =  op;
  return builder.makeBinary(AndInt32, curr->ifFalse, binary);
}
```
and same for second case

src/passes/OptimizeInstructions.cpp
MaxGraey: However now I'm not sure. We already have reversed rule in OptimizeInstructions for: `x ^ 1` -> `eqz(x)`, if `getMaxBits(x) == 1`. But I guess we could hide it under opt options. I have an assumption that eqz(x) could be best option if it combined with if, br_if or select but suboptimal for rest cases

test/lit/passes/optimize-instructions.wast
```diff
@@ -922,13 +948,67 @@
...
+        (local.get $y)
+        (i32.const 1337)
+      )
+      ;; With a 1 here, we must negate the condition, but the condition here
+      ;; cannot be negated in a simple way, so skip.
+      (i32.const 1)
+      (i32.shr_u
+        (local.get $x)
+        (i32.const 31)
+      )
```
MaxGraey, line 1002: Btw what do you think add `(unsigned)x >> 31` -> `(signed)x < 0` to canonicalize method?

MaxGraey: I created PR for this: #4160

MaxGraey:
Will be great also add this rules:

```
x ? x : y   ==>   x | y
x ? y : x   ==>   x & y
```
x, y <- 1-bit types

MaxGraey:
@kripken could you merge this?

src/passes/OptimizeInstructions.cpp
```diff
@@ -2404,17 +2404,53 @@ struct OptimizeInstructions
         Bits::getMaxBits(curr->ifFalse, this) <= 1) {
       // The condition and both arms are i32 booleans, which allows us to do
       // boolean optimizations.
-      Expression* x;
-      Expression* y;
 
       // x ? y : 0   ==>   x & y
-      if (matches(curr, select(any(&y), ival(0), any(&x)))) {
```
tlively, line 2411: Starting the match with `select` even though we already know it is select is intentional so we can read the whole pattern directly without any context. It also lets us use variable names in the output that directly match the names in the input pattern comment.

src/passes/OptimizeInstructions.cpp
```diff
@@ -2125,17 +2125,53 @@ struct OptimizeInstructions
...
+
+          // x ? y : 1  ==>   z | y   where z is the inverse of x
+          if (matches(curr->ifFalse, ival(1))) {
+            auto inv = invertBinaryOp(binary->op);
+            if (inv != InvalidBinary) {
+              binary->op = inv;
+              return builder.makeBinary(OrInt32, curr->ifTrue, binary);
+            }
+          }
+        }
```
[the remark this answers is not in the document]
tlively: +1 this seems like it would be simpler.

src/passes/OptimizeInstructions.cpp
```diff
@@ -4110,6 +4146,8 @@ struct OptimizeInstructions
     }
   }
 
+  // Invert (negate) the opcode, so that it has the exact negative meaning as it
+  // had before.
   BinaryOp invertBinaryOp(BinaryOp op) {
     // use de-morgan's laws
```
tlively, line 4152: This function does not look directly related to De Morgan's laws!

src/passes/OptimizeInstructions.cpp
```diff
@@ -2404,17 +2404,53 @@ struct OptimizeInstructions
...
+      //       first though, to see if we need that.
+      //
+      // (Note that aside from a relational binary which we can flip, we could
+      // also check for an eqz unary, but a select with an eqz condition would
+      // have already been removed in the cases relevant to us, since the true
+      // and false arms can be flipped given that one of them is a constant.)
+      if (auto* binary = curr->condition->dynCast<Binary>()) {
+        if (binary->isRelational()) {
+          // x ? 0 : y   ==>   z & y   where z is the inverse of x
+          if (matches(curr->ifTrue, ival(0))) {
```
tlively, line 2436: It would be great to use `matches` more aggressively here and below:

```c++
if (matches(curr, select(ival(0), any(&y), binary(&x)))) {
  if (auto op = invertBinaryOp(x->op); op != invalidBinary) {
    ...
```

test/lit/passes/optimize-instructions.wast
```diff
@@ -874,9 +874,35 @@
...
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:   (i32.ge_u
+  ;; CHECK-NEXT:    (local.get $x)
+  ;; CHECK-NEXT:    (i32.const 20)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $select-or-negation (param $x i32) (param $y i32) (result i32)
+    (select
+      ;; We can turn this select into an and, by negating the condition.
```
tlively, line 891:
```suggestion
      ;; We can turn this select into an and by negating the condition.
```

src/passes/OptimizeInstructions.cpp
kripken: Hmm, interesting. That is less efficient potentially but maybe it is more readable...

src/passes/OptimizeInstructions.cpp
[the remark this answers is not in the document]
kripken: Yes, that might be simpler, fair point.

Meanwhile I think this PR can be even simpler by never even calling `makeBinary`. We can instead canonicalize a 0 in a select to the ifFalse and a 1 to the ifTrue, when the condition allows it. Then rely on the existing optimization to do the rest:
```
x >= y ? 0 : z
  => new operation here
x < y ? z : 0
  => existing opt from before
(x < y) & z
```

That would make this a pure canonicalization PR.

Overall from all the comments here it seems like there's a bit of work left to get this landed. Not sure when I'll have time as this seems quite low priority compared to other optimizations I'm working on. If someone else is interested to pick this up that would be welcome.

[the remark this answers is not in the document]
MaxGraey: How about merge this as is with TODO about canonicalization?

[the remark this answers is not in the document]
kripken: I don't think we should merge this as-is because there are several outstanding comments. I can try to get to them though.

tlively (approved):

PR #4155, opened 2021-09-15
DWARF: Canonicalize tombstone values

kripken:
This uses LLVM-style tombstone values in more places, and avoids
doing math on tombstone values incorrectly. This reduces a large
amount of dwarfdump verification errors that currently happen even on
hello world.

cc @pfaffe

### DIFF
--- a/test/passes/class_with_dwarf_noprint.bin.txt
+++ b/test/passes/class_with_dwarf_noprint.bin.txt
@@ -336,7 +336,7 @@ Abbrev table for offset: 0x00000000
 
 0x0000011f:     DW_TAG_lexical_block [22] *
                   DW_AT_low_pc [DW_FORM_addr]	(0x0000000000000030)
-                  DW_AT_high_pc [DW_FORM_data4]	(0xffffffd0)
+                  DW_AT_high_pc [DW_FORM_data4]	(0x00000000)
 
 0x00000128:       DW_TAG_variable [21]  
                     DW_AT_location [DW_FORM_sec_offset]	(0x00000046: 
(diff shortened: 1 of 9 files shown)


test/passes/class_with_dwarf_noprint.bin.txt
```diff
@@ -336,7 +336,7 @@ Abbrev table for offset: 0x00000000
 
 0x0000011f:     DW_TAG_lexical_block [22] *
                   DW_AT_low_pc [DW_FORM_addr]	(0x0000000000000030)
-                  DW_AT_high_pc [DW_FORM_data4]	(0xffffffd0)
+                  DW_AT_high_pc [DW_FORM_data4]	(0x00000000)
```
kripken, line 339: This was a case of doing math with a tombstone.

test/passes/dwarfdump_roundtrip_dwarfdump.bin.txt
```diff
@@ -141,7 +141,7 @@ Abbrev table for offset: 0x00000000
               DW_AT_high_pc [DW_FORM_data4]	(0x00000000)
 
 0x00000026:   DW_TAG_subprogram [2]  
-                DW_AT_low_pc [DW_FORM_addr]	(0x0000000000000000)
+                DW_AT_low_pc [DW_FORM_addr]	(0x00000000ffffffff)
```
kripken, line 144: This looks odd, but is what LLVM does, and LLVM's current dwarfdump (newer than the dumping code we are using here) will actually print `(dead code)` for this.

Issue #4061, opened 2021-08-06
Some opportunities for TrapsNeverHappen


askeksa-google:
I tried out the TrapsNeverHappen mode from https://github.com/WebAssembly/binaryen/pull/4059 on the Dart barista3 benchmark, and it successfully reduced the number of `(drop (ref.cast ...))`, `(drop (ref.as_non_null ...))` or `(drop (struct.get ...))` occurrences (as measured by searches in the `.wat` output) from more than 3000 to less than 150.

The remaining occurrences all seem to be cases where the inner expression has some side effect, but it should still be sound to reduce these to just a `drop` with the same contents.

Another situation sometimes seen is where the `ref.cast` or `ref.as_non_null` is unnecessary, in the sense that the code will still validate if the check is removed. These can also be removed in TrapsNeverHappen mode.

I found a function in the optimized module which (along with one it calls) illustrates some of these opportunities:
```wat
 (func $5602 (param $0 dataref) (result (ref null $type$5))
  (local $1 (ref null $type$574))
  (drop
   (ref.as_non_null
    (struct.get $type$574 0
     (local.tee $1
      (ref.cast
       (local.get $0)
       (rtt.canon $type$574)
      )
     )
    )
   )
  )
  (drop
   (call $5603
    (ref.as_non_null
     (local.get $1)
    )
   )
  )
  (ref.null $type$5)
 )
 (func $5603 (param $0 dataref) (result (ref null $type$5))
```
Remove the `ref.as_non_null`:
```wat
 (func $5602 (param $0 dataref) (result (ref null $type$5))
  (local $1 (ref null $type$574))
  (drop
   (struct.get $type$574 0
    (local.tee $1
     (ref.cast
      (local.get $0)
      (rtt.canon $type$574)
     )
    )
   )
  )
  (drop
   (call $5603
    (ref.as_non_null
     (local.get $1)
    )
   )
  )
  (ref.null $type$5)
 )
 (func $5603 (param $0 dataref) (result (ref null $type$5))
```
Remove the `struct.get`:
```wat
 (func $5602 (param $0 dataref) (result (ref null $type$5))
  (local $1 (ref null $type$574))
  (drop
   (local.tee $1
    (ref.cast
     (local.get $0)
     (rtt.canon $type$574)
    )
   )
  )
  (drop
   (call $5603
    (ref.as_non_null
     (local.get $1)
    )
   )
  )
  (ref.null $type$5)
 )
[...]

PR #4157, opened 2021-09-15
[Wasm GC] Optimize away ref.as_non_null going into local.set in TNH mode

kripken:
If we can remove such traps, we can remove `ref.as_non_null` if the local
type is nullable anyhow.

If we support non-nullable locals, however, then do not do this, as it
could inhibit specializing the local type later.

Background: https://github.com/WebAssembly/binaryen/issues/4061#issuecomment-920033473


src/passes/OptimizeInstructions.cpp
```diff
@@ -1054,21 +1054,32 @@ struct OptimizeInstructions
...
           as->finalize();
           replaceCurrent(as);
+          return;
+        }
+
+        // Otherwise, if this is not a tee, then no value falls through. The
+        // ref.as_non_null acts as a null check here, basically. If we are
+        // ignoring such traps, we can remove it. Note that we only do so when
+        // we do not allow non-nullable locals, as removing the cast of the type
+        // to be non-nullable would prevent specializing the local's type later.
```
aheejin, line 1078: Where do we specialize local types?

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -1247,7 +1247,6 @@
         (ref.as_non_null
           (local.get $x)
         )
-        (rtt.canon $struct)
```
aheejin, line 1250: What was this? (Was this a correct syntax?)

aheejin (approved):

src/passes/OptimizeInstructions.cpp
kripken: The LocalSubtyping pass does that:

https://github.com/WebAssembly/binaryen/blob/1ec93d2ed521431ffea3152759a5b693e085d7a1/src/passes/LocalSubtyping.cpp#L18-L22

test/lit/passes/optimize-instructions-gc.wast
kripken: This was just incorrect syntax I happened to notice. Our parser atm will ignore extra items at the end in most cases (unless it has an explicit check for the length), and this is one such case.

src/passes/OptimizeInstructions.cpp
aheejin: Thanks! It might be helpful to mention the pass name (`LocalSubtyping`) in the comments for readers like me 😅

kripken: Oh yes, very good point! Added.

tlively (approved):
All the new code LGTM, but I have a question about the pre-existing code.

edit: whoops, looks like I never saved my question! Fixing...

src/passes/OptimizeInstructions.cpp
```diff
@@ -1054,21 +1054,33 @@ struct OptimizeInstructions
...
+        // can be reordered to
+        //   (ref.as_non_null (local.tee ..))
+        // if the local is nullable anyhow. The reordering allows the ref.as to
+        // be potentially optimized further based on where the value flows to.
+        if (curr->isTee()) {
           curr->value = as->value;
           curr->finalize();
           as->value = curr;
           as->finalize();
           replaceCurrent(as);
```
tlively, line 1070: Doesn't this transformation on tees inhibit the specialization of the local type just like we are trying to avoid in this PR?

kripken: Good point, yeah, the pre-existing code does have that issue. I think it was written before we have LocalSubtyping...

Updated the PR to do this consistently.

tlively (approved):

PR #4158, opened 2021-09-15
Fix regression from #4130

kripken:
That PR reused the same node twice in the output, which fails on the
assertion in `BINARYEN_PASS_DEBUG=1` mode.

No new test is needed because the existing test suite fails already
in that mode. That the PR managed to land seems to say that we are
not testing pass-debug mode on our lit tests?

A second critical thing is that as I mentioned in

https://github.com/WebAssembly/binaryen/pull/4130#issuecomment-920407688

the testcase fails with `--enable-bulk-memory` but works with `-all`. We
need to figure out why that is.


tlively (commented):
I think this needs to either use a new temp local or have a check that `memFill->dest` does not have side effects that can't be duplicated.

kripken:
Oh right, sorry, I was doing this too rushed... yes, this needs a new local.

test/lit/passes/optimize-instructions-bulk-memory.wast
```diff
@@ -1,5 +1,6 @@
 ;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
-;; RUN: wasm-opt %s --all-features --optimize-instructions --enable-bulk-memory -S -o - | filecheck %s
+;; RUN: wasm-opt %s --optimize-instructions --enable-bulk-memory --enable-simd -S -o - | filecheck %s
+;; RUN: wasm-opt %s --optimize-instructions --enable-bulk-memory -S -o - | filecheck %s --check-prefix NOSIMD
```
kripken, line 3: I added a NOSIMD mode to avoid changing the existing results (which allowed SIMD as they had `--all-features`).

tlively (approved):

Issue #4073, opened 2021-08-10
Fuzzing: Focus on recent initial contents


kripken:
The [fuzzer script](https://github.com/WebAssembly/binaryen/blob/52a90ac33416322c1be103fba6ab32764d3d1c9e/scripts/fuzz_opt.py#L174) will pick initial contents some of the time: an initial testcase to add fuzz on top of. This can be useful as often testcases are corner cases.

I had the idea that instead of randomly choosing a testcase with equal probability, we could give a higher chance for ones that are recently updated. That is, if the OS timestamp on the file is recent, increase the probability. This would have the effect of doing more fuzzing on changes that just landed, as the changes usually adjust tests.

aheejin:
This sounds good, but I think it would be less confusing to make this possible to be turned on and off with an option, and make the option disabled by default.

tlively:
I like the idea, but wouldn't this make it harder to share fuzz seeds usefully? The timestamps might be different on different people's machines.

kripken:
@aheejin @tlively Good points - yeah, this is not as simple as I thought. I guess this should be opt-in.

kripken:
Let's close this, as the points made show it is not a great idea.

Instead, in

https://github.com/WebAssembly/binaryen/pull/4111/files#diff-3c641dc74f54559f1df2b85afcac1c8e199a33edfbbf80a481167ad2e1b01f90R168-R175

something simpler is proposed: we can manually add new test files to
the list of important ones. Some time later we can then remove them.
This avoids the above issues I believe.

PR #4173, opened 2021-09-21
Select initial contents automatically in fuzzer

aheejin:
This is another attempt to address #4073. Instead of relying on the
timestamp, this examines git log to gather the list of test files added
or modified within some fixed number of days. The number of days is
currently set to 30 (= 1 month) but can be changed. This will be enabled
by `--auto-initial-contents`, which is now disabled by default.

I'm not entirely sure if this will be useful, but uploading this because
this is a simple thing to do. Feedbacks are welcome, and if it turns out
not to be useful, that's also fine.

Fixes (?) #4073.


scripts/fuzz_opt.py
```diff
@@ -178,6 +196,12 @@ def randomize_fuzz_settings():
     os.path.join('lit', 'passes', 'optimize-instructions-gc.wast'),
     os.path.join('lit', 'passes', 'inlining_splitting.wast'),
 ]
+
+if shared.options.auto_initial_contents:
+    IMPORTANT_INITIAL_CONTENTS += auto_select_recent_initial_contents()
```
kripken, line 201: It might be nice to add some logging of what files were selected here. But it would be hidden fast by other output, so maybe it should pause for user input to continue or something like that..?

scripts/test/shared.py
```diff
@@ -92,6 +92,11 @@ def parse_args(args):
         '--filter', dest='test_name_filter', default='',
         help=('Specifies a filter. Only tests whose paths contains this '
               'substring will be run'))
+    parser.add_argument(
+        '--auto-initial-contents', dest='auto_initial_contents',
+        action='store_true', default=False,
+        help='Select important initial contents automaticaly in fuzzer. '
+             'Default: disabled.')
```
kripken, line 100: This is what we discussed yesterday, I guess - yeah, it's unfortunate that the current code doesn't let a specific script add an option. I guess we'd need to make a major refactoring to manually parse the args instead of automatically, so there is a chance to add more arguments. I guess for now this is ok, but maybe someone else will have an idea. (If not, we should probably add a comment here that this is just for `fuzz_opt.py`)

kripken (commented):
Nice idea!

I think this would be useful. Some comments below though.

scripts/test/shared.py
aheejin: Yeah this is unfortunate; I tried my hand at refactoring this for.. like 10 mins and stopped. For the moment I added a comment.

aheejin:
Added the prompt.
```
aheejin@aheejin:~/binaryen$ ./scripts/fuzz_opt.py 
- Perenially-important initial contents:
  lit/passes/optimize-instructions.wast
  passes/optimize-instructions_fuzz-exec.wast

- Recently added or modified initial contents (manually selected):
  lit/passes/once-reduction.wast
  passes/remove-unused-brs_enable-multivalue.wast
  lit/passes/optimize-instructions-bulk-memory.wast
  lit/passes/optimize-instructions-ignore-traps.wast
  lit/passes/optimize-instructions-gc.wast
  lit/passes/inlining_splitting.wast

Do you want to proceed with these initial contents? (y/n) n
```
```
aheejin@aheejin:~/binaryen$ ./scripts/fuzz_opt.py --auto-initial-contents
- Perenially-important initial contents:
  lit/passes/optimize-instructions.wast
  passes/optimize-instructions_fuzz-exec.wast

- Recently added or modified initial contents (automatically selected: within last 30 days):
  lit/passes/inlining-optimizing_optimize-level=3.wast
  lit/passes/inlining_splitting.wast
  lit/passes/optimize-instructions-gc-iit.wast
  lit/passes/optimize-instructions-gc-tnh-nn.wast
  lit/passes/optimize-instructions-gc.wast
  lit/passes/optimize-instructions-bulk-memory.wast
  lit/passes/optimize-instructions-gc.wast
  passes/Oz_fuzz-exec_all-features.wast
  lit/passes/optimize-instructions-bulk-memory.wast
  lit/passes/optimize-instructions-ignore-traps.wast
  passes/remove-unused-brs_enable-multivalue.wast
  lit/passes/optimize-instructions.wast
  lit/passes/intrinsic-lowering.wast
  lit/passes/vacuum-intrinsics.wast
  heap-types.wast
  passes/Oz_fuzz-exec_all-features.wast
  lit/passes/merge-blocks.wast
  lit/passes/Oz.wast
  lit/passes/optimize-instructions.wast
  lit/wasm-split/imported-memory.wast
  lit/wasm-split/instrument-funcs.wast
  lit/wasm-split/instrument-in-memory.wast
  lit/wasm-split/profile-guided.wast
  lit/passes/once-reduction.wast
  multi-table.wast
  lit/passes/vacuum-tnh.wast
  lit/stdin-stdout.wast
  lit/stdin-stdout.wast
  lit/passes/inlining-optimizing.wast
  lit/passes/optimize-instructions-call_ref.wast
  lit/passes/optimize-instructions-gc-iit.wast
  lit/passes/optimize-instructions-gc.wast
  lit/wasm-split/basic.wast

Do you want to proceed with these initial contents? (y/n)
```

scripts/fuzz_opt.py
```diff
@@ -164,23 +164,64 @@ def randomize_fuzz_settings():
...
+    if shared.options.auto_initial_contents:
+        print(f'(automatically selected: within last {RECENT_DAYS} days):')
+        recent_contents += auto_select_recent_initial_contents()
+    else:
+        print('(manually selected):')
+        recent_contents = MANUAL_RECENT_INITIAL_CONTENTS
+    for test in recent_contents:
+        print('  ' + test)
+    print()
+    if input('Do you want to proceed with these initial contents? (y/n) ').lower() != 'y':
```
kripken, line 217: How about only adding this query when the user changed `shared.options.auto_initial_contents`? Otherwise it would make the default way to run the fuzzer slower.

kripken (approved):
Thanks! lgtm % comment

scripts/fuzz_opt.py
aheejin: > when the user changed `shared.options.auto_initial_contents`

You mean the user set `shared.options.auto_initial_contents`?

I'm.. little unsure, because I hope this can be the default in the long run. I made it not the default in case it disrupts the recent workflow on GC, but I think, making this automatic is better than expecting every PR submitter to include their test case in this fuzz_opt.py or code reviewer to suggest them to include it, so I didn't want to treat the one more special... But if you think it's inconvenient I can remove it.

And even now, if we set `RECENT_DAYS` to 14 or something, the list is not too different to your manual recent contents. Should we reduce the default to 14?
```
aheejin@aheejin:~/binaryen$ ./scripts/fuzz_opt.py 
- Perenially-important initial contents:
  lit/passes/optimize-instructions.wast
  passes/optimize-instructions_fuzz-exec.wast

- Recently added or modified initial contents (manually selected):
  lit/passes/once-reduction.wast
  passes/remove-unused-brs_enable-multivalue.wast
  lit/passes/optimize-instructions-bulk-memory.wast
  lit/passes/optimize-instructions-ignore-traps.wast
  lit/passes/optimize-instructions-gc.wast
  lit/passes/optimize-instructions-gc-iit.wast
  lit/passes/optimize-instructions-call_ref.wast
  lit/passes/inlining_splitting.wast

Do you want to proceed with these initial contents? (y/n) n
```
```
aheejin@aheejin:~/binaryen$ ./scripts/fuzz_opt.py --auto-initial-contents
- Perenially-important initial contents:
  lit/passes/optimize-instructions.wast
  passes/optimize-instructions_fuzz-exec.wast

- Recently added or modified initial contents (automatically selected: within last 14 days):
  lit/passes/inlining_splitting.wast
  lit/passes/Oz.wast
  lit/wasm-split/imported-memory.wast
  lit/passes/merge-blocks.wast
  heap-types.wast
  lit/passes/optimize-instructions-bulk-memory.wast
  lit/passes/vacuum-intrinsics.wast
  lit/passes/optimize-instructions.wast
  lit/passes/intrinsic-lowering.wast
  lit/passes/optimize-instructions-gc.wast
  lit/passes/optimize-instructions-call_ref.wast
  lit/passes/optimize-instructions-gc-iit.wast
  lit/passes/optimize-instructions-gc-tnh-nn.wast
  lit/passes/inlining-optimizing_optimize-level=3.wast
  passes/Oz_fuzz-exec_all-features.wast
  lit/passes/optimize-instructions-ignore-traps.wast
  passes/remove-unused-brs_enable-multivalue.wast

Do you want to proceed with these initial contents? (y/n) n
```

Oh and one thing; I forgot to `uniq` the auto-computed recent list before, so there were duplicates and the list was longer.

scripts/fuzz_opt.py
```diff
@@ -164,23 +164,66 @@ def randomize_fuzz_settings():
...
+        # Returns the list of test wast/wat files added or modified within the
+        # RECENT_DAYS number of days
+        # 1. Print 'git log' with changed file status and without commit message,
+        #    with commits within RECENT_DAYS number of days
+        # 2. Pick up lines in the form of
+        #    A       test/../something.wast
+        #    M       test/../something.wast
+        #    (wat extension is also included)
+        p = re.compile(r'^[AM]\stest' + os.sep + r'(.*\.(wat|wast))$')
+        log = run(['git', 'log', '--name-status', '--format=', '--date=relative', '--no-renames', f'--since="{RECENT_DAYS} days ago"'], silent=True).splitlines()
```
tlively, line 197: My git knowledge about time-based queries is rusty - will this look at `RECENT_DAYS` before the commit (or authored?) date of `HEAD` or at `RECENT_DAYS` since the current system time? I hope it's the former so that issues can be reproduced even months later by checking out the original commit.

aheejin: > will this look at `RECENT_DAYS` before the commit (or authored?)

Before the commit.

> date of `HEAD` or at `RECENT_DAYS` since the current system time?

The current system time.

tlively: Wait, so is it before the commit or before the current system time?

scripts/fuzz_opt.py
aheejin: Or, I can make this `(Y/n)`, meaning if you just enter, it will be considered as you answered `y`.

aheejin: Made `y` the default option.

scripts/fuzz_opt.py
aheejin: Sorry I maybe misunderstood your question..?

So this only considers the commit time. So if you authored a commit 6 months ago but committed it 3 days ago, it will show up. Also "%d days ago" is compared to the current time, not the time of `HEAD`. So if you run this `git log` command on a local git repo you checked out more than a month ago and haven't touched, it will turn up nothing.

tlively: Got it, thanks. Would it be possible to make it compare to the time of `HEAD` instead? That way the fuzzer will always behave the same way for a given commit and seed no matter when it is run. Otherwise I think we'll have the same reproducibility and determinism issues that previous solutions for this had.

aheejin: Oh, that's a very good point... Yeah, the result of the specific version(commit) of the fuzzer for a given seed shouldn't change. Will fix.

scripts/fuzz_opt.py
kripken: I think back when we discussed the idea either you or @tlively mentioned that reproducibility would be hurt by enabling this by default. Would we ever be able to enable this by default, that is, can we make this deterministic? The specific problem I worry about is if we look at OS times for files, can we rely on that to be identical between systems? Like if the time resolution is different, maybe the choices could be different, which would be hard to debug. Unless we are certain of determinism, I think this would be useful but should be opt-in.

aheejin: Per @tlively's suggestion, I changed the script to gather files added/modified within the fixed number of days _counting from `HEAD`_ instead of counting from the current system time: https://github.com/WebAssembly/binaryen/pull/4173/commits/a2689eae17a12747660bfefd7e64117241a56cf4

After this I think this will be deterministic given the Binaryen version (commit SHA). I'm not suggesting to change this to default right now though.

kripken: Oh, nice idea!

kripken (approved):
lgtm. If the `input` prompt gets annoying we can reconsider later.

Another thought, before we enable this by default we'd want to add a check for `git` with a clear error perhaps.

aheejin:
> lgtm. If the `input` prompt gets annoying we can reconsider later.

You can just press enter now instead of `y`, but yeah, we can remove it if it's annoying.

> Another thought, before we enable this by default we'd want to add a check for `git` with a clear error perhaps.

In which situation? For when `git` does not exist in the system?

aheejin:
Also, do you think 30 is an appropriate value for `RECENT_DAYS` default? Or might that be too high?

kripken:
> For when git does not exist in the system?

I was thinking of when the code is not from a git checkout. For a dev that's rare I guess...

> Also, do you think 30 is an appropriate value for RECENT_DAYS default? Or might that be too high?

30 days seems good to me. Erroring on the side of larger seems better to me.

scripts/test/shared.py
tlively: It would be nice if `shared.py` just created the `parser` object and added the standard options to it, but then returned it back up to the main script instead of going ahead and parsing the arguments. That would be a large refactoring, though. Maybe a simpler change for now would be to add an optional parameter to `parse_args` that takes a lambda to be applied to the `parser` object (for example to add more options) before parsing.

aheejin:
> > For when git does not exist in the system?
> 
> I was thinking of when the code is not from a git checkout. For a dev that's rare I guess...

Made it warn when `--auto-initial-contents` is specified and the directory where fuzz_opt.py is located is not within a git repo, saying it will just use the manual contents. It will look like this:
```
aheejin@aheejin:~/binaryen$ ./scripts/fuzz_opt.py --auto-initial-contents
Warning: The current directory is not a git repository, so you cannot use "--auto-initial-contents". Using the manually selected contents.

- Perenially-important initial contents:
  lit/passes/optimize-instructions.wast
  passes/optimize-instructions_fuzz-exec.wast

- Recently added or modified initial contents (manually selected):
  lit/passes/once-reduction.wast
  passes/remove-unused-brs_enable-multivalue.wast
  lit/passes/optimize-instructions-bulk-memory.wast
  lit/passes/optimize-instructions-ignore-traps.wast
  lit/passes/optimize-instructions-gc.wast
  lit/passes/optimize-instructions-gc-iit.wast
  lit/passes/optimize-instructions-call_ref.wast
  lit/passes/inlining_splitting.wast

Do you want to proceed with these initial contents? (Y/n)
```

scripts/test/shared.py
aheejin: That's an option, but note that this is a preexisting problem; fuzz_opt.py only used two of the default options and others were ignored. Other scripts also use a subset of options each. So `parse_args` was already a big superset of all options taken by all scripts, so I thought it'd be weird if we make a parameter only for this option.

Also, even though this option is shown when you run `--help`, scripts under scripts/ directory are mostly used by developers, so the normal users are not gonna see it when they run things like `wasm-opt`. Actually there is an option to hide an option so that it doesn't show up in `--help`, but then it will need a separate comment on what it does, so thought it wasn't worth it.

And yeah refactoring seems to be a daunting task..

kripken:
> Made it warn when --auto-initial-contents is specified

sgtm, thanks.

tlively (approved):

kripken:
An issue we missed here, that I noticed now while reducing something:  The input prompt breaks reduction since if I copy-paste the reduction command when the fuzzer finds a bug, it times out, I assume because it runs the fuzzer script internally (with a seed and a given wasm) and then it hangs on waiting for the input prompt.

aheejin:
Ah I see. Then it will be a problem for both with `--auto-initial-contents` and without, right? Should we just remove the prompt? That way the list will be hard to read but we can't help it I guess...

tlively:
I would be fine with removing the prompt.

kripken:
Maybe we could avoid showing the prompt when the user provides a fuzzer seed?

aheejin:
> Maybe we could avoid showing the prompt when the user provides a fuzzer seed?

Sounds good!

PR #4179, opened 2021-09-22
Use UniqueDeferringQueue in Precompute

kripken:
This makes the code a little shorter, and more consistent with other
similar queue situations we have in the codebase.

### DIFF
--- a/src/passes/Precompute.cpp
+++ b/src/passes/Precompute.cpp
@@ -33,6 +33,7 @@
 #include <ir/properties.h>
 #include <ir/utils.h>
 #include <pass.h>
+#include <support/unique_deferring_queue.h>
 #include <wasm-builder.h>
 #include <wasm-interpreter.h>
 #include <wasm.h>
@@ -282,18 +283,16 @@ private:
     localGraph.computeInfluences();
     // prepare the work list. we add things here that might change to a constant
     // initially, that means everything
-    std::unordered_set<Expression*> work;
+    UniqueDeferredQueue<Expression*> work;
     for (auto& pair : localGraph.locations) {
       auto* curr = pair.first;
-      work.insert(curr);
+      work.push(curr);
     }
     // the constant value, or none if not a constant
     std::unordered_map<LocalSet*, Literals> setValues;
     // propagate constant values
     while (!work.empty()) {
-      auto iter = work.begin();
-      auto* curr = *iter;
-      work.erase(iter);
+      auto* curr = work.pop();
       // see if this set or get is actually a constant value, and if so,
       // mark it as such and add everything it influences to the work list,
       // as they may be constant too.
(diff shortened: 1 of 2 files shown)


src/support/unique_deferring_queue.h
```diff
@@ -23,6 +23,7 @@
 #ifndef wasm_support_unique_deferring_queue_h
 #define wasm_support_unique_deferring_queue_h
 
+#include <cassert>
```
aheejin, line 26: Why was this added?

src/passes/Precompute.cpp
```diff
@@ -282,18 +283,16 @@ struct Precompute
...
+      work.push(curr);
     }
     // the constant value, or none if not a constant
     std::unordered_map<LocalSet*, Literals> setValues;
     // propagate constant values
     while (!work.empty()) {
-      auto iter = work.begin();
-      auto* curr = *iter;
-      work.erase(iter);
+      auto* curr = work.pop();
```
aheejin, line 295: Is this the same semantics as the `unordered_set::erase`? It looks if `UniqueDeferredQueue` contains a count for each item and the item is removed only if the count reaches zero, whereas in `unordered_set` the item is erased right away. 

At the same time I don't think I have a full grasp on when `UniqueDeferredQueue` should be used...

src/support/unique_deferring_queue.h
kripken: To fix compilation. The file uses assert, and by chance the other users had it included anyhow, so this wasn't a problem before.

src/passes/Precompute.cpp
kripken: The semantics should be the same, yes. Erase removes the item from a set, and a `UniqueDeferredQueue` has a queue of unique items - so it is effectively like a set in that popping an item removes it and there cannot be others like it.

In general `UniqueDeferredQueue` is nice because it has a simpler API (replacing the 3 lines here with one), and also it is more "deterministic" in that using an `unordered_set` will cause the order to change based on the user's c++ standard hashing. A `UniqueDeferredQueue` operates as a queue, so the order is always the same (with clear semantics for how to handle a push of an existing item - it is deferred to the end; it could be either way).

aheejin: > The semantics should be the same, yes. Erase removes the item from a set, and a `UniqueDeferredQueue` has a queue of unique items - so it is effectively like a set in that popping an item removes it and there cannot be others like it.

So it is guaranteed there are no duplicates pushed to the queue in this code? If there are duplicates the semantics can be different right?

kripken: It handles duplicates itself. If the someone pushes `A, B, C, D, B` then `B` will still only be popped once. In this example the first pop will be `A`, then `C` (*not* `B`, which is deferred), then `D`, and only then `B`, and then the queue is empty.

aheejin: Ah I see. I was misunderstanding how `UniqueDeferredQueue` works.

aheejin (approved):

PR #4183, opened 2021-09-23
Add feature flag for relaxed-simd

ngzhian:


aheejin:
There seem to be a few more tests that tests every feature flag:

https://github.com/WebAssembly/binaryen/blob/main/test/unit/test_features.py (But it looks this one maybe actually require some instruction support?)
https://github.com/WebAssembly/binaryen/blob/main/test/example/c-api-kitchen-sink.c
https://github.com/WebAssembly/binaryen/blob/main/test/binaryen.js/kitchen-sink.js

ngzhian:
Thanks for the pointers Heejin, for the test_features.py, we don't actually have any relaxed simd instruction yet, so i just added to the test that emits all features.

tlively (approved):
Nice! I wonder if at some point we will want a mechanism like the `SIMDLevel` we use in Clang and LLVM, but I don't think we need to figure that out now.

PR #4184, opened 2021-09-23
Precompute: Only run a single LocalGraph iteration

kripken:
Precompute has a mode in which it propagates results from `local.set`s to
`local.get`s. That constructs a LocalGraph which is a non-trivial amount of
work. We used to run multiple iterations of this, but investigation shows that
such opportunities are extremely rare, as doing just a single propagation
iteration has no effect on the entire emscripten benchmark suite, nor on
j2cl output. Furthermore, we run this pass twice in the normal pipeline (once
early, once late) so even if there are such opportunities they may be
optimized already. And, `--converge` is a way to get additional iterations of
all passes if a user wants that, so it makes sense not to costly work for more
iterations automatically.

In effect, 99.99% of the time before this pass we would create the LocalGraph
twice: once the first time, then a second time only to see that we can't
actually optimize anything further. This PR makes us only create it once, which
makes `precompute-propagate` 10% faster on j2cl and even faster on other things
like poppler (33%) and LLVM (29%).

See the change in the test suite for an example of a case that does require
more than one iteration to be optimized. Note that even there, we only manage
to get benefit from a second iteration by doing something that overlaps with
another pass (optimizing out an if with condition 0), which shows even more
how unnecessary the extra work was.

See #4165

### DIFF
--- a/src/passes/Precompute.cpp
+++ b/src/passes/Precompute.cpp
@@ -114,8 +114,6 @@ struct Precompute
 
   GetValues getValues;
 
-  bool worked;
-
   void doWalkFunction(Function* func) {
     // Walk the function and precompute things.
     super::doWalkFunction(func);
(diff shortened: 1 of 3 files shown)


src/passes/Precompute.cpp
```diff
@@ -390,6 +395,7 @@ struct Precompute
           for (auto* set : localGraph.getInfluences[get]) {
             work.push(set);
           }
+          worked = true;
```
tlively, line 398: Can we have `optimizeLocals` return this bool rather than using separate state?

kripken: Good idea, done.

tlively (approved):

PR #4185, opened 2021-09-24
[Fuzzer] Don't prompt user when seed is given

aheejin:
Fuzzer shows the initial contents and prompts the user if they want to
proceed after #4173, but when the fuzzer is used within a script called
from `wasm-reduce`, we shouldn't pause for the user input. This shows
the prompt only when there is no seed given.

To do that, we now initialize the important initial contents from
`main`. We used to assign those variables before we start `main`.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -44,6 +44,8 @@ INPUT_SIZE_MAX = 5 * INPUT_SIZE_MEAN
 
 PRINT_WATS = False
 
+given_seed = None
+
 
 # utilities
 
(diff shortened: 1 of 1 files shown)


scripts/fuzz_opt.py
```diff
@@ -238,15 +240,18 @@ def is_git_repo():
...
     print()
-    ret = input('Do you want to proceed with these initial contents? (Y/n) ').lower()
-    if ret != 'y' and ret != '':
-        sys.exit(1)
-
-    initial_contents = FIXED_IMPORTANT_INITIAL_CONTENTS + recent_contents
-    return [os.path.join(shared.get_test_dir('.'), t) for t in initial_contents]
 
+    # We prompt the user only when there is no seed given. This fuzz_opt.py is
+    # often used with seed in a scirpt called from wasm-reduce, in which case we
```
kripken, line 245:
```suggestion
    # often used with seed in a script called from wasm-reduce, in which case we
```

kripken (approved):
Thanks!

PR #4191, opened 2021-09-27
Disable partial inlining by default and add a flag for it.

kripken:
Locally I saw a 10% speedup on j2cl but reports of regressions have
arrived, so let's disable it for now pending investigation. The option added
here should make it easy to experiment.


src/tools/optimization-options.h
```diff
@@ -171,6 +171,16 @@ struct OptimizationOptions : public ToolOptions {
...
            })
+      .add("--partial-inlining-ifs",
+           "-pii",
+           "Number of ifs allowed in partial inlining (zero means partial "
+           "inlining is disabled) (default: " +
+             std::to_string(InliningOptions().partialInliningIfs) + ')',
+           Options::Arguments::One,
+           [this](Options* o, const std::string& argument) {
+             passOptions.inlining.partialInliningIfs =
+               static_cast<Index>(atoi(argument.c_str()));
```
aheejin, line 182:
```suggestion
               static_cast<Index>(std::stoi(argument));
```

aheejin (approved):

PR #4193, opened 2021-09-29
Exclude test/ from clang-tidy

aheejin:
We don't keep files within test/ clang-tidy compliant, so including this
directory can generate CI errors when test files change.

### DIFF
--- a/scripts/clang-tidy-diff.sh
+++ b/scripts/clang-tidy-diff.sh
@@ -24,15 +24,16 @@ function realpath() {
 
 CLANG_DIR=$(dirname $(dirname $(realpath $CLANG_TIDY)))
 CLANG_TIDY_DIFF=$CLANG_DIR/share/clang/clang-tidy-diff.py
+ARG="-quiet -p1 -regex=src/.*"
 if [ ! -e "$CLANG_TIDY_DIFF" ]; then
   echo "Failed to find clang-tidy-diff.py ($CLANG_TIDY_DIFF)"
   exit 1
 fi
-TIDY_MSG=$(git diff -U0 $BRANCH... | $CLANG_TIDY_DIFF -quiet -p1 2> /dev/null)
+TIDY_MSG=$(git diff -U0 $BRANCH... | $CLANG_TIDY_DIFF $ARG 2> /dev/null)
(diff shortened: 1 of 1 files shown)


sbc100:
I general I would say it a good idea to keep your test code compliant and readable.  Is there some reason not to with binaryen?

aheejin:
Actually the only thing we check from clang-tidy is braces (+ basic errors from clang). When I first added .clang-tidy, my plan was to add more rules and checks, but it kind of fizzled out... We can pick it up anytime though. But anyway there are many lines that don't use `{}` for `if`s and `for`s, which we can fix, but we have to do that manually.

The other reason is clang-tidy errors out on lines like
```
#include <emscripten.h>
```
in some files. But at the same time I wonder if those files are even being used. If we can remove those files maybe this problem goes away. I asked the question here: https://github.com/WebAssembly/binaryen/pull/4192#issuecomment-929733703

sbc100:
> Actually the only thing we check from clang-tidy is braces (+ basic errors from clang). When I first added .clang-tidy, my plan was to add more rules and checks, but it kind of fizzled out... We can pick it up anytime though. But anyway there are many lines that don't use `{}` for `if`s and `for`s, which we can fix, but we have to do that manually.
> 
> The other reason is clang-tidy errors out on lines like
> 
> ```c++
> #include <emscripten.h>
> ```
> 
> in some files. But at the same time I wonder if those files are even being used. If we can remove those files maybe this problem goes away. I asked the question here: [#4192 (comment)](https://github.com/WebAssembly/binaryen/pull/4192#issuecomment-929733703)

I didn't know clang-tidy required the files to actually compile (i.e. find all the includes)... clang-format doesn't need that ability but I guess clang-tidy is different.

In that case I support excluding those files under test from clang-tidy... since there is no cmake command to build them and it doesn't seem worth while to add a cmake target for them.

sbc100 (approved):

aheejin:
It looks there is no way to easily exclude a directory or a file from clang-tidy. The way I did here is actually not an option of clang-tidy but [an option of the wrapper script](https://github.com/llvm/llvm-project/blob/f9b68304a22ef82d827d282dd9014efc2e7ac875/clang-tools-extra/clang-tidy/tool/clang-tidy-diff.py#L128-L130) provided from LLVM.

PR #4195, opened 2021-09-29
Implement table.get

kripken:
Adds the part of the spec test suite that this passes (without `table.set` we
can't do it all).

We'll need this eventually anyhow, and in the short term I have some ideas
for GC opts that may use it.


kripken:
Did we change something in clang-tidy recently? The errors here do not make sense to me...

aheejin:
Sorry, I think #4193 might be the culprit. It tried to exclude test/ from clang-tidy by giving it the regex `src/*`, but it looks I accidentally overrided the existing extension-based regex there... Will fix soon.

tlively:
Yep, in https://github.com/WebAssembly/binaryen/pull/4193, @aheejin, do you know what could be going on here?

src/ir/effects.h
```diff
@@ -589,6 +589,11 @@ class EffectAnalyzer {
     void visitRefIs(RefIs* curr) {}
     void visitRefFunc(RefFunc* curr) {}
     void visitRefEq(RefEq* curr) {}
+    void visitTableGet(TableGet* curr) {
+      // TODO: track readsTable/writesTable, like memory?
```
tlively, line 593: Yep, tracking that sounds important. Want to model this as reading from memory for now? `call_indirect` should probably be modeled as reading from memory to match if so.

src/wasm-interpreter.h
```diff
@@ -2528,7 +2532,37 @@ template<typename GlobalManager, typename SubType> class ModuleInstanceBase {
...
+    ExternalInterface* interface;
+    // The name the table has in that interface.
+    Name name;
+  };
+
+  TableInterfaceInfo getTableInterfaceInfo(Name name) {
+    auto* table = wasm.getTable(name);
+    if (table->imported()) {
+      auto& importedInstance = linkedInstances.at(table->module);
+      auto* tableExport = importedInstance->wasm.getExport(table->base);
```
tlively, line 2550: Do we need to handle link errors here or can we assume that the various modules link correctly?

src/wasm/wasm-validator.cpp
```diff
@@ -2031,6 +2032,20 @@ void FunctionValidator::visitRefEq(RefEq* curr) {
                   "ref.eq's right argument should be a subtype of eqref");
 }
 
+void FunctionValidator::visitTableGet(TableGet* curr) {
+  shouldBeTrue(getModule()->features.hasReferenceTypes(),
+               curr,
+               "table.eq requires reference types to be enabled");
```
tlively, line 2038:
```suggestion
               "table.get requires reference types to be enabled");
```

test/table-operations.wast
```diff
@@ -0,0 +1,25 @@
+(module
```
tlively, line 1: Can this be a lit test instead?

kripken: I believe we haven't used lit tests for this because we depend on this type of test to do both text and binary roundtripping? But if there's a way, maybe we could convert them...

src/ir/effects.h
kripken: Until we get `table.set` this doesn't seem like a potential problem, though? At that time, I think we should add `readsTable` etc. I suppose.

src/wasm-interpreter.h
kripken: This is after validation and linking, so we can assume those things succeeded.

(This code is just moved around into a utility function - no known issues with it AFAIK.)

test/table-operations.wast
tlively: We haven't focused on porting the existing tests of this type, but for new tests it would still be good to use lit. Using multiple run lines should be able to cover both text and binary round trips.

src/ir/effects.h
tlively: Well `call_indirect` reads the table, but I guess in practice it doesn't matter since it also does a call. Not doing anything until we also have `table.set` sgtm, then.

tlively (approved):
LGTM!

aheejin:
Clang-tidy bug is fixed in #4197. Sorry for the inconvenience.

test/table-operations.wast
kripken: I'm afraid it's not obvious to me how to do all that we need with a lit test here. Is there an existing lit test that does it?

(And, would doing this in a lit test be compatible with auto-updating it? If not, I'd rather not...)

tlively: I think something like this should do approximately what the test runner currently does. It would also work well with the auto update script.
```
; RUN: wasm-as %s -all -g -o %t.wasm 
; RUN: wasm-dis %t.wasm -all -o %t.wast
; RUN: wasm-opt %t.wast -all -o %t.opt.wasm
; RUN: cat %t.wast | FileCheck %s
; RUN: cat %t.opt.wast | FileCheck %s --check-prefix CHECK-OPT
```

kripken: Thanks... I still can't get it to work, though. It does read and write the file, but no `CHECK` stuff is emitted - it makes no changes at all. I spent some time now trying to figure out what's different in this file from others that do work (like having `--all-items` or not in the first line, I tried both), but didn't find anything useful. Rather than start to debug the python scripts here, I think I'll leave this for now...

tlively: Oops, there were a few issues with that. Can you try this?

```
;; RUN: wasm-as %s -all -g -o %t.wasm
;; RUN: wasm-dis %t.wasm -all -o %t.wast
;; RUN: wasm-opt %t.wast -all -o %t.opt.wasm
;; RUN: cat %t.wast | filecheck %s
;; RUN: cat %t.opt.wast | filecheck %s --check-prefix CHECK-OPT
```

Note the `;;` instead of `;` at the beginning of each line and the `filecheck` instead of `FileCheck`.

kripken: Ah, I did fix all those issues except for the lowercase/uppercase issue with `filecheck`... now it works, thanks! Why doesn't it error though if the case is wrong?

tlively: The update script doesn't know that `FileCheck` is not a real command, but it also doesn't recognize it as `filecheck`. I can make a PR to have it detect this error as a quality of life improvement.

PR #4196, opened 2021-09-29
Fuzzer: Add --auto-initial-contents automatically to reduction script

kripken:
Without this, the reduction script won't use the same initial contents,
which means it may see different things.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -1343,6 +1343,9 @@ on valid wasm files.)
                 original_wasm = os.path.abspath('original.wasm')
                 shutil.copyfile('a.wasm', original_wasm)
                 # write out a useful reduce.sh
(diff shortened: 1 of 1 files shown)


scripts/fuzz_opt.py
```diff
@@ -1383,6 +1386,7 @@ def randomize_opt_flags():
                   ''' % {'wasm_opt': in_bin('wasm-opt'),
                          'bin': shared.options.binaryen_bin,
                          'seed': seed,
+                         'auto_init': auto_init,
```
tlively, line 1389: It would be good to switch to using f-strings here, perhaps in a separate PR.

tlively (approved):

aheejin (approved):
Oh sorry.. Thanks for the fix.

PR #4197, opened 2021-09-30
Fix clang-tidy regex

aheejin:
While making clang-tidy CI only check for src/, #4193 accidentally
overrided the existing default regex that checks only c/cpp/h... files:
https://github.com/llvm/llvm-project/blob/8971b99c8387f3daf2e802956f2688b3b77335a4/clang-tools-extra/clang-tidy/tool/clang-tidy-diff.py#L131-L132

This fixes the problem by appending all those extensions after `src/`.

### DIFF
--- a/scripts/clang-tidy-diff.sh
+++ b/scripts/clang-tidy-diff.sh
@@ -24,7 +24,7 @@ function realpath() {
 
 CLANG_DIR=$(dirname $(dirname $(realpath $CLANG_TIDY)))
 CLANG_TIDY_DIFF=$CLANG_DIR/share/clang/clang-tidy-diff.py
-ARG="-quiet -p1 -regex=src/.*"
+ARG="-quiet -p1 -iregex=src/.*\.(cpp|cc|c\+\+|cxx|c|cl|h|hpp|m|mm|inc)"
 if [ ! -e "$CLANG_TIDY_DIFF" ]; then
   echo "Failed to find clang-tidy-diff.py ($CLANG_TIDY_DIFF)"
   exit 1


kripken (approved):
Thanks!

scripts/clang-tidy-diff.sh
```diff
@@ -24,7 +24,7 @@ function realpath() {
 
 CLANG_DIR=$(dirname $(dirname $(realpath $CLANG_TIDY)))
 CLANG_TIDY_DIFF=$CLANG_DIR/share/clang/clang-tidy-diff.py
-ARG="-quiet -p1 -regex=src/.*"
+ARG="-quiet -p1 -iregex=src/.*\.(cpp|cc|c\+\+|cxx|c|cl|h|hpp|m|mm|inc)"
```
tlively, line 27: Does the `\` need to be escaped?

aheejin: Why?

tlively: I may misunderstand bash, but doesn't it interpret backslash interpreted as an escape character, so clang-tidy will see `-iregex=src/.*.(cpp|...`? This happens to work, but I think it would also include files like `src/not_a_real_cpp`.

aheejin: I see. I checked and it seems it's fed to clang-tidy-diff.py without disappearing, and I also tested with src/not_a_real_cpp and it didn't show up.

tlively: Thanks for checking!

tlively: Oh yes, I misunderstood bash. From the reference manual:

> The backslash retains its special meaning only when followed by one of the following characters: ‘$’, ‘`’, ‘"’, ‘\’, or newline. Within double quotes, backslashes that are followed by one of these characters are removed. Backslashes preceding characters without a special meaning are left unmodified.

PR #4198, opened 2021-10-01
Detect erroneous use of "FileCheck" in update script

tlively:
The upstream name of the command in LLVM is FileCheck, but we use a Python
package called filecheck instead and using `FileCheck` in lit tests will not
work. To make catching this common error easier, add a warning about it in
scripts/update_lit_checks.py.

### DIFF
--- a/scripts/update_lit_checks.py
+++ b/scripts/update_lit_checks.py
@@ -146,7 +146,7 @@ def get_command_output(args, test, lines, tmp):
         if len(commands) > 1 and commands[1].startswith('filecheck '):
             filecheck_cmd = commands[1]
             commands = commands[:1]
-        if len(commands) > 1 and commands[1].startswith('FileCheck '):
+        elif len(commands) > 1 and commands[1].startswith('FileCheck '):
             warn('`FileCheck` is not a known command. '
                  'Did you mean to use `filecheck` instead?')
 
--- a/scripts/update_lit_checks.py
+++ b/scripts/update_lit_checks.py
@@ -44,7 +44,7 @@ ITEM_RE = re.compile(r'(^\s*)\((' + '|'.join(items) + r')\s+(\$?[^\s()]*).*$',
 
 
 def warn(msg):
-    print(f'WARNING: {msg}', file=sys.stderr)
+    print(f'warning: {msg}', file=sys.stderr)
 
 
 def itertests(args):
@@ -146,6 +146,9 @@ def get_command_output(args, test, lines, tmp):
         if len(commands) > 1 and commands[1].startswith('filecheck '):
             filecheck_cmd = commands[1]
             commands = commands[:1]
+        if len(commands) > 1 and commands[1].startswith('FileCheck '):
+            warn('`FileCheck` is not a known command. '
+                 'Did you mean to use `filecheck` instead?')
 
         prefix = ''
         if filecheck_cmd.startswith('filecheck '):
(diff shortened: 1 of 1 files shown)


scripts/update_lit_checks.py
```diff
@@ -146,6 +146,9 @@ def get_command_output(args, test, lines, tmp):
         if len(commands) > 1 and commands[1].startswith('filecheck '):
             filecheck_cmd = commands[1]
             commands = commands[:1]
+        if len(commands) > 1 and commands[1].startswith('FileCheck '):
+            warn('`FileCheck` is not a known command. '
+                 'Did you mean to use `filecheck` instead?')
```
kripken, line 152: could these be a chain if if-elif ending in an error? Or is the set of commands we need too long to be written out?

tlively: The set of commands is unbounded. Every command except for `filecheck` is passed through directly to the underlying system. So the only "known command" is `filecheck`, but there are no invalid commands.

kripken: Doesn't the underlying system error, though, if that command is wrong? Or do we silently swallow errors?

tlively: Right now we're silently swallowing errors. I'll push a fix for that as well.

kripken (approved):
Thanks, I think showing those errors could be very useful.

Lgtm as it is though, depending on if you want to split this up or not.

tlively:
After that last commit, I don't think there's anything else to do to show errors. We already use `subprocess.check_output`, which throws if there is an underlying error. The actual problem was that we only ever run the first command in a pipe sequence, so in your case `FileCheck` was never run because it was the second command in the sequence. The new warning should cover that, but I don't think there's any more information we can give.

kripken:
> The actual problem was that we only ever run the first command in a pipe sequence, so in your case FileCheck was never run because it was the second command in the sequence.

Huh, I'm confused then. Fixing the typo `FileCheck => filecheck` make it go from silent nothing to actually getting output through filecheck. If you're saying that it never executed before, then it sounds like fixing the typo made it start to execute, is that right? (how?)

tlively:
No, this script never actually executes the `filecheck` commands (or any commands besides the first in a pipe sequnce). Fixing the typo allowed this script to recognize that it should generate checks based on the output of the command that is piped into the `filecheck` command, whereas with the typo the script ran that preceding command but did not generate checks for it. That preceding instruction was correct all along, so there were no errors to report.

kripken:
Oh, is ` | filecheck` magic syntax? So it's not a real pipe that is executed? Nothing runs `filecheck`, it's just a signal to add the checks then.

If that's right, what other values are allowed in place of `filecheck`? I'm having a hard time squaring that with the fact that we have an unbounded list of possible things that we "pipe (but not really)" to. That is, if we really did pipe, it could be anything on the system, but if it's magical internal syntax, the magic must have a list...?

Sorry for all the questions. If you're annoyed feel free to land...

tlively:
No problem! This script is kind of weird about this. When we run `lit`, `| filecheck` is not magic; it actually runs filecheck. But for this script, we actually need to read the output that would be piped into filecheck, not run filecheck itself. So yes, `| filecheck` is magic syntax for the purposes of this script.

No other values are allowed in place of `filecheck`. In particular, this script doesn't really support pipes at all. If you have a run line like `cmd1 | cmd2 | cmd3 | ...`, this script will only run `cmd1`. Previously it silently dropped `cmd2` and friends, but after this PR it at least prints a warning about it.

kripken:
> So yes, | filecheck is magic syntax for the purposes of this script.

Ah, now I see... so it's magic in the auto-updater, but not magic in lit... hence the difference. Thanks!

lgtm

PR #4201, opened 2021-10-01
Implement standalone nominal types

tlively:
These new nominal types do not depend on the global type sytem being changed
with the --nominal flag. Instead, they can coexist with the existing
equirecursive structural types, as required in the new milestone 4 spec. This PR
implements subtyping, upper bounding, canonicalizing, and other type operations
but using the new types in the parsers and elsewhere in Binaryen is left to a
follow-on PR.

### DIFF
--- a/src/wasm-type.h
+++ b/src/wasm-type.h
@@ -382,7 +382,18 @@ public:
   const Struct& getStruct() const;
   Array getArray() const;
 
+  // Return whether there is a nontrivial (i.e. non-basic) nominal supertype and
+  // store it in `out` if there is. Nominal types (in the sense of isNominal,
+  // i.e. Milestone 4 nominal types) may always have supertypes and other types
+  // may have supertypes in `TypeSystem::Nominal` mode but not in
+  // `TypeSystem::Equirecursive` mode.
   bool getSuperType(HeapType& out) const;
+
+  // Whether this is a nominal type in the sense of being a GC Milestone 4
+  // nominal type. Although all non-basic HeapTypes are nominal in
+  // `TypeSystem::Nominal` mode, this will still return false unless the type is
+  // specifically constructed as a Milestone 4 nominal type.
+  bool isNominal() const;
 
   constexpr TypeID getID() const { return id; }
   constexpr BasicHeapType getBasic() const {
@@ -576,6 +587,12 @@ struct TypeBuilder {
   // equirecursive mode.
   void setSubType(size_t i, size_t j);
 
+  // Make this type nominal in the sense of the Milestone 4 GC spec, independent
+  // of the current TypeSystem configuration. `supertype` is the index of the
+  // nontrivial supertype, or `NoSupertype` if there is no nontrivial supertype.
+  enum : size_t { NoSupertype = -1ull };
+  void setNominal(size_t i, size_t supertype);
+
   // Returns all of the newly constructed heap types. May only be called once
   // all of the heap types have been initialized with `setHeapType`. In nominal
   // mode, all of the constructed HeapTypes will be fresh and distinct. In
@@ -614,6 +631,11 @@ struct TypeBuilder {
       builder.setSubType(index, other.index);
       return *this;
     }
+    Entry& setNominal(size_t supertype) {
+      builder.setNominal(index, supertype);
+      return *this;
+    }
+    Entry& setNominal(Entry supertype) { return setNominal(supertype.index); }
   };
 
   Entry operator[](size_t i) { return Entry{*this, i}; }
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -104,6 +104,7 @@ struct HeapTypeInfo {
   // Otherwise, the type definition tree is still being constructed via the
   // TypeBuilder interface, so hashing and equality use pointer identity.
   bool isFinalized = true;
+  bool isNominal = false;
   // In nominal mode, the supertype of this HeapType, if it exists.
   HeapTypeInfo* supertype = nullptr;
   enum Kind {
@@ -590,6 +591,7 @@ bool TypeInfo::operator==(const TypeInfo& other) const {
 HeapTypeInfo::HeapTypeInfo(const HeapTypeInfo& other) {
   kind = other.kind;
   supertype = other.supertype;
+  isNominal = other.isNominal;
   switch (kind) {
     case BasicKind:
       new (&basic) auto(other.basic);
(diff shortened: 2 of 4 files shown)


kripken:
I think I've misunderstood the spec, then. In what sense do they coexist? I've assumed we'd just flip the default to nominal at some point and that's it.

Specifically, I think I see this PR  annotate types as "isNominal" or not. But how can we tell?

tlively:
The nominal types can coexist and be differentiated from the equirecursive types because they have different type codes in the binary format and different names in the text format. The new spec is designed to allow for a graceful transition and also to experiment with things like equirecursive function types and nominal other types.

kripken:
Is https://docs.google.com/document/d/1afthjsL_B9UaMqCA5ekgVmOm75BVFu6duHNsN9-gnXw/edit# the spec? If so, where are the text format changes there?

tlively:
Yes, that's the spec. Look for `func_subtype`, `struct_subtype`, and `array_subtype`. Those are the new nominal type constructors. The existing `func`, `struct`, and `array` are unchanged and still equirecursive.

kripken:
But those just create subtypes - they are basically another notation for `(extends $name)`, aren't they? That is, if `A -> B`, how can I define `A` as nominal if it is not a subtype of anything?

Or are you saying that's not possible, and we still have type canonicalization for types with no parent? (that would be bad for optimization)

tlively:
From the document:

> the supertype may be dataref (-0x19) for struct_subtype and array_subtype, or funcref (-0x10) for func_subtype, to indicate "no nontrivial supertype", i.e. the type created this way is a root of an inheritance tree. The type is still a subtype of dataref and eqref (for struct and array types) or funcref (for function types) and anyref.

So you still use e.g. `struct_subtype` to declare a nominal struct even if it doesn't have a meaningful supertype. To do that, you would pass the sentinel value `data` in the supertype position. (This was chosen because in fact the immediate supertype of such a struct would be `data`.)

kripken:
Oh, thanks, I missed that entirely...

src/wasm-type.h
```diff
@@ -576,6 +587,12 @@ struct TypeBuilder {
   // equirecursive mode.
   void setSubType(size_t i, size_t j);
 
+  // Make this type nominal in the sense of the Milestone 4 GC spec, independent
+  // of the current TypeSystem configuration. `supertype` is the index of the
+  // nontrivial supertype, or `NoSupertype` if there is no nontrivial supertype.
+  enum : size_t { NoSupertype = static_cast<size_t>(-1) };
+  void setNominal(size_t i, size_t supertype);
```
kripken, line 594: Can we use `setSubType` (right above) to define subtyping relationships in all modes, instead of adding a new function? I guess we need a new one to mark as "milestone 4" but that could take no parameters.

src/wasm/wasm-type.cpp
```diff
@@ -2052,6 +2082,11 @@ bool FiniteShapeEquator::eq(const TypeInfo& a, const TypeInfo& b) {
 }
 
 bool FiniteShapeEquator::eq(const HeapTypeInfo& a, const HeapTypeInfo& b) {
+  if (a.isNominal != b.isNominal) {
+    return false;
+  } else if (a.isNominal) {
+    return &a == &b;
+  }
```
kripken, line 2089: To be sure I follow, there is no old code for typesystem==nominal here because we never even got here in the first place?

src/wasm/wasm-type.cpp
```diff
@@ -2213,9 +2248,11 @@ struct TypeBuilder::Impl {
       // value.
       info = std::make_unique<HeapTypeInfo>(Signature());
       set(Signature());
+      initialized = false;
```
kripken, line 2251: This change looks unrelated... but is it?

src/wasm/wasm-type.cpp
```diff
@@ -2807,16 +2868,18 @@ void ShapeCanonicalizer::translatePartitionsToTypes() {
...
     };
     HeapType root = asHeapType(info);
     ChildUpdater(*this).walkRoot(&root);
+
+    // If this is a nominal type, we may need to update its supertype as well.
+    if (info->supertype) {
+      HeapType heapType(uintptr_t(info->supertype));
+      replaceHeapType(&heapType);
+      info->supertype = getHeapTypeInfo(heapType);
+    }
```
kripken, line 2882: Why is there no code for typesystem==nominal around here? (i.e., why was this not an issue before)

test/example/type-builder-nominal-new.cpp
```diff
@@ -0,0 +1,435 @@
+#include <cassert>
+#include <iostream>
+
+#include "wasm-type.h"
+
+/* This file is identical to type-builder-nominal.cpp, except that instead of
+ * setting the global type system to be nominal, it instead individually sets
```
kripken, line 7: Diffing the output, there are a few changes, like:
```diff
 Before setting heap types:
-(ref $sig) => (ref [T](func))
-(ref $struct) => (ref [T](func))
-(ref $array) => (ref [T](func))
-(ref null $array) => (ref null [T](func))
-(rtt 0 $array) => (rtt 0 [T](func))
+(ref $sig) => [T](ref [T](func))
+(ref $struct) => [T](ref [T](func))
+(ref $array) => [T](ref [T](func))
+(ref null $array) => [T](ref null [T](func))
+(rtt 0 $array) => [T](rtt 0 [T](func))
```
Is that intentional to indicate a milestone4 type?

kripken (commented):
Nice work!

Please fuzz this before landing.

src/wasm-type.h
tlively: Yes, that seems cleaner. I'll make that change.

src/wasm/wasm-type.cpp
tlively: That's right. `FiniteShapeEquator` and `FiniteShapeHasher` are only used as part of equirecursive canonicalization, so they're skipped entirely in `--nominal` mode.

src/wasm/wasm-type.cpp
tlively: No, this is just a drive by fix. The `initialized` field is checked at `build()` time to make sure that there are no user-uninitialized entries, but we were accidentally setting it to `true` by default because we used `set` to initialize the entries to their placeholder values. The reason we don't have a test for this is that we don't have a good way to test fatal errors in the C++ API.

src/wasm/wasm-type.cpp
tlively: `ShapeCanonicalizer` is also only used for equirecursive canonicalization, so `--nominal` mode never uses this code.

test/example/type-builder-nominal-new.cpp
tlively: The `[T]` signifies a temporary type or heap type. In `--nominal` mode, the "temporary" types constructed by a `TypeBuilder` are actually allocated in the global store, so they are not truly temporary. But in equirecursive canonicalization, temporary types are in fact initially temporary, and that difference is reflected here.

kripken (approved):

PR #4207, opened 2021-10-04
Optimize call_ref+table.get => call_indirect

kripken:
Rather than load from the table and call that reference, call using the table.

Am I missing some corner case that prevents doing this? (Would be sad if so, as I
hope to build on this to do more devirtualization.)

### DIFF
--- a/test/lit/passes/optimize-instructions-call_ref.wast
+++ b/test/lit/passes/optimize-instructions-call_ref.wast
@@ -5,21 +5,27 @@
 ;; remove-unused-names is to allow fallthrough computations to work on blocks
 
 (module
- ;; CHECK:      (type $none_=>_i32 (func (result i32)))
- (type $none_=>_i32 (func (result i32)))
-
  ;; CHECK:      (type $i32_i32_=>_none (func (param i32 i32)))
  (type $i32_i32_=>_none (func (param i32 i32)))
+
+ ;; CHECK:      (type $none_=>_i32 (func (result i32)))
+ (type $none_=>_i32 (func (result i32)))
 
  ;; CHECK:      (type $none_=>_none (func))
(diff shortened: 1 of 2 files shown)


test/lit/passes/optimize-instructions-call_ref.wast
```diff
@@ -5,21 +5,27 @@
...
  (type $data_=>_none (func (param (ref data))))
 
- ;; CHECK:      (type $i32_=>_none (func (param i32)))
+ ;; CHECK:      (table $table-1 10 (ref null $i32_i32_=>_none))
+ (table $table-1 10 (ref null $i32_i32_=>_none))
+ ;; CHECK:      (elem $elem-1 (table $table-1) (i32.const 0) (ref null $i32_i32_=>_none) (ref.func $foo))
+ (elem $elem-1 (table $table-1) (i32.const 0) (ref null $i32_i32_=>_none)
+  (ref.func $foo))
 
- ;; CHECK:      (elem declare func $fallthrough-no-params $fallthrough-non-nullable $foo $return-nothing)
```
tlively, line 22: Do you know why `$foo` was removed from the table?

tlively (approved):
LGTM! I don't think there's anything wrong with this transformation.

test/lit/passes/optimize-instructions-call_ref.wast
kripken: Yes, this is just the `elem declare` - it's not a real table. It just declares items that a reference is taken to them, but that are not in an actual table (this helps VMs optimize, to know that up front IIUC). As this added a use of an item, it no longer needs to be here.

(I could have worked around this to avoid a change on this line, but `elem declare` is not part of this test, it's just an artifact of the wasm text & binary formats.)

PR #4208, opened 2021-10-04
Optimize call_indirect of a select of two constants

kripken:
```wat
(call_indirect
 ..args..
  (select
    (i32.const x)
    (i32.const y)
    (condition)
  )
)

=>

(if
  (condition)
  (call $func-for-x
    ..args..
  )
  (call $func-for-y
    ..args..
  )
)
```

To do this we must reorder the condition with the args, and also use
the args more than once, so place them all in locals.

This works towards the goal of polymorphic devirtualization, that is,
turning an indirect call of more than one possible target into more
than one direct call.


src/passes/Directize.cpp
```diff
@@ -50,37 +51,77 @@ struct FunctionDirectizer : public WalkerPass<PostWalker<FunctionDirectizer>> {
...
+      replaceCurrent(makeDirectCall(operands, curr->target, flatTable, curr));
+      return;
+    }
+
+    // If the target is a select of two different constants, we can emit two
+    // direct calls.
+    // TODO: handle 3+
+    // TODO: handle the case where just one arm is a constant?
+    if (auto* select = curr->target->dynCast<Select>()) {
+      if (select->ifTrue->is<Const>() && select->ifFalse->is<Const>()) {
```
tlively, line 67: You could use `match.h` here if you'd like.

src/passes/Directize.cpp
```diff
@@ -50,37 +51,77 @@ struct FunctionDirectizer : public WalkerPass<PostWalker<FunctionDirectizer>> {
...
+          return;
+        }
+        auto conditionLocal = builder.addVar(func, Type::i32);
+        blockContents.push_back(
+          builder.makeLocalSet(conditionLocal, select->condition));
+
+        // Build the calls.
+        auto numOperands = curr->operands.size();
+        std::vector<Expression*> newOperands(numOperands);
+        auto getOperands = [&]() {
```
tlively, line 96: It seems strange that this lambda closes over but also returns `newOperands`. Can we move `newOperands` into the lambda and create it fresh each time?

src/passes/Directize.cpp
```diff
@@ -89,14 +130,48 @@ struct FunctionDirectizer : public WalkerPass<PostWalker<FunctionDirectizer>> {
...
+    // remove them, at least not by default).
+    if (index >= flatTable.names.size()) {
+      return replaceWithUnreachable(operands);
+    }
+    auto name = flatTable.names[index];
+    if (!name.is()) {
+      return replaceWithUnreachable(operands);
+    }
+    auto* func = getModule()->getFunction(name);
+    if (original->sig != func->getSig()) {
```
tlively, line 155: Might as well use a subtype check here to make this more future-proof while we're changing it. wdyt?

test/lit/passes/directize_all-features.wast
```diff
@@ -494,3 +498,314 @@
...
+ ;; CHECK-NEXT:   (local.get $x)
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT:  (local.set $4
+ ;; CHECK-NEXT:   (local.get $y)
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT:  (local.set $5
+ ;; CHECK-NEXT:   (local.get $z)
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT:  (if
+ ;; CHECK-NEXT:   (local.get $5)
```
tlively, line 538: It doesn't seem necessary to put the condition in a local. Does that simplify the implementation?

tlively (approved):
LGTM apart from comments.

src/passes/Directize.cpp
kripken: Fair enough, yeah, that would be clearer. And this is not that performance-sensitive I guess. (Also maybe the compiler optimizes it?)

src/passes/Directize.cpp
kripken: Hmm, perhaps... but it's just two things to check, and adding Match to here seems like a larger change.

src/passes/Directize.cpp
kripken: Good idea to do that, but I'd prefer to do it in a followup.

test/lit/passes/directize_all-features.wast
kripken: Ah, good point, yeah, if all the others are in locals then we can just use this immediately. Fixing.

src/passes/Directize.cpp
kripken: Actually, can we do this check here? `HeapType::isSubType` would check nominally, but this must be a structural check.

kripken: Or did you mean to do a subtyping check on each item of the params and results separately, or something like that?

tlively: Oh yeah we would need a structural signature subtype check even in nominal mode. I guess we don't have an API for that yet since we don't actually support function subtyping.

tlively (approved):

PR #4209, opened 2021-10-04
Refactor generic functionality out of ConstantFieldPropagation. NFC

kripken:
This just moves code outside and makes it more generic. One set of
functionality are "struct utils", which are tools to scan wasm for info
about the usage of struct fields, and to analyze that data. The other
tool is a general analysis of nominal subtypes.

The code will be useful in a few upcoming passes, so this will avoid a
significant amount of code duplication.


src/ir/struct-utils.h
```diff
@@ -0,0 +1,251 @@
...
+};
+
+// Helper class to propagate information about fields to sub- and/or super-
+// classes. While propagating it calls a method
+//
+//  to.combine(from)
+//
+// which combines the information from |from| into |to|, and should return true
+// if we changed something.
+template<typename T> class StructValuePropagator {
```
tlively, line 196: Can we get something about `TypeHierarchy` into the name to hint at how the propagation is happening?

tlively (approved):
LGTM apart from those comments.

src/ir/struct-utils.h
```diff
@@ -0,0 +1,251 @@
...
+    }
+  }
+};
+
+// A generic scanner that finds struct operations and calls hooks to update
+// information. Subclasses must define these methods:
+//
+// * Note an expression written into a field.
+//
+//   void noteExpression(Expression* expr, HeapType type, Index index, T& info);
```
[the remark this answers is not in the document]
kripken: Perhaps, but they are all writes, so all three names would begin in `noteWritten*` which seems excessive to me. That is, we note a written expression, or a written default value (during creation; for this we just need the type) or a written copy (for this we don't need the value, as we know it is writing a copy, hence one parameter less).

src/ir/struct-utils.h
kripken: Ah, nice idea, renaming to `TypeHierarchyPropagator`.

PR #4211, opened 2021-10-04
Update nominal binary format to match milestone 4

tlively:
Update the binary format used in --nominal mode to match the format of nominal
types in milestone 4. In particular, types without declared supertypes are now
emitted using the nominal type codes with either `func` or `data` as their
supertypes. This change is hopefully enough to get --nominal mode code running
on V8's milestone 4 implementation until the rest of the type system changes can
be implemented for use without --nominal.

### DIFF
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -219,16 +219,15 @@ void WasmBinaryWriter::writeTypes() {
     return;
   }
   BYN_TRACE("== writeTypes\n");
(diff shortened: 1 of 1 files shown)


kripken (approved):
Nice!

Please fuzz this well before landing.

tlively:
Fuzzing with --nominal is difficult since we have some tests in the recently modified set that do not validate under --nominal. That being said, I fuzzed this for a bit and didn't see any new problems. I expect that bugs here would have made themselves obvious.

PR #4212, opened 2021-10-04
Fix roundtripping specialized element segments of table zero

kripken:
Before this fix, the first table (index 0) is counted as its element segment
having "no index". That makes us not emit a type for it, which makes the
testcase added here (found by the fuzzer) fail, as the table has a specialized
type, so the element segment must have the same or more.

Or am I missing something and the wasm spec special-cases table zero?


tlively (commented):
Yes, table 0 is special and does not require a table index on active segments. See `elem` variant 0x00 in https://webassembly.github.io/spec/core/binary/modules.html#element-section. This is for backward compatibility with MVP modules that do not support multiple tables or table indices at all, so it is important that we continue eliding the index when we can.

src/wasm/wasm-binary.cpp
```diff
@@ -590,7 +590,7 @@ void WasmBinaryWriter::writeElementSegments() {
     bool hasTableIndex = false;
     if (!isPassive) {
       tableIdx = getTableIndex(segment->table);
-      hasTableIndex = tableIdx > 0;
+      hasTableIndex = true;
```
tlively, line 593:
```suggestion
      hasTableIndex = tableIdx > 0 || /* type != funcref */;
```

kripken:
Oh wow, thanks, so 0 is special-cased... ok, updated.

src/wasm/wasm-binary.cpp
```diff
@@ -590,7 +590,14 @@ void WasmBinaryWriter::writeElementSegments() {
     bool hasTableIndex = false;
     if (!isPassive) {
       tableIdx = getTableIndex(segment->table);
-      hasTableIndex = tableIdx > 0;
+      // Table index 0 is special for backwards compatibility: only tables with
```
tlively, line 593: Maybe something like this would be clearer: `// The table index can and should be elided for active segments of table 0 when table 0 has type funcref. This was the only type of segment supported by the MVP, which also did not support table indices in the segment encoding.`

tlively (approved):
Code LGTM.

src/wasm/wasm-binary.cpp
kripken: Thanks, added, that does read better.

PR #4213, opened 2021-10-04
[Wasm GC] GlobalTypeOptimization: Turn fields immutable when possible

kripken:
Add a new pass to perform global type optimization. So far this just
does one thing, to find fields with no `struct.set` and to turn them
immutable (where possible - sub and supertypes must agree).

To do that, this adds a `GlobalTypeRewriter` utility which rewrites
all the heap types in the module, allowing changes while doing so.
In this PR, the change is to flip the mutable field. Otherwise, the
utility handles all the boilerplate of creating temp heap types using
a `TypeBuilder`, and it handles replacing the types in every place
they are used in the module.

This is not enabled by default yet as I don't see enough of a benefit
on j2cl. This PR is basically the simplest thing to do in the space of
global type optimization, and the simplest way I can think of to
fully test the GlobalTypeRewriter (which can't be done as a unit
test, really, since we want to emit a full module and validate it etc.).

This PR builds the foundation for more complicated things like
removing unused fields, subtyping fields, and more.


src/passes/pass.cpp
```diff
@@ -154,6 +154,9 @@ void PassRegistry::registerPasses() {
     createGenerateI64DynCallsPass);
   registerPass(
     "generate-stack-ir", "generate Stack IR", createGenerateStackIRPass);
+  registerPass("gto",
+               "rewrite global GC types to be more optimal",
```
tlively, line 158: More concisely:
```suggestion
               "globally optimize GC types",
```

tlively (commented):
Nice!

PR #4218, opened 2021-10-07
Directize: Do not optimize if a table has a table.set

kripken:


aheejin:
Can we gather all set indices and optimize `call_indirect`s that don't contain one of those indices?

PR #4223, opened 2021-10-08
[finalize] Add imported tag symbols to 'declares'

aheejin:
This adds imported tags in `declares` section of
wasm-emscripten-finalize metadata, as with imported function. This is to
signal to Emscripten that these symbols are imported and make this
appear in `asmLibraryArg` in Emscripten-generated JS file.

This is to make Wasm EH work with dynamic linking.

Companion patches:
- https://reviews.llvm.org/D111388
- emscripten-core/emscripten#15266


aheejin:
> I thought we decided not to take this approach?

I can do that but I thought this might be simpler... By the way I think you said there are two ways of doing it in Emscripten:
1. Add the symbol to `DEFAULT_LIBRARY_FUNCS_TO_INCLUDE` + define the tag in library.js
2. Add the symbol to `send_items_map` + define the tag in preamble.js

I tried to do 1, but somehow I also need to add the symbol to `send_items_map`. So I have to add it to `DEFAULT_LIBRARY_FUNCS_INCLUDE`, also add it to `send_items_map`, and define it in library.js. If I do all these three, I can do it without this Binaryen patch. Please let me know if you prefer that over this.

sbc100:
yes, lets close this for now and stick with the emscripten side change.

If we want to change how it works and use this method instead I think should to it for `memory` and `__indirect_function_table` too.

PR #4236, opened 2021-10-12
Allow parsing of blocks with a type reference

pmatos:
Blocks are not allowed to have input parameters.

Currently only blocks of the form are supported:
```
(block $label (result ...)
  ...)
```

This patch enables this type of block declaration:
```
(type $ty (func (param) (result ...)))

(block $label (type $ty)
  ...)
```

### DIFF
--- a/src/wasm-s-parser.h
+++ b/src/wasm-s-parser.h
@@ -303,6 +303,7 @@ private:
 
   // Helper functions
   Type parseOptionalResultType(Element& s, Index& i);
+  Type parseOptionalTypeRef(Element& s, Index& i);
   Index parseMemoryLimits(Element& s, Index i);
   Index parseMemoryIndex(Element& s, Index i);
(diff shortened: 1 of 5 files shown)


pmatos:
This is still in draft mode as I am still working out an issue with the build. 

@tlively @kripken What's the best place for a test for this?

pmatos:
This is still breaking the validator. There's some strange case where apparently the parser is for:

```
(type $ty (func (param) (result i32)))
(func $withblock-typeref (param) (result i32)
    (block $label (type $ty)
      (i32.const 0)))) 
```

generating a `Rtt` for `(type $ty)` rather than a `Ref`.

tlively:
Probably a new test in `test/lit` would be good for this. For a very thorough test script, see the `RUN` lines in test/lit/table-first-special.wast.

PR #4243, opened 2021-10-13
Precompute: Track reference identity

kripken:
Precompute will run the interpreter on `struct.new` etc. repeatedly,
as it keeps doing so while it propagates constant values around (if one
of the operands to the `struct.new` becomes constant, that could have
a noticeable effect). But creating new GC data means we lose track of
their identity, and so `ref.eq` would not work, and we disabled basically
all struct operations. This implements identity tracking so we can start
to optimize there, which is a step towards using it for immutable field
propagation.

To track identity, always store the data representing each `struct.new`
in the source using the same GCData structure. That keeps identity
consistent no matter how many times we execute.


tlively:
Does this work for comparing equality of a gc param with itself?

PR #4251, opened 2021-10-14
[Wasm GC] Propagate immutable fields

kripken:
Very simple with the work so far, just add StructGet/ArrayGet code to check
if the field is immutable, and allow the get to go through in that case. 99% of
this PR is tests...


kripken:
Oh, oops, this PR was begun before the `stuct_subtype` etc. stuff landed... so to update the test I need to rebase as well it seems.

PR #4252, opened 2021-10-15
Add a --structural flag

tlively:
Just as the --nominal flag forces all types to be parsed as nominal, the
--structural flag forces all types to be parsed as equirecursive. This is the
current default behavior, but a future PR will change the default to parse types
as either structural or nominal according to their syntax or encoding. This new
flag will then be necessary to get the current behavior.

Also take this opportunity to deduplicate more flags in the help tests.


tlively:
It's for when users need to downgrade nominal types to structural to run on older Chrome versions. AFAIK, that is still a requirement for some of our benchmarking.

kripken:
Sorry I mean to say earlier that this lgtm aside from the comment. I see you have automerge so I didn't mark it approved?

tlively:
Oh shoot, I forgot to actually upload the fix for that comment. Sorry about that.

PR #4271, opened 2021-10-22
Fuzz nontrivial GC types

tlively:
The fuzzer quickly finds many bugs with this patch, but I haven't dug in to sort out which ones are bugs with this implementation and which are existing bugs found by the newly empowered fuzzer. The next steps will be to work through bugs until the fuzzer is stable with this patch, then we can start thinking about landing this.


tlively:
With the new type fuzzer stuff, this PR is largely obsolete. The only thing this PR has that we don't have landed is the use of nontrivial types in the main fuzzer.

PR #4280, opened 2021-10-26
Switch binaryen.js/wasm to ESM

dcodeIO:
As discussed a while ago, this PR aims to switch browser builds to ESM. New usage is ~:

```js
import Binaryen from "binaryen";
const binaryen = await Binaryen();
...
```

The `await` is not strictly necessary for the .js / wasm2js build, but when provided allows switching out the two builds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,13 @@ full changeset diff at the end of each section.
 Current Trunk
 -------------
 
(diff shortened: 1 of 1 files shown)


RefinedSoftwareLLC:
This might be related:
[https://github.com/AssemblyScript/assemblyscript/issues/1306](https://github.com/AssemblyScript/assemblyscript/issues/1306)

PR #4286, opened 2021-10-28
Effects: Differentiate mutable from immutable globals

kripken:
Similar to what we do with structs, if a global is immutable then we know it
cannot interact with calls.

This will unlock a bunch of j2wasm optimizations, as their `struct.new`s
often use `global.get`s of vtables, itables, rtts, etc.


src/ir/effects.h
```diff
@@ -128,8 +129,11 @@ class EffectAnalyzer {
   bool accessesLocal() const {
     return localsRead.size() + localsWritten.size() > 0;
   }
+  bool accessesMutableGlobal() const {
+    return globalsWritten.size() + mutableGlobalsRead.size() > 0;
```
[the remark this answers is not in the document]
tlively: Yeah that sounds fine.

test/passes/licm.wast
```diff
@@ -393,9 +394,34 @@
...
+      (drop (local.get $x))
+      (br_if $loop (local.get $x))
+    )
+  )
+  (func $global-call-immutable
+    (local $x i32)
+    (loop $loop
+      ;; As above, but now the global is immutable, so the call does not
+      ;; prevent us from optimizing.
+      ;; mutable global state.
```
tlively, line 420: Is this last comment line left over from a copy-paste?

tlively (approved):

test/lit/passes/gto-mutability.wast
```diff
@@ -14,14 +14,15 @@
   (type $two-params (func (param (ref $struct)) (param (ref $struct))))
 
   ;; Test that we update tag types properly.
+  (table 0 funcref)
```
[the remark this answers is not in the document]
tlively: Huh ok. Doesn't seem worth looking into, so let's just go with it.

test/passes/licm.wast
kripken: Oops, yes, a copy-paste leftover...

kripken:
```suggestion
```

kripken:
I figured out why the JS API test was failing here. `getSideEffects()` there had an optional module parameter. It was broken, however: passing in the module, it just pass that object to the compiled C code, so it was coerced to 0 or 1, and not a pointer to a module... To fix that, this now does `module.ptr` to actually get the pointer, and this is now actually tested as without a module we cannot compute the effects of a global.

This also makes the module param mandatory, as again, without a module we can't compute global effects.

tlively (approved):

PR #4292, opened 2021-10-29
ChildLocalizer: Do not use locals needlessly

kripken:
We only need to use locals if there are effects we can't remove, or if they
interact with other children. Improve the comment to explain what the
ChildLocalizer is working towards: a state where all the children of the
expression can be reordered or removed freely (`local.get`s have that
property, as do other things if they have no relevant effects).

Aside from avoiding wasteful locals, this is necessary for running
GlobalTypeOptimization on j2wasm: That code will do a `global.get`
of an rtt, and those cannot be placed in locals.


test/lit/passes/gto-removals.wast
```diff
@@ -475,6 +482,82 @@
...
+  ;; CHECK-NEXT:     (global.get $imm-i32)
+  ;; CHECK-NEXT:     (rtt.canon $struct)
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $new-side-effect-global-imm
+    ;; As above, the 2nd&3rd fields here will be removed. The first field does
+    ;; a global.get, which has effects, but those effects do not interact with
+    ;; anything else (since it is an immutable global), so we do not need a
```
aheejin, line 510: Should we consider reading immutable globals "effect" in the first place? Can we not count it in `EffectAnalyzer`?

src/ir/localize.h
```diff
@@ -44,13 +44,20 @@ struct Localizer {
...
-// Replaces all children with gets of locals, if they have any effects. After
-// this, the original input has only local.gets as inputs, or other things that
-// have no interacting effects, and so those children can be reordered.
+// Replaces all children with gets of locals, if they have any effects that
+// interact with any of the others, or if they have side effects which cannot be
+// removed.
+//
+// After this, the original input has only local.gets as inputs, or other things
+// that have no interacting effects, and so those children can be reordered
+// and/or removed as needed.
```
aheejin, line 53: So is the objective of this class is make all children of an expression freely reorderable? What use cases do you imagine for this other than field removal?

test/lit/passes/gto-removals.wast
kripken: Hmm, good point. We don't use immutable read effects at all right now. I can think of possible future uses, but maybe we should remove it for now. I can look into that separately.

src/ir/localize.h
kripken: For example when we turn an if into a select or vice versa then the condition moves around.

Also, removal and reordering are related, since if we want to remove a field from an instruction but it has effects then we need to move it to another position (like a drop either before or after), which can require reordering with the other children.

In general this pattern of "replace an instruction with a new one using a modified subset of the children" is common in OptimizeInstructions and RemoveUnusedBrs, and hopefully we can generalize and share that code.

aheejin (approved):
Sorry for the delayed reply!

Issue #4290, opened 2021-10-29
Cannot perform RTT casts on GC data created without an RTT


tlively:
`doCast` in wasm-interpreter.h assumes that if a cast uses an RTT, the casted reference must have been created with an RTT and that if a cast does not use an RTT, the casted reference must not have been created with an RTT. These assumptions are stricter than they should be. Looking at the [M4 prototype spec](https://docs.google.com/document/d/1DklC3qVuOdLHSXB5UXghM_syCh-4cMinQ50ICiXnK3Q/edit#), it looks like the precise semantics are not described, but it does say that rtt-consuming instructions should work with nominal types:

> Nominal types using RTT-consuming instructions; this could be useful as a gradual migration path, or as a potential final configuration if we decide to continue to use RTTs for the benefit of future extensibility.

Furthermore, the spec defines `rtt.canon` to have these semantics:

> If $t was defined with structtype, returns an RTT with no super-RTTs. 
> If $t was defined with struct_subtype, returns an RTT with super RTTs according to the type's static supertype 

The missing piece is that the allocation instructions without explicit RTTs implicitly use the canonical rtt for the relevant heap type, as defined above.

I plan to work on this, but I just wanted to document the issue first.

tlively:
This is fixed by #4294.

PR #4293, opened 2021-10-29
[Wasm GC] Fix casts of rtt.canon with nominal typing

kripken:
I am not sure this PR is right or not. We seem to have a few notions of casts atm, one for the original structural typing, one for nominal typing, and also for rtt-free instructions that might involve either or both of structural and nominal subtyping. The specific problem this PR tries to address is that j2wasm used to do
```wat
(ref.test
  (struct.new $subclass ;; declared as struct_subtype of $class
    (rtt.sub $subclass (rtt.canon $class))
  )
  (rtt.canon $class)
)
```
That worked (i.e. it returns `1`). They are now switching to
```wat
(ref.test
  (struct.new $subclass
    (rtt.canon $subclass)) ;; this line changed to a canon of the subclass
  )
  (rtt.canon $class)
)
```
This apparently works in V8, though I am not entirely sure which spec to read to understand why. This PR adds a possible hack to get things to work, that when we use nominal typing, then use the nominal info at the base of rtt chains..? :shrug:

### DIFF
--- a/src/wasm/literal.cpp
+++ b/src/wasm/literal.cpp
@@ -2562,7 +2562,17 @@ bool Literal::isSubRtt(const Literal& other) const {
   if (otherSupers.size() < supers.size()) {
     return other.type == supers[otherSupers.size()].type;
   } else {
-    return other.type == type;
+    if (getTypeSystem() == TypeSystem::Nominal) {
+      // If the type system is nominal, then allow not just type equality here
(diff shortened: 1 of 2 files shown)


tlively:
The problem here is related to #4290. I am working on a more comprehensive fix and hope to have it ready soon.

kripken:
Sgtm! Closing.

Feel free to take the tests from here if that would be helpful.

PR #4294, opened 2021-10-30
Fix RTTs for RTT-less instructions

tlively:
Allocation and cast instructions without explicit RTTs should use the canonical
RTTs for the given types. Furthermore, the RTTs for nominal types should reflect
the static type hierarchy. Previously, however, we implemented allocations and
casts without RTTs using an alternative system that only used static types
rather than RTT values. This alternative system would work fine in a world
without first-class RTTs, but it did not properly allow mixing instructions that
use RTTs and instructions that do not use RTTs as intended by the M4 GC spec.

This PR fixes the issue by using canonical RTTs where appropriate and cleans up
the relevant casting code using std::variant.

-----

The code in this PR should be complete, but it will need tests.

### DIFF
--- a/src/wasm-interpreter.h
+++ b/src/wasm-interpreter.h
@@ -26,6 +26,7 @@
 #include <cmath>
 #include <limits.h>
 #include <sstream>
+#include <variant>
 
 #include "ir/module-utils.h"
 #include "support/bits.h"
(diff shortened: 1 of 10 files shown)


src/wasm-interpreter.h
```diff
@@ -1404,154 +1405,141 @@ class ExpressionRunner : public OverriddenVisitor<SubType, Flow> {
...
-      auto* func = module
-                     ? module->getFunctionOrNull(cast.originalRef.getFunc())
-                     : nullptr;
+    // externref or an i31. The cast definitely fails in these cases.
+    if (!original.isData() && !original.isFunction()) {
+      return typename Cast::Failure{original};
+    }
+    Literal actualRtt;
+    if (original.isFunction()) {
+      // Function references always have the canonical rTTs of the functions
```
kripken, line 1470:
```suggestion
      // Function references always have the canonical RTTs of the functions
```

src/wasm-interpreter.h
```diff
@@ -1404,154 +1405,141 @@ class ExpressionRunner : public OverriddenVisitor<SubType, Flow> {
...
     NOTE_ENTER("RefTest");
     auto cast = doCast(curr);
-    if (cast.outcome == cast.Break) {
-      return cast.breaking;
+    if (auto* breaking = cast.getBreaking()) {
+      return *breaking;
+    } else if (cast.getSuccess()) {
+      return Literal(int32_t(1));
+    } else {
+      return Literal(int32_t(0));
```
kripken, line 1505:
```suggestion
    } else {
      return Literal(int32_t(cast.getSuccess() ? 1 : 0));
    }
```

kripken (commented):
Nice!

src/wasm-interpreter.h
tlively: Will use the similar `bool(getSuccess())`.

tlively:
Added tests building on top of #4301. Here's the diff specific to this PR: https://github.com/WebAssembly/binaryen/pull/4294/files/4a53df9c4b83bff7c8f028814dcb1af5ab90391a..21e9c5bb5d620e79bafbed10d590dd04e33bc098

kripken:
Oops, I see you wrote that in a comment...

kripken (approved):

PR #4298, opened 2021-11-01
[Wasm GC] LUBFinder helper. NFC

kripken:
This is a minor refactoring in DAE to have a helper class that does the
incremental LUB calculation. The class is also used in LocalSubtyping,
where it has the effect of making the work incremental which it was not
before (that would have no observable consequence, but it should make
us faster in the common case where we fail to find a new LUB).

This refactoring by itself does little, but it provides a central place to
further optimize this LUB calculation, which will be in a later PR.

### DIFF
--- a/src/ir/lubs.h
+++ b/src/ir/lubs.h
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2021 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


tlively (approved):
What kind of optimizations do you plan for this?

kripken:
The key optimization is to ignore/update nulls. The LUB of `(ref.null any)` and `(ref data)` can be `(ref null data)`, if we update that null to be `(ref null data)` - which we can do as all nulls are identical.

tlively:
Oh interesting. If you do that, you will need to update the types on the `ref.null` instructions to keep the module valid, right?

kripken:
Yes, exactly.

PR #4301, opened 2021-11-02
Generate lit checks for fuzz-exec output

tlively:
Use the new capability to port the bulk of
test/passes/Oz_fuzz-exec_all-features.wast to lit.


tlively:
This test seemed to previously be doing two things. First, it tested -Oz on some interesting patterns, and second, it tested that the interpreter was correct. I've basically ignored that first part in this PR, so I could use some guidance on how better to split the test.

tlively:
Instead of using the test currently in this PR, I plan to introduce the test from #4294 early. I'll get to that tomorrow.

kripken (approved):

PR #4303, opened 2021-11-03
Fuzz more basic GC types

tlively:
Generate both nullable and non-nullable references to basic HeapTypes and
introduce `i31` and `data` HeapTypes. Generate subtypes rather than exact types
for all concrete-typed children.

### DIFF
--- a/src/tools/fuzzing/fuzzing.cpp
+++ b/src/tools/fuzzing/fuzzing.cpp
@@ -16,6 +16,9 @@ constexpr int MAX_GLOBALS = 20;
 // The maximum number of tuple elements.
 constexpr int MAX_TUPLE_SIZE = 6;
 
+// The maximum rtt depth.
+constexpr int MAX_RTT_DEPTH = 3;
+
 // some things require luck, try them a few times
 constexpr int TRIES = 10;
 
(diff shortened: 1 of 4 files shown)


tlively:
Interestingly, my local `git clang-format` disagrees with CI here, but for now I've just applied the diff that CI suggested.

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -1916,76 +1931,143 @@ Literal TranslateToFuzzReader::makeLiteral(Type type) {
...
-      // For now, create a null, and if it must be non-null, cast it to such
-      // even though that traps at runtime.
-      auto nullable = Type(type.getHeapType(), Nullable);
-      Expression* ret = builder.makeRefNull(nullable);
-      if (!type.isNullable()) {
-        ret = builder.makeRefAs(RefAsNonNull, ret);
-      }
-      return ret;
+Expression* TranslateToFuzzReader::makeRefFuncConst(Type type) {
+  // Neither ref.func nor ref.as_non_null are allowed in globals, so we can only
```
kripken, line 1935: `ref.func` isn't allowed in globals? j2wasm vtables and itables use them. Is it just at the top level that `ref.func` isn't allowed? That seems odd if so.

tlively: Actually, I think this is just a temporary restriction on the fuzzer, since we create globals before we create functions. I'll clarify the comment and look at lifting this restriction in a follow-up.

kripken (approved):
sgtm about the comment.

PR #4305, opened 2021-11-04
[Wasm GC] Enable GlobalTypeOptimization

kripken:
Now that all known issues with that pass are fixed, enable it by
default. This adds it in a place that seems to make sense on j2wasm,
but in general multiple cycles of optimization will be needed.

This adds a test showing that we run this pass and that it helps
ConstantFieldPropagation by running before it.


tlively (approved):
Can we give the test file a more descriptive name rather than just using the passes it runs as the name? We have lots of ported legacy tests that just use the pass names, but for new tests I think we can do better.

kripken:
Do you have an idea for a better name? I think `O_gc` is actually good enough as it is about the sequence of GC passes we run by default in the optimization pipeline. The current name is not just the names of passes.

tlively:
Maybe something like `gto_and_cfp_in_O.wast`, since it's testing that GTO and CFP are included in -O? That's not a wonderful name either, though, so if we can't think of anything better, then the current name also seems ok.

kripken:
That name sgtm, done.

PR #4306, opened 2021-11-04
Add GlobalSubtyping pass

kripken:
This specializes the fields of structs based on the types written to them. That is,
if a field is of type A but in practice we always write some subtype B to it
then we can change the type of the field to that.

On j2wasm this manages to improve at least one field in 2% of types. Not a
large amount, but this does lead to further benefits in later opts (e.g. about a third
of the improvements are to turn a field non-nullable).


src/passes/GlobalSubtyping.cpp
```diff
@@ -0,0 +1,249 @@
...
+        }
+      }
+
+      if (auto super = type.getSuperType()) {
+        auto& superFields = super->getStruct().fields;
+        for (Index i = 0; i < superFields.size(); i++) {
+          auto newSuperType = finalInfos[*super][i].get();
+          auto& info = finalInfos[type][i];
+          auto newType = info.get();
+          if (!Type::isSubType(newType, newSuperType)) {
```
ajklein, line 182: Having stared at this for awhile I'm still having trouble understanding under what circumstances this happens (I think I'm not fully modeling the propagation steps above). Which test case(s) cover this?

tlively (commented):
Posting all my pending comments from when I started taking a look last week, but I haven't finished reading through this yet.

src/passes/GlobalSubtyping.cpp
kripken: The testcase beginning on line 422 covers this,

https://github.com/WebAssembly/binaryen/pull/4306/files#diff-f4453daffbce437bfcb3bf0a2117bfdc32fe747707ecfc546f3771c0d34b38e3R422-R466

This is a pretty subtle point and the code should explain it better. I'll push a comment with an example in a minute.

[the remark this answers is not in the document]
kripken: (see my response to @ajklein right below, and the comment I just pushed)

src/passes/GlobalSubtyping.cpp
```diff
@@ -0,0 +1,249 @@
...
+#include "ir/lubs.h"
+#include "ir/struct-utils.h"
+#include "ir/type-updating.h"
+#include "ir/utils.h"
+#include "pass.h"
+#include "support/unique_deferring_queue.h"
+#include "wasm-type.h"
+#include "wasm.h"
+
+using namespace std;
```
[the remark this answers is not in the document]
kripken: Ah, sorry about this namespace stuff, I think it's a leftover from some debugging that I missed...

src/passes/GlobalSubtyping.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+  // Force the lub to a particular type.
+  void set(Type type) { lub = type; }
+
+  void dump(std::ostream& o) {
+    std::cout << "FieldInfo(" << lub << ", " << nullDefault << ")";
+  }
+};
+
+struct FieldInfoScanner : public Scanner<FieldInfo, FieldInfoScanner> {
```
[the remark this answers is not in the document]
kripken: Sounds good. That's a refactoring that crosses a few other existing files, so I'll do it separately?

src/passes/GlobalSubtyping.cpp
```diff
@@ -0,0 +1,249 @@
...
+    while (!work.empty()) {
+      auto type = work.pop();
+
+      auto& fields = type.getStruct().fields;
+      for (Index i = 0; i < fields.size(); i++) {
+        auto oldType = fields[i].type;
+        auto& info = finalInfos[type][i];
+        auto newType = info.get();
+        if (newType == Type::unreachable) {
+          // We have no info on this field, so use the old type.
```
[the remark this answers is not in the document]
kripken: Note that while the field has no writes, but a super's field might - so we have no guarantee of GTO being able to remove it.

Overall, we have to pick *some* type for the field, which is all this does - we have no better idea of a type than the old  type, which we just leave there.

I'll add more comments to help here.

kripken:
Thanks for the feedback, I pushed some more comments and clarifications now. I also fixed the merge conflict.

src/passes/GlobalSubtyping.cpp
```diff
@@ -0,0 +1,266 @@
...
+  }
+};
+
+struct GlobalSubtyping : public Pass {
+  StructValuesMap<FieldInfo> finalInfos;
+
+  void run(PassRunner* runner, Module* module) override {
+    if (getTypeSystem() != TypeSystem::Nominal) {
+      Fatal() << "GlobalSubtyping requires nominal typing";
+    }
```
aheejin, line 114: Basic question: Why does this pass (and `GlobalTypeOptimization` too) require nominal typing?

aheejin (commented):
I started to read this but I'm not sure if I'll finish until you merge, and at the moment I only have a few simple nits and questions. I'll post them anyway but don't wait for me for the merge.

src/passes/GlobalSubtyping.cpp
kripken: Nominal typing lets us see exactly which types are subtypes of others. Technically we could compute that for structural subtyping too, but it takes a lot more work (we'd need to duplicate the logic in the canonicalization code, I assume), and also it would be less effective anyhow as when it merges types it limits optimization opportunities (since we can only optimize the merged type in one way, and not each of the original types differently as is best for it).

kripken:
Prelude PR: https://github.com/WebAssembly/binaryen/pull/4317

ajklein (commented):
Looks good to me, but I naturally defer to @tlively and @aheejin on proper approval here.

test/lit/passes/global-subtyping.wast
```diff
@@ -0,0 +1,538 @@
...
+    )
+    (struct.set $child 0
+      (local.get $child)
+      (local.get $child)
+    )
+  )
+)
+
+(module
+  ;; As in GP, write to the parent a child, and to the child a parent, but now
```
tlively, line 225: What is `GP`?

tlively (approved):
Very cool to see this working.

src/passes/GlobalSubtyping.cpp
```diff
@@ -0,0 +1,268 @@
...
+            // LUB of $A to take it into account, but as here $C is never even
+            // created, that does not happen. And then if we update $A's type
+            // to something more specific than $C's old type, we end up with the
+            // problem that this code path fixes: we just need to get $C's type
+            // to be identical to its super so that validation works.
+            info.set(newSuperType);
+          } else if (fields[i].mutable_ == Mutable) {
+            // Mutable fields must have identical types, so we cannot
+            // specialize.
+            info.set(newSuperType);
```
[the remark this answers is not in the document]
kripken: Interesting, yeah, maybe this should really be on fields. I'll add a TODO.

test/lit/passes/global-subtyping.wast
kripken: Clearly I spend too much time online... "GrandParent" (or really "Grandparent Post[er]"). That is, not the preceding testcase but one before it. I'll clarify.

test/lit/passes/global-subtyping.wast
```diff
@@ -0,0 +1,538 @@
...
+  ;; As above, but all writes are of $child-A, which allows more optimization
+  ;; up to that type.
+
+  ;; CHECK:      (type $struct (struct_subtype (field (mut (ref $child-A))) data))
+  (type $struct (struct_subtype (field (mut dataref)) data))
+
+  ;; CHECK:      (type $child-A (struct_subtype (field (mut (ref $child-A))) $struct))
+  (type $child-A (struct_subtype (field (mut dataref)) $struct))
+
+  (type $child-B (struct_subtype (field (mut dataref)) $struct))
```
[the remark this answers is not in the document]
kripken: Sounds good, added a mention of $child-B. That shows that it has been updated too, as expected.

test/lit/passes/global-subtyping.wast
tlively: I've only ever seen "OP" in the wild. Clearly I need to spend more time online!

PR #4312, opened 2021-11-05
Effects.h: Fix RefAs

kripken:
We marked that as only trapping if the input as nullable. But `ref.as_func`
will trap if it isn't a func, for example.

We could in theory try to check if a trap is possible, like checking if the
input is already non-nullable or already a function, etc., but we have
optimization passes to get rid of RefAs when they are not needed
anyhow, so there is no point to duplicate that here.


aheejin:
> but we have optimization passes to get rid of RefAs when they are not needed anyhow

Which pass is it? This contains a test for Vacuum but I couldn't find something from it.. (Maybe it's one of the functions called from it?)

But this `Effects` is used in a lot more places, and the cleanup happens in a specific pass, can this affect other optimizations in a negative way?

kripken:
> Which pass is it? 

In general, vacuum and other passes that use effects.h can call `hasRemovableSideEffects`, and that will return false if the only effect is a trap, which can remove a RefAs. In addition, we have specific stuff in OptimizeInstructions,

https://github.com/WebAssembly/binaryen/blob/b604ad935d1adc8498872e8afdb18df2f9b3b572/src/passes/OptimizeInstructions.cpp#L1304-L1306

> But this Effects is used in a lot more places, and the cleanup happens in a specific pass, can this affect other optimizations in a negative way?

Possibly yes. But in the worst case this would take another iteration of optimizations to get to optimal behavior. That is, every case in which we can report "no side effects" here is something we already have logic for actually removing the node, so after we run that logic we are in an optimal state. So I think it's better to avoid duplicating that logic here, but it is a tradeoff.

aheejin (approved):

kripken:
I added a comment for this - the question was a good one given the tradeoff is not obvious in the code.

tlively (approved):

PR #4315, opened 2021-11-08
[EH] Improve catch validation

aheejin:
This improves validation of `catch` bodies mostly by checking the
validity of `pop`s.

For every `catch` body:
- Checks if its tag exists
- If the tag's type is none:
  - Ensures there shouldn't be any `pop`s
- If the tag's type is not none:
  - Checks if there's a single `pop` within the catch body
  - Checks if the tag type matches the `pop`'s type
  - Checks if the `pop`'s location is valid

For every `catch_all` body:
- Ensures there shuldn't be any `pop`s

This uncovers several bugs related to `pop`s in existing tests, which
this PR also fixes.

### DIFF
--- a/src/ir/eh-utils.cpp
+++ b/src/ir/eh-utils.cpp
@@ -16,8 +16,6 @@
 
 #include "ir/eh-utils.h"
 #include "ir/branch-utils.h"
-#include "ir/find_all.h"
-#include "ir/properties.h"
 
 namespace wasm {
 
--- a/src/ir/eh-utils.cpp
+++ b/src/ir/eh-utils.cpp
@@ -66,7 +66,6 @@ bool isPopValid(Expression* catchBody) {
     }
     firstChild = *it.begin();
   }
-  assert(false); // Shouldn't be reached
 }
 
 } // namespace EHUtils
--- a/src/ir/eh-utils.cpp
+++ b/src/ir/eh-utils.cpp
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2021 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "ir/eh-utils.h"
+#include "ir/branch-utils.h"
+#include "ir/find_all.h"
+#include "ir/properties.h"
+
+namespace wasm {
+
+namespace EHUtils {
+
+bool isPopValid(Expression* catchBody) {
+  Expression* firstChild = nullptr;
+  auto* block = catchBody->dynCast<Block>();
+  if (!block) {
+    firstChild = catchBody;
+  } else {
+    // When there are multiple expressions within a catch body, an implicit
+    // block is created within it for convenience purposes, and if there are no
+    // branches that targets the block, it will be omitted when written back.
+    // But if there is a branch targetting this block, this block cannot be
+    // removed, and 'pop''s location will be like
+    // (catch $e
+    //   (block $l0
+    //     (pop i32) ;; within a block!
+    //     (br $l0)
+    //     ...
+    //   )
+    // )
+    // which is invalid.
+    if (BranchUtils::BranchSeeker::has(block, block->name)) {
+      return false;
+    }
+    // There should be a pop somewhere
+    if (block->list.empty()) {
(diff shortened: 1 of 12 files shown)


src/ir/eh-utils.cpp
```diff
@@ -0,0 +1,74 @@
...
+    }
+    // We use ValueChildIterator in order not to go into block/loop/try/if
+    // bodies, because a pop cannot be in those control flow expressions.
+    ValueChildIterator it(firstChild);
+    if (it.begin() == it.end()) {
+      return false;
+    }
+    firstChild = *it.begin();
+  }
+  assert(false); // Shouldn't be reached
```
kripken, line 69:
```suggestion
```

I don't think this is needed as the compiler will warn if it's reached anyhow since there is a return value.

src/wasm/wasm-validator.cpp
```diff
@@ -2143,6 +2144,66 @@ void FunctionValidator::visitTry(Try* curr) {
                 curr,
                 "try cannot have both catch and delegate at the same time");
 
+  // Given a catch body, find pops corresponding to the catch
+  auto findPops = [](Expression* expr) {
+    std::vector<Pop*> pops;
+    std::vector<Expression*> work;
```
kripken, line 2150: These could be SmallVectors perhaps? But probably the difference wouldn't be noticeable.

src/wasm/wasm-validator.cpp
```diff
@@ -2143,6 +2144,66 @@ void FunctionValidator::visitTry(Try* curr) {
...
+    }
+    return pops;
+  };
+
+  for (Index i = 0; i < curr->catchTags.size(); i++) {
+    Name tagName = curr->catchTags[i];
+    auto* tag = getModule()->getTagOrNull(tagName);
+    std::string msg =
+      std::string("catch's tag name is invalid: ") + tagName.c_str();
+    shouldBeTrue(tag != nullptr, curr, msg.c_str());
```
kripken, line 2175:
```suggestion
    if (!shouldBeTrue(tag != nullptr, curr)) {
      getStream() << "tag name is invalid: " << tagName << "\n";
    }
```

src/wasm/wasm-validator.cpp
```diff
@@ -2143,6 +2144,66 @@ void FunctionValidator::visitTry(Try* curr) {
...
+    if (tag->sig.params == Type::none) {
+      std::string msg = std::string("catch's tag (") + tagName.c_str() +
+                        ") doesn't have any params, but there are pops";
+      shouldBeTrue(pops.empty(), curr, msg.c_str());
+    } else {
+      std::string msg =
+        std::string("catch's tag (") + tagName.c_str() +
+        ") has params, so there should be a single pop within the catch body";
+      shouldBeTrue(pops.size() == 1, curr, msg.c_str());
+      if (pops.size() == 1) {
```
kripken, line 2188:
```suggestion
      if (shouldBeTrue(pops.size() == 1, curr, msg.c_str())) {
```

src/wasm/wasm-validator.cpp
aheejin: Done

src/wasm/wasm-validator.cpp
```diff
@@ -2143,6 +2144,65 @@ void FunctionValidator::visitTry(Try* curr) {
...
+    Name tagName = curr->catchTags[i];
+    auto* tag = getModule()->getTagOrNull(tagName);
+    if (!shouldBeTrue(tag != nullptr, curr, "")) {
+      getStream() << "tag name is invalid: " << tagName << "\n";
+    }
+
+    auto* catchBody = curr->catchBodies[i];
+    SmallVector<Pop*, 1> pops = findPops(catchBody);
+    if (tag->sig.params == Type::none) {
+      std::string msg = std::string("catch's tag (") + tagName.c_str() +
```
kripken, line 2180: I'd suggest using the same pattern here to avoid allocating a string even when there isn't an error. Looks like there are some more cases below too.

kripken (approved):
lgtm with that comment

tlively (approved):

src/wasm/wasm-validator.cpp
aheejin: Done

src/wasm/wasm-validator.cpp
```diff
@@ -2143,6 +2144,65 @@ void FunctionValidator::visitTry(Try* curr) {
...
+      shouldBeTrue(pops.empty(), curr, msg.c_str());
+    } else {
+      std::string msg =
+        std::string("catch's tag (") + tagName.c_str() +
+        ") has params, so there should be a single pop within the catch body";
+      if (shouldBeTrue(pops.size() == 1, curr, msg.c_str())) {
+        auto* pop = *pops.begin();
+        msg = std::string("catch's tag (") + tagName.c_str() +
+              ")'s pop doesn't have the same type as the tag's params";
+        shouldBeTrue(pop->type == tag->sig.params, curr, msg.c_str());
```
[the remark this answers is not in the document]
aheejin: Thanks for pointing this out! Added a test for this too.

kripken (approved):

PR #4317, opened 2021-11-09
[NFC] Add StructUtils namespace, and Scanner => StructScanner

kripken:
This avoids cluttering the main `wasm` namespace, and clarifies what the
scanner does.


tlively (approved):
Thanks!

aheejin (approved):

PR #4318, opened 2021-11-10
CoalesceLocals: Remove a redundant tee of the same local as a parent set

kripken:
Tiny followup to #4314

Existing tests happen to already cover this pretty well.

Also updates some function types in test output - I suspect that is from a
race in landing of PRs.

### DIFF
--- a/test/lit/passes/optimize-instructions-gc.wast
+++ b/test/lit/passes/optimize-instructions-gc.wast
@@ -2428,7 +2428,7 @@
   ;; CHECK-NEXT:   )
   ;; CHECK-NEXT:  )
   ;; CHECK-NEXT: )
-  ;; NOMNL:      (func $ref-cast-static-fallthrough-remaining (param $x eqref)
+  ;; NOMNL:      (func $ref-cast-static-fallthrough-remaining (type $eqref_=>_none) (param $x eqref)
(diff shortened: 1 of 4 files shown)


test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -2428,7 +2428,7 @@
   ;; CHECK-NEXT:   )
   ;; CHECK-NEXT:  )
   ;; CHECK-NEXT: )
-  ;; NOMNL:      (func $ref-cast-static-fallthrough-remaining (param $x eqref)
+  ;; NOMNL:      (func $ref-cast-static-fallthrough-remaining (type $eqref_=>_none) (param $x eqref)
```
tlively, line 2431: Are the changes in this file related to the PR?

tlively (approved):

test/lit/passes/optimize-instructions-gc.wast
kripken: No, I think that's what I mentioned at the end of the description, `Also updates some function types in test output - I suspect that is from a race in landing of PRs.` So this fixes breakage on main. (Could be a separate PR, but the benefit to that seems slim.)

PR #4324, opened 2021-11-12
Remove python from CMake build

### LINKED ISSUE #3036: Don't require Python in build (if not found, skip those targets)
This is entirely similar to https://github.com/WebAssembly/wabt/issues/1385

On my Windows machine, CMake fails with:

```
CMake Error at C:/Program Files/CMake/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
  Could NOT find PythonInterp: Found unsuitable version "2.7.15", but
  required is at least "3.5" (found C:/Python27/python.exe)
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:391 (_FPHSA_FAILURE_MESSAGE)
  C:/Program Files/CMake/share/cmake-3.16/Modules/FindPythonInterp.cmake:169 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  src/passes/CMakeLists.txt:2 (find_package)
```

I have both Python2 and Python3 installed. The latter apparently needs to be invoked with `py -3`.

But more importantly, one should be able to build a C++/CMake project without needing Python at all. If it is used for tests, then it simply won't be able to run those tests if the user has no Python installed.

phated:
Closes #3036

I'm trying to remove the python dependency from the Cmake build process (to aid in some OCaml repackaging for Grain).

This changes the values inside `WasmIntrinsics.cpp` from integers to hex, but I hope that doesn't matter. If it does, then we won't be able to use cmake to produce this file.


phated:
Somehow I broke everything 🙈 🙉 🙊

phated:
Excellent! It was just the location of the generated file. All green now 🎉

kripken (commented):
Thanks!

lgtm, but two concerns:

* Do you know which version of cmake this requires? @sbc100 we'd need to verify that matches the other emsdk requirements. Skimming the [docs](https://cmake.org/cmake/help/latest/command/configure_file.html) it seems to be in at least 3.10.
* Does this change in any way how often that file is rebuilt? I mean, does it still recognize that after a build it does not manually do this step each time, and only does it when that input wat changes.

phated:
Binaryen's cmake requirement seems to be 3.10.2, so I set that version in their docs while I reviewed the implementation.

I'm not sure how often the file is rebuilt. I wonder if there's a way for me to test that 🤔

phated:
I just did a little bit of poking around and found some info:

From [this stackoverflow](https://stackoverflow.com/a/31187095):

> `configure_file()` is executed immediately at configuration step

And then from the [cmake mailing list](https://cmake.org/pipermail/cmake/2012-May/050221.html):

> I am using `configure_file` to fill in the required values.
> Unfortunately, every time the file is modified, it triggers CMake to do
> a complete configuration and generation step.

So it seems that it is executed at the configuration step, but will cause a re-configuration when the `wat` file changes before a build will be executed. I'm definitely not a cmake expert, so it'd help a ton if anyone else can provide more details.

sbc100:
Can you explain why have a python dependency is do difficult?   Shouldn't the focus be on fixing the downstream build system to be able to depend on python?

As an alternative to this approach I guess would could instead checkin the result of embed-wat?   I would gen it almost never changes.  Then we could skip the auto-rebuild step if python is not found?

phated:
> Can you explain why have a python dependency is do difficult?

In grain, we use a build system that sandboxes absolutely everything, so we would need to build python from source just to build this single file. Why would you want an entire python dependency to build 1 file? Up until the Mac M1 came out, we just built archives in CI, but that's not possible right now.

We're not the only people that have this problem (see the issue I linked in the original.

I'm fine with checking in the file instead, but this was what I proposed and @kripken said it was worth a try. I also don't know how to change your setup to do that.

sbc100:
> > Can you explain why have a python dependency is do difficult?
> 
> In grain, we use a build system that sandboxes absolutely everything, so we would need to build python from source just to build this single file. Why would you want an entire python dependency to build 1 file? Up until the Mac M1 came out, we just built archives in CI, but that's not possible right now.

I'm not aware of grain I'm afraid.   Are you trying to bootstrap and entire software ecosystem?  And if so, won't you need to build python anyway?   I assume you are already building sandboxed version of all the other dependencies like cmake and ninja?   Are you saying that those need to be built from scratch every time you build binaryen and of those builds doesn't work somehow because of Mac M1?

phated:
Python is needed nothing in our build, except this single file in binaryen. Cmake is built from source. Ninja isn't required, as we use the "unix makefiles" option. 

> Are you saying that those need to be built from scratch every time you build binaryen and that caching of those builds doesn't work somehow because of Mac M1?

I don't understand the question.

phated:
I understand that you are trying to understand why we want this, but if you check the original thread, it's already identified as a problem. I just want to remove the dependency to solve it.

sbc100:
> I understand that you are trying to understand why we want this, but if you check the original thread, it's already identified as a problem. I just want to remove the dependency to solve it.

In this case I'm happy to go forward with this change because we use so little python.   As I say, I'm mostly just curious because in general I think using python in a build system is not an unreasonable thing to do.

phated:
I guess that the last thing to decide is if this is an acceptable solution or if someone wants to implement the "check in the file" solution.

Note that this is a blocker on https://github.com/grain-lang/libbinaryen/issues/15

sbc100 (approved):
This solution seems reasonable to me.

kripken:
@phated 

> I'm not sure how often the file is rebuilt. I wonder if there's a way for me to test that thinking

I would just do a simple check here: build everything, then modify some C++ file, then do a rebuild, and verify (maybe with verbose build output) that we only rebuild that C++ file and perform linking (and do not rebuild anything related to the changes in this PR). If that looks good it should be fine.

phated:
@kripken sorry for the delay here. I was finally able to test your suggestion and here are the results:

Building from scratch:
```bash
$ cmake --build out --config Release --target install
[222/223] Install the project...
````

And then I added a comment to `src/ir/ExpressionAnalyzer.cpp` and re-ran the build:

```bash
$ cmake --build out --config Release --target install
[13/14] Install the project...
```

I didn't run it in verbose mode, but I believe the drastically reduced file count means only the relevant files were rebuilt.

Is that enough to get this merged? We are hoping to get this into a new release of Binaryen soon so we can finish M1 support  for Grain!

kripken (approved):
lgtm @phated , thanks for checking that, and sorry for the holiday delay!

phated:
🎉 Excellent! Thanks! Hope you had a good holiday.

PR #4326, opened 2021-11-12
[Wasm GC] Signature Subtyping pass

kripken:
This is fairly short and simple after the recent refactorings. This basically
just finds all uses of each signature/function type, and then sees if it
receives more specific types as params. It then rewrites the types if so.

This just handles arguments so far, and not return types.

This differs from DeadArgumentElimination's refineArguments() in that
that pass modifies each function by itself, changing the type of the
function as needed. That is only valid if the type is not observable, that
is, if the function is called indirectly then DAE ignores it. This pass will
work on the types themselves, so it considers all functions sharing a
type as a whole, and when it upgrades that type it ends up affecting them
all.

This finds optimization opportunities on 4% of the total signature
types in j2wasm. Those lead to some benefits in later opts, but the
effect is not huge.


kripken:
Btw, this is the last of the 4 "subtyping" passes I had planned (DAE function subtyping, local subtyping, struct subtyping, and now signature subtyping). The only idea I have left after that is to unify them all in one big pass, but that would be a large amount of work and the benefit is not yet clear.

src/passes/SignatureSubtyping.cpp
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    // Compute optimal LUBs.
+    std::unordered_set<HeapType> seen;
+    for (auto& func : module->functions) {
+      auto type = func->type;
+      if (seen.count(type)) {
+        continue;
+      }
+      seen.insert(type);
```
tlively, line 101:
```suggestion
      if (seen.insert(type).second) {
        continue;
      }
```

src/passes/SignatureSubtyping.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+      auto sig = type.getSignature();
+
+      auto params = sig.params;
+      auto numParams = params.size();
```
tlively, line 106: It looks like `params` isn't used after this, so it probably doesn't need to be named.
```suggestion
      auto numParams = sig.params.size();
```

src/passes/SignatureSubtyping.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+    // Rewrite the types.
+    class TypeRewriter : public GlobalTypeRewriter {
+      SignatureSubtyping& parent;
+
+    public:
+      TypeRewriter(Module& wasm, SignatureSubtyping& parent)
+        : GlobalTypeRewriter(wasm), parent(parent) {}
+
+      virtual void modifySignature(HeapType oldSignatureType, Signature& sig) {
```
tlively, line 182:
```suggestion
      void modifySignature(HeapType oldSignatureType, Signature& sig) override {
```

tlively (commented):
It doesn't look like this handles updating table types, which will require considering and updating indirect calls as well, since we currently depend on indirect calls having the same signature as the table they point to. The existence of a `funcref` table should disable this optimization (until we start storing heap types on call_indirect) because that makes it impossible or at least untenably fragile to determine the heap type that will be emitted for its indirect calls.

Maybe this PR should bail out if there are any tables at all, and that can be relaxed in the future?

 I think this should also consider indirect calls to tables with particular signature types, since those indirect call signatures will have to work with the new type as well.

kripken:
Good points about tables... yes, I was assuming the simple case without any tables. Thinking on it, maybe this should just do nothing if tables exist, I agree.

kripken:
Ok, updated for all the feedback here so far. This now bails out if there are tables, as suggested - good idea. Also updated for latest main which now has null updating.

test/lit/passes/signature-refining.wast
```diff
@@ -0,0 +1,493 @@
...
+  ;; CHECK-NEXT:    (struct.new_default $struct)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:   (ref.func $func)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $caller
+    (call $func
+      (block (result (ref null $struct))
+        (ref.null $struct)
+      )
```
tlively, line 107: Can we use a local instead of a block?

tlively (approved):

PR #4328, opened 2021-11-13
Add a fuzzer specifically for types

tlively:
Add a new fuzzer binary that repeatedly generates random types to find bugs in
the type system implementation. Each iteration creates some number of root types
followed by some number of subtypes thereof. Each built type can contain
arbitrary references to other built types, regardless of their order of
construction.

Right now the fuzzer only finds fatal errors in type building (and in its own
implementation), but it is meant to be extended to check other properties in the
future, such as that LUB calculations work as expected.

The logic for creating types is also intended to be integrated into the main
fuzzer in a follow-on PR so that the main fuzzer can fuzz with arbitrarily more
interesting GC types.


src/tools/CMakeLists.txt
```diff
@@ -25,5 +26,6 @@ binaryen_add_executable(wasm-as wasm-as.cpp)
 binaryen_add_executable(wasm-dis wasm-dis.cpp)
 binaryen_add_executable(wasm-ctor-eval wasm-ctor-eval.cpp)
 binaryen_add_executable(wasm-reduce wasm-reduce.cpp)
+binaryen_add_executable(binaryen-fuzz-types "${fuzzing_SOURCES};fuzzing/binaryen-fuzz-types.cpp")
```
kripken, line 29: I'm unclear on why we want a new executable here? (And also why we want one with a new prefix, `binaryen-`, and that it is the first whose main file resides outside of `tools/`). In general I think we have a high bar for adding new executables, as we don't want to keep accumulating them over time.

What is the purpose of the new executable?

If the purpose is mainly to simplify testing, could we add a cpp unit test that does the same as the executable?

tlively: It seemed better to create a new executable than to clutter wasm-opt with new functionality unrelated to anything else it does. The purpose of the new executable is to fuzz the type system specifically, and doing that separately from the current fuzzing infrastructure lets us fuzz thousands of iterations per second. I imagine that in the future we could add additional fuzzers of specific subsystems to this tool for the same benefit. I chose the `binaryen-` prefix to try to hint that this is an internal development tool that is probably not generally useful, but that could probably be clearer. I'm open to suggestions on the name. I can also move the executable main file out to the `tools` directory.

kripken: I see, thanks.

What does LLVM do for such "mostly-internal" tools?

tlively: LLVM doesn't have very clear conventions here. Many LLVM tools start with the `llvm-` prefix, but both "internal" tools such as `llvm-mc` and drop-in replacements for well-known tools such as `llvm-cxxfilt` follow that convention. Other LLVM tools have entirely different naming schemes, such as `obj2yaml`, which does not hint at an LLVM connection at all.

kripken: I see. Yeah, `obj2yamp` and `opt` are weird to not have `llvm-` prefixes... anyhow, if they have "internal" tools with their prefix, then I see no reason for us not to either. I think I'd prefer `wasm-fuzz*` over starting to consider what a special prefix might look like and other bikeshedding about the location of the files etc.

tlively: Sure, I'll rename to `wasm-fuzz-types` (and we can make it more general if we want in the future) and move the main file up to `tools`.

src/wasm-type.h
```diff
@@ -576,6 +576,13 @@ struct TypeBuilder {
   void setHeapType(size_t i, Struct&& struct_);
   void setHeapType(size_t i, Array array);
 
+  // This is an ugly hack around the fact that temp heap types initialized with
+  // BasicHeapTypes are not themselves considered basic, so `HeapType::isBasic`
+  // and `HeapType::getBasic` do not work as expected with them. Call these
+  // methods instead.
+  bool isBasic(size_t i);
+  HeapType::BasicHeapType getBasic(size_t i);
```
kripken, line 584: Is fixing this issue not practical? I've run into it too and it is annoying.

src/tools/fuzzing/heap-types.cpp
```diff
@@ -0,0 +1,492 @@
...
+                     HeapType::ext,
+                     HeapType::any,
+                     HeapType::eq,
+                     HeapType::i31,
+                     HeapType::data);
+  }
+
+  Type::BasicType generateBasicType() {
+    // Exclude basic reference types because we will construct them via the
+    // TypeBuilder.
```
kripken, line 64: I don't understand this comment. We're constructing everything with the TypeBuilder, aren't we? And, this lack of symmetry between generateBasicHeapType/generateBasicType is confusing.

src/tools/fuzzing/heap-types.cpp
```diff
@@ -0,0 +1,492 @@
...
+      case 2:
+        return generateRttType();
+    }
+    WASM_UNREACHABLE("unexpected");
+  }
+
+  Type generateTupleType() {
+    std::vector<Type> types(2 + rand.upTo(MAX_TUPLE_SIZE - 1));
+    for (auto& type : types) {
+      // Make sure tuples are defaultable. See comment in getTupleType.
```
kripken, line 107: This is no `getTupleType` function in this file..?

src/tools/fuzzing/heap-types.cpp
```diff
@@ -0,0 +1,492 @@
...
+      type = generateSingleType(/*defaultable=*/true);
+    }
+    return builder.getTempTupleType(Tuple(types));
+  }
+
+  Type generateControlFlowType() {
```
kripken, line 113: Perhaps `generateReturnType`?

kripken (commented):
Nice! Mostly minor comments.

There are some code movements here (like the picker stuff). I assume those are all moved as-is / NFC?

test/lit/fuzz-types/nominal.test
```diff
@@ -0,0 +1,24 @@
...
+;; CHECK-NEXT: 10: (struct (field (ref eq) i8 (rtt func)))
+;; CHECK-NEXT: 11: (array (mut i16))
+;; CHECK-NEXT: 12: (struct (field (ref (array (mut i16))) i8 f32 (mut i8) (mut (rtt (struct (field (ref any) i8)))) v128))
+;; CHECK-NEXT: 13: (struct (field i31ref i8 (mut (ref func))))
+;; CHECK-NEXT: 14: (func (result anyref))
+;; CHECK-NEXT: 15: (struct (field (ref any) i8))
+;; CHECK-NEXT: 16: (func (result anyref))
+;; CHECK-NEXT: 17: (array (mut i16))
+;; CHECK-NEXT: 18: (array (mut i16))
+;; CHECK-NEXT: 19: (struct (field (ref (array (mut i16))) i8 f32 (mut (rtt i31))))
```
kripken, line 24: Perhaps this could also print the index of the super, if there is one?

tlively: Yes, these tests make it clear that there is room for improvement when printing groups of interrelated HeapTypes. Do you mind if we leave improvements there to a follow-on PR?

src/tools/fuzzing/heap-types.cpp
tlively: This is copied from the method of the same name in `fuzzing.cpp`, but in this context you're right that `generateReturnType` probably makes more sense. Will update this and add a comment about the similarity to the other method.

src/tools/fuzzing/heap-types.cpp
tlively: Oops, this must be stale or copied from elsewhere incorrectly. Thanks for the catch.

src/wasm-type.h
tlively: `isBasic` is currently a single integer comparison that does not need to know any of the implementation details in `wasm-type.cpp`, but fixing this would require making it more complicated. I should probably check whether the performance difference is noticeable, and if it's not, just fix this the "right" way.

tlively:
> There are some code movements here (like the picker stuff). I assume those are all moved as-is / NFC?

Correct; there should be no changes there.

src/tools/fuzzing/heap-types.cpp
tlively: I'll undo this. The goal was to have just one way to generate e.g. (ref null any), but actually we get better fuzz coverage if we allow it to be generated directly as a basic type as well as indirectly as a nullable reference to a table slot that happens to contain `any`.

tlively:
The simple comments should be taken care of. I will run performance experiments with `isBasic` to figure out what to do there and I will work on a follow-up PR to improve the type printing by giving the types names and properly printing their supertypes.

kripken (approved):
Followups sgtm since this is a new feature anyhow.

src/tools/fuzzing/heap-types.cpp
```diff
@@ -0,0 +1,499 @@
...
+  struct Ref {
+    HeapType type;
+    Nullability nullability;
+  };
+
+  Ref generateSubRef(Ref super) {
+    auto nullability = super.nullability == NonNullable
+                         ? NonNullable
+                         : rand.oneIn(2) ? Nullable : NonNullable;
+    return {pickSubHeapType(super.type), nullability};
```
aheejin, line 305: When do we `generate` things and when do we `pick` things? There are `generate***` methods and `pick***` methods, and this place looks like an entry to callsites of `pick***` methods. Does picking mean we have already generated all types?

aheejin (commented):
Sorry for the late comments! I haven't reviewed much type system-related code for a while, so most of them are just basic questions on how we deal with that.

src/tools/wasm-fuzz-types.cpp
```diff
@@ -0,0 +1,122 @@
...
+    "Use the nominal type system (default)",
+    Options::Arguments::Zero,
+    [&](Options*, const std::string& arg) { system = TypeSystem::Nominal; });
+  options.add("--structural",
+              "",
+              "Use the equirecursive type system",
+              Options::Arguments::Zero,
+              [&](Options*, const std::string& arg) {
+                system = TypeSystem::Equirecursive;
+              });
```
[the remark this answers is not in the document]
tlively: If the user gives both, then the last one wins. This is consistent with how we do this in wasm-opt and other tools, so I think that's ok.

src/tools/fuzzing/heap-types.cpp
```diff
@@ -0,0 +1,499 @@
...
+    // then after that start creating subtypes of existing types. Determine the
+    // top-level kind of each type in advance so that we can appropriately use
+    // types we haven't constructed yet.
+    // TODO: Determine individually whether each HeapType is nominal once
+    // mixing type systems is expected to work.
+    typeKinds.reserve(builder.size());
+    std::vector<std::optional<Index>> supertypeIndices(builder.size());
+    Index numRoots = 1 + rand.upTo(builder.size());
+    for (Index i = 0; i < builder.size(); ++i) {
+      typeIndices.insert({builder[i], i});
```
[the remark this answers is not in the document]
tlively: HeapTypes are represented as integers that are either one of the BasicHeapType enum variants or a pointer to a `HeapTypeInfo` on the heap. When you cast a `TypeBuilder::Entry` to `HeapType`, you are essentially just taking a pointer to the `HeapTypeInfo` for some slot in the `TypeBuilder`. When that backing `HeapTypeInfo` is mutated via the `TypeBuilder` API, the `HeapType` that points to it does not need to be updated.

src/tools/fuzzing/heap-types.cpp
```diff
@@ -0,0 +1,499 @@
...
+  Array generateSubArray(Array super) {
+    if (rand.oneIn(2)) {
+      return super;
+    }
+    return {generateSubField(super.element)};
+  }
+
+  HeapTypeKind generateHeapTypeKind() {
+    // Building basic heap types is only allowed in equirecursive mode.
+    // TODO: Relax this.
```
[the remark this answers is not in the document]
tlively: This is an artificial limitation of the current implementation. When I wrote the nominal implementation, I cut some corners and assumed / asserted that none of the TypeBuilder slots would be basic types. This is fine for normal use because the only time wasm-opt has to build a basic type is in equirecursive LUB calculations. Nominal typing should never need to build a basic type, but it turns out that having this difference between the type systems makes it harder to generalize code to work with both of them.

src/tools/fuzzing/heap-types.cpp
```diff
@@ -0,0 +1,499 @@
...
+    if (rand.oneIn(2)) {
+      return super;
+    }
+    std::vector<Field> fields;
+    // Depth subtyping
+    for (auto field : super.fields) {
+      fields.push_back(generateSubField(field));
+    }
+    // Width subtyping
+    Index extra = rand.upTo(MAX_STRUCT_SIZE + 1 - fields.size());
```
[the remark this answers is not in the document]
tlively: `upTo` is noninclusive (except for `upTo(0)`, which always returns 0).

src/tools/fuzzing/heap-types.cpp
```diff
@@ -0,0 +1,499 @@
...
+  }
+
+  Signature generateSubSignature(Signature super) {
+    // TODO: Update this once we support nontrivial function subtyping.
+    return super;
+  }
+
+  Field generateSubField(Field super) {
+    if (super.mutable_ == Mutable) {
+      // Only immutable fields support subtyping.
```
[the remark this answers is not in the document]
tlively: This has been the case all along because mutable subtyping would be unsound. For example, if `(struct (mut funcref))` were a subtype of `(struct (mut anyref))`, then I could write a non-func value via the `(mut anyref)`, then read it as a funcref from the `(mut funcref)`.

src/tools/fuzzing/heap-types.cpp
```diff
@@ -0,0 +1,499 @@
...
+    } else {
+      return builder[rand.upTo(builder.size())];
+    }
+  }
+
+  // TODO: Create an enum for `Defaultable, NonDefaultable` in wasm-type.h.
+  Type generateRefType(bool defaultable = false) {
+    auto heapType = generateHeapType();
+    auto nullability = (defaultable || rand.oneIn(2)) ? Nullable : NonNullable;
+    return builder.getTempRefType(heapType, nullability);
```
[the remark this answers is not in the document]
tlively: Sure!

Unlike normal `HeapType`s, the `HeapType`s owned by the `TypeBuilder` are mutable (or at least the `HeapTypeInfo`s that back them are). These `HeapType`s are also non-canonical, meaning they haven't been deduplicated (if necessary) with the normal `HeapTypes`. Similarly, the `Type`s owned by the `TypeBuilder` have not been deduplicated with the normal `Type`s, and they can't be until the `HeapType`s are all initialized and won't change any more. Since the temporary `Type`s and `HeapType`s owned by the TypeBuilder are not canonicalized, they cannot be used like normal `Type`s and `HeapType`s. In particular, type equality is implemented as integer equality for normal `Type`s and `HeapType`s, but that doesn't work before they've been deduplicated. We use "temp" in all the of the methods for creating these temporary types as a reminder that they should not be mixed with normal types.

src/tools/fuzzing/heap-types.cpp
```diff
@@ -0,0 +1,499 @@
...
+  HeapType::BasicHeapType generateBasicHeapType() {
+    return rand.pick(HeapType::func,
+                     HeapType::ext,
+                     HeapType::any,
+                     HeapType::eq,
+                     HeapType::i31,
+                     HeapType::data);
+  }
+
+  Type::BasicType generateBasicType() {
```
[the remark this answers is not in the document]
tlively: I don't have plans to share them directly, but I do plan for `fuzzing.cpp` to call `HeapTypeFuzzer::generateHeapTypes`. There is a lot of similarity between these smaller type generation functions and the type generations functions in `fuzzing.cpp`, so maybe they should be more directly shared, though.

src/tools/fuzzing/heap-types.cpp
```diff
@@ -0,0 +1,499 @@
...
+                   ? super.depth
+                   : rand.oneIn(2) ? Rtt::NoDepth : rand.upTo(MAX_RTT_DEPTH);
+    return {depth, super.heapType};
+  }
+
+  Type generateSubtype(Type type) {
+    if (type.isBasic()) {
+      // We do not construct types with basic reference types (we go through the
+      // TypeBuilder for those instead), so this must be a non-reference basic
+      // type, which means it has no other subtypes.
```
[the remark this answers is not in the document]
tlively: Oh good catch! This is a stale comment (and stale code as well). We used to not generate reference types in `getSingleType`, but now we do, so this code should be updated to generate strict subtypes of those basic reference types as well.

src/tools/fuzzing/heap-types.cpp
```diff
@@ -0,0 +1,499 @@
...
+  Ref generateSubRef(Ref super) {
+    auto nullability = super.nullability == NonNullable
+                         ? NonNullable
+                         : rand.oneIn(2) ? Nullable : NonNullable;
+    return {pickSubHeapType(super.type), nullability};
+  }
+
+  Rtt generateSubRtt(Rtt super) {
+    auto depth = super.hasDepth()
+                   ? super.depth
```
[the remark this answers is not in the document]
tlively: I don't remember there ever being a rule for subtyping between rtts of different depths. This makes sense because the rtt value in the engine must have some particular depth, so it would be unsound to type that value with two different depths.

src/tools/fuzzing/heap-types.cpp
```diff
@@ -0,0 +1,499 @@
...
+        case HeapType::i31:
+          return HeapType::i31;
+        case HeapType::data:
+          return pickSubData();
+      }
+      WASM_UNREACHABLE("unexpected kind");
+    }
+  }
+
+  // TODO: Make this part of the wasm-type.h API
```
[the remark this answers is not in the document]
tlively: Yes, and that should probably be moved to the public API in wasm-type.h, along with e.g. RTT.

src/tools/fuzzing/heap-types.cpp
```diff
@@ -0,0 +1,499 @@
...
+        return pickSubEq();
+    }
+    WASM_UNREACHABLE("unexpected index");
+  }
+
+  HeapType pickSubHeapType(HeapType type) {
+    auto it = typeIndices.find(type);
+    if (it != typeIndices.end()) {
+      // This is a constructed type, so we know where its subtypes are.
+      return builder[rand.pick(subtypeIndices[typeIndices[type]])];
```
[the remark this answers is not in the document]
tlively: Everything is at least a subtype of itself, so `subtypeIndices[x]` will never be empty.

src/tools/fuzzing/heap-types.cpp
```diff
@@ -0,0 +1,499 @@
...
+      }
+    }
+    if (candidates.size()) {
+      return rand.pick(candidates);
+    } else {
+      return {};
+    }
+  }
+
+  HeapType pickSubFunc() {
```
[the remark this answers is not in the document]
tlively: `pickSubFunc` is chosen to be consistent with `pickSubData`, `pickSubEq`, etc. They are all responsible for picking subtypes of a particular `BasicHeapType`.

src/tools/fuzzing/heap-types.cpp
```diff
@@ -0,0 +1,499 @@
...
+  // For each type we will build, the indices of its subtypes we will build.
+  std::vector<std::vector<Index>> subtypeIndices;
+
+  // Abstract over all the types that may be assigned to a builder slot.
+  using Assignable =
+    std::variant<HeapType::BasicHeapType, Signature, Struct, Array>;
+
+  // Top-level kinds, chosen before the types are actually constructed. This
+  // allows us to choose HeapTypes that we know will be subtypes of data or func
+  // before we actually generate the types.
```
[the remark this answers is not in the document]
tlively: Knowing the kinds beforehand lets us pick subtypes of `HeapType::func` and `HeapType::data` that we haven't initialized yet. Otherwise we wouldn't know which of the uninitialized types would end up being valid subtypes. We don't need to differentiate between arrays and structs at this stage because `data` covers both of them.

src/tools/fuzzing/heap-types.cpp
```diff
@@ -0,0 +1,499 @@
...
+  }
+
+  Type generateReturnType() {
+    // This is similar to TranslateToFuzzreader::getControlFlowType.
+    if (rand.oneIn(10)) {
+      return Type::none;
+    } else if (features.hasMultivalue() && rand.oneIn(5)) {
+      return generateTupleType();
+    } else {
+      return generateSingleType(/*defaultable=*/true);
```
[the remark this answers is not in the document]
tlively: Good question. Looking at `getControlFlowType` in fuzzing.cpp, which this is similar to, perhaps they don't need to be defaultable.

src/tools/fuzzing/heap-types.cpp
tlively: `pick***` means we are choosing from among a finite set of options (such as one of the slots of the type builder), whereas `generate` means we are creating something fresh. I agree it's not a very clear distinction, though.

tlively:
Thanks for your comments, @aheejin! I'll create a follow-up PR cleaning some of these things up.

src/tools/fuzzing/heap-types.cpp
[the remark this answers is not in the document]
aheejin: I don't get it... The subtyping is `funcref`<:`anyref`, then how can we write a non-func value into `mut anyref` and read it as a `funcref` from `mut funcref`?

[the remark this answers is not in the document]
tlively: Here's an example:

```
(module
  (type $func (func))
  (type $super (array (mut anyref)))
  (type $sub (array (mut (ref null $func)))) ;; subtype of $super
  (import "some" "extern" (global $extern externref))
  
  ;; Write an `externref` at index 0 of a `$super` value.
  (func $write-extern (param $x (ref $super))
    (array.set
      (local.get $x)
      (i32.const 0)
      (global.get $extern)
    )
  )

  ;; Call $write-extern, then read a (ref null func) out of the $sub value and call it.
  (func $unsound (param $x (ref $sub))
    (call $write-extern
      (local.get $sub)
    )
    (call_ref
      (array.get
        (local.get $sub)
        (i32.const 0)
      )
    )
  )
)
```

PR #4330, opened 2021-11-15
Directize: Fix handling of non-nullable locals and unreachability

kripken:
The order of operations could allow us to add vars but then later decide
not to do the optimization due to unreachability. And then we did not do a
fixup for non-nullability for those args, leading to a fuzzer error.

### DIFF
--- a/test/lit/passes/directize_all-features.wast
+++ b/test/lit/passes/directize_all-features.wast
@@ -728,8 +728,6 @@
   )
  )
  ;; CHECK:      (func $select-unreachable-condition (param $x i32) (param $y i32) (param $z i32)
- ;; CHECK-NEXT:  (local $3 i32)
- ;; CHECK-NEXT:  (local $4 i32)
  ;; CHECK-NEXT:  (call_indirect $0 (type $ii)
  ;; CHECK-NEXT:   (local.get $x)
  ;; CHECK-NEXT:   (local.get $y)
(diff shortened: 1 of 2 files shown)


test/lit/passes/directize_all-features.wast
```diff
@@ -807,6 +808,52 @@
...
+ ;; CHECK-NEXT:  (call_indirect $0 (type $F)
+ ;; CHECK-NEXT:   (unreachable)
+ ;; CHECK-NEXT:   (select
+ ;; CHECK-NEXT:    (i32.const 10)
+ ;; CHECK-NEXT:    (i32.const 11)
+ ;; CHECK-NEXT:    (local.get $x)
+ ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT: )
+ (func $select-non-nullable-unreachable-arg (param $x i32)
```
aheejin, line 845: Did we do anything for this in this PR?

aheejin (approved):

test/lit/passes/directize_all-features.wast
kripken: Technically the PR didn't do anything for it, correct - this wasn't broken before (since we checked unreachability before adding any vars). Still, for completeness this test means we check all the corner cases in this area to verify we never create a var if we don't optimize.

PR #4331, opened 2021-11-15
[NFC] Rename GlobalSubtyping => HeapRefining

kripken:
This pass specializes fields in heap types, so "heap subtyping" seems better.
Especially since I just realized we actually need a pass that does subtyping
on globals, which would rightly deserve the name "global subtyping" to
parallel "local subtyping" etc.


tlively (approved):
While we're considering a name change, what do you think of `HeapTypeRefining`?

kripken:
I don't feel strongly. Why do you prefer that?

From my side, "HeapSubtyping" seems reasonable as applying subtypes to heap type fields is literally what it does. I agree HeapType could be more explicit than Heap, but I don't think there's ambiguity personally.

tlively:
I'm more interested in the "refining" part, since that seems like a more descriptive verb than "subtyping." I usually don't use "subtyping" as a verb, but if I do, it's to point out a relationship between two types, e.g. "A subtypes B". In contrast, "refining" describes the transformation that pass is doing, i.e. making the types more specific.

kripken:
I see. To me the difference isn't obvious but I'm not a native speaker so HeapRefining sgtm.

tlively:
I would prefer `TypeRefining` or `HeapTypeRefining` over `HeapRefining`, since it's the types that get refined rather than the heap.

kripken:
Oh sorry, I thought we'd agreed on the final name so I did automerge... will open another PR.

PR #4335, opened 2021-11-16
Handle removed/renamed initial contents in fuzzer

kripken:
(I hit this after recently renaming a lit test...)


aheejin (approved):
Oh thanks for the catch..

PR #4345, opened 2021-11-19
[NFC] Move value numbering code to a header

kripken:
This just moves code out of RedundantSetElimination. This code will be
useful in a future PR to improve CoalesceLocals in the direction suggested
by @tlively in https://github.com/WebAssembly/binaryen/pull/4314#discussion_r745944545

### DIFF
--- a/src/ir/numbering.h
+++ b/src/ir/numbering.h
@@ -54,9 +54,7 @@ public:
   // Return a new unique value. Normally this is called internally, but there
   // are also use cases for the user of the class to call this, when they want
   // to get a new value that will not collide with any others.
-  Index getUniqueValue() {
-    return nextValue++;
-  }
+  Index getUniqueValue() { return nextValue++; }
 
 private:
   // Cache the value numbers of literals and expressions.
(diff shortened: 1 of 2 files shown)


aheejin (approved):

src/ir/numbering.h
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  // Return a new unique value. Normally this is called internally, but there
+  // are also use cases for the user of the class to call this, when they want
+  // to get a new value that will not collide with any others.
+  Index getUniqueValue() { return nextValue++; }
+
+private:
+  // Reserve 0 for possible future use as a "null value".
+  Index nextValue = 1;
```
tlively, line 59: There should be no need to reserve a value now, even if we plan to introduce a special meaning in the future. These values should be opaque to all users, so changing them shouldn't break anything.

tlively (approved):

src/ir/numbering.h
kripken: Fair point, yeah. I do find it useful to not use 0 when debugging locally, I guess mentally 0 feels "special" to me. I'll improve the comment.

kripken: I just made it 0 for now, actually.

PR #4346, opened 2021-11-19
Allow building basic HeapTypes in nominal mode

tlively:
As we work toward allowing nominal and structural types to coexist, any
difference in how they can be built or used will be an inconvenient footgun that
we will have to work around. In the spirit of reducing the differences between
the type systems, allow TypeBuilder to construct basic HeapTypes in nominal mode
just as it can in equirecursive mode.

Although this change is a net increase in code complexity for not much
benefit (wasm-opt never needs to build basic HeapTypes), it is also an
incremental step toward getting rid of separate type system modes, so I expect
it to simplify other PRs in the near future.

This change also uncovered a bug in how the type fuzzer generated subtypes of
basic HeapTypes. The generated subtypes did not necessarily have the intended
`Kind`, which caused failures in nominal subtype validation in the fuzzer.


src/wasm/wasm-type.cpp
```diff
@@ -3172,16 +3164,90 @@ std::vector<HeapType> buildNominal(TypeBuilder& builder) {
...
+  for (auto& info : infos) {
+    struct BasicTypeReplacer : HeapTypeChildWalker<BasicTypeReplacer> {
+      void noteChild(HeapType* child) {
+        if (child->isBasic()) {
+          return;
+        }
+        auto* info = getHeapTypeInfo(*child);
+        if (info->kind == HeapTypeInfo::BasicKind) {
+          *child = info->basic;
+        }
```
kripken, line 3177: Reading this code, it checks if a child is basic. Then it checks if the info of the child is basic. Can you add a comment explaining when these two diverge, as that seems non-obvious?

tlively: Done.

kripken (approved):
Lgtm, although I admit to not following all the details here. But the extra complexity does seem necessary and I agree it should be beneficial in the long run.

Running this on j2wasm, it is 3% slower to parse the optimized wasm and 10% slower to parse the unoptimized one. The unoptimized one goes from 5.7 seconds to 6.3. That's a significant slowdown, but I suppose it's acceptable, unless you have ideas?

tlively:
Thanks for measuring the performance impact of this. Once we have the ability to have equirecursive + nominal types simultaneously (and/or switch CallIndirect to storing a HeapType), we will be able to stop trying to save "canonical" nominal signature types. That will let us simplify nominal HeapTypeInfo hashing and comparisons and even skip hashing entirely when moving nominal heapTypes to the global store, which will probably help us recover some performance.

PR #4347, opened 2021-11-19
Add ChildPointerIterator/ValueChildPointerIterator

aheejin:
This adds `ChildPointerIterator` and `ValueChildPointerIterator` to
iteration.h. This allows `AbstractChildIterator` take an additional
template parameter and make the existing
`ChildIterator`/`ValueChildIterator` use it.

It's hard to add tests, but `ChildIterator`/`ValueChildIterator` are
already being used in the codebase, and `ChildPointerIterator` will be
used in my follow-up PR.

### DIFF
--- a/src/ir/iteration.h
+++ b/src/ir/iteration.h
@@ -40,8 +40,21 @@ namespace wasm {
 //                        this instruction. For example, includes If::condition
 //                        but not If::ifTrue.
 //
-template<class Specific> class AbstractChildIterator {
-  using Self = AbstractChildIterator<Specific>;
+// ChildPointerIterator and ValueChildPointerIterator iterate the same way
+// respectively, but they return pointers to children instead.
+
+template<bool isPointer> struct DereferenceKind {};
+template<> struct DereferenceKind<true> {
+  using type = Expression**;
+  static type ref(Expression** item) { return item; }
+};
+template<> struct DereferenceKind<false> {
+  using type = Expression*&;
+  static type ref(Expression** item) { return *item; }
+};
+
+template<class Specific, bool isPointer> class AbstractChildIterator {
+  using Self = AbstractChildIterator<Specific, isPointer>;
 
   struct Iterator {
     const Self& parent;
@@ -57,8 +70,11 @@ template<class Specific> class AbstractChildIterator {
 
     void operator++() { index++; }
 
-    Expression*& operator*() {
-      return *parent.children[parent.mapIndex(index)];
+    // Depending on which specilization of DereferenceKind we use, this returns
+    // either Expression* or Expression**.
+    typename DereferenceKind<isPointer>::type operator*() {
+      return DereferenceKind<isPointer>::ref(
+        parent.children[parent.mapIndex(index)]);
     }
   };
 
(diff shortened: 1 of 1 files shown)


aheejin:
Thanks for the template help @tlively!

src/ir/iteration.h
```diff
@@ -57,8 +70,11 @@ template<class Specific> class AbstractChildIterator {
 
     void operator++() { index++; }
 
-    Expression*& operator*() {
-      return *parent.children[parent.mapIndex(index)];
+    // Depending on which specilization of DereferenceKind we use, this returns
+    // either Expression* or Expression**.
```
tlively, line 74: Why do we need both `Expression*&` and `Expression**` return types? They seem fairly similar.

```suggestion
    // either Expression*& or Expression**.
```

src/ir/iteration.h
```diff
@@ -57,8 +70,11 @@ template<class Specific> class AbstractChildIterator {
 
     void operator++() { index++; }
 
-    Expression*& operator*() {
```
kripken, line 60: I'm not sure I understand the motivation here yet. The old code returns an `Expression*&` which also allows replacing the child, the same as returning `**` - the reference is to the actual pointer in the parent expression. This works as the `children` vector contains `Expression**` already (which I see this PR does not change).

Is the goal here to return `**` instead for clarity reasons? Or something else?

src/ir/iteration.h
```diff
@@ -40,8 +40,21 @@ namespace wasm {
 //                        this instruction. For example, includes If::condition
 //                        but not If::ifTrue.
 //
-template<class Specific> class AbstractChildIterator {
-  using Self = AbstractChildIterator<Specific>;
+// ChildPointerIterator and ValueChildPointerIterator iterate the same way
+// respectively, but they return pointers to children instead.
+
+template<bool isPointer> struct DereferenceKind {};
+template<> struct DereferenceKind<true> {
```
tlively, line 47: We could slightly simplify this by not using a `bool` and instead passing these structs directly as the template argument that determines what kind of dereference is done. Sorry for not thinking of this yesterday!

```suggestion
struct PointerDereference {
```

src/ir/iteration.h
```diff
@@ -40,8 +40,21 @@ namespace wasm {
...
+  using type = Expression**;
+  static type ref(Expression** item) { return item; }
+};
+template<> struct DereferenceKind<false> {
```
tlively, line 51:
```suggestion
struct NonPointerDereference {
```

src/ir/iteration.h
```diff
@@ -40,8 +40,21 @@ namespace wasm {
...
+  using type = Expression*&;
+  static type ref(Expression** item) { return *item; }
+};
+
+template<class Specific, bool isPointer> class AbstractChildIterator {
```
tlively, line 56:
```suggestion
template<class Specific, class DereferenceKind> class AbstractChildIterator {
```

src/ir/iteration.h
```diff
@@ -57,8 +70,11 @@ template<class Specific> class AbstractChildIterator {
 
     void operator++() { index++; }
 
-    Expression*& operator*() {
-      return *parent.children[parent.mapIndex(index)];
+    // Depending on which specilization of DereferenceKind we use, this returns
+    // either Expression* or Expression**.
+    typename DereferenceKind<isPointer>::type operator*() {
+      return DereferenceKind<isPointer>::ref(
```
tlively, line 76:
```suggestion
    typename DereferenceKind::type operator*() {
      return DereferenceKind::ref(
```

src/ir/iteration.h
aheejin: I tried to do this using iterator assignment based on the existing iterator but it was hard. So what I'm doing is this: https://github.com/aheejin/binaryen/blob/844cda5efcafda97c796bf20bb926d6a14749abb/src/ir/eh-utils.cpp#L26-L90

So what I've been trying to do is make a function that returns a pointer (`Expression**`) to a nested `pop` and use that pointer to assign another expression (`local.get`) in that place. The function's signature is like
```cpp
Expression* getFirstPop(Expresion* catchBody, bool& isPopNested, Expression**& popPtr);
```
So this returns the pointer to `pop`, and also returns a pointer to it as an output parameter. I don't particularly like this design though. The reason I made it return both of them is, if the catch's body itself is a `pop`, it is possible that we can get `Expression*` but not `Expression**`

I tried to make this function to return `ChildIterator::Iterator` at your suggestion, but it was somehow not easy. First of all, `ChildIterator::Iterator` is a private type so we have to make it public to use it as a function return type. Maybe that's fine. But the current structure of the `while` loop in that function is, you create an iterator at the bottom of the loop and return something at the top of the loop; the reason it is structured in this way is to handle the case that the given catch body itself is a `pop`. To make an iterator, we need a way for an iterator variable to hold its values across loop iterations, and because `ChildIterator::Iterator` doesn't have a default constructor, it was hard... Maybe I'm missing something and there's an easier way.


I guess I think I can return an iterator instead, so that I can use it to both dereference it and assign something else to its place...? I'll try that.

kripken: I think I see, thanks.

For getting the `**`, I think in C++ it is possible to convert a reference to a pointer, that is,
```cpp
void foo(T*& t) {
  T** p = &t; // convert &* into **
}
```
It's possible that's undefined behavior, though..? Not sure. But at the machine level those are the same... And in practice this is how we replace children today using this API, effectively.

For `getFirstPop`, maybe instead of returning `**` as an out param you can return a `std::variant` that has one option be a `*&`? Not sure if that's possible though, or if it's a good idea if it is...

aheejin: Oh, yeah, I can do that... Why was I thinking that was not possible.. 😨 Thanks, and sorry @tlively for taking your time... I guess than we can close this?

Not sure `std::variant` would be good for that purpose... Anyway we can talk more on that PR.

aheejin:
I'll close this, because maybe I don't need this after all: https://github.com/WebAssembly/binaryen/pull/4347#discussion_r753498490

Sorry @tlively for taking your time 😢

PR #4348, opened 2021-11-19
Add fixup function for nested pops in catch

### LINKED ISSUE #4237: Pop in catch emitted inside a block ("not enough arguments on the stack" from VM)
```wat
(module
 (tag $tag (param i32))

 (func $foo
  (local $temp i32)
  (try
   (do
    (nop)
   )
   (catch $tag
    (call $call
     (block (result i32)
      (local.set $temp
       (pop i32)
      )
      (call $nop)
      (local.get $temp)
     )
    )
   )
  )
 )

 (func $call (param i32))
 (func $nop)
)
```
In this IR the pop exists in a place that Binaryen accepts but wasm does not: we use a pop to get the catch value, but we are inside a block so the wasm does not validate, and a VM will error on something like this:
```
CompileError: WebAssembly.instantiate(): Compiling function #0 failed: not enough arguments on the stack for local.set (need 1, got 0) @+43)
```
That is with just translating that wat into a binary, `bin/wasm-opt a.wat -all -o a.out.wasm`. However, if we add a `--roundtrip` in that command, then the IR becomes this:
```wat
   (catch $tag$0
    (local.set $temp
     (pop i32)
    )
    (call $call
     (block $label$3 (result i32)
      (call $nop)
      (local.get $temp)
     )
    )
   )
```
Note how the pop was moved into a proper position. So it seems like we have some logic in the binary reading code that handles this somehow, and so using `--roundtrip` is a workaround, at least, so this is probably not urgent.

I noticed this in wasm GC that had this:
```wat
   (catch $tag$0
    (throw $tag$0
     (ref.cast
      (pop (ref null $type$16))
      (global.get $global$1)
     )
    )
   )
```
That cast can be statically removed, leading to:
```wat
   (catch $tag$0
    (throw $tag$0
     (block (result (ref null $type$16))
      (local.set $4
       (pop (ref null $type$16))
      )
      (drop
       (global.get $global$1)
      )
      (local.get $4)
     )
    )
   )
```
And that is how the problem can occur.

This does not seem specific to wasm GC, but perhaps we have more peephole opts that end up emitting small blocks, which increase the risk.

aheejin:
This adds `EHUtils::handleBlockNestedPops`, which can be called at the
end of passes that has a possibility to put `pop`s inside `block`s. This
method assumes there exists a `pop` in a first-descendant line, even
though it can be nested within a block. This allows a `pop` to be nested
within a `block` or a `try`, but not a `loop`, since that means the
`pop` can run multile times. In case of `if`, `pop` can exist only in
its condition; if a `pop` is in its true or false body, that's not in
the first-descendant line.

This can be useful when optimization passes create blocks to do          
transformations. Wrapping expressions wiith a block does not change      
semantics most of the time, but if pops happen to be inside a block      
generated by those passes, they can result in invalid binaries.

To test this, this adds `passes/test_passes.cpp`, which is intended to
contain multiple test passes that test a single (or more) utility
functions separately. Without this kind of pass, it is hard to test
various cases in which nested `pop`s can be generated in existing
passes. This PR also adds `PassRegistry::registerTestPass`, which
registers a pass that's intended only for internal testing and does not
show up in `wasm-opt --help`.

Fixes #4237.


src/ir/eh-utils.h
```diff
@@ -20,6 +20,8 @@
 namespace wasm {
 
 class Expression;
+class Function;
+class Module;
```
kripken, line 24: I don't think I noticed this before, but looks like it avoids including wasm.h by adding these forward declarations? Given there are 3 of them I think a single line that includes "wasm.h" would be shorter.

src/ir/eh-utils.cpp
```diff
@@ -16,53 +16,137 @@
...
-    // bodies, because a pop cannot be in those control flow expressions.
-    ValueChildIterator it(firstChild);
+
+    if (Properties::isControlFlowStructure(firstChild)) {
+      // If's condition is a value child who comes before an 'if' instruction
+      // in binary, it is fine if a 'pop' is in there. We don't allow a 'pop' to
+      // be in an 'if''s then or else body because they are not
+      // first descendants.
+      if (auto* if_ = firstChild->dynCast<If>()) {
+        firstChild = if_->condition;
```
kripken, line 59:
```suggestion
      if (auto* iff = firstChild->dynCast<If>()) {
        firstChild = iff->condition;
```
(this is the convention used most elsewhere, I believe)

src/ir/eh-utils.cpp
```diff
@@ -16,53 +16,137 @@
...
+      }
+      // We don't allow the pop to be included in a loop, because it cannot be
+      // run more than once
+      if (firstChild->is<Loop>()) {
+        return nullptr;
+      }
+      if (firstChild->is<Block>()) {
+        // If there are no branches that targets the implicit block, it will be
+        // removed when written back. But there are branches that target the
+        // implicit block,
```
kripken, line 70:
```suggestion
        // implicit block in examples like this:
```
(or maybe I'm not following the intent of the text here?)

src/ir/eh-utils.cpp
```diff
@@ -16,53 +16,137 @@
...
-    ValueChildIterator it(firstChild);
+
+    if (Properties::isControlFlowStructure(firstChild)) {
+      // If's condition is a value child who comes before an 'if' instruction
+      // in binary, it is fine if a 'pop' is in there. We don't allow a 'pop' to
+      // be in an 'if''s then or else body because they are not
+      // first descendants.
+      if (auto* if_ = firstChild->dynCast<If>()) {
+        firstChild = if_->condition;
+        continue;
```
kripken, line 60: I think we need to update `firstChildPtr` before continuing?

src/ir/eh-utils.cpp
```diff
@@ -16,53 +16,137 @@
...
+        } else {
+          isPopNested = true;
+        }
+      }
+      if (firstChild->is<Try>()) {
+        isPopNested = true;
+      }
+    }
+    ChildIterator it(firstChild);
     if (it.begin() == it.end()) {
```
kripken, line 93: We recently added a helper that can help here:
```suggestion
    if (it.getNumChildren() == 0) {
```

src/ir/eh-utils.cpp
```diff
@@ -16,53 +16,137 @@
...
+      //     (local.get $new)
+      //   )
+      // )
+      if (isPopNested) {
+        assert(popPtr);
+        Index newLocal = builder.addVar(func, pop->type);
+        try_->catchBodies[i] =
+          builder.makeSequence(builder.makeLocalSet(newLocal, pop), catchBody);
+        *popPtr = builder.makeLocalGet(newLocal, pop->type);
+      }
```
kripken, line 148: This should call `handleNondefaultableLocals` as well for the non-nullable local case, as I assume it's possible to pop a non-nullable type?

(All these corner cases stack up, sadly...)

src/pass.h
```diff
@@ -39,8 +39,12 @@ struct PassRegistry {
   typedef std::function<Pass*()> Creator;
 
   void registerPass(const char* name, const char* description, Creator create);
+  // Register a pass that's used for internal testing. These pass do not show up
```
kripken, line 42:
```suggestion
  // Register a pass that's used for internal testing. These passes do not show up
```

src/pass.h
```diff
@@ -39,8 +39,12 @@ struct PassRegistry {
...
+  // in --help.
+  void
+  registerTestPass(const char* name, const char* description, Creator create);
   std::unique_ptr<Pass> createPass(std::string name);
-  std::vector<std::string> getRegisteredNames();
+  std::vector<std::string> getRegisteredNames(bool includeHidden = false);
```
kripken, line 47: How about an enum for Hidden/Visible? Can be used on line 58 too.

test/lit/passes/catch-pop-fixup.wast
```diff
@@ -0,0 +1,303 @@
...
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $pop-within-block1
+    (try
+      (do)
+      (catch $e-i32
+        (throw $e-i32
+          ;; The pop is wihtin a block, so it will be handled
```
kripken, line 33:
```suggestion
          ;; The pop is within a block, so it will be handled
```

kripken (commented):
Nice, generally lgtm % comments (some of which I'm not sure about).

aheejin:
Hmm, the `hidden` property should be also in `Option` I think after all...

src/ir/eh-utils.h
aheejin: Done. I tried to avoid includes if possible (just as general recommendations), but in this case that might not matter much anyway...

src/ir/eh-utils.cpp
aheejin: Oh right, thanks for the catch... I added a new test `$pop-within-block-within-if-condition` that fails in the current code but runs correctly with this fix.

src/ir/eh-utils.cpp
aheejin: Oh, what I tried to say was "But _if_ there are branches that target the implicit block, ..."... "if" was missing. Fixed.

src/pass.h
aheejin: I moved the `hidden` property to `option`, so this code is gone now.. PTAL at the new code and let me know if that looks fine.

aheejin:
I haven't addressed all comments but I have to run for the shuttle... I'll fix the others shortly.

src/ir/eh-utils.cpp
aheejin: Thanks. I added `$pop-non-defaultable-type-within-block` to test this case too.

src/ir/eh-utils.cpp
```diff
@@ -16,54 +16,144 @@
...
+                                             implicitBlock->name)) {
+            isPopNested = true;
+          }
+        } else {
+          isPopNested = true;
+        }
+      }
+      if (firstChild->is<Try>()) {
+        isPopNested = true;
+      }
```
kripken, line 92: How about switching the sequence of ifs to a chain of `if-else`, and then add WASM_UNREACHABLE at the end? Then if we add a new control flow instruction in the future it would error here, which could help avoid some future debugging.

test/lit/passes/catch-pop-fixup.wast
```diff
@@ -1,12 +1,18 @@
 ;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
-;; RUN: wasm-opt %s --no-validation --catch-pop-fixup -all -S -o - | filecheck %s
+;; RUN: wasm-opt %s --catch-pop-fixup --no-validation -all -S -o - | filecheck %s
```
kripken, line 2: Maybe add a comment about why no-validation is used here?

kripken (approved):
lgtm % comments

test/lit/passes/catch-pop-fixup.wast
aheejin: Done

src/ir/eh-utils.cpp
aheejin: Done

PR #4350, opened 2021-11-20
Check for correct subtyping in the type fuzzer

tlively:
Check that types that were meant to have a subtype relationship actually do. To
expose the intended subtyping to the fuzzer, expose `subtypeIndices` in the
return value of the type generation function.

### DIFF
--- a/src/tools/wasm-fuzz-types.cpp
+++ b/src/tools/wasm-fuzz-types.cpp
@@ -49,14 +49,35 @@ struct Fuzzer {
     Random rand(std::move(bytes));
 
     // TODO: Options to control the size or set it randomly.
-    std::vector<HeapType> types =
-      HeapTypeFuzzer::generateHeapTypes(rand, FeatureSet::All, 20);
+    HeapTypeGenerator generator =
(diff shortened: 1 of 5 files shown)


src/tools/wasm-fuzz-types.cpp
```diff
@@ -49,14 +49,35 @@ struct Fuzzer {
...
+    for (size_t i = 0; i < types.size(); ++i) {
+      std::cout << i << ": " << types[i] << "\n";
+    }
+  }
+
+  void checkSubtypes(const std::vector<HeapType>& types,
+                     const std::vector<std::vector<Index>>& subtypeIndices) {
+    for (size_t super = 0; super < types.size(); ++super) {
+      for (size_t j = 0; j < subtypeIndices[super].size(); ++j) {
+        size_t sub = subtypeIndices[super][j];
```
kripken, line 74: Since you don't use `j`, I think this is possible:
```suggestion
      for (auto sub : subtypeIndices[super]) {
```

kripken (approved):

PR #4355, opened 2021-11-23
CoalesceLocals: Use ValueNumbering

kripken:
This removes the old hardcoded value numbering in that pass and makes
it use the new code that was split into helper code. The immediate benefit
of this is to make the code aware of identical constants: if two locals have
the same constant then they do not interfere. Future improvements to
numbering will also automatically help here.

This changes some constants in existing tests so that they keep testing
what they were testing before, and adds new tests for the new benefit here.

This implements a proposed TODO by @tlively in #4314


tlively (approved):
Nice!

PR #4357, opened 2021-11-23
wasm2js: Don't assume the existence of js assert function

sbc100:
Its seems that with this emscripten change DCE is able to remove
the `assert` JS runtime function making this call to assert fail
with `ReferenceError: assert is not defined`:

https://github.com/emscripten-core/emscripten/pull/15588

I guess this generated code is not included in the DCE graph?

### DIFF
--- a/src/wasm2js.h
+++ b/src/wasm2js.h
@@ -2868,7 +2868,7 @@ void Wasm2JSGlue::emitSpecialSupport() {
     } else if (import->base == ABI::wasm2js::ATOMIC_RMW_I64) {
       out << R"(
   function wasm2js_atomic_rmw_i64(op, bytes, offset, ptr, valueLow, valueHigh) {
-    assert(bytes == 8); // TODO: support 1, 2, 4 as well
+    // TODO: support bytes=1, 2, 4 as well as 8.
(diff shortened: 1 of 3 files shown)


src/wasm2js.h
```diff
@@ -2870,7 +2870,7 @@ void Wasm2JSGlue::emitSpecialSupport() {
     } else if (import->base == ABI::wasm2js::ATOMIC_RMW_I64) {
       out << R"(
   function wasm2js_atomic_rmw_i64(op, bytes, offset, ptr, valueLow, valueHigh) {
-    assert(bytes == 8); // TODO: support 1, 2, 4 as well
+    if (typeof assert !== undefined) assert(bytes == 8); // TODO: support 1, 2, 4 as well
```
kripken, line 2873:
```suggestion
    // TODO: support bytes=1, 2, 4 as well as 8.
```

kripken (commented):
Good find. I think the issue is that the wasm2js setup code is "special" in how we process it. Probably not worth getting into - I think we can just remove this assert. It was useful when we were bringing up wasm2js atomics, but now it's just a normal TODO in the code. lgtm with the suggestion in the comment.

PR #4365, opened 2021-12-02
SimplifyGlobals: Handle nested read-only-to-write patterns

kripken:
The general pattern is

```js
if (!global) { global = 1 }
```

This PR generalizes that to handle nested appearances,

```js
if ({
  if (!global) { global = 1 }
  !global
}) {
  global = 1
}
```

With this I can finally see no more "once" global operations on the hottest
function in the currently slowest j2wasm benchmark ("filter").

Also added a testcase for the issue you mentioned @tlively about adding
more random global.gets. That is not handled yet by this pass.


src/passes/SimplifyGlobals.cpp
```diff
@@ -208,7 +213,7 @@ struct GlobalUseScanner : public WalkerPass<PostWalker<GlobalUseScanner>> {
             // We found the get of the global. Check where its value flows to,
             // and how it is used there.
             assert(expressionStack.back() == get);
-            for (Index i = 0; i < expressionStack.size() - 1; i++) {
+            for (int i = int(expressionStack.size()) - 2; i >= 0; i--) {
```
tlively, line 216: Why reverse the iteration order?

tlively (approved):

src/passes/SimplifyGlobals.cpp
kripken: Before the order did not matter. Now it does: we start at the expression itself, then see what its parent is, and so forth, and we stop when we either hit a problem or we figure out that there is no problem.

tlively: Sorry I'm missing it, but why does the order matter now?

kripken: Imagine we have
```wat
(a
  (b
    (c)
  )
)
```
Imagine that `c` is our starting point, a get of the proper global. We look at its parent `b`. If we find that `b` uses that value and does not pass it further up (as is the case of a nested pattern) then we want to stop looking. If we did not stop looking, or if we looped in the opposite direction, we'd scan `a`, and imagine that `a` does something that looks dangerous - then we'd think that `c`'s value was consumed indirectly by `a`, and not optimize.

tlively: aha, thank you! That makes sense.

PR #4369, opened 2021-12-06
[EH] Make interpreter handle uncaught exceptions

aheejin:
When a wasm exception is thrown and uncaught in the interpreter, it
caused the whole interpreter to crash, rather than gracefully reporting
it. This fixes the problem, and also compares whether an uncaught
exception happened when comparing the results before and after
optimizations in `--fuzz-exec`. To do that, when `--fuzz-exec` is given,
we now compare results even when the function does not have return
values. Logs for some existing test have changed because of this.


src/tools/execution-results.h
```diff
@@ -106,26 +110,20 @@ struct ExecutionResults {
...
-            auto resultType = func->getResults();
-            if (resultType.isRef()) {
-              // Don't print reference values, as funcref(N) contains an index
-              // for example, which is not guaranteed to remain identical after
-              // optimizations.
-              std::cout << resultType << '\n';
-            } else {
-              std::cout << ret << '\n';
-            }
+        FunctionResult ret = run(func, wasm, instance);
```
aheejin, line 113: This part is mostly indentation changes due to the removal of `if`

aheejin:
When comparing results, now we compare exceptions. Should we compare whether a trap has occurred too? I think this can depend on some trap-related options, so if we do that, it'd better be done as a followup.

src/tools/execution-results.h
```diff
@@ -236,9 +237,12 @@ struct ExecutionResults {
         }
         arguments.push_back(Literal::makeZero(param));
       }
-      return instance.callFunction(func->name, arguments);
+      return {instance.callFunction(func->name, arguments), false};
     } catch (const TrapException&) {
       return {};
+    } catch (const WasmException& e) {
+      std::cout << "[exception thrown: " << e << "]" << std::endl;
+      return {{}, true};
```
kripken, line 245: Separate from this PR, I like this idea of tracking an exception, and I think maybe we should track traps too. That is, we could have an enum "normal, exception, trap", and then line 242 would mark "trap" like this line would mark "exception".

kripken (approved):

src/tools/execution-results.h
aheejin: Do you mind if I change to that in a followup?

kripken: Yes, followup sounds better.

aheejin: Done in #4405.

PR #4374, opened 2021-12-07
[NFC] Deduplicate Store insertion logic

tlively:
Types and HeapTypes are inserted into their respective stores either by copying
a reference to a `TypeInfo` or `HeapTypeInfo` or by moving a
`std::unique_ptr<TypeInfo>` or `std::unique_ptr<HeapTypeInfo>`. Previously these
two code paths had separate, similar logic. To reduce deduplication, combine
both code paths into a single method.


kripken (approved):
Somewhat fancy `is_same` and `constexpr` stuff, but looks like a valid use of them...

PR #4375, opened 2021-12-07
[NFC] Add a separate cache for nominal signature types

tlively:
We have always cached nominal signature types keyed on their signatures to avoid
creating extra nominal types through the `HeapType::HeapType(Signature)`
constructor. However, that logic was previously built into the HeapTypeInfo
canonicalization system. To allow that system to be simplified in future PRs,
separate the caching into its own explicit layer.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -730,8 +730,8 @@ private:
 using TypeStore = Store<TypeInfo>;
 using HeapTypeStore = Store<HeapTypeInfo>;
 
-TypeStore globalTypeStore;
-HeapTypeStore globalHeapTypeStore;
+static TypeStore globalTypeStore;
+static HeapTypeStore globalHeapTypeStore;
 
 // Specialized to simplify programming generically over Types and HeapTypes.
 template<typename T> struct MetaTypeInfo {};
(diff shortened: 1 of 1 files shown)


src/wasm/wasm-type.cpp
```diff
@@ -3241,15 +3263,24 @@ std::vector<HeapType> TypeBuilder::build() {
...
   }
-  return result;
+
+  // Note built signature types. See comment in `HeapType::HeapType(Signature)`.
+  for (auto type : results) {
+    if (type.isSignature() &&
+        (type.isNominal() || getTypeSystem() == TypeSystem::Nominal)) {
+      nominalSignatureCache.insertType(type);
+    }
+  }
```
kripken, line 3281: Why is this needed when we have it handled in the constructor of HeapType?

tlively: `TypeBuilder` only uses the lower-level constructor that takes a `uintptr_t` pointing to an underlying `HeapTypeInfo`, so that logic in the other constructor does not otherwise get a chance to run.

kripken: Thanks, right, that makes sense...

kripken (approved):

PR #4376, opened 2021-12-07
Do not track effects of immutable things

kripken:
As suggested by @aheejin , we don't use those effects now in any way, and
if we need them some day we can add them back. For now they just add
overhead and complexity.

### DIFF
--- a/src/ir/effects.h
+++ b/src/ir/effects.h
@@ -72,7 +72,6 @@ public:
   std::set<Index> localsRead;
   std::set<Index> localsWritten;
   std::set<Name> mutableGlobalsRead;
-  std::set<Name> immutableGlobalsRead;
   std::set<Name> globalsWritten;
   bool readsMemory = false;
   bool writesMemory = false;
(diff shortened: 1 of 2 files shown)


src/ir/effects.h
```diff
@@ -805,7 +789,7 @@ class EffectAnalyzer {
     if (localsWritten.size() > 0) {
       effects |= SideEffects::WritesLocal;
     }
-    if (mutableGlobalsRead.size() + immutableGlobalsRead.size() > 0) {
+    if (mutableGlobalsRead.size()) {
       effects |= SideEffects::ReadsGlobal;
```
aheejin, line 793: How about changing this to `SideEffects::ReadsMutableGlobal` not to be confusing?

aheejin (approved):

src/ir/effects.h
kripken: This part is for the C API, where I think we lean towards not making changes as much as possible, so I'm not sure. I'd rather not change it unless there is a strong reason.

aheejin: Sounds good

PR #4394, opened 2021-12-14
[NFC] Simplify HeapTypeInfo hashing and comparison

tlively:
Now that caching of "canonical" nominal signatures is handled at a separate
layer, we can remove the separate code paths for hashing and comparing
HeapTypeInfos based on their structure even in nominal mode. Now hashing and
comparing of HeapTypeInfos is uniformly handled by FiniteShapeHasher and
FiniteShapeEquator.


kripken (approved):
Great!

PR #4402, opened 2021-12-20
Adds support for blame ignore using ignoreRevsFile

kateinoigakukun:
Since version 2.23, git-blame has a feature to ignore certain commits. This feature is useful to ignore large formatting changes when blaming.


kripken (approved):
Nice, I didn't even know about that option!

PR #4404, opened 2021-12-20
[EH] Handle nested pops after inlining

aheejin:
Inlining creates additional `block`s at inlined call sites, which can be
inside a `catch`. For example:
```wast
(try
  (do)
  (catch $tag
    (call $callee
      (pop i32)
    )
  )
)
```

After inlining, this becomes
```wast
(try
  (do)
  (catch $tag
    (block $__inlined_func$callee
      (local.set $0
        (pop i32) ;; Invalid!!
      )
    (nop)
    )
  )
)
```

Now the `pop` is nested in a `block`, which makes this invalid. This PR
runs `EHUtils::handleBlockNestedPops` at the end to assign the `pop` to
a local right after the `catch`, making the code valid again:
```wast
(try
  (do)
  (catch $tag
    (local.set $new ;; New local to store `pop` result
      (pop i32)
    )
    (block $__inlined_func$callee
      (local.set $0
        (local.get $new)
      )
    (nop)
    )
  )
)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -535,3 +535,5 @@ Windows and OS X as most of the core devs are on Linux.
 [minification]: https://kripken.github.io/talks/binaryen.html#/2
 [unreachable]: https://github.com/WebAssembly/binaryen/issues/903
 [binaryen_ir]: https://github.com/WebAssembly/binaryen/issues/663
+
+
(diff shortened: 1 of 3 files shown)


tlively (approved):
LGTM!

PR #4405, opened 2021-12-21
Compare traps in ExecutionResults

aheejin:
We used to only compare return values, and in #4369 we started comparing
whether an uncaught exception was thrown. This also adds whether a trap
occurred to `ExecutionResults`. So in `--fuzz-exec`, if a program with a
trap loses the trap or vice versa, it will error out saying the result
has changed, unless either of `--ignore-implicit-traps` or
`--trans-never-happen` is set.


src/tools/execution-results.h
```diff
@@ -87,15 +87,22 @@ struct LoggingExternalInterface : public ShellExternalInterface {
 // we can only get results when there are no imports. we then call each method
 // that has a result, with some values
 struct ExecutionResults {
+  enum FunctionResultKind { Normal, Trap, Exception };
   struct FunctionResult {
     Literals values;
-    bool exception; // Whether an exception is uncaught and the function crashes
+    FunctionResultKind kind;
```
tlively, line 93: Since the `values` are only meaningful if the `FunctionResultKind` is `Normal`, wdyt of using a `std::variant<Literals, Trap, Exception>` as `FunctionResult`?

tlively (approved):

src/tools/execution-results.h
aheejin: `Trap` and `Exception` are enum values and not types though..?

tlively: You would also have to change them from being enum variants to being types:

```
struct Trap {};
struct Exception {};
using FunctionResult = std::variant<Literals, Trap, Exception>;
```

aheejin: Done

src/tools/execution-results.h
```diff
@@ -189,10 +196,19 @@ struct ExecutionResults {
         return false;
       }
       std::cout << "[fuzz-exec] comparing " << name << '\n';
-      if (!areEqual(results[name].values, other.results[name].values)) {
-        return false;
+      if (results[name].index() != other.results[name].index()) {
+        if (ignoreTrap) {
+          if (!std::holds_alternative<Trap>(results[name]) &&
+              !std::holds_alternative<Trap>(other.results[name])) {
```
tlively, line 202: You can also use `std::get_if` here and let the returned pointer convert to bool instead of using `std::holds_alternative` if you want to use a shorter function name.

tlively (approved):
Nice! LGTM.

src/tools/execution-results.h
aheejin: Done

PR #4406, opened 2021-12-23
[Fuzzer] Allow empty data in --translate-to-fuzz

aheejin:
When a parameter and a member variable have the same name within a
constructor, to access (and change) the member variable, we need to
either use `this->` or change the name of the parameter. The current
code ended up changing the parameter and didn't affect the status of the
member variable, which remained empty.

### DIFF
--- a/src/tools/fuzzing/random.cpp
+++ b/src/tools/fuzzing/random.cpp
@@ -20,8 +20,8 @@
 
 namespace wasm {
 
-Random::Random(std::vector<char>&& bytes, FeatureSet features)
-  : bytes(std::move(bytes)), features(features) {
+Random::Random(std::vector<char>&& bytes_, FeatureSet features)
+  : bytes(std::move(bytes_)), features(features) {
   // Ensure there is *some* input to be read.
   if (bytes.empty()) {
     bytes.push_back(0);
--- a/test/unit/test_fuzz_empty_data.py
+++ b/test/unit/test_fuzz_empty_data.py
@@ -1,5 +1,4 @@
 import os
-import subprocess
 import tempfile
 from scripts.test import shared
 from . import utils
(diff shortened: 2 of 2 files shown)


src/tools/fuzzing/random.cpp
```diff
@@ -20,8 +20,8 @@
 
 namespace wasm {
 
-Random::Random(std::vector<char>&& bytes, FeatureSet features)
-  : bytes(std::move(bytes)), features(features) {
```
tlively, line 24: I'm really surprised if this was incorrect. This is part of the initializer list rather than the constructor body, so I would expect it to properly initialize the member `bytes` in terms of the parameter `bytes`. This works for `features(features)`, so why wouldn't it work for `bytes(std::move(bytes))`?

test/unit/test_fuzz_empty_data.py
```diff
@@ -0,0 +1,14 @@
+import os
+import tempfile
+from scripts.test import shared
+from . import utils
+
+
+class EmptyDataFuzzTest(utils.BinaryenTestCase):
```
tlively, line 7: This is another example of a place where it would be great to have `gtest` set up :(

src/tools/fuzzing/random.cpp
aheejin: Oh no, this is fine per se, but then `bytes` on line 27 ends up pointing to the parameter `bytes`, not the member variable `bytes`. The other way we can fix this is to leave this name as is and change line 27 to `this->bytes`. Let me know if you prefer that.

test/unit/test_fuzz_empty_data.py
aheejin: How would gtest help this case? This case doesn't need things like asserts, but it's rather we check if the process crashes or not. But I have limited knowledge on gtest so there can be features I'm not aware of..

tlively: yeah, I think gtest can test for crashes or their absence as well.

src/tools/fuzzing/random.cpp
tlively: ohhh, makes sense.

tlively (approved):

PR #4408, opened 2021-12-24
[EH] Support try-delegate in interpreter

aheejin:


test/spec/exception-handling.wast
```diff
@@ -216,6 +298,10 @@
 (assert_return (invoke "rethrow_target_test1") (i32.const 2))
 (assert_return (invoke "rethrow_target_test2") (i32.const 1))
 (assert_return (invoke "rethrow_target_test3") (i32.const 1))
+(assert_return (invoke "try_delegate_caught") (i32.const 3))
+(assert_trap (invoke "try_delegate_to_catchless_try"))
```
tlively, line 302: Is there a plan to implement a new kind of assertion for inspecting the contents of exceptions?

tlively (approved):

test/spec/exception-handling.wast
aheejin: Yeah that'd be good for a followup.

PR #4409, opened 2021-12-24
[EH] Enable fuzzer with initial contents

aheejin:
This enables fuzzing EH with initial contents. fuzzing.cpp/h does not
yet support generation of EH instructions, but with this we can still
fuzz EH based on initial contents.

The fuzzer ran successfully for more than 1,000,000 iterations, with my
local modification that always enables EH and lets the fuzzer select
only EH tests for its initial contents.


src/tools/fuzzing.h
```diff
@@ -182,8 +183,12 @@ class TranslateToFuzzReader {
   // type, but should not do so for certain types that are dangerous. For
   // example, it would be bad to add an RTT in a tuple, as that would force us
   // to use temporary locals for the tuple, but RTTs are not defaultable.
+  // Also, 'pop' pseudo instruction for EH is supposed to exist only at the
+  // beginning of a 'catch' block, so it shouldn't be moved around or deleted
+  // freely.
   bool canBeArbitrarilyReplaced(Expression* curr) {
-    return curr->type.isDefaultable();
+    return curr->type.isDefaultable() &&
+           !EHUtils::containsValidDanglingPop(curr);
```
tlively, line 191: The doc comment for `containsValidDanglingPop` says "This assumes there should be at least a single pop. So given a catch body whose tag type is void or a catch_all's body, this returns false." That makes me think that `canBeArbitrarilyReplaced` will be false for any expression that contains no pops, but I assume that's not correct. Can the doc comment be clarified?

tlively (commented):
LGTM besides that one comment.

src/tools/fuzzing.h
aheejin: Given a `catch` body whose tag type is void or a `catch_all`'s body, `containsValidDanglingPop` returns false, so `canBeArbitrarilyReplaced` returns true (Note the `!`), so they are fine to be replaced. I'm not sure exactly which part is confusing or needs rewriting. Can you clarify?

tlively: Oh I see, thanks. I confused myself.

tlively (approved):

aheejin:
Just realized that I haven't enabled EH on CompareVMs, so I was only testing code validity but not actual execution. Will test for more days before landing this.

kripken (approved):

aheejin:
This ran more than 1,900,000 iterations, with my local modifications that
- Always use an initial content (The default is 50% of the time)
- Always select one of EH tests as an initial content
- Always enable EH
- Disable [adding more functions](https://github.com/WebAssembly/binaryen/blob/28665b1d8f0632216ceb2de475560c64dc260b9d/src/tools/fuzzing/fuzzing.cpp#L172-L176) (The fuzzer doesn't support generating EH instructions, these newly added functions wouldn't contain EH instructions)

Will land this.

PR #4414, opened 2021-12-26
merge-functions size optimization pass, which merges similar functions

kateinoigakukun:
Merge similar functions that only differs constant values (like immediate
operand of const and call insts) by parameterization.
Performing this pass at post-link time can merge more functions across
objects. Inspired by Swift compiler's optimization which is derived from
LLVM's one:
https://github.com/apple/swift/blob/main/lib/LLVMPasses/LLVMMergeFunctions.cpp
https://github.com/llvm/llvm-project/blob/main/llvm/docs/MergeFunctions.rst

The basic ideas here are constant value parameterization and direct callee
parameterization by indirection.

Constant value parameterization is like below:

```webassembly
;;  Before
(func $big-const-42 (result i32)
  [[many instr 1]]
  (i32.const 44)
  [[many instr 2]]
)
(func $big-const-43 (result i32)
  [[many instr 1]]
  (i32.const 45)
  [[many instr 2]]
)

;; After
(func $byn$mgfn-shared$big-const-42 (result i32)
  [[many instr 1]]
  (local.get $0) ;; parameterized!!
  [[many instr 2]]
)
(func $big-const-42 (result i32)
  (call $byn$mgfn-shared$big-const-42
   (i32.const 42)
  )
)
(func $big-const-43 (result i32)
  (call $byn$mgfn-shared$big-const-42
   (i32.const 43)
  )
)
```

Direct callee parameterization is similar to the constant value parameterization,
but it parameterizes callee function i by `ref.func` instead. Therefore it is enabled
only when reference-types and typed-function-references features are enabled.

I saw more 1 ~ 2 % reduction for SwiftWasm binary and Ruby's wasm port
using wasi-sdk, and 3 ~ 4.5% reduction for Unity WebGL binary when `Oz`.

## Oz Benchmark

- original: The original .wasm file (dwarf sections and name section are stripped)
- old: Optimized by the latest version 104 wasm-opt
- new: Optimized by wasm-opt built from this branch
- new-with-features: Optimized by wasm-opt built from this branch and direct callee parameterization enabled 
  (with `--enable-reference-types --enable-typed-function-references`)

### Code size

| | original (B) | old | new | new-with-features |
| :-- | --:|--:|--:|--: |
| [`rust_game_of_life_bg.wasm`](https://github.com/rustwasm/wasm_game_of_life) | 36765|36150 (-1.67%)|35855 (-2.48%)|35855 (-2.48%)|
| [`swift-hello.wasm`](https://github.com/swiftwasm/swift) | 4825211|2968719 (-38.47%)|2928842 (-39.3%)|2907489 (-39.74%)|
| [`ruby.wasm`](https://github.com/kateinoigakukun/ruby/pull/3) | 8592839|8428193 (-1.92%)|8302506 (-3.38%)|8270963 (-3.75%)|
| `unity-3d-webgl-template.wasm` | 7986610|7725181 (-3.27%)|7482497 (-6.31%)|7374920 (-7.66%)|

`unity-3d-webgl-template.wasm` is built from Unity's 3D project template targeting WebGL.

### Optimizer performance

|                                                  | old (ms) | new (ms) |
| :-- |--:|--: |
| `rust_game_of_life_bg.wasm` | 59.8 ± 1.2 | 57.4 ± 1.5 | 
| `swift-hello.wasm` | 5.176 ± 0.160 | 4.892 ± 0.074 | 
| `ruby.wasm` | 13.510 ± 0.278 | 12.557 ± 0.335 |
| `unity-3d-webgl-template.wasm` | 9.801 ± 0.179 | 9.291 ± 0.064 |

Measured on Mac mini (M1 2020)

### Runtime performance

|   | old (ms) | new (ms) |
| :-- |--:|--: |
| `swift-hello.wasm` | 55.4 ± 0.7 | 56.6 ± 0.5 | 
| `ruby.wasm` | 87.3 ± 0.6 | 90.8 ± 0.9 |

Measured on Mac mini (M1 2020) by wasmtime 0.31.0


kripken:
Interesting, thanks for the PR @kateinoigakukun ! I have not looked at the code yet, but some initial thoughts:

* Have you seen https://github.com/WebAssembly/binaryen/issues/1142 ? There is some discussion and investigation there of related things. The numbers there are significantly different than here, so I'm curious how the approach (or the data) differs.
* Is there data on how good the Swift one is?
* This could be done as a change to the existing DuplicateFunctionElimination pass perhaps: extending the hashing and adding more replacement there. I'm not sure if that's better or not, just curious if you've considered it?

kateinoigakukun:
@kripken 
> Have you seen Explore Similar Function Elimination (SFE) optimization #1142 ?

I haven't seen it yet, thank you for sharing!
I'm not familiar with asm.js, but as far as I read the code of the tool, my approach does almost the same things.
One possible reason for the regression of my approach is that it accepts only the difference of literals and callees, but the SFE tool does for literals and *identifiers*. So it gets more merge opportunities, I guess. (I'm not sure what kinds of things can appear as identifiers)

> Is there data on how good the Swift one is?

According to the PR, which introduces the pass, it gives ~7% reduction in Swift stdlib. https://github.com/apple/swift/pull/2462


> This could be done as a change to the existing DuplicateFunctionElimination pass perhaps: extending the hashing and adding more replacement there. I'm not sure if that's better or not, just curious if you've considered it?

Yes, I've considered it before. `DuplicateFunctionElimination` is always beneficial from performance and code size aspects. but `MergeFunctions` is not beneficial in the performance aspect due to thunk overhead, so I decided to create a new pass.

kripken:
> (I'm not sure what kinds of things can appear as identifiers)

Hmm, I think for asm.js probably function names and global names were relevant.

> Yes, I've considered it before. DuplicateFunctionElimination is always beneficial from performance and code size aspects. but MergeFunctions is not beneficial in the performance aspect due to thunk overhead, so I decided to create a new pass.

Makes sense. I agree they should be run at different times/contexts. But I think we can still share code between them if that is useful. See for example the inlining pass which has an "optimize" parameter, and the simplify-locals pass has several. But I'm not sure how much code could be shared, so it's just a thought.

kateinoigakukun:
> But I think we can still share code between them if that is useful.

I took a look at DuplicateFunctionElimination again, but I can't see a sharable code because DFE is pretty simple, I think.

src/passes/MergeFunctions.cpp
```diff
@@ -0,0 +1,731 @@
...
+#if 0
+#define MERGEFUNC_DEBUG(X)                                                     \
+  do {                                                                         \
+    X;                                                                         \
+  } while (0)
+#else
+#define MERGEFUNC_DEBUG(X)                                                     \
+  do {                                                                         \
+  } while (0)
+#endif
```
kripken, line 85: looks like this is just for debugging and can be removed?

src/passes/MergeFunctions.cpp
```diff
@@ -0,0 +1,731 @@
...
+
+  CheapHasher(Map* output) : output(output) {}
+
+  CheapHasher* create() override { return new CheapHasher(output); }
+
+  static Map createMap(Module* module) {
+    Map hashes;
+    for (auto& func : module->functions) {
+      // ensure an entry for each function - we must not modify the map shape in
+      // parallel, just the values
```
kripken, line 107:
```suggestion
      // Ensure an entry for each function - we must not modify the map shape in
      // parallel, just the values.
```
(older code is not consistent here, but we try to do better in new stuff)

src/passes/MergeFunctions.cpp
```diff
@@ -0,0 +1,731 @@
...
+        rehash(digest, 0);
+        continue;
+      }
+      hash_combine(digest, hashExpression(curr, stack));
+    }
+    return digest;
+  }
+
+  static size_t hashExpression(const Expression* curr, ExpressionStack& stack) {
+// Visit an expr to collect children
```
kripken, line 147: This can use `ChildIterator` instead which is a simpler, higher-level API

kripken (commented):
I'm not done reading yet but here are some comments so far.

src/passes/MergeFunctions.cpp
```diff
@@ -0,0 +1,731 @@
...
+    return hashes;
+  }
+
+  void doWalkFunction(Function* func) { output->at(func) = hashFunction(func); }
+
+  static size_t hashFunction(Function* func) {
+    size_t digest = 0;
+    for (Type param : func->getParams()) {
+      rehash(digest, param);
+    }
```
kripken, line 119: It is not correct to hash params like this - instead, the type should be hashed. In nominal typing in wasm GC the params may match but the type may cause them to differ.

That is handled correctly in this existing code:

https://github.com/WebAssembly/binaryen/blob/3168fa227eeb3b964b6d3c773d95e0a2014b396a/src/ir/hashed.h#L51-L53

It might be good to refactor so that code can be shared and not duplicated.

src/passes/MergeFunctions.cpp
```diff
@@ -0,0 +1,731 @@
...
+#define DELEGATE_FIELD_NAME(id, field)           // noop
+#define DELEGATE_FIELD_TYPE(id, field)           // noop
+#define DELEGATE_FIELD_HEAPTYPE(id, field)       // noop
+#define DELEGATE_FIELD_ADDRESS(id, field)        // noop
+#define DELEGATE_FIELD_SCOPE_NAME_DEF(id, field) // noop
+#define DELEGATE_FIELD_SCOPE_NAME_USE(id, field) // noop
+
+#include "wasm-delegations-fields.def"
+
+    return hash(curr->_id);
```
kripken, line 169: Maybe I'm missing something, but it looks like this just hashes the ID of the expression, but not any of its immediates? Things like the offset in a load/store etc. should be hashed as well, or they'd be considered identical incorrectly. The main expression hashing code does that, and hopefully this can reuse that code: that is, hash most things normally using that code, but switch to custom logic for the things we want to see as identical, like Consts.

src/passes/pass.cpp
```diff
@@ -569,6 +572,12 @@ void PassRunner::addDefaultGlobalOptimizationPostPasses() {
   if (options.optimizeLevel >= 2 || options.shrinkLevel >= 2) {
     addIfNoDWARFIssues("inlining-optimizing");
   }
+
+  // perform after the number of functions is reduced by inlining-optimizing
+  if (options.shrinkLevel >= 2) {
+    addIfNoDWARFIssues("merge-functions");
+  }
```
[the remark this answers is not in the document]
kateinoigakukun: Makes sense!

test/passes/merge-functions.wast
```diff
@@ -0,0 +1,68 @@
+(module
```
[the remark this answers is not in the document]
kateinoigakukun: Ok, changed them to be lit tests

src/passes/MergeFunctions.cpp
kateinoigakukun: removed

src/passes/MergeFunctions.cpp
kateinoigakukun: Makes sense. I refactored to reuse FunctionHasher as much as possible.

src/passes/pass.cpp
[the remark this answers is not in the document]
kateinoigakukun: There was a small improvement!
Screen Shot 2022-01-14 at 10 12 48

test/lit/passes/merge-functions.wast
```diff
@@ -0,0 +1,215 @@
...
+  ;; CHECK-NEXT: )
+  (func $take-param-and-local-1 (param $0 i32) (result i32)
+    (local $1 i32)
+    (nop) (nop) (nop) (nop) (nop) (nop)
+    (nop) (nop) (nop) (nop) (nop) (nop)
+    (i32.add (i32.add (i32.const 43) (local.get $0)) (local.get $1))
+  )
+)
+
+;; allowing locals indexes order difference by reoder-locals
```
kripken, line 80:
```suggestion
;; allowing locals indexes order difference by reorder-locals
```

test/lit/passes/merge-functions_all-features.wast
```diff
@@ -0,0 +1,253 @@
...
+  ;; CHECK-NEXT: )
+  (func $yes-call-callee-take-arg-0 (result i32)
+    (nop) (nop) (nop) (nop) (nop) (nop)
+    (nop) (nop) (nop) (nop) (nop) (nop)
+    (nop) (nop) (nop) (nop) (nop) (nop)
+    (call $callee-take-arg-0 (i32.const 0))
+  )
+  ;; CHECK:      (func $yes-call-callee-take-arg-1 (result i32)
+  ;; CHECK-NEXT:  (call $byn$mgfn-shared$yes-call-callee-take-arg-0
+  ;; CHECK-NEXT:   (ref.func $callee-take-arg-1)
```
kripken, line 82: There is already a test for `ref.func` in the other test file?

test/lit/passes/merge-functions_all-features.wast
```diff
@@ -0,0 +1,253 @@
...
+  ;; CHECK-NEXT:  (nop)
+  ;; CHECK-NEXT:  (call $callee-take-arg-0
+  ;; CHECK-NEXT:   (block $block (result i32)
+  ;; CHECK-NEXT:    (drop
+  ;; CHECK-NEXT:     (i32.const 0)
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:    (i32.const 0)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
```
kripken, line 121: maybe add a comment about why this case is not optimized?

src/passes/MergeFunctions.cpp
```diff
@@ -0,0 +1,591 @@
...
+#include <variant>
+#include <vector>
+
+namespace wasm {
+
+struct ConstValueDiff : public std::monostate {};
+struct ConstCalleeDiff : public std::monostate {};
+
+using ConstDiff =
+  std::variant<ConstValueDiff, Literals, ConstCalleeDiff, std::vector<Name>>;
```
kripken, line 83: It's a little hard to read this code as it is not obvious to me what `ConstDiff` is meant to be. A comment explaining it would avoid needing to infer that from the usage below. Same for the monostates above.

src/passes/MergeFunctions.cpp
```diff
@@ -0,0 +1,591 @@
...
+
+// Describes a parameter which we create to parameterize the merged function.
+struct ParamInfo {
+  // Actual values of the parameter ordered by the EquivalentClass's
```
kripken, line 87: The first mention of `EquivalentClass` is here, referred to indirectly, where the reader doesn't yet know what it is going to be.

Perhaps a large toplevel comment could explain how all the pieces fit together? Figuring it out from the code atm is not easy for me at least, sorry.

kripken (commented):
Please measure the speed impact of these hashing changes. Probably fine, but just to be sure. You can try `--duplicate-function-elimination` on a large wasm file (let me know if you need a link to one) as that will hash a lot.

kateinoigakukun:
@kripken Sorry for late reply.

> Please measure the speed impact of these hashing changes. Probably fine, but just to be sure. You can try --duplicate-function-elimination on a large wasm file (let me know if you need a link to one) as that will hash a lot.

OK, I measured DFE performance on chibi-link.wasm (which contains 63135 functions), and doesn't see big regression 👍 

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `wasm-opt.new --duplicate-function-elimination chibi-link.wasm -o /dev/null` | 1.595 ± 0.010 | 1.584 | 1.618 | 1.00 ± 0.01 |
| `wasm-opt.old --duplicate-function-elimination chibi-link.wasm -o /dev/null` | 1.595 ± 0.012 | 1.586 | 1.626 | 1.00 |

src/passes/MergeFunctions.cpp
```diff
@@ -0,0 +1,609 @@
...
+  std::vector<Expression**> uses;
+
+  ParamInfo(ConstDiff values, std::vector<Expression**> uses)
+    : values(std::move(values)), uses(uses) {}
+
+  // Returns the type of the parameter value.
+  Type getValueType(Module* module) const {
+    if (const auto literals = std::get_if<Literals>(&values)) {
+      return (*literals)[0].type;
+      return Type::i32;
```
kripken, line 119:
```suggestion
```

src/passes/MergeFunctions.cpp
```diff
@@ -0,0 +1,609 @@
...
+  }
+};
+
+// Describes the set of functions which are considered as "equivalent" (i.e.
+// only differing by some constants).
+struct EquivalentClass {
+  // Primary function in the `functions`, which will be the base for the merged
+  // function.
+  Function* primaryFunction;
+  // List of functions belonging to a same class.
```
kripken, line 149:
```suggestion
  // List of functions belonging to this equivalence class.
```

src/passes/MergeFunctions.cpp
```diff
@@ -0,0 +1,609 @@
...
+
+struct MergeFunctions : public Pass {
+  bool invalidatesDWARF() override { return true; }
+
+  void run(PassRunner* runner, Module* module) override {
+    {
+      // Canonicalize locals indices to make comparison easier.
+      PassRunner runner(module);
+      runner.setIsNested(true);
+      runner.add("reorder-locals");
```
kripken, line 183: I think this is not needed as the pass is scheduled in the global post passes, so it is after this function pass has been run. I guess it might help if the pass is run manually by itself, but I don't think it's worth the extra wasted work in the normal case for that.

kripken (commented):
How about `MergeFunctions => MergeSimilarFunctions`? I think that would be less ambiguous as it clarifies this is about similarity (otherwise it's not clear how this differs from the pass for duplicate function elimination, which also merges functions).

src/passes/MergeFunctions.cpp
```diff
@@ -0,0 +1,609 @@
...
+        assert((*it)->_id == primary->_id);
+        ++it;
+      }
+      continue;
+    }
+    // If all values are the same, skip to parameterize it.
+    if (isAllSame) {
+      continue;
+    }
+    // If the derived param is already in the params, reuse it.
```
kripken, line 435: I don't follow this. Perhaps a small example?

kripken (commented):
Looking very good! Sorry for the late review.

src/passes/MergeFunctions.cpp
kateinoigakukun: Makes sense. And I confirmed no size regression after removing reorder-locals in this pass.

src/passes/MergeFunctions.cpp
kateinoigakukun: Added an example for this case

src/passes/MergeFunctions.cpp
```diff
@@ -0,0 +1,609 @@
...
+    }
+
+    Expression*& operator*() { return *tasks.back(); }
+    bool empty() { return tasks.empty(); }
+  };
+
+  if (primaryFunction->imported()) {
+    return false;
+  }
+  DeepValueIterator primaryIt(&primaryFunction->body);
```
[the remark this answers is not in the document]
kateinoigakukun: I couldn't figure out a simpler way 🤔

src/passes/MergeSimilarFunctions.cpp
```diff
@@ -0,0 +1,626 @@
...
+ */
+
+//
+// Merge similar functions that only differs constant values (like immediate
+// operand of const and call insts) by parameterization.
+// Performing this pass at post-link time can merge more functions across
+// objects. Inspired by Swift compiler's optimization which is derived from
+// LLVM's one:
+// https://github.com/apple/swift/blob/main/lib/LLVMPasses/LLVMMergeSimilarFunctions.cpp
+// https://github.com/llvm/llvm-project/blob/main/llvm/docs/MergeSimilarFunctions.rst
```
kripken, line 25: I think a find-replace changed these but I don't think they should have been?

kripken (commented):
Thanks!

Code looks good to me. The last thing is to fuzz this carefully. Basically you can run `./scripts/fuzz_opt.py` for several hours to see if it finds anything. I'll also run some fuzzing locally.

src/passes/MergeSimilarFunctions.cpp
kateinoigakukun: Good catch 🤦‍♀ Thanks

kateinoigakukun:
@kripken Yeah, I've also started fuzz test 👍

src/passes/MergeSimilarFunctions.cpp
```diff
@@ -0,0 +1,626 @@
...
+        continue;
+      }
+
+      clazz.merge(module, params);
+    }
+  }
+
+  // Parameterize direct calls if the module supports func ref values.
+  bool isCallIndirectionEnabled(Module* module) const {
+    return module->features.hasTypedFunctionReferences();
```
kripken, line 205:
```suggestion
    return module->features.hasReferenceTypes();
```

kripken: (fuzzer noticed this before me...)

kripken: Oh, actually you use call_ref as well... so you do need both features. Apologies for the noise here.

kripken: And I think you should check them both, as you had before - again, sorry for my confusion there.

@tlively I guess we don't have a mechanism for making a feature like typed-references automatically enable a feature it depends on like reference types?

kateinoigakukun: AFAIU, "TypedFunctionReferences requires ReferenceTypes", and callee parameterization requires TypedFunctionReferences, so I thought `hasTypedFunctionReferences` is appropriate here. If you overlook something, let me know 🙏

kripken:
This is looking pretty good to me. I've fuzzed 40,000 iterations, and also ran the emscripten `-Oz` and `other` test suites. Maybe worth leaving the the fuzzer overnight as a last check.

kateinoigakukun:
Update: I've fuzzed 120,000 iterations for now, and no bug related to this was found

src/passes/MergeSimilarFunctions.cpp
tlively: > @tlively I guess we don't have a mechanism for making a feature like typed-references automatically enable a feature it depends on like reference types?

That's correct, and somewhat deliberate to keep the mental model of enabling and disabling features simpler and more predictable.

kripken:
Thanks, looks good in my fuzzing as well. Landing, thanks @kateinoigakukun ! This is a very nice feature for `-Oz` builds.

kripken (approved):

kateinoigakukun:
Thank you @kripken !

PR #4416, opened 2021-12-30
Factor out a utility for replacing types during canonicalization

tlively:
Equirecursive canonicalization generates new minimal type definitions for each
distinct type, but then it must replace the old, non-minimial definitions with
the new ones. That second part where the types are replaced is not unique to
equirecursive canonicalization; the same replacement happens with BasicKind
types and in the future the hybrid isorecursive system will also need to do that
kind of replacement. To improve code reuse and separation of concerns, factor
the type replacement logic out into a separate utility.

This change slightly slows down shape canonicalization, but shape
canonicalization is still much faster than global canonicalization. Nominal
typing performance is not affected.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -2362,7 +2362,7 @@ namespace {
 // provides logic for replacing them gradually with more canonical types.
 struct CanonicalizationState {
   // The list of types being built and canonicalized. Will eventually be
-  // returned to the user TypeBuilder user.
+  // returned to the TypeBuilder user.
   std::vector<HeapType> results;
   // The newly constructed, temporary HeapTypeInfos that still need to be
   // canonicalized.
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -2358,6 +2358,8 @@ void TypeBuilder::setNominal(size_t i) {
 
 namespace {
 
+// Helper for TypeBuilder::build() that keeps track of temporary types and
+// provides logic for replacing them gradually with more canonical types.
 struct CanonicalizationState {
   // The list of types being built and canonicalized. Will eventually be
   // returned to the user TypeBuilder user.
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -22,6 +22,7 @@
 #include <sstream>
 #include <unordered_map>
 #include <unordered_set>
+#include <variant>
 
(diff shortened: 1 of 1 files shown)


src/wasm/wasm-type.cpp
```diff
@@ -2357,6 +2358,114 @@ void TypeBuilder::setNominal(size_t i) {
 
 namespace {
 
+struct CanonicalizationState {
+  // The list of types being built and canonicalized. Will eventually be
+  // returned to the user TypeBuilder user.
```
kripken, line 2363:
```suggestion
  // returned to the TypeBuilder user.
```

src/wasm/wasm-type.cpp
```diff
@@ -2357,6 +2358,114 @@ void TypeBuilder::setNominal(size_t i) {
 
 namespace {
 
+struct CanonicalizationState {
```
kripken, line 2361: Maybe add a comment that this is a helper struct for TypeBuilder specifically?

src/wasm/wasm-type.cpp
```diff
@@ -2357,6 +2358,114 @@ void TypeBuilder::setNominal(size_t i) {
...
+};
+
+void CanonicalizationState::update(ReplacementMap& replacements) {
+  if (replacements.empty()) {
+    return;
+  }
+
+  // Update the results vector.
+  for (auto& type : results) {
+    if (auto it = replacements.find(type); it != replacements.end()) {
```
kripken, line 2413:
```suggestion
    auto it = replacements.find(type);
    if (it != replacements.end()) {
```
perhaps? I don't think we use `;` inside if anywhere in the codebase atm. (I actually didn't know that was possible...)

tlively: This is a new C++17 that we should start using. It avoids `it` being in-scope unnecessarily after the `if {}`.

kripken: I see. A downside though is the line is longer which seems less readable. And in this case we don't need to limit the scope. But I don't feel strongly, if this is common in c++17 then sgtm.

src/wasm/wasm-type.cpp
tlively: Done!

src/wasm/wasm-type.cpp
tlively: Done!

kripken (approved):

PR #4417, opened 2021-12-31
fix BinaryenTypeCreate argument numTypes type, all other numXxxs argement

chai2010:
fix BinaryenTypeCreate argument numTypes type, all other numXxxs argement


kripken (approved):
Thanks!

PR #4420, opened 2022-01-04
[EH] Fixup nested pops after reading stacky binary

aheejin:
When [reading stacky code](https://github.com/WebAssembly/binaryen/blob/6137b338c7fb37ba98b63c31225ec9cfda8cfa59/src/wasm/wasm-binary.cpp#L2656-L2684) in the binary reader, we create `block`s to
make it fit into Binaryen AST, within which `pop`s can be nested, making
the resulting AST invalid. This PR runs the fixup function after reading
each `Try` to fix this.


kripken (approved):
Using stack IR opts to generate stacky code is a nice idea here. Perhaps though it makes sense to save the output of those opts as a binary? A binary would work forever, while depending on `--optimize-stack-ir` might require updates later if we remove stack IR and replace it with something else. But I don't feel strongly as it would be very little work at that future time, I guess.

aheejin:
Hmm I didn't know we can also add binaries to lit tests... I'll change to that.

kripken (approved):

Issue #4410, opened 2021-12-25
A abort failure in wasm::WasmBinaryBuilder::visitRethrow(wasm::Rethrow*)


ZFeiXQ:
**Version:**
```
version_104
```

**command:**
```
./bin/wasm-ctor-eval POC5
```

[POC5.zip](https://github.com/WebAssembly/binaryen/files/7775566/POC5.zip)

**Result**
```
76703 abort 
```
**bt**

```
Program received signal SIGABRT, Aborted.
[----------------------------------registers-----------------------------------]
RAX: 0x0 
RBX: 0x7ffff6d47fc0 (0x00007ffff6d47fc0)
RCX: 0x7ffff6ee118b (<__GI_raise+203>:	mov    rax,QWORD PTR [rsp+0x108])
RDX: 0x0 
RSI: 0x7fffffffcbe0 --> 0x0 
RDI: 0x2 
RBP: 0x7ffff7056588 ("%s%s%s:%u: %s%sAssertion `%s' failed.\n%n")
RSP: 0x7fffffffcbe0 --> 0x0 
RIP: 0x7ffff6ee118b (<__GI_raise+203>:	mov    rax,QWORD PTR [rsp+0x108])
R8 : 0x0 
R9 : 0x7fffffffcbe0 --> 0x0 
R10: 0x8 
R11: 0x246 
R12: 0x7ffff7e4e3d8 ("/home/zxq/CVE_testing/source/binaryen/src/wasm/wasm-binary.cpp")
R13: 0x1964 
R14: 0x7ffff7e4f588 ("curr->target != DELEGATE_CALLER_TARGET")
R15: 0x5555555f7030 --> 0x30246c6562616c ('label$0')
EFLAGS: 0x246 (carry PARITY adjust ZERO sign trap INTERRUPT direction overflow)
[-------------------------------------code-------------------------------------]
   0x7ffff6ee117f <__GI_raise+191>:	mov    edi,0x2
   0x7ffff6ee1184 <__GI_raise+196>:	mov    eax,0xe
   0x7ffff6ee1189 <__GI_raise+201>:	syscall 
=> 0x7ffff6ee118b <__GI_raise+203>:	mov    rax,QWORD PTR [rsp+0x108]
   0x7ffff6ee1193 <__GI_raise+211>:	xor    rax,QWORD PTR fs:0x28
   0x7ffff6ee119c <__GI_raise+220>:	jne    0x7ffff6ee11c4 <__GI_raise+260>
[...]

PR #4422, opened 2022-01-05
Turn an assertion on not colliding with an internal name into an error

kripken:
Without this, the result in a build without assertions might be quite
confusing. See #4410

Also make the internal names more obviously internal names.


aheejin (approved):

aheejin:
It looks tests have not been updated

Issue #4413, opened 2021-12-25
A abort failure in wasm::Builder::makeFunction


ZFeiXQ:
**Version:**
```
version_104
```

**command:**
```
 wasm-dis POC8
```

[POC8.zip](https://github.com/WebAssembly/binaryen/files/7775621/POC8.zip)

**Result**
```
Aborted.
```
**bt**

```
Program received signal SIGABRT, Aborted.
[----------------------------------registers-----------------------------------]
RAX: 0x0 
RBX: 0x7ffff4416040 (0x00007ffff4416040)
RCX: 0x7ffff446018b (<__GI_raise+203>:	mov    rax,QWORD PTR [rsp+0x108])
RDX: 0x0 
RSI: 0x7fffffffb890 --> 0x0 
RDI: 0x2 
RBP: 0x7ffff45d5588 ("%s%s%s:%u: %s%sAssertion `%s' failed.\n%n")
RSP: 0x7fffffffb890 --> 0x0 
RIP: 0x7ffff446018b (<__GI_raise+203>:	mov    rax,QWORD PTR [rsp+0x108])
R8 : 0x0 
R9 : 0x7fffffffb890 --> 0x0 
R10: 0x8 
R11: 0x246 
R12: 0x7ffff799dc40 ("/home/zxq/CVE_testing/project/binaryen/src/wasm-builder.h")
R13: 0x31 ('1')
R14: 0x7ffff799dc00 ("type.isSignature()")
R15: 0xfffff700 --> 0x0
EFLAGS: 0x246 (carry PARITY adjust ZERO sign trap INTERRUPT direction overflow)
[-------------------------------------code-------------------------------------]
   0x7ffff446017f <__GI_raise+191>:	mov    edi,0x2
   0x7ffff4460184 <__GI_raise+196>:	mov    eax,0xe
   0x7ffff4460189 <__GI_raise+201>:	syscall 
=> 0x7ffff446018b <__GI_raise+203>:	mov    rax,QWORD PTR [rsp+0x108]
   0x7ffff4460193 <__GI_raise+211>:	xor    rax,QWORD PTR fs:0x28
   0x7ffff446019c <__GI_raise+220>:	jne    0x7ffff44601c4 <__GI_raise+260>
   0x7ffff446019e <__GI_raise+222>:	mov    eax,r8d
   0x7ffff44601a1 <__GI_raise+225>:	add    rsp,0x118
[------------------------------------stack-------------------------------------]
[...]

PR #4423, opened 2022-01-05
Add binary format parse check for imported function types

kripken:
Without this we hit an assertion later, which is less clear.

See #4413

### DIFF
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -2056,7 +2056,11 @@ void WasmBinaryBuilder::readImports() {
         Name name(std::string("fimport$") + std::to_string(functionCounter++));
         auto index = getU32LEB();
         functionTypes.push_back(getTypeByIndex(index));
-        auto curr = builder.makeFunction(name, getTypeByIndex(index), {});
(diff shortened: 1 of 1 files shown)


src/wasm/wasm-binary.cpp
```diff
@@ -2056,7 +2056,13 @@ void WasmBinaryBuilder::readImports() {
         Name name(std::string("fimport$") + std::to_string(functionCounter++));
         auto index = getU32LEB();
         functionTypes.push_back(getTypeByIndex(index));
-        auto curr = builder.makeFunction(name, getTypeByIndex(index), {});
+        auto type = getTypeByIndex(index);
+        if (!type.isSignature()) {
+          throwError(std::string("Imported function ") + module.str + '.' +
+                     base.str +
```
aheejin, line 2062: Can `module.str` and `base.str` be possibly null?

aheejin (approved):

src/wasm/wasm-binary.cpp
kripken: In general it can be, but not for an imported function, which is the case here.

PR #4424, opened 2022-01-05
[GC] Move heap-types.wast out of lit/test/binary/

aheejin:
Apparently it is not a binary test?


kripken:
Hmm, it does do `--roundtrip` which goes through the binary format. So maybe the question is should that directory contain only tests with binary inputs, or tests that check the binary format in some way. I don't have a preference either way.

aheejin:
To me it is more intuitive for binary/ directory to conain `wasm` and `test` files; `--roundtrip` doesn't store any binary in that directory and I think it is more consistent to treat it as just any other passes. But I don't have a strong preference either.

kripken (approved):
Makes sense, yeah, that might be more consistent.

Issue #4418, opened 2022-01-02
wasm-opt: Imported function params cause std::out_of_range when using --debuginfo


timkendrick:
I'm trying to use `wasm-opt` to optimize a hand-coded `.wat` file, but when I specify the `--debuginfo` wasm-opt CLI flag it causes an uncaught exception for WASM modules that contain imported functions that specify parameters in the type signature.

This only happens when an imported function signature specifies one or more input parameters. It seems to happen at all optimization levels (including unoptimized).

I've confirmed this behavior on `wasm-opt` v104 and v103 (I'm on macOS 10.15.7).

Minimal test case to reproduce:

```
echo '(module (func $foo (import "imports" "foo") (param i32)))' > ./foo.wat
wat2wasm --debug-names --output ./foo.wasm ./foo.wat
wasm-opt --debuginfo --output ./bar.wasm ./foo.wasm
```

Output: (error code 134)

```
warning: no passes specified, not doing any work
libc++abi.dylib: terminating with uncaught exception of type std::out_of_range
zsh: abort      wasm-opt --debuginfo --output ./bar.wasm ./foo.wasm
```

...note that this example works fine if you remove the `(param i32)` from the imported function signature.

kripken:
Interesting, this looks like three separate issues: First, wasm2wat is emitting an empty "name" for parameter zero there (zero-length string). Please file a bug for that, unless I'm missing something and that is valid somehow... Second, we should have better error messages here, which I'll look into. Third, this only happens on param names of imported functions, which it looks like we don't support yet.

timkendrick:
Thanks for taking the time to look into this!

 I've been looking into the WASM [binary encoding spec](https://webassembly.github.io/spec/core/binary/modules.html#binary-import), and I'm not sure I fully understand your explanation - when you say wat2wasm emits an empty "name" for parameter zero are you talking about the generated `functype` in the "Type Section" (I wasn't aware type params could have names)? Or are you talking about the debug name in the `.name` custom section?

As far as I understand it, the wasm binary emitted by wat2wasm for my example contains no "Function section" (/ `func` definitions). Instead it just contains an "Import section" that contains a single `import`, whose `importdesc` contains the type id `0x00 0x00` (which is the ID of the `functype` generated for the import). So given that there are no function definitions I'm not sure where it's trying to read the parameter names from?

[...]

PR #4426, opened 2022-01-06
Warn about and ignore empty local/param names in name section

kripken:
Fixes the crash in #4418

Also replace the `.at()` there with better logic to handle imported functions.


timkendrick:
Thanks for the speedy fix! I've compiled this version and can confirm that it allows modules containing imported functions to compile successfully with the `--debuginfo` argument.

As explained in [my comment on the original issue](https://github.com/WebAssembly/binaryen/issues/4418#issuecomment-1006402970) I think it's probably the wrong behavior for this to emit a warning seeing as as far as I understand it the WASM binary is perfectly legitimate - I'll leave that to you guys though; from my perspective I'm fine with spurious warnings so long as it compiles.

Thanks again for your help on this, and for all your work on `wasm-opt`, it's a lifesaver!

test/lit/binary/empty-param-name.test
```diff
@@ -0,0 +1,6 @@
+;; Test that we show a warning on an empty local name, and do not crash.
```
aheejin, line 1: - Would be good to include the contents of the wasm file
- I think this depends on the wabt's behavior that emits an empty local name? Would it be good to reproduce the behavior from our end?

kripken: Hmm, I consider wabt's behavior to be a bug. Maybe not literally, it's valid with the spec, but an empty name is just worse for debugging than an index. So it's better to not emit it. I opened a wabt issue meanwhile: https://github.com/WebAssembly/wabt/issues/1799

But I guess it's debatable. I'm ok with removing the warning (and skipping such names so we don't crash).

kripken: (added content details to the testcase as requested, thanks)

aheejin: If it's a bug from the wabt side, isn't this test relying on that bug? Would this test still work after that bug is fixed?

kripken: Yes, if wabt fixes the bug then recreating the wasm won't cause a warning. Still, I think it's good to test that we don't crash + that we warn on an invalid input in a clear way. I don't feel strongly though.

I also am ok with either warning or not warning.

aheejin: Ah right, we still have the binary.. I was confused and thought if the wabt bug is fixed this test would no longer work, sorry. This sounds OK, especially given that the test file description says the file was generated from the bug.

aheejin (approved):

kripken:
I'll land this with the warning for now. It it turns out annoying, or discussion on the wabt issue decides this isn't a wabt bug, we can always remove it.

timkendrick:
Thanks @kripken, makes sense. Thanks for all your help fixing this!

PR #4427, opened 2022-01-06
Allow import mutable globals used in Asyncify pass

nokotan:
This PR will be a solution to <https://github.com/emscripten-core/emscripten/issues/15594>.

emscripten Asyncify won't work properly in side modules, because the globals, `__asyncify_state` and `__asyncify_data`, are not synchronized between main-module and side-modules.

New pass arg, `asyncify-side-module`, will be available to make `__asyncify_state` and `__asyncify_data` imported in the instrumented wasm.


src/passes/Asyncify.cpp
```diff
@@ -1555,6 +1561,28 @@ struct Asyncify : public Pass {
...
   }
 
+  void addImportedGlobals(Module* module) {
+    Builder builder(*module);
+    {
+      auto asyncify_state = builder.makeGlobal(ASYNCIFY_STATE,
+                                               Type::i32,
+                                               builder.makeConst(int32_t(0)),
+                                               Builder::Mutable);
+      asyncify_state->module = "env";
```
kripken, line 1571:
```suggestion
      asyncify_state->module = ENV;                                         
```
I believe there is a global string for this.

src/passes/Asyncify.cpp
```diff
@@ -1555,6 +1561,28 @@ struct Asyncify : public Pass {
                                          Builder::Mutable));
   }
 
+  void addImportedGlobals(Module* module) {
+    Builder builder(*module);
+    {
+      auto asyncify_state = builder.makeGlobal(ASYNCIFY_STATE,
```
kripken, line 1567: `asyncifyState` (camelCase) is more consistent I think, for the source code (even though the wasm name might be different).

src/passes/Asyncify.cpp
```diff
@@ -1488,7 +1490,11 @@ struct Asyncify : public Pass {
                             verbose);
 
     // Add necessary globals before we emit code to use them.
-    addGlobals(module);
+    if (sideModule) {
+      addImportedGlobals(module);  
+    } else {
+      addGlobals(module);
```
kripken, line 1496: Maybe a single function taking a param `bool imported` might be simpler?

kripken (commented):
Thanks for the PR!

Overall I think this is the right direction. @sbc100 any concerns about the dynamic linking side? I think Asyncify should just work with dynamic linking once the global is properly shared, but maybe I'm missing something.

Please add a test. (see for example `./test/lit/passes/asyncify_pass-arg=asyncify-imports@env.import,env.import2.wast` for an asyncify test, and similar names)

nokotan:
I used the lit test suits for the first time, I'm not sure that the test case have been generated properly.

src/passes/Asyncify.cpp
```diff
@@ -1543,16 +1546,32 @@ struct Asyncify : public Pass {
...
-                                         Type::i32,
-                                         builder.makeConst(int32_t(0)),
-                                         Builder::Mutable));
-    module->addGlobal(builder.makeGlobal(ASYNCIFY_DATA,
-                                         Type::i32,
-                                         builder.makeConst(int32_t(0)),
-                                         Builder::Mutable));
+    {
+      auto asyncifyState = builder.makeGlobal(ASYNCIFY_STATE,
+                                               Type::i32,
```
sbc100, line 1553: Indentation look wrong here.  Perhaps `git clang-format origin/main`?

src/passes/Asyncify.cpp
```diff
@@ -1543,16 +1546,32 @@ struct Asyncify : public Pass {
...
+                                               builder.makeConst(int32_t(0)),
+                                               Builder::Mutable);
+      if (imported) {
+        asyncifyData->module = ENV;                                         
+        asyncifyData->base = ASYNCIFY_DATA;  
+      }                                           
+
+      module->addGlobal(std::move(asyncifyData)); 
+    }
```
sbc100, line 1574: Why these extra scopes?

sbc100 (commented):
Seems reasonable to me.

src/passes/Asyncify.cpp
nokotan: It was the avoidance of variable name typo. Removed because it is no longer necessary.

test/lit/passes/asyncify_pass-arg=asyncify-side-module@true.wast
```diff
@@ -0,0 +1,106 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
+;; NOTE: This test was ported using port_test.py and could be cleaned up.
+
+;; RUN: foreach %s %t wasm-opt --enable-mutable-globals --asyncify --pass-arg=asyncify-side-module@true -S -o - | filecheck %s
```
kripken, line 4:
```suggestion
;; RUN: foreach %s %t wasm-opt --enable-mutable-globals --asyncify --pass-arg=asyncify-side-module -S -o - | filecheck %s
```

kripken (approved):
Nice, lgtm except for that small suggestion.

PR #4428, opened 2022-01-06
[ctor-eval] Add --ignore-external-input option

kripken:
This is meant to address one of the main limitations of wasm-ctor-eval in
emscripten atm, that libc++ global ctors will read env vars, which means they
call an import, which stops us from evalling,

https://github.com/emscripten-core/emscripten/pull/15403#issuecomment-969585580

To handle that, this adds an option to ignore external input. When set, we can
assume that no env vars will be read, no reading from stdin, no arguments to
main(), etc. Perhaps these could each be separate options, but I think keeping it
simple for now might be good enough.


src/wasm-interpreter.h
```diff
@@ -35,10 +35,6 @@
 #include "wasm-traversal.h"
 #include "wasm.h"
 
-#ifdef WASM_INTERPRETER_DEBUG
-#include "wasm-printing.h"
-#endif
-
```
kripken, line 41: Drive-by fix I noticed while debugging. We removed that header a long time ago.

sbc100 (approved):
I don't love the name of the option but I can't think of a better one.

Do you expect us to have much use for this new option? How do you expect emscripten users to use it?

kripken:
Yeah, I'm not 100% happy with the name either... but can't seem to find anything nicer.

I was thinking maybe `EVAL_CTORS=2`, where 2 means "unsafe opts"? Or a new option `EVAL_CTORS_IGNORE_EXTERNAL_INPUT`?

I actually hope that this can become the common way to use `EVAL_CTORS`. Most programs don't use commandline args or env vars, I think? Certainly any pure-computation library won't. Other programs might want to refactor how they get input - say, turning commandline args into a mechanism called at runtime later. If ctor-eval is beneficial enough then that could be worth their time. I don't have numbers yet - will need several more PRs to improve things - but I'm hopeful that now that we do sbrk and the stack etc. all in wasm that ctor-eval could be highly useful.

PR #4429, opened 2022-01-06
[ctor-eval] Remove stack hacks

kripken:
Remove some hackish code for fastcomp's stack handling. The stack pointer arrives
in an imported global there. Upstream does not do this, so this code is completely
unneeded these days (and, frankly, kind of scary as I read it now... it modeled the
stack as separate memory from the heap...).

Remove the tests for this as well. I verified that there was nothing else in those
tests that we need to keep.


sbc100 (approved):
Not sure its important here, but the `__stack_pointer` can still be imported in some cases with the wasm backend (e.g. with dynamic linking).

kripken:
Oh, thanks, good point about dynamic linking...

If we want ctor-eval to work with dynamic linking then we'd need to figure that out. I don't think the code removed here would work, we'd need something like a manual update to the stack pointer during startup (of the final position as recorded by ctor-eval).

Issue #4401, opened 2021-12-17
Thread stack size of prebuilt binary for linux may not be enough


kateinoigakukun:
Hi, I'm using binaryen's Asyncify to port CRuby to WebAssembly.

I found that when using [prebuilt binary of binaryen 1.0.3](https://github.com/WebAssembly/binaryen/releases/download/version_103/binaryen-version_103-x86_64-linux.tar.gz), it exits with segmentation fault due to stack overflow with my binary on Linux. However, it succeed when using self-built binaryen and version 0.9.9's prebuilt binary.

Here is how to reproduce the seg fault.
```console
$ wasm-opt --asyncify -O3 --pass-arg=asyncify-ignore-imports -o miniruby.opt.wasm miniruby.wasm
```

[miniruby.wasm.zip](https://github.com/WebAssembly/binaryen/files/7732420/miniruby.wasm.zip)

After some investigations, I found that

-  the prebuilt binary links musl instead of glibc and musl's default thread stack size is much smaller than glibc's one: https://wiki.musl-libc.org/functional-differences-from-glibc.html
- a function in my binary nested too deeply https://gist.github.com/kateinoigakukun/fb8b9ddd98e96183694dde408d97e428, 
- Asyncify processes its transformation recursively: https://github.com/WebAssembly/binaryen/blob/80101484099e9451072eb8c10df69499031b3475/src/passes/Asyncify.cpp#L883

So, is it acceptable to extend the default thread size by adding `-Wl,-z,stack-size=N` for building prebuilt binaries?

kripken:
Interesting...

I think it might be better to limit the recursion there. We could rewrite the `process(..)` calls to use a stack. Specifically though I am 99% sure that this is due to heavily nested blocks (that's the only real pattern in wasm where this is an issue), and so using a stack just in the code for blocks might be enough, which would be less work.

kripken:
Side question, what part of CRuby needs asyncify? File operations? (just curious)

kateinoigakukun:
@kripken OK, I'll try to rewrite the `process` method in a few days.

> Side question, what part of CRuby needs asyncify? File operations? (just curious)

CRuby uses Asyncify to implement setjmp and longjmp in userland for non-local exits, and also to support context-switching for Fiber.
c.f. https://github.com/kateinoigakukun/ruby/tree/e3b5c3c232f151bd4de4e6bf2d48790987a4ae96/wasm

ko1:
and conservative GC

[...]

PR #4433, opened 2022-01-08
Asyncify: Use stack instead of recursive call to avoid stack overflow

kateinoigakukun:
Rewrite `AsyncifyFlow.process` to use stack instead of recursive call.

This patch resolves https://github.com/WebAssembly/binaryen/issues/4401

I did not see significant performance regressions at my end for CRuby port.

```console
$ hyperfine "wasm-opt.old --asyncify --pass-arg=asyncify-ignore-imports -o miniruby.opt miniruby" "wasm-opt.new --asyncify --pass-arg=asyncify-ignore-imports -o miniruby.opt miniruby" --warmup 2
Benchmark #1: wasm-opt.old --asyncify --pass-arg=asyncify-ignore-imports -o miniruby.opt miniruby
  Time (mean ± σ):      6.525 s ±  1.779 s    [User: 13.730 s, System: 0.800 s]
  Range (min … max):    4.493 s …  9.517 s    10 runs

Benchmark #2: wasm-opt.new --asyncify --pass-arg=asyncify-ignore-imports -o miniruby.opt miniruby
  Time (mean ± σ):      5.922 s ±  2.218 s    [User: 13.149 s, System: 0.711 s]
  Range (min … max):    4.204 s … 10.905 s    10 runs

Summary
  'wasm-opt.new --asyncify --pass-arg=asyncify-ignore-imports -o miniruby.opt miniruby' ran
    1.10 ± 0.51 times faster than 'wasm-opt.old --asyncify --pass-arg=asyncify-ignore-imports -o miniruby.opt miniruby'
```

### DIFF
--- a/src/passes/Asyncify.cpp
+++ b/src/passes/Asyncify.cpp
@@ -882,9 +882,6 @@ private:
   Index callIndex = 0;
 
   Expression* process(Expression* curr) {
-    if (!analyzer->canChangeState(curr, func)) {
-      return makeMaybeSkip(curr);
-    }
     // The IR is in flat form, which makes this much simpler: there are no
     // unnecessarily nested side effects or control flow, so we can add
     // skips for rewinding in an easy manner, putting a single if around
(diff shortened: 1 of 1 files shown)


kripken (commented):
Thanks for the PR!

Reading this, I think it could perhaps be simpler to keep the control flow closer to the old form. That is, the resume/phase mechanism adds some unnecessary complexity. What I mean is that there could be a stack of work and a stack of results. Both can contain `Expression*`. When we reach an expression the first time we push its children as additional work. When we process a node, we pop its children from the result stack. Then we use those processed children and push this node on the result stack, etc.

`ChildIterator` is a generic utility for getting the children of a node (or their number), and could help here.

I think that could be shorter, and also would avoid the new classes and virtual methods.

kateinoigakukun:
@kripken Thank you for your review!

I've tried that approach to make it simpler. (and I thought PostWalker and UnifiedExpressionVisitor are suitable for the purpose)

However I didn't employ that because deciding whether or not the expression is `process`ed depends on the state of the parent expression.

For example, if a child expression of Block expression is `!canChangeState`, it should not be `process`ed to avoid unnecessary state checks.
Therefore, in order to push works to the stack of work, we need to not only iterate through the children, but also insert checks that depend on the parent of the expression.

Such conditional pushes also introduce complexity due to push/pop balancing between the stacks of work and result.
In the case of a simple push/pop based on the number of children, it can be simply written by the child iterator. But the conditional push/pop, both pop and push sides checks the same condition.

I couldn't figure out a way to write it in a clean way...

My PR approach requires a hand-written asymmetric coroutine-like style, but without the need to consider balancing of work and result.

kateinoigakukun:
@kripken Gentle Reminder :)

kripken:
Ok, I didn't look at the entire problem yet but I wrote a patch to handle deeply nested blocks, which I suspect is the actual issue here, as mentioned previously. @kateinoigakukun can you please check if this patch fixes your use case?

```diff
diff --git a/src/passes/Asyncify.cpp b/src/passes/Asyncify.cpp
index fc867f32e..3b6718b42 100644
--- a/src/passes/Asyncify.cpp
+++ b/src/passes/Asyncify.cpp
@@ -919,10 +919,25 @@ private:
     // particular they are saved and loaded. That way if we resume from the
     // first if arm we will avoid the second.
     if (auto* block = curr->dynCast<Block>()) {
+      // Look into an inner block, if there is one. This lets us handle the
+      // common case of massively-nested blocks forming a switch in an efficient
+      // manner that avoid massive recursion.
+      auto* innerBlock = block;
+      while (1) {
+        if (innerBlock->list.size() != 1) {
+          break;
+        }
+        if (auto* next = innerBlock->list[0]->dynCast<Block>()) {
+          innerBlock = next;
+        } else {
+          break;
+        }
+      }
+
       // At least one of our children may change the state. Clump them as
       // necessary.
       Index i = 0;
-      auto& list = block->list;
+      auto& list = innerBlock->list;
       while (i < list.size()) {
         if (analyzer->canChangeState(list[i], func)) {
           list[i] = process(list[i]);
```

kateinoigakukun:
@kripken Thank you for making a patch! Unfortunately, it doesn't solve my situation because a block contains another block and br instr, and it doesn't run the path you added. (Sorry for the confusion. The disasm log I attached in the issue only shows the nested blocks, but it actually has br instr)

You can reproduce the issue with a non-asyncified ruby binary attached here: https://github.com/WebAssembly/binaryen/issues/4401#issue-1082932822

kripken:
I see, ok.

I can't reproduce the issue on my linux machine, but I guess that's because linux has large stacks... :smile: 

Here is a partial patch - I don't have time to finish and get it fully working atm, but hopefully it gives you a direction for how to implement this?

```diff
diff --git a/src/passes/Asyncify.cpp b/src/passes/Asyncify.cpp
index fc867f32e..c934256d4 100644
--- a/src/passes/Asyncify.cpp
+++ b/src/passes/Asyncify.cpp
@@ -918,14 +918,57 @@ private:
     // Note that temp locals added in this way are just normal locals, and in
     // particular they are saved and loaded. That way if we resume from the
     // first if arm we will avoid the second.
+
+    // To avoid recursion, we use stacks here. We process the work for each
+    // node in two phases as follows:
+    //
+    //  1. The "Scan" phase finds children we need to process (ones that may
+    //     change the state), and adds Scan tasks for them to the work stack.
+    //  2. The "Finish" phase runs after all children have been Scanned and
+    //     Finished. It pops the children's results from the results stack (if
+    //     there were relevant children), and then it pushes its own result.
+    //
+    struct Work {
+      Expression* curr;
+      enum {
+        Scan,
+        Finish
+      } phase;
+    };
+
+    std::vector<Work> work;
+    std::vector<Expression*> results;
+
+    work.push_back(Work{curr, Work::Scan});
+
+    while (!work.empty()) {
+      auto item = work.back();
+      work.pop_back();
+      auto* curr = work.curr;
+      auto phase = work.phase;
+
       if (auto* block = curr->dynCast<Block>()) {
+        auto& list = block->list;
+
+        // Find the children we need to process. They will be Scanned and
+        // Finished before we reach our own Finish phase.
+        if (phase == Work::Scan) {
+          work.push_back(Work{child, Work::Finish});
+          for (auto* child : list) {
+            if (analyzer->canChangeState(child, func)) {
+              work.push_back(Work{child, Work::Scan});
+            }
+          }
+          continue;
+        }
+
         // At least one of our children may change the state. Clump them as
         // necessary.
         Index i = 0;
-      auto& list = block->list;
         while (i < list.size()) {
           if (analyzer->canChangeState(list[i], func)) {
-          list[i] = process(list[i]);
+            list[i] = results.back();
+            results.pop_back();
             i++;
           } else {
             Index end = i + 1;
@@ -951,7 +994,7 @@ private:
             i = end;
           }
         }
-      return block;
+        results.push_back(block);
       } else if (auto* iff = curr->dynCast<If>()) {
         // The state change cannot be in the condition due to flat form, so it
         // must be in one of the children.
@@ -999,6 +1042,10 @@ private:
       WASM_UNREACHABLE("unexpected expression type");
     }
 
+    assert(results.size() == 1);
+    return results.back();
+  }
+
   // Possibly skip some code, if rewinding.
   Expression* makeMaybeSkip(Expression* curr) {
     return builder->makeIf(builder->makeStateCheck(State::Normal), curr);
```

kateinoigakukun:
@kripken 

> I can't reproduce the issue on my linux machine, but I guess that's because linux has large stacks... 😄

Are you linking musl libc? It's only reproducible with musl. I used musl toolchain distributed here: https://musl.cc/


I updated this PR based on your patch. Does this solve your concerns?

kripken:
Ah, ok, my not using musl must be the issue then, thanks.

Ok, the code here looks good! I also benchmarked it locally and did some fuzzing and that looks good as well. However, I do see differences in the build output, which I don't understand, so I think we need to understand those.

Attached is a wasm file containing SQLite. Running `wasm-opt --asyncify src.wasm -o a.wasm` with and without this patch leads to different results. The new build is almost 2x larger. Adding `-O1` to both commands reduces the difference to a small one, but it still remains, and is a little larger. At a quick glance, the main difference is more loads and stores - as if more locals are being spilled perhaps?

[src.wasm.zip](https://github.com/WebAssembly/binaryen/files/7928423/src.wasm.zip)

kateinoigakukun:
@kripken Hmm, I can't reproduce the output difference 🤔 
I processed the SQLite binary with main and this branch, but they produces the exactly same binary on my end.

Could I know which revision you are comparing with?

```
total 8468
-rwxr-xr-x 1 katei katei  920589 Jan 25 02:07 src.wasm
-rw-rw-r-- 1 katei katei 3874095 Jan 25 04:43 src.wasm.main
-rw-rw-r-- 1 katei katei 3874095 Jan 25 04:38 src.wasm.mypr
```

kripken:
Oh, sorry! I re-tested now and you are 100% right. I think what must have happened is that I had some local change which got mixed in with the diff from your code...

kripken (approved):

PR #4438, opened 2022-01-10
[ctor-eval] Partial evaluation

kripken:
This lets us eval part of a function but not all, which is necessary to handle
real-world things like `__wasm_call_ctors` in LLVM output, as that is the
single ctor that is exported and it has calls to the actual ctors.

To do so, we look for a toplevel block and execute its items one by one, in
a FunctionScope. If we stop in the middle, then we are performing a partial
eval. In that case, we only remove the parts of the function that we removed,
and we also serialize the locals whose values we read from the
FunctionScope.

For example, consider this:
```js
function foo() {
  return 10;
}

function __wasm_call_ctors() {
  var x;
  x = foo();
  x++;
  // We stop evalling here.
  import1();
  import2(x);
}
```
We can eval `x = foo()` and `x++`, but we must stop evalling when
we reach the first of those imports. The partially-evalled function
then looks like this:
```js
function __wasm_call_ctors() {
  var x;
  x = 11;
  import1();
  import2(x);
}
```
That is, we evalled two lines of executing code and simply removed
them, and then we wrote out the value of the local at that point, and then
the rest of the code in the function is as it used to be.


src/tools/wasm-ctor-eval.cpp
```diff
@@ -66,8 +72,9 @@ class EvallingGlobalManager {
     if (dangerousGlobals.count(name) > 0) {
       std::string extra;
       if (name == "___dso_handle") {
-        extra = "\nrecommendation: build with -s NO_EXIT_RUNTIME=1 so that "
-                "calls to atexit that use ___dso_handle are not emitted";
+        extra = RECOMMENDATION
+          "build with -s NO_EXIT_RUNTIME=1 so that "
+          "calls to atexit that use ___dso_handle are not emitted";
```
sbc100, line 77: Unrelated to this change, but I'm not sure this is true anymore is it?  

Today NO_EXIT_RUNTIME means that atexit is a dummy function but IIRC that calls to atexit are not effected and wasm-opt/LTO is requires to clean them up.

src/tools/wasm-ctor-eval.cpp
```diff
@@ -302,10 +309,10 @@ struct CtorEvalExternalInterface : EvallingModuleInstance::ExternalInterface {
 
     std::string extra;
     if (import->module == ENV && import->base == "___cxa_atexit") {
-      extra = "\nrecommendation: build with -s NO_EXIT_RUNTIME=1 so that calls "
-              "to atexit are not emitted";
+      extra = RECOMMENDATION "build with -s NO_EXIT_RUNTIME=1 so that calls "
+                             "to atexit are not emitted";
```
sbc100, line 313: Ditto

src/wasm-interpreter.h
```diff
@@ -3609,9 +3609,11 @@ template<typename GlobalManager, typename SubType> class ModuleInstanceBase {
     return flow.values;
   }
 
+  // The maximum call stack depth to evaluate into.
+  static const Index maxDepth = 250;
```
sbc100, line 3613: Should we prefer `constexpr` over `const` these days?

test/ctor-eval/results.wast.out
```diff
@@ -0,0 +1,7 @@
+(module
```
sbc100, line 1: Are these tests a good candidate for converting to `lit`?    Should we do that before add all these new ones?  @tlively ?

sbc100 (approved):

src/tools/wasm-ctor-eval.cpp
kripken: Oh, right, your recent change to atext is relevant here. I think this code is not needed actually. I can remove it in a followup.

test/ctor-eval/results.wast.out
kripken: Hmm, these might work as lit tests, yeah.

Do we want to do that now? I'm not sure of the overall priority of that migration atm.

Would we port these tests using a script anyhow like other groups? If so then I think there's no harm to adding more now. But if it would be manual then perhaps new tests should be lit...

src/wasm-interpreter.h
kripken: (I don't have an opinion on const/constexpr myself...)

sbc100: I heard that `constexpr` can be win because it means that compiler can assume to really is const.  (as opposed to `const` which can be cast away with `const_cast` and modified.. not sure that applies to basic types like..)

test/ctor-eval/results.wast.out
sbc100: No worries.. I don't want to block this.. I just recently was reminded how annoying it is to try to run one of these tests in isolation so I was hoping we could keep the momentum going, but it doesn't have to happen as part of this change.

sbc100: BTW how do you run individual tests like this?  Or do you just run the whole suite?

kripken: For an individual test I just write out the command, something like `wasm-ctor-eval test/ctor-eval/NAME.wat --ctors test1 -o /dev/stdout -S`.

Is there a way to run individual lit tests?

sbc100: Yes, you just name the file on the command line.  Each lit test is by-definition described by a single file so it fairly obvious and you can use tab completion and wildcards.   This is probably my number 1 reason for wanting to make the switch:  For example, you see a test failure on the bot (it has a name which is the name of the file) and you can simply run that one test locally to reproduce.   With the current binaryen test suite (the non-lit parts) its really hard to see what exactly failed, and how exactly one might re-run the failing part (IMHO).

For example, when I work on lld I'm always running commands like this:

```
$ ../llvm-build/bin/llvm-lit lld/test/wasm/
$ ../llvm-build/bin/llvm-lit lld/test/wasm/*pic*
```

binaryen-lit works exactly the same way I believe.

sbc100: Anyway, this change lgtm :)

kripken:
Ok, this looks good now. Before the fuzzer (custom setup in a side branch `cefuzz` for now) couldn't run for 100 iterations without breaking, and now I see it get to 10,000 and it's still going.

PR #4439, opened 2022-01-10
SafeHeap: Avoid instrumenting functions directly called from the "start"

sbc100:
The `__wasm_memory_init` function is not always used as the start
function.  Sometimes it is just one of a sequence of functions that are
called from the start function.  In order to handle this case we can
instead avoid instrumenting the start function as well as any functions
call from it.

This fixes the SAFE_HEAP + USE_PTHREADS + MAIN_MODULE configuration
which is currently broken in emscripten.

### DIFF
--- a/src/passes/SafeHeap.cpp
+++ b/src/passes/SafeHeap.cpp
@@ -100,11 +100,10 @@ struct AccessInstrumenter : public WalkerPass<PostWalker<AccessInstrumenter>> {
   }
 };
 
-struct FindCalledFunctions
-  : public WalkerPass<PostWalker<FindCalledFunctions>> {
+struct FindDirectCallees : public WalkerPass<PostWalker<FindDirectCallees>> {
 public:
-  void visitCall(Call* curr) { called.insert(curr->target); }
-  std::set<Name> called;
+  void visitCall(Call* curr) { callees.insert(curr->target); }
+  std::set<Name> callees;
 };
 
 struct SafeHeap : public Pass {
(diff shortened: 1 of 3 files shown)


src/passes/SafeHeap.cpp
```diff
@@ -117,12 +122,21 @@ struct SafeHeap : public Pass {
     // add imports
     addImports(module);
     // instrument loads and stores
-    AccessInstrumenter(getSbrkPtr).run(runner, module);
+    std::set<Name> ignoreFunctions;
+    if (module->start.is()) {
+      FindCalledFunctions findCalledFunctions;
+      findCalledFunctions.walkFunctionInModule(
+        module->getFunction(module->start), module);
```
kripken, line 129: Doesn't this need to compute the "closure" here, that is if A calls B which calls C then we want C in this set as well?

Taking a step back, though, maybe another approach is to do this at runtime. That seems safer than statically ignoring a potentially large chunk of the program. What I mean here is that if the issue is that the sbrk location has not yet been set then we can just early exit from SAFE_HEAP functions. Then they will start to check the sbrk location only after it is set to a valid (>0) value?

sbc100: Actually for my use case its enough to only avoid instrumenting the direct callees because the format of the start function is relatively fixed.  Its either:

a. `__wasm_memory_init` itself
b. A function that directly calls `__wasm_memory_init`

Its never more complex than that today.   So this change fixes the current issue.  If its OK with you we can land this as is, and we can look into a smarter/runtime solution separately?

sbc100: BTW the start function can never call user-generated code, only very specific linker-generated code, so the risk of this "infecting" other part of the program don't exist.

kripken: Good points, I guess we don't need to handle anything more generic then. Please add a comment though explaining that.

sbc100: Done.

src/passes/SafeHeap.cpp
```diff
@@ -117,12 +115,27 @@ struct SafeHeap : public Pass {
...
-    AccessInstrumenter(getSbrkPtr).run(runner, module);
+    // We avoid instrumenting the module start function of any function
+    // that it directly calls.  This is because in some cases the linker
+    // generates `__wasm_init_memory` (either as the start function or
+    // a function directly called from it) and this function is used in shared
+    // memory builds to load the passive memory segments, which in turn means
+    // that value of sbrk() is not available until after it has run.
+    std::set<Name> ignoreFunctions;
+    if (module->start.is()) {
+      FindCalledFunctions findCalledFunctions;
```
kripken, line 126:
```suggestion
      // Note that this only looks into a depth of a single call. That is enough in
      // current LLVM output as start will only contain things from a very limited
      // set of code (__wasm_init_memory etc. as mentioned above).
      FindCalledFunctions findCalledFunctions;
```

src/passes/SafeHeap.cpp
```diff
@@ -109,6 +100,13 @@ struct AccessInstrumenter : public WalkerPass<PostWalker<AccessInstrumenter>> {
   }
 };
 
+struct FindCalledFunctions
```
kripken, line 103:
```suggestion
struct FindDirectlyCalledFunctions
```

kripken (commented):
lgtm % suggestions

kripken (approved):
(clicked the wrong button)

PR #4442, opened 2022-01-11
[ctor-eval] Stop if there are any memory.init instructions

kripken:
This tool depends (atm) on flattening memory segments. That is not compatible
with memory.init which cares about segment identities.

This changes `flatten()` only by adding the check for MemoryInit. The rest is
unchanged, although I saw the other two params are not needed and I removed
them while I was there.

Diff in wasm-ctor-eval.cpp is smaller without whitespace.

Found by the fuzzer.


src/ir/memory-utils.cpp
```diff
@@ -0,0 +1,72 @@
...
+      hasMemoryInit = FindAll<MemoryInit>(func->body).list.size() > 0;
+    });
+
+  for (auto& [func, hasMemoryInit] : analysis.map) {
+    if (hasMemoryInit) {
+      return false;
+    }
+  }
+
+  auto& memory = wasm.memory;
```
kripken, line 40: Nothing is changed from here.

tlively (approved):

PR #4448, opened 2022-01-12
[ctor-eval] Use std::optional for EvalCtorOutcome

kripken:
A bit less explicit I think but definitely more idiomatic.


tlively (approved):
Thanks!

PR #4450, opened 2022-01-12
Add --no-emit-metadata option to wasm-emscripten-finalize

sbc100:
This is useful for the case where we might want to finalize
without extracting metadata.

See: https://github.com/emscripten-core/emscripten/pull/15918


kripken (approved):
Test might be nice, but sounds like maybe this is experimental still and not worth it yet?

kripken:
Can run `scripts/update_help_checks.py` to update the help tests - those are not updated by the main auto updater script yet. (Perhaps they should be?)

Issue #4452, opened 2022-01-13
SmallVector could call constructors only when relevant


kripken:
The fixed storage there is a `std::array` which constructs the items immediately. We could instead use a different approach that only calls the constructor when the size actually increases.

This looks like it might be a noticeable speedup as I see profile data in which `Literals`, a `SmallVector` of `Literal`, is fairly high. The `Literal` constructor is apparently not cheap.

kripken:
Here is the benchmark I am looking at:

```
wasm-ctor-eval --ctors=__wasm_call_ctors,main --kept-exports=main --ignore-external-input a.out.wasm -o b.out.wasm --mvp-features
```
[benchmark.zip](https://github.com/WebAssembly/binaryen/files/7866105/benchmark.zip)

kripken:
The benchmark stresses interpreter performance. That matters for wasm-ctor-eval if you run it on computationally-intensive code. But any speedups there should also help with compile times in general as it will help `--precompute` too.

kripken:
It would be great if someone wanted to pick this up. It requires some C++ "magic" that I'm not very familiar with (calling a constructor on existing memory, something like that is needed here I assume).

PR #4453, opened 2022-01-13
Add a fast path to isSubType

kripken:
Apparently constructing that object is expensive? And we call it very
frequently in the interpreter.

This is a 25% speedup on the benchmark in https://github.com/WebAssembly/binaryen/issues/4452

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -1049,6 +1049,10 @@ Type Type::get(unsigned byteSize, bool float_) {
 }
 
 bool Type::isSubType(Type left, Type right) {
+  // As an optimization, in the common case do not even construct a SubTyper.
+  if (left == right) {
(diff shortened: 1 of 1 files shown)


tlively (approved):
This is really surprising. The only thing a subtyper contains is an unordered_map that would never be touched in this otimized case. What in the world is the unordered_map constructor doing?

kripken:
I'm not an expert on x86 assembly but I tried to read it now. It looks like there are a lot of checks and jumps that happen in `SubTyper().isSubType` before we exit, even if the inputs are equal. Perhaps the optimizer reordered stuff for some reason. In any case, it may not be a huge amount of work, but it is a lot of work compared to a guaranteed early exit. That is, we call this method in each interpreter execution of an instruction, so even just adding a call + a few jumps is apparently a big deal...

PR #4454, opened 2022-01-13
Add fast paths for Literals::getType

kripken:
In the common case, avoid allocating a vector and calling malloc.

This makes us over 3x faster on the benchmark in https://github.com/WebAssembly/binaryen/issues/4452

### DIFF
--- a/src/literal.h
+++ b/src/literal.h
@@ -693,6 +693,12 @@ public:
   Literals(size_t initialSize) : SmallVector(initialSize) {}
 
   Type getType() {
+    if (empty()) {
+      return Type::none;
+    }
+    if (size() == 1) {
+      return (*this)[0].type;
+    }
     std::vector<Type> types;
     for (auto& val : *this) {
       types.push_back(val.type);


tlively (approved):

src/literal.h
```diff
@@ -693,6 +693,12 @@ class Literals : public SmallVector<Literal, 1> {
   Literals(size_t initialSize) : SmallVector(initialSize) {}
 
   Type getType() {
+    if (empty()) {
+      return Type::none;
+    }
+    if (size() == 1) {
+      return (*this)[0].type;
+    }
     std::vector<Type> types;
```
aheejin, line 702: Would it be better to make it `SmallVector` too?

aheejin (approved):

src/literal.h
kripken: The class itself here is a SmallVector. The problem is the code it calls only accepts a normal one, so it converts, which is kind of annoying... perhaps we could add a template to support more inputs there, or something like that. Could be worth looking into.

(Regardless this change remains useful I think since it avoids the Type() constructor entirely in common cases.)

PR #4455, opened 2022-01-13
Refactor ModuleUtils::collectHeapTypes

tlively:
Move the code into a new .cpp file because it is too complicated for a header
and update the API to make both the type indices and optimized sorting optional.
It will become more important to avoid unnecessary sorting once isorecursive
types have been implemented because they will make the sorting more complicated.

### DIFF
--- a/src/ir/CMakeLists.txt
+++ b/src/ir/CMakeLists.txt
@@ -6,6 +6,7 @@ set(ir_SOURCES
   intrinsics.cpp
   lubs.cpp
   memory-utils.cpp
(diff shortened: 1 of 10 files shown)


kripken:
Would you mind making an NFC change first that just moves it to another file? Just that the complexity of that code is not small so I'd like to see the actual changes clearly.

tlively:
sure!

tlively:
@kripken, the diff is much smaller now. PTAL!

kripken (approved):

PR #4456, opened 2022-01-13
Optimize Literal constructors and destructor

kripken:
Handle the `isBasic()` case first - that inlined function is very fast to call,
and it is the common case. Also, do not do unnecessary work there: just
write out what we need, instead of always doing a memcpy of 16 bytes.

This makes us over 2x faster on the benchmark in https://github.com/WebAssembly/binaryen/issues/4452

### DIFF
--- a/src/wasm/literal.cpp
+++ b/src/wasm/literal.cpp
@@ -30,18 +30,40 @@ namespace wasm {
 template<int N> using LaneArray = std::array<Literal, N>;
 
 Literal::Literal(Type type) : type(type) {
-  if (type == Type::i31ref) {
-    // i31ref is special in that it is non-nullable, so we construct with zero
-    i32 = 0;
-  } else {
-    assert(type != Type::unreachable && !type.isNonNullable());
-    if (isData()) {
-      new (&gcData) std::shared_ptr<GCData>();
-    } else if (type.isRtt()) {
-      new (this) Literal(Literal::makeCanonicalRtt(type.getHeapType()));
-    } else {
-      memset(&v128, 0, 16);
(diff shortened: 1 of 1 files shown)


tlively (approved):
It's unfortunate that moving these things around has such a big performance impact. It would be easily to accidentally make performance worse again with an innocuous-looking change. I also wonder what effect using `std::variant` instead of a raw union would have on performance.

src/wasm/literal.cpp
```diff
@@ -93,49 +140,18 @@ Literal::Literal(const Literal& other) : type(other.type) {
       }
     }
   }
-  TODO_SINGLE_COMPOUND(type);
```
aheejin, line 96: Was this deleted intentionally?

aheejin (approved):

src/wasm/literal.cpp
kripken: Yes. Before it was guarding the `type.getBasic()` call right after it, but now there is nothing there to guard.

kripken:
@tlively 

Agreed, those things would be good to look into as well. Indeed it is odd that this matters so much for perf.

src/wasm/literal.cpp
aheejin: I mean, shouldn't we move this with the `switch` too?

kripken: We already have a check `if (type.isBasic())` right before it, so it is safe to do `type.getBasic()` there, unless I'm missing something.

aheejin: Ah right, sorry for the confusion!

PR #4457, opened 2022-01-13
Remove old EffectAnalyzer hacks for asm.js debugInfo

kripken:
In asm2wasm we modelled debugInfo using special imports. And we tried
 to not move them around much. Current debugInfo is tracked on
instructions and is not affected by removing this.

This may have some tiny effect beneficial effect on code size in debug
builds, perhaps.


sbc100 (commented):
rubberstamp lgtm

PR #4463, opened 2022-01-19
SAFE_HEAP: Avoid annotating any function reachable from start function

sbc100:
Since https://reviews.llvm.org/D117412 landed it has causes a bunch of
SAFE_HEAP tests in emscripten to start failing, because
`__wasm_apply_data_relocs` can now sometimes be called from with
`__wasm_init_memory` as opposed to directly from the start function.

### DIFF
--- a/src/passes/SafeHeap.cpp
+++ b/src/passes/SafeHeap.cpp
@@ -22,6 +22,7 @@
 
 #include "asmjs/shared-constants.h"
 #include "ir/bits.h"
+#include "ir/find_all.h"
 #include "ir/import-utils.h"
 #include "ir/load-utils.h"
 #include "pass.h"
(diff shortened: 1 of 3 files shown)


src/passes/SafeHeap.cpp
```diff
@@ -100,10 +101,30 @@ struct AccessInstrumenter : public WalkerPass<PostWalker<AccessInstrumenter>> {
...
+  void addFunction(Name name) {
+    functions.insert(name);
+    toVisit.insert(name);
+  }
+
+  void visitCall(Call* curr) {
+    if (functions.count(curr->target) > 0) {
+      return;
+    }
+    addFunction(curr->target);
```
kripken, line 115: Given how simple this operation is, I suggest using `FindAll<Call>`. No need to even define this class etc.

sbc100: Done

src/passes/SafeHeap.cpp
```diff
@@ -100,11 +101,31 @@ struct AccessInstrumenter : public WalkerPass<PostWalker<AccessInstrumenter>> {
...
-};
+static std::set<Name> findCalledFunctions(Module* module, Name startFunc) {
+  std::set<Name> called;
+  std::vector<Name> toVisit;
+
+  auto addFunction = [&](Name name) {
+    if (called.count(name) == 0) {
+      called.insert(name);
+      toVisit.push_back(name);
+    }
```
kripken, line 112: I am pretty sure @tlively would like you to do `if (called.insert(name).second) toVisit.push...`

src/passes/SafeHeap.cpp
```diff
@@ -100,11 +101,31 @@ struct AccessInstrumenter : public WalkerPass<PostWalker<AccessInstrumenter>> {
...
+  };
+
+  if (startFunc.is()) {
+    addFunction(startFunc);
+    while (!toVisit.empty()) {
+      auto next = toVisit.back();
+      toVisit.pop_back();
+      auto func = module->getFunction(next);
```
kripken, line 120:
```suggestion
      auto* func = module->getFunction(next);
```

src/passes/SafeHeap.cpp
```diff
@@ -100,11 +101,31 @@ struct AccessInstrumenter : public WalkerPass<PostWalker<AccessInstrumenter>> {
...
+      for (auto call : FindAll<Call>(func->body).list) {
```
kripken, line 121:
```suggestion
      for (auto* call : FindAll<Call>(func->body).list) {
```

kripken (commented):
lgtm % comments

kripken (commented):
lgtm with a test, unless this is tested elsewhere?

sbc100:
Updated test.

PR #4464, opened 2022-01-19
Parse, create, and print isorecursive recursion groups

tlively:
In `--hybrid` isorecursive mode, associate each defined type with a recursion
group, represented as a `(rec ...)` wrapping the type definitions in the text
format. Parse that text format, create the rec groups using a new TypeBuilder
method, and print the rec groups in the printer.

The only semantic difference rec groups currently make is that if one type in a
rec group will be included in the output, all the types in that rec group will
be included. This is because changing a rec group in any way (for example by
removing a type) changes the identity of the types in that group in the
isorecursive type system. Notably, rec groups do not yet participate in
validation, so `--hybrid` is largely equivalent to `--nominal` for now.


src/passes/Print.cpp
```diff
@@ -2582,7 +2582,9 @@ struct PrintSExpression : public UnifiedExpressionVisitor<PrintSExpression> {
 
   void handleSignature(HeapType curr, Name name = Name()) {
     Signature sig = curr.getSignature();
-    if (!name.is() && getTypeSystem() == TypeSystem::Nominal) {
+    bool nominal = !name.is() && (getTypeSystem() == TypeSystem::Nominal ||
```
kripken, line 2585: `nominal` seems like a confusing name here. That is, `foo = isFoo() || isBar()` just feels odd...

How about `notStructural`, if nothing better?

src/tools/fuzzing/heap-types.cpp
```diff
@@ -42,8 +42,12 @@ struct HeapTypeGeneratorImpl {
...
   // before we actually generate the types.
   using BasicKind = HeapType::BasicHeapType;
-  struct SignatureKind {};
-  struct DataKind {};
+  struct SignatureKind {
+    SignatureKind() = default;
+  };
+  struct DataKind {
+    DataKind() = default;
+  };
```
kripken, line 50: I am afraid to ask, but what weird C++ reason that I won't understand required this change?

src/wasm-type.h
```diff
@@ -427,6 +431,18 @@ class HeapType {
...
 
+// A recursion group consisting of one or more HeapTypes.
+class RecGroup {
+  uintptr_t id;
+
+public:
+  explicit RecGroup(uintptr_t id) : id(id) {}
+  bool operator==(const RecGroup& other) { return id == other.id; }
+  bool operator!=(const RecGroup& other) { return id != other.id; }
+  std::vector<HeapType> getHeapTypes() const;
```
kripken, line 442: It's a little odd to not see the types stored in this RecGroup struct. Perhaps add a comment explaining where they are stored? In particular that would help explain the complexity of getHeapTypes here (if it's doing a global lookup that might be slow, etc., and worth mentioning).

Also, is this returning by value intentionally, I assume? (any chance it can be `const&`?)

src/wasm/wasm-type.cpp
```diff
@@ -1241,6 +1252,35 @@ HeapType HeapType::getLeastUpperBound(HeapType a, HeapType b) {
...
+    // Mark the low bit to signify that this is a trivial recursion group and
+    // points to a heap type info rather than a vector of heap types.
+    return RecGroup(id | 1);
+  }
+}
+
+std::vector<HeapType> RecGroup::getHeapTypes() const {
+  if (id & 1) {
+    // This is a trivial recursion group. Mask off the low bit to recover the
+    // single HeapType.
```
kripken, line 1269: Where is this encoding defined/documented? Sorry if I missed it.

test/lit/isorecursive-whole-group.wast
```diff
@@ -0,0 +1,21 @@
...
+
+;; CHECK:      (rec
+;; CHECK-NEXT:  (type $used (struct_subtype data))
+;; CHECK-NEXT:  (type $unused (struct_subtype data))
+;; CHECK-NEXT: )
+
+ (rec
+  (type $used (struct_subtype data))
+  (type $unused (struct_subtype data))
+ )
```
kripken, line 18: Please add a test of a rec group of size 1, which gets different code paths in this PR. Also of size 0, if that is valid?

src/tools/fuzzing/heap-types.cpp
tlively: Oh, actually this is left over from when my compiler started emitting spurious warnings about unitialized memory. This was my attempt to fix it, but it didn't work. I ended up just switching compilers for the time being. Will revert this.

src/wasm-type.h
tlively: Will add a comment. Unfortunately we have to return by value because the singleton rec groups do not have a vector materialized anywhere in memory and have to be created on the fly. (Alternatively we could change the API to return nothing if the rec group is a singleton or we could materialize permanent vectors for single rec groups).

src/wasm/wasm-type.cpp
tlively: It's not, really 😅 Will add documentation.

test/lit/isorecursive-whole-group.wast
tlively: Sounds good, will add a test for a group of size 1. Groups of size zero cannot exist, since the only way to get a rec group is to query a `HeapType` that must be a member of the returned group.

src/ir/module-utils.cpp
```diff
@@ -159,11 +159,53 @@ Counts getHeapTypeCounts(Module& wasm) {
...
+        newTypes.insert(type);
+        counts.note(type);
+      }
+    }
   }
 
   return counts;
 }
 
+void coalesceRecGroups(IndexedHeapTypes& indexedTypes) {
```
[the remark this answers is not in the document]
tlively: This is in an anonymous namespace, so `static` shouldn't be necessary.

src/passes/Print.cpp
tlively: Hmm, I guess I can use `hasSupertype`.

tlively:
@kripken, your comments should be addressed now!

src/wasm-type.h
kripken: Ah, interesting, the singleton case is what I didn't connect. Makes sense.

I do worry about speed here. I suppose the common case is going to be a single huge rec group for everything, so finding a fast API for that (and not focusing on speed for singletons) would be good enough. Returning nothing if the rec group is a singleton, as you said, sounds good, but maybe the name would need to be changed (separate functions for the singleton and multiple cases?)

src/wasm/wasm-type.cpp
```diff
@@ -1232,6 +1242,41 @@ HeapType HeapType::getLeastUpperBound(HeapType a, HeapType b) {
   return TypeBounder().getLeastUpperBound(a, b);
 }
 
+// Recursion groups with single elements are encoded as that single element's
+// type ID with the low bit set and other recursion groups are encoded with the
+// address of the vector containing their members. These encodings are disjoint
+// because the alignment of the vectors is greater than 1.
+static_assert(alignof(std::vector<HeapType>) > 1);
```
kripken, line 1249: If the case of singleton rec groups is rare (again, I assume a single big rec group will be most common), maybe it's not important to optimize for this? That is, we could have normal rec groups for singletons.

src/wasm-type.h
tlively: My hope is that J2CL (and other producers) would be able to use some new form of `--nominal` so that the input they send to Binaryen is still nominal types, then Binaryen could emit them as the standard isorecursive types. I don't know how big the rec groups will be in that case, but at least it will be smaller than if we just had a single huge rec group.

But yeah, we can do better here. I'll implement a custom iterator so we can still treat trivial and nontrivial recursion groups uniformly (which is going to be important for keeping users uncomplicated).

kripken: I'm missing something - why wouldn't there be a single big rec group in that case?

tlively: The nominal-like input in that case would not have any rec groups (or equivalently would only have single-member rec groups) with relaxed validation rules that would allow forward references and mutual recursion.

kripken: Is this discussed anywhere? It sounds unfamiliar to me, and I don't understand it.

Here is what I know (or think I know): Right now j2cl emits nominal typing, which is equivalent to a single big rec group. We want to continue supporting that because that gives the maximally nominal/precise types. And j2cl does not have a use case for linking, which is what multiple rec groups are meant to help with. If so, what is the use case for getting j2cl to emit something else than that?

tlively: I don't think this is laid out anywhere, but it probably should be.

Yes, we want J2CL to emit a single rec group and we want Binaryen to optimize the types within it as if they were nominal types, meaning that Binaryen should be free to make changes to types in that rec group and to reorder types within that rec group. But if we internally represented types in that use case as belonging to a single large rec group, changing a single type or reordering types would require regenerating and replacing every type in the module because rec groups are immutable, ordered, and part of a type's identity. So for that use case we actually want to literally use nominal types in Binaryen and not use rec groups in the internal representation.

The rec groups in this PR are necessary for a correct and safe implementation of the hybrid system that does not make externally visible changes to types during optimization. That's not a code path optimized J2CL builds will ever use, but it is important for our benchmarking and validation of the hybrid type system proposal.

I can write a short doc about how I see this working in the long term and my plan for benchmarking the hybrid types system proposal, if that would be helpful.

kripken: Interesting... not sure I follow yet though.

Yes, rec groups are immutable, but so are types, and I'd expect them to be modified in tandem when they are modified at all, which is very rare. That is, the existing TypeUpdater class, used in the few passes that it is, is the one place where we actually modify types, and that would be the exact place where we also modify rec groups. (For the single big rec group case, the work there would be trivial.)

The second issue is reordering. I'm not sure I see the problem there - the order only matters when emitting the binary, so we just need to pick a deterministic order in a rec group (keeping the existing order would be ok, if rec groups store the order).

And if j2cl won't emit multiple rec groups as I think you just agreed, I'm not sure how we can realistically do much work for the multi-rec-group case. My vague notion is that we'd test as best we can on artificial stuff, but that's about it?

A short doc might help! I still don't see the plan here.

tlively: Hmm, good point about the types all being regenerated and replaced at once anyway. Not being able to reorder types when emitting the binary would lose us our code size optimization of emitting the most frequently referenced types first, but I suppose the effect of that is probably small. I'll go ahead and simplify the code here to materialize each rec group rather than optimizing for singleton groups.

tlively: Oh, and using a single rec group means that we would need a new optimization pass to explicitly drop unused types, whereas that's currently done implicitly when we collect heap types to emit. That might have a larger effect on code size.

tlively: Actually, there's another reason to special-case the singleton rec groups. It lets users of the types API uniformly assume that all non-basic types have rec groups, which saves a bunch of `if (type system is isorecursive)` guards in a bunch of places. It also simplifies the code here in wasm-type.cpp because it avoids having to create a recursion group on every code path we have for creating a heap type. Instead, we can create rec groups only when there is a nontrivial rec group to create. Overall the code is much simpler if we concentrate this tiny bit of complexity in the `RecGroup` class.

src/wasm/wasm-type.cpp
```diff
@@ -3036,14 +3105,22 @@ void canonicalizeEquirecursive(CanonicalizationState& state) {
...
+
   // Nominal types do not require separate canonicalization, so just validate
   // that their subtyping is correct.
 
 #if TIME_CANONICALIZATION
   auto start = std::chrono::steady_clock::now();
 #endif
 
-  assert(typeSystem == TypeSystem::Nominal);
+  // assert(typeSystem == TypeSystem::Nominal);
```
kripken, line 3123: Was this intentional? Or should it be removed?

kripken (approved):

src/wasm/wasm-type.cpp
tlively: Yeah this can probably just be removed. It's not a very valuable assertion.

PR #4467, opened 2022-01-20
Update heuristic for finding `__stack_pointer` to allow exports. NFC

sbc100:
There is no reason the `__stack_pointer` global can't be exported from
the module, and in fact I'm experimenting with a non-relocatable main
module that requires this.

See https://github.com/emscripten-core/emscripten/issues/12682

This heuristic still kind of sucks but should always be good enough
for llvm output that always puts the stack pointer first.

### DIFF
--- a/src/wasm/wasm-emscripten.cpp
+++ b/src/wasm/wasm-emscripten.cpp
@@ -55,7 +55,7 @@ bool isExported(Module& wasm, Name name) {
 
 Global* getStackPointerGlobal(Module& wasm) {
   // Assumption: The stack pointer is either imported as __stack_pointer or
-  // we just assume its the first non-imported global.
+  // we just assume it's the first non-imported global.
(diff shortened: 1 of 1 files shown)


sbc100:
Verified that the `other` and `core2` test suites both continue to pass with this change.

src/wasm/wasm-emscripten.cpp
```diff
@@ -55,15 +55,16 @@ bool isExported(Module& wasm, Name name) {
 
 Global* getStackPointerGlobal(Module& wasm) {
   // Assumption: The stack pointer is either imported as __stack_pointer or
-  // its the first non-imported and non-exported global.
+  // we just assume its the first non-imported global.
```
kripken, line 58:
```suggestion
  // we just assume it's the first non-imported global.
```

kripken (approved):

PR #4468, opened 2022-01-20
Reset global type state between tests

tlively:
Add a `destroyAllTypes` function to clear the global state of the type system
and use it in a custom gtest test fixture to ensure that each test starts and
ends with a fresh state.

### DIFF
--- a/src/wasm-type.h
+++ b/src/wasm-type.h
@@ -48,7 +48,7 @@ TypeSystem getTypeSystem();
 // Dangerous! Frees all types and heap types that have ever been created and
 // resets the type system's internal state. This is only really meant to be used
 // for tests.
-void destroyAllTypes();
+void destroyAllTypesForTestingPurposesOnly();
 
 // The types defined in this file. All of them are small and typically passed by
 // value except for `Tuple` and `Struct`, which may own an unbounded amount of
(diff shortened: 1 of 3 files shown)


test/gtest/type-builder.cpp
```diff
@@ -3,18 +3,36 @@
...
+  TypeSystem originalSystem;
+
+protected:
+  void SetUp() override {
+    destroyAllTypes();
+    originalSystem = getTypeSystem();
+    setTypeSystem(system);
+  }
+  void TearDown() override {
+    destroyAllTypes();
```
kripken, line 17: Seems somewhat unnecessary to destroy in both places, but can't hurt I guess...

src/wasm-type.h
```diff
@@ -45,6 +45,11 @@ void setTypeSystem(TypeSystem system);
 
 TypeSystem getTypeSystem();
 
+// Dangerous! Frees all types and heap types that have ever been created and
+// resets the type system's internal state. This is only really meant to be used
+// for tests.
+void destroyAllTypes();
```
kripken, line 51: How about `_destroyAllTypes` or `DO_NOT_CALL_destroyAllTypes` or such? That would make it clearer at the callsites (so someone reading through the code that uses this wouldn't get the wrong impression).

tlively:
I went with `destroyAllTypesForTestingPurposesOnly`, which should stick out like a sore thumb if anyone tries to call it 😁

kripken (approved):
Great!

PR #4469, opened 2022-01-21
Create `ParentIndexIterator` to reduce iterator boilerplate

tlively:
Add a utility class for defining all the common operations like pre- and post-
increment and decrement, addition and subtraction, and assigning addition and
subtraction for iterators that are comprised of a parent object and an index
into that parent object. Use the new utility to reduce the boilerplate in
wasm-type.h. Add a new test of the iterator behavior.

### DIFF
--- a/src/support/parent_index_iterator.h
+++ b/src/support/parent_index_iterator.h
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2022 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef wasm_support_parent_index_iterator_h
+#define wasm_support_parent_index_iterator_h
+
+#include <cstddef>
+#include <iterator>
+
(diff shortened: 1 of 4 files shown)


src/support/parent_index_iterator.h
```diff
@@ -0,0 +1,96 @@
...
+  bool operator!=(const Iterator& other) const { return !(*this == other); }
+  Iterator& operator++() {
+    ++index;
+    return self();
+  }
+  Iterator& operator--() {
+    --index;
+    return self();
+  }
+  Iterator operator++(int) {
```
kripken, line 64: What is the parameter here, do you know? It existed before this PR, but just curious.

src/wasm-type.h
```diff
@@ -270,70 +271,24 @@ class Type {
...
+  Iterator end() const { return Iterator{{this, size()}}; }
   std::reverse_iterator<Iterator> rbegin() const {
     return std::make_reverse_iterator(end());
   }
   std::reverse_iterator<Iterator> rend() const {
     return std::make_reverse_iterator(begin());
   }
-  size_t size() const { return end() - begin(); }
-  const Type& operator[](size_t i) const;
+  const Type& operator[](size_t i) const { return *Iterator{{this, i}}; }
```
kripken, line 291: This change seems like it might be slower now, perhaps worth benchmarking?

kripken (approved):

src/support/parent_index_iterator.h
tlively: The parameter is an unused marker to differentiate the signature for the post-increment operator from the the signature for the pre-increment operator. Without it, there'd be no way to tell the two operator overloads apart.

src/wasm-type.h
tlively: My unscientific benchmark running `-O2` on a poppler binary had the new version running faster, but I assume that's just noise. I think observing a difference here would take a microbenchmark, but even then it shouldn't have an appreciable impact on real-world performance.

Issue #4462, opened 2022-01-19
Closed-world assumption and GC


kripken:
The GlobalTypeOptimization pass will remove fields from GC types when they are never used. @askeksa-google noticed that this can break code where JS reads a struct from the outside - the field indexes have changed.

We should probably have an option for this. Should we have a single "closed-world" option perhaps? That could also cover other things like assuming the table is not modified from the outside. Or should we just add a GC-specific option to assume no struct accesses from outside? ("fixed struct ABI")

Alternatively, we could scan the module and see if any GC objects actually escape. But I suspect that will be common, e.g. if the program holds on to JS strings or a canvas context or whatever.

Even more alternatively, if the JS API used names instead of indexes then the initial issue here would have been avoided, but that's probably a larger topic... but cc @jakobkummerow @tlively if there have been recent thoughts there?

tlively:
I've been assuming we would eventually do the escape analysis to figure out what types we can safely modify, but I agree that lots of types will probably escape. A closed-world option would make sense, especially if we could also either emit some useful mapping (or even just JS?) to help users translate types to their expected ABI or if we could insert fixups ourselves at the module boundary.

tlively:
We now have a closed-world flag, and new ideas in this are are tracked in #6965.

PR #4472, opened 2022-01-21
Add --closed-world flag, default to off

kripken:
The GlobalTypeOptimization pass changes observable struct object layout, which
is not safe to do if we are not in a closed world. This adds a flag for that. The flag
may cover more things later, but perhaps we will want more specific flags as well,
see https://github.com/WebAssembly/binaryen/issues/4462

This will regress j2wasm until they enable this flag, as the default is now the
conservative one of not performing this optimization.

This should unbreak dart2wasm, cc @askeksa-google


test/lit/passes/gto_and_cfp_in_O.wast
```diff
@@ -1,18 +1,38 @@
...
   ))
 
+  ;; OPENED:      (export "main" (func $main))
+
+  ;; OPENED:      (func $by-ref (type $none_=>_none) (; has Stack IR ;)
+  ;; OPENED-NEXT:  (struct.set $struct 1
+  ;; OPENED-NEXT:   (global.get $glob)
+  ;; OPENED-NEXT:   (i32.const 200)
+  ;; OPENED-NEXT:  )
+  ;; OPENED-NEXT: )
```
aheejin, line 35: I wonder why `CLOSED` check lines are generated with this function?

aheejin (approved):

src/tools/optimization-options.h
```diff
@@ -241,6 +241,14 @@ struct OptimizationOptions : public ToolOptions {
            Options::Arguments::Zero,
            [this](Options*, const std::string&) {
              passOptions.zeroFilledMemory = true;
+           })
+      .add("--closed-world",
+           "-cw",
+           "Assume that we can optimize without thinking of anything outside",
```
tlively, line 247:
```suggestion
           "Assume that we can optimize without preserving the external module interface",
```

test/lit/passes/gto_and_cfp_in_O.wast
tlively: Do you mean why there are no `CLOSED` check lines for this function? It's because it is optimized out with the closed world flag and does not exist in that output.

src/pass.h
```diff
@@ -160,6 +160,11 @@ struct PassOptions {
   // creates it and we know it is all zeros right before the active segments are
   // applied.)
   bool zeroFilledMemory = false;
+  // Whether to assume that we can optimize this module regardless of anything
+  // outside - we can consider it as a closed world. When this is enabled the
+  // optimizer will do things like change the shape of Struct types, which can
+  // be observable if a GC object escapes to the outside and is inspected.
+  bool closedWorld = false;
```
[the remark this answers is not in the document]
kripken: Interesting... But *some* export must exist, unless all the wasm does is run its start function, so I'm not sure?

kripken:
Thinking more on this, I'd like to flip the flag. That is, that by default we'd consider the world closed. That is technically not safe, but it is only unsafe in the case of dynamic linking, more or less. That is, by default wasm builds from LLVM do not support dynamic linking, and we save a lot of code size thanks to that. You have to opt in to dynamic linking to get that overhead. I'd like the same to happen in wasm GC, basically - otherwise, if the default is dynamic linking just works then much code on the web will end up bloated...

This might require framing the flag as focusing more on wasm GC, perhaps, since we don't want to break anything in wasm MVP?

tlively:
It's unsafe in the case of dynamic linking and also in the case of trying to use the types from JS, which is probably more common. Whether doing the safe thing ends up bloating code depends on what the code imports and exports. If a module is careful to only use very simple types on its boundary, then there wouldn't be any bloat from doing safe optimizations. For now, I would prefer to be safe by default and encourage producers to produce safely optimizable modules.

Do we know what kinds of types J2CL currently makes externally visible?

kripken:
J2Wasm, do you mean? I don't think they access structs, and I know a JS API is not a high priority there. We could ask to find out more.

I guess we should know more about common use patterns here. If everyone will be using structs from JS, then that code size overhead will just need to be paid, and the wrong default would be annoying.

How about if we add a temporary flag for now, since it doesn't sound like we know what we want here, just to unblock @askeksa-google ? The flag could be literally what we were asked for, to disable the GlobalTypeOptimization pass.

kripken:
Actually, checking the J2Wasm output, no import or export is a GC type. Only MVP types + externref appear. A GC type cannot be passed as an externref (correct?) so I think that means we can see statically that there is no problem here. If so, maybe we don't even need a flag?

tlively:
> Actually, checking the J2Wasm output, no import or export is a GC type. Only MVP types + externref appear. A GC type cannot be passed as an externref (correct?) so I think that means we can see statically that there is no problem here. If so, maybe we don't even need a flag?

Oh nice, if they're not exposing complex GC types in their imports and exports, then there shouldn't be a problem. A temporary flag to unblock @askeksa-google until we can implement a module escape analysis in the global type optimizations sounds good.

kripken:
Thanks @askeksa-google , that's good to know. Ok, then for now I think let's not worry about this, and close this PR. When we think more about the actual multi-module case we'll want to consider specific fine-grained flags perhaps.

PR #4474, opened 2022-01-25
Make `TypeBuilder::build()` fallible

tlively:
It is possible for type building to fail, for example if the declared nominal
supertypes form a cycle or are structurally invalid. Previously we would report
a fatal error and kill the program from inside `TypeBuilder::build()` in these
situations, but this handles errors at the wrong layer of the code base and is
inconvenient for testing the error cases.

In preparation for testing the new error cases introduced by isorecursive
typing, make type building fallible and add new tests for existing error cases.
Also fix supertype cycle detection, which it turns out did not work correctly.

### DIFF
--- a/src/ir/type-updating.cpp
+++ b/src/ir/type-updating.cpp
@@ -75,8 +75,7 @@ void GlobalTypeRewriter::update() {
   }
 
   auto buildResults = typeBuilder.build();
-  assert(std::get_if<std::vector<HeapType>>(&buildResults));
-  auto newTypes = *std::get_if<std::vector<HeapType>>(&buildResults);
+  auto& newTypes = *buildResults;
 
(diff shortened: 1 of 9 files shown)


src/ir/type-updating.cpp
```diff
@@ -74,7 +74,9 @@ void GlobalTypeRewriter::update() {
     }
   }
 
-  auto newTypes = typeBuilder.build();
+  auto buildResults = typeBuilder.build();
+  assert(std::get_if<std::vector<HeapType>>(&buildResults));
+  auto newTypes = *std::get_if<std::vector<HeapType>>(&buildResults);
```
kripken, line 79:
```suggestion
  auto& newTypes = *buildResults;
```

kripken (approved):
lgtm % comment

PR #4475, opened 2022-01-25
Isorecursive type validation

tlively:
When building isorecursive types, validate their relationships according to the
rules described in https://github.com/WebAssembly/gc/pull/243. Specifically,
supertypes must be declared before their subtypes to statically prevent cycles
and child types must be declared either before or in the same recursion group as
their parents.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -3238,13 +3238,11 @@ std::optional<TypeBuilder::Error> canonicalizeIsorecursive(
     for (size_t index = groupStart; index < groupEnd; ++index) {
       HeapType type = state.results[index];
       for (HeapType child : type.getHeapTypeChildren()) {
-        // Basic children, globally canonical children, and children defined in
-        // this or previous recursion groups are allowed.
-        if (!isTemp(child) || indexOfType.count(child)) {
-          continue;
+        // Only basic children, globally canonical children, and children
+        // defined in this or previous recursion groups are allowed.
+        if (isTemp(child) && !indexOfType.count(child)) {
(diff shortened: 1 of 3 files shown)


src/wasm/wasm-type.cpp
```diff
@@ -3112,6 +3116,50 @@ void canonicalizeEquirecursive(CanonicalizationState& state) {
...
+        if (!typer.isSubType(sub->struct_, super->struct_)) {
+          return fail();
+        }
+        break;
+      case HeapTypeInfo::ArrayKind:
+        if (!typer.isSubType(sub->array, super->array)) {
+          return fail();
+        }
+        break;
+    }
```
tlively, line 3158: This is moved unchanged from below.

src/wasm/wasm-type.cpp
```diff
@@ -3212,10 +3224,59 @@ std::optional<TypeBuilder::Error> canonicalizeIsorecursive(
...
+  // Validate the children of all types in a recursion group after all the types
+  // have been registered in `indexOfType`.
+  auto finishGroup = [&](size_t groupEnd) -> std::optional<TypeBuilder::Error> {
+    for (size_t index = groupStart; index < groupEnd; ++index) {
+      HeapType type = state.results[index];
+      for (HeapType child : type.getHeapTypeChildren()) {
+        // Basic children, globally canonical children, and children defined in
+        // this or previous recursion groups are allowed.
+        if (!isTemp(child) || indexOfType.count(child)) {
+          continue;
```
kripken, line 3244: How about flipping this condition? That is, "if error, error". Then we don't need the continue and the code would be slightly simpler.

kripken (approved):

PR #4476, opened 2022-01-25
[Wasm GC] Skip GlobalTypeOptimization if it is not safe

kripken:
It is not safe to run it if a struct might escape to the outside, in which case someone
could notice if we altered or removed a field.

Fixes #4472 - I realized it's easy enough to just do this, and avoid a temporary flag.


kripken:
@askeksa-google I believe this should be enough for Dart? (that is, GTO should be skipped because Dart is escaping a struct type in one of the ways checked here)

kripken:
CI error here looks unrelated,
```
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package clang-format
E: Unable to locate package clang-tidy
Error: Process completed with exit code 100.
```
Anyone seen that before?

src/passes/GlobalTypeOptimization.cpp
```diff
@@ -114,11 +115,72 @@ struct GlobalTypeOptimization : public Pass {
...
 
+    // Removing fields or even changing their mutability can be noticeable, if
+    // struct types escape to the outside.
+    auto mightBeStruct = [](Type type) {
+      if (!type.isRef()) {
+        return false;
+      }
+
+      return type.isStruct() ||
+             HeapType::isSubType(HeapType::data, type.getHeapType());
```
tlively, line 131: Is the idea here that `any`, `eq`, or `data` might be passed across the boundary and downcasted on the other side, which would be able to observe changes? I hadn't considered this possibility and it makes things potentially much worse than if we only had to track exact types crossing the boundary.

I had been hoping we could be more precise and still optimize any struct that does not escape, but if exposing a single `any` or `data` reference locks down every struct type, then that might not be very useful.

I wonder if it would be useful to have a flag to assume that the other side won't downcast values we pass it? I would even be ok having that flag on by default.

Either way, we'll need a similar analysis for when we change signature types in the pass that does that.

tlively:
I'm also seeing these same CI failures on https://github.com/WebAssembly/binaryen/pull/4475.

src/passes/GlobalTypeOptimization.cpp
kripken: I think this is an issue even without a downcast in the case of JS. If a struct can be placed in such a position, JS will simply receive the object, the type does not affect anything there.

Meanwhile, I see that I was actually wrong about j2wasm. One of the function *does* actually escape a struct. The name of the function, "Math.max", confused me... I'll ask them. If this is not a glitch, then this PR won't actually help :cry:

tlively (commented):
This PR also doesn't handle the case where a struct escapes indirectly because it's reachable from an array that escapes or something like that.

kripken:
Ah, good point about Array's containing references...

src/passes/GlobalTypeOptimization.cpp
tlively: > I think this is an issue even without a downcast in the case of JS. If a struct can be placed in such a position, JS will simply receive the object, the type does not affect anything there.

Do we have a JS API spec to verify this against? Even if the JS does have visibility to the additional fields, depending on their existence (the equivalent of downcasting) would imply an unspoken contract with the module much the same way downcasting a type passed out from the module would imply an unspoken contract.

kripken: I haven't followed the JS API side. Are they considering adding downcasting?

My guess here was just based on this: if I pass a JS object - of some interesting JS class - into an externref in wasm, then get it back out as an externref, I just get the object - the type doesn't matter. It's not wrapped in any way. If it were wrapped then JS would need a way to downcast on JS types, which it doesn't have. I guess in theory we could do something different from wasm types in JS, though intuitively that feels a little odd.

tlively: Sure, I would expect `JS -> externref -> JS` to work like that, but I'm not sure `struct(i32, i64) -> struct(i32) -> JS` would necessarily expose the `i64` field to JS or that `struct(i32, i64) -> data -> JS` would expose anything about the struct. But this is pure conjecture because we don't have a JS API spec.

I guess JS that reads functions out of the table and calls them already depends on being able to (implicitly) downcast from `funcref` to a specific signature, so maybe just saying "JS doesn't downcast" isn't nuanced enough :/

kripken: Ah, yes, funcrefs are a good existing example. That's the behavior that seems natural to me, too - downcasts are an odd concept to consider for JS.

askeksa-google: AFAIK, V8 currently does not even support typing parameters/returns of imports and exports with custom reference types (if they are to be transferred to/from JS values), so in dart2wasm, these all use either `eqref` or `dataref`.

The interop feature I currently use is the ability to access Wasm struct fields as `.$field0`, `.$field1`, etc. This is just a temporary solution until we have proper interop. The interference with field elimination in Binaryen could be an argument not to retain this feature in V8.

kripken:
Closing as this does not block dart, and we will leave this topic for later when we consider the multiple-module case with wasm GC more carefully. We may want some form of flags there anyhow, and so might not need this check in GTO.

PR #4477, opened 2022-01-26
[NFC] Templatize/generalize RuntimeExpressionRunner

kripken:
Add a base class for it, that is templated and can be extended in general
ways, and make `callFunction` templated on the runner to use as well.

This allows the interpreter's behavior to be customized in a way that we
couldn't so far. wasm-ctor-eval wants to use a special Runner when it
evals a function, so that it can track certain operations, which this will
enable.

### DIFF
--- a/src/wasm-interpreter.h
+++ b/src/wasm-interpreter.h
@@ -3560,16 +3560,18 @@ public:
   };
 
   // Call a function, starting an invocation.
+  template<typename Runner=RuntimeExpressionRunner>
   Literals callFunction(Name name, const Literals& arguments) {
     // if the last call ended in a jump up the stack, it might have left stuff
     // for us to clean up here
     callDepth = 0;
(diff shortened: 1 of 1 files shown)


tlively (approved):
LGTM, but I hope we can clean up all the interpreter stuff at some point. Even before this change, I've found it difficult to keep track of all the templates and interfaces involved in the interpreter code.

kripken:
Agreed, that would be good to do. I did a quick attempt locally actually today, trying to extract this class from ModuleInstance where it is nested, but even that was not easy - the instance info is used in the runner, which I guess it has to be. But perhaps we can find a simpler design for all this eventually...

This is btw an example of something that I think would be simpler in a less-typed language... I think in Go it could "just work" with interfaces given how they are duck-typed, with no CRTP or nesting or anything.

PR #4479, opened 2022-01-26
[NFC] Refactor ModuleInstanceBase+RuntimeExpressionRunner into a single class

kripken:
As recently discussed, the interpreter code is way too complex. Trying to add
ctor-eval stuff I need, I got stuck and ended up spending some time to get rid
of some of the complexity.

We had a ModuleInstanceBase class which was basically an instance of a
module, that is, an execution of it. And internally we have RuntimeExpressionRunner
which is a runner that integrates with the ModuleInstanceBase - basically, it uses
the runtime info to execute code. For example, the MIB has globals info, and the
RER would read it from there.

But these two classes are really just one functionality - an execution of a module.
We get rid of some complexity by removing the separation between them, ending
up with a class that can run a module.

One set of problems we avoid is that we can now extend the single class in a
simple way. Before, we would need to extend both - and inform each other of
those changes. That gets "fun" with CRTP which we use everywhere. In other
words, each of the two classes depended on the other / would need to be
templated on the other. Specifically, `MIB.callFunction` would need to be given
the RER to run with, and so that would need to be templated on it. This ends up
leading to a bunch more templating all around - all complexity that we just
don't need. See the simplification to the wasm-ctor-eval for some of that (and
even worse complexity would have been needed without this PR in the next
steps for that tool to eval GC stuff).

The final single class could probably be called ModuleRunner or such. For now
I left the external name the same to keep this PR small.

Note: Diff without whitespace is far, far smaller.


kripken:
Added a commit from an issue the fuzzer found. It was a pre-existing issue, but this PR made it noticeable. We had the delegate target on the runner, but it should be tied to a function scope. This happened to not be a problem if one always created a new runner for each scope, but this PR makes the runner longer-lived, so the stale data ended up mattering. The commit moves that data to the proper place.

src/wasm-interpreter.h
```diff
@@ -2680,899 +2681,875 @@ template<typename GlobalManager, typename SubType> class ModuleInstanceBase {
...
-        Export* globalExport = inst->wasm.getExport(global->base);
-        global = inst->wasm.getGlobal(globalExport->value);
-      }
+  // Stack of <caught exception, caught catch's try label>
+  SmallVector<std::pair<WasmException, Name>, 4> exceptionStack;
+
+protected:
+  // Returns the instance that defines the memory used by this one.
+  SubType* getMemoryInstance() {
+    auto* inst = self();
```
tlively, line 2705: Probably no need for a separate `inst` variable here and below.

tlively (approved):
To summarize my understanding of the changes:

 - `ModuleInstanceBase` is now an `ExpressionRunner` rather than creating a `RuntimeExpressionRunner` parameterized with itself.
 - `RuntimeExpressionRunner` is folded into the normal `ExpressionRunner`

kripken:
This part is not accurate:

> RuntimeExpressionRunner is folded into the normal ExpressionRunner

`RuntimeExpressionRunner` is now folded into `ModuleInstanceBase`. (As mentioned above, it should probably be renamed to `ModuleRunnerBase` - I can do that in this PR if we want, or later.)

src/wasm-interpreter.h
kripken: `inst` is modified (`inst = ..`) in the loop. We are basically starting at this instance, and then looking through the imports to find the proper instance that defines the memory (which may be this one, or another).

tlively:
Oh, that makes sense. In such a large file it's hard to keep track of where classes begin and end 😬. Maybe as a follow-up we could move the function bodies out of line with the declarations. LGTM modulo that other comment regardless!

kripken:
I realized from your comment that this refactoring does make things less clear with keeping the same name. I pushed a renaming to ModuleRunner, and added a proper comment that I hope also explains the difference between ModuleRunner and ExpressionRunner (one runs an expression by itself, the other an entire module - the class names now are almost self-explanatory).

kripken:
Btw, about moving out of line: sounds good to try, but I'm not sure it's possible with a templated class like this.

tlively:
It is possible, but the syntax is non-obvious. This example works:

```C++
template<int I> struct Foo {
  int foo();
};

template<int I> int Foo<I>::foo() {
  return I;
}

int main() {
  return Foo<0>{}.foo();
}
```

src/wasm-interpreter.h
kripken: I think this is the only open question on this PR - any remaining concerns after my response?

tlively: ah, thanks. Makes sense.

tlively (approved):
LGTM!

aheejin:
> As recently discussed, the interpreter code is way too complex.

Sorry for catching up late! Can you point out where the discussion took place?

kripken:
Just a comment or two, but it was here: https://github.com/WebAssembly/binaryen/pull/4477#pullrequestreview-864052938

PR #4481, opened 2022-01-27
Isorecursive canonicalization

tlively:
Isorecursive canonicalization is similar to equirecursive canonicalization in
that it deduplicates types based on their structure, but unlike equirecursive
canonicalization, isorecursive canonicalization does not need to minimize the
type definition first. Another difference is that structures are deduplicated at
the level of recursion groups rather than individual types, so we cannot reuse
the existing `FiniteShapeHasher` and `FiniteShapeEquator` utilities. Instead,
introduce a new `RecGroupStructure` wrapper that provides equality comparison
and hashing very similar to the former utilities.

Another feature of the isorecursive type system is that its constraints on the
order of recursion groups means that they are already topologically sorted and
can be incrementally canonicalized in a bottom-up manner. This incremental
canonicalization means that the `RecGroupStructure` utility can assume that all
child `HeapTypes` have already been canonicalized and can avoid traversing
anything but the top-level type definitions in the wrapped recursion group. The
only exception is self-references into the wrapped recursion group itself, which
may not be canonical. That special case is detected and handled without any
nontrivial extra work.

Overall, the canonicalization algorithm traverses each type definition once to
canonicalize its `HeapType` use sites, then once to hash its recursion group
structure, then finally once to canonicalize its `Type` use sites in
`globallyCanonicalize`, giving only linear time complexity.


src/wasm/wasm-type.cpp
```diff
@@ -772,15 +827,45 @@ struct SignatureTypeCache {
 static SignatureTypeCache nominalSignatureCache;
 
 // Keep track of the constructed recursion groups.
-static std::mutex recGroupsMutex;
-static std::vector<std::unique_ptr<RecGroupInfo>> recGroups;
+struct RecGroupStore {
+  std::mutex mutex;
+  std::unordered_set<RecGroupStructure> canonicalGroups;
+  std::vector<std::unique_ptr<RecGroupInfo>> builtGroups;
```
kripken, line 833: I'm not sure how to read this. What are "canonical groups" and what are "built groups"?

kripken:
Btw, what is the speed of loading the j2wasm output in nominal mode vs the new isorecursive mode (1 rec group), if we can measure that now?

src/wasm/wasm-type.cpp
tlively: I'll add comments. `canonicalGroups` is used to do the actual hashing and deduplication of groups, including singleton groups that do not have associated `RecGroupInfo`s, while `builtGroups` exists just to hold the unique pointers keeping the `RecGroupInfo`s alive.

tlively:
> Btw, what is the speed of loading the j2wasm output in nominal mode vs the new isorecursive mode (1 rec group), if we can measure that now?

We can't quite measure it as-is because the nominal output does not respect the constraint that supertypes must come before their subtypes. I'll try hacking the nominal output to make it work locally; we need that logic to get proper isorecursive output anyways.

src/wasm/wasm-type.cpp
```diff
@@ -2203,6 +2291,258 @@ bool FiniteShapeEquator::eq(const Rtt& a, const Rtt& b) {
...
+  }
+  if (a.isBasic() || b.isBasic()) {
+    return false;
+  }
+  return eq(*getTypeInfo(a), *getTypeInfo(b), other);
+}
+
+bool RecGroupStructure::eq(HeapType a, HeapType b, RecGroup other) const {
+  // Do not recurse into the structure of children `a` and `b`, but check
+  // whether their recursion groups and indices match. Since the wrapped group
```
kripken, line 2332: To make sure I follow, the "wrapped group" here refers to the group the RecGroupStructure itself wraps?

src/wasm/wasm-type.cpp
```diff
@@ -284,6 +284,60 @@ struct FiniteShapeEquator {
...
+// groups reachable from this one have been canonicalized, except for the
+// wrapped group itself.
+struct RecGroupStructure {
+  RecGroup group;
+
+  bool operator==(const RecGroupStructure& other) const;
+  size_t hash() const;
+
+private:
+  bool topLevelEq(HeapType a, HeapType b, RecGroup other) const;
```
kripken, line 299: How should I interpret "other" in all of these? How does it relate to a, b, and the group in this wrapper?

src/wasm/wasm-type.cpp
tlively: Yes, correct.

src/wasm/wasm-type.cpp
tlively: `other` needs to be piped down through all of these intermediate functions to reach the leaf `eq` function that compares `HeapType`s without further recursion. That function uses it to detect self-references in that other rec group so they can compare equal to self-references in the wrapped rec group.

kripken: Sorry, but I must be still missing something obvious here. The name "topLevelEq" suggests to me that it compares a and b at the top level. But what is "other" here? The group of a? Or of b? Is a in this group, and b in other, maybe?

tlively: Ultimately we are comparing the wrapped rec group and the `other` rec group. In all the `eq` helper functions, `a` comes from the wrapped group and `b` comes from `other`.

kripken: I see, thanks. Maybe mentioning that in a comment would be helpful.

kripken: Or, `bGroup` might clarify that it is the group of b? not sure that's great though.

src/wasm/wasm-type.cpp
```diff
@@ -2203,6 +2291,258 @@ bool FiniteShapeEquator::eq(const Rtt& a, const Rtt& b) {
...
+  // Do not recurse into the structure of children `a` and `b`, but check
+  // whether their recursion groups and indices match. Since the wrapped group
+  // may not be canonicalized, explicitly check whether `a` and `b` are in the
+  // respective recursion groups of the respective top-level groups we are
+  // comparing, in which case the structure is still equivalent.
+  if (a.getRecGroupIndex() != b.getRecGroupIndex()) {
+    return false;
+  }
+  auto groupA = a.getRecGroup();
+  auto groupB = b.getRecGroup();
```
kripken, line 2340: If `other` is the group of B, why do we need this line? The next line even compares the two. So is `other` not simply the group of `b`, piped down into all these `eq()` functions - I thought that's what you said before?

tlively: `other` is the rec group that `b` is being used in, not necessarily the rec group that `b` belongs to.

kripken: I think this might be the core of my misunderstanding then. So this method is called where a is used in the wrapped group, and b is used in the `other` group, but they both might belong to other groups? Or is that true only for b and not a?

Maybe adding concrete examples of rec groups in a comment might help. And/or, a comment that explains when these eq methods are used - what are the presuppositions about groups and canonicalization state and so forth are (e.g., "b must belong to .., but a is still being"). This may be super-obvious to you, sorry about that! But the isorecursive semantics are odd enough for me that I can't just get it from reading this code I guess.

kripken: Also, perhaps `other => bUser` to clarify it is the group that uses `b`?

src/wasm/wasm-type.cpp
```diff
@@ -284,6 +284,63 @@ struct FiniteShapeEquator {
   bool eq(const Rtt& a, const Rtt& b);
 };
 
+// A wrapper around a RecGroup that provides equality and hashing based on the
+// structure of the group such that isorecursively equivalent recursion groups
+// will compare equal and will have the same hash. Assumes that all recursion
+// groups reachable from this one have been canonicalized, except for the
+// wrapped group itself.
+struct RecGroupStructure {
```
kripken, line 292: Thinking on this, what do you think of the following design: Make this a class `RecGroupComparer(existing, new)` which represents a series of comparisons between two rec groups, one existing (and canonicalized) and one new (and in the process of being compared for purposes of canonicalization). This is similar to the current class except that `other` is passed in the constructor, and so it is not passed in any of the methods. I think this might be easier to understand, at least for me, and would be slightly simpler code as well. The move from "wrapping a rec group" to "comparing two rec groups" also makes things clearer since "wrapper" is less obvious I think.

tlively:
I reorganized things so that hashing and equality comparison are handled by separate utilities more similar to `FiniteShapeHasher` and `FiniteShapeEquator`. PTAL!

src/wasm/wasm-type.cpp
```diff
@@ -2203,6 +2313,239 @@ bool FiniteShapeEquator::eq(const Rtt& a, const Rtt& b) {
...
+size_t RecGroupHasher::innerHash(HeapType type) const {
+  // Do not recurse into the structure of this child type, but rather hash it as
+  // an index into a rec group. Only take the rec group identity into account if
+  // the child is not a member of the top-level group because in that case the
+  // group may not be canonicalized yet.
+  size_t digest = wasm::hash(type.getRecGroupIndex());
+  auto currGroup = type.getRecGroup();
+  if (currGroup != group) {
+    wasm::rehash(digest, currGroup.getID());
+  }
```
kripken, line 2343: I think it would be good to add an else here with a rehash of something that cannot overlap with the getID in the if. Without that, if this hash is combined with other things then it can compose in ways that happen to overlap.

src/wasm/wasm-type.cpp
```diff
@@ -284,6 +284,82 @@ struct FiniteShapeEquator {
...
+  RecGroup group;
+
+  RecGroupHasher(RecGroup group) : group(group) {}
+
+  // Perform the hash.
+  size_t operator()() const;
+
+  size_t topLevelHash(HeapType type) const;
+  size_t innerHash(HeapType type) const;
+  size_t hash(Type type) const;
```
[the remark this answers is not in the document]
tlively: That `hash` is the helper for hashing `Type` as opposed to `HeapType`. Like `Tuple`, `Struct`, `Field`, and `Array`, `Type` is neither the top-level item being hashed nor the bottom-level item being hashed, but rather is some intermediate structure.

src/wasm/wasm-type.cpp
```diff
@@ -2203,6 +2313,239 @@ bool FiniteShapeEquator::eq(const Rtt& a, const Rtt& b) {
...
+
+size_t RecGroupHasher::hash(const TypeInfo& info) const {
+  size_t digest = wasm::hash(info.kind);
+  switch (info.kind) {
+    case TypeInfo::TupleKind:
+      hash_combine(digest, hash(info.tuple));
+      return digest;
+    case TypeInfo::RefKind:
+      rehash(digest, info.ref.nullable);
+      hash_combine(digest, innerHash(info.ref.heapType));
```
kripken, line 2365: Reading this, I am not clear on when `hash` is called vs `innerHash`, which I guess is the same question as above.

src/wasm/wasm-type.cpp
```diff
@@ -2203,6 +2313,239 @@ bool FiniteShapeEquator::eq(const Rtt& a, const Rtt& b) {
...
+  // whether their recursion groups and indices match. Since the wrapped group
+  // may not be canonicalized, explicitly check whether `a` and `b` are in the
+  // respective recursion groups of the respective top-level groups we are
+  // comparing, in which case the structure is still equivalent.
+  if (a.getRecGroupIndex() != b.getRecGroupIndex()) {
+    return false;
+  }
+  auto groupA = a.getRecGroup();
+  auto groupB = b.getRecGroup();
+  return groupA == groupB || (groupA == newGroup && groupB == otherGroup);
```
kripken, line 2467: This is the "magic" line you mentioned offline, I think? Where we accept them as equal even if we don't know the entire group matches yet, and then at the end if nothing mismatches then the entire groups canonicalize so it all works out. I think it would be good to add a comment here about that.

src/wasm/wasm-type.cpp
```diff
@@ -3306,13 +3653,39 @@ std::optional<TypeBuilder::Error> canonicalizeIsorecursive(
...
 
-  // Move the recursion groups into the global store. TODO: after proper
-  // isorecursive canonicalization, some groups may no longer be used, so they
-  // will need to be filtered out.
-  std::lock_guard<std::mutex> lock(recGroupsMutex);
-  for (auto& info : recGroupInfos) {
-    recGroups.emplace_back(std::move(info));
+  // Now that we know everything is valid, start canonicalizing recursion
+  // groups. Before canonicalizing each group, update all the use sites with it
+  // to make sure it only refers to other canonical groups or to itself. To
```
kripken, line 3658: How could the uses point to anything else? We have already validated that they do not refer to any later rec group, I think - or am I wrong?

src/wasm/wasm-type.cpp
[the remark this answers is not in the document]
kripken: I think I see. Then when calling `hash` on a type, that should call `innerHash` on the HeapType in the type (if there is one), is that right?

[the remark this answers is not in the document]
tlively: Yep, that's right.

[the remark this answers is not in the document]
kripken: Thanks, I think I get it. I think comments mentioning what you said in this sub-thread would be good to have in the source - would have saved me time.

kripken:
Ok, I finished a full pass now. Overall shape and logic looks good. I think I get almost everything, so lgtm % my open questions.

src/wasm/wasm-type.cpp
kripken: Specifically, it is surprising to see hash on one and innerHash on another of the switch cases. I'd expect them all to be the same. I guess the issue is that HeapTypes get special treatment. But I wonder if `hash=>innerHash` might be less surprising when reading?

kripken: Or, innerHash=hash could make sense too, really, with a comment.

src/wasm/wasm-type.cpp
tlively: I don't believe there's any way such a composition could exist (modulo weaknesses in the hash function that could cause collisions in any other circumstance). In other words, not having a rehash here is just as good as adding a rehash of some arbitrary constant I could come up with.

src/wasm/wasm-type.cpp
tlively: `hash` is for some intermediate structure (which is never a `HeapType`) and `innerHash` is for `HeapType`s. If this is causing confusion, I could also rename `innerHash` back to `hash`.

src/wasm/wasm-type.cpp
tlively: Yes, this is magic line, but the big comment above already describes what is happening. (Separately, I notice that comment is stale since the refactoring, so I'll fix that.)

src/wasm/wasm-type.cpp
tlively: If recursion group `B` refers to recursion group `A`, then yes, we must have already canonicalized `A`. But perhaps `A` was equivalent to some previously canonicalized recursion group `A'`. In that case, we would have thrown out `A` and we need to replace all references to `A` with equivalent references to `A'`.

src/wasm/wasm-type.cpp
tlively: Oh, I hadn't seen the last two comments when I wrote this. I'll rename `innerX` back to just `X`.

src/wasm/wasm-type.cpp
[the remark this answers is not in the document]
tlively: Ended up renaming `innerHash` to just `hash`, so now the only special method is the "topLevel" one. I'll add comments about why those exist.

tlively:
All comments now addressed.

src/wasm/wasm-type.cpp
kripken: If you have a struct with 3 child hash types, then we'll hash them in sequence. Without ensuring the same "structure" of hashing operations, that might end up overlapping with a struct with say 4 children, if the one with 3 does an extra rehashing for one of its children.

Is `getID()` a pointer? Then rehashing with a 0 would be a value that cannot collide.

Here are places where we do similar things:

https://github.com/WebAssembly/binaryen/blob/f2d53918155cfe6e7fc333be22f25ae3776b1438/src/ir/ExpressionAnalyzer.cpp#L275-L277

https://github.com/WebAssembly/binaryen/blob/f2d53918155cfe6e7fc333be22f25ae3776b1438/src/ir/ExpressionAnalyzer.cpp#L343-L346

Or are you saying there is a logical reason stemming from the isorecursive structure that prevents the problem here? I may be just missing that then...

src/wasm/wasm-type.cpp
kripken: Oh, wow, I didn't even think about that. Yes, makes sense.

Maybe worth clarifying in the comment? Might be just me though...

src/wasm/wasm-type.cpp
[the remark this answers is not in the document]
kripken: Yes, this looks good now, just special-casing `topLevel` feels right.

src/wasm/wasm-type.cpp
```diff
@@ -3306,13 +3659,39 @@ std::optional<TypeBuilder::Error> canonicalizeIsorecursive(
...
```
kripken, line 3664:
```suggestion
  // to make sure it only refers to other canonical groups or to itself. That
  // is, it may have references to groups that have been canonicalized
  // with something before them; we want to only refer to the canonical
  // group for all things equal to it. To
```

src/wasm/wasm-type.cpp
```diff
@@ -3306,13 +3659,39 @@ std::optional<TypeBuilder::Error> canonicalizeIsorecursive(
...
+    groupStart = 0;
+    for (auto group : groups) {
+      size_t size = group.size();
+      for (size_t i = 0; i < size; ++i) {
+        state.updateUses(replacements, state.newInfos[groupStart + i]);
+      }
+      groupStart += size;
+      RecGroup canonical(0);
+      if (auto it = groupInfoMap.find(group); it != groupInfoMap.end()) {
+        // Move ownership to the global store if this group is canonical.
```
kripken, line 3680: I don't get this. `groupInfoMap` is not a place where presence indicates being canonical or not, does it? The entire canonicalization aspect is handled a few lines down where you do `if (group != canonical)`, I think. Then why does this line mention "canonical"? I am also puzzled by the word. Does "is canonical" mean "is the canonical one" or "is canonicalized to something else"?

src/wasm/wasm-type.cpp
tlively: In the locations you linked to, it is possible that the code path is reached _n_ times in a row and without hashing in a constant, different values of _n_ would result in the same hash. In this case, the hashing of the index above this branch ensures that calling this method _n_ times returns different results for different values of _n_. Given that, not modifying the hash is just as good as hashing in a constant.

src/wasm/wasm-type.cpp
tlively: Sure, I'll call this out more specifically in the comment.

tlively: done.

src/wasm/wasm-type.cpp
kripken: What you say is right, but it is an additional issue to the one I mentioned. Maintaining the "structural form" of hashing has been noticed to help in practice. I mean, we don't want these to hash the same:
```
[1 2 3 4]
[1 2 3] [4]
```
The second line has a shorter logical first item and a second short item that happens to contain 4. The same 4 numbers are hashed in each case, but logically the result should differ. This is solved by ensuring the structure to keep the right size:
```
[1 2 3 4]
[1 2 3 0] [4]
```
Here `0` is a value we know cannot overlap with other things possible in that position.

src/wasm/wasm-type.cpp
tlively: I added some different text to this comment before I saw your suggestion. Let me know if it makes sense!

kripken: Your text lgtm!

src/wasm/wasm-type.cpp
tlively: > Does "is canonical" mean "is the canonical one" or "is canonicalized to something else"?

The former.

`globalRecGroupStore.insert(std::unique_ptr<RecGroupInfo>&&)`, called on the line after this comment, moves the info into the store if and only if it is the info for the first rec group of this structure to be inserted, i.e. if the rec group becomes the canonical rec group for its structure. I'll rename the method `maybeInsert` to make it clearer that it only conditionally takes ownership of its argument.

tlively: But actually this is already consistent with how the `insert` methods of `globalHeapTypeStore` and `globalTypeStore` already work, so maybe it would be better to keep the naming consistent (or to change it everywhere in a follow up).

tlively: I updated the comment to hopefully make it clearer.

kripken: I see, existing naming seems ok then.

I am still having trouble with the logic, though. `auto it = groupInfoMap.find(group); it != groupInfoMap.end()` returns true if we find the item in that map. If we find it, doesn't that mean it is *not* canonical? But the move is in that arm of the if.

tlively: This condition is about whether or not there exists a `RecGroupInfo` for this group, in which case it would live in the `groupInfoMap`. The alternative is that this is a singleton group so there is no `RecGroupInfo`. In either case, the group may or may not end up being the canonical group for its structure.

kripken: I see, thanks. I missed that this was about the singleton issue...  code is quite complex so I'd suggest a comment about that as well, as the current ones just speak about canonicalization here.

tlively: Added a bigger comment about info ownership to replace this comment.

kripken: "If it is canonical" - based on your last comment, shouldn't it be "is not a singleton"?

kripken (approved):
Thanks! Last comment makes sense to me. Sorry for all the back and forth!

tlively:
Thanks for taking the time for a thorough review!

PR #4483, opened 2022-01-27
wasm-emscripten-finalize: Remove legacy --new-pic-abi option

sbc100:


test/lit/help/wasm-fuzz-types.test
```diff
@@ -0,0 +1,29 @@
+;; RUN: wasm-fuzz-types --help | filecheck %s
```
kripken, line 1: Looks like this was added as the auto-updater emits it but it was not checked in? Sgtm.

kripken (approved):

PR #4486, opened 2022-01-28
Interpreter: Remove GlobalManager

kripken:
GlobalManager is another class that added complexity in the interpreter logic,
and did not help. In fact it hurts extensibility, as when one wants to extend the
interpreter one has another class to customize, and it is templated on the main
runner, so again as #4479 we end up with annoying template cycles.

This simply removes that class. That makes the interpreter code strictly
simpler. Applying that change to wasm-ctor-eval also ends up fixing a
pre-existing bug, so this PR gets testing through that.

The ctor-eval issue was that we did not extend the GlobalManager properly
in the past: we checked for accesses on imported globals there, but not in
the main class, i.e., not on `global.get` operations. Needing to do things in
two places is an example of the previous complexity. The fix is simply to
implement `visitGlobalGet` in one place, and remove all the GlobalManager
logic added in ctor-eval, which then gets a lot simpler as well.

The new `imported-global-2.wast` checks for that bug (a `global.get` of an
import should stop us from evalling). Existing tests cover the other cases,
like it being ok to read a non-imported global, etc. The existing test
`indirect-call3.wast` required a slight change: There was a global.get of
an imported global, which was ignored in the place it happened (an init
of an elem segment); the new code checks all global.gets, so it now
catches that.


tlively (approved):
Nice!

PR #4487, opened 2022-02-01
Isorecursive type output ordering

tlively:
Generally we try to order types by decreasing use count so that frequently used
types get smaller indices. For the equirecursive and nominal systems, there are
no contraints on the ordering of types, so we just have to sort them according
to their use counts. For the isorecursive type system, however, there are a
number of ordering constraints that have to be met for the type section to be
valid.

First, types in the same recursion group must be adjacent so they can be grouped
together. Next, groups must be ordered topologically. Finally, supertypes must
appear before their subtypes, even within a single group. Only after satisfying
these constraints can we additionally order by use count.

Update type ordering to produce a valid and optimal ordering. In principle it
would be possible to do less work for use cases that don't require the optimal
ordering, such as in type updating passes, but skipping unnecessary work
wouldn't improve the algorithmic complexity of the sorting, so don't bother for
now.

### DIFF
--- a/src/ir/module-utils.cpp
+++ b/src/ir/module-utils.cpp
@@ -33,6 +33,11 @@ struct Counts : public InsertOrderedMap<HeapType, size_t> {
       note(ht);
     }
   }
+  void include(HeapType type) {
+    if (!type.isBasic()) {
+      (*this)[type];
+    }
+  }
 };
 
 Counts getHeapTypeCounts(Module& wasm) {
@@ -156,46 +161,21 @@ Counts getHeapTypeCounts(Module& wasm) {
         // is in flux, skip counting them to keep the type orderings in nominal
         // test outputs more similar to the orderings in the equirecursive
         // outputs. FIXME
-        counts.note(*super);
+        counts.include(*super);
       }
     }
 
-    // Make sure we've noted the complete recursion group of each type as well.
+    // Make sure we've included the complete recursion group of each type.
     auto recGroup = ht.getRecGroup();
     for (auto type : recGroup) {
       if (!counts.count(type)) {
         newTypes.insert(type);
-        counts.note(type);
+        counts.include(type);
       }
     }
   }
 
   return counts;
-}
-
-void coalesceRecGroups(IndexedHeapTypes& indexedTypes) {
-  if (getTypeSystem() != TypeSystem::Isorecursive) {
-    // No rec groups to coalesce.
-    return;
-  }
-
-  // TODO: Perform a topological sort of the recursion groups to create a valid
-  // ordering rather than this hack that just gets all the types in a group to
-  // be adjacent.
-  assert(indexedTypes.indices.empty());
-  std::unordered_set<HeapType> seen;
-  std::vector<HeapType> grouped;
-  grouped.reserve(indexedTypes.types.size());
-  for (auto type : indexedTypes.types) {
-    if (seen.insert(type).second) {
-      for (auto member : type.getRecGroup()) {
-        grouped.push_back(member);
-        seen.insert(member);
-      }
-    }
-  }
-  assert(grouped.size() == indexedTypes.types.size());
-  indexedTypes.types = grouped;
 }
 
 void setIndices(IndexedHeapTypes& indexedTypes) {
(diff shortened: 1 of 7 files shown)


src/ir/module-utils.cpp
```diff
@@ -33,6 +33,11 @@ struct Counts : public InsertOrderedMap<HeapType, size_t> {
       note(ht);
     }
   }
+  void include(HeapType type) {
```
kripken, line 36: I had to read the body 3 times before I saw the difference is in the lack of `++`, so I think perhaps a comment could help here. "include" does not explain enough by itself, at least for me, the difference with "note".

Or, how about `note => count` and `include => ensure`?

src/ir/module-utils.cpp
```diff
@@ -156,48 +161,23 @@ Counts getHeapTypeCounts(Module& wasm) {
         // is in flux, skip counting them to keep the type orderings in nominal
         // test outputs more similar to the orderings in the equirecursive
         // outputs. FIXME
-        counts.note(*super);
+        counts.include(*super);
```
kripken, line 164: Is the comment above this still valid?

test/lit/isorecursive-output-ordering.wast
```diff
@@ -0,0 +1,97 @@
...
+
+ (rec
+  (type $twig (struct_subtype data))
+  (type $used-a-bit (struct_subtype (ref $leaf) data))
+ )
+
+ (rec
+  (type $root (struct_subtype data))
+  (type $used-a-lot (struct_subtype $twig))
+ )
```
kripken, line 72: The groups here remain in the same order after sorting. How about flipping them in the input, just to show the change? (otherwise, a stable sort might arrive at the same result)

kripken (approved):
Nice, lgtm % comments.

The `N^2` stuff is somewhat worrying... but I guess when you measure it we'll see how urgent it is to fix.

src/ir/module-utils.cpp
```diff
@@ -216,37 +196,153 @@ std::vector<HeapType> collectHeapTypes(Module& wasm) {
...
+    }
   }
 
-  coalesceRecGroups(indexedTypes);
-  setIndices(indexedTypes);
-  return indexedTypes;
-}
+  // Fix up the cumulative counts to be an average instead.
+  for (auto& [group, info] : groupInfos) {
+    info.count /= group.size();
```
kripken, line 262: Why average here?

src/ir/module-utils.cpp
tlively: Unfortunately `count` is already the name of the method we use to determine whether a type is already present in the `Counts`, since it inherits from a map type.

I'll at least add a comment about the difference for now.

src/ir/module-utils.cpp
tlively: Yep, previously we "skipped counting" by only calling `note` once, which actually set the count to 1. Now we set the count to 0, so the comment is more accurate than before.

test/lit/isorecursive-output-ordering.wast
tlively: We can't change their order in the input because then the input wouldn't validate. By using each group with different frequencies, I can at least show that the dependencies prevent reordering based on the counts. I'll add an unrelated group in to show more differences as well.

src/ir/module-utils.cpp
tlively: Reasonable alternatives might be to take the max or min count among group members instead, but average is a better heuristic.

If we took the max count, then we could end up with huge groups with many infrequently used types and a few commonly used types hogging the valuable beginning of the index space.

If we took the min count, then we could end up with groups with many frequently used types and a few infrequently used types appearing very late in the index space, which would be bad for code size.

Using the average means that the cumulative count for the group has to justify the amount of index space the group takes up.

tlively: I'll add a comment in the code.

tlively:
I think the error the MacOS builder ran into also showed up intermittently on a previous PR. Hopefully it goes away for now and shows up again if I add fuzzing support for isorecursive types.

tlively:
Ugh, the binary relation on types implemented in this PR is not a total ordering. Here's a counterexample:

```
(rec
 (type $A (struct_subtype data)) (; count = 0 ;)
 (type $B (struct_subtype $A)) (; count = 2 ;)
 (type $C (struct_subtype data)) (; count = 1 ;)
)
```
`$A < $B` because `$A` is a supertype of `$B`. `$B < $C` because `$B` has a higher count than `$C`. `$C < $A` because `$C` has a greater count than `$A`. But that makes a cycle, so `<` cannot be both transitive and antisymmetric, so it is not a partial order.

src/ir/module-utils.cpp
kripken: The supertype is encoded in the binary, though, isn't it? Why don't we want to bump the count here?

src/ir/module-utils.cpp
kripken: I see, makes sense.

The averaging using integers loses information, though. In fact I wonder if it might be common for the average to end up mostly 0 or 1. Maybe using doubles would make sense? The overhead isn't large, and they are deterministic.

src/ir/module-utils.cpp
tlively: It's what the comment explains; supertypes appear in the binary only in nominal mode but not in equirecursive mode, so don't count them in either mode to keep the outputs similar.

tlively:
I largely rewrote this PR as #4492, which fixes the correctness issue and is more efficient.

PR #4492, opened 2022-02-02
Topological sorting of types in isorecursive output

tlively:
Generally we try to order types by decreasing use count so that frequently used
types get smaller indices. For the equirecursive and nominal systems, there are
no contraints on the ordering of types, so we just have to sort them according
to their use counts. For the isorecursive type system, however, there are a
number of ordering constraints that have to be met for the type section to be
valid. First, types in the same recursion group must be adjacent so they can be
grouped together. Second, groups must be ordered topologically so that they only
refer to types in themselves or prior groups.

Update type ordering to produce a valid isorecursive output by performing a
topological sort on the recursion groups. While performing the sort, prefer to
visit and finish processing the most used groups first as a heuristic to improve
the final ordering.

Do not reorder types within groups, since doing so would change type identity
and could affect the external interface of the module. Leave that reordering to
an optimization pass (not yet implemented) that users can explicitly opt in to.

### DIFF
--- a/src/ir/module-utils.cpp
+++ b/src/ir/module-utils.cpp
@@ -197,7 +197,10 @@ void setIndices(IndexedHeapTypes& indexedTypes) {
 // finished as well and should be popped.
 template<typename T> struct TopologicalSortStack {
   std::list<T> workStack;
+  // Map items to their locations in the work stack so that we can make sure
+  // each appears and is finished only once.
   std::unordered_map<T, typename std::list<T>::iterator> locations;
+  // Remember which items we have finished so we don't visit them again.
   std::unordered_set<T> finished;
 
   bool empty() const { return workStack.empty(); }
--- a/src/ir/module-utils.cpp
+++ b/src/ir/module-utils.cpp
@@ -33,6 +33,12 @@ struct Counts : public InsertOrderedMap<HeapType, size_t> {
       note(ht);
     }
   }
+  // Ensure a type is included without increasing its count.
+  void include(HeapType type) {
+    if (!type.isBasic()) {
+      (*this)[type];
+    }
+  }
 };
 
 Counts getHeapTypeCounts(Module& wasm) {
@@ -128,14 +134,16 @@ Counts getHeapTypeCounts(Module& wasm) {
 
   // Recursively traverse each reference type, which may have a child type that
   // is itself a reference type. This reflects an appearance in the binary
-  // format that is in the type section itself.
-  // As we do this we may find more and more types, as nested children of
-  // previous ones. Each such type will appear in the type section once, so
-  // we just need to visit it once.
+  // format that is in the type section itself. As we do this we may find more
+  // and more types, as nested children of previous ones. Each such type will
+  // appear in the type section once, so we just need to visit it once. Also
+  // track which recursion groups we've already processed to avoid quadratic
+  // behavior when there is a single large group.
   InsertOrderedSet<HeapType> newTypes;
   for (auto& [type, _] : counts) {
     newTypes.insert(type);
   }
+  std::unordered_set<RecGroup> includedGroups;
   while (!newTypes.empty()) {
     auto iter = newTypes.begin();
     auto ht = *iter;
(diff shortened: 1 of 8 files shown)


src/ir/module-utils.cpp
```diff
@@ -156,54 +164,68 @@ Counts getHeapTypeCounts(Module& wasm) {
...
+// A utility data structure for performing topological sorts. The elements to be
+// sorted should all be `push`ed to the stack, then while the stack is not
+// empty, predecessors of the top element should be pushed. On each iteration,
+// if the top of the stack is unchanged after pushing all the predecessors, that
+// means the predecessors have all been finished so the current element is
+// finished as well and should be popped.
+template<typename T> struct TopologicalSortStack {
+  std::list<T> workStack;
+  std::unordered_map<T, typename std::list<T>::iterator> locations;
+  std::unordered_set<T> finished;
```
kripken, line 201: comments on the last two would be good, as their names do not obviously explain themselves (at least to me)

src/ir/module-utils.cpp
```diff
@@ -156,54 +164,68 @@ Counts getHeapTypeCounts(Module& wasm) {
...
+
+  T& peek() { return workStack.back(); }
+
+  void pop() {
+    finished.insert(workStack.back());
+    locations.erase(workStack.back());
+    workStack.pop_back();
+  }
+};
+
```
kripken, line 228: This seems almost like https://github.com/WebAssembly/binaryen/blob/348131b0a383d2946f18923f7acfd963089f4f5d/src/support/unique_deferring_queue.h#L70 except that it does not defer, is that right?

src/wasm/wasm-binary.cpp
```diff
@@ -1847,6 +1847,7 @@ void WasmBinaryBuilder::readTypes() {
   size_t numTypes = getU32LEB();
   BYN_TRACE("num: " << numTypes << std::endl);
   TypeBuilder builder(numTypes);
+  builder.createRecGroup(0, numTypes);
```
kripken, line 1850: What is this?

tlively: Oh, oops, this is one my local hacks to get measurements. Will remove.

src/ir/module-utils.cpp
```diff
@@ -216,37 +238,117 @@ std::vector<HeapType> collectHeapTypes(Module& wasm) {
...
+    info.useCount /= group.size();
+  }
 
-IndexedHeapTypes getOptimizedIndexedHeapTypes(Module& wasm) {
-  Counts counts = getHeapTypeCounts(wasm);
+  // Sort the preds of each group by increasing use count so the topological
+  // sort visits the most used first. Break ties with the group's appearance
+  // index to ensure determinism.
+  auto sortGroups = [&](std::vector<RecGroup>& groups) {
+    std::sort(groups.begin(), groups.end(), [&](auto& a, auto& b) {
```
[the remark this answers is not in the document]
tlively: No need, because no two groups will compare equal due to their different assigned indices.

src/ir/module-utils.cpp
kripken: I see, thanks. What do you think about adding a UniqueDeferringStack, then, as a LIFO version? Having the code more similar would be nice too, like using a count instead of `locations`.

src/ir/module-utils.cpp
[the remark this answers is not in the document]
kripken: Ah, there is the index there... I missed that before.

kripken (commented):
Lgtm aside from the refactoring question about the generic helper class. I think something that fits in better with the existing code would be nicer here.

src/ir/module-utils.cpp
tlively: As discussed offline, there are a few options for generalizing and unifying this code that I would like to play around with. Let's land this PR as-is, then I will investigate those options for a follow-up improvement.

tlively:
(automerge enabled)

tlively:
Merged manually since we agreed on a plan for the refactoring offline.

PR #4494, opened 2022-02-02
Isorecursive binary format

tlively:
Write and parse recursion groups in binary type sections. Unlike in the text
format, where we ignore recursion groups when not using isorecursive types, do
not allow parsing binary recursion group when using other type systems. Doing so
would produce incorrect results because recursions groups only count as single
entries in the type system vector so we dynamically grow the TypeBuilder when we
encounter them. That would change the mapping of later indices to types, and
would change the meaning of previous type definitions that use those later
indices. This is not a problem in the isorecursive system because in that system
type definitions are not allowed to use later indices.

### DIFF
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -219,16 +219,33 @@ void WasmBinaryWriter::writeTypes() {
   if (indexedTypes.types.size() == 0) {
     return;
   }
+  // Count the number of recursion groups
+  std::optional<RecGroup> lastGroup;
+  size_t numGroups = 0;
+  for (auto type : indexedTypes.types) {
+    auto currGroup = type.getRecGroup();
+    numGroups += lastGroup != currGroup;
+    lastGroup = currGroup;
+  }
   BYN_TRACE("== writeTypes\n");
   auto start = startSection(BinaryConsts::Section::Type);
-  o << U32LEB(indexedTypes.types.size());
+  o << U32LEB(numGroups);
+  lastGroup = {};
   for (Index i = 0; i < indexedTypes.types.size(); ++i) {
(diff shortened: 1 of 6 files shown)


src/wasm/wasm-binary.cpp
```diff
@@ -219,16 +219,33 @@ void WasmBinaryWriter::writeTypes() {
...
+  size_t numGroups = 0;
+  for (auto type : indexedTypes.types) {
+    auto currGroup = type.getRecGroup();
+    numGroups += lastGroup != currGroup;
+    lastGroup = currGroup;
+  }
   BYN_TRACE("== writeTypes\n");
   auto start = startSection(BinaryConsts::Section::Type);
-  o << U32LEB(indexedTypes.types.size());
+  o << U32LEB(numGroups);
```
kripken, line 232: In the non-hybrid case, this is just the number of types, is that right? (one type per "rec group") If so, it might be good to clarify that.

src/wasm/wasm-binary.cpp
```diff
@@ -219,16 +219,33 @@ void WasmBinaryWriter::writeTypes() {
...
+  lastGroup = {};
   for (Index i = 0; i < indexedTypes.types.size(); ++i) {
     auto type = indexedTypes.types[i];
-    bool nominal = getTypeSystem() == TypeSystem::Nominal;
+    // Check whether we need to start a new recursion group.
+    auto currGroup = type.getRecGroup();
+    if (lastGroup != currGroup && currGroup.size() > 1) {
```
kripken, line 238: ` > 1` is worth a comment. That handles the non-Isorecursive case also, correct?

tlively:
Comments improved!

kripken (approved):

PR #4495, opened 2022-02-02
[NFC] Simplify TopologicalSortStack

tlively:
Using a vector and lazily skipping finished items in `pop` rather than using a
linked list and eagerly removing duplicated items makes the code simpler and
more than twice as fast on my test case. The algorithmic complexity is unchanged
because the work to skip duplicates remains linear in the number of duplicates
added, it's just not spread out over the linear duplicate pushes any more.

### DIFF
--- a/src/ir/module-utils.cpp
+++ b/src/ir/module-utils.cpp
@@ -196,10 +196,7 @@ void setIndices(IndexedHeapTypes& indexedTypes) {
 // means the predecessors have all been finished so the current element is
 // finished as well and should be popped.
 template<typename T> struct TopologicalSortStack {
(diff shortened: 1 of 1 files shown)


tlively:
It turns out we cannot use this to replace `UniqueNonrepeatingDeferredQueue` because its only user, `Heap2Local`, depends on the FIFO ordering of the queue. A quick grep does not turn up any other instances of topological sorting in the code base, although I certainly plan to add at least one for optimizing the ordering of types within a rec group, so I'll move this to a header.

tlively:
Closing this in favor of a PR from an upstream branch so I can proper stacked PRs.

PR #4498, opened 2022-02-03
[NFC] Simplify TopologicalSortStack

tlively:
Using a vector and lazily skipping finished items in `pop` rather than using a
linked list and eagerly removing duplicated items makes the code simpler and
more than twice as fast on my test case. The algorithmic complexity is unchanged
because the work to skip duplicates remains linear in the number of duplicates
added, it's just not spread out over the linear duplicate pushes any more.


tlively:
This is identical to and supersedes #4495.

kripken (approved):

PR #4499, opened 2022-02-03
CRTP topological sort utility

tlively:
Refactor the `TopologicalSortStack` into a `TopologicalSort` CRTP utility that
manipulates the DFS stack internally instead of exposing it to the user. The
only method users have to implement to use the utility is `pushPredecessors`,
which should call the provided `push` method on all the predecessors of a given
item. The public interface to `TopologicalSort` is an input iterator, so it can
easily be used in range-based for loops.

### DIFF
--- a/src/support/topological_sort.h
+++ b/src/support/topological_sort.h
@@ -19,6 +19,17 @@
 
 namespace wasm {
 
+// CRTP utility that provides an iterator through arbitrary directed acyclic
+// graphs of data that will visit the data in a topologically sorted order
+// (https://en.wikipedia.org/wiki/Topological_sorting). In other words, the
+// iterator will produce each item only after all that items predecessors have
+// been produced.
+//
+// Subclasses should call `push` on all the root items in their constructors and
+// implement a `void pushPredecessors(T item)` method that calls `push` on all
+// the immediate predecessors of `item`.
+//
+// Cycles in the graph are not detected and will result in an infinite loop.
 template<typename T, typename Subtype> struct TopologicalSort {
 private:
   // The DFS work list.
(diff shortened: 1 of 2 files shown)


src/support/topological_sort.h
```diff
@@ -0,0 +1,103 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef wasm_support_topological_sort_h
+#define wasm_support_topological_sort_h
+
+namespace wasm {
+
+template<typename T, typename Subtype> struct TopologicalSort {
```
kripken, line 22: A comment here would be good, even if just a link to a wikipedia article on topological sort (if that matches the terminology used here).

src/support/topological_sort.h
```diff
@@ -0,0 +1,103 @@
...
+  // Should be overridden by `Subtype`.
+  void pushPredecessors(T item) {
+    static_assert(&TopologicalSort<T, Subtype>::pushPredecessors !=
+                    &Subtype::pushPredecessors,
+                  "TopologicalSort subclass must implement `pushPredecessors`");
+    WASM_UNREACHABLE("Derived class must implement pushPredecessors");
+  }
+
+  // Pop until the stack is empty or it has an unfinished item on top.
+  void pop() {
```
kripken, line 39: About the name, this doesn't return an item, which popping usually does (or should). It is also not a single pop. Maybe `finishTop`? (as it finishes the top item, then does the logical work resulting from that)

kripken (commented):
Nice! lgtm % names

kripken (approved):
Great!

PR #4500, opened 2022-02-03
[Wasm GC] Fix TypeRefining corner case with uncreated types

kripken:
This pass ignores reads from structs - it only cares about writes (during a
create or a struct.set). That makes sense since we want to refine the type
of fields to more specific things based on what is actually written to them.
However, a corner case was missed: If we ignore reads, the pass may
"cleverly" optimize to something that is no longer valid to read from. How
that happens is if there is no info at all for a type - no sets or news, so all
we have is a read, which as mentioned before we ignore, so we think we
have nothing at all for that type, and can do arbitrary stuff with it. But then
the arbitrary replacement can be invalid to read from, say if it has fewer
fields.

To handle that, just emit an unreachable. If all we have is a get but no
new then there cannot be an instance here at all. (That's only true in a
closed world, of course, but this entire pass assumes that anyhow.)


src/passes/TypeRefining.cpp
```diff
@@ -199,10 +199,59 @@ struct TypeRefining : public Pass {
...
+        auto oldType = curr->ref->type.getHeapType();
+        auto newFieldType =
+          parent.finalInfos[oldType][curr->index].getBestPossible();
+        if (!Type::isSubType(newFieldType, curr->type)) {
+          // This instruction is invalid, so it must be the result of the
+          // situation described above: we ignored the read during our
+          // inference, and optimized accordingly, and so now we must remove it
+          // to keep the module validating. It doesn't matter what we emit here,
+          // since there are no struct.new or struct.sets for this type, so this
+          // code is logically unreachable.
```
tlively, line 238: It sounds like this would change in a world where the value could come in from outside the module, but I guess we should have avoided optimizing the type in that case. So I guess this code won't have to change after we handle open world modules correctly, right?

test/lit/passes/type-refining.wast
```diff
@@ -768,3 +768,70 @@
...
+  ;; CHECK-NEXT:    (struct.new_default $G)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $func
+    (drop
+      ;; We never create an instance of $I, and we only have a get of its
+      ;; field. This optimization ignores the get (to not be limited by it).
+      ;; It will then optimize $I's field of $H (another struct for which we
+      ;; have no creation, and only a get) into $G, which is driven by the fact
```
tlively, line 818: Why do we change the field to a completely unrelated type? This issue wouldn't happen if we didn't do that, right? Maybe we should be even more explicit and change the field type to `unreachable` or some other fresh type in cases where it can't possibly exist, then go fix up all the dead code (and make the type valid again) afterwards.

src/passes/TypeRefining.cpp
kripken: Yes, in an open world I think we can't run this entire pass, so we'd handle it properly that way eventually.

test/lit/passes/type-refining.wast
kripken: We basically don't know what to do when we have zero constraints. We have to pick *some* type, so we pick the one used by our supertype, see details here: https://github.com/WebAssembly/binaryen/blob/4e8b85a8d3953234bbfa490cb1d4f5c32380fb3f/src/passes/TypeRefining.cpp#L134-L172 And the supertype might have picked up a type from another branch of the type tree, which is not a subtype of ours.

The key issue, though, is simple, I think: Our constraints are writes, and we ignore reads. So if we have only reads, we seem to have no constraints, and logically we can do "anything". That is the right way to model it, I think, as we've done. But we do need to fix up the case where "anything" happens to cause a validation error.

I don't see how to mark such types as a special type without altering the entire algorithm here, but maybe I'm missing something. The issue is that a type may have no reads or writes, but its parent or subtypes might. When we propagate information we have something to propagate from the parent or subtypes - again, they do have information - so it overrides us. So we wouldn't end up with the special type any more. A more complex encoding might handle that, I guess, but again, I think the key issue is simpler to think about as mentioned in the last paragraph.

tlively: Aha, "We have to pick some type, so we pick the one used by our supertype... And the supertype might have picked up a type from another branch of the type tree, which is not a subtype of ours." makes it clear. Can you add that at the beginning of the comment here? It also might help to give the types names describing their role rather than just letters. Also, would it be possible to construct a test case with fewer types?

kripken: Added.

I managed to remove one middle type, and another field I turned from a ref to an i32, while keeping the bug present, so it is simpler now. I also added names as you suggested, and a little diagram. Hopefully it's clearer!

tlively (approved):
Awesome, thanks!

PR #4501, opened 2022-02-03
Isorecursive type fuzzing

tlively:
Add support for isorecursive types to wasm-fuzz-types by generating recursion
groups and ensuring that children types are only selected from candidates
through the end of the current group. For non-isorecursive systems, treat all
the types as belonging to a single group so that their behavior is unchanged.

Also fix two small bugs found by the fuzzer: LUB calculation was taking the
wrong path for isorecursive types and isorecursive validation was not handling
basic heap types properly.


src/tools/fuzzing/heap-types.cpp
```diff
@@ -47,6 +47,13 @@ struct HeapTypeGeneratorImpl {
   using HeapTypeKind = std::variant<BasicKind, SignatureKind, DataKind>;
   std::vector<HeapTypeKind> typeKinds;
 
+  // For each type, the index one past the end of its recursion group, used to
+  // determine what types could be valid children.
+  std::vector<Index> recGroupEnds;
```
kripken, line 52: One past the end feels a little odd here. Maybe it's because it's not a pair, it's a single value?

src/tools/fuzzing/heap-types.cpp
```diff
@@ -80,35 +86,68 @@ struct HeapTypeGeneratorImpl {
...
+            recGroupEnds.push_back(newGroupStart);
+          }
+          groupStart = newGroupStart;
+        }
+      }
+      // Finish the last group.
+      builder.createRecGroup(groupStart, builder.size() - groupStart);
+      for (Index j = groupStart; j < builder.size(); ++j) {
+        recGroupEnds.push_back(builder.size());
+      }
```
kripken, line 118: I think you can avoid the extra code on the outside for the last group by doing a more efficient loop that picks group sizes, rather than tests on each item for "is end of group", like this:
```python
position = 0
while position < builder.size():
  curr_size = std::min(upTo(groupSize), builder.size())
  add_group([position, position + curr_size])
  position += curr_size
```
But if you have a preference for your current loop, how about replacing `if (rand.oneIn(expectedSize))` with `if (rand.oneIn(expectedSize) || i == builder.size() - 1)`?

test/lit/fuzz-types/isorecursive.test
```diff
@@ -0,0 +1,24 @@
...
+;; CHECK-NEXT: 10: (array i16)
+;; CHECK-NEXT: 11: (array (rtt ...1))
+;; CHECK-NEXT: 12: (array i16)
+;; CHECK-NEXT: 13: (array (rtt 2 (array (rtt ...1))))
+;; CHECK-NEXT: 14: (array i16)
+;; CHECK-NEXT: 15: (func (result (rtt extern) (ref (func (result (rtt extern) ...1 i32 (rtt 0 (array (rtt 2 (array (rtt ...1))))) (ref null (array (rtt ...1))) (rtt (array i16))))) i32 (rtt 0 (array (rtt 2 (array (rtt ...1))))) (ref null (array (rtt ...1))) (rtt (array i16))))
+;; CHECK-NEXT: 16: (array (rtt (array ...1)))
+;; CHECK-NEXT: 17: (array i16)
+;; CHECK-NEXT: 18: (array i16)
+;; CHECK-NEXT: 19: (func (result (rtt extern) (ref (func (result (rtt extern) ...1 i32 (rtt 0 (array (rtt 2 (array (rtt ...1))))) (ref null (array (rtt ...1))) (rtt (array i16))))) i32 (rtt 0 (array (rtt 2 (array (rtt ...1))))) (ref null (array (rtt ...1))) (rtt (array i16))))
```
kripken, line 24: why are there only arrays and a few funcs, and no structs?

tlively: Just a quirk of this seed. Other seeds look more reasonable. As a follow-up, I'm going to improve the printing here so we can see the groups and supertypes.

src/tools/fuzzing/heap-types.cpp
tlively: I like the current loop because it gives a nice geometric distribution for the size of the group, but I'll see if I can deduplicate the code.

src/tools/fuzzing/heap-types.cpp
tlively: Another way to look at it is cumulative sizes of the current and all prior rec groups for each index.

test/lit/fuzz-types/isorecursive.test
kripken: Hmm, perhaps pick another seed then? One with nice output would be good to see it and to check for regressions later.

tlively: The change I made to the loop you commented on resulted in one less call to `rand.oneIn`, so now the output looks better.

tlively:
(enabled automerge)

kripken (commented):
lgtm % changing seed, unless there is a reason not to do so?

kripken (approved):
Ah, just saw the comment and test update now. lgtm!

PR #4503, opened 2022-02-04
Generate heap type names when printing types

tlively:
The previous printing system in the Types API would print the full recursive
structure of a Type or HeapType with special markers using de Bruijn indices to
avoid infinite recursion and a separate special marker for when the size
exceeded an arbitrary upper limit. In practice, the types printed by that system
were not human readable, so all that complexity was not useful.

Replace that system with a new system that always emits a HeapType name rather
than recursing into the structure of inner HeapTypes. Add methods for printing
Types and HeapTypes with custom HeapType name generators. Also add a new
wasm-type-printing.h header with off-the-shelf type name generators that
implement simple naming schemes sufficient for tests and the type fuzzer.

Note that these new printing methods and the old printing methods they augment
are not used for emitting text modules. Printing types as part of expressions
and modules is handled by separate code in Print.cpp and the printing API
modified in this PR is mostly used for debugging. However, the new printing
methods are general enough that Print.cpp should be able to use them as well, so
update the format used to print types in the modified printing system to match
the text format in anticipation of making that change in a follow-up PR.


src/wasm/wasm-type.cpp
```diff
@@ -196,42 +196,57 @@ struct TypeBounder {
...
-  size_t currDepth = 0;
-  std::unordered_map<TypeID, size_t> depths;
+  // Whether to print explicit supertypes.
+  bool hasSupertype;
 
   // The stream we are printing to.
   std::ostream& os;
 
-  TypePrinter(std::ostream& os) : os(os) {}
+  // The default generator state if no other generated is provided.
```
kripken, line 217:
```suggestion
  // The default generator state if no other generator is provided.
```

src/wasm/wasm-type.cpp
```diff
@@ -196,42 +196,57 @@ struct TypeBounder {
...
+
+  void getName(std::ostream& os, HeapType type);
+};
+
+// Helper for printing types.
 struct TypePrinter {
-  size_t currDepth = 0;
-  std::unordered_map<TypeID, size_t> depths;
+  // Whether to print explicit supertypes.
+  bool hasSupertype;
```
kripken, line 212: This confuses me. This is a TypePrinter for multiple types, while `hasSupertype` sounds like a property of a single type?

tlively: I'll change this to `printSupertypes` to make it more clear.

kripken (approved):
Great!

tlively:
@kripken, PTAL at the latest commit. I realized that updating the example tests would have been really burdensome with the existing API, so I added a new wasm-type-printing.h API layered on top of it to handle the common cases.

I also changed the signature of the `HeapTypeNameGenerator` function to return a `TypeNames`, which I have moved from wasm.h to wasm-type.h. That change is more future-proof and makes the composition of name generators a lot simpler.

tlively:
The reason to use size_t instead of Index is kind of silly: Index is defined in wasm.h and so far we've avoided having wasm-types.h depend on anything in wasm.h.

kripken:
Hmm, it does increase some data structures unnecessarily, though - that may have a memory or performance cost.

Perhaps have wasm-types define Index, or have a tiny header they both use?

tlively:
Sure, I'll split it off into a tiny header.

kripken (approved):

PR #4507, opened 2022-02-08
Eagerly canonicalize basic types

tlively:
We were already eagerly canonicalizing basic HeapTypes when building types so
the more complicated canonicalization algorithms would not have to handle
noncanonical heap types, but we were not doing the same for Types. Equirecursive
canonicalization was properly handling noncanonical Types everywhere, but
isorecursive canonicalization was not. Rather than update isorecursive
canonicalization in multiple places, remove the special handling from
equirecursive canonicalization and canonicalize types in a single location.

### DIFF
--- a/test/gtest/type-builder.cpp
+++ b/test/gtest/type-builder.cpp
@@ -459,19 +459,23 @@ TEST_F(IsorecursiveTest, CanonicalizeTypesBeforeSubtyping) {
 }
 
 static void testCanonicalizeBasicTypes() {
-  TypeBuilder builder(3);
+  TypeBuilder builder(5);
 
   Type externref = builder.getTempRefType(builder[0], Nullable);
+  Type externrefs = builder.getTempTupleType({externref, externref});
 
   builder[0] = HeapType::ext;
   builder[1] = Struct({Field(externref, Immutable)});
   builder[2] = Struct({Field(Type::externref, Immutable)});
+  builder[3] = Signature(externrefs, Type::none);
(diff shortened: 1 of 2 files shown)


test/gtest/type-builder.cpp
```diff
@@ -457,3 +457,30 @@ TEST_F(IsorecursiveTest, CanonicalizeTypesBeforeSubtyping) {
...
+  TypeBuilder builder(5);
+
+  Type externref = builder.getTempRefType(builder[0], Nullable);
+  Type externrefs = builder.getTempTupleType({externref, externref});
+
+  builder[0] = HeapType::ext;
+  builder[1] = Struct({Field(externref, Immutable)});
+  builder[2] = Struct({Field(Type::externref, Immutable)});
+  builder[3] = Signature(externrefs, Type::none);
+  builder[4] = Signature(Type({Type::externref, Type::externref}), Type::none);
```
kripken, line 471: Which Type constructor is called here? Reading wasm-type.h I can't figure that out.

tlively: `Type(std::initializer_list<Type>)`, which constructs a tuple type.

kripken: I see, thanks. I've never really understood that part of C++ I guess...

kripken (approved):

test/gtest/type-builder.cpp
tlively: I removed the `Type` here to bring it more in line with how we use this in other places.

PR #4510, opened 2022-02-09
Fuzz structural type canonicalization

tlively:
Add a new fuzz checker to wasm-type-fuzzer that builds copies of the originally
built types, randomly selecting for each child type from all potential sources,
including both the originally built types and the not-yet-built duplicate types.
After building the new types, check that they are indeed identical to the old
types, which means that nothing has gone wrong with canonicalization.


src/tools/wasm-fuzz-types.cpp
```diff
@@ -189,6 +197,292 @@ void Fuzzer::checkLUBs(const std::vector<HeapType>& types) {
...
+
+  // Helper for creating new definitions of existing types, randomly choosing
+  // between canonical and temporary components.
+  struct Copier {
+    Random& rand;
+    const std::vector<HeapType>& types;
+    TypeBuilder& builder;
+
+    // For each type, the indices in `types` at which it appears.
+    std::unordered_map<HeapType, std::vector<Index>> typeIndices;
```
kripken, line 219: Maybe `SmallVector<1>`?

kripken (approved):

src/tools/wasm-fuzz-types.cpp
tlively: Repeats aren't super uncommon because the fuzzer goes slightly out of its way to make them more likely, so I think that would document the wrong intent.

kripken: Ah, I see, sgtm then.

PR #4511, opened 2022-02-09
Make IndexedTypeNameGenerator more powerful

tlively:
Allow IndexedTypeNameGenerator to be configured with a custom prefix and also
allow it to forward additional arguments to initializing its fallback generator.
This allows multiple IndexedTypeNameGenerators to be composed together, for
example.

### DIFF
--- a/src/wasm-type-printing.h
+++ b/src/wasm-type-printing.h
@@ -67,17 +67,23 @@ struct DefaultTypeNameGenerator
 template<typename FallbackGenerator = DefaultTypeNameGenerator>
 struct IndexedTypeNameGenerator
   : TypeNameGeneratorBase<IndexedTypeNameGenerator<FallbackGenerator>> {
-  FallbackGenerator fallback;
+  DefaultTypeNameGenerator defaultGenerator;
+  FallbackGenerator& fallback;
   std::unordered_map<HeapType, TypeNames> names;
-  template<typename T, typename... Args>
+
+  template<typename T>
   IndexedTypeNameGenerator(T& types,
-                           const std::string& prefix = "",
-                           Args&&... args)
-    : fallback(std::forward<Args>(args)...) {
+                           FallbackGenerator& fallback,
+                           const std::string& prefix = "")
+    : fallback(fallback) {
     for (size_t i = 0; i < types.size(); ++i) {
       names.insert({types[i], {prefix + std::to_string(i), {}}});
     }
   }
+  template<typename T>
+  IndexedTypeNameGenerator(T& types, const std::string& prefix = "")
+    : IndexedTypeNameGenerator(types, defaultGenerator, prefix) {}
+
   TypeNames getNames(HeapType type) {
     if (auto it = names.find(type); it != names.end()) {
       return it->second;
--- a/src/wasm-type-printing.h
+++ b/src/wasm-type-printing.h
@@ -69,9 +69,13 @@ struct IndexedTypeNameGenerator
   : TypeNameGeneratorBase<IndexedTypeNameGenerator<FallbackGenerator>> {
   FallbackGenerator fallback;
   std::unordered_map<HeapType, TypeNames> names;
-  template<typename T> IndexedTypeNameGenerator(T& types) {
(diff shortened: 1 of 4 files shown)


test/gtest/type-builder.cpp
```diff
@@ -101,6 +102,46 @@ TEST_F(TypeTest, TypeIterator) {
...
+  auto result = builder.build();
+  ASSERT_TRUE(result);
+  auto built = *result;
+
+  std::vector<HeapType> structs{built[0], built[1]};
+  std::vector<HeapType> arrays{built[2], built[3]};
+
+  using ArrayPrinter = IndexedTypeNameGenerator<DefaultTypeNameGenerator>;
+  using StructPrinter = IndexedTypeNameGenerator<ArrayPrinter>;
+  StructPrinter print(structs, "struct", arrays, "array");
```
kripken, line 126: hmm...
```cpp
AP = ITNG<Default>;
SP = ITNG<AP>;
```
It's like a chain?

What actually ends up consuming the args on this line? `IndexedTypeNameGenerator`'s constructor just forwards them, and I'm missing where a parent does so.

Is this C++ type magic necessary? :smile:

tlively: Yep, they compose into a chain. The full type that gets instantiated is `IndexedTypeNameGenerator<IndexedTypeNameGenerator<DefaultTypeNameGenerator>>`, but of course that's too long to write out all at once. In principle you can make it as long as you want.

What actually consumes the args is the initializer for the inner `fallback` name generator: `fallback(std::forward<Args>(args)...)`. In the case of the test here, that means that `fallback` is initialized with `arrays, "array"`.

> Is this C++ type magic necessary? 😄

Not really, no 😁 I did find this useful when debugging the new type fuzzing stuff, though. I created an indexed type printer for the new temporary types that fell back to a different indexed type printer (with a different prefix) for the original types and that made it easy to see which types came from where and how they all fit together.

tlively:
@kripken, I think you'll like the latest change. Instead of using `std::forward` to forward arguments to the constructor for a fallback generator, now we just take a reference to a fallback generator in the more general constructor with another constructor for the special case of not having an explicit fallback. The API is slightly more dangerous now because dangling references could be a problem, but it's also more intuitive and easier to read this way.

src/wasm-type-printing.h
```diff
@@ -67,13 +67,23 @@ struct DefaultTypeNameGenerator
...
+    : fallback(fallback) {
     for (size_t i = 0; i < types.size(); ++i) {
-      names.insert({types[i], {std::to_string(i), {}}});
+      names.insert({types[i], {prefix + std::to_string(i), {}}});
     }
   }
+  template<typename T>
+  IndexedTypeNameGenerator(T& types, const std::string& prefix = "")
+    : IndexedTypeNameGenerator(types, defaultGenerator, prefix) {}
+
```
kripken, line 86: This is much simpler, thanks!

We could use std::unique_ptr if you're worried about dangling references here I guess. But I am not too worried.

test/gtest/type-builder.cpp
```diff
@@ -101,6 +102,47 @@ TEST_F(TypeTest, TypeIterator) {
...
+  ASSERT_TRUE(result);
+  auto built = *result;
+
+  std::vector<HeapType> structs{built[0], built[1]};
+  std::vector<HeapType> arrays{built[2], built[3]};
+
+  using ArrayPrinter = IndexedTypeNameGenerator<DefaultTypeNameGenerator>;
+  ArrayPrinter printArrays(arrays, "array");
+  using StructPrinter = IndexedTypeNameGenerator<ArrayPrinter>;
+  StructPrinter print(structs, printArrays, "struct");
```
kripken, line 127: I not sure how to read the intent of this code. Why do we need a special array and  struct printers? Why are those types not printed by the default printer?

tlively: Yeah the names here are very similar to the names produced by the default printer, but the point of the test is to test `IndexedTypeNameGenerator` specifically. I updated the name of the test and added a comment.

PR #4513, opened 2022-02-09
[wasm-split] Add an --asyncify option

tlively:
Add an option for running the asyncify transformation on the primary module
emitted by wasm-split. The idea is that the placeholder functions should be able
to unwind the stack while the secondary module is asynchronously loaded, then
once the placeholder functions have been patched out by the secondary module the
stack should be rewound and end up in the correct secondary function.

### DIFF
--- a/src/tools/wasm-split/wasm-split.cpp
+++ b/src/tools/wasm-split/wasm-split.cpp
@@ -280,6 +280,16 @@ void splitModule(const WasmSplitOptions& options) {
   adjustTableSize(wasm, options.initialTableSize);
   adjustTableSize(*secondary, options.initialTableSize);
 
+  // Run asyncify on the primary module
+  if (options.asyncify) {
+    PassOptions passOptions;
+    passOptions.optimizeLevel = 1;
+    passOptions.arguments.insert({"asyncify-ignore-imports", ""});
+    PassRunner runner(&wasm, passOptions);
+    runner.add("asyncify");
+    runner.run();
+  }
+
   if (options.symbolMap) {
     writeSymbolMap(wasm, options.primaryOutput + ".symbols");
     writeSymbolMap(*secondary, options.secondaryOutput + ".symbols");
--- a/test/lit/wasm-split/asyncify.wast
+++ b/test/lit/wasm-split/asyncify.wast
@@ -1,0 +1,174 @@
+;; RUN: wasm-split %s --export-prefix='%' -g -o1 %t.1.wasm -o2 %t.2.wasm --keep-funcs=foo --asyncify
+;; RUN: wasm-dis %t.1.wasm | filecheck %s --check-prefix PRIMARY
+;; RUN: wasm-dis %t.2.wasm | filecheck %s --check-prefix SECONDARY
+
+;; Check that the --asyncify option instruments the primary module but not the
+;; secondary module.
+
+(module
+ (func $foo (param i32) (result i32)
+  (call $bar (i32.const 0))
+ )
+ (func $bar (param i32) (result i32)
+  (call $foo (i32.const 1))
+ )
+)
+
+;; PRIMARY:      (module
+;; PRIMARY-NEXT:  (type $i32_=>_i32 (func (param i32) (result i32)))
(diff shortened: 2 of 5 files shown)


test/lit/wasm-split/asyncify.wast
```diff
@@ -0,0 +1,174 @@
...
+;; PRIMARY-NEXT:  (table $0 1 funcref)
+;; PRIMARY-NEXT:  (elem (i32.const 0) $placeholder_0)
+;; PRIMARY-NEXT:  (export "%foo" (func $foo))
+;; PRIMARY-NEXT:  (export "%table" (table $0))
+;; PRIMARY-NEXT:  (export "asyncify_start_unwind" (func $asyncify_start_unwind))
+;; PRIMARY-NEXT:  (export "asyncify_stop_unwind" (func $asyncify_stop_unwind))
+;; PRIMARY-NEXT:  (export "asyncify_start_rewind" (func $asyncify_start_rewind))
+;; PRIMARY-NEXT:  (export "asyncify_stop_rewind" (func $asyncify_stop_rewind))
+;; PRIMARY-NEXT:  (export "asyncify_get_state" (func $asyncify_get_state))
+;; PRIMARY-NEXT:  (func $foo (param $0 i32) (result i32)
```
tlively, line 35: This is a shocking amount of code to instrument a single indirect call. Is there room for improvement here? It looks like there is a stack pointer being manipulated, probably into the stack of spilled locals, but this function doesn't need to spill any locals (AFAIK), so I'm surprised that code is generated.

kripken: Yeah, quite a lot...

This is with both wasm-split and asyncify, though - do you have the code after just wasm-split? I think asyncify is seeing more than just a single call, as it seems to be spilling locals - does wasm-split add any?

tlively: This is the code after just wasm-split:
```
 (func $foo (param $0 i32) (result i32)
  (call_indirect (type $i32_=>_i32)
   (i32.const 0)
   (i32.const 0)
  )
 )
 ```

File: src/tools/wasm-split/wasm-split.cpp [hunk 2]
```diff
@@ -280,6 +280,16 @@ void splitModule(const WasmSplitOptions& options) {
   adjustTableSize(wasm, options.initialTableSize);
   adjustTableSize(*secondary, options.initialTableSize);
 
+  // Run asyncify on the primary module
+  if (options.asyncify) {
+    PassOptions passOptions;
+    passOptions.optimizeLevel = 1;
+    passOptions.arguments.insert({"asyncify-ignore-imports", ""});
```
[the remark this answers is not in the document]
tlively: wasm-split transforms direct calls to secondary functions into indirect calls that either go to placeholder functions or to the original secondary functions, depending on the contents of the table. Those indirect calls and any other indirect calls that might end up in a placeholder function can cause a pause.

test/lit/wasm-split/asyncify.wast
kripken: Ah, ok, thanks...

Reading the output more carefully, it could indeed be better. It is handling the case of a return value from a call carefully by adding another local, which forces it to add spilling code. That doesn't matter here I think since the code is so trivial, but it doesn't take advantage of that. Second, it writes and reads the "pause position" index in the function, but there is just a single possible position here, so that code is redundant as well - again, we don't have a specific optimization to benefit from the case of a single possible pause site, so the code is more general than it needs to be...

src/tools/wasm-split/wasm-split.cpp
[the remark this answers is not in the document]
kripken: Ah, right, thanks, so this ignores all imports and just the indirect calls are the possible pausers. Makes sense.

kripken (approved):
Lgtm on the code. Were you also looking for a review on the general direction of integrating Asyncify in this tool (as opposed to running it on the emscripten side)? (if so I'd need to know more about the plans here)

tlively:
Talked about the general direction offline, and I think this PR makes sense to land.

PR #4514, opened 2022-02-09
DeadArgumentElimination: Remove removable effects

kripken:


test/lit/passes/dae_tnh.wast
```diff
@@ -0,0 +1,35 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
+;; NOTE: This test was ported using port_test.py and could be cleaned up.
```
tlively, line 2: Was this generated with `port_test.py` or that left over from a copy-paste?

tlively (approved):

test/lit/passes/dae_tnh.wast
kripken: Ah, heh, that's a copy-paste... good catch.

Issue #4285, opened 2021-10-28
Clarify that -O is equivalent to -Os in --help


RReverser:
From `--help` alone, it's unclear whether `-O` will optimize for size like `-Os` / `-Oz` or speed like `-O{number}`. It would be good to clarify that to make the choice easier for the user.

PR #4516, opened 2022-02-15
#4285: Clarify that `-O` is equivalent to `-Os` in tools help message.

t4lz:
Also update two relevant tests.

#4285

### DIFF
--- a/src/tools/optimization-options.h
+++ b/src/tools/optimization-options.h
@@ -27,6 +27,8 @@ namespace wasm {
 
 struct OptimizationOptions : public ToolOptions {
   static constexpr const char* DEFAULT_OPT_PASSES = "O";
+  static constexpr const int OS_OPTIMIZE_LEVEL = 2;
+  static constexpr const int OS_SHRINK_LEVEL = 1;
 
   std::vector<std::string> passes;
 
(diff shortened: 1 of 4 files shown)


src/tools/optimization-options.h
```diff
@@ -39,7 +39,7 @@ struct OptimizationOptions : public ToolOptions {
     (*this)
       .add("",
            "-O",
-           "execute default optimization passes",
+           "execute default optimization passes (equivalent to -Os)",
```
kripken, line 42: I worry this might get stale if we change the default but forget to change this help text. How about adding some assertions to verify that, say after line 46 where we set the default opts, we can check that `optimizeLevel == 2, shrinkLevel == 1`?

t4lz: Good point, how is the new version?
With the static constants in `OptimizationOptions`, we also catch changes that are done on `-Os`'s side.
Added another commit also introducing static constants with the default options to `PassOptions`, in order to make the assertion a `static_assert`. How is that? I suggest I would squash that second commit if you like that solution and delete it otherwise.

kripken (approved):
Nice! Yes, this looks better, good idea to add constants.

src/tools/optimization-options.h
kripken: (side note, about landing, we squash all PRs anyhow, to guarantee the main branch is 100% bisectable. so no need to worry about the commits in the PR, that is.)

kripken:
Looks like there's a minor lint issue - running `clang-format` should fix it, or you can copy-paste the diff in the error message.

kripken (approved):
Thanks!

PR #4517, opened 2022-02-18
Include type names in error messages from building

tlively:
Instead of just reporting the type index that causes an error when building
types, report the name of the responsible type when parsing the text format.

### DIFF
--- a/src/wasm/wasm-s-parser.cpp
+++ b/src/wasm/wasm-s-parser.cpp
@@ -936,7 +936,16 @@ void SExpressionWasmBuilder::preParseHeapTypes(Element& module) {
 
   auto result = builder.build();
   if (auto* err = result.getError()) {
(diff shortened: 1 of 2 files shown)


src/wasm/wasm-s-parser.cpp
```diff
@@ -936,7 +936,16 @@ void SExpressionWasmBuilder::preParseHeapTypes(Element& module) {
 
   auto result = builder.build();
   if (auto* err = result.getError()) {
-    Fatal() << "Invalid type: " << err->reason << " at index " << err->index;
+    // Find the name to provide a better error message.
+    std::stringstream msg;
+    msg << "Invalid type: " << err->reason;
+    for (auto& [name, index] : typeIndices) {
```
kripken, line 942: Heh, `O(N)` seems ok for reporting an error...

kripken (approved):
Nice!

PR #4520, opened 2022-02-24
[Wasm GC] Optimize static casts in br_on_cast*

kripken:
We were missing this particular case, which we can in fact handle
when the cast is static.

Noticed by @jakobkummerow

### DIFF
--- a/test/lit/passes/remove-unused-brs-gc.wast
+++ b/test/lit/passes/remove-unused-brs-gc.wast
@@ -183,4 +183,36 @@
    (unreachable)
   )
  )
+
+ ;; CHECK:      (func $br_on_cast_dynamic (result (ref $struct))
+ ;; CHECK-NEXT:  (local $temp (ref null $struct))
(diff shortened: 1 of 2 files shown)


test/lit/passes/remove-unused-brs-gc.wast
```diff
@@ -111,5 +111,108 @@
...
+ ;; CHECK-NEXT:    (struct.new_default $struct)
+ ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:   (unreachable)
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT: )
+ (func $br_on_cast_fail_static (result (ref $struct))
+  (local $temp (ref null $struct))
+  (block $block (result (ref $struct))
+   (drop
+    ;; As $br_on_cast_static, but this is a failing cast, so we know it will
```
tlively, line 177:
```suggestion
    ;; As $br_on_cast_static, but this checks for a failing cast, so we know it will
```

tlively (approved):

PR #4523, opened 2022-03-01
feat(wasm2js): export global var

### LINKED ISSUE #4522: [wasm2js] globals are not exported from generated js file
All globals are missing and cause the generated JS not working.

```lisp
(module
    (global $global0 i32 (i32.const 655360))
    (export "HELLO" (global $global0))
    (func (result i32) (i32.const 42))
    (export "helloWorld" (func 0))
)
```

Run it with

> wasm2js -O0

```js
function asmFunc(env) {
 var Math_imul = Math.imul;
 var Math_fround = Math.fround;
 var Math_abs = Math.abs;
 var Math_clz32 = Math.clz32;
 var Math_min = Math.min;
 var Math_max = Math.max;
 var Math_floor = Math.floor;
 var Math_ceil = Math.ceil;
 var Math_trunc = Math.trunc;
 var Math_sqrt = Math.sqrt;
 var abort = env.abort;
 var nan = NaN;
 var infinity = Infinity;
 var global$0 = 655360;
 function $0() {
  return 42 | 0;
 }

 return {
  "HELLO": global$0,
  "helloWorld": $0
 };
}

var retasmFunc = asmFunc(  { abort: function() { throw new Error('abort'); }
  });
export var helloWorld = retasmFunc.helloWorld;
```

It should also exports HELLO

magic-akari:
- fix #4522

### DIFF
--- a/test/wasm2js/export_global.2asm.js
+++ b/test/wasm2js/export_global.2asm.js
@@ -19,7 +19,11 @@ function asmFunc(env) {
  }
  
  return {
-  "HELLO": global0, 
+  "HELLO": {
+   get value() {
+    return global0;
+   }
+  }, 
   "helloWorld": $0
  };
 }
--- a/test/wasm2js/export_global.2asm.js
+++ b/test/wasm2js/export_global.2asm.js
@@ -22,6 +22,9 @@ function asmFunc(env) {
   "HELLO": {
    get value() {
     return global0;
+   }, 
+   set value(_global0) {
+    global0 = _global0;
    }
   }, 
   "helloWorld": $0
--- a/test/wasm2js/export_global.2asm.js
+++ b/test/wasm2js/export_global.2asm.js
@@ -1,0 +1,30 @@
+
+function asmFunc(env) {
+ var Math_imul = Math.imul;
+ var Math_fround = Math.fround;
+ var Math_abs = Math.abs;
+ var Math_clz32 = Math.clz32;
+ var Math_min = Math.min;
+ var Math_max = Math.max;
+ var Math_floor = Math.floor;
+ var Math_ceil = Math.ceil;
+ var Math_trunc = Math.trunc;
+ var Math_sqrt = Math.sqrt;
+ var abort = env.abort;
+ var nan = NaN;
+ var infinity = Infinity;
+ var global0 = 655360;
+ function $0() {
+  return 42 | 0;
+ }
+ 
+ return {
+  "HELLO": global0, 
+  "helloWorld": $0
+ };
+}
+
+var retasmFunc = asmFunc(  { abort: function() { throw new Error('abort'); }
+  });
+export var HELLO = retasmFunc.HELLO;
+export var helloWorld = retasmFunc.helloWorld;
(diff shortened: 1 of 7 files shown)


test/wasm2js/export_global.2asm.js
```diff
@@ -0,0 +1,30 @@
...
+ var abort = env.abort;
+ var nan = NaN;
+ var infinity = Infinity;
+ var global0 = 655360;
+ function $0() {
+  return 42 | 0;
+ }
+ 
+ return {
+  "HELLO": global0,
```
kripken, line 22: I think in wasm an exported global is an object `{ value: .. }`. That allows mutable globals to be exported. If I am not wrong, then we should export such an object here.

However, I think for mutable globals we'd need more work, as we'd need to access the same object inside the module. But I think it is ok to not handle that in this PR. We can handle just the immutable case for now. (But I do think we should export an object as the wasm JS API expects.)

Jack-Works: if export an object instead, it will make the exports of the generated JS doesn't match the WASM. for mutable globals, I suggest generating this code:

```diff
+export var HELLO
 function asmFunc(env) {
  // ...
- var global0 = 655360;
+ HELLO = 655360;
  function $0() {
   return 42 | 0;
  }
 
  return {
-  "HELLO": global0, 
   "helloWorld": $0
  };
 }

 var retasmFunc = asmFunc(  { abort: function() { throw new Error('abort'); }
   });
-export var HELLO = retasmFunc.HELLO;
 export var helloWorld = retasmFunc.helloWorld;
```

kripken: What do you mean by "won't match the wasm"?

IIANM, the wasm version emits an object with `.value` as described here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Global

Jack-Works: my wasm glue code use them as `wasm.HELLO` not `wasm.value.HELLO`, so if it exports in this way, our JS glue code will still be broken...

Jack-Works (approved):

test/wasm2js/export_global.2asm.js
Jack-Works: oh I reinvestigated again, it do match the shape.

image

thanks!

kripken (approved):
Nice use of a getter! I had thought to do it a different way, but this is better.

magic-akari:
I will try to add a setter.

magic-akari:
Currently, we support getter and setter for both immutable and mutable global. I think it is OK.

We don't support quoted getters/setters yet, they are valid JS syntax as well.
I think this is acceptable for now, since we are actually only using the known value of `value`.

kripken (approved):
Thanks! Looks good. Yes, let's land this, and any other improvements needed can be done later.

Jack-Works:
hi @kripken can you release a new version? thanks!

kripken:
Sure, I opened https://github.com/WebAssembly/binaryen/pull/4533 now

PR #4525, opened 2022-03-04
--enable-shared-memory pass

caiiiycuk:
Hi, I want to use shared memory for my WebAssembly module. It's needed for js-dos. Currently emulation works in worker and frame updates are send using postMessage. I want to use SharedArrayBuffer to avoid copying/sending big rgba data. I dont' need to use pthread or threading, I just want to share memory between worker and main process. 

I found similar question here: https://github.com/emscripten-core/emscripten/issues/12098
As suggested I tried to use wasm2wat and then wat2wasm, but it not working for me. Looks like it affects  -s INVOKE_RUN=0, because after wasm2wat my module immediately starts and I can't provide own memory.

Without wasm2wat I have this error:

```LinkError: WebAssembly.instantiate(): mismatch in shared state of memory, declared = 0, imported = 1```

After wasm2wat this:

```LinkError: WebAssembly.instantiate(): mismatch in shared state of memory, declared = 1, imported = 0```

So, I decided to add this pass to binaryen.

--
[google groups source](https://groups.google.com/g/emscripten-discuss/c/IsRC-MRGQjw)

### DIFF
--- a/src/passes/EnableSharedMemory.cpp
+++ b/src/passes/EnableSharedMemory.cpp
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2017 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "pass.h"
+#include "wasm-features.h"
+
+namespace wasm {
+
+struct EnableSharedMemory : public Pass {
+  void run(PassRunner* runner, Module* module) override {
+    if (module->memory.shared) {
+      return;
+    }
+
+    module->memory.shared = true;
+    module->features.set(FeatureSet::Atomics, true);
+  };
+};
+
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/passes/EnableSharedMemory.cpp [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+namespace wasm {
+
+struct EnableSharedMemory : public Pass {
+  void run(PassRunner* runner, Module* module) override {
+    if (module->memory.shared) {
+      return;
+    }
+
+    module->memory.shared = true;
+    module->features.set(FeatureSet::Atomics, true);
```
tlively, line 29: We generally avoid adding new features that were not present in the original module or added on the command line, so perhaps this should instead check whether `atomics` is enabled and print a warning if it not?

caiiiycuk: Without it validation fails in [wasm-validator.cpp](https://github.com/WebAssembly/binaryen/blob/6247e7cd9be619d53c926975690981aa267917f9/src/wasm/wasm-validator.cpp#L3036-L3040)

```
  if (curr.shared) {
    info.shouldBeTrue(module.features.hasAtomics(),
                      "memory",
                      "memory is shared, but atomics are disabled");
  }
```

I am not sure if it affects something. Can I turn it into warning?

tlively: I have something like this in mind:

```
if (atomics)
  memory.shared = true;
else
  warn("Cannot make the memory shared without --enable-atomics")
```

tlively:
Also, it would be good to add a test for this in test/lit/passes

sbc100:
It sounds like https://github.com/emscripten-core/emscripten/pull/16419 should solve this issue.

caiiiycuk:
Cancelled, because `-sSHARED_MEMORY=1 -sUSE_THREADS=1` works fine. Thank you.

Issue #3059, opened 2020-08-18
Replacing Stack IR with Poppy IR


tlively:
The latest code pushed to my [`stackify` branch](https://github.com/tlively/binaryen/tree/stackify) ([diff](https://github.com/tlively/binaryen/compare/master...tlively:stackify)) implements a new form of IR ("Poppy IR") that directly represents the stack machine structure of WebAssembly and is meant to replace Stack IR for performing final stack machine optimizations before binary writing. This new system is motivated primarily by our need to emit efficient multivalue blocks and functions, which Stack IR cannot optimize.

The code I've written so far is sufficient to demonstrate that Poppy IR works and can generate smaller code with simpler passes than the current Stack IR pipeline, but it is very much a prototype. There are no tests, the validator is incomplete, the unstackify pass is buggy, and the pass runner does not prevent users from running passes on the incorrect form of IR. That being said, most of what has been written has been extensively fuzzed (although is not bug-free) and I expect that most of the code will eventually be able to land with minimal modifications. You can try it yourself by building my branch and running with `BINARYEN_USE_STACKIFY=1` in your environment.

Here is a potential roadmap of discrete patches I plan to land to get Poppy IR upstream. I welcome discussion of how to better split up the work. As I land patches, I will check them off here and add links to the respective PRs.

- [ ] Stack utils for analyzing Poppy IR with C++ API tests
- [ ] Implement parsing of wast files directly to Poppy IR and printing
- [ ] Validation
- [ ] Stackification with tests and fuzzing
- [ ] Unstackification with tests and fuzzing
- [ ] Poppy IR optimization passes with tests and fuzzing
- [ ] Pass runner integration and Stack IR replacement via environment variable
- [ ] Remove Stack IR

Follow up work to immediately take advantage of Stack IR:

- [ ] Move Binaryen to C++17 (prerequisite)
- [ ] Replace wast parser with [wasp](https://github.com/binji/wasp)

Follow up related investigative work:

- [ ] Look at why Binaryen's DCE pass is so complicated and whether it could be shared with Poppy IR
- [ ] Investigate removing unreachable control flow from Binaryen entirely.

Additional cleanup work to be done if at any point it would be helpful for achieving the other goals:

- [ ] Move various test suites to a lit + FileCheck test framework

PR #4526, opened 2022-03-08
Add CHANGELOG entry for v102 OptimizeInstructions

ospencer:
We (the Grain team) ran into a funny thing when upgrading to Binaryen 102—we occasionally use tuples to use the wasm stack via a direct `BinaryenTupleExtract` on a `BinaryenTupleMake` to avoid managing an extra local. We support some runtimes that still don't have multivalue, but by never returning a tuple from an `if` or `block` we've avoided any actual use of multivalue in the generated binaries. The update to OptimizeInstructions allows the `BinaryenTupleExtract` to be pulled outside of an `if`, which makes the `if` actually return multiple values. We shouldn't have been abusing tuples in this way, but we figured we'd call it out in the changelog for anyone else who might have done the same 😅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@ v102
 
 - Replace `BinaryenExpressionGetSideEffects`'s features parameter with a module
   parameter.
+  
+- OptimizeInstructions now lifts identical code in `select`/`if` arms (#3828). This may cause direct `BinaryenTupleExtract(BinaryenTupleMake(...))` to [use multivalue types](https://github.com/grain-lang/grain/pull/1158).
 
 v101
 ----


tlively:
@ospencer, out of curiosity, do you plan to take advantage of multivalue in the future? (or alternatively, would you use it now if it were better supported by tools + engines?)

ospencer:
Yes, absolutely. In fact, something I was working on just this weekend I really wished I could use multivalue to do codegen for. It's mostly an engine thing—multivalue would be fantastic for avoiding allocations when passing around ADT variants, would be helpful for cleaner FFIs, would be excellent for tuples and records, etc. At _some_ point we'll decide to leave some engines behind, likely the moment Wasm GC hits stage 5 😄

tlively:
Interesting! Yes, I can see that multivalue would be useful for all those things. I was working about a year and a half ago to improve our multivalue codegen by transforming Binaryen IR into a stack machine format called Poppy IR and performing stack-machine-specific optimizations on it that can't be expressed in Binaryen IR's AST format. The tracking issue is here: https://github.com/WebAssembly/binaryen/issues/3059.

Unfortunately, I never finished the project because it wasn't very important for the things we were working on at the time. In particular, clang is not in a great position to use multivalue in C/C++ and we estimated that multivalue optimizations in Binaryen would only decrease code size by about 1.5%. If you start using Binaryen's existing multivalue support heavily, it would probably be worth picking that project back up.

PR #4528, opened 2022-03-15
Fix errors when building in C++20 mode

kubasz:
C++:
 * use [[noreturn]] available since C++11 instead of compiler-specific attributes
 * replace deprecated std::is_pod with is_trivial&&is_standard_layout (also available since C++11/14)
 * explicitly capture this in [=] lambdas
 * extra const functions in FeatureSet, fix ambiguous operator& call error

CMake:
 * Use CMAKE_CXX_STANDARD set in the directory scope to set the default CXX_STANDARD for all targets
 * The above made it unnecessary for the -std=c++17 flag to be passed directly in compiler options
 * Explicitly disable vendor c++ standard extensions (GNU/MSVC) in directory scope as they are on by default in CMake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,8 +13,11 @@ include(GNUInstallDirs)
 # The C++ standard whose features are required to build Binaryen.
 # Keep in sync with scripts/test/shared.py cxx_standard
 # The if condition allows embedding in a project with a higher default C++ standard set
-if((NOT CMAKE_CXX_STANDARD) OR (CMAKE_CXX_STANDARD LESS 17))
-  set(CMAKE_CXX_STANDARD 17)
+set(REQUIRED_CXX_STANDARD 17)
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD ${REQUIRED_CXX_STANDARD})
+elseif(CMAKE_CXX_STANDARD LESS ${REQUIRED_CXX_STANDARD})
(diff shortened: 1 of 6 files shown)


tlively (approved):
This is great, thanks!

CMakeLists.txt
```diff
@@ -12,7 +12,12 @@ include(GNUInstallDirs)
 
 # The C++ standard whose features are required to build Binaryen.
 # Keep in sync with scripts/test/shared.py cxx_standard
-set(CXX_STANDARD 17)
+# The if condition allows embedding in a project with a higher default C++ standard set
+if((NOT CMAKE_CXX_STANDARD) OR (CMAKE_CXX_STANDARD LESS 17))
```
sbc100, line 16: If somebody asked for less that 17 maybe it would be nicer to error out rather than slightly ignore?

sbc100 (commented):
What is wrong with sticking with ` std::is_pod` given its more readable? Is it deprecated perhaps?

kubasz:
> What is wrong with sticking with ` std::is_pod` given its more readable? Is it deprecated perhaps?

It has been deprecated in C++20 with this paper: <http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0767r1.html>

src/tools/tool-options.h
```diff
@@ -155,7 +155,7 @@ struct ToolOptions : public Options {
            std::string("Enable ") + description,
            ToolOptionsCategory,
            Arguments::Zero,
-           [=, this](Options*, const std::string&) {
```
kubasz: C++17 only supports `[=]` while C++20 throws a warning that `[=, this]` should be used for clarity since the this-object could also be captured by-value. As there is only one capture + `this` in this lambda anyway, I switched it to an explicit list that won't warn in either mode

sbc100 (approved):

PR #4529, opened 2022-03-15
Add support for extended-const proposal

sbc100:


sbc100:
I'm not sure where the best place to add tests for this would be..  so this change still lacks tests.  But I welcome feedback on the implemenation.

sbc100:
Added test

src/wasm/wasm-validator.cpp
```diff
@@ -2767,12 +2767,18 @@ void FunctionValidator::visitFunction(Function* curr) {
...
+  if (!Properties::isValidInConstantExpression(curr, features)) {
+    return false;
+  }
   if (curr->is<GlobalGet>()) {
     return true;
   }
   auto* c = curr->dynCast<Const>();
   if (!c) {
+    if (features.hasExtendedConst()) {
+      return true;
```
kripken, line 2780: Shouldn't this check for Add etc. and not accept anything that is not a const? That could be a TODO though for now.

kripken (approved):
lgtm % comment.

Test looks sufficient to me!

src/wasm/wasm-validator.cpp
sbc100: isValidInConstantExpression already take care of check only limited instructions are accepted.

This code already gives up if it finds a `global.get` .. so I think its reasonable to also give up in the face of other non-const things (in theory the global.get could result in a const but we don't try to handle this case here).

I'll add a TODO.

sbc100: Added comment.

sbc100: I also restructured the code a bit to make to more clear what is going on here.

src/wasm/wasm-validator.cpp
```diff
@@ -2767,13 +2767,21 @@ void FunctionValidator::visitFunction(Function* curr) {
...
+                               Address max,
+                               FeatureSet features) {
+  if (!Properties::isValidInConstantExpression(curr, features)) {
+    return false;
   }
   auto* c = curr->dynCast<Const>();
   if (!c) {
-    return false;
+    // Unless the instruction is actually a const instruction, we don't
+    // currently try to evaluate it.
```
kripken, line 2780:
```suggestion
    // currently try to validate it.
```

src/wasm/wasm-validator.cpp
```diff
@@ -2767,13 +2767,21 @@ void FunctionValidator::visitFunction(Function* curr) {
...
+    // TODO: Attempt to evaluate other expressions that might also be const
```
kripken, line 2781:
```suggestion
    // TODO: Attempt to validate other expressions that might also be const
```

kripken (approved):

src/wasm/wasm-validator.cpp
sbc100: Actually the validation is done already.. this part of the code is trying to evaluate to a absolute constant.  Does that make sense?

sbc100: To clarify, the validation of the instructions is done in isValidInConstantExpression.. this part is trying to evaluate the expression to get an absolute offset so that we can bounds check it.    This not always possible... for example if the expression is a global.get we currently just skip the bounds check and assume is not possible to evaluate.

kripken: Oh, right... yes, ignore my comments then. lgtm as is.

kripken:
Not sure about CI, but maybe the force-push confused it somehow?

sbc100:
Looks like github actions is having issues:  https://www.githubstatus.com/

Issue #4530, opened 2022-03-16
[memory64] Optimization failures on memory.size


backes:
wasm-opt fails to process this code with `--enable-memory64 -O3`:
```
(module
 (type $i32_i64_=>_i32 (func (param i32 i64) (result i32)))
 (type $none_=>_none (func))
 (memory $0 i64 256 256)
 (export "main" (func $0))
 (func $0 (param $0 i32) (param $1 i64) (result i32)
  (call $1)
  (i32.const 0)
 )
 (func $1
  (if
   (i32.wrap_i64
    (i64.shl
     (memory.size)
     (i64.const 16)
    )
   )
   (unreachable)
  )
 )
)
```

The error is:
```
[wasm-validator error in function 0] i32 != i64: binary child types must be equal, on 
(i64.shl
 (memory.size)
 (i64.const 16)
)
[wasm-validator error in function 0] i32 != i64: i64 op, on 
(i64.shl
 (memory.size)
 (i64.const 16)
)
[wasm-validator error in function 0] i32 != i64: wrap type must be correct, on 
(i32.wrap_i64
 (i64.shl
  (memory.size)
  (i64.const 16)
 )
)
Fatal: error after opts
```

backes:
Seems like this is the fix:
```
diff --git a/src/wasm-delegations-fields.def b/src/wasm-delegations-fields.def
index 3d7504c60..422bb521c 100644
--- a/src/wasm-delegations-fields.def
+++ b/src/wasm-delegations-fields.def
@@ -461,12 +461,16 @@ switch (DELEGATE_ID) {
   }
   case Expression::Id::MemorySizeId: {
     DELEGATE_START(MemorySize);
+    DELEGATE_FIELD_TYPE(MemorySize, type);
+    DELEGATE_FIELD_TYPE(MemorySize, ptrType);
     DELEGATE_END(MemorySize);
     break;
   }
   case Expression::Id::MemoryGrowId: {
     DELEGATE_START(MemoryGrow);
+    DELEGATE_FIELD_TYPE(MemoryGrow, type);
     DELEGATE_FIELD_CHILD(MemoryGrow, delta);
+    DELEGATE_FIELD_TYPE(MemoryGrow, ptrType);
     DELEGATE_END(MemoryGrow);
     break;
   }
```

tlively:
Looks like this was fixed.

PR #4531, opened 2022-03-16
[memory64] Keep type of memory.size and memory.grow on copy

backes:
When copying a MemorySize or MemoryGrow instruction (e.g. for inlining),
transfer the memory type also to the copy. Otherwise it will always be
i32, even if memory64 should be used.

This fixes issue #4530.


backes:
Let me know if this should have a regression test somewhere.

src/wasm-delegations-fields.def
```diff
@@ -461,12 +461,16 @@ switch (DELEGATE_ID) {
   }
   case Expression::Id::MemorySizeId: {
     DELEGATE_START(MemorySize);
+    DELEGATE_FIELD_TYPE(MemorySize, type);
```
kripken, line 464: `type` is not needed in both cases, as it is present in all these classes and handled automatically.

kripken (commented):
Thanks for the PR!

For a test you can use this, in `test/lit/passes/inlining_memory64.wat`:
```wat

;; RUN: foreach %s %t wasm-opt --inlining --enable-memory64 -S -o - | filecheck %s

(module
  ;; CHECK:      (type $none_=>_i64 (func (result i64)))

  ;; CHECK:      (memory $0 i64 256 256)
  (memory $0 i64 256 256)

  ;; CHECK:      (func $call-size (result i64)
  ;; CHECK-NEXT:  (block $__inlined_func$size (result i64)
  ;; CHECK-NEXT:   (memory.size)
  ;; CHECK-NEXT:  )
  ;; CHECK-NEXT: )
  (func $call-size (result i64)
    (call $size)
  )

  (func $size (result i64)
    ;; Upon inlining this code is copied, and as the memory is 64-bit the copied
    ;; instruction must remain 64-bit (and not have the default 32-bit size). If
    ;; we get that wrong then validation would fail here.
    (memory.size)
  )

  ;; CHECK:      (func $call-grow (result i64)
  ;; CHECK-NEXT:  (block $__inlined_func$grow (result i64)
  ;; CHECK-NEXT:   (memory.grow
  ;; CHECK-NEXT:    (i64.const 1)
  ;; CHECK-NEXT:   )
  ;; CHECK-NEXT:  )
  ;; CHECK-NEXT: )
  (func $call-grow (result i64)
    ;; As above, but for grow instead of size.
    (call $grow)
  )

  (func $grow (result i64)
    (memory.grow
      (i64.const 1)
    )
  )
)
```

src/wasm-delegations-fields.def
backes: Great, thanks!

backes:
Thanks for providing the test! I remember that syntax from my LLVM days :)
Can you take another look please?

src/wasm-delegations-fields.def
```diff
@@ -461,12 +461,14 @@ switch (DELEGATE_ID) {
...
   case Expression::Id::MemorySizeId: {
     DELEGATE_START(MemorySize);
+    DELEGATE_FIELD_TYPE(MemorySize, ptrType);
     DELEGATE_END(MemorySize);
     break;
   }
   case Expression::Id::MemoryGrowId: {
     DELEGATE_START(MemoryGrow);
     DELEGATE_FIELD_CHILD(MemoryGrow, delta);
+    DELEGATE_FIELD_TYPE(MemoryGrow, ptrType);
```
kripken, line 471: The order here should be flipped (as the stack traversal pushes first), that is, the new line should be before `delta`.

kripken, line 471:
```suggestion
    DELEGATE_FIELD_TYPE(MemoryGrow, ptrType);
    DELEGATE_FIELD_CHILD(MemoryGrow, delta);
```

kripken: Oh, I guess github allows the reviewer to apply stuff like this, so I'll just do that.

kripken (approved):

PR #4534, opened 2022-03-21
Fix merge-similar-functions test expections

kripken:
The tests in #4414 have different results than expected when I run
the auto updater script on them. In this PR is what I get locally when
I run that script.

This seems wrong, as when running the test manually - for example,
running `-all -merge-similar-functions` on the first module in
`test/lit/passes/merge-similar-functions.wast` - then I do see the
function in the output. The function has to be there, in fact, as it
is called by others after the pass runs, so it would not validate without
it. However, for some reason the auto update script removes it.
@tlively any idea?


tlively:
Looks like it's because that function has a synthesized name that doesn't correspond to any of the input functions. The fix would be to use `update_lit_checks.py --all-items`. Updating the comment at the top of the test to add that flag and rerunning the auto update script should fix it.

kripken:
Thanks! I forgot about that flag...

PR updated to fix this.

tlively (approved):
The formatting is unfortunate because it makes the merged function look like it comes from the following module, but at least it is included in the test now.

PR #4536, opened 2022-03-21
Add BUILD_TESTS CMake option

tlively:
Turning it off removes the build dependency on the third-party googletest
library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,7 +320,7 @@ cmake . && make
 
 A C++17 compiler is required. Note that you can also use `ninja` as your generator: `cmake -G Ninja . && ninja`.
 
-To avoid the gtest dependency, you can pass `-DBUILD_TESTS=NO` to cmake.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -318,7 +318,9 @@ After that you can build with CMake:
 cmake . && make
 ```
 
-A C++14 compiler is required. Note that you can also use `ninja` as your generator: `cmake -G Ninja . && ninja`.
+A C++17 compiler is required. Note that you can also use `ninja` as your generator: `cmake -G Ninja . && ninja`.
+
+To avoid the gtest dependency, you can pass `-DBUILD_TESTS=NO` to cmake.
```
sbc100, line 323: `-DBUILD_TESTS=OFF` maybe?

truboxl:
I think https://github.com/WebAssembly/binaryen/blob/main/third_party/CMakeLists.txt also needs to be guarded

tlively:
Ah yeah, thanks for the catch @truboxl! https://github.com/WebAssembly/binaryen/pull/4552

PR #4538, opened 2022-03-21
Run auto-update script on CI to check for uncommited changes

kripken:
This would have caught #4534 early.

This just runs the `lit` tests atm. Those are the ones we update the most
so I think it has the most value, and it's a fast suite since it's parallel.
Running all the suites would give more coverage, but would almost
double our build times which I'm not sure is worth it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,7 +108,7 @@ jobs:
     # XXX  run: python check.py --binaryen-bin=out/bin
 
     - name: verify auto-updating causes no changes
-      run: ./auto_update_tests.py lit | git diff test/
+      run: ./auto_update_tests.py lit | git diff -- test/
       if: matrix.os == 'ubuntu-latest'
 
   build-clang:


.github/workflows/ci.yml
```diff
@@ -104,8 +104,12 @@ jobs:
     - name: test binaryen-lit
       run: python out/bin/binaryen-lit test/lit/parse-error.wast
 
-    - name: test
-      run: python check.py --binaryen-bin=out/bin
+    # XXX disable momentarily to verify the below stage fails - name: test
+    # XXX  run: python check.py --binaryen-bin=out/bin
+
+    - name: verify auto-updating causes no changes
+      run: ./auto_update_tests.py lit | git diff test/
```
tlively, line 111: I'm not sure if `test/` could ever be parsed as something other than a directory (e.g. as some git ref), but adding a `--` will ensure that it is always parsed as a file path.
```suggestion
      run: ./auto_update_tests.py lit | git diff -- test/
```

tlively (approved):

.github/workflows/ci.yml
```diff
@@ -104,8 +104,12 @@ jobs:
...
+      if: matrix.os == 'ubuntu-latest'
```
aheejin, line 112: Why only on ubuntu?

aheejin (approved):

.github/workflows/ci.yml
kripken: This is to prevent it from trying to run on windows, I think. We have the same for other similar scripts.

PR #4542, opened 2022-03-22
[NFC] Refactor DeadArgumentElimination code to use LocalGraph

kripken:
I believe this old code was written before we had LocalGraph. It does a flow
analysis to see if the values arriving in parameters are actually used, but
there is no reason to do it manually since we have LocalGraph now which
makes that trivial, as it matches the values for each get, so we can just check
if any get can read the param's incoming value.


tlively (approved):
Nice simplification!

PR #4543, opened 2022-03-22
Pass v8 flag for extended-const

kripken:
This flag is needed now that we have testcases using extended-const
in our test suite, as the fuzzer will grab random testcases, modify them,
and run them through v8.


sbc100 (approved):

sbc100:
> Oops, this wasn't enough. The fuzzer can still fail on this, because there is a possibility of the wasm trapping during startup if the base is not in bounds. I added an ignore-list to just ignore this new testcase for now, how does that sound?

I know nothing about how the fuzzer works so don't really have an opinion on this.

kripken:
Ok, let's check with @aheejin and @tlively then - does an ignore-list for the new extended-const testcase sound reasonable? (see comments in the code in this PR for more details)

tlively:
Can't these traps on startup already happen if there is a start function or if an active data section is initialized out of bounds? How do we handle those cases?

kripken:
I think we just don't handle them, and it became much more likely with this new testcase. (The fuzzer will initialize all imported globals and params etc. to 0, so segments ended up in bounds due to that I think in the past.)

Handling them is not that trivial. It's basically a randomly generated wasm that fails if you do `wasm-opt --fuzz-exec` on it. We'd need to detect which errors are ok to skip in that case, and which aren't. But it's also bad if we end up skipping a lot of things because of this, so silent skipping isn't great either. For now I'm hoping just ignoring this one testcase is enough.

aheejin:
Why is trapping during startup function different from trapping in running and why only the former is the problem?

sbc100:
Perhaps we could update the test so that it always evaluates to in-bouds/const.

e.g.`(i32.sub (i32.add (global.get 1) (i32.const 42)) (global.get 1))` is always 42

kripken:
> Why is trapping during startup function different from trapping in running and why only the former is the problem?

Hmm, good question about the startup function. So this issue is not just for segment initialization. I'm actually not sure why we don't see errors from that. Perhaps we don't have any testcases with nontrivial start functions? But the fuzzer would modify them, so that's not it...

But ignoring the startup function, a trap during initialization breaks the fuzzer's assumption that we start from a valid wasm. It assumes a valid wasm, then runs some passes on it, then sees if it is still valid after that - if not, the passes are blamed. So we have to avoid wasms that are invalid to begin with.

sbc100:
> > Why is trapping during startup function different from trapping in running and why only the former is the problem?
> 
> Hmm, good question about the startup function. So this issue is not just for segment initialization. I'm actually not sure why we don't see errors from that. Perhaps we don't have any testcases with nontrivial start functions? But the fuzzer would modify them, so that's not it...
> 
> But ignoring the startup function, a trap during initialization breaks the fuzzer's assumption that we start from a valid wasm. It assumes a valid wasm, then runs some passes on it, then sees if it is still valid after that - if not, the passes are blamed. So we have to avoid wasms that are invalid to begin with.

Start function or const expressions that trap with out-of-bounds memory accesses are not invalid in that they do pass validation and are valid module,  they just trap on startup, right?

aheejin:
Yeah I have the same question w/ @sbc100.. They are not invalid, they just trap during initialization... Apparently trapping in normal functions is fine right? We just make sure the behavior is preserved, so if the starting wasm traps, we have to make sure (unless options that ignore traps are given) the optimized wasm also traps. Can't we do the same for wasm that traps during startup/initialization?

kripken:
We do allow traps during function execution. We then check that after we run passes we see the same traps as before. The difference here is that it traps during startup, not in a function, which is handled elsewhere. But, you're both right that we should probably find a way to refactor the test runner to handle that case too...

kripken:
Looking into this more carefully after the above comments, I think perhaps the issue just happens in the asyncify handler. I don't seem to see errors outside of it. What I think happens is it is the only TestcaseHandler that runs d8 without first running the binaryen interpreter. I updated this PR to add a proper skip there, PTAL.

kripken:
Oh, actually this is not enough, nm... we need to actually check what causes the error. It must be a problem during startup, but we can ignore traps later (maybe the program is supposed to trap). This probably requires some new fuzzer feature. Or maybe someone else has a better idea?

PR #4544, opened 2022-03-23
[NFC] Move and generalize parameter-removing logic from DeadArgumentElimination

kripken:
In preparation for removing dead arguments from all functions sharing a heap
type (which seems useful for j2wasm output), first this PR refactors that code
so it is reusable. This moves the code out of the pass into FunctionUtils, and
also generalizes it slightly by

* supporting a set of functions and not just a single one, and
* receiving a list of call_refs and not just calls

(no other changes to anything).

### DIFF
--- a/src/ir/function-utils.h
+++ b/src/ir/function-utils.h
@@ -104,11 +104,11 @@ inline bool removeParameter(const std::vector<Function*> funcs,
                             PassRunner* runner) {
   assert(funcs.size() > 0);
   auto* first = funcs[0];
-#ifndef NDEBUG
+//#ifndef NDEBUG
   for (auto* func : funcs) {
     assert(func->type == first->type);
   }
-#endif
+//#endif
 
   // Great, it's not used. Check if none of the calls has a param with
   // side effects that we cannot remove (as if we can remove them, we
@@ -134,7 +134,7 @@ inline bool removeParameter(const std::vector<Function*> funcs,
   }
 
   // We can do it!
-
+//std::cout << "remove one\n";
   // Remove the parameter from the function. We must add a new local
   // for uses of the parameter, but cannot make it use the same index
   // (in general).
@@ -144,6 +144,7 @@ inline bool removeParameter(const std::vector<Function*> funcs,
   params.erase(params.begin() + index);
   // TODO: parallelize some of these loops?
   for (auto* func : funcs) {
+  //std::cout << " in func " << func->name << '\n';
     func->setParams(Type(params));
 
     // It's cumbersome to adjust local names - TODO don't clear them?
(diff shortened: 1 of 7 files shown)


src/ir/function-utils.h
```diff
@@ -43,6 +46,188 @@ inline bool equal(Function* left, Function* right) {
...
+// ensure to pass in all relevant calls and call_refs).
+//
+// This does not check if removing the parameter would change the semantics
+// (say, if the parameter's value is used), which the caller is assumed to do.
+//
+// This assumes that the set of functions all have the same signature. The main
+// use cases are either to send a single function, or to send a set of functions
+// that all have the same heap type (and so if they all do not use some
+// parameter, it can be removed from them all).
+inline bool removeParameter(const std::vector<Function*> funcs,
```
tlively, line 99: There are an awful lot of preconditions that have to be met to use this function correctly. I wonder if there is some more encapsulated abstraction that would work as well.

tlively (commented):
Can you move the function definitions into a .cpp file?

src/ir/function-utils.h
kripken: Maybe the comment isn't clear enough - I think it's not that complex? Or maybe I don't understand your concern.

Another way to phrase the comment might be: This function assumes you have already checked for "logical" correctness. It then simply removes the parameter, and if you made a mistake then you would be breaking the wasm. The function does check for "hard constraints" and returns false if it hits them, such constraints are things that basically would make the wasm not validate if applied, or otherwise be obviously wrong.

tlively: The preconditions:

 1. All calls and call_refs must be passed in.
 2. Removing the parameter should not change semantics.
 3. All functions must have the same signature.
 
I guess this isn't as bad as I initially thought, but this is more preconditions than I would expect from a general-purpose utility that's supposed to be useful in many potential situations. I guess it's more of a special-purpose utility that we only expect to ever call in a couple places.

Why can't we collapse the call sites into a single DAE optimization? It seems unfortunate to have such similar optimizations that share so much code but are still separate.

kripken: Thanks, I see what you mean better now.

I think this is somewhat specialized and not fully general purpose, yes. But it is a way to avoid duplication between the existing DAE which optimizes single functions, and a future signature-pruning pass which does the same logical operation but on entire heap types at once. With this shared code split out, that new pass will be very simple.

We could in principle put all that code into a single file, DAE.cpp, and then these methods would go in there. Then it would be obvious it's not truly general purpose. I'd still prefer to have a new file SignaturePruning.cpp, however, just to keep that pass nice and separate and simple.

One option that might achieve both goals is to put this code in a separate file, but not in a "general" place like `ir/function-utils.h`? It could perhaps be in a file `remove-parameters.cpp` which would live alongside `DAE.cpp` and `SignaturePruning.cpp`?

tlively: `remove-parameters.cpp` sounds good to me!

tlively (approved):
LGTM with the code living in a more specialized location.

kripken:
I named it `passes/param-utils` as it also has the code to check if a param is used, and not just remove them. I also added a comment about how tied it is to the opt passes that use it.

src/passes/param-utils.cpp
```diff
@@ -0,0 +1,177 @@
...
+    }
+  }
+
+  // Remove the arguments from the calls.
+  for (auto* call : calls) {
+    call->operands.erase(call->operands.begin() + index);
+  }
+  for (auto* call : callRefs) {
+    call->operands.erase(call->operands.begin() + index);
+  }
```
aheejin, line 135: This doesn't seem to update `call_ref`s' target types. Is that OK?

kripken: Ah, good point. Yes, this is intentional (the caller is expected to update types in all the places, its easier if a single place does it), but it should be documented. I'll open a PR.

PR #4549, opened 2022-03-25
Generalize PossibleConstantValues for immutable globals

kripken:
This moves more logic from ConstantFieldPropagation into PossibleConstantValues,
that is, instead of handling the two cases of a Literal or a Name before calling
PossibleConstantValues, move that code into the helper class. That way all users of
PossibleConstantValues can benefit from it. In particular, this makes
DeadArgumentElimination now support optimizing immutable globals, as well as
`ref.func` and `ref.null`.

(Changes to `test/lit/passes/dae-gc-refine-params.wast` are to avoid the new
optimizations from kicking in, so that it still tests what it tested before.)


test/lit/passes/dae_all-features.wast
```diff
@@ -538,3 +538,115 @@
...
+;; constant value.
+(module
+ ;; CHECK:      (type $none_=>_none (func))
+
+ ;; CHECK:      (type $i32_=>_none (func (param i32)))
+
+ ;; CHECK:      (type $i32_i32_=>_none (func (param i32 i32)))
+
+ ;; CHECK:      (global $imm i32 (i32.const 42))
+ (global $imm i32 (i32.const 42))
```
aheejin, line 552: Super nit: How about `immut` instead of `imm` in this file? `immut`/`mut` feels more intuitive and `imm` is often used to denote immdiates elsewhere in Wasm spec, so..

aheejin (approved):

PR #4551, opened 2022-03-28
[Wasm GC] Signature Pruning: Remove params passed constant values

kripken:
This basically just adds a call to `ParamUtils::applyConstantValues`, however,
we also need to be careful to not optimize in the presence of imports or
exports, so this adds a boolean that indicates unoptimizability.


src/passes/SignaturePruning.cpp
```diff
@@ -116,32 +116,53 @@ struct SignaturePruning : public Pass {
...
+      }
+
+      if (usedParams.size() == numParams) {
         // All parameters are used, give up on this one.
         continue;
       }
 
-      // We found possible work! Find the specific params that are unused try to
+      // We found possible work! Find the specific params that are unused&try to
       // prune them.
```
aheejin, line 162:
```suggestion
      // We found possible work! Find the specific params that are unused & try
      // to prune them.
```

aheejin (approved):

PR #4553, opened 2022-03-29
[Wasm GC] GlobalTypeOptimization: Remove fields from end based on subtypes

kripken:
Previously we'd remove a field from a type if that field has no uses in any
sub- or super-type. In that case we'd remove it from all the types at once.
However, there is a case where we can remove a field only from a parent
but not from its children, if the field is at the end: if `A` has fields `{x, y, z}`
and its subtype `B` has fields `{x, y, z, w}`, and `A` pointers only access
field `y` while `B` pointers access all the fields, then we can remove `z`
from `A`. Removing it from the end is safe, and then `B` will not only add
`w` as it did before but also add `z`. Note that we cannot remove `x`,
because it is not at the end: removing it from just `A` but not `B` would
shift the indexes, making them incompatible.

### DIFF
--- a/src/passes/GlobalTypeOptimization.cpp
+++ b/src/passes/GlobalTypeOptimization.cpp
@@ -158,7 +158,7 @@ struct GlobalTypeOptimization : public Pass {
     auto subSupers = combinedSetGetInfos;
     propagator.propagateToSuperAndSubTypes(subSupers);
     auto subs = std::move(combinedSetGetInfos);
-    propagator.propagateToSuperAndSubTypes(subs);
+    propagator.propagateToSubTypes(subs);
 
     // Process the propagated info.
     for (auto type : propagator.subTypes.types) {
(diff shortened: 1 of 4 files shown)


src/passes/GlobalTypeOptimization.cpp
```diff
@@ -175,16 +189,39 @@ struct GlobalTypeOptimization : public Pass {
...
+          removableIndexes.insert(i);
+        }
+      }
+      for (int i = int(fields.size()) - 1; i >= 0; i--) {
+        // Unlike above, a write would stop us here: above we propagated to both
+        // sub- and super-types, which means if we see no reads then there is no
+        // possible read of the data at all. But here we just propagated to
+        // subtypes, and so we need to care about the case where the parent
+        // writes to a field but does not read from it - we still need those
+        // writes to happen as children may read them.
```
tlively, line 210: Can't we look at the children and determine precisely whether or not there is a read in a child? I don't see why we need to be conservative here.

kripken: If children do not read this, then that means nothing reads this at all, and it would be handled in the other case anyhow. I'll add to the comment.

kripken:
Yes, that is what `propagator.propagateToSubTypes(..)` handles: we propagate reads to subtypes. If we see a read on A, then it could be a B or a C that is pointed to by that A reference, so we have to consider them possible reads of B and C as well.

I added a test for that now, and also a variation that puts the read on one of the children, showing we can optimize the parent and the other child properly.

tlively (approved):
Thanks for the explanations!

PR #4555, opened 2022-03-30
[Wasm GC] Fix non-nullable tuples

### LINKED ISSUE #4554: Assert with `--enable-gc-nn-locals`
If I try to optimize e.g. [this module](https://github.com/WebAssembly/binaryen/files/8382268/m63.wasm.gz) with `wasm-opt -g -all -O3 -tnh --nominal --enable-gc-nn-locals`, I hit this assert:
```
wasm-opt: /usr/local/google/home/askesc/checkout/binaryen/src/wasm-builder.h:1058: static wasm::Index wasm::Builder::addVar(wasm::Function*, wasm::Name, wasm::Type): Assertion `type.isConcrete()' failed.
```
Without `--enable-gc-nn-locals` it works fine. It seems to do the same both with and without `--nominal`, and whether or not the input module contains any non-nullable locals.

kripken:
Fixes #4554

### DIFF
--- a/src/tools/wasm-shell.cpp
+++ b/src/tools/wasm-shell.cpp
@@ -134,7 +134,7 @@ protected:
     lastModule = module->name;
     builders[moduleName] = builder;
     modules[moduleName].swap(module);
-    modules[moduleName]->features = FeatureSet::All;
+    modules[moduleName]->features.setAll();
     bool valid = WasmValidator().validate(*modules[moduleName]);
     if (!valid) {
       std::cout << *modules[moduleName] << '\n';
@@ -237,7 +237,7 @@ protected:
 
   void parseModuleAssertion(Element& s) {
     Module wasm;
-    wasm.features = FeatureSet::All;
+    wasm.features.setAll();
     std::unique_ptr<SExpressionWasmBuilder> builder;
     auto id = s[0]->str();
 
@@ -358,7 +358,7 @@ protected:
       "memory", spectest->memory.name, ExternalKind::Memory));
 
     modules["spectest"].swap(spectest);
-    modules["spectest"]->features = FeatureSet::All;
+    modules["spectest"]->features.setAll();
     instantiate(modules["spectest"].get());
     linkedInstances["spectest"] = instances["spectest"];
     // print_* functions are handled separately, no need to define here.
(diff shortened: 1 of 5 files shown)


tlively (commented):
Shouldn't we count RTTs as non-nullable for these purposes? I would be very surprised if we supported non-defaultable ref types but not non-defaultable rtt types.

kripken:
Fixing this also forces us to fix an existing bug in the spec test suite: `FeatureSet::All` is an internal bitmask of all features including nn-locals. wasm-shell should not use that, and instead use the `.setAll()` API like wasm-opt etc. do, which handles nn-locals in a special way (disabled unless explicitly opted into).

kripken:
> Shouldn't we count RTTs as non-nullable for these purposes?

I don't think so, on. RTTs are neither nullable nor non-nullable - they are not references at all. (In the spec we implement, which is before upstream changes since then that I'm not sure have been finalized or discussed fully yet.)

This PR does not change that - it just adds a helper function for convenience. It's always been the case that nn-locals allows non-nullable references but not RTTs.

kripken:
> Can we make FeatureSet::All an internal implementation detail of the API to avoid that failure mode?

I agree this API atm is not great. Yes, that seems reasonable to consider, but I'd want to find the relevant old discussion and see why we didn't do it then.

> The reason I'm confused about this is that defaultability should be the only property that matters for locals. Reference types shouldn't be special among non-defaultable types

I thought that's what the point of the non-nullable locals flag has always been. That's why it's not called non-defaultable, for example. It's possible I've misunderstood this all along :smile: 

> The reason this isn't a problem in the upstream proposal repo is that RTTs have been turned into reference types there.

Yes - has that been fully discussed? I remember it landed without a full discussion.

In any case, as the status of RTTs in the MVP is still unclear, perhaps it's best to keep them as they are for now? Again, this PR changes nothing about them. We were already doing the right (AFAIK) behavior for single values, and this PR just makes us apply that behavior to tuples so we do the same everywhere.

tlively (approved):
Bug fix lgtm

tlively:
> > The reason this isn't a problem in the upstream proposal repo is that RTTs have been turned into reference types there.
> 
> Yes - has that been fully discussed? I remember it landed without a full discussion.

There was some more recent discussion about this here: https://github.com/WebAssembly/gc/issues/276. I'm not sure many people saw the discussion, but the conclusion was that it makes sense for RTTs to be reference types.

PR #4556, opened 2022-03-30
[NFC][MemoryPacking] Avoid unnecessarily enormous memory allocations

tlively:
The memory packing pass previously used vectors with a slot for each data
segment to easily map segment indices to lists of "referrers," i.e. bulk memory
expressions that referred to the segments. The parallel analysis pass to collect
these referrers allocated one of these vectors for each function in a module.
Unfortunately, for a particularly large module with over 200k functions and over
75k data segments, this resulted in hundreds of gigabytes of memory allocations.
The vast majority of functions contain no referrers, so using a
std::unordered_map makes much more sense than using a vector to perform this
mapping.

### DIFF
--- a/src/passes/MemoryPacking.cpp
+++ b/src/passes/MemoryPacking.cpp
@@ -65,8 +65,11 @@ using Replacement = std::function<Expression*(Function*)>;
 // Maps each bulk memory op to the replacement that must be applied to it.
 using Replacements = std::unordered_map<Expression*, Replacement>;
 
-// A collection of bulk memory operations referring to a particular segment
+// A collection of bulk memory operations referring to a particular segment.
 using Referrers = std::vector<Expression*>;
+
+// Map segment indices to referrers.
+using ReferrersMap = std::unordered_map<Index, Referrers>;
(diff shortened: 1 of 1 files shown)


src/passes/MemoryPacking.cpp
```diff
@@ -459,48 +461,53 @@ void MemoryPacking::getSegmentReferrers(Module* module,
...
       }
     } else {
+      // Active segment.
       used = true;
     }
     if (used) {
-      usedSegments.push_back(segments[i]);
-      usedReferrers.push_back(referrers[i]);
-    } else {
+      usedSegments.push_back(std::move(segments[i]));
```
kripken, line 504: This modifies an input parameter which seems slightly surprising. Maybe worth mentioning in a comment on the definition?

kripken (approved):

src/passes/MemoryPacking.cpp
kripken: Oh, I see there is a `swap` below, so this always modified it...

PR #4557, opened 2022-03-31
[NFC] Refactor Feature::All to match FeatureSet.setAll()

kripken:
As we recently noted in #4555, that `Feature::All` and `FeatureSet.setAll()`
are different is potentially confusing...

I think the best thing is to make them identical. This does that, and adds a
new `Feature::AllPossible` which is everything possible and not just the
set of all features that are enabled by `-all`.

This undoes part of #4555 as now the old/simpler code works properly.


tlively (approved):
Great, thanks!

PR #4563, opened 2022-03-31
[Wasm GC] Fix non-nullable tuple locals in CoalesceLocals

### LINKED ISSUE #4562: Assert with `--enable-gc-nn-locals`, take 3
After the fix in https://github.com/WebAssembly/binaryen/pull/4561, the case in https://github.com/WebAssembly/binaryen/issues/4554 now hits a different assert:
```
wasm-opt: /usr/local/google/home/askesc/checkout/binaryen/src/literal.h:252: static wasm::Literal wasm::Literal::makeNull(wasm::Type): Assertion `type.isNullable()' failed.
```

kripken:
That pass calls `makeZeros` and it only checked if the type was itself
a non-nullable reference, as opposed to a tuple that contains such a
thing. I added a new helper function for that. It's possible we should use
it in more places, but that would require further investigation.

Fixes #4562


test/lit/passes/coalesce-locals-gc-nn.wast
```diff
@@ -0,0 +1,84 @@
...
+ ;; CHECK-NEXT:   (tuple.extract 1
+ ;; CHECK-NEXT:    (local.get $2)
+ ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT: )
+ (func $nn-locals (param $any (ref any))
+  ;; When computing interferences, coalesce locals should not error on tuples
+  ;; that contain non-nullable locals.
+  (local $x ((ref any) (ref any)))
+  (local $y ((ref any) (ref any)))
```
aheejin, line 46: These two locals are set and used in exactly the same way... Aren't they supposed to be "coalesced" by `CoalesceLocals`?

aheejin: Then how about adding a comment in this file that this does not test coalescing tuples atm (despite the file name) but only tests it does not crash?

aheejin (approved):

test/lit/passes/coalesce-locals-gc-nn.wast
aheejin: On sorry, line 48 already has it... Nevermind 😅

kripken:
Actually I think there is a better solution than this PR, sorry for the noise. Will open replacement shortly.

PR #4567, opened 2022-04-01
Lift the restriction in liveness-traversal.h that supported max 65535 locals in a function.

juj:
Lift the restriction in liveness-traversal.h that supported max 65535 locals in a function.

A customer of Unity is compiling some massive C# libraries, and are hitting this limit.

~~Yes, this will result in massive 4GB+ memory usage, but I am not overly concerned about that. Previously they'd crash, after this change they'll be able to go and buy some new RAM sticks.~~

Edit: Added a sparse_square_matrix container that intelligently handles sparse or dense memory backing, so memory usage will be minimized (currently using 8k*8k = 64MB as a limit to switch to a sparse backing).

Existing users with functions with small number of locals won't be impacted/regressed by this change: `std::vector::operator []` index type is already 64-bit if `size_t` is 64-bit, so performance implication is the same before and after this PR.

Also a small drive-by micro-optimization to avoid double zero filling the arrays when they grow.


kripken:
To make this change we'd need to also figure out a solution for the fuzzer (that is the reason the limit was added in the first place, also maybe there was a spec test). The problem is that a fuzz testcase might crash due to failing to allocate here, which would look like the fuzzer found a bug. I'm not sure offhand how to ignore that situation.

But taking a step back, even a huge real-world module should not hit this, I hope, as the optimization pipeline will run `simplify-locals, reorder-locals` before running any liveness analyses. That combination will generally make the amount of locals very reasonable - it is why we've never seen this assertion fire on real-world code before. So I think it's worth investigating why this limit is being hit. Can you provide that wasm perhaps?

juj:
I could alternatively write the code to use `std::unordered_map<std::pair<Index, Index>, u8>`, but I don't have the possibility to undertake performance comparisons if that might be a worry.

conrad-watt:
My understanding is that Web engines have similar limits - e.g. [here](https://chromium.googlesource.com/v8/v8/+/refs/heads/main/src/wasm/wasm-limits.h#48) in V8 there appears to be a limit of ~50,000 locals + 1000 params per function. So even if the code gets past Binaryen there may still be issues.

kripken:
@juj 

> Before this PR, the fuzzer would hit an assertion failure on this line

I believe it should not hit that assertion because the relevant passes should call `canRun()` before. The assertion checks the same function so it basically ensures that they did in fact make that call (and if we cannot run, they would skip the liveness analysis).

@conrad-watt 

> My understanding is that Web engines have similar limits - e.g. [here](https://chromium.googlesource.com/v8/v8/+/refs/heads/main/src/wasm/wasm-limits.h#48) in V8 there appears to be a limit of ~50,000 locals + 1000 params per function. So even if the code gets past Binaryen there may still be issues.

True, good point. However, Binaryen allows a larger limit internally (before this PR: 64K with liveness, 4GB without liveness) because running the optimizer will usually reduce the number of locals in the final output below the web limit. It's common for autogenerated code to have massive amounts of locals but they almost all optimize away. So the same might happen with @juj's case.

juj:
> I believe it should not hit that assertion because the relevant passes should call canRun() before. The assertion checks the same function so it basically ensures that they did in fact make that call (and if we cannot run, they would skip the liveness analysis).

Oh thanks, now I see the code flow. Yeah, it is likely that the messages we are seeing are just warnings, but then the actual crash/error for the build for the customer is something else - hard to say if the actual failure were due to that since I was not able to locally repro identically same behavior.

> My understanding is that Web engines have similar limits - e.g. [here](https://chromium.googlesource.com/v8/v8/+/refs/heads/main/src/wasm/wasm-limits.h#48) in V8 there appears to be a limit of ~50,000 locals + 1000 params per function. So even if the code gets past Binaryen there may still be issues.

That is an interesting observation. Though I would think that even if that is the case, it should certainly not be a reason to keep/have the limit in Binaryen esp. since ridding from the limit is as easy as this. Those defaults can change in the future, and the data in the tools could be flowing to other VMs, or to other passes that the final locals count or the limits won't be the same.

juj:
> The problem is that a fuzz testcase might crash due to failing to allocate here, which would look like the fuzzer found a bug. I'm not sure offhand how to ignore that situation.

Now I do see this limit on the CI actually. I suppose the CI bots won't have a swap space like desktop setups do or idk. To resolve this, I opted to do a hybrid backing storage: when the number of locals is low, use a dense N*N matrix, and if the number of locals is high (>8k locals), then use a std::unordered_map.

This way it should be fairly safe to assume that performance will not be affected, and memory usage will also stay low.

src/cfg/liveness-traversal.h
```diff
@@ -171,30 +172,18 @@ struct LivenessWalker : public CFGWalker<SubType, VisitorType, Liveness> {
...
-    if (uint64_t(numLocals) * uint64_t(numLocals) <=
-        std::numeric_limits<Index>::max()) {
-      return true;
-    }
-    std::cerr << "warning: too many locals (" << numLocals
-              << ") to run liveness analysis in " << this->getFunction()->name
-              << '\n';
-    return false;
-  }
+  bool canRun(Function*) { return true; }
```
kripken, line 175: I think we can just remove this from all callers, if it's always true.

src/cfg/liveness-traversal.h
```diff
@@ -171,30 +172,18 @@ struct LivenessWalker : public CFGWalker<SubType, VisitorType, Liveness> {
...
     numLocals = func->getNumLocals();
-    assert(canRun(func));
-    copies.resize(numLocals * numLocals);
-    std::fill(copies.begin(), copies.end(), 0);
+    copies.recreate(numLocals);
+    std::fill(totalCopies.begin(),
+              totalCopies.begin() +
+                std::min<uint64_t>(totalCopies.size(), numLocals),
+              0);
     totalCopies.resize(numLocals);
```
kripken, line 186: The code from before looks simpler I think, where it resizes first, then does a fill on the entire thing after that - was there a reason not to do it that way?

src/support/sparse_square_matrix.h
```diff
@@ -0,0 +1,63 @@
...
+template<typename Ty> class sparse_square_matrix {
+  std::vector<Ty> denseStorage;
+  std::unordered_map<uint64_t, Ty> sparseStorage;
+  uint32_t N;
+
+public:
+  sparse_square_matrix() : N(0) {}
+
+  explicit sparse_square_matrix(int N) : N(N) {
+    if (N < 8192) {
```
kripken, line 34: 8192 could perhaps be `static const size_t DenseLimit = 8192` and that constant used in the relevant places.

src/support/sparse_square_matrix.h
```diff
@@ -0,0 +1,63 @@
...
+      denseStorage.resize(N * N);
+    }
+  }
+
+  void set(uint32_t i, uint32_t j, const Ty& value) {
+    if (denseStorage.empty()) {
```
kripken, line 40: maybe add a helper
```cpp
bool usingDenseStorage() { return !denseStorage.empty(); }
```
?

src/support/sparse_square_matrix.h
```diff
@@ -0,0 +1,63 @@
...
+#include <unordered_map>
+#include <vector>
+
+template<typename Ty> class sparse_square_matrix {
+  std::vector<Ty> denseStorage;
+  std::unordered_map<uint64_t, Ty> sparseStorage;
+  uint32_t N;
+
+public:
+  sparse_square_matrix() : N(0) {}
```
kripken, line 31:
```suggestion
  sparse_square_matrix() : N(0) {
    assert(N > 0);
  }
```
(just to avoid corner cases with ambiguity around dense or sparse storage being used)

src/support/sparse_square_matrix.h
```diff
@@ -0,0 +1,63 @@
...
+
+  const Ty get(uint32_t i, uint32_t j) const {
+    if (denseStorage.empty()) {
+      auto iter = sparseStorage.find(i * N + j);
+      return iter == sparseStorage.end() ? Ty() : iter->second;
+    }
+    return denseStorage[i * N + j];
+  }
+
+  // Resizes the matrix, and discards all previous data entries.
```
kripken, line 55: Might be worth mentioning that the resize will initialize the dense storage's values to the default while it does so.

src/support/sparse_square_matrix.h
```diff
@@ -0,0 +1,63 @@
...
+  void recreate(uint32_t N) {
```
kripken, line 56: The N passed in here might differ from the one before, so we should update the N on the instance.

kripken (commented):
For testing, see e.g. `test/example/small_set.*` which tests SmallSet.

src/cfg/liveness-traversal.h
juj: It is redundant work. After resizing, the resized elements will be zeros already, so one only needs to clear the old entries.

src/support/sparse_square_matrix.h
juj: Sorry, probably some typo here, or? The assert would always trigger here, as default initialization initializes N to zero, and 0 > 0 would be false.

src/support/sparse_square_matrix.h
juj: That was what the "discards all previous data entries" attempted to convey. Updated the wording to better message this.

src/support/sparse_square_matrix.h
juj: Thanks, that was a clear bug, fixed now.

src/cfg/liveness-traversal.h
juj: Check

src/support/sparse_square_matrix.h
juj: Oh now I think I understand your intent.

If the matrix is of 0*0 size, then it does not matter if it is categorized to use dense or sparse storage, since there won't be any elements stored in it, and it won't be valid to call set() or get() with any index. Added assert()s to the get() and set() functions to check that the indices that the functions are called with will be within bounds.

src/cfg/liveness-traversal.h
juj: Adjusted to do a vector.clear() followed by a vector.resize(), maybe that looks the simplest. And looks like is quite close to the same performance. Authored a second PR #4580 to systematically do the same to other places as well.

kripken (commented):
Nice! Just a few things before merging:

1. We should fuzz this. I started the fuzzer up now on this PR.
2. Let's measure compilation speed to make sure this does not regress. I don't think it should, but let's make sure. (Of course, it's fine if it's slower on the case > 8192, since it will actually be doing more work.)
3. Please add some unit testing for this new class. As mentioned before `test/example/small_set.*` is an example of such a test and can be used as a template.

juj:
Updated the PR with unit tests.

I don't have good workloads to use as a performance benchmarks. The main reason for investing to implement this kind of hybrid backing storage approach (instead of unconditionally using `std::unordered_map` always) was to produce a good guarantee that performance would not be impacted. Even if there is an observable performance impact to this PR, I am not sure how this code could be changed to help that.

test/example/sparse_square_matrix.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+  // Recreating should resize the matrix.
+  m.recreate(100);
+  assert(m.width() == 100);
+
+  // Small matrices should use dense storage.
+  assert(m.usingDenseStorage());
+
+  // Setting and getting element values in dense storage should work.
+  for(int y = 0; y < 100; ++y)
```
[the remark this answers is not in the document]
juj: Great, thanks for benchmarking!

I am mainly working on a Windows desktop and a macOS laptop, and looks like clang-format script does not seem to run on either of them (or I don't know how to install the prerequisites) Good thing that the CI gives a diff though, so I just manually type in the CI suggested clang-format changes.

juj:
Looks like tsan did not like the new test. For good measure, put that to a rerun, although very surprising if it would randomly fluke on the new added test specifically. Not completely sure yet what would be the issue with tsan on the new code.

PR #4570, opened 2022-04-01
Restore and fix SpillPointers pass

kripken:
Reverts the removal in #3261, fixes compile errors in internal API
changes since then, and flips the direction of the stack for the
wasm backend.

(Can see individual commits for each part of that.)

We should figure out a way to actually test this, though, if we want
to restore it.


ChrisJefferson:
Here is an example of a function which I might want to work with this code -- this works fine with GCC at `-O3`, but fails when compiled with emcc. This is too big at the moment to be a test (and is filled with printfs), but hopefully shows the principle. 

https://gist.github.com/ChrisJefferson/33a8909364f686358da7f68a56ae221b

This PR fixes this code with no optimisation, but not when compiled at `-O1` (but that might be because the compiler is clever enough to optimise out the secret_stack_value altogether? Sorry, I'm not an expert in how to read wasm).

ChrisJefferson:
I tried making an example test (this code is far too big of course!)

https://gist.github.com/ChrisJefferson/3d0813c40b130d1da5e04dc92cc5b36c -- also a second file containing a function `id` which just returns it's argument, just to stop the compiler optimising away the local variable.

With `emcc -O1` this code fails, then passes with the "Spill Pointers" pass. With `-O2` however it fails with "spill pointers", and more worryingly it output looks corrupted (the global pointer which should be constant changes value at some point while printing).

kripken:
@ChrisJefferson Thanks!

The challenge will be testing it here in binaryen though. I'm not sure atm how to do that. I also probably won't have time to look into it soon, sorry, but perhaps someone else can take this PR as a starting point.

ChrisJefferson:
Thanks. Is there any place where a test might live? (I could take a look at existing tests), or is there not really a test framework to run an optimiser on some code, and check the result?

kripken:
@ChrisJefferson Sorry for the late response. We do have tests to run the optimizer and see the code that results (`tests/lit/`, `tests/passes/` - and this PR has a test there) but only running the code would have noticed the wrong stack direction... and that requires a full program that doesn't really fit in the test suite here. So this is an awkward situation, and I don't have a good idea.

However, this has come up again in https://github.com/emscripten-core/emscripten/issues/17131#issuecomment-1145703382 and so I think we should just land this, and figure out more detailed testing later (perhaps on the emscripten side?)

src/abi/stack.h
```diff
@@ -0,0 +1,145 @@
...
+// FIXME: This function assumes that the stack grows upward, per the convention
+// used by fastcomp.  The stack grows downward when using the WASM backend.
+
+inline void
+getStackSpace(Index local, Function* func, Index size, Module& wasm) {
+  // Attempt to locate the stack pointer by recognizing code idioms
+  // used by Emscripten.  First, look for a global initialized to an
+  // imported variable named "STACKTOP" in environment "env".
+  auto* stackPointer =
+    GlobalUtils::getGlobalInitializedToImport(wasm, ENV, "STACKTOP");
```
sbc100, line 51: `STACKTOP` is not a thing in emscripten anymore.

src/abi/stack.h
```diff
@@ -0,0 +1,145 @@
...
+    if (stackSaveFunctionExport &&
+        stackSaveFunctionExport->kind == ExternalKind::Function) {
+      auto* stackSaveFunction =
+        wasm.getFunction(stackSaveFunctionExport->value);
+      assert(!stackSaveFunction->imported());
+      auto* globalGet = stackSaveFunction->body->dynCast<GlobalGet>();
+      if (globalGet) {
+        stackPointer = wasm.getGlobal(globalGet->name);
+      }
+    }
```
sbc100, line 77: We have getStackPointerGlobal in `src/wasm-emscripten.h` today

sbc100:
Would be nice to had a more readable/smaller test in `tests/lit`.. but not going to block this change on that.

Shame about re-introducing those two header files just for this one pass :(

sbc100 (approved):

kripken:
Nice, thanks, `getStackPointerGlobal` makes this a lot cleaner...

I think in this PR it's best to keep the test identical to before, so this is a clean reversion of the removal. We can improve the test later if we want.

Updated after feedback.

kripken:
Oh, we can at least remove one header here, I did that now: it is easy to get the pointer type since we've added `indexType` to memories (we added that for wasm64, I believe).

PR #4574, opened 2022-04-04
[Wasm GC] Fix unreachable local.gets of non-nullable locals in CoalesceLocals

### LINKED ISSUE #4573: Crash with `--nominal --enable-gc-nn-locals`
If I optimize [this module](https://github.com/WebAssembly/binaryen/files/8396224/b63_flow.wasm.gz) (same as in https://github.com/WebAssembly/binaryen/issues/4564) with `wasm-opt -g -all -O3 -tnh --nominal --enable-gc-nn-locals` (using e6a62de43ef66b1c1ad0d740e3deeeb6365ded1b), I get a segmentation fault, or sometimes a message like
```
malloc(): unaligned fastbin chunk detected
Aborted
```
Either `--nominal` or `--enable-gc-nn-locals` works fine, but not both. Giving both with `-O1` or `-O2` also works.

kripken:
Normally we just replace unreachable `local.get`s with a constant (0, or null), but if
the local is non-nullable we can't do that.

Fixes #4573

### DIFF
--- a/src/cfg/liveness-traversal.h
+++ b/src/cfg/liveness-traversal.h
@@ -118,7 +118,24 @@ struct LivenessWalker : public CFGWalker<SubType, VisitorType, Liveness> {
     auto* curr = (*currp)->cast<LocalGet>();
     // if in unreachable code, ignore
     if (!self->currBasicBlock) {
-      *currp = Builder(*self->getModule()).replaceWithIdenticalType(curr);
+      Builder builder(*self->getModule());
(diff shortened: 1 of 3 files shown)


tlively (commented):
Why do we need to preserve the type by wrapping the unreachable in a block?

kripken:
We could emit a "raw" unreachable here but then we'd need to update types further up (say by running ReFinalize()). This pass doesn't do any type updating though, and it seems simplest to keep it that way.

tlively (approved):

Issue #4576, opened 2022-04-04
Sanitizers fail to load libbinaryen.so on CI


kripken:
```
/home/runner/work/binaryen/binaryen/out/bin/wasm-as: error while loading shared libraries: libbinaryen.so: cannot open shared object file: No such file or directory
```
https://github.com/WebAssembly/binaryen/runs/5823984219?check_suite_focus=true

That is from https://github.com/WebAssembly/binaryen/pull/4574 which does not change anything related AFAIK, so it seems like a CI issue.

kripken:
#4577 shows that this is the same as #4575 - they both error on failing to load `libbinaryen.so`. That proves this isn't sanitizer-related. Perhaps instead the issue is that the passing builds use `cc/c++` - I assume `gcc` - while the failing ones all use `clang`?

PR #4577, opened 2022-04-04
Add '-vv' verbosity to binaryen-lit tests

tlively:
To help debug them when they go wrong, as seems to be happening on #4574.


kripken (approved):
Good idea!

Looks like this shows the issue is related to #4576

tlively:
I'll merge this despite the test failures since there doesn't seem to be any point in waiting.

Issue #4575, opened 2022-04-04
Lit error on unrelated PR


kripken:
The clang test on https://github.com/WebAssembly/binaryen/pull/4574 is failing for an odd lit reason, and it doesn't seem to provide any details as to why:

https://github.com/WebAssembly/binaryen/runs/5820870416?check_suite_focus=true

That PR does not modify anything related to that test AFAICT.

The test is not required I see, so I guess we can land things without it meanwhile.

I see that test was added in #3426 cc @sbc100 @tlively

tlively:
Not sure what's going on with this, but adding more verbose test output by merging https://github.com/WebAssembly/binaryen/pull/4577 might help us figure it out.

PR #4578, opened 2022-04-04
Fix CMake on CI (see #4575, #4576)

kripken:


kripken:
Does not seem to help to build with gcc...

kripken:
Well, I have no idea what's going on here, but through a "clever" method of just bisecting on differences between working and failing builds, apparently we need some extra incantation in the cmake now, which we did not need just a day ago. cmake update perhaps?

The magic:
```
-DCMAKE_INSTALL_PREFIX=out/install
``` 
Note that we don't even do `make install` in these jobs, so it's quite bizarre. But the jobs seem to start passing now..

download

tlively (approved):
Well that's a strange one! Thanks for tracking it down.

dschuff:
Interesting. what specifically was the error you saw that was fixed by this (or a link to the relevant build? I i clicked a few above but there were several different kinds of errors)

kripken:
error was
```
/home/runner/work/binaryen/binaryen/out/bin/wasm-as: error while loading shared libraries: libbinaryen.so: cannot open shared object file: No such file or directory
```

kripken:
https://github.com/WebAssembly/binaryen/runs/5825162539?check_suite_focus=true

PR #4580, opened 2022-04-05
Avoid a code pattern of vec.resize() followed by std::fill() as suboptimal.

juj:
Avoid a code pattern of vec.resize() followed by std::fill() as suboptimal. Instead do a clear()+resize().

Calling
```cpp
vec.resize(x);
std::fill(vec.begin(), vec.end(), 0);
```
is suboptimal, because .resize() already adds default-initialized (zero-initialized) values onto the vector. std::filling the same elements right after will be redundant.

Also, if the vector will need to reallocate, it will reallocate the vector to a new capacity, then memcpy the old data over, then immediately after fill that copied data to zero.

C++ STL vectors do not support a notion of "resize to new size, discard old data", so the best that one can do is one of:

```cpp
std::fill(vec.begin(), std::min(vec.size(), x), 0);
vec.resize(x);
```
or
```cpp
vec.clear();
vec.resize(x);
```
The first pattern also has the same issue of memcpying junk data over in case of a reallocate. The second pattern avoids that, and is more readable.

This PR switches the code to use the .clear() + .resize() pattern in places where std::fill() is being used. In a [synthetic microbenchmark](https://quick-bench.com/q/TUX7JBB0mhDc0nJ-W00ZFvq3pf0) this does show up, although not sure what the real world impact will be.

image

### DIFF
--- a/src/passes/LoopInvariantCodeMotion.cpp
+++ b/src/passes/LoopInvariantCodeMotion.cpp
@@ -75,7 +75,6 @@ struct LoopInvariantCodeMotion
     // FIXME: also the loop tail issue from above.
     auto numLocals = getFunction()->getNumLocals();
     std::vector<Index> numSetsForIndex(numLocals);
-    std::fill(numSetsForIndex.begin(), numSetsForIndex.end(), 0);
     LoopSets loopSets;
     {
       FindAll<LocalSet> finder(loop);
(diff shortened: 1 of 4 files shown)


src/passes/LoopInvariantCodeMotion.cpp
```diff
@@ -75,7 +75,6 @@ struct LoopInvariantCodeMotion
     // FIXME: also the loop tail issue from above.
     auto numLocals = getFunction()->getNumLocals();
     std::vector<Index> numSetsForIndex(numLocals);
-    std::fill(numSetsForIndex.begin(), numSetsForIndex.end(), 0);
```
juj, line 78: Creating a `std::vector(size)` with an initial size does already default-initialize to zeros, so here std::fill was redundant.

PR #4581, opened 2022-04-05
Refined HeapTypes with depth bounds

tlively:
Add first-class support for tracking additional information through the type
system in the form of refined heap types, i.e. heap types with additional
information that restricts the sets of values that can inhabit those types.
These refined types are implemented as a new kind of HeapType whose HeapTypeInfo
contains an unrefined base HeapType and a `Refinement` object containing the
additional information. For now that information is a "subtype depth" that
controls how deep in the subtype hierarchy the dynamic type of values inhabiting
the refined type can be. Tracking subtype depth is a generalization over
tracking whether or not types are "exact" that is still relatively cheap. We may
add additional information to `Refinement` objects in the future, for example to
differentiate between duplicate type definitions.

To avoid having to change all users of the type API to account for the
possibility that any HeapType may be refined, update most of the functions in
the API to look through any refinement.

Update subtyping and LUB calculation to take subtype depth refinements into
account.

A follow-on PR will start using refined types in the optimizer.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -120,12 +120,18 @@ struct HeapTypeInfo {
     SignatureKind,
     StructKind,
     ArrayKind,
+    RefinedKind,
   } kind;
+  struct RefinedType {
+    HeapType base;
+    Refinement refinement;
+  };
   union {
     HeapType::BasicHeapType basic;
     Signature signature;
     Struct struct_;
     Array array;
+    RefinedType refined;
   };
 
   HeapTypeInfo(HeapType::BasicHeapType basic) : kind(BasicKind), basic(basic) {}
@@ -134,6 +140,8 @@ struct HeapTypeInfo {
   HeapTypeInfo(Struct&& struct_)
     : kind(StructKind), struct_(std::move(struct_)) {}
   HeapTypeInfo(Array array) : kind(ArrayKind), array(array) {}
+  HeapTypeInfo(HeapType base, Refinement refinement)
+    : kind(RefinedKind), refined{base, refinement} {}
   HeapTypeInfo(const HeapTypeInfo& other);
   ~HeapTypeInfo();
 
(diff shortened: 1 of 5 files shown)


src/wasm/wasm-type.cpp
```diff
@@ -159,7 +167,7 @@ struct SubTyper {
   // otherwise.
   std::unordered_set<std::pair<HeapType, HeapType>> seen;
   bool isSubType(Type a, Type b);
-  bool isSubType(HeapType a, HeapType b);
+  std::optional<size_t> isSubType(HeapType a, HeapType b);
```
kripken, line 170: It's not obvious what this returns now?

kripken (commented):
Interesting approach!

I had imagined something quite different, that like `Type` has `nullable` we'd also add `exact` (so this would be an addition to `Type` and not `HeapType`). I'm not sure if that would be simpler overall though, and not sure it could be expanded to varying depths like this can.

If this code were simple I'd say let's land it and measure later, however, given the complexity I think we should measure the benefit + compare this to the alternatives first.

For measuring the benefit, I think the main pass is CFP. Here is (uncompiled, untested) code that should give the general idea:

https://gist.github.com/kripken/d592f5882eca40f255881cf070204adb

Even simpler than that code, though, it would be interesting to just see how often an oracle can actually find a precise type at the location in the diff where the oracle is queried.

src/wasm/wasm-type.cpp
tlively: The distance in the subtype hierarchy between `a` and `b` iff `a <: b`. I'll add a comment.

tlively:
> I had imagined something quite different, that like `Type` has `nullable` we'd also add `exact` (so this would be an addition to `Type` and not `HeapType`). I'm not sure if that would be simpler overall though, and not sure it could be expanded to varying depths like this can.

Adding the refinement to `Type` instead of `HeapType` would probably also work, but would be more complicated. The subtype hierarchy is defined on heap types, so keeping modifiers to it alongside the heap types reduces the distance between the related information and reduces the complexity of any code that has to consider both.

I'll work on a suite of follow-up PRs that will let us measure the benefit and costs of this change before we land it.

tlively:
Closing this for now, but might revive again in a different form in the future once we want to do nominal-style optimizations on isorecursive modules.

PR #4586, opened 2022-04-07
Implement relaxed SIMD dot product instructions

tlively:
As proposed in https://github.com/WebAssembly/relaxed-simd/issues/52.

### DIFF
--- a/src/wasm/literal.cpp
+++ b/src/wasm/literal.cpp
@@ -2372,15 +2372,31 @@ Literal Literal::pmaxF64x2(const Literal& other) const {
   return binary<2, &Literal::getLanesF64x2, &Literal::pmax>(*this, other);
 }
 
-Literal Literal::dotSI16x8toI32x4(const Literal& other) const {
-  LaneArray<8> lhs = getLanesSI16x8();
-  LaneArray<8> rhs = other.getLanesSI16x8();
-  LaneArray<4> result;
-  for (size_t i = 0; i < 4; ++i) {
-    result[i] = Literal(lhs[i * 2].geti32() * rhs[i * 2].geti32() +
-                        lhs[i * 2 + 1].geti32() * rhs[i * 2 + 1].geti32());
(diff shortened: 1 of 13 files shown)


src/wasm/literal.cpp
```diff
@@ -2372,17 +2372,33 @@ Literal Literal::pmaxF64x2(const Literal& other) const {
...
-  LaneArray<8> lhs = getLanesSI16x8();
-  LaneArray<8> rhs = other.getLanesSI16x8();
-  LaneArray<4> result;
-  for (size_t i = 0; i < 4; ++i) {
-    result[i] = Literal(lhs[i * 2].geti32() * rhs[i * 2].geti32() +
-                        lhs[i * 2 + 1].geti32() * rhs[i * 2 + 1].geti32());
+template<size_t Lanes,
+         size_t Factor,
+         LaneArray<Lanes * Factor> (Literal::*IntoLanes)() const>
+static Literal dot(const Literal& left, const Literal& right) {
```
aheejin, line 2378: Adding a brief comment about what this does can be helpful, like, "lane-wise multiply signed n-bit integers in the two input vectors and add adjacent pairs of the full (n*2)-bit results". I wasn't aware that adjacent pairs are added in the result before checking the instruction description myself.

aheejin (approved):
Why is the relaxed version instruction `i7x16` and not `i8x16`?

tlively:
> Why is the relaxed version instruction `i7x16` and not `i8x16`?

Apparently it ignores the high bit on one of the inputs. Actually, now that I write that out, I realize that this is not what I implemented in the interpreter. Will fix that.

tlively:
Actually, the text on the proposal issue is confusing, so I think the code is already correct. The output of the instruction is deterministic if the input is up to 7 bits, but otherwise it may have one of two behaviors. The code in this PR implements the same behavior as ARM, so I don't think it needs to change.

PR #4589, opened 2022-04-08
[Inlining] Preserve return_calls when possible

### LINKED ISSUE #4587: Tail calls not respected after inlining
I ran into a bit of an issue with `return_call` instructions not being honored after inlining. This can cause code that otherwise works without blowing the stack to cause a stack overflow after being optimized.

A fairly small reproduction is available here (and my apologies for the OCaml 😶): https://github.com/ospencer/binaryen_tailcall_bug_repro/blob/main/lib/Util.re

If you have `esy` installed, you can run the example with `esy && esy test`, and you'll get this output:

```wat
Unoptimized Binaryen IR:
(module
 (export "wrapper" (func $wrapper))
 (func $is_odd (param $0 i32) (result i32)
  (if (result i32)
   (i32.le_u
    (local.get $0)
    (i32.const 1)
   )
   (i32.eq
    (local.get $0)
    (i32.const 1)
   )
   (return_call $is_even
    (i32.sub
     (local.get $0)
     (i32.const 1)
    )
   )
  )
 )
 (func $is_even (param $0 i32) (result i32)
  (if (result i32)
   (i32.le_u
    (local.get $0)
    (i32.const 1)
   )
   (i32.eq
    (local.get $0)
    (i32.const 0)
   )
   (return_call $is_odd
    (i32.sub
     (local.get $0)
     (i32.const 1)
    )
   )
  )
 )
 (func $wrapper (result i32)
  (call $is_odd
   (i32.const 99999)
  )
 )
)

Optimized Binaryen IR:
(module
 (export "wrapper" (func $wrapper))
 (func $is_odd (; has Stack IR ;) (param $0 i32) (result i32)
  (if (result i32)
   (i32.le_u
    (local.get $0)
    (i32.const 1)
   )
   (i32.eq
    (local.get $0)
    (i32.const 1)
   )
   (if (result i32)
    (i32.le_u
     (local.tee $0
      (i32.sub
       (local.get $0)
       (i32.const 1)
      )
     )
     (i32.const 1)
    )
    (i32.eqz
     (local.get $0)
    )
    (call $is_odd
     (i32.sub
      (local.get $0)
      (i32.const 1)
     )
    )
   )
  )
 )
 (func $wrapper (; has Stack IR ;) (result i32)
  (call $is_odd
   (i32.const 99999)
  )
 )
)
```

I haven't looked at the code just yet, but my intuition is that this happens because it wouldn't be safe to copy over a `return_call` in the body of a function, as it likely would change the semantics of the function it's being inlined into. I can't really think of something more clever than just not inlining functions with a `return_call` instruction, though in "simple" cases like this one where the result of the inlined function is returned, the `return_call` could be kept, but that analysis seems a bit annoying to write 😅 

Let me know if there's any additional information I can provide or any way I can help.

tlively:
We can preserve return_calls in inlined functions when the inlined call site is
itself a return_call, since the call result types must transitively match in
that case. This solves a problem where the previous inlining logic could
introduce stack exhaustion by downgrading recursive return_calls to normal
calls.

Fixes #4587.

cc @ospencer

### DIFF
--- a/src/passes/Inlining.cpp
+++ b/src/passes/Inlining.cpp
@@ -247,6 +247,7 @@ struct Updater : public PostWalker<Updater> {
   Module* module;
   std::map<Index, Index> localMapping;
   Name returnName;
+  bool isReturn;
   Builder* builder;
   void visitReturn(Return* curr) {
     replaceCurrent(builder->makeBreak(returnName, curr->value));
@@ -257,6 +258,11 @@ struct Updater : public PostWalker<Updater> {
   // not cause unbounded stack growth because inlining and return calling both
   // avoid creating a new stack frame.
   template<typename T> void handleReturnCall(T* curr, HeapType targetType) {
+    if (isReturn) {
+      // If the inlined callsite was already a return_call, then we can keep
+      // return_calls in the inlined function rather than downgrading them.
+      return;
+    }
     curr->isReturn = false;
     curr->type = targetType.getSignature().results;
     if (curr->type.isConcrete()) {
@@ -329,6 +335,7 @@ doInlining(Module* module, Function* into, const InliningAction& action) {
   Updater updater;
   updater.module = module;
   updater.returnName = block->name;
+  updater.isReturn = call->isReturn;
   updater.builder = &builder;
   // Set up a locals mapping
   for (Index i = 0; i < from->getNumLocals(); i++) {
--- a/src/passes/Inlining.cpp
+++ b/src/passes/Inlining.cpp
@@ -261,6 +261,9 @@ struct Updater : public PostWalker<Updater> {
     if (isReturn) {
       // If the inlined callsite was already a return_call, then we can keep
       // return_calls in the inlined function rather than downgrading them.
+      // That is, if A->B and B->C and both those calls are return_calls
+      // then after inlining A->B we want to now have A->C be a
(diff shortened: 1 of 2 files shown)


ospencer:
The logic you mentioned on the issue makes sense to me. Thanks for the quick fix! 😄

src/passes/Inlining.cpp
```diff
@@ -257,6 +258,11 @@ struct Updater : public PostWalker<Updater> {
   // not cause unbounded stack growth because inlining and return calling both
   // avoid creating a new stack frame.
   template<typename T> void handleReturnCall(T* curr, HeapType targetType) {
+    if (isReturn) {
+      // If the inlined callsite was already a return_call, then we can keep
+      // return_calls in the inlined function rather than downgrading them.
```
kripken, line 263:
```suggestion
      // return_calls in the inlined function rather than downgrading them.
      // That is, if A->B and B->C and both those calls are return_calls
      // then after inlining A->B we want to now have A->C be a
      // return_call.
```

kripken (commented):
(not marking approve due to automerge)

lgtm!

Might be good to fuzz this before landing.

tlively:
Fuzzed for a bit with no issues, so I'll go ahead and land this.

PR #4591, opened 2022-04-12
Revert "Re-enable a previously flaky type test"

tlively:
Reverts WebAssembly/binaryen#4582.

This test is still sporadically failing on at least the mac builder.


tlively:
Landing TBR.

PR #4596, opened 2022-04-15
Remane asyncify-side-module to asyncify-relocatable

nokotan:
Related: <https://github.com/emscripten-core/emscripten/pull/15893#discussion_r849517443>

`--pass-arg=asyncify-side-module` option will be used not only from side modules, but also from main modules.


src/passes/Asyncify.cpp
```diff
@@ -1532,7 +1532,7 @@ struct Asyncify : public Pass {
     auto verbose =
       runner->options.getArgumentOrDefault("asyncify-verbose", "") != "";
     auto sideModule =
```
kripken, line 1534: Please also rename this local and other places it appears to `relocatable`.

kripken (commented):
Thanks! This will definitely be better. lgtm % comment.

src/passes/Asyncify.cpp
nokotan: I'd forgot the local variable name. Thanks!

kripken (approved):
Perfect!

PR #4598, opened 2022-04-18
Grand Unified Flow Analysis (GUFA)

kripken:
This is far from ready for review, but @tlively was curious to see the current status, so posting.

This tracks the possible contents in the entire program all at once using a single IR. That is in contrast to say DeadArgumentElimination of LocalRefining etc., all of whom look at one particular aspect of the program (function params and returns in DAE, locals in LocalRefining). The cost is to build up an entire new IR, which takes a lot of code - ~2000 lines atm, but should be close to done. At least all that code is separable from everything else and could fit entirely in the new pass. Another cost is this new IR is very big and requires a lot of time and memory to process. The benefit is that this can find opportunities that are only obvious when looking at the entire program, and also it can track information that is more specialized than the normal type system in the IR - in particular, this can track an ExactType, which is the case where we know the value is of a particular type exactly and not a subtype. Both may end up useful, but it's too early to tell.

This passes fuzzing (but we don't fuzz `--nominal` well atm, so that is somewhat limited) and a large amount of new tests, but removes too much code in dart2wasm and breaks things there. It is also too large/slow to run on j2wasm atm.

**edit**: now passes on dart2wasm. it removes 2% of code size and 5% of vars.

**edit edit**: This speeds up j2wasm microbenchmarks by 20%, and is ready for review.


kripken:
Ok, this is now ready for review. Apologies for the large amount of code and tests in this PR, but I can't see a way to split this up any more...

kripken:
Some annoyances with a CI-specific test failure, which was apparently that my cpp test used global constructors as well as Type/HeapType, and some global ctor ordering issue happened... Fixed now, I believe.

src/ir/drop.h
```diff
@@ -0,0 +1,65 @@
...
+#define wasm_ir_drop_h
+
+#include "ir/effects.h"
+#include "ir/iteration.h"
+#include "wasm-builder.h"
+#include "wasm.h"
+
+namespace wasm {
+
+// Returns the dropped children of a node (in a block, if there is more than
```
tlively, line 27: Initially parsed this as meaning that it would return the children that are _already_ dropped, which is not correct.

```suggestion
// Given an expression, returns a new expression that drops the given expression's children that cannot be removed outright due to their side effects.
```

tlively (commented):
I've started taking a look. Right off the bat, would you be able to turn the `example` tests into `gtest` tests instead?

src/ir/possible-contents.h
```diff
@@ -0,0 +1,577 @@
...
+      return;
+    }
+
+    auto type = getType();
+    auto otherType = other.getType();
+
+    if (!type.isRef() || !otherType.isRef()) {
+      // At least one is not a reference. We could in principle try to find
+      // ExactType here, say as the combination of two different constants, but
+      // as subtyping does not exist, Many is good enough anyhow, so do that.
```
tlively, line 162:
```suggestion
      // ExactType here, say as the combination of two different constants, but
      // since there is no subtyping between non-ref types, ExactType would not carry any more information than Many.
```

Alternatively, it seems like it would be pretty harmless to get the shared ExactType in the case of two literals, and that might be less surprising to readers.

src/ir/possible-contents.h
```diff
@@ -0,0 +1,577 @@
...
+    if (isNull() || other.isNull()) {
+      // If only one is a null, but the other's type is known exactly, then the
+      // combination is to add nullability (if the type is *not* known exactly,
+      // like for a global, then we cannot do anything useful here).
+      if (!isNull() && isTypeExact()) {
+        value = ExactType(Type(type.getHeapType(), Nullable));
+        return;
+      } else if (!other.isNull() && other.isTypeExact()) {
+        value = ExactType(Type(otherType.getHeapType(), Nullable));
+        return;
```
tlively, line 179: This repetition could be avoided if this were a function operating on two `PossibleContent`s rather than a mutating method by canonicalizing to have the first of the two combined values be the null one, for instance. That would also let you combine the `value = *` lines with the `return` lines. Another way to combine those lines would be to implement e.g.

```c++
template<T> PossibleContents& operator=(T val) {
  *this = PossibleContents(val); return *this;
}
```

src/ir/possible-contents.h
```diff
@@ -0,0 +1,577 @@
...
+      // Literal; or both might be ExactTypes and only one might be nullable).
+      // In these cases we can emit a proper ExactType here, adding nullability
+      // if we need to.
+      value = ExactType(Type(
+        type.getHeapType(),
+        type.isNullable() || otherType.isNullable() ? Nullable : NonNullable));
+      return;
+    }
+
+    // Nothing else possible combines in an interesting way; emit a Many.
```
tlively, line 203: What about a literal and an exact type? Those combine to another exact type if the literal is of the same type.

edit: Ah, I see that this is covered since `isTypeExact` is different from `isExactType`.

src/ir/possible-contents.h
```diff
@@ -0,0 +1,577 @@
...
+    value = Many();
+  }
+
+  bool isNone() const { return std::get_if<None>(&value); }
+  bool isLiteral() const { return std::get_if<Literal>(&value); }
+  bool isGlobal() const { return std::get_if<GlobalInfo>(&value); }
+  bool isExactType() const { return std::get_if<Type>(&value); }
```
tlively, line 210: This would make these methods strictly more useful, since they could be used in all the same contexts as the `is*` variants, but could also be used in conditions that perform assignment such as in `getType` below.

```suggestion
  Literal* getLiteral() const { return std::get_if<Literal>(&value); }
  GlobalInfo* getGlobal() const { return std::get_if<GlobalInfo>(&value); }
  Type* getExactType() const { return std::get_if<Type>(&value); }
```

src/ir/possible-contents.h
```diff
@@ -0,0 +1,577 @@
...
+    return value == other.value;
+  }
+
+  bool operator!=(const PossibleContents& other) const {
+    return !(*this == other);
+  }
+
+  // Combine the information in a given PossibleContents to this one. The
+  // contents here will then include whatever content was possible in |other|.
+  void combine(const PossibleContents& other) {
```
tlively, line 135: I think it would make sense to move this out of the header and into the .cpp file.

src/ir/possible-contents.h
```diff
@@ -0,0 +1,577 @@
...
+    } else {
+      WASM_UNREACHABLE("bad value");
+    }
+  }
+
+  // Returns whether the type we can report here is exact, that is, nothing of a
+  // strict subtype might show up.
+  //
+  // This returns false for None and Many, for whom it is not well-defined.
+  bool isTypeExact() const { return isExactType() || isLiteral(); }
```
tlively, line 251: This is confusingly similar to `isExactType`. Can we change it to `hasExactType` maybe? Or maybe this should be rolled into `getExactType` as suggested above so that it covers both `ExactType` and `Literal`.

src/ir/possible-contents.h
```diff
@@ -0,0 +1,577 @@
...
+
+  // Whether we can make an Expression* for this containing the proper contents.
+  // We can do that for a Literal (emitting a Const or RefFunc etc.) or a
+  // Global (emitting a GlobalGet), but not for anything else yet.
+  bool canMakeExpression() const { return isLiteral() || isGlobal(); }
+
+  Expression* makeExpression(Module& wasm) {
```
tlively, line 258: Can we roll these together into `maybeMakeExpression` that returns nullptr if no expression can be made? If `canMakeExpression` is ever used independently of `makeExpression` we shouldn't do this because making an expression is nontrivial work.

src/ir/possible-contents.h
```diff
@@ -0,0 +1,577 @@
...
+    Builder builder(wasm);
+    if (isLiteral()) {
+      return builder.makeConstantExpression(getLiteral());
+    } else {
+      auto name = getGlobal();
+      return builder.makeGlobalGet(name, wasm.getGlobal(name)->type);
+    }
+  }
+
+  size_t hash() const {
```
tlively, line 268: Is there any reason to have the method separate from the `std::hash` specialization or could we combine them? Also, `std::variant` provides a specialization of `std::hash` already if all of its variants have one, so that might be a simpler and more modular way to provide hashing.

tlively (commented):
I have finished reading possible-contents.h!

kripken:
I opened #4685 that contains the first part of this, for easier review. I'll read and address the comments already written here in that PR.

src/ir/possible-contents.h
```diff
@@ -0,0 +1,577 @@
...
+//                     If the type here is nullable then null is also allowed.
+//                     TODO: Add ConeType, which would include subtypes.
+//                     TODO: Add ExactTypePlusContents or such, which would be
+//                           used on e.g. a struct.new with an immutable field
+//                           to which we assign a constant: not only do we know
+//                           the exact type, but also certain field's values.
+//
+//  * Many:            Anything else. Many things are possible here, and we do
+//                     not track what they might be, so we must assume the worst
+//                     in the calling code.
```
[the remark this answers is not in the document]
kripken: Yes, tracking a type instead of Many may help. The "cone" type mentioned a little lower down in a TODO is basically that (though it just applies to reference types, but I think that's all we need).

src/ir/possible-contents.h
kripken: I lean towards keeping this as a member function rather than static on two inputs, since the by-far common case is to keep merging in new contents to a particular location. It's possible clang is smart enough to avoid overhead of a static method + a copy, but I'd rather not risk it given how slow this operation already is...

Interesting idea to define `operator=(T val)` like that. I agree it could save a line or two. However, a worry I've had during development here is to make a mistake and do something like `PossibleContents(Type(x))` - that turns into an ExactType atm, but should probably be a ConeType when we add one later. So I think explicit constructors and assignments are safer than a template that might match more things than we expect.

src/ir/possible-contents.h
kripken: Interesting, but I think it's slightly safer as it is, where `getLiteral` etc. will assert instead of returning null.

src/ir/possible-contents.h
kripken: This is by far the most performance-sensitive function in this class. It is hammered on constantly by the flow analysis. So I think it is important to keep it inlineable?

src/ir/possible-contents.h
kripken: I agree the current name has some risk of confusion. `hasExactType` can also make sense, interesting idea. I feel like it has similar risks, though, so I don't have a preference.

@aheejin `isExactType` checks if something is an exact type, and then we have two options for a function that checks if something's type is known exactly (which is true for ExactType, but also constant values as well):
* `isTypeExact` (sort of "is the type exact")
* `hasExactType` (sort of "does it have an exact type")

Which do you think is clearer?

src/ir/possible-contents.h
kripken: Yes, the `can` method is used independently, so these are intentionally separate. Can see that in `GUFA.cpp` - we rule out the case where we just can't make an expression early, then only later if other things line up do we make an expression, and we actually call that on a potentially different value than we called `can` on.

src/ir/possible-contents.h
kripken: The main reason to separate the internals is that when we do the `namespace std` stuff we'd need to write a lot more `wasm::` namespacing etc., making the code less clear. Doing it in the class is the most concise I think.

You know, I did read that `std::variant` has a default hash impl... but I couldn't get it to work. C++ error messages are not very helpful for stuff like that :cry: It would just dump tons of lines of complaints at me...

kripken:
Thanks for the feedback! All comments so far should be addressed or replied to. 95% of them are in the other PR, but one is here since it is on `drop.h` which is not in that other PR.

kripken:
(Though I did merge the other PR to here, so the changes are visible here as well. I guess we can discuss in either place if I keep doing that.)

src/ir/possible-contents.h
tlively: Why? Any mistakes would turn into segfaults (at least on native builds), which are similar to assertions in terms of usefulness.

src/ir/possible-contents.h
kripken: (though if we start to use PossibleContents in more places, that might change... but we can refactor later if so)

tlively: Oh right, I missed that. Thanks for investigating!

src/ir/possible-contents.h
kripken: I guess I don't feel strongly. One issue, though, is that `get_if` returns a pointer to something whose lifetime might end before the pointer is discarded. It's just not a safe/nice API, I guess is how I feel about it. To make it safe we could make copies of the things before returning them here (they all fit in register or two), but at that point I feel like it's making the code less clear...

tlively: Yes, another alternative is to return a copy in a `std::optional`. That seems like overkill though. By far the most common use case I've seen for this kind of code is

```
if (auto* val = var.getSomeVariant()) {
  // Use val without it escaping
}
```

kripken: That's definitely the most common case. I'm worried about bugs in the less common paths though. And it is a simpler mental model to not have to think about the lifetime of the pointer returned here, by not returning a pointer.

If you feel strongly about this, though, then I think a copy + std::optional might be a reasonable compromise? (I should benchmark it if we want to go that way, though.)

kripken: Oh, a downside of `std::optional`, however, is the lack of asserts... I've been bitten by this a few times already. Doing `*thing` on an empty optional is undefined behavior, and in practice very strange things happen :cry:

src/ir/possible-contents.h
aheejin: `isTypeExact` wasn't very confusing when I read it, so I don't have a strong preference, but I guess I slightly prefer the latter.

kripken: Sounds good, thanks, I'll change it to `hasExactType` then.

src/ir/possible-contents.h
tlively: Too bad we're not using Rust, where neither lifetimes nor accidental dereferences can ever be problems :crab: :cry: :crab:

kripken:
Ok, after the first PR landed this should be ready for review now. This is mostly test code, as the new pass that is added does most of its work using the last PR's utility code.

tlively:
Taking a look now. Sorry for the delay again :(

test/lit/passes/gufa-optimizing.wast
```diff
@@ -0,0 +1,62 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
+;; RUN: foreach %s %t wasm-opt -all --gufa            -S -o - | filecheck %s --check-prefix NO__OPT
+;; RUN: foreach %s %t wasm-opt -all --gufa-optimizing -S -o - | filecheck %s --check-prefix YES_OPT
```
tlively, line 3: Maybe `DO_OPT` and `NO_OPT`? The double `_` is a little distracting.

tlively (commented):
Non-test code LGTM. Will take a peek at the tests next.

src/passes/GUFA.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+      // TODO: would emitting a more specific null be useful when valid?
+    }
+
+    auto* c = contents.makeExpression(wasm);
+
+    // We can only place the constant value here if it has the right type. For
+    // example, a block may return (ref any), that is, not allow a null, but in
+    // practice only a null may flow there if it goes through casts that will
+    // trap at runtime.
```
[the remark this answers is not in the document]
kripken: Good question. I believe right now it does not, since we don't have full optimization/analysis of `ref.as` etc. Once we have all that, we might be able to assert here on the type being valid. I'll add a TODO.

src/passes/GUFA.cpp
```diff
@@ -0,0 +1,308 @@
...
+      return;
+    }
+
+    // This is reachable. Check if we can emit something optimized for it.
+    // TODO: can we handle more general things here too?
+    if (!contents.canMakeExpression()) {
+      return;
+    }
+
+    if (contents.isNull() && curr->type.isNullable()) {
```
aheejin, line 163: Can this ever be non-nullable when the contents is null?

src/passes/GUFA.cpp
```diff
@@ -0,0 +1,308 @@
...
+      // Null values are all identical, so just fix up the type here (the null's
+      // type might not fit in this expression).
+      contents =
+        PossibleContents::literal(Literal::makeNull(curr->type.getHeapType()));
```
aheejin, line 167: Not sure what this does. If this expression was null before and the code was valid, that means it was a `Literal` of `curr->type.getHeaptype()` or one of its subtypes. If it was one of the subtypes, doesn't this effectively lose the more fine-grained type info?

src/passes/GUFA.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+      // Note that if curr's type is *not* nullable, then the code will trap at
+      // runtime (the null must arrive through a cast that will trap). We handle
+      // that below, so we don't need to think about it here.
```
aheejin, line 171: When can this happen? Does that validate then?

aheejin (commented):
Sorry for the late comments! Haven't looked at the tests yet, and I think I'm not understanding some crucial part about nullability, so I think I'll ask that first.

Another question: We don't use `ExactType` info for optimizations? We don't call `hasExactInfo` anywhere, and we don't do anything if `canMakeExpression` returns false.

kripken:
> We don't use ExactType info for optimizations? We don't call hasExactInfo anywhere, and we don't do anything if canMakeExpression returns false.

We don't use it directly, correct. We just check if we know the contents are equal to a simple expression like a constant.

The benefit of exact types is internal in the flow analysis. Imagine that we do `(struct.get $A 3 (local.get $foo))`. If we have exact type info on `$foo`, say that it must be of class `$B`, then that may prove that field `3` must contain a particular constant. `ContentOracle` would then return that constant for us here.

(But maybe we'll find uses for exact type info in this pass? Not sure.)

src/passes/GUFA.cpp
kripken: It should trap in that case before we get here (see comment on line 169 below, which should maybe be moved up here).

That is, a null can be cast to have a non-nullable type, and so the flow might think that a null can arrive here. On line 182 there is a TODO to filter out such values in the flow.

src/passes/GUFA.cpp
kripken: Say,
```wat
(block (result (ref $A)) ;; non-nullable
  (ref.as_non_null
    (ref.null $A)))
```
That validates but will trap at runtime.

aheejin: Thanks, I think some of my questions effectively asked the same thing. I didn't think about cast instructions so I was assuming `ref.null` is the only instruction that can contain a null value and confused.

src/passes/GUFA.cpp
aheejin: Nevermind, I think it was the same confusion as this: https://github.com/WebAssembly/binaryen/pull/4598#discussion_r910467140

aheejin: Oh no, I think my question was this:
```wast
(ref.cast $struct.Parent (ref.null $struct.Child))
```
Then the possible contents for this location contains `(ref.null $struct.Child)`, but this replaces the `ref.cast` with `(ref.null $struct.Parent)` Wouldn't it be better to make this `(ref.null $struct.Child)`, which is a more fine-grained type? (I might be mistaken from somewhere)

kripken:
The fuzzer found a nondeterminism bug here due to nulls. Fixed, see details in last commit.

aheejin (commented):
Some more questions:

src/passes/GUFA.cpp
```diff
@@ -0,0 +1,311 @@
...
+        contents = PossibleContents::literal(
+          Literal::makeNull(curr->type.getHeapType()));
+      }
+
+      // Note that if curr's type is *not* nullable, then the code will trap at
+      // runtime (the null must arrive through a cast that will trap). We handle
+      // that below, so we don't need to think about it here.
+
+      // TODO: would emitting a more specific null be useful when valid?
+    }
```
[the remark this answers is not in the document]
kripken: This code handles the case of the contents being equal to a null, and the null happens to have a bad type for this location. We basically "preprocess" the contents here, in hopes of allowing more optimizations to work. If we didn't do this then the type would be wrong and lower down we wouldn't optimize - but nulls are special, we can change their types and then optimize with that.

We don't set `optimized` here since this just does "preprocessing". We just rewrite `contents`, that is, we haven't written anything to Binaryen IR yet, so we haven't optimized that IR yet.

src/passes/GUFA.cpp
```diff
@@ -0,0 +1,311 @@
...
+      }
+      optimized = true;
+    } else {
+      // The type is not compatible: we cannot place |c| in this location, even
+      // though we have proven it is the only value possible here.
+      if (Properties::isConstantExpression(c)) {
+        // The type is not compatible and this is a simple constant expression
+        // like a ref.func. That means this code must be unreachable. (See below
+        // for the case of a non-constant.)
+        replaceWithUnreachable();
```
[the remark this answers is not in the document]
kripken: The casts would be along the way:
```wat
(block (result (ref data))
  (ref.as_data
    (ref.func $foo)))
```
The contents `c` here are a `ref.func $foo`. But the type at the block is `data` which is impossible. The cast along the way allowed the contents to get here in the GUFA flow, and also they allowed the wasm to validate, but at runtime the cast will trap.

In general, if we see the contents must have a type that is not valid for the declared type in the wasm, then either we have a bug, or the wasm spec has a bug, or the code is unreachable, since the wasm type system should prevent this situation.

src/passes/GUFA.cpp
```diff
@@ -0,0 +1,311 @@
...
+    // Ignore a trap, as the unreachable replacement would trap too.
+    effects.trap = false;
+    return !effects.hasUnremovableSideEffects();
+  }
+
+  void visitExpression(Expression* curr) {
+    // Skip things we can't improve in any way.
+    auto type = curr->type;
+    if (type == Type::unreachable || type == Type::none ||
+        Properties::isConstantExpression(curr)) {
```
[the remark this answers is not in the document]
kripken: True, we can't optimize those cases either. Bailing out would make sense. However, `isValidInConstantExpression` is meant for stuff that validates in wasm global locations, so I'm not sure it is the right meaning for us here. We could add a list of stuff to bail out on, but these are not super-common instructions so the perf benefit would be small I think?

src/passes/GUFA.cpp
```diff
@@ -0,0 +1,311 @@
...
+        // the global's type is nullable, so the global.get's type will be as
+        // well. When we get to the ref.as_non_null, we then want to replace it
+        // with a global.get - in fact that's what its child already is, showing
+        // it is the right content for it - but that global.get would not have a
+        // non-nullable type like a ref.as_non_null must have, so we cannot
+        // simply replace it.
+        //
+        // For now, do nothing here, but in some cases we could probably
+        // optimize (e.g. by adding a ref.as_non_null in the example) TODO
+        assert(c->is<GlobalGet>());
```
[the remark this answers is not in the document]
kripken: For now this can only be a real constant, like `i32.const` or `ref.func`, or `global.get` of an immutable global (which is "effectively constant", but we don't know the value at compile time). The type isn't the issue here, it's figuring out how to infer such a constant (or "effectively constant") value properly. I'm not sure how to do that for `struct.new` (but maybe there's a way).

src/passes/GUFA.cpp
```diff
@@ -0,0 +1,311 @@
...
+      }
+    }
+  }
+
+  // TODO: If an instruction would trap on null, like struct.get, we could
+  //       remove it here if it has no possible contents. That information
+  //       is present in OptimizeInstructions where it removes redundant
+  //       ref.as_non_null, so maybe there is a way to share that.
```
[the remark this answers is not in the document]
kripken: Good catch, yeah, this is wrong. The missing context is "in traps-never-happen mode". In that mode we can remove more stuff. I'll add a comment.

src/passes/GUFA.cpp
```diff
@@ -0,0 +1,311 @@
...
+    //       filtering content not just on ref.cast as it does now, but also
+    //       ref.as etc. Once it does those we could assert on the type being
+    //       valid here.
+    if (Type::isSubType(c->type, curr->type)) {
+      if (canRemove(curr)) {
+        replaceCurrent(getDroppedChildrenAndAppend(curr, wasm, options, c));
+      } else {
+        // We can't remove this, but we can at least drop it and put the
+        // optimized value right after it.
+        replaceCurrent(builder.makeSequence(builder.makeDrop(curr), c));
```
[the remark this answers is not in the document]
kripken: There is some risk here, that is true. Perhaps this should only be done when not optimizing for size? However, propagating the constant to here will potentially open up opportunities to remove other code, so I'm not sure. Will add a comment now but let's keep thinking about it.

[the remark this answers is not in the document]
aheejin: Maybe we disable this part of code and do a quick check on perf and size, if that's easily doable?

src/passes/GUFA.cpp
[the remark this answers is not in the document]
aheejin: Ah right, sorry, I keep confusing the original expression with the expression created by the contents oracle...

src/passes/GUFA.cpp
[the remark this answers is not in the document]
aheejin: Yeah, I was mostly wondering what the criteria is for bailing out and not bailing out. Then what happens to these expressions? Do we eventually bail out here for them?
```cpp
    // This is reachable. Check if we can emit something optimized for it.
    // TODO: can we handle more general things here too?
    if (!contents.canMakeExpression()) {
      return;
    }
```

src/passes/GUFA.cpp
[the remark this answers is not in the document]
aheejin: Ah right we only assigned it to `contents` and didn't do `replaceCurrent` yet... Missed that part. Thanks.

src/passes/GUFA.cpp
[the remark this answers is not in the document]
aheejin: Sorry I again confused the content `c` with the original expression `curr`... 😅 Nevermind.

src/passes/GUFA.cpp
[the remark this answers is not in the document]
kripken: Testing now, code size is 1.6% smaller with this optimization. It's harder to test speed (we don't have great wasm GC benchmarks for this kind of thing), but emitting more constants is almost always better. So at least on this one example (j2wasm application) it looks worth doing.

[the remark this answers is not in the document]
aheejin: Hmm does that mean DCE or Vacuum was eventually able to remove those side-effect-having expressions? I was like, even if we emit an additional constant, if we cannot remove the original expression, that original expression still needs to be run..

src/passes/GUFA.cpp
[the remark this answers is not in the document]
kripken: I think what happens is `canMakeExpression()` returns false, and we stop there. That function only returns true for the two cases we support so far, constants and immutable globals. A `struct.new` would end up as an exact type which is not enough information for us to do something with, at least for now.

src/passes/GUFA.cpp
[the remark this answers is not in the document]
kripken: Yeah, there's no guarantee we can remove it. But sometimes dropping an expression lets us remove parts of it:
```wat
(select
  (call $get-A) ;; returns ref.func $foo
  (call $get-B) ;; also returns ref.func $foo
  (..condition..)
)
=>
(drop
  (select
    (call $get-A) ;; returns ref.func $foo
    (call $get-B) ;; also returns ref.func $foo
    (..condition..)
  )
)
(ref.func $foo)
```
Now that the select is dropped we don't need it, and vacuum will remove it and just leave the calls (and maybe the condition).

(Not sure if that's the main factor here though. Could also be that we don't remove any dropped code, but the constant lets us remove other code...)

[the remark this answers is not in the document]
MaxGraey: I'm wondering what if move all meaningful reduction code from vacuum pass to some common util file and reuse routines from this file for vacuum and this pass? At least for those functions that don't require a global analysis of the program, but only subexpressions. This will allow you to avoid code overgrowth and its subsequent removal through a separate pass. It will also allow you to evaluate the benefit of constant propagation ahead of time (at least in local cases).

src/passes/GUFA.cpp
```diff
@@ -0,0 +1,319 @@
...
+    // example, a block may return (ref any), that is, not allow a null, but in
+    // practice only a null may flow there if it goes through casts that will
+    // trap at runtime.
+    // TODO: GUFA should eventually do this, but it will require it properly
+    //       filtering content not just on ref.cast as it does now, but also
+    //       ref.as etc. Once it does those we could assert on the type being
+    //       valid here.
+    if (Type::isSubType(c->type, curr->type)) {
+      if (canRemove(curr)) {
+        replaceCurrent(getDroppedChildrenAndAppend(curr, wasm, options, c));
```
aheejin, line 191: Why can't we remove its children too? This is within `canRemove`, which checks `EffectAnalyzer` to see if it has any side effects. So that we are here means `curr` can be removed safely, no? The same question for this function's use in `replaceWithUnreachable`.

aheejin (commented):
Haven't finished reading the tests yet. But some more questions:

src/passes/GUFA.cpp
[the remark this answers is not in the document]
kripken: @MaxGraey The vacuum code is already reusable right now, isn't it? We reuse it from here already just by doing `runner.add("vacuum");`. If you mean we just need a subset of vacuum, I'm not sure that's true, we need basically all of it.

test/lit/passes/gufa.wast
```diff
@@ -0,0 +1,967 @@
...
+        (i32.const 1)
+      )
+    )
+    ;; As above, but now the function is not exported. That means it has no
+    ;; callers, so the only possible contents for $param are the local.set here,
+    ;; as this code is unreachable. We will infer a constant of 1 for all values
+    ;; of $param here. (With an SSA analysis, we could see that the first
+    ;; local.get must be unreachable, and optimize even better; as things are,
+    ;; we see the local.set and it is the only thing that sends values to the
+    ;; local.)
```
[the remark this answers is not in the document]
kripken: Without SSA we consider a local index either reachable or not as a whole, not individual sets and gets. This local index seems reachable due to the set here. But the set itself is not really reachable since it's behind an unreachable get - but we need an SSA-like analysis to see that.

src/passes/GUFA.cpp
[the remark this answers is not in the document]
MaxGraey: `runner.add("vacuum")` run on per function level right? But It can't provide a feedback. I mean something like `tryVacuum(transform(expr))` and if `tryVacuum` return false (which mean you can't cleanup code)just don't replace current expression to transformed due to code overgrowth

test/lit/passes/gufa-no_names.wast
```diff
@@ -0,0 +1,115 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
+;; RUN: foreach %s %t wasm-opt -all --remove-unused-names --gufa -S -o - | filecheck %s
+
+;; Tests cases where the lack of names matters (hence --remove-unused-names)
+
+;; Two tags with different values. Names are added by text format parsing, which
+;; would inhibit optimizations, hence this pass requires unused names to be
+;; removed.
```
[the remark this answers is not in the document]
kripken: Good catch!

I think this mattered somehow in the past... perhaps I was using logic like `getFallthrough` which can be limited by names. I guess I changed things so this isn't an issue any more, though I don't remember when.

I removed the unneeded pass and renamed this to `gufa-tags`.

src/passes/GUFA.cpp
kripken: Very good point! Yes, this could be a lot better. I fixed it to use `ShallowEffectAnalyzer` which is all we need since we'll keep the children if we actually need them. This optimizes more code + it's faster.

aheejin: I think using `ShallowEffectAnalyzer` is better, but what I was asking about was, why do we even need to keep children if `EffectAnalyzer.hasUnremovableSideEffects()` returns `false`? 

Some expression as a whole, including children, may not have a side effect even when some of children has one. For example, 
```wast
(block $lable0
  (br $label0)
)
```
`(br $label0)` has a side effects, but the whole `(block ... )` doesn't. I guess we don't encounter this case in this pass because we don't remove named blocks anyway, but actually in this case this block may be removable, because this `block` as a whole does not have any side effects.

So, if we are to preserve children anyway, using `ShallowEffectAnalyer` is better for sure. But in the previous code, when we were using `EffectAnalyer`, if that doesn't have any side effects, can't we remove the expression, including children, altogether? Do we even need to preserve children in that case as well?

kripken: Yes, you're right that the earlier code was silly, since if `EffectAnalyzer.hasUnremovableSideEffects()` returns `false` then we don't need the children.

The new code seems optimal: we check for shallow effects which ignores the children, which allows us to remove more things, and then we do keep the children (if we need them).

aheejin: Isn't it more optimal if we don't keep children, because we can remove more code that way? I'm not sure why we need to keep children when `EffectAnalyzer.hasUnremovableSideEffects` returns false, which means the expression as a whole, including all children, doesn't have any side effects.

aheejin: > so the new code handles more cases, without losing anything (anything but the eager removal).

Is that always the case? There are cases when a child has a side effect but the parent, including all children, may not. Like the case I wrote above:
```wast
(block $lable0
  (br $label0)
)
```

Also `try`-`delegate` is a similar case; a single `try`-`delegate` may have a side effect (because `delegate` can target an outer `try`) but the outer `try` as a whole may not have any side effects.

But yeah, these may not matter because we exclude `block`s and `try`s in `canRemoveStructurally` or somewhere. But I wasn't sure it's always the case that "a child is not needed it will be removed by Vacuum eventually". Maybe it's true except the case of `block` and `try`, which we handle separately.

I think using `ShallowEffectAnalyzer` is better in terms of speed, because it's not O(n^2).

kripken: Oh, right! Sorry for not understanding you before. I forgot about the possibility of the combination having fewer effects.

Yes, I think you're right. In principle this is an issue. However, `canRemoveStructurally` rules those cases out as you said. I think any time that the combined expression has no effects, but the pieces of it do if they are separate, are cases that we can't remove other pieces.

I added a comment here.

aheejin: OK now I think I understand better. Then what's the difference between the ones we exclude in `canRemoveStructurally` and the ones in `getDroppedUnconditionalChildrenAndAppend`?

In `canRemoveStructurally`, we exclude named `block`s, `try`s, and `pop`s. In `getDroppedUnconditionalChildrenAndAppend`, we exclude `if`s and `try`s. But I think the thing they check should be the same, namely, they check whether the expression itself can be removed, regardless of its chlidren can be removed or not, which will can be dropped separately.

So my question is, 1. Are they really the same? 2. If so, do we need two places to check this? 3. And, doesn't `getDroppedUnconditionalChildrenAndAppend` need to exclude `pop` and named `block`s as well, in case it is used in places other than GUFA? So what I mean is, can we check all these within `getDroppedUnconditionalChildrenAndAppend` and remove `canRemoveStructurally` altogether? We are currently excluding expressions in three places (`canRemove`, `canRemoveStructurally`, and `getDroppedUnconditionalChildrenAndAppend`), which is confusing. If we can remove one of them that will be an improvement.

aheejin: Hmm, I think this can actually be simplified a lot more, but it may be better to do that as a follow-up, given that this PR was hold up for too many weeks due to my laziness 😅 Sorry for that. I tried to do the follow-up in #4787.

test/lit/passes/gufa-vs-cfp.wast
```diff
@@ -0,0 +1,2753 @@
...
+    ;; ref is needed: the optimizer can see that the struct.get cannot trap, as
+    ;; its reference is non-nullable).
+    (drop
+      (struct.get $struct 0
+        (struct.new_default_with_rtt $struct
+          (rtt.canon $struct)
+        )
+      )
+    )
+  )
```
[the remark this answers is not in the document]
kripken: The specific reason in the test you mentioned is that dropping each result leads to GUFA seeing that the result does not reach anywhere. And that null is an opaque placeholder for CFP, but GUFA sees that it will trap. So we need to connect results to uses, which "compresses" the test.

More details here:

https://github.com/WebAssembly/binaryen/blob/adef90c1e53a163381537939f9640228221f8a43/test/lit/passes/gufa-vs-cfp.wast#L6-L22

aheejin (approved):
Very impressive framework! 😮 LGTM and sorry for the super delayed review! 😅

kripken:
Thanks for the thorough review @aheejin !

aheejin:
Umm, I think I messed up something while merging...

kripken:
Thanks @aheejin !

kripken:
Friendly ping @tlively , did you want to take a look here?

tlively:
Thanks for the ping. I'll take a look today.

tlively (approved):
Half-rubberstamp LGTM. I don't want to hold this up any longer :disappointed:

test/lit/passes/gufa-no_names.wast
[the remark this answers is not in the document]
tlively: (This was a very old pending comment, feel free to ignore)

PR #4601, opened 2022-04-21
[NominalFuzzing] SignatureRefining: Ignore exported functions

kripken:
This hits the fuzzer when it tries to call reference exports with a null.

### DIFF
--- a/src/ir/export-utils.h
+++ b/src/ir/export-utils.h
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2022 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


tlively (commented):
Will this handle the case where an exported function has a signature type `A` that contains another signature type `B` and we need to prevent `B` from changing as well? If so, it would be good to add a test for that.

kripken:
Ah, good question... no, this does not handle that. I guess we'd need to recurse into the type. Do we have a helper for that?

tlively:
Actually, this is a general problem for all our type optimizations, right? IIRC, we punted on the problem of not modifying types that escape the module because our users don't have an interesting enough module boundary for there to be problems. Maybe a TODO to handle nested types is good enough for now.

No, I don't think we have a public type visitor utility. There is a private one in wasm-type.cpp, but it's not fit for public consumption because it observes mutable pointers to types in publicly-immutable data.

kripken:
Yeah, it is likely a general issue. I added a TODO now - also it's related to the "closed-world" flag we need to decide about. I agree for now a TODO seems enough as it will unblock nominal fuzzing.

kripken:
@tlively PTAL, did you have other comments aside from the TODO on this one?

tlively (approved):
Nope, LGTM!

PR #4602, opened 2022-04-21
[NominalFuzzing] MergeSimilarFunctions: handle nominal types properly

kripken:
This fixes two bugs: First, we need to compare the nominal types of function
constants when looking for constants to "merge", not just their structure.
Second, when creating the new function we must use the proper type of
those constants, and not just another type (or the canonicalized type, I
guess) of that structural signature,


src/passes/MergeSimilarFunctions.cpp
```diff
@@ -246,8 +246,16 @@ bool MergeSimilarFunctions::areInEquvalentClass(Function* lhs,
...
       auto* lhsCallee = module->getFunction(lhsCast->target);
       auto* rhsCallee = module->getFunction(rhsCast->target);
-      if (lhsCallee->getSig() != rhsCallee->getSig()) {
-        return false;
+      // In nominal typing we need the types to match perfectly. Otherwise,
+      // structurally is sufficient.
+      if (getTypeSystem() == TypeSystem::Nominal) {
+        if (lhsCallee->type != rhsCallee->type) {
+          return false;
+        }
```
tlively, line 254: We should unconditionally compare the types for equality. Under equirecursive typing this will be equivalent to comparing the signature structures and for the other two type systems comparing the signatures is presumably wrong.

test/lit/passes/merge-similar-functions_types.wast
```diff
@@ -0,0 +1,540 @@
...
+ ;; NOMNL:      (memory $0 (shared 16 17))
+ (memory $0 (shared 16 17))
+ ;; CHECK:      (elem declare func $2 $3)
+
+ ;; CHECK:      (export "func_406_invoker" (func $0))
+ ;; NOMNL:      (export "func_406_invoker" (func $0))
+ (export "func_406_invoker" (func $0))
+ ;; CHECK:      (export "func_448_invoker" (func $1))
+ ;; NOMNL:      (export "func_448_invoker" (func $1))
+ (export "func_448_invoker" (func $1))
```
tlively, line 39: Are these necessary? It would be good to minimize the test cases.

src/passes/MergeSimilarFunctions.cpp
kripken: Ah, thanks, I forgot how that worked in the structural case... done.

tlively (approved):

PR #4603, opened 2022-04-21
[NominalFuzzing] Don't compare nominal types in the fuzzer

kripken:
The same module will have a different type after some transformations, even
though that is not observable, like `--roundtrip`. Basically, we should not be
comparing types between separate modules, which is what the fuzzer does.

### DIFF
--- a/src/tools/execution-results.h
+++ b/src/tools/execution-results.h
@@ -151,9 +151,14 @@ struct ExecutionResults {
   bool areEqual(Literal a, Literal b) {
     // We allow nulls to have different types (as they compare equal regardless)
     // but anything else must have an identical type.
-    if (a.type != b.type && !(a.isNull() && b.isNull())) {
-      std::cout << "types not identical! " << a << " != " << b << '\n';
-      return false;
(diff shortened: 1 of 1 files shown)


src/tools/execution-results.h
```diff
@@ -151,9 +151,14 @@ struct ExecutionResults {
   bool areEqual(Literal a, Literal b) {
     // We allow nulls to have different types (as they compare equal regardless)
     // but anything else must have an identical type.
-    if (a.type != b.type && !(a.isNull() && b.isNull())) {
-      std::cout << "types not identical! " << a << " != " << b << '\n';
-      return false;
+    // We cannot do this in nominal typing, however, as different modules will
+    // have different types in general. We could perhaps compare them
+    // structurally, but that would not be right either.
```
tlively, line 156: Comparing structures one pair of types at a time wouldn't work, but we can at least have a TODO about checking for structural equivalence between the definition graphs.

tlively (approved):

src/tools/execution-results.h
kripken: Sounds good, added TODO.

PR #4604, opened 2022-04-21
[NominalFuzzing] Fix TranslateToFuzzReader::getSubType(Rtt)

kripken:
Randomly selecting a depth is ok for structural typing, but in nominal it
must match the actual hierarchy of types.

### DIFF
--- a/src/tools/fuzzing/fuzzing.cpp
+++ b/src/tools/fuzzing/fuzzing.cpp
@@ -3041,10 +3041,10 @@ HeapType TranslateToFuzzReader::getSubType(HeapType type) {
 }
 
 Rtt TranslateToFuzzReader::getSubType(Rtt rtt) {
-  if (getTypeSystem() == TypeSystem::Nominal) {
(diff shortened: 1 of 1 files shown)


src/tools/fuzzing/fuzzing.cpp
```diff
@@ -3041,6 +3041,12 @@ HeapType TranslateToFuzzReader::getSubType(HeapType type) {
 }
 
 Rtt TranslateToFuzzReader::getSubType(Rtt rtt) {
+  if (getTypeSystem() == TypeSystem::Nominal) {
```
tlively, line 3044: We should include `TypeSystem::Isorecursive` in this condition as well, since it will have the same problem.

tlively (approved):
LGTM modulo that comment. The latest upstream proposal gets rid of RTT depth entirely, so I doubt we will ever need to add more logic here.

src/tools/fuzzing/fuzzing.cpp
kripken: done.

PR #4606, opened 2022-04-21
[NominalFuzzing] Add a validation error on ref.cast's intended type

kripken:
`(ref.cast func (..))` for example is not allowed (and the binary format emitter will
crash when it looks for the heap type index).

### DIFF
--- a/src/wasm/wasm-validator.cpp
+++ b/src/wasm/wasm-validator.cpp
@@ -2375,6 +2375,8 @@ void FunctionValidator::visitRefTest(RefTest* curr) {
                     HeapType(),
                     curr,
                     "static ref.test must set intendedType field");
+    shouldBeTrue(
+      !curr->intendedType.isBasic(), curr, "ref.test must test a non-basic");
   }
 }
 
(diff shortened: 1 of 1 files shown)


tlively (commented):
We should make the same change for `ref.test` and other cast-like instructions. We can also add tests in `test/lit/validation`.

kripken:
Added `ref.test` and `br_on_cast*`.

I am a little torn about adding tests for this. I expect when wasm GC stabilitizes we'll have a solid test suite upstream and we'll migrate to that, so any specific tests here would end up unnecessary. For things that we are likely to regress and/or are tricky to debug I definitely think we should add even temporary tests, but validation stuff like this seems less critical to me. What do you think?

tlively (approved):
Seems fair about testing, although we haven't actually been good about pulling in upstream tests to my knowledge.

kripken:
Yeah, we should do better about pulling in upstream. I think we could have a utility that converts tests from arbitrary wat into structured wat (just read and write them using wabt, say) and then it should be easy.

PR #4610, opened 2022-04-22
[NominalFuzzing] Switch to nominal fuzzing by default

kripken:
This adds a parameter `NOMINAL` in the fuzzer, turns it on by default, and
applies `--nominal` when it is active. It also adds all the new GC passes to
the list of fuzzable passes.

This should be ready to land once #4601 #4609 do, as with them and all
the other recent PRs I ran the fuzzer overnight for well over 100,000 iterations
without issues.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -35,13 +35,11 @@ assert sys.version_info.major == 3, 'requires Python 3!'
 
 # parameters
 
-NOMINAL = True
+TYPE_SYSTEM_FLAG = '--nominal'
 
 # feature options that are always passed to the tools.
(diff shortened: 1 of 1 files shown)


scripts/fuzz_opt.py
```diff
@@ -35,9 +35,14 @@
 
 # parameters
 
+NOMINAL = True
```
tlively, line 38: Are we ready for something like this?

```python
# --structural, --nominal, or --hybrid
TYPE_SYSTEM = '--nominal'
```

kripken: Sounds good to generalize, done. That should be helpful later. (I haven't done any hybrid fuzzing yet, though.)

tlively (approved):

PR #4611, opened 2022-04-22
OptimizeInstructions: Refinalize after a cast removal

kripken:
See details in the comment in the code.

We have avoided any Refinalize calls in OptimizeInstructions, but the case
handled here requires it sadly. I considered moving it to another pass, but this
is a peephole optimization so there isn't really a better place.


tlively (approved):
Alternatively we could loosen our validation rules to allow expressions to be typed with supertypes of most precisely knowable type, but it's probably not worth losing that guaranteed type precision.

PR #4614, opened 2022-04-25
Lit tests: print failing modules

kripken:
This is probably not the right fix, but it's been extremely useful for me locally. With this change, the auto-update script will print out a failing module, which is nice if the failing module was one out of dozens in a single file - without this change, there is no hint as to which of the modules failed, just the filename.

I suspect this is wrong as it stops at the first failing module, but we don't want that for the check script, I assume. If this is the right direction I could look into fixing that part.

### DIFF
--- a/scripts/foreach.py
+++ b/scripts/foreach.py
@@ -39,6 +39,12 @@ def main():
         result = subprocess.run(new_cmd)
         if result.returncode != 0:
             returncode = result.returncode
+            # For ease of debugging, print out the module that failed (otherwise
+            # the user would need to sift through all the modules in the wat to
+            # find the failing one).
+            print("[Failing module]")
+            print(module)
+            break
     sys.exit(returncode)
 
 
--- a/scripts/foreach.py
+++ b/scripts/foreach.py
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
 import sys
 import subprocess
 
@@ -31,6 +32,7 @@ def main():
     tempfile = sys.argv[2]
     cmd = sys.argv[3:]
     returncode = 0
+    all_modules = open(infile).read()
     for i, (module, asserts) in enumerate(support.split_wast(infile)):
         tempname = tempfile + '.' + str(i)
         with open(tempname, 'w') as temp:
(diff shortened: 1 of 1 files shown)


tlively (commented):
I don't think we want to do this by default - it would make the diffs reported by filecheck for failed tests more confusing, or in the worst case could hide failures by printing the expected output.

How do you use this printing in your workflow? Normally if the total input with all modules is small enough, I just re-run the command including the `foreach.py` on the full input file while debugging. If the unmodified input file is too long, I usually run the same full command to find the last passing module, then comment out all but the interesting module in the input file.

kripken:
> How do you use this printing in your workflow?

For example, `test/lit/passes/cfp.wast` has 34 small modules. Say there is a crash bug in one, which I can fake now by adding `(foobar)` to a random module. Running the auto update script then shows this:
```
bad module element foo
[parse exception: unknown module element (at 97:2)]
Fatal: error parsing wasm
[..]
subprocess.CalledProcessError: Command '/home/azakai/Dev/3-binaryen/scripts/foreach.py /home/azakai/Dev/3-binaryen/test/lit/passes/cfp.wast /tmp/tmpzd48xs4n wasm-opt --nominal --remove-unused-names --cfp -all -S -o -' returned non-zero exit status 1.
```
There is no indication of which module out of the 34 failed, and line 97 is inside that module, not the absolute number in the file. Trying to debug this, I can run that foreach command manually (if I replace `wasm-opt` with `bin/wasm-opt`):
```
/home/azakai/Dev/3-binaryen/scripts/foreach.py /home/azakai/Dev/3-binaryen/test/lit/passes/cfp.wast /tmp/tmpzd48xs4n bin/wasm-opt --nominal --remove-unused-names --cfp -all -S
```
That emits this:
```
[..]
warning: no output file specified, not emitting output
warning: no output file specified, not emitting output
bad module element foo
[parse exception: unknown module element (at 97:2)]
Fatal: error parsing wasm
warning: no output file specified, not emitting output
warning: no output file specified, not emitting output
[..]
```
Still not clear enough. But with this patch, I get this:
```
warning: no output file specified, not emitting output
warning: no output file specified, not emitting output
bad module element foo
[parse exception: unknown module element (at 97:2)]
Fatal: error parsing wasm
[Failing module]
(module
  ;; CHECK:      (type $struct1 (struct_subtype (field (mut i32)) data))
  (type $struct1 (struct (mut i32)))
  ;; CHECK:      (type $struct2 (struct_subtype (field (mut i32)) (field f64) $struct1))
  (type $struct2 (struct_subtype (mut i32) f64 $struct1))
  ;; CHECK:      (type $none_=>_none (func_subtype func))
[..]
```
I can then copy the module text and search for it in the big wast. 

> Normally if the total input with all modules is small enough, I just re-run the command including the foreach.py on the full input file while debugging. If the unmodified input file is too long, I usually run the same full command to find the last passing module, then comment out all but the interesting module in the input file.

Not sure I follow this - how do you find out which is the bad module? (How do you even find the "last passing module"?)

tlively:
I see, thanks. I guess normally I only have problems with files with a handful of modules, so I can just count to e.g. the third module in the file and know that one has the problem. Maybe the most helpful and least dangerous option would be to print an error message with the line number of the bad module rather than its contents.

kripken:
Good idea, I implemented it that way now.

PR #4620, opened 2022-04-27
[NominalFuzzing] SignatureRefining can modify the IR while running a parallel analysis

kripken:
Normally ParallelFunctionAnalysis is just an analysis, and has no effects. However, in
SignatureRefining we actually do have side effects, due to an internal limitation of the
helper code it runs. This adds a template parameter to the class so users can note that
they do modify the IR. The parameter is added in the middle as it is easier to add this
param than to add the last one (the map).

### DIFF
--- a/src/ir/module-utils.h
+++ b/src/ir/module-utils.h
@@ -303,7 +303,7 @@ template<typename T> inline void iterImports(Module& wasm, T visitor) {
 // Helper class for performing an operation on all the functions in the module,
 // in parallel, with an Info object for each one that can contain results of
 // some computation that the operation performs.
-// The operation performend should not modify the wasm module in any way, by
+// The operation performed should not modify the wasm module in any way, by
 // default - otherwise, set the Mutability to Mutable. (This is not enforced at
 // compile time - TODO find a way - but at runtime in pass-debug mode it is
 // checked.)
--- a/src/ir/module-utils.h
+++ b/src/ir/module-utils.h
@@ -303,10 +303,13 @@ template<typename T> inline void iterImports(Module& wasm, T visitor) {
 // Helper class for performing an operation on all the functions in the module,
 // in parallel, with an Info object for each one that can contain results of
 // some computation that the operation performs.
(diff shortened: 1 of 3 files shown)


kripken:
(diff without whitespace is smaller)

src/ir/module-utils.h
```diff
@@ -303,10 +303,15 @@ template<typename T> inline void iterImports(Module& wasm, T visitor) {
 // Helper class for performing an operation on all the functions in the module,
 // in parallel, with an Info object for each one that can contain results of
 // some computation that the operation performs.
-// The operation performend should not modify the wasm module in any way.
-// TODO: enforce this
+// The operation performend should not modify the wasm module in any way, by
+// default - otherwise, ModifiesIR should be set to true. (This is not enforced
+// at compile time - TODO find a way - but at runtime in pass-debug mode it is
+// checked.)
+using Modifies = bool;
```
tlively, line 310: Could we use an enum here instead? I forget whether enums can be template parameters in C++17.

tlively (approved):

src/ir/module-utils.h
kripken: Hmm, in that case, what would the enum values be? I'm having trouble thinking of anything good.

One option:
```cpp
enum class Modifies {
  True,
  False
};
```
(or Yes/No etc) but I'm not sure that's better than the current code. Another is
```cpp
enum class PossibleModification {
  DoesModify,
  DoesNotModify
};
```
But that seems more verbose?

With stuff like mutability we have a natural pair of values for the true/false cases (Mutable/Immutable), but here I can't seem to find anything better than `Modifies(true) / Modifies(false)` as in the code now...

tlively: Maybe `ReadWrite` and `ReadOnly`? Alternatively `Mutable`/`Immutable`. Anyway a bool is probably good enough if you don't like those options.

kripken: Ok, sgtm to use the Mutability enum here, if it sounds right to your ear. Done.

PR #4621, opened 2022-04-27
RemoveUnusedModuleElements: Track CallRef/RefFunc more precisely

kripken:
If we see `(ref.func $foo)` that does not mean that `$foo` is reachable - we
must also see a `(call_ref ..)` of the proper type. Only after seeing both should
we mark the function as reachable, which this PR does.

This adds some complexity as we need to track intermediate state as we go,
since we could see the RefFunc before the CallRef or vice versa. We also
need to handle the case of a RefFunc without a CallRef properly: We cannot
remove the function, as the RefFunc must refer to it, but at least we can
empty out the body since we know it is never reached.

This removes an old wasm-opt test which is now superseded by a new lit
test.

On J2Wasm output this removes 3% of all functions, which account for
2.5% of total code size.


src/passes/RemoveUnusedModuleElements.cpp
```diff
@@ -199,15 +260,42 @@ struct RemoveUnusedModuleElements : public Pass {
...
+    // TODO: Compute this based on which ElementSegments are actually reachable,
+    //       and which functions have a call_indirect of the proper type.
     ElementUtils::iterAllElementFunctionNames(module, [&](Name& name) {
       roots.emplace_back(ModuleElementKind::Function, name);
     });
     // Compute reachability starting from the root set.
     ReachabilityAnalyzer analyzer(module, roots);
+
+    // RefFuncs that are never called are a special case: We cannot remove the
+    // function, since then (ref.func $foo) would not validate. But if we know
```
aheejin, line 272: Can't we use `(ref.null func)`?

aheejin (commented):
This does not handle a case of a signature whose `call_ref`s only exist in to-be-removed functions, right? For example, an unexported function `$a`'s signature is `vii` and it's only `call_ref`s are within `func $a`. I guess handling this case can be nontrivial and might not bring much benefit anyway, but just wanted to check.

src/passes/RemoveUnusedModuleElements.cpp
kripken: Hmm, I don't think so. The type of a null is nullable while `(ref.func $foo)` is non-nullable. Also, someone might do `ref.is_null` on the ref, even if the ref is never called.

kripken:
> This does not handle a case of a signature whose call_refs only exist in to-be-removed functions, right? 

If I understand your example, I think that case is handled. This pass starts from the reachable roots and goes from there, so anything we scan is reachable. A to-be-removed function would never be scanned, so we'd never see the RefFuncs or the CallRefs in it.

aheejin (approved):

tlively (approved):

PR #4624, opened 2022-04-29
Handle call.without.effects in RemoveUnusedModuleElements

kripken:
We assume a closed world atm in the GC space, but the `call.without.effects`
intrinsic sort of breaks that: that intrinsic looks like an import, but we really
need to care about what is sent to it even in a closed world:
```wat
(call $call-without-effects
  (ref.func $target-keep)
)
```
That reference cannot be ignored, as logically it is called just as if there
were a `call_ref` there. This adds support for that, fixing the combination of
#4621 and using `call.without.effects`.

This contains a bunch of new corner-case tests that are not directly related
to this PR, that I wrote while investigating things. I can split those out if
people prefer?

Also flip the vector of ref.func names to a set. I realized that in a very
large program we might see the same name many times.


test/lit/passes/remove-unused-module-elements-refs.wast
```diff
@@ -224,3 +224,780 @@
...
+    ;; This function is not reachable, but has a reference.
+  )
+
+  (func $target-C-2 (type $C)
+    ;; This function is not reachable.
+  )
+)
+
+(module
+  ;; A j2wasm-like itable pattern: An itable is an array of (possibly-null)
```
tlively, line 328: What is the purpose of the itable tests? Were they previously failing?

tlively (commented):
I do think it would be helpful to split the unrelated tests into a separate PR.

kripken:
Ok, extra tests removed for now, and other feedback addressed.

test/lit/passes/remove-unused-module-elements-refs.wast
kripken: They were not failing, but during investigation of this bug I thought it might be useful to have tests that check this real-world pattern. Maybe it is excessive though. (Tests removed from this PR meanwhile.)

tlively (approved):

tlively:
I don't love how complicated this intrinsic design makes things. Perhaps in the future we can consider removing intrinsics as part of parsing and storing the side effect info directly on the `call` expression to be taken into account by the side effect analyzer.

kripken:
Yes, I agree the current complexity due to intrinsics is not great. Storing the info on the call, or on the function, are other options we could consider. I think the main downside to them is that we'd need to add special code for parsing and writing the module (while right now we use the existing import parsing code for free).

PR #4625, opened 2022-04-29
Update the type section binary format

tlively:
Print subtype declarations using the standards-track format with a vector of
supertypes followed by a normal type declaration rather than our interim nominal
format that used alternative versions of the func, struct, and array forms.

Desugar the nominal format to additionally emit all the types into a single
large recursion group. Currently V8 is performing this desugaring, but after
this change and a future change that fixes the order of nominal types to ensure
supertypes precede subtypes, it will no longer need to.

### DIFF
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -253,7 +253,8 @@ void WasmBinaryWriter::writeTypes() {
     // Check whether we need to start a new recursion group. Recursion groups of
     // size 1 are implicit, so only emit a group header for larger groups. This
     // gracefully handles non-isorecursive type systems, which only have groups
-    // of size 1.
+    // of size 1 internally (even though nominal types are emitted as a single
+    // large group).
     auto currGroup = type.getRecGroup();
     if (lastGroup != currGroup && currGroup.size() > 1) {
       o << S32LEB(BinaryConsts::EncodedType::Rec) << U32LEB(currGroup.size());
@@ -1985,10 +1986,6 @@ void WasmBinaryBuilder::readTypes() {
       uint32_t groupSize = getU32LEB();
       if (getTypeSystem() == TypeSystem::Equirecursive) {
         throwError("Recursion groups not allowed with equirecursive typing");
-      }
-      if (getTypeSystem() == TypeSystem::Nominal && i > 0) {
-        throwError(
-          "Only a single recursion group is allowed under nominal typing");
       }
       if (groupSize == 0u) {
         // TODO: Support groups of size zero by shrinking the builder.
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -219,20 +219,35 @@ void WasmBinaryWriter::writeTypes() {
   if (indexedTypes.types.size() == 0) {
     return;
   }
-  // Count the number of recursion groups, which is always the number of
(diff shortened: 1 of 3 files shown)


src/wasm/wasm-binary.cpp
```diff
@@ -219,20 +219,35 @@ void WasmBinaryWriter::writeTypes() {
...
+    }
   }
   BYN_TRACE("== writeTypes\n");
   auto start = startSection(BinaryConsts::Section::Type);
   o << U32LEB(numGroups);
-  lastGroup = std::nullopt;
+  if (getTypeSystem() == TypeSystem::Nominal) {
+    // The nominal recursion group contains every type.
+    o << S32LEB(BinaryConsts::EncodedType::Rec)
+      << U32LEB(indexedTypes.types.size());
```
kripken, line 248: Does this duplicate what happens on (new) line 259, in the case of nominal typing? Naively I'd guess 259 is also reached since lastGroup begins as null (different from currGroup) and the size of currGroup is greater than 1 (since in nominal typing, the group contains all the types (as the comment on 258 says). So at first reading it looks like we emit duplicate data here. If that's not the case, then I'd guess while the nominal group logically contains every type, we don't actually contain that data in currGroup (I assume due to https://github.com/WebAssembly/binaryen/blob/e7c4f9da6d4a98f58b6320b5769f653a6bf26d3e/src/wasm/wasm-type.cpp#L1449-L1451)? If so, a comment would be good, as this is all fairly complex with all the three type systems...

src/wasm/wasm-binary.cpp
```diff
@@ -1869,6 +1877,9 @@ void WasmBinaryBuilder::readMemory() {
 void WasmBinaryBuilder::readTypes() {
   BYN_TRACE("== readTypes\n");
   TypeBuilder builder(getU32LEB());
+  if (getTypeSystem() == TypeSystem::Nominal && builder.size() > 1) {
+    throwError("Nominal type sections must have a single element");
```
kripken, line 1881: Is that element the singleton big rec group? But then this seems to overlap with the error on (new) line 1990, so I'm not sure.

kripken (commented):
lgtm % questions

src/wasm/wasm-binary.cpp
tlively: Right, we internally represent nominal types as each being in their own group of size one since that is equivalent to not having a recursion group for most purposes. That means that new line 159 is never reached for nominal types. The comment on new lines 254-256 tries to get at that, but I'll make it more explicit.

src/wasm/wasm-binary.cpp
tlively: Yes, that element is the big rec group. I guess that error on line 1990 is unreachable since we already have this check here. I'll remove it.

kripken (approved):
Thanks, sgtm with those changes!

kripken:
I see fuzzer errors with this, e.g.

```
./scripts/fuzz_opt.py --auto-initial-contents 7210303614564672651
```

Error:
```
/usr/local/google/home/azakai/Dev/binaryen/bin/wasm-opt a.wasm.temp.wasm --fuzz-exec-before
[parse exception: Recursion groups not allowed with equirecursive typing (at 0:14)]
```

tlively:
I would expect errors when trying to run nominal types on V8 (since the supertypes aren't ordered correctly), but that error is unexpected. Will investigate.

PR #4627, opened 2022-05-02
Missing include

martinRenou:
This makes a build on Windows fail with `Errors : 'isdigit': is not a member of 'std'`


kripken (approved):
Thanks!

PR #4631, opened 2022-05-02
Update nominal type ordering

tlively:
V8 requires that supertypes come before subtypes when it parses
isorecursive (i.e. standards-track) type definitions. Since 2268f2a we are
emitting nominal types using the standard isorecursive format, so respect the
ordering requirement.

### DIFF
--- a/src/ir/module-utils.cpp
+++ b/src/ir/module-utils.cpp
@@ -203,9 +203,10 @@ std::vector<HeapType> collectHeapTypes(Module& wasm) {
 }
 
 IndexedHeapTypes getOptimizedIndexedHeapTypes(Module& wasm) {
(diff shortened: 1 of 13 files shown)


src/ir/module-utils.cpp
```diff
@@ -236,7 +240,7 @@ IndexedHeapTypes getOptimizedIndexedHeapTypes(Module& wasm) {
       if (useCount != other.useCount) {
         return useCount < other.useCount;
       }
-      return index < other.index;
+      return index > other.index;
```
tlively, line 243: Switching this comparison makes the diff on the tests much smaller.

tlively (commented):
Diff is slightly smaller ignoring whitespace.

kripken (approved):

Issue #4632, opened 2022-05-03
test/passes/65536_locals_for_liveness.wasm is suuuuper slow


tlively:
To the point where I haven't been patient enough for it to finish locally and I've temporarily deleted it to unblock running the rest of the test suite. Did something change with coalesce-locals recently?

conrad-watt:
Just as a drive-by, maybe the change in https://github.com/WebAssembly/binaryen/pull/4567 is related? Presumably before this change the test wasn't actually running.

kripken:
Yes, it seems likely it is that PR. Running say `./auto_update_tests.py wasm-opt` pauses on it for a very long time which is indeed very noticeable...

cc @juj

juj:
I wonder if it would make sense to remove that test, since it is now testing an arbitrary limit that no longer exists? Originally the issue was large memory usage, but now it does look like the optimization pass is not fast enough to operate on a >64K locals case in a timely manner.

For end users that might have such many locals: if running the coalesce-locals pass for 64K locals is unrealistically slow, it is very likely also unrealistically slow for, say, 63K locals, so reinstating a limit on 64K does not seem based on actual data. Maybe there should be some kind of cmdline tuneable option that would enable users to choose a ceiling to disable that pass at? And by default disable liveness already at something like 8K or 32K or wherever a realistic perf. limit would happen to fall at?

conrad-watt:
@juj out of curiosity, given the original (very sensible) motivation for the change, is the [5.5M LOC example](https://github.com/WebAssembly/binaryen/pull/4567#issuecomment-1086068895) behaving better now that it's in place?

juj:
> @juj out of curiosity, given the original (very sensible) motivation for the change, is the https://github.com/WebAssembly/binaryen/pull/4567#issuecomment-1086068895 behaving better now that it's in place? Presumably it involves more than 64K locals.

I did test that the change resolved the build warnings that the user was having, but like mentioned in the thread, I was unable to reproduce the fault that the user had reported. We are in the process of getting an updated Unity version with the fix in to the customer to be able to test, but that will take some weeks to get through the release cycles.

tlively:
This test no longer seems to exist.

PR #4635, opened 2022-05-03
Replace 64K sparse matrix testcase with 8K

kripken:
Helps #4632

This makes it take 4 seconds instead of 5 minutes.


kripken:
This also makes tsan take just 16 minutes vs an hour before, so it resolves that issue as well...

tlively (approved):

PR #4636, opened 2022-05-03
Update StackCheck for memory64

sbc100:


src/passes/StackCheck.cpp
```diff
@@ -87,16 +87,23 @@ struct EnforceStackLimits : public WalkerPass<PostWalker<EnforceStackLimits>> {
     } else {
       handlerExpr = builder.makeUnreachable();
     }
+
+    BinaryOp GtPtr = BinaryOp::GtUInt32;
+    BinaryOp LtPtr = BinaryOp::LtUInt32;
+    if (stackPointer->type == Type::i64) {
+      GtPtr = BinaryOp::GtUInt64;
+      LtPtr = BinaryOp::LtUInt64;
+    }
```
kripken, line 96: This could use `src/ir/abstract.h`: `Abstract::getBinary(type, GtU)` will return the i32 or i64 version accordingly.

sbc100:
Added a test.

test/lld/basic_safe_stack64.s
```diff
@@ -0,0 +1,38 @@
+# Test that wasm-binaryen-finalize --check-stack-overflow correctly
+# inserts stack chekc handlers.
```
kripken, line 2:
```suggestion
# inserts stack check handlers.
```

sbc100:
> I think it would be good to test the pass directly, see the existing `./test/passes/stack-check_enable-mutable-globals.wast` to which we can add more contents.
> 
> I'd prefer that over the `.s` test, but I'd also be ok with both if you think it's worth ensuring the end-to-end behavior as well.

Done!

kripken (approved):
Ah, sorry, I didn't notice the old test was not `lit`. We're trying to only add new `lit` tests. lgtm with that to replace the one in `test/passes/`

test/lit/passes/stack-check-memory64.wast
```diff
@@ -0,0 +1,93 @@
...
+ ;; CHECK:      (type $i64_i64_=>_none (func (param i64 i64)))
+
+ ;; CHECK:      (global $sp (mut i64) (i64.const 0))
+ (global $sp (mut i64) (i64.const 0))
+ (func "use_stack" (result i64)
+  (global.set $sp (i64.const 42))
+  (global.get $sp)
+ )
+)
+;; CHECK:      (global $__stack_base (mut i64) (i64.const 0))
```
sbc100, line 19: @tlively  are these automatically generated checks supposed to sit after the module or inside it?

tlively: I think it's expected that they're after the module, since they're all that's left when the auto update script has finished matching up outputs with inputs.

It would be nice to improve the auto update script to produce nicer output, but it's already very complex. Rewriting it in a statically typed language rather than python might help manage the complexity and help us extend it.

tlively (approved):

PR #4637, opened 2022-05-03
chore: Add Cmake target for js_of_ocaml build target

phated:
I lost my previous patch and needed to recreate it, and I'm submitting this upstream so I don't lose it again. We need to have a special build target for [libbinaryen compiled for js_of_ocaml](https://github.com/grain-lang/libbinaryen) and our [binaryen.ml](https://github.com/grain-lang/binaryen.ml) bindings (also compiled with js_of_ocaml).

This shouldn't add any overhead to normal workflows, as you need to use emcmake with the specific target to run it and it would make my life a lot easier for generating everything in the Grain libraries. What do you think?


CMakeLists.txt
```diff
@@ -412,6 +412,32 @@ if(EMSCRIPTEN)
   set_property(TARGET binaryen_js PROPERTY CXX_STANDARD ${CXX_STANDARD})
   set_property(TARGET binaryen_js PROPERTY CXX_STANDARD_REQUIRED ON)
   install(TARGETS binaryen_js DESTINATION ${CMAKE_INSTALL_BINDIR})
+
+  # binaryen.js JavaScript variant
```
phated, line 416:
```suggestion
  # binaryen.es5.js JavaScript variant
```

src/js/binaryen.jsoo-extern-pre.js
```diff
@@ -0,0 +1,7 @@
+//Provides: binaryen
+var binaryen = {};
```
kripken, line 2: What are the differences between this and the main binaryen.js build? This line seems to be added, and also `-s WASM=0` - anything else? (If just those two then maybe we can add a build flag with an ifdef for each of them, which would avoid duplication.)

Also, what does this line do? I'm not sure why it's needed in your build but not the default.

phated: `WASM=0` is actually specified in the `binaryen_js` target, but even binaryen.js only uses the wasm version now.

A few things are going on with this line:
1. If a `var` matching `EXPORT_NAME` doesn't exist, the value is attached to the global, which doesn't work for js_of_ocaml
2. js_of_ocaml uses the `//Provides: ` syntax to treeshake values from the program

Other changes include:
1. `ENVIRONMENT` to only `web` and `worker` because js_of_ocaml tries to get smart about node globals and it breaks the usage.
2. Compiling to ES5, which was changed previously.
3. `EXPORT_ES6=0` because we need plain ES5 javascript until js_of_ocaml finishes their ES6 support
4. `NO_FILESYSTEM=1` - since node is not supported, this ensures there is no filesystem related code in the file that might make js_of_ocaml think it is running in node.

I'm happy to use ifdef or any other solution you want, this was just the patch that I had to generate the file we need.

CMakeLists.txt
```diff
@@ -412,6 +412,32 @@ if(EMSCRIPTEN)
...
+  target_link_libraries(binaryen.es5 "-s ENVIRONMENT=web,worker")
+  target_link_libraries(binaryen.es5 "--post-js ${CMAKE_CURRENT_SOURCE_DIR}/src/js/binaryen.js-post.js")
+  target_link_libraries(binaryen.es5 "--extern-pre-js ${CMAKE_CURRENT_SOURCE_DIR}/src/js/binaryen.jsoo-extern-pre.js")
+  target_link_libraries(binaryen.es5 optimized "--closure 1")
+  target_link_libraries(binaryen.es5 optimized "--closure-args \"--language_in=ECMASCRIPT6 --language_out=ECMASCRIPT5\"")
+  target_link_libraries(binaryen.es5 optimized "-flto")
+  target_link_libraries(binaryen.es5 debug "--profiling")
+  target_link_libraries(binaryen.es5 debug "-s ASSERTIONS")
+  set_property(TARGET binaryen.es5 PROPERTY CXX_STANDARD ${CXX_STANDARD})
+  set_property(TARGET binaryen.es5 PROPERTY CXX_STANDARD_REQUIRED ON)
```
phated, line 439:
```suggestion
```

Looks like these were removed between 106 and 107

src/js/binaryen.jsoo-extern-pre.js
kripken: I see, thanks. Ok, then I think a new build flag would be best. It sounds like these are specific to js_of_ocaml, so perhaps literally a new flag `JS_OF_OCAML`? Then we'd ifdef based on that on each of the bullet points you mention 1-2 and 1-4. That would avoid duplication and make it clear what the differences actually are (comments on each would be good as well).

phated: Thanks! Should be updated now (I also updated this branch to with `main`)

kripken (commented):
lgtm with a declaration of that new variable, something like
```cmake
// (comment with more details?)
option(JS_OF_OCAML "Build binaryen.js for js_of_ocaml" OFF)
```
I think that help text shows up in cmake's help for the project, or something like that.

phated:
Cool! Didn't know about that. TIL about `cmake -LAH` 😄

kripken (approved):
Great!

PR #4640, opened 2022-05-04
Setup git submodules so googletest is fetched, unbreaking alpine in releases

### LINKED ISSUE #4639: GitHub Releases missing Linux binaries
For versions 106 and 107, Mac and Windows binaries are attached to the GitHub Release, but the Linux binary that was present in earlier releases (e.g. https://github.com/WebAssembly/binaryen/releases/download/version_105/binaryen-version_105-x86_64-linux.tar.gz ) is missing.

kripken:
Without this the CMake step fails on not finding

`googletest/googletest/src/gtest_main.cc`

Fixes #4639 

This won't be tested in this PR. What's the best way to test it? Make a fake release I guess?

### DIFF
--- a/.github/workflows/create_release.yml
+++ b/.github/workflows/create_release.yml
@@ -111,6 +111,8 @@ jobs:
       with:
         python-version: '3.x'
     - uses: actions/checkout@v1
+      with:
+        submodules: true
     - name: start docker
       run: |
         docker run -w /src -dit --name alpine -v $PWD:/src node:lts-alpine
@@ -124,11 +126,6 @@ jobs:
 
     - name: install python dev dependencies
       run: ./alpine.sh pip3 install -r requirements-dev.txt
-
-    - name: git submodules
-      run: |
-        ./alpine.sh git submodule init
-        ./alpine.sh git submodule update
 
     - name: cmake
       run: |
--- a/.github/workflows/create_release.yml
+++ b/.github/workflows/create_release.yml
@@ -125,6 +125,11 @@ jobs:
     - name: install python dev dependencies
       run: ./alpine.sh pip3 install -r requirements-dev.txt
 
+    - name: git submodules
+      run: |
+        ./alpine.sh git submodule init
+        ./alpine.sh git submodule update
+
     - name: cmake
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/create_release.yml
```diff
@@ -125,6 +125,11 @@ jobs:
     - name: install python dev dependencies
       run: ./alpine.sh pip3 install -r requirements-dev.txt
 
+    - name: git submodules
+      run: |
+        ./alpine.sh git submodule init
+        ./alpine.sh git submodule update
+
```
sbc100, line 132: Better to modify the `checkout` step above. e.g:

```
    - uses: actions/checkout@v1                                                     
      with:                                                                         
        submodules: true      
```

sbc100 (approved):

kripken:
Any idea how to test this? I created a draft release "Testing alpine git integration" but it doesn't seem to start any actions,

https://github.com/WebAssembly/binaryen/releases/tag/untagged-dc99bed2246bfe1042b4 (details only shown in the edit display, https://github.com/WebAssembly/binaryen/releases/edit/untagged-dc99bed2246bfe1042b4)

There seems to be a previous draft release "test_tag" which *does* have builds for it, though. Maybe it's just slow?

sbc100:
That action is triggered by tag creation..  and I don't currently see a tag for your test release (I though releases even drafts were based on tags.. but I guess not).

I'm not sure this change needs testing though.. its fairly obviously correct.

kripken:
Hmm, I did create a tag when I drafted the release. Can see it on https://github.com/WebAssembly/binaryen/releases/edit/untagged-dc99bed2246bfe1042b4 - it is "alpgit". So that is odd...

I'd prefer to test it just to be sure we don't forget this next release, if somehow it doesn't work - the silent missing files from releases (happened in the last 2) are annoying. But yeah, this seems trivial, so if I can't figure it out I'll leave it I guess.

sbc100:
I looks like the draft release process creates the tag on publish.. so there is no tag as of yet.

kripken:
Oh, ok... I guess I'll publish it then, then delete it.

kripken:
Thanks, yes, I see actions are starting now...

PR #4642, opened 2022-05-04
Fix fuzzer's choosing of reference types

kripken:
1. Don't emit "`i31`" or "`data`" if GC is not enabled, as only the GC feature adds those.
2. Don't emit "`any`" without GC either. While it is allowed, fuzzer limitations prevent
  this atm (see details in comment - it's fixable).

Not sure why this wasn't noticed before, I guess we were just lucky...

### DIFF
--- a/src/tools/fuzzing/fuzzing.cpp
+++ b/src/tools/fuzzing/fuzzing.cpp
@@ -1919,7 +1919,12 @@ Expression* TranslateToFuzzReader::makeConst(Type type) {
           Nullability nullability = getSubType(type.getNullability());
           HeapType subtype;
           if (funcContext || nullability == Nullable) {
-            subtype = pick(HeapType::func, HeapType::i31, HeapType::data);
+            subtype = pick(FeatureOptions<HeapType>()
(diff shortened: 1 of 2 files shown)


src/tools/fuzzing/fuzzing.cpp
```diff
@@ -2900,7 +2905,13 @@ Type TranslateToFuzzReader::getSingleConcreteType() {
                      WeightedOption{Type::f32, VeryImportant},
                      WeightedOption{Type::f64, VeryImportant})
                 .add(FeatureSet::SIMD, WeightedOption{Type::v128, Important})
-                .add(FeatureSet::ReferenceTypes, Type::funcref, Type::anyref)
+                // Avoid Type::anyref without GC enabled, as in some contexts
+                // we will not be able to emit any constant for it (with GC, we
```
tlively, line 2909: Can't we use `ref.null`?

kripken: In this code path I suppose null is an option, yes, I can revert this part.

tlively (approved):

PR #4644, opened 2022-05-04
Parse the prototype nominal binary format

tlively:
In f124a11ca3 we removed support for the prototype nominal binary format
entirely, but that means that we can no longer parse older binary modules that
used that format. Fix this regression by restoring the ability to parse the
prototype binary format.


tlively:
* main
  * **PR #4644** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4644" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4644?utm_source=stack-comment).

kripken (approved):
lgtm with fuzzing

tlively:
The fuzzer is finding a lot of unrelated `unexpected false: array.init requires gc to be enabled...` and similar errors. I ran it a few times and didn't find anything else, but I also don't expect that it would find anything else since we never emit the binary format this PR parses.

PR #4645, opened 2022-05-05
[NominalFuzzing] Add typesystem flag to wasm-dis

kripken:
`wasm-dis` does enable all features by default, so we don't need the
feature flags, but we do need `--nominal` etc. since we emit such
modules now.


tlively (approved):
Thank you!

PR #4646, opened 2022-05-05
Revert "Reduce iterations required for dead argument elimination convergence"

arsnyder16:
Reverts WebAssembly/binaryen#4629


sbc100:
If you would prefer to fix forward rather than revert the current breakage is not actually hurting emscripten yet, its just preventing new changes from binaryen making their way to emscripten users. Just FYI.

arsnyder16:
@sbc100 How quickly do you need the fix ? I don’t want to hold anything up

sbc100:
I would say its worth reverting if its going to remain unfixed for more than a day or two.    Maybe safest just to land this revert if you are ok with that?

sbc100 (approved):

arsnyder16:
Yea no problem

PR #4647, opened 2022-05-05
[Fuzzer] Fix another reference types vs gc types issue

kripken:

We can't emit `HeapType::data` without GC. Fixing that by switching to `func`,
another problem was uncovered: `makeRefFuncConst` had a TODO to handle
the case where we need a function to refer to but have created none yet. In
fact that TODO was done at the end of the function. Fix up the logic in
between to actually get there.

### DIFF
--- a/src/tools/fuzzing/fuzzing.cpp
+++ b/src/tools/fuzzing/fuzzing.cpp
@@ -1856,50 +1856,50 @@ Literal TranslateToFuzzReader::makeLiteral(Type type) {
 }
 
 Expression* TranslateToFuzzReader::makeRefFuncConst(Type type) {
-  // ref.as_non_null is allowed in globals and we don't yet support func.ref in
-  // globals because we create globals before we create functions. As a result,
-  // we can only create non-nullable function references if we are in a function
-  // context for now.
-  // TODO: Generate trivial functions to support ref.func in globals.
-  assert(type.isNullable() || funcContext);
-  if (!funcContext || (type.isNullable() && oneIn(8))) {
-    return builder.makeRefNull(type);
-  }
-
   auto heapType = type.getHeapType();
   if (heapType == HeapType::func) {
     // First set to target to the last created function, and try to select
     // among other existing function if possible.
-    Function* target = funcContext->func;
-    if (!wasm.functions.empty() && !oneIn(wasm.functions.size())) {
+    Function* target = funcContext ? funcContext->func : nullptr;
+    // If there is no last function, and we have others, pick between them. Also
+    // pick between them with some random probability even if there is a last
+    // function.
+    if (!target || (!wasm.functions.empty() && !oneIn(wasm.functions.size()))) {
       target = pick(wasm.functions).get();
     }
-    return builder.makeRefFunc(target->name, target->type);
-  } else {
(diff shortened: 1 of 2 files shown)


src/tools/fuzzing/fuzzing.cpp
```diff
@@ -1856,50 +1856,50 @@ Literal TranslateToFuzzReader::makeLiteral(Type type) {
...
   if (heapType == HeapType::func) {
     // First set to target to the last created function, and try to select
     // among other existing function if possible.
-    Function* target = funcContext->func;
-    if (!wasm.functions.empty() && !oneIn(wasm.functions.size())) {
+    Function* target = funcContext ? funcContext->func : nullptr;
+    // If there is no last function, and we have others, pick between them. Also
+    // pick between them with some random probability even if there is a last
+    // function.
+    if (!target || (!wasm.functions.empty() && !oneIn(wasm.functions.size()))) {
```
tlively, line 1867: Why does the probability of picking a random function decrease with the number of functions? I would expect it to be constant or maybe even increase with the number of functions.

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -1856,50 +1856,50 @@ Literal TranslateToFuzzReader::makeLiteral(Type type) {
...
-      builder.makeFunction(Names::getValidFunctionName(wasm, "ref_func_target"),
-                           heapType,
-                           {},
-                           builder.makeUnreachable()));
-    return builder.makeRefFunc(func->name, heapType);
   }
+  // We don't have a matching function, so create a null with high probability
+  // if the type is nullable or otherwise create and cast a null with low
+  // probability.
+  if ((type.isNullable() && !oneIn(8)) || oneIn(8)) {
```
tlively, line 1889: Should we bump up the odds on the second `oneIn(8)` since it will produce a trapping code sequence?

tlively (approved):
lgtm % comments.

src/tools/fuzzing/fuzzing.cpp
kripken: I think the probability does increase with the number of functions. If there are say 100 functions then `oneIn(100)` is `1/100`, and then `!oneIn(100)` is `99/100`. I think the code is trying to give an equal chance across the combined set of the current function context + all previous ones.

src/tools/fuzzing/fuzzing.cpp
kripken: Hmm, makes sense. I'd rather do it in a separate PR though to keep this focused on not changing anything to the distribution (aside from the bugfix)?

kripken: (What would a good value be, though?)

src/tools/fuzzing/fuzzing.cpp
tlively: Oh yeah the `!` confused me. Sounds good!

src/tools/fuzzing/fuzzing.cpp
tlively: I would probably pick 16, although of course that's not based on any particular principle.

PR #4649, opened 2022-05-06
Add SubTypes::getAllSubTypes variant which includes the type itself

kripken:
This also includes the type itself in the returned vector. This will be
useful in a future PR.


src/ir/subtypes.h
```diff
@@ -51,6 +54,13 @@ struct SubTypes {
     return ret;
   }
 
+  // Like getAllSubTypes, but also includes the type itself.
+  std::vector<HeapType> getAllSubTypesInclusive(HeapType type) {
```
tlively, line 58: How about we name this `getAllSubtypes` and rename the current `getAllSubTypes` to `getStrictSubtypes`?

(Changing to a lowercase `t` in `Subtype` is intentional, but also we can leave that to be a global renaming in a future PR)

test/example/type-builder-nominal.cpp
```diff
@@ -433,6 +435,26 @@ void test_subtypes() {
       built = *builder.build();
     }
     assert(LUB(built[1], built[0]) == built[0]);
+
+    // Test SubTypes utility code, which also verifies the above. Create a tiny
+    // module that uses the types in the locals of a function, so the types are
+    // used.
```
tlively, line 441: I would strongly prefer adding this as a gtest. test/gtest/type-builder.cpp already tests many other type-adjacent utilities like this, so this could be added to that file.

kripken: Ok, moved to gtest.

src/ir/subtypes.h
kripken: Ok, renamed to `getStrictSubTypes`.

(About removing the capital `T`, I'm not opposed to that, but is it the right spelling? Is "subtype" a single world in English?)

tlively: Yes, "subtype" is a single word.

test/gtest/type-builder.cpp
```diff
@@ -498,3 +499,31 @@ TEST_F(EquirecursiveTest, CanonicalizeBasicTypes) {
...
 }
+
+// Test SubTypes utility code.
+TEST_F(NominalTest, testSubTypes) {
+  // Build type types, the second of which is a subtype.
+  TypeBuilder typeBuilder(2);
+  typeBuilder[0] = Struct({Field(Type::anyref, Immutable)});
+  typeBuilder[1] = Struct({Field(Type::funcref, Immutable)});
+  typeBuilder[1].subTypeOf(typeBuilder[0]);
+  auto built = *typeBuilder.build();
```
tlively, line 510: Nit: We usually use the name `builder` for the `TypeBuilder`s in these tests.

test/gtest/type-builder.cpp
```diff
@@ -498,3 +499,31 @@ TEST_F(EquirecursiveTest, CanonicalizeBasicTypes) {
...
+  Module wasm;
+  Builder wasmBuilder(wasm);
+  wasm.addFunction(wasmBuilder.makeFunction(
+    "func",
+    Signature(Type::none, Type::none),
+    {Type(built[0], Nullable), Type(built[1], Nullable)},
+    wasmBuilder.makeNop()));
+  SubTypes subTypes(wasm);
+  auto subTypes0 = subTypes.getStrictSubTypes(built[0]);
+  assert(subTypes0.size() == 1 && subTypes0[0] == built[1]);
```
tlively, line 523: Using the gtest macros will ensure that we get nice output and don't stop other tests from running if this test ever starts failing. The `ASSERT` macro variants will return from the test function if they fail and the `EXPECT` macro variants will continue to execute the rest of the function even if they fail, so generally it's best to prefer `EXPECT` except in cases where a failed check means the rest of the test isn't meaningful, like for a bounds check.

```suggestion
  ASSERT_EQ(subTypes0.size(), 1u);
  EXPECT_EQ(subTypes0[0], built[1]);
```

src/ir/subtypes.h
kripken: Is "supertype" also a single word?

Looks like we consider them both as type words atm in the codebase. But sgtm to fix to whatever is the common usage.

test/gtest/type-builder.cpp
kripken: Thanks, sounds good. Rewrote to use those.

src/ir/subtypes.h
tlively: Yes, supertype is also a single word.

test/gtest/type-builder.cpp
```diff
@@ -498,3 +499,32 @@ TEST_F(EquirecursiveTest, CanonicalizeBasicTypes) {
...
+    Signature(Type::none, Type::none),
+    {Type(built[0], Nullable), Type(built[1], Nullable)},
+    wasmBuilder.makeNop()));
+  SubTypes subTypes(wasm);
+  auto subTypes0 = subTypes.getStrictSubTypes(built[0]);
+  EXPECT_TRUE(subTypes0.size() == 1 && subTypes0[0] == built[1]);
+  auto subTypes0Inclusive = subTypes.getAllSubTypes(built[0]);
+  EXPECT_TRUE(subTypes0Inclusive.size() == 2 &&
+              subTypes0Inclusive[0] == built[1] &&
+              subTypes0Inclusive[1] == built[0]);
```
tlively, line 527: Splitting these up into separate checks would provide more helpful error messages if the test ever fails.

tlively (approved):

test/gtest/type-builder.cpp
kripken: I see, yes. It would make the code more verbose, though, as we'd need to size check in each one, etc. I don't have a strong sense for what's the right tradeoff here. If you feel strongly I'm happy to add it as a followup.

PR #4652, opened 2022-05-09
Add BUILD_TOOLS compile flag

TheGreatRambler:
Adds a `BUILD_TOOLS` compile flag so that compiling the library only is possible. Saves both time and storage in cases when the library is embedded in another program. Is set by default


CMakeLists.txt
```diff
@@ -270,7 +273,9 @@ else()
   if(WIN32)
     add_compile_flag("-D_GNU_SOURCE")
     add_compile_flag("-D__STDC_FORMAT_MACROS")
-    add_link_flag("-Wl,--stack,8388608")
+    if(NOT CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+      add_link_flag("-Wl,--stack,8388608")
+    endif()
```
kripken, line 278: This change looks unrelated?

kripken (commented):
Looks like a reasonable option to add.

Curious if people have thoughts on the name of the new option? cc @sbc100

CMakeLists.txt
```diff
@@ -37,6 +37,9 @@ option(BUILD_TESTS "Build GTest-based tests" ON)
 # Turn this on to build binaryen.js as ES5, with additional compatibility configuration for js_of_ocaml.
 option(JS_OF_OCAML "Build binaryen.js for js_of_ocaml" OFF)
 
+# Turn this off to build only the library
+option(BUILD_TOOLS "Build tools" ON)
```
sbc100, line 41: Maybe group this with `BUILD_TESTS` above?

sbc100 (commented):
Seems reasonable to me.

CMakeLists.txt
TheGreatRambler: Ah my bad, should have made a separate PR for this. `--stack` doesn't exist as a linker option on clang, and because clang in mingw64 isn't considered MSVC, this flag is used and causes errors on linking

TheGreatRambler: Pushed a new commit to add back the stack, using `-stack_size` which clang supports

sbc100: Can you make this a completely separate PR since it seems unrelated?

Also, I think maybe using `if(MSVC)` is better than checking or `NOT Clang` here?

TheGreatRambler: `MSVC` is not set but clang is in my case. Perhaps that's an edge case not worth supporting but I'm using a standard clang install with the msys shell which works for my other libs. Also yeah I'll split that into another PR, how can I do that

sbc100:
I still think the BUILD_TOOLS change is worth landing.

PR #4653, opened 2022-05-09
[Fuzzer] Reduce trap probability in function ref fallback code

kripken:
Also improve comments.

As suggested in #4647


src/tools/fuzzing/fuzzing.cpp
```diff
@@ -1885,15 +1885,18 @@ Expression* TranslateToFuzzReader::makeRefFuncConst(Type type) {
   }
   // We don't have a matching function, so create a null with high probability
   // if the type is nullable or otherwise create and cast a null with low
-  // probability.
-  if ((type.isNullable() && !oneIn(8)) || oneIn(8)) {
+  // probability (which would trap).
+  if ((type.isNullable() && !oneIn(16)) || oneIn(16)) {
```
tlively, line 1889: I would keep the probability of using a `ref.null` for a nullable reference low, too. It's more interesting to have a non-null value for a nullable reference than a null value for a non-null reference.

tlively (approved):

src/tools/fuzzing/fuzzing.cpp
kripken: Ok, I adjusted that probability now.

I also made us not trap if no value is returned - we can just have a nop there instead of an unreachable.

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -1883,25 +1883,26 @@ Expression* TranslateToFuzzReader::makeRefFuncConst(Type type) {
...
-  // the last option of a cast of a null that will trap even if the reference is
-  // not called.
-  auto* func = wasm.addFunction(
-    builder.makeFunction(Names::getValidFunctionName(wasm, "ref_func_target"),
-                         heapType,
-                         {},
-                         builder.makeUnreachable()));
+  // As a final option, create a new function with the correct signature. If it
+  // returns a value, write a trap as we do not want to create any more code
+  // here (we might end up recursing). Note that a trap is still better than the
```
tlively, line 1899: It would be good to be more specific about what "better" means here. I guess it means something like "executes more code before trapping"?

tlively (approved):

src/tools/fuzzing/fuzzing.cpp
kripken: Yes. Clarified the comment now.

tlively: lgtm!

PR #4659, opened 2022-05-11
Global Struct Inference optimization idea

kripken:
This implements an idea for devirtualization in the megamorphic case: when we
know two function references are possible, we could in theory do this:
```wat
(select
  (ref.func A)
  (ref.func B)
  (ref.eq
    (..ref value..)
    (ref.func A))
```
That is, compare to one of the values, and emit the two possible values there.
Other optimizations can then turn a `call_ref` on this select into an if over
two direct calls, and two direct calls may be more easily optimizable.

Just a draft - uploading for experimentation.


src/passes/GlobalStructInference.cpp
```diff
@@ -0,0 +1,239 @@
...
+        typeGlobals[type].push_back(global->name);
+      }
+    }
+
+    // A struct.get might also read from any of the subtypes. As a result, an
+    // unoptimizable type makes all its supertypes unoptimizable as well.
+    // TODO: this could be specific per field (and not all supers have all
+    //       fields)
+    auto unoptimizableCopy = unoptimizable;
+    for (auto type : unoptimizableCopy) {
```
tlively, line 136: Why not iterate over `unoptimizable`? It's not modified in the loop.

src/passes/GlobalStructInference.cpp
```diff
@@ -0,0 +1,239 @@
...
+      }
+    }
+
+    // Similarly, propagate global names: if one type has [global1], then a get
+    // of any supertype might access that, so propagate to them.
+    auto typeGlobalsCopy = typeGlobals;
+    for (auto& [type, globals] : typeGlobalsCopy) {
+      auto curr = type;
+      while (1) {
+        auto super = curr.getSuperType();
```
tlively, line 153: It might be useful to add an `iterSupertypes` utility for this common loop.

tlively (commented):
Comments so far. Looking at the tests now.

test/lit/passes/gsi.wast
```diff
@@ -0,0 +1,584 @@
...
+        (ref.null $struct)
+      )
+    )
+  )
+)
+
+;; We ignore imports, as we assume a closed world, but that might change in the
+;; future. For now, we will optimize here.
+;; TODO: test a reference type here that is imported, and not an i32, but that
+;;       is all that validates atm
```
tlively, line 178: We don't allow reference-typed imports? That seems like a bug.

test/lit/passes/gsi.wast
```diff
@@ -0,0 +1,584 @@
...
+    (drop
+      (struct.get $struct 0
+        (ref.null $struct)
+      )
+    )
+  )
+)
+
+;; One global for each of the type and the subtype. The optimization will pick
+;; between their 2 values.
```
tlively, line 482: It would also be good to test with globals for two distinct subtypes of a common supertype.

test/lit/passes/gsi.wast
```diff
@@ -0,0 +1,584 @@
...
+    )
+    (drop
+      (struct.get $super-struct 0
+        (ref.null $super-struct)
+      )
+    )
+  )
+)
+
+;; One global has a non-constant field, so we cannot optimize.
```
tlively, line 542: I would be good to add a test that mutable fields also prevent optimization.

src/passes/GlobalStructInference.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+    // We cannot optimize types that appear in a struct.new in a function, which
+    // we just collected and merge now.
+    HeapTypes unoptimizable;
+
+    for (auto& [func, types] : analysis.map) {
+      for (auto type : types) {
+        unoptimizable.insert(type);
+      }
+    }
```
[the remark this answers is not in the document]
kripken: I think I tried this and ended up with some trickiness in the template type... might require some C++ magic.

src/passes/GlobalStructInference.cpp
kripken: Good point, done.

test/lit/passes/gsi.wast
kripken: Added now.

kripken:
I experimented with combing this with GUFA, and saw no benefit. While doing both at once should help in theory, perhaps those situations are too rare. Given that, perhaps the best thing is to move forward with this in a separate/module way as it is written in this PR. We can land that and test it on real-world code, and after GUFA lands we can consider a refactor at that point.

test/lit/passes/gsi.wast
```diff
@@ -0,0 +1,809 @@
...
+      )
+    )
+  )
+)
+
+;; We ignore imports, as we assume a closed world, but that might change in the
+;; future. For now, we will optimize here.
+;; TODO: We can't import (ref $struct) here due to
+;;       https://github.com/WebAssembly/binaryen/issues/4676
+;;       So we test an imported i32 for now.
```
tlively, line 216: This has since been resolved!

tlively (approved):

test/lit/passes/gsi.wast
kripken: Right, thanks!

PR #4660, opened 2022-05-11
OptimizeInstructions: Turn call_ref of a select into an if over two direct calls 

kripken:
Just a draft, uploading for experimentation.

### DIFF
--- a/src/passes/Directize.cpp
+++ b/src/passes/Directize.cpp
@@ -63,9 +63,7 @@ struct FunctionDirectizer : public WalkerPass<PostWalker<FunctionDirectizer>> {
     // Emit direct calls for things like a select over constants.
     if (auto* calls = CallUtils::convertToDirectCalls(
           curr,
-          [&](Expression* target) {
-            return getNameInTable(target, flatTable);
-          },
+          [&](Expression* target) { return getNameInTable(target, flatTable); },
           *getFunction(),
           *getModule())) {
       replaceCurrent(calls);
@@ -93,7 +91,7 @@ private:
   // index in the given flat table. If so, return the name there, and otherwise
   // a null name.
   Name getNameInTable(Expression* curr,
-                    const TableUtils::FlatTable& flatTable) {
+                      const TableUtils::FlatTable& flatTable) {
     auto* c = curr->dynCast<Const>();
     if (!c) {
       return Name();
--- a/src/passes/Directize.cpp
+++ b/src/passes/Directize.cpp
@@ -63,7 +63,9 @@ struct FunctionDirectizer : public WalkerPass<PostWalker<FunctionDirectizer>> {
     // Emit direct calls for things like a select over constants.
     if (auto* calls = CallUtils::convertToDirectCalls(
           curr,
-          [&](Expression* target) { return getTargetInfo(target, flatTable); },
+          [&](Expression* target) {
+            return getTargetInfo(target, flatTable, curr);
+          },
           *getFunction(),
           *getModule())) {
       replaceCurrent(calls);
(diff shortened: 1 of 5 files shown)


kripken:
Ok, this is now ready for review. It refactors the existing Directize logic into a helper that can be used from both places (for CallIndirect and CallRef respectively).

This is NFC for CallIndirect, which Directize did before, but while doing this I realized that in the "trap" case we don't need to add drops of `local.get`s of the params. So the directize test gets a little shorter in the output. But functionally it is the same.

src/passes/call-utils.h
```diff
@@ -0,0 +1,153 @@
...
+    // TODO: Perhaps emitting a direct call for one arm is enough even if the
+    //       other remains indirect?
+    return nullptr;
+  }
+
+  auto& operands = curr->operands;
+
+  // We must use the operands twice, and also must move the condition to
+  // execute first; use locals for them all. While doing so, if we see
+  // any are unreachable, stop trying to optimize and leave this for DCE.
```
tlively, line 101: The comment makes it sound like this is all happening in one loop, but it happens in separate loops. `operandLocals` is not used until down in the next loop.

src/passes/call-utils.h
```diff
@@ -0,0 +1,153 @@
...
+    for (Index i = 0; i < numOperands; i++) {
+      newOperands[i] =
+        builder.makeLocalGet(operandLocals[i], operands[i]->type);
+    }
+    return newOperands;
+  };
+
+  auto makeCall = [&](IndirectCallInfo info) -> Expression* {
+    if (std::get_if<Trap>(&info)) {
+      return builder.makeUnreachable();
```
tlively, line 134: Why not handle `Trap` earlier where we handled `Unknown`?

src/passes/call-utils.h
```diff
@@ -0,0 +1,153 @@
...
+    } else {
+      return builder.makeCall(std::get<Known>(info).target,
```
tlively, line 136: Did we solve the problem where `std::get` pulled in some library code that didn't exist on the Mac platforms we were targeting?

src/passes/Directize.cpp
```diff
@@ -59,62 +60,19 @@ struct FunctionDirectizer : public WalkerPass<PostWalker<FunctionDirectizer>> {
...
+    if (auto* calls = CallUtils::convertToDirectCalls(
+          curr,
+          [&](Expression* target) {
+            return getTargetInfo(target, flatTable, curr);
+          },
+          *getFunction(),
+          *getModule())) {
+      replaceCurrent(calls);
+      // Note that types may have changed, as the utility here can add locals
+      // which require fixups if they are non-nullable, for example.
```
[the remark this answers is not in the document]
kripken: I think it's probably not worth it. It would add complexity to the API, and this is a fairly rare optimization, and the cost of the fixups we do are linear and fast.

src/passes/call-utils.h
kripken: Good catch! This was an issue in the code before it was moved in this PR, but worth fixing. Done.

src/passes/call-utils.h
kripken: I hope so :smile: A grep shows 3 places we use `std::get` before this PR adds another that makes it 4.

src/passes/call-utils.h
tlively: If we checked for `Trap` where we check for `Unknown`, could we either replace the entire expression with `unreachable` at that point or skip the optimization to let DCE do that? That would save us the work of going through and doing the rest of the optimization.

kripken: Oh, the trap might only happen in one arm:
```wat
(call_indirect
  (select
    (i32.const 1) ;; goes to $valid_target
    (i32.const 99999) ;; invalid
    (..condition..)))

=>

(if
  (..condition..)
  (call $valid_target)
  (unreachable))
```
We have a [test for that](https://github.com/WebAssembly/binaryen/blob/ac2190c1bd1a0be12a1d25e62695f791f270d050/test/lit/passes/directize_all-features.wast#L619-L659) (existing before this PR).

tlively: oh I see, thanks!

tlively (approved):

PR #4662, opened 2022-05-12
Ensure symmetric results in PossibleConstantValues

kripken:
Previously we could return different results depending on the order we
noted things:
```js
note(anyref.null);
note(funcref.null);
get() => anyref.null

note(funcref.null);
note(anyref.null);
get() => funcref.null
```
This is correct, as nulls are equal anyhow, and any could be used in
the location we are optimizing. However, it can lead to nondeterminism
if the caller's order of notes is nondeterministic. That is the case in
DeadArgumentElimination, where we scan functions in parallel, then
merge them without special ordering.

To fix this, make the note operation symmetric. That seems simplest and
least likely to be confusing. We can use the LUB to do that.

To avoid duplicating the null logic, refactor `note()` to use `combine()`.

### DIFF
--- a/src/ir/possible-constant.h
+++ b/src/ir/possible-constant.h
@@ -112,7 +112,7 @@ public:
       auto otherType = other.getConstantLiteral().type.getHeapType();
       auto lub = HeapType::getLeastUpperBound(type, otherType);
       if (lub != type) {
-        value = Literal::makeNull(Type(type, Nullable));
+        value = Literal::makeNull(Type(lub, Nullable));
         return true;
       }
       return false;
--- a/src/ir/possible-constant.h
+++ b/src/ir/possible-constant.h
@@ -105,8 +105,11 @@ public:
       return true;
     }
 
-    // Nulls compare equal, but we must combine them into a null of the LUB,
-    // both for determinism (the order should not matter) and to validate.
+    // Nulls compare equal, and we could consider any of the input nulls as the
+    // combination of the two (as any of them would be valid to place in the
+    // location we are working to optimize). In order to have simple symmetric
+    // behavior here, which does not depend on the order of the inputs, use the
+    // LUB.
     if (isNull() && other.isNull()) {
       auto type = getConstantLiteral().type.getHeapType();
       auto otherType = other.getConstantLiteral().type.getHeapType();
--- a/src/ir/possible-constant.h
+++ b/src/ir/possible-constant.h
@@ -72,22 +72,9 @@ public:
 
   // Note either a Literal or a Name.
   template<typename T> void note(T curr) {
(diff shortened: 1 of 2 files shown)


tlively (approved):
LGTM, although I guess we will have to update this code somehow if we remove `func <: any` and end up with multiple unrelated nulls.

aheejin:
Is `func` not a subtype anymore? And I noticed `externref` was deleted from Binaryen recently. Have we decided to use `anyref` for that purpose and get rid of `externref` eventually?

src/ir/possible-constant.h
```diff
@@ -118,6 +105,22 @@ struct PossibleConstantValues {
...
 
+    // Nulls compare equal, and we could consider any of the input nulls as the
+    // combination of the two (as any of them would be valid to place in the
+    // location we are working to optimize). In order to have simple symmetric
+    // behavior here, which does not depend on the order of the inputs, use the
+    // LUB.
+    if (isNull() && other.isNull()) {
+      auto type = getConstantLiteral().type.getHeapType();
+      auto otherType = other.getConstantLiteral().type.getHeapType();
+      auto lub = HeapType::getLeastUpperBound(type, otherType);
```
aheejin, line 116: Can we directly use `Type::getLeastUpperBound`?

aheejin (approved):

tlively:
> Is `func` not a subtype anymore?

`func` is still a subtype of `any` for now, although we're discussing changing this: https://github.com/WebAssembly/gc/issues/293

> And I noticed `externref` was deleted from Binaryen recently. Have we decided to use `anyref` for that purpose and get rid of `externref` eventually?

Yep, we decided to make `extern` and `any` the same type, so we didn't need both of them in Binaryen anymore.

src/ir/possible-constant.h
kripken: We can, but I think the clearer thing is to LUB on the heap type, because nullability is not relevant here... but maybe the real issue here is that `makeNull` gets a type and not a heap type - if it got a heap type then we wouldn't have two options here, which seems better. Let me look into that as a refactor PR.

aheejin: Both of the types are null at this point. Doesn't that mean both types are nullable anyway here?

kripken: Yes, exactly, the types must be nullable anyhow. So we really just need the least upper bound of the heap type, which is the only thing that can be different between them.

After #4664 the code here will be something like this, which I think is good:
```cpp
auto lub = HeapType::getLeastUpperBound(type, otherType)
if (lub != type) {
  value = Literal::makeNull(lub);
  return true;
}
```

PR #4665, opened 2022-05-13
[Nominal Fuzzing] Ensure local types are always included in collectHeapTypes

kripken:
See detailed comments in the code and test for the specifics, but basically,
this is to ensure we work properly when updating signature types. We are in
an awkward situation there in which we've updated `local.get` types, but not
updated signature types yet at some point in time (since we have separate
utility code for each - we can't do it all at once monolothically). If a `local.get`
is the only place a type is used, then the old code would miss it and assert
later.

This is not 100% elegant, but it is the smallest and simplest solution I can
think of, at least. Other ideas welcome!

The change in an unrelated test is just a minor ordering change - the new
`.include(type)` call ends up changing the order in which we notice types,
which changes the order in which we emit them.


src/ir/module-utils.cpp
```diff
@@ -75,6 +75,42 @@ struct CodeScanner
...
+      // only place in the entire program that uses this type then we'd run into
+      // an internal error later, so make sure the type is included here.
+      //
+      // This problem is specific to local.get because the type of local.get is
+      // dependent on the function's signature. Therefore local.get types must
+      // be changed atomically with a signature change to one of its params.
+      // However, for modularity we have separate utility code for each and not
+      // a single big monolithic helper that does both, so we need to be a
+      // little flexible here to handle IR that is partially updated,
+      // specifically, has local.get types updated but not the signature yet.
```
tlively, line 99: I took a look at SignatureRefining.cpp, but it wasn't obvious to me how the code modifications interact with heap type collection, so a more detailed walkthrough of the problem would be helpful.

Does the fact that `LUB::getResultsLUB` have side effects play into this? We might want to consider having it return a list of Expressions that need to be updated rather than having it perform the updates itself to keep that parallel analysis read-only.

tlively (commented):
It seems like a pretty bad abstraction leak to have extra logic here to work around a particular pass violating typing invariants, so I think it would be worth trying to understand how we can fix the pass instead.

src/ir/module-utils.cpp
kripken: Sorry, yeah, it's a little complicated, both what the underlying issue is but also how the fuzz testcase actually manages to cause the underlying issue to hit an error.

The key issue is that `local.get` types are tied to the function signature. That means that we need to change them atomically when we change the signature. More specifically, we update `local.get` in `TypeUpdating::updateParamTypes()`, which if we are changing function signatures means we do a "partial update" - the IR will only be in a consistent state after we also change the signatures, which `GlobalTypeRewriter::updateSignatures` does later. So another approach here would be to avoid a partial update and leave more to that later step. That is more code, but might be cleaner - I pushed that now.

Summary of current state of the code in this PR:

Optionally avoid updating types in `TypeUpdating::updateParamTypes()`. That update is incomplete if the function signature is also changing, which is the case in SignatureRefining (but not DeadArgumentElimination). "Incomplete" means that we updated the `local.get` type, but the function signature does not match yet. That incomplete state can hit an internal error in `GlobalTypeRewriter::updateSignatures` where it updates types. To avoid that, do the entire full update only there (in `GlobalTypeRewriter::updateSignatures`).

 (diff is smaller without whitespace).

> Does the fact that LUB::getResultsLUB have side effects play into this? 

The refinalize there helps the fuzz bug hit the error, but it's not the key issue (changes there would just be a workaround, in other words).

kripken: Friendly ping on this PR.

test/lit/passes/signature-refining.wast
```diff
@@ -674,3 +674,50 @@
...
+      (block (result (ref ${}))
+        (unreachable)
+      )
+      (i32.const 0)
+    )
+    ;; Write something of type eqref into $ref. When we refine the type of the
+    ;; parameter from eqref to ${} we must do something here, as we can no
+    ;; longer just write this (ref.null eq) into a parameter of the more
+    ;; refined type. While doing so, we must not be confused by the fact that
+    ;; there is nothing remaining in the module of type ${} (since as mentioned
```
tlively, line 715: This comment looks like it might be out of date since we no longer observe old types here.

tlively (approved):

PR #4668, opened 2022-05-16
Allow TypeBuilder::grow to take 0 as an argument

### LINKED ISSUE #4667: Roundtripping tiny nominal modules can fail
Testcase:
```wat
(module
 (func $0)
)
```
Command:
```shell
$ bin/wasm-opt -all --nominal c.wat --roundtrip
wasm-opt: src/wasm/wasm-type.cpp:2813: void wasm::TypeBuilder::grow(size_t): Assertion `size() + n > size()' failed.
Aborted
```

cc @tlively - looks like `n` is zero there. Is that a larger problem, or should we just do nothing when growing by size 0?

tlively:
There's no reason not to allow growing by zero slots, but previously doing so
would trigger an assertion. This caused a crash when roundtripping a trivial
module.

Fixes #4667.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -2810,7 +2810,7 @@ TypeBuilder::TypeBuilder(TypeBuilder&& other) = default;
 TypeBuilder& TypeBuilder::operator=(TypeBuilder&& other) = default;
 
 void TypeBuilder::grow(size_t n) {
-  assert(size() + n > size());
+  assert(size() + n >= size());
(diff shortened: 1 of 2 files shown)


tlively:
* main
  * **PR #4668** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4668" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4668?utm_source=stack-comment).

kripken (approved):

PR #4672, opened 2022-05-17
Fuzzer: Show a clear error if a given wasm fails to run

kripken:
Without this the error from a bad given wasm file - either by the user, or during
a reduction where smaller wasms are given - could be very confusing. A bad
given wasm file during reduction, in particular, indicates a bug in the reducer
most likely.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -1048,7 +1048,11 @@ def test_one(random_input, given_wasm):
         # apply properties like not having any NaNs, which the original fuzz
         # wasm had applied. that is, we need to preserve properties like not
         # having nans through reduction.
-        run([in_bin('wasm-opt'), given_wasm, '-o', 'a.wasm'] + FUZZ_OPTS + FEATURE_OPTS)
+        try:
+            run([in_bin('wasm-opt'), given_wasm, '-o', 'a.wasm'] + FUZZ_OPTS + FEATURE_OPTS)
+        except Exception as e:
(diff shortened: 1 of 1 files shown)


scripts/fuzz_opt.py
```diff
@@ -1048,7 +1048,11 @@ def test_one(random_input, given_wasm):
         # apply properties like not having any NaNs, which the original fuzz
         # wasm had applied. that is, we need to preserve properties like not
         # having nans through reduction.
-        run([in_bin('wasm-opt'), given_wasm, '-o', 'a.wasm'] + FUZZ_OPTS + FEATURE_OPTS)
+        try:
+            run([in_bin('wasm-opt'), given_wasm, '-o', 'a.wasm'] + FUZZ_OPTS + FEATURE_OPTS)
+        except Exception as e:
```
aheejin, line 1053: How is this case different from what this is testing? https://github.com/WebAssembly/binaryen/blob/bfa956c250e0aa9c9dcedc2287ab74e72a2deeb5/scripts/fuzz_opt.py#L1313

kripken: I think line 1313 is the generic error we show when we see any failure. Those failures should be errors that the fuzzer found. The extra error in this PR, that will be written before the generic error, is about the case where we error even before doing anything with the given wasm, so we've been given a bad wasm basically, which is likely an internal error.

aheejin (approved):

PR #4673, opened 2022-05-17
wasm-reduce: Fix order in shrinkByReduction call

kripken:
The old code would short-circuit and not do anything after we managed
any reduction in the loop here. That would end up doing entire iterations of
the whole pipeline before removing another element segment, which could
be slow.


aheejin (approved):
Maybe one-line comment on the order there is good?

kripken:
Good idea, added a comment now.

PR #4677, opened 2022-05-18
Validator: Check features for ref.null's type

kripken:
Yet another case of this. `eq` only requires reference types, so we can't
unconditionally pick `i31` or `data` as subtypes of it, we need GC to be
enabled.

### DIFF
--- a/src/tools/fuzzing/fuzzing.cpp
+++ b/src/tools/fuzzing/fuzzing.cpp
@@ -3045,12 +3045,9 @@ HeapType TranslateToFuzzReader::getSubType(HeapType type) {
                  HeapType::data));
       case HeapType::eq:
         // TODO: nontrivial types as well.
-        return pick(
-          FeatureOptions<HeapType>()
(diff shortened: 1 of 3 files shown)


tlively:
`eq` should require gc, not just reference types. So if we're trying to get a subtype of `eq`, we can probably assume that `i31` is valid as well.

kripken:
> `eq` should require gc, not just reference types.

Ah, I was confused... it is mentioned in the reference-types repo, and even [fully defined](https://github.com/WebAssembly/reference-types/blob/master/proposals/reference-types/Overview.md)... but in a "future features" section :rofl: whose title I did not scroll back to read

kripken:
PR updated to contain assertions instead. The second assertion still fails for me, so I'll need to investigate a proper fix for this.

kripken:
Ok, the issue was missing validation. Fixed + added a test, PTAL.

tlively (approved):
LGTM!

PR #4678, opened 2022-05-19
[NFC] Deduplicate imported global parsing

### LINKED ISSUE #4676: Cannot import globals with GC types in text format
```wat
(module
  (type $struct (struct funcref))
  (import "a" "b" (global $global-import (ref $struct)))
)
```
That fails to validate with
```wat
[parse exception: expected mut (at 3:41)]
```

tlively:
Share the logic for parsing imported and non-imported globals of the formats:

  (import "module" "base" (global $name? type))

  (global $name? type init)

This fixes #4676, since the deleted logic for parsing imported globals did not
handle parsing GC types correctly.


tlively:
FWIW I tried to deduplicate the function parsing logic as well, but gave up because it was too complicated :/

kripken (approved):

PR #4682, opened 2022-05-23
Add C and JS API functions for accessing memory info

eyebrowsoffire:
Bringing https://github.com/WebAssembly/binaryen/issues/3573 back to life. I think it has all the requested changes in the previous PR. I also modified the boolean return values to be actually boolean, but other than that this is the same, just updated and merged with the main branch.


src/js/binaryen.js-post.js
```diff
@@ -2500,9 +2500,26 @@ function wrapModule(module, self = {}) {
...
   };
+  self['hasMemory'] = function() {
+    return Boolean(Module['_BinaryenHasMemory'](module));
+  };
+  self['getMemoryInfo'] = function() {
+    var memoryInfo = {
+      'module': UTF8ToString(Module['_BinaryenMemoryImportGetModule'](module)),
+      'base': UTF8ToString(Module['_BinaryenMemoryImportGetBase'](module)),
+    };
+    if (memoryInfo['module'] === '' && memoryInfo['base'] === '') {
```
kripken, line 2511: Imported memories still declare initial and max sizes, so perhaps this `if` is not necessary?

eyebrowsoffire: Yep, you're right. I'll change that.

src/binaryen-c.cpp
```diff
@@ -3812,6 +3812,37 @@ uint32_t BinaryenGetMemorySegmentByteOffset(BinaryenModuleRef module,
...
+  } else {
+    return "";
+  }
+}
+const char* BinaryenMemoryImportGetBase(BinaryenModuleRef module) {
+  auto& memory = ((Module*)module)->memory;
+  if (memory.imported()) {
+    return memory.base.c_str();
+  } else {
+    return "";
```
tlively, line 3840: I think it would make more sense to return `NULL` here and above? IIUC, empty string is a perfectly valid string for import modules and names.

OTOH, I see that there is plenty of precedent for using empty string in this context, so we probably shouldn't change that in this PR.

tlively (approved):
Also cc @ashleynh, since this is related to what you're working on.

src/js/binaryen.js-post.js
kripken: I think a test may need updating after this change, see CI.

src/binaryen-c.cpp
eyebrowsoffire: Yeah I agree with you, but it looks like this is the convention used for other entities like tables, functions, etc. It's particularly annoying that from the JS perspective you get fields with empty strings for base/name when it would make way more sense for those fields to be undefined. That being said, unless you feel strongly I don't really want to change this expectation in this PR. Who knows what clients rely on the empty string behavior? Hyrum's law and all that.

PR #4685, opened 2022-05-24
PossibleContents + ContentOracle

kripken:
This pulls out the core PossibleContents and ContentOracle classes from the
very large #4598, making a smaller PR that can be reviewed first. This includes
unit tests for the code, but comprehensive testing will only appear in the later
PR, when a new pass is added that uses all this.

PossibleContents tracks the possible contents at particular locations in the
program. It can track constant values as well as "this must contain this
exact type", which is more than wasm itself can indicate.

*Location structs are provided to declare locations in the wasm, like the
location of a local or of a function argument.

ContentOracle analyzes the entire program, and can then map a Location
to the PossibleContents there, which a later pass will use to optimize.


tlively (commented):
I don't think copying all my previous comments across to this PR would be worth it, but they should all still apply. Let's continue existing discussions on the tests and header on the previous PR and I'll add any new comments to this PR.

kripken:
`example` test converted to `gtest`.

test/gtest/possible-contents.cpp
```diff
@@ -0,0 +1,296 @@
...
+// Asserts a == b, in any order.
+template<typename T> void assertEqualSymmetric(const T& a, const T& b) {
+  std::cout << "\nassertEqualSymmetric\n";
+  a.dump(std::cout);
+  std::cout << '\n';
+  b.dump(std::cout);
+  std::cout << '\n';
+
+  EXPECT_TRUE(a == b);
+  EXPECT_TRUE(b == a);
```
tlively, line 17:
```suggestion
  EXPECT_EQ(a == b);
  EXPECT_EQ(b == a);
```
This will make gtest print a more helpful error message if the test fails. It would be good to keep the test output clean by removing the unconditional printing above and relying on gtest to do any necessary printing. If you implement `operator<<(std::ostream&)` to call `dump`, then gtest will find and use it automatically.

test/gtest/possible-contents.cpp
```diff
@@ -0,0 +1,296 @@
...
+    Fatal() << "error in parsing wasm text";
+  }
+  return wasm;
+};
+
+// We want to declare a bunch of globals that are used in various tests. Doing
+// so in the global scope is risky, as their constructors use types, and the
+// type system itself uses global constructors. Instead, put all the tests
+// inside a "testMain()" function whose scope will contain those reused
+// "globals."
```
tlively, line 79: This sounds like a good job for a test fixture. Docs are here: https://google.github.io/googletest/primer.html#same-data-multiple-tests and you can copy the structure of our type test fixture here: https://github.com/WebAssembly/binaryen/blob/main/test/gtest/type-test.h

test/gtest/possible-contents.cpp
kripken: I am a little unhappy about using `EXPECT_EQ` for a test that focuses on the implementation of the `==` operator itself. That is, `EXPECT_EQ` might be implemented internally as `assert(!(x != y))` - yes, that seems a little silly, but it could actually make sense for `EXPECT_NE` to not use `!=`. But my point is that it's best to make sure we actually see the `==` that is tested here.

I agree that using `operator<<` is useful. It would depend on the above point, however (if we need to keep the existing unconditional printing, then we don't need it).

test/gtest/possible-contents.cpp
kripken: Thanks, refactored now to use a fixture.

test/gtest/possible-contents.cpp
tlively: Thankfully the assertions are documented as testing the operators specifically: https://google.github.io/googletest/reference/assertions.html#EXPECT_EQ, so I don't think we need to worry about this. (Without that documentation I would agree that using EXPECT_TRUE would be better.)

test/gtest/possible-contents.cpp
```diff
@@ -74,223 +74,215 @@ static std::unique_ptr<Module> parse(std::string module) {
...
+// type system itself uses global constructors. Instead, we use a fixture for
+// that.
+class PossibleContentsTest : public testing::Test {
+protected:
+  void SetUp() override {
+    // Use nominal typing to test struct types.
+    wasm::setTypeSystem(TypeSystem::Nominal);
+  }
+
+  PossibleContents none_ = PossibleContents::none();
```
tlively, line 86: Do we need the underscore here?
```suggestion
  PossibleContents none = PossibleContents::none();
```

test/gtest/possible-contents.cpp
```diff
@@ -0,0 +1,288 @@
...
+  assertCombination(anyGlobal, funcNull, many);
+}
+
+TEST_F(PossibleContentsTest, TestOracleMinimal) {
+  // A minimal test of the public API of PossibleTypesOracle. See the lit test
+  // for coverage of all the internals (using lit makes the result more
+  // fuzzable).
+  auto wasm = parse(R"(
+    (module
+      (type $struct (struct))
```
tlively, line 242: Does the struct serve any purpose here?

tlively (commented):
Tests lgtm modulo two small comments. Will look at the implementation next.

test/gtest/possible-contents.cpp
kripken: Ah, fair enough. To be honest even with such docs I am not happy depending on a third-party library that uses macros for something like this... but I suspect I'm overly paranoid here. I'll refactor in the more pragmatic way as you suggest.

kripken: Although, wait... how would we check all the combinations? Atm we have this:
```cpp
  EXPECT_TRUE(a == b);
  EXPECT_FALSE(a != b);
```
The first line can be `EXPECT_EQ(a, b)`. But how can we write the second line? We want to write "not not equal" in order to test `operator!=`...

tlively: Yep, we need to keep using `EXPECT_FALSE` for the second line.

kripken: In that case is it worth changing stuff here? We'll still need the full manual logging for that case. The code would be more consistent to not switch between the "modes".

tlively: It would be nice to get the better automatic error reporting where we can, even if we can't get it everywhere. And I hope we can avoid unconditionally printing anything.

But actually, looking at the gtest docs, it looks like what we want here is `EXPECT_PRED`, which lets us test an arbitrary condition and still get nice automatic printing: https://google.github.io/googletest/reference/assertions.html#EXPECT_PRED.

aheejin (commented):
Haven't read the impl yet, but some questions in the header and the test:

src/ir/possible-contents.h
```diff
@@ -0,0 +1,490 @@
...
+           tupleIndex == other.tupleIndex;
+  }
+};
+
+// The location of a branch target in a function, identified by its name.
+struct BranchLocation {
+  Function* func;
+  Name target;
+  // As in ExpressionLocation, the index inside the tuple, or 0 if not a tuple.
+  Index tupleIndex;
```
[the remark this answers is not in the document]
kripken: Yes, exactly. I'll improve the comment.

test/gtest/possible-contents.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+  PossibleContents i32Global1 =
+    PossibleContents::global("i32Global1", Type::i32);
+  PossibleContents i32Global2 =
+    PossibleContents::global("i32Global2", Type::i32);
+  PossibleContents f64Global = PossibleContents::global("f64Global", Type::f64);
+  PossibleContents anyGlobal =
+    PossibleContents::global("anyGlobal", Type::anyref);
+
+  PossibleContents func = PossibleContents::literal(
```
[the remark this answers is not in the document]
kripken: Good point, thanks, fixed.

test/gtest/possible-contents.cpp
```diff
@@ -0,0 +1,288 @@
...
+          )
+          (i32.const 0)
+        )
+      )
+    )
+  )");
+  ContentOracle oracle(*wasm);
+  // The function's body should be Many.
+  EXPECT_TRUE(
+    oracle.getContents(ResultLocation{wasm->getFunction("foo"), 0}).isMany());
```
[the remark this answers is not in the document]
kripken: `ExactType` means it has that type and not a subtype. But the actual contents will be one of the 4 possible subtypes here.

In the future we should add a "ConeType" which is I think what you expect here. `ConeType(Type::anyref)` would say that it is anyref or any subtype.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1610 @@
...
+
+} // namespace std
+
+namespace wasm {
+
+namespace {
+
+// The data we gather from each function, as we process them in parallel. Later
+// this will be merged into a single big graph.
+struct CollectedInfo {
```
aheejin, line 193: If this only pertains to a single function, maybe name this `CollectedFuncInfo` or something?

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1610 @@
...
+  }
+
+  bool isRelevant(Signature sig) {
+    return isRelevant(sig.params) || isRelevant(sig.results);
+  }
+
+  bool isRelevant(Expression* curr) { return curr && isRelevant(curr->type); }
+
+  template<typename T> bool isRelevant(const T& vec) {
+    return true;
```
aheejin, line 264:
```suggestion
```
Is this a typo?

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1610 @@
...
+               PossibleContents contents = PossibleContents::many()) {
+    if (isRelevant(curr)) {
+      addRoot(ExpressionLocation{curr, 0}, contents);
+    }
+  }
+
+  // As above, but given an arbitrary location and not just an expression.
+  void addRoot(Location loc,
+               PossibleContents contents = PossibleContents::many()) {
+    info.roots.emplace_back(loc, contents);
```
aheejin, line 831: Can this`roots` be duplicated?

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1610 @@
...
+
+  // First, collect information from each function.
+  ModuleUtils::ParallelFunctionAnalysis<CollectedInfo> analysis(
+    wasm, [&](Function* func, CollectedInfo& info) {
+      InfoCollector finder(info);
+
+      if (func->imported()) {
+        // Imports return unknown values.
+        for (Index i = 0; i < func->getResults().size(); i++) {
+          finder.addRoot(ResultLocation{func, i}, PossibleContents::many());
```
aheejin, line 1039: If an imported function returns a value of a type that doesn't have subtypes, such as plain old types like `i32`, can't we create `ExactType` for this?

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1610 @@
...
+  void visitSIMDLoadStoreLane(SIMDLoadStoreLane* curr) { addRoot(curr); }
+  void visitMemoryInit(MemoryInit* curr) {}
+  void visitDataDrop(DataDrop* curr) {}
+  void visitMemoryCopy(MemoryCopy* curr) {}
+  void visitMemoryFill(MemoryFill* curr) {}
+  void visitConst(Const* curr) {
+    addRoot(curr, PossibleContents::literal(curr->value));
+  }
+  void visitUnary(Unary* curr) { addRoot(curr); }
+  void visitBinary(Binary* curr) { addRoot(curr); }
```
aheejin, line 330: Can't we infer `ExactType` if the type of these don't have subtypes, like `i32`? The same for most other `addRoot`s as well. Maybe I'm mistaken about some assumptions.

aheejin (commented):
Haven't finished reading, but it would help my understanding to ask these first, because I might be fundamentally mistaken about something.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1610 @@
...
+  //    (local.get $ref)
+  //    (local.get $value)
+  //  )
+  //
+  // Imagine that the first local.get, for $ref, receives a new value. That can
+  // affect where the struct.set sends values: if previously that local.get had
+  // no possible contents, and now it does, then we have DataLocations to
+  // update. Likewise, when the second local.get is updated we must do the same,
+  // but again which DataLocations we update depends on the ref passed to the
+  // struct.get. To handle such things, we set add a childParent link, and then
```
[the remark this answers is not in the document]
kripken: Yes, good catch!

src/ir/possible-contents.cpp
kripken: Sounds good, renamed.

src/ir/possible-contents.cpp
kripken: Good catch! Yes, looks like a typo... so the code was considering too many things relevant, and doing wasted work...

src/ir/possible-contents.cpp
kripken: Do you mean can it contain duplicate values? There wouldn't be any harm if there were, but I think there can't be, since we walk the code and add roots as we find things, and we don't walk things twice.

src/ir/possible-contents.cpp
kripken: In principle we could, yes. There is a comment in `possible-contents.h` about it, in `combine()` I believe. While we could create `ExactType{i32}`, that doesn't add any information since without subtyping we always know the exact type. And there would be a downside to doing so: "Many" is a clear signal that we can't do any better, that we've hit the worst case. "Many" indicates that for all types. If instead we emitted `ExactType{i32}` we'd need to check more carefully if it is the worst case, by seeing it is an ExactType of a type with no subtypes - not complicated, but it's simpler the other way I think.

I can add a comment here if the other isn't enough?

src/ir/possible-contents.cpp
kripken: Hmm, it probably is worth an extra comment here - since this "use Many" principle is used in all these `addRoot` places. I'll add one here, which will hopefully address the previous comment too.

kripken: I ended up writing a longer comment at the top of the header, and linking to it. The old comment was actually refactored together with some code into the cpp, so there was no longer a good comment in the header, which is the right place for it... fixed.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1607 @@
...
+    }
+#endif
+
+    auto iter = workQueue.begin();
+    auto locationIndex = *iter;
+    workQueue.erase(iter);
+
+    flowAfterUpdate(locationIndex);
+
+    updateNewLinks();
```
aheejin, line 1218: We do all processing from `locationIndex` -> targets in `flowUpdate()` above. Then does adding new links after we've done all the processing anything? (I guess I am misunderstand something.)

src/ir/possible-contents.h
```diff
@@ -0,0 +1,513 @@
...
+//
+// Note that we prefer not to emit ExactType{i32} etc., or any other type that
+// has no subtyping: instead, we emit Many for such things. While ExactType{i32}
+// would be correct, it adds no additional information: we know the type from
+// the wasm, and we know no subtype is possible, so we know it exactly.
+// Furthermore, using Many instead simplifies the logic in various places, in
+// particular, Many is a clear signal that we've hit the worst case and can't
+// do any better in that location, and can stop looking at it. If instead of
+// Many we emitted ExactType{i32} we'd need to also check for an exact type of
+// a type without subtypes, when trying to reason about hitting the worst case.
```
aheejin, line 69: Sorry I still don't really understand this paragraph.

> While ExactType{i32} would be correct, it adds no additional information: we know the type from the wasm, and we know no subtype is possible, so we know it exactly.

Yeah we know the type exactly, but I've never seen we query it in `Flower`. If a child gets `Many` type, it just stops flowing the value to its targets. Then isn't this loss of information?

> Furthermore, using Many instead simplifies the logic in various places, in particular, Many is a clear signal that we've hit the worst case and can't do any better in that location, and can stop looking at it. If instead of Many we emitted ExactType{i32} we'd need to also check for an exact type of a type without subtypes, when trying to reason about hitting the worst case.

Don't really get this part either. Why `ExactType{i32}` means we can't do any better? If we assign `ExactType` to a location, it will be propagated to its target locations, but we instead say it's `Many` and don't propagate anything anymore.

Maybe my misunderstanding stems from that I don't really know how this class is used for optimization. (I haven't looked at the GUFA PR yet) But I assumed the purpose of this analysis is to have more specific info on what can go to each location; correct me if I'm wrong.

src/ir/possible-contents.cpp
aheejin: I saw your new comments. Let's talk more there.

aheejin (commented):
Sorry, too many questions 😅
I think many of them stem from that I don't really know who this data will be used for optimization.

src/ir/possible-contents.h
```diff
@@ -0,0 +1,513 @@
...
+//  * Global:          The name of a global whose value is here. We do not know
+//                     the actual value at compile time, but we know it is equal
+//                     to that global. Typically we can only infer this for
+//                     immutable globals.
+//
+//  * ExactType:       Any possible value of a specific exact type - *not*
+//                     including subtypes. For example, (struct.new $Foo) has
+//                     ExactType contents of type $Foo.
+//                     If the type here is nullable then null is also allowed.
+//                     TODO: Add ConeType, which would include subtypes.
```
aheejin, line 51: Is cone type a type system terminology? I have never heard of it, so..

src/ir/possible-contents.h
aheejin: Is `ExactType` is supposed to only reference types? If so, would renaming this to `ExactRefType` be clearer?

src/ir/possible-contents.h
```diff
@@ -0,0 +1,513 @@
...
+  // As in ExpressionLocation, the index inside the tuple, or 0 if not a tuple.
+  Index tupleIndex;
+  bool operator==(const LocalLocation& other) const {
+    return func == other.func && index == other.index &&
+           tupleIndex == other.tupleIndex;
+  }
+};
+
+// The location of a branch target in a function, identified by its name.
+struct BranchLocation {
```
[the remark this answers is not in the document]
kripken: Good point, thanks. Renamed to `BreakTargetLocation`.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1607 @@
...
+
+    // See visitPop().
+    assert(handledPops == totalPops);
+  }
+
+  // Helpers
+
+  // Handles the value sent in a break instruction. Does not handle anything
+  // else like the condition etc.
+  void handleBreakValue(Expression* curr) {
```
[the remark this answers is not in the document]
kripken: You're right, it should be here too. Fixing.

Note that `isRelevant` is mostly an optimization - just to avoid passing around values that we don't want to optimize. So the fuzzer won't catch missing places...

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1607 @@
...
+      [&](Index i) {
+        return LocalLocation{target, i, 0};
+      },
+      [&](Index i) {
+        return ResultLocation{target, i};
+      });
+  }
+  void visitCallIndirect(CallIndirect* curr) {
+    // TODO: the table identity could also be used here
+    auto target = curr->heapType;
```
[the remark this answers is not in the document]
kripken: Sounds good, will make it the same as `visitCallRef`.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1607 @@
...
+      });
+    }
+    addRoot(curr, PossibleContents::exactType(curr->type));
+  }
+  void visitArrayNew(ArrayNew* curr) {
+    if (curr->type == Type::unreachable) {
+      return;
+    }
+    auto type = curr->type.getHeapType();
+    if (curr->init) {
```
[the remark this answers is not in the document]
kripken: I think we don't need to check for a tuple here, since a tuple would not validate as correct wasm anyhow.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1607 @@
...
+    }
+    addRoot(curr, PossibleContents::exactType(curr->type));
+  }
+
+  // Struct operations access the struct fields' locations.
+  void visitStructGet(StructGet* curr) {
+    if (!isRelevant(curr->ref)) {
+      // We are not tracking references, and so we cannot properly analyze
+      // values read from them, and must assume the worst.
+      addRoot(curr);
```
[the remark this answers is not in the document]
kripken: "Tracking" means that we notice those values as relevant, and we represent them and flow them around. I'll replace "tracking" with "is relevant" which would make the connection to that function better.

Yes, we only consider references relevant in nominal mode. In other modes we don't have enough information to do much useful anyhow (since we don't have a way to see subtypes).

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1607 @@
...
+  void visitArraySet(ArraySet* curr) {
+    if (curr->ref->type == Type::unreachable) {
+      return;
+    }
+    addSpecialChildParentLink(curr->ref, curr);
+    addSpecialChildParentLink(curr->value, curr);
+  }
+
+  void visitArrayLen(ArrayLen* curr) {
+    // TODO: optimize when possible
```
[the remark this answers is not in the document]
kripken: Yes, this returns `Many` atm (the default value for `addRoot`). The TODO is to potentially optimize cases where we can, like if we see the length of the array, we could return a Literal. (I think that might help some Java itable patterns, which I guess is why I wrote the TODO here.)

src/ir/possible-contents.h
kripken: I don't remember where I heard it... maybe from Dart?

Does that line explain what it is well enough, or do you think it should be improved?

src/ir/possible-contents.h
kripken: On the one hand, when we have a wasm location, like an `i32.add`, and we have `PossibleContents` for it, then `ExactType{i32}` adds nothing since the `i32.add` already has that type. That's why we don't use it in most places. However, on the other hand, there may be cases where we only have a `PossibleContents` but *not* have a wasm location for it. That's the case in our unit tests for `PossibleContents`, and it might make sense elsewhere. In such contexts I think `ExactType{i32}` is useful, as we do want a way to express that.

So overall, I think when optimizing we don't want `ExactType{i32}`, but in general we don't want to disallow it entirely.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1607 @@
...
+      auto params = getModule()->getTag(tag)->sig.params;
+      if (params.size() == 0) {
+        return;
+      }
+
+      // Find the pop of the tag's contents. The body must start with such a
+      // pop, which might be of a tuple.
+      FindAll<Pop> pops(body);
+      assert(!pops.list.empty());
+      auto* pop = pops.list[0];
```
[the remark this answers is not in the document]
kripken: Good catch!

I did what you suggested in a new PR #4704 (and also applied it in this PR, but that one should land first).

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1607 @@
...
+    for (Index i = 0; i < curr->operands.size(); i++) {
+      info.links.push_back(
+        {ExpressionLocation{operands[i], 0}, TagLocation{tag, i}});
+    }
+  }
+  void visitRethrow(Rethrow* curr) {}
+
+  void visitTupleMake(TupleMake* curr) {
+    if (isRelevant(curr->type)) {
+      for (Index i = 0; i < curr->operands.size(); i++) {
```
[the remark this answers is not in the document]
kripken: We don't validate nested tuples, correct, so this is the same as the above.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1607 @@
...
+    for (auto& [root, value] : info.roots) {
+      roots[root] = value;
+
+      // Ensure an index even for a root with no links to it - everything needs
+      // an index.
+      getIndex(root);
+    }
+    for (auto [child, parent] : info.childParents) {
+      childParents[getIndex(ExpressionLocation{child, 0})] =
+        getIndex(ExpressionLocation{parent, 0});
```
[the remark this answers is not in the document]
kripken: Good point, that's not clear here... It is the case, but we can't really see that at this point in the code. I'll add an assertion.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1607 @@
...
+  for (auto& func : wasm.functions) {
+    for (Index i = 0; i < func->getParams().size(); i++) {
+      links.insert(getIndexes({SignatureParamLocation{func->type, i},
+                               LocalLocation{func.get(), i, 0}}));
+    }
+    for (Index i = 0; i < func->getResults().size(); i++) {
+      links.insert(getIndexes({ResultLocation{func.get(), i},
+                               SignatureResultLocation{func->type, i}}));
+    }
+  }
```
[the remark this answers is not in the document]
kripken: Sorry, I'm not sure what you mean here?

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1607 @@
...
+
+    updateNewLinks();
+  }
+
+  // TODO: Add analysis and retrieval logic for fields of immutable globals,
+  //       including multiple levels of depth (necessary for itables in j2wasm).
+}
+
+bool Flower::updateContents(LocationIndex locationIndex,
+                            PossibleContents newContents) {
```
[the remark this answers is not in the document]
kripken: There is one place where we can actually remove targets and get a benefit from it - where we send stuff to the targets in the main loop. All the other places are different in some way, like when we process roots we will never visit them again, so there is nothing to remove. And when we compute the dynamic targets for StructGet etc., because the list is dynamic there is no benefit to pruning it (we throw that dynamic list away immediately).

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1607 @@
...
+    // or anything else.
+    // TODO: The Global case may have a different cone type than the heapType,
+    //       which we could use here.
+    // TODO: A Global may refer to an immutable global, which we can read the
+    //       field from potentially (reading it from the struct.new/array.now
+    //       in the definition of it, if it is not imported; or, we could track
+    //       the contents of immutable fields of allocated objects, and not just
+    //       represent them as ExactType).
+    //       See the test TODO with text "We optimize some of this, but stop at
+    //       reading from the immutable global"
```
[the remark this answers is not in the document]
kripken: Ah, yes, that test is one of those that I split out into the later PR, sorry about that.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1607 @@
...
+
+void Flower::connectDuringFlow(Location from, Location to) {
+  // Add the new link to a temporary structure on the side to avoid any
+  // aliasing as we work. updateNewLinks() will be called at the proper time
+  // to apply these links to the graph safely.
+  auto newLink = LocationLink{from, to};
+  auto newIndexLink = getIndexes(newLink);
+  if (links.count(newIndexLink) == 0) {
+    newLinks.push_back(newIndexLink);
+    links.insert(newIndexLink);
```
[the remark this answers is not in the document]
kripken: Good point, this was not clear at all. The relevant aliasing is on the `targets` vectors: the flow iterates on such a vector as it sends contents, and we can't modify it at that time. I'm fixing all the relevant comments to mention that.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1607 @@
...
+  PossibleContents filtered;
+  if (contents.isMany()) {
+    // Just pass the Many through.
+    // TODO: we could emit a cone type here when we get one, instead of
+    //       emitting a Many in any of these code paths
+    filtered = contents;
+  } else {
+    auto intendedType = cast->getIntendedType();
+    bool mayBeSubType =
+      HeapType::isSubType(contents.getType().getHeapType(), intendedType);
```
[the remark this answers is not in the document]
kripken: "maybe" here is because it might also be null, and it might also be simply incompatible. For example we might be trying to cast a function ref to a struct ref.

src/ir/possible-contents.cpp
kripken: I'm updating the comment on that method to this, hopefully it is clearer?
```cpp
  // New links must be added to the appropriate |targets| vector (that is, the
  // |targets| of the source must get a new target added to it). We are
  // iterating on |targets| for most of our work, however, so we must not alter
  // it at that time. This method is called later, after there is no risk of
  // aliasing, and it appends to the appropriate |targets| vector.
  void updateNewLinks();
```

src/ir/possible-contents.h
kripken: > Maybe my misunderstanding stems from that I don't really know how this class is used for optimization. (I haven't looked at the GUFA PR yet)

Yes, maybe it was not best to split up the PRs... this PR is much smaller but then the motivation is not present.

But hopefully my last comment in this thread is helpful? The key thing is that when we optimize we also have the context of the location that the content is for. And that location has a type. So if subtyping is not an issue, Many tells us no more than an exact type.

kripken:
All comments so far addressed/replied. Thanks for the feedback!

src/ir/possible-contents.cpp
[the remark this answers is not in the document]
aheejin: Hmm, sorry, I somehow thought this returns `ExactType{i32}` when writing this comment, so I asked like "didn't you say `ExactType{i32}` is useless?". Not sure why I thought that way. Nevermind.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1607 @@
...
+    }
+  }
+  void visitThrow(Throw* curr) {
+    auto& operands = curr->operands;
+    if (!isRelevant(operands)) {
+      return;
+    }
+
+    auto tag = curr->tag;
+    for (Index i = 0; i < curr->operands.size(); i++) {
```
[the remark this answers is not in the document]
aheejin: Sounds good as is. Didn't know it doesn't validate.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1607 @@
...
+    }
+  };
+
+  for (auto& ex : wasm.exports) {
+    if (ex->kind == ExternalKind::Function) {
+      calledFromOutside(ex->value);
+    } else if (ex->kind == ExternalKind::Table) {
+      // If a table is exported, assume any function in tables can be called
+      // from the outside.
+      // TODO: This could be more precise about which tables
```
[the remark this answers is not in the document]
aheejin: But aren't we iterating over exports in this loop anyway? Then aren't all tables that end up here exported?

src/ir/possible-contents.cpp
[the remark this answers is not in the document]
aheejin: Nevermind, I was confused 😅

aheejin:
> All comments so far addressed or replied. Thanks for the feedback!

Thanks for the answers! Btw I think there are two more comments from @tlively at the top.

src/ir/possible-contents.cpp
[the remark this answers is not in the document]
aheejin: Where are we iterating on the `target` vectors? It looks we are iterating on `targets` here:
https://github.com/WebAssembly/binaryen/pull/4685/files#diff-4d9d6d08d863caa22d1fa86a99f2b7d37d1013d36c98e177aca6edd6eaff54e1R1305-R1311

And `connectDuringFlow` is called from `readFromData` here: https://github.com/WebAssembly/binaryen/pull/4685/files#diff-4d9d6d08d863caa22d1fa86a99f2b7d37d1013d36c98e177aca6edd6eaff54e1R1354

Both are in `flowAfterUpdate`, but it looks the iteration precedes `connectDuringFlow` calls. No?

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1607 @@
...
+  } else {
+    assert(refContents.isMany() || refContents.isGlobal());
+
+    // Update all possible subtypes here.
+    auto type = ref->type.getHeapType();
+    for (auto subType : subTypes->getAllSubTypes(type)) {
+      auto heapLoc = DataLocation{subType, fieldIndex};
+      updateContents(heapLoc, valueContents);
+    }
+  }
```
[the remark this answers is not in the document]
aheejin: I am still confused.

1. I guess you mean `DataLocation` by `StructLocation`, right? (There's no `StructLocation`)

2. Why do we use `SpecialLocation`? I think it makes sense if `SpecialLocation`s are uniqued, but it looks they are created every time we process `struct.get` / `array.get` (per subtype): https://github.com/WebAssembly/binaryen/pull/4685/files#diff-4d9d6d08d863caa22d1fa86a99f2b7d37d1013d36c98e177aca6edd6eaff54e1R1489
Then what makes this more efficient? For example, a `struct.get`'s field 0 can have 3 subtypes. Let's call them A, B, and C. And let's call the read expression's location E. And we created a special location S. If we do the optimization, we will create connections of A->S, B->S, and C->S, and then S->E. If we don't do the optimization, we will create A->E, B->E, and C->E. And for another `struct.get`, we will create another special location S2, and then this will repeat. What's the benefit? And we will even do this multiple times for the same `struct.get`, every time this `struct.get` is pushed to the `workQueue`.

3. About why we don't use `connectDuringFlow`, `connectDuringFlow` adds new links that will update targets later. It's OK not to do that in `writeToData`?

4. Why do we do this (creating `SpecialLocation` and stuff) later? Can't we do this when we first run `InfoCollector`? It looks simpler, and in the current way it looks we end up doing the same structure building (creating special locations and links and such) for expressions like `struct.get`/`struct.set` every time these exprs are updated.

src/ir/possible-contents.cpp
aheejin: Yeah the comment looks clearer, but I have another question about this behavior. I asked in in another thread.

src/ir/possible-contents.h
aheejin: > On the one hand, when we have a wasm location, like an `i32.add`, and we have `PossibleContents` for it, then `ExactType{i32}` adds nothing since the `i32.add` already has that type. That's why we don't use it in most places.

This looks the repetition of the comment above, so I have the same question from my previous comment:
Yeah we know the type exactly, but I've never seen we query it in `Flower`. If a child gets `Many` type, it just stops flowing the value to its targets. Then isn't this loss of information?

> However, on the other hand, there may be cases where we only have a `PossibleContents` but _not_ have a wasm location for it. That's the case in our unit tests for `PossibleContents`, and it might make sense elsewhere.

Can you point what test is this?

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1628 @@
...
+      assert(child->type.size() == parent->type.size());
+      for (Index i = 0; i < child->type.size(); i++) {
+        info.links.push_back(
+          {ExpressionLocation{child, i}, ExpressionLocation{parent, i}});
+      }
+    }
+  }
+
+  // See the comment on CollectedFuncInfo::childParents.
+  void addSpecialChildParentLink(Expression* child, Expression* parent) {
```
tlively, line 817: Is there a more descriptive word than "special" that we can use in this name? Alternatively, would dropping it and having `addChildParentLink` work? Or even `addParentLink`?

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1628 @@
...
+  // As above, but given an arbitrary location and not just an expression.
+  void addRoot(Location loc,
+               PossibleContents contents = PossibleContents::many()) {
+    info.roots.emplace_back(loc, contents);
+  }
+};
+
+// Main logic for building data for the flow analysis and then performing that
+// analysis.
+struct Flower {
```
tlively, line 842: 🌹🌷💮

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1628 @@
...
+  // to have already written the new contents to the main data structure. That
+  // avoids larger data here, and also, updating the contents as early as
+  // possible is helpful as anything reading them meanwhile (before we get to
+  // their work item in the queue) will see the newer value, possibly avoiding
+  // flowing an old value that would later be overwritten.
+#ifdef POSSIBLE_CONTENTS_INSERT_ORDERED
+  InsertOrderedSet<LocationIndex> workQueue;
+#else
+  std::unordered_set<LocationIndex> workQueue;
+#endif
```
tlively, line 936: Could we move the `ifdef` out of the main code by wrapping two alternative `using WorkQueue = ...` statements instead?

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1628 @@
...
+  // for each struct.get of a cone. If there are M such gets then we have N * M
+  // edges for this. Instead, make a single canonical "cone read" location, and
+  // add a single link to it from each get, which is only N + M (plus the cost
+  // of adding "latency" in requiring an additional step along the way for the
+  // data to flow along).
+  std::unordered_map<std::pair<HeapType, Index>, LocationIndex>
+    canonicalConeReads;
+
+  // Creates a new special location that is different from all others so far.
+  LocationIndex makeSpecialLocation() {
```
tlively, line 951: Maybe `makeFreshLocation` to be more descriptive?

tlively (commented):
Here are my comments after reading the first thousand lines or so of possible-contents.cpp.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1628 @@
...
+      auto params = getModule()->getTag(tag)->sig.params;
+      if (params.size() == 0) {
+        return;
+      }
+
+      // Find the pop of the tag's contents. The body must start with such a
+      // pop, which might be of a tuple.
+      auto pops = EHUtils::findPops(body);
+      assert(pops.size() == 1);
+      auto* pop = pops[0];
```
aheejin, line 691:
```suggestion
      auto* pop = EHUtils::findPop(body);
      if (!pop) {
        continue;
      }
```

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1628 @@
...
+    // notified us that we do not need to send them any more updates.
+    assert(targets.empty());
+  }
+
+  // We are mostly done, except for handling interesting/special cases in the
+  // flow, additional operations that we need to do aside from sending the new
+  // contents to the statically linked targets.
+
+  if (auto* exprLoc = std::get_if<ExpressionLocation>(&location)) {
+    auto* child = exprLoc->expr;
```
aheejin, line 1335: Nit: This can be moved down after the `if` below

src/ir/possible-contents.cpp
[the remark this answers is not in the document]
kripken: Oh, I see now. Yes, it's not clear from the comment, and the code is non-obvious. Yes, we loop over exports, but once we find one table that is exported, we "give up" and mark all element segment's contents as callable from outside. So if one table is exported, it "ruins things" for all others. This is good enough for now and simpler. I'll clarify the comment.

test/gtest/possible-contents.cpp
kripken: Ah, good point, this is not needed since the `none` type is namespaced in `wasm::Type`, so no conflict... fixing.

src/ir/possible-contents.cpp
kripken: Oh, sorry, I did do this change to use `findPop` and not `Pops` but looks like I didn't push it yesterday. But it was pushed today among the other updates.

Good point about the case of null, I didn't handle that, but I'm adding a commit now. I think we can assert on `pop` existing since we exited earlier if the tag had no params, is that correct?

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1628 @@
...
+                            LocalLocation{getFunction(), curr->index, i}});
+
+      // Tees also flow out the value (receiveChildValue will see if this is a
+      // tee based on the type, automatically).
+      receiveChildValue(curr->value, curr);
+    }
+  }
+
+  // Globals read and write from their location.
+  void visitGlobalGet(GlobalGet* curr) {
```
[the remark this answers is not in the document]
kripken: Oh, do we allow tuples in globals? I didn't know that...

I'll add a TODO here. Fixing it will require adding a tupleIndex to GlobalLocation, I think, and maybe more. Probably not urgent as I don't think any code exists that uses tuple globals though.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1628 @@
...
+    //       length)
+    addRoot(curr);
+  }
+  void visitArrayCopy(ArrayCopy* curr) {
+    if (curr->type == Type::unreachable) {
+      return;
+    }
+    // Our flow handling of GC data is not simple: we have special code for each
+    // read and write instruction. Therefore, to avoid adding special code for
+    // ArrayCopy, model it as a combination of an ArrayRead and ArrayWrite, by
```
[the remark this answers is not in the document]
kripken: No, at least not except for manual debugging (but I've managed without it even there).

src/ir/possible-contents.cpp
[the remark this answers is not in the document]
aheejin: 2. Ah, right, I missed that we query `canonicalConeReads` map before creating a new `SpecialLocation`. Thanks.

3. `readFromData` uses `connectDuringFlow` but `writeToData` does not.

---
Also I still don't understand your earlier explanation on why we use `SpecialLocation` only in `readFromData` and not in `writeToData`. As you pointed out, if # of subtypes is `n` and `readFromData` is called from `m` `struct.get`s, if we use `SpecialLocation`s, the # of edges is `O(n+m)`, but if we use `SpecialLocation`s, the # of edges is `O(n*m)`. (Assume we only think about a single field index, or an `array.get`.)

But doesn't this apply in the same way to `writeToData`? If # of subtypes is `n` and `writeToData` is called from `m` `struct.set`s, in the current impl, the # of edges is `O(n*m)`, and it can be reduced to `O(n+m)` if we use `SpecialLocation`s there too.

src/ir/possible-contents.h
aheejin: > For example, see `declaredHeapType` and how we use it on cone reads: if we have `Many` there then we just fall back to the declared type in the wasm. (I think that answers your point `I've never seen we query it in Flower` - if not, perhaps I've misunderstood you?)

Ah I see, we do use it in `readFromData`. What I meant was I have not seen it being used in all the other places. For example, binary or unary operations. You mean the user of this `PossibleContents` class, which is presumably in the original GUFA PR uses it?

src/ir/possible-contents.cpp
aheejin: Ah right, we have an early exit in case of no params, in which case we don't need this `if`... By the way is it ok to return? Shouldn't we continue?

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1618 @@
...
+    }
+
+    auto numTags = curr->catchTags.size();
+    for (Index tagIndex = 0; tagIndex < numTags; tagIndex++) {
+      auto tag = curr->catchTags[tagIndex];
+      auto* body = curr->catchBodies[tagIndex];
+
+      auto params = getModule()->getTag(tag)->sig.params;
+      if (params.size() == 0) {
+        return;
```
aheejin, line 687:
```suggestion
        continue;
```
Shouldn't we check other tags, even if the current tag doesn't have any params?

src/ir/possible-contents.cpp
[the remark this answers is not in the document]
kripken: Maybe a way to get an intuition for why read is not symmetric with write here is this: When we do a write, we are sending data forward. A read not only reads stuff right now, but also must receive data in the future. So it needs to modify the graph, which a write doesn't.

In more detail, in a write we know exactly where to send it to: if it's a precise type in the ref of the write, then exactly one `DataLocation` (for that exact type, and its field). If it's not precise, then all N subtypes. We **must** do that work, since our data representation looks like `DataLocation{type, fieldIndex}`. So we have no choice but to do N operations here (at least with this representation). But, this is not so bad actually :smile: It is N operations, but at least **the graph size** is not N, because we can just do a simple loop on the subtypes, and we're done. We don't need to add anything to the graph.

Reads are much harder, because we are not just passing data forward. We could do a loop over subtypes here (if we need subtypes), and read from those `DataLocations` right now, but that is not enough: we also need future updates to those `DataLocations` to reach us. For that, we need to add connections to the graph. This is where things can start to take lots of memory: we want to avoid adding N connections for every `struct.get`, and we use canonical cone reads to avoid that: one canonical cone read has N connections while the M `struct.get`s have just one (to the canonical one).

src/ir/possible-contents.h
kripken: Yes, fair point, it's just used there.

Ok, I did some experimentation and it seems like it's practical to do it as you suggest. I pushed that now. This is about the same amount of code as before, and I benchmarked it as being the same speed. And it is a simpler mental model, to not special-case exact types of non-references. So I agree your approach is best here.

This is done in https://github.com/WebAssembly/binaryen/pull/4685/commits/6c5affd482c96e7a1a076fb84fee86391c835b3f and then optimized in https://github.com/WebAssembly/binaryen/pull/4685/commits/ec3de0213b0bed1177add2a315c834476b7737cb - without that, it would be slightly slower, I measured - and comments are simplified in https://github.com/WebAssembly/binaryen/pull/4685/commits/2c3fb7827d1305022f25a2f2add7f048450e0b02

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1628 @@
...
+        // everything is modeled as if at index 0.
+        return DataLocation{type, 0};
+      });
+    }
+    addRoot(curr, PossibleContents::exactType(curr->type));
+  }
+
+  // Struct operations access the struct fields' locations.
+  void visitStructGet(StructGet* curr) {
+    if (!isRelevant(curr->ref)) {
```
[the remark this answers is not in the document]
tlively: Nominal or isorecursive! It would be good to start allowing isorecursive types in these optimizations because we should switch to them by default at some point.

src/ir/possible-contents.cpp
kripken: Good catch, thanks!

src/ir/possible-contents.cpp
tlively: Can we have something like this up at the top where we currently have the conditional include?

```
#ifdef POSSIBLE_CONTENTS_INSERT_ORDERED
#include ...
using WorkQueue = ...
#else
using WorkQueeu = ...
#endif
```

It seems fine to define `WorkQueue` up at the top outside of any namespaces since it can get included anywhere outside of this file.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1628 @@
...
+        info.links.push_back(
+          {getNullLocation(fields[i].type), DataLocation{type, i}});
+      }
+    } else {
+      // Link the operands to the struct's fields.
+      handleChildList(curr->operands, [&](Index i) {
+        return DataLocation{type, i};
+      });
+    }
+    addRoot(curr, PossibleContents::exactType(curr->type));
```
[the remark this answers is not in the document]
kripken: Good point, if we wanted to track fields in allocated structs then we'd need to do so dynamically, and include new information that flows around later. Not simple to do, probably... might require expanding the graph. And yeah, depending on how we do that, maybe this would not be a root, somehow, I'm not sure.

src/ir/possible-contents.cpp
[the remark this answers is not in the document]
kripken: Can we find all the subtypes of an isorecursive type? Doesn't that depend on other isorecursive recursion groups etc.? I guess new groups can't be added if we assume a closed world, is that the idea?

Regardless, this depends on `SubTyping` which atm has an assertion that it only works in nominal typing, so I guess that should be changed as well? Not sure how we'd want to handle the closed world concept there, though, if that's relevant - it's not a pass, it's a utility (for a pass, we can imagine a simple mechanism some day to not run closed-world passes if a flag says the world is open).

src/ir/possible-contents.cpp
kripken: We can, but the annoyance is that those `using WorkQueue = ..` must refer to `LocationIndex`. That is defined lower down in a double namespace. So what you are suggesting would look like this:
```cpp
namespace wasm {
namespace {

using LocationIndex = ..;

} // anonymous namespace
} // namespace wasm

#ifdef POSSIBLE_CONTENTS_INSERT_ORDERED
#include ...
using WorkQueue = InsertOrderedQueue<LocationIndex>;
#else
using WorkQueue = std::unordered_set<LocationIndex>;
#endif
```

Then we have an include after non-include code... so this just feels awkward to me, and not that helpful given this type is used in exactly one spot... But if you feel strongly I don't object.

> It seems fine to define WorkQueue up at the top outside of any namespaces since it can get included anywhere outside of this file.

Should that be "can't"? If not, why?

src/ir/possible-contents.cpp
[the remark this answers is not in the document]
tlively: I think it's literally just a matter of fixing that assertion. Since the isorecursive types have declared supertypes, all the supertype-related APIs work identically between nominal and isorecursive types.

src/ir/possible-contents.cpp
tlively: Oh right, I meant "can't." I see what you mean about the annoyance now. Leaving this as-is sgtm.

src/ir/possible-contents.cpp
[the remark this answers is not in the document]
kripken: Ok, sounds good, done.

I guess when we get to open-world situations we'll have a bunch more work on this stuff.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1636 @@
...
+
+  auto type = getType();
+  auto otherType = other.getType();
+
+  if (!type.isRef() || !otherType.isRef()) {
+    // At least one is not a reference. The only possibility left for a useful
+    // combination here is if they have the same type (since we've already ruled
+    // out the case of them being equal). If they have the same type then
+    // neither is a reference and we can emit an exact type (since subtyping is
+    // not relevant for non-references.
```
aheejin, line 73: > since subtyping is not relevant for non-references.

What does this mean?

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1636 @@
...
+
+    // The value may also flow through in a br_if (the type will indicate that,
+    // which receiveChildValue will notice).
+    receiveChildValue(curr->value, curr);
+  }
+  void visitSwitch(Switch* curr) { handleBreakValue(curr); }
+  void visitLoad(Load* curr) {
+    // We could infer the exact type here, but as no subtyping is possible, it
+    // would have no benefit, so just add a generic root (which will be "Many").
+    addRoot(curr);
```
aheejin, line 307: Can't this be an `ExactType` too?

src/ir/possible-contents.cpp
aheejin: This name is confusing to me... Other locations are not unique? How do you define uniqueness in location sense? For example, are `ExpressionLocation`s not unique?

I prefer `SpecialLocation`, because it is... well, special, but if we want to emphasize that this function makes a new unique location each time it is called, how about keeping `SpecialLocation` struct name as before and rename this function to `makeNewSpecialLocation` or `makeUniqueSpecialLocation` or something? Sorry for reopening the resolved discussions, and I don't feel strongly, but this name is little confusing to me.

src/ir/possible-contents.h
aheejin: It's not that I was insisting we do this; I was mostly asking why the current code is like this.. But anyway if this is clearer and doesn't have any performance hit that should be OK I guess.

But what this fixes is only a single place among many. There are all `visit***` methods in which we just call `addRoot` with `Many` by default, for example, `visitUnary`, `visitBinary`, ..., basically all the places, where they can be `ExactType`. Again, I'm not insisting we should change all this, I'm just asking... So what you meant by earlier comments is that when we actually use this in GUFA, we don't only query `PossibleContents` but also query the actual type of each expression, so that's fine?

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1636 @@
...
+  // worst possible case, as no future value could cause any change.
+  //
+  // Many is always the worst possible case. An exact type of a non-reference is
+  // also the worst case, since subtyping is not relevant there, and so if we
+  // know only the type then we already know nothing beyond what the type in the
+  // wasm tells us (and from there we can only go to Many).
+  bool worthSendingMore = !contents.isMany();
+  if (!contents.getType().isRef() && contents.isExactType()) {
+    worthSendingMore = false;
+  }
```
aheejin, line 1276: > An exact type of a non-reference is also the worst case,

Why? Sorry I think I'm asking the same question over and over... Maybe we should take this and some other remaining questions offline, because I feel I'm holding this PR back for too long time.

The simple mental model I had in mind when I was reading this is a simple LUB kind of model, so you start from ⊥, which is `None`, and you combine types, and then it's possible end state is ⊤, which is `Many`.

src/ir/possible-contents.cpp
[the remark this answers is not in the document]
aheejin: I asked this in the previous comment but I'm not sure if this was answered... `readFromData` uses `connectDuringFlow`, but `writeToData` doesn't. Why?

> Maybe a way to get an intuition for why read is not symmetric with write here is this: When we do a write, we are sending data forward. A read not only reads stuff right now, but also must receive data in the future. So it needs to modify the graph, which a write doesn't.

I'm not sure what you mean by "sending data forward" vs. "receive data in the future". Does "forward" here man the past, because you contrasted it with "future"?

And why does this "receive from future" matter only for `struct.set`/`array.set`? Isn't this the same for all expressions? Like, if any expression's child gets a new `PossibleContents` at some point in the future, the expression's `PossibleContents` should be updated... No?

I'm having a hard time understanding why they are not symmetric, and the concepts of forward vs. future.

> But, this is not so bad actually 😄 It is N operations, but at least **the graph size** is not N, because we can just do a simple loop on the subtypes, and we're done. We don't need to add anything to the graph.

I'm not sure what "the graph size" here means.

I guess this is taking too long, sorry, my understanding is bad. 😢 Maybe it's better to take remaining things offline..

kripken:
@aheejin 

> I prefer SpecialLocation, because it is... well, special, but if we want to emphasize that this function makes a new unique location each time it is called, how about keeping SpecialLocation struct name as before and rename this function to makeNewSpecialLocation or makeUniqueSpecialLocation or something? Sorry for reopening the resolved discussions, and I don't feel strongly, but this name is little confusing to me.

I'd be ok either way. @tlively ?

About an offline meeting, yeah, that sounds good - it does seem like we're not quite understanding each other here, sorry about that.

src/ir/possible-contents.cpp
aheejin: We talked briefly about this offline. Other names like `FreshLocation`, `ConeLocation`, and `AbstractLocation` were suggested. Among the candidates so far (`SpecialLocation`, `UniqueLocation`, `FreshLocation`, `AbstractLocation`, `ConeLocation`), I personally prefer one of `SpecialLocation`, `AbstractLocation`, or `ConeLocation`. I think `UniqueLocation` is confusing because it implies other locations are not unique and it's hard to define uniqueness in the location sense, and `FreshLocation` does not say much about what that is... Are other locations not fresh?

But anyway, it's my personal preferences, and if you too prefer something else, I'm fine with that too. And it's not that this name is super important...

src/ir/possible-contents.cpp
aheejin: The meeting today made this clearer. I think this can be summarized along the lines of "we don't need to send redundant information that can be supplied by the expression types."

src/ir/possible-contents.cpp
aheejin: Sorry, I think this is good as is. I'm not sure why I confused before..

src/ir/possible-contents.h
aheejin: We talked about this offline and I think I now understand better about this thanks to @kripken's patient explanation. I'll try to summarize what I understood:

Generally, we don't want to flow redundant info with `PossibleContents` when that info can be supplied from the expression/local/whatever's type. For example, we can query the expression to see if an expression is `i32`, so flowing `ExactType{i32}` is more work and doesn't really add benefits. When we need additional info in `PossibleContents` is when the expression/local/whatever's type is a reference type, in which case it can be exactly that type or one of its subtypes. So we flow `ExactType` only in those cases, and don't flow `ExactType`s for expressions that deal with non-references, such as `visitUnary`, `visitBinary`, or `visitLoad`.

I asked then why we don't rename `ExactType` to something like `ExactRefType`, given that this is only used for references. @kripken's rationale is we want to make `PossibleContents` as general as possible, so even if GUFA doesn't use `ExactType` in `PossibleContents` for references, other analysis that use `PossibleContents` can make use of that feature, so we don't want to remove an existing capability. In the same vein, we want to make `combine` improvements on `ExactType` done in https://github.com/WebAssembly/binaryen/commit/6c5affd482c96e7a1a076fb84fee86391c835b3f, because this makes `PossibleContents` class fully general and its `combine` now does standard type lattice computations.

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1644 @@
...
+      // type, so we'll pass them through.
+      // TODO: When we get cone types, we could filter the cone here.
+      filtered.combine(contents);
+    }
+    bool mayBeNull = contents.getType().isNullable();
+    if (mayBeNull) {
+      // A null is possible, so pass that along.
+      filtered.combine(
+        PossibleContents::literal(Literal::makeNull(intendedType)));
+    }
```
aheejin, line 1582: There are other places we call `combine`. Why do we do this null check only here?

src/ir/possible-contents.cpp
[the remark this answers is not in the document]
aheejin: We talked about this offline. Thanks @kripken for the patient explanation! I thought about this a little more after the meeting and now I think I understand things better! I'll start with summarizing what I understood, to make sure what I understood is correct. (Let me know if not.)

- The reason we don't pre-create edges with `DataLocation`s for all possible subtypes for `struct.get`/`struct.set` (I'll omit `array.get` and `array.set`, but think of them as included too) is it can create unnecessary many edges for each of possible subtypes in places in which we possibly only end up with only an `ExactType`.
- The reason `writeToData` doesn't use `connectDuringFlow` is we don't want to increase the size of the graph, more precisely the # of edges in the graph. We transfer contents from `value` -> `ref`'s heap type's `DataLocation`, but we don't need to _materialize_ the connections, meaning adding them to `links`, to do this flowing, because we can do that manually each time a `struct.set` is visited in `updateContents`.
- The reason `readFromData` is different from `writeToData` is, `struct.get`'s destination expression needs to be updated each time its heap type's (+ possibly its subtypes') `DataLocation`s are updated. (This is different from other expressions because other expressions generally need to be updated when one of their children is updated.) So we create dynamic connections from the heap type's `DataLocation` -> `struct.get`'s `ExpressionLocation` and, unlike `writeFromData`, also _materialize_ them (= add them to `links`), to _subscribe_ to future updates from the heap type's (+ possibly its subtypes') `DataLocation`s. (These `DataLocation` updates will be done when updating `struct.set`s in the future.) But this can create `O(n*m)` edges; so we introduce `SpecialLocation` that represents cone types, with which we only need to create `O(n+m)` edges.

src/ir/possible-contents.cpp
kripken: I also don't feel strongly. How about `ConeReadLocation`? (just to clarify it is just for reads and not writes)

src/ir/possible-contents.cpp
kripken: Normally we just call `combine()` on all the arriving contents. The tricky thing here is that `ref.cast` filters some things out. So we basically check for the things it would allow through, and pass them through (and so nothing else gets by). Those things are either a proper subtype, or a null, so we have logic for both of those things here.

src/ir/possible-contents.h
```diff
@@ -477,6 +477,34 @@ namespace wasm {
...
+// Therefore the only useful information we can provide on top of the info
+// already in the wasm is either that nothing can be there (1, above), or that a
+// constant must be there (2, above), and so we do not make an effort to track
+// non-reference types here. This makes the internals of ContentOracle simpler
+// and faster. A noticeable outcome of that is that querying the contents of an
+// i32 local will return Many and not ExactType{i32} (assuming we could not
+// infer either that there must be nothing there, or a constant). Again, the
+// caller is assumed to know the wasm IR type anyhow, and also other
+// optimization passes work on the types in the IR, so we do not focus on that
+// here.
```
aheejin, line 507: I think this is very helpful for understanding, but one thing, didn't you say this is not about `PossibleContents` itself but how the user of this class will use it (in our case `InfoCollector`), and you wanted to make this class as general as possible so that it can be possibly used for something else?

That this comment block is attached to `PossibleContents` sounds like all this is an intrinsic property of `PossibleContents`.

>  This makes the internals of ContentOracle simpler and faster.

This is confusing too. I don't think the internals of `ContentsOracle` has changed (except for the `combine` change, which will make this more general); what was made simpler was the internals of `InfoCollector`.. No? But that `InfoCollector` class is also in `possible-contents.cpp`, so yeah, the line between them can be fuzzy though.

> A noticeable outcome of that is that querying the contents of an i32 local will return Many and not ExactType{i32} (assuming we could not infer either that there must be nothing there, or a constant). 

Again, the same thing. These inferences happen not in `PossibleContents` but in `InfoCollector`..

src/ir/possible-contents.cpp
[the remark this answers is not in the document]
aheejin: We seem to be checking two things: 1. Is it a subtype? 2. Can it be a null?

2 is checked separately below in `if (maybeNull)`. Also if they are simply incompatible `HeapType::isSubType` will return false. So if `mayBeSubType` is true, I think there's no possible case this is not a subtype, right? Then shouldn't this be just `isSubType`?

src/ir/possible-contents.cpp
aheejin: I like that too!

src/ir/possible-contents.cpp
aheejin: We talked about this offline and I summarized what I understood in https://github.com/WebAssembly/binaryen/pull/4685#discussion_r891979161.

src/ir/possible-contents.h
kripken: > That this comment block is attached to PossibleContents sounds like all this is an intrinsic property of PossibleContents.

The comment is on `ContentOracle` oracle, though, not `PossibleContents`? But they are in the same file - do you mean we should separate `ContentOracle` out to its own header?

The same with `InfoCollector` - it's only used in `ContentOracle`. Could be clearer in its own file..?

src/ir/possible-contents.cpp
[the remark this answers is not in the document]
kripken: > We seem to be checking two things: 1. Is it a subtype? 2. Can it be a null?

Yes, we check both those things, separately. It might be a subtype but be non-nullable, or it could be a nullable type that is not a subtype, etc., so all 4 combinations can happen here.

> So if mayBeSubType is true, I think there's no possible case this is not a subtype, right? Then shouldn't this be just isSubType?

I can change it to `isSubType` if you prefer, but my reasoning was that it could also be null (which has no type, really). And I wanted to keep the two checks, of subtype and null, separate - so each of the checks is phrased as "may be X". We are only thinking about one aspect of the contents at each time. But, happy to do it either way.

src/ir/possible-contents.cpp
kripken: Ok, refactored to `ConeReadLocation`. This does make the code simpler, to just focus on the concrete use case for now. Seems worth it, as I'm not sure if we'll need something more later that would benefit from a generic "unique/special" location.

src/ir/possible-contents.h
aheejin: Oh, sorry, I was confused, I thought it was `PossibleContents`... why... anyway thanks nevermind. 😅

src/ir/possible-contents.cpp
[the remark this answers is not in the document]
aheejin: No I'm mostly asking.

- Subtype + non-nullable: `mayBeSubType` is true
- Subtype + nullable: `mayBeSubType` is true
- Not subtype + non-nullable: `mayBeSubType` is false
- Not subtype + nullable: `mayBeSubType` is false

What I meant was, is there any case when this is not a subtype when `mayBeSubType` is true? If this is a subtype whenever `mayBeSubType` is true, why is there 'maybe'?

---

However we end up, how about changing this to `maybe` (not `mayBe`)? We use 'maybe` in other places: https://github.com/WebAssembly/binaryen/blob/0b66981d0e3aa958877b5c04769e75836497b4c0/src/wasm-binary.h#L1637-L1681

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1652 @@
...
+    // clear our targets vector. But we should have already removed all the
+    // targets in the above remove_if operation, since they should have all
+    // notified us that we do not need to send them any more updates.
+    assert(targets.empty());
+  }
+}
+
+void Flower::connectDuringFlow(Location from, Location to) {
+  auto newLink = LocationLink{from, to};
+  auto newIndexLink = getIndexes(newLink);
```
tlively, line 1385: Why do links need indices, again? I would expect each location to just store a vector of its successor location indices with no need to separately materialize links.

tlively (commented):
Read another few hundred lines...

aheejin (approved):
Except for this minor thing, I think all my questions are addressed: https://github.com/WebAssembly/binaryen/pull/4685#discussion_r892971155
(It's getting hard to track unresolved comment threads in this PR, which has 190 comments now...)

Thanks a lot for your patience!

src/ir/possible-contents.cpp
```diff
@@ -0,0 +1,1676 @@
...
+  }
+
+#ifdef POSSIBLE_CONTENTS_DEBUG
+  std::cout << "struct phase\n";
+#endif
+
+  if (getTypeSystem() == TypeSystem::Nominal ||
+      getTypeSystem() == TypeSystem::Isorecursive) {
+    subTypes = std::make_unique<SubTypes>(wasm);
+  }
```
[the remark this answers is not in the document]
kripken: ContentOracle is useful even on non-GC wasm. It can infer i32 constants, for example.

(Tests in the next PR show that explicitly.)

src/ir/possible-contents.cpp
kripken: Correct, each location just stores a vector of successor indexes. That's what the code
```cpp
// Add it to the |targets| vector.
```
does a few lines down. However, we also keep `links` around simply to be able to tell what links exist in the program. The comment on `links` explains that,
```cpp
  // All existing links in the graph. We keep this to know when a link we want
  // to add is new or not.
  std::unordered_set<IndexLink> links;
```
(Instead, we could scan through the `targets` vector to avoid adding a link that already exists, but that would take linear time.)

src/ir/possible-contents.cpp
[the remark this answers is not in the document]
aheejin: Is null a not subtype? Then, `(ref.null func)` is not a subtype of `funcref`? Didn't know that...? Anyway, sorry if this made the comment more verbose and harder to understand... I was mostly asking so please revert that if you see that is clearer.

On `maybe` vs. `mayBe`, my English is bad, sorry 😅 So nevermind.. I remembered I had seen `maybe` in other places in the code so I suggested that but I guess the meaning here was different after all.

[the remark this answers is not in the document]
kripken: `(ref null func)` is a subtype of `funcref`. But the runtime value, `null`, has no type. That is, there is a single `null` for all types. So, to me at least, it seems weird to say that it must be a subtype of the heap type, if it can be a null (but it makes sense to say it's a subtype of the *type* - since that includes nullability). Anyhow, again, maybe that's a matter of definition or perspective...

Let's see what @tlively thinks. I'm fine either way here.

aheejin:
Because this now has hundreds of comments and most of them are folded, I'll paste links to my comments with which I tried to summarize what I learned from @kripken in our offline meeting last week (it was actually not 'offline', but anyway), mainly for my own reference in case I forget these later.
- Asymmetry between `readFromData` and `writeToData`: https://github.com/WebAssembly/binaryen/pull/4685#discussion_r892001319
- Why we don't flow `ExactType`s for non-reference types: https://github.com/WebAssembly/binaryen/pull/4685#discussion_r891979161

kripken:
@tlively 

Friendly ping. If you don't have time that's fine too, as I think this PR is in good shape now. The next PR may be more "interesting" as that is where the actual GUFA optimization pass + tests arrive.

tlively:
yikes! Sorry for the long wait. Taking a last look now.

tlively (approved):
Didn't read the last bit too carefully, but I don't want to hold this up any longer.

src/ir/possible-contents.cpp
kripken: I've never used a bloom filter before - interesting!

PR #4686, opened 2022-05-25
[Parser] Lex idchar and identifiers

tlively:


tlively:
* main
  * **PR #4686** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4686" target="_blank">Graphite</a>  👈
    * **PR #4687** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4687" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4686?utm_source=stack-comment).

src/wasm/wat-parser-internal.h
```diff
@@ -372,6 +372,72 @@ std::optional<LexIntResult> integer(std::string_view in) {
...
+//          | 'a' | ... | 'z'
+//          | '!' | '#' | '$' | '%' | '&' | ''' | '*' | '+'
+//          | '-' | '.' | '/' | ':' | '<' | '=' | '>' | '?'
+//          | '@' | '\' | '^' | '_' | '`' | '|' | '~'
+std::optional<LexResult> idchar(std::string_view in) {
+  LexCtx ctx(in);
+  if (ctx.empty()) {
+    return {};
+  }
+  char c = ctx.next()[0];
```
aheejin, line 386: Would it be better to create a method like `peek(int)`, which checks the string without consuming it?

src/wasm/wat-parser-internal.h
```diff
@@ -372,6 +372,72 @@ std::optional<LexIntResult> integer(std::string_view in) {
...
+      case '=':
+      case '>':
+      case '?':
+      case '@':
+      case '\\':
+      case '^':
+      case '_':
+      case '`':
+      case '|':
+      case '~':
```
aheejin, line 414: Can we use all these chars within names? Didn't know that...

src/wasm/wat-parser-internal.h
```diff
@@ -405,8 +471,16 @@ struct IntTok {
   }
 };
 
+struct IdTok {
```
aheejin, line 474: `IdTok` doesn't have the actual ID string attached. Then why do we need `n` for `IntTok`?

src/wasm/wat-parser-internal.h
tlively: Good idea, I'll add that.

src/wasm/wat-parser-internal.h
tlively: `IdTok` doesn't need to perform any transformations on the span stored in the parent `Token`, so it doesn't need to carry any data. In contrast, it would be unfortunate to have to re-parse an integer token every time we wanted its value.

src/wasm/wat-parser-internal.h
```diff
@@ -68,6 +68,9 @@ struct LexCtx {
   // The next input that has not already been lexed.
   std::string_view next() const { return input.substr(lexedSize); }
 
+  // Get the next character without consuming it.
+  uint8_t peek() const { return next()[0]; }
```
aheejin, line 72: I think this can even take an optional integer argument for length (which defaults to 1)

tlively: So far I haven't ever needed to peek beyond the next character, but if a use case for that comes up, I will definitely update the method.

PR #4688, opened 2022-05-25
[Parser] Lex keywords

tlively:
Also include reserved words that look like keywords to avoid having to find and
enumerate all the valid keywords. Invalid keywords will be rejected at a higher
level in the parser instead.

### DIFF
--- a/src/wasm/wat-parser-internal.h
+++ b/src/wasm/wat-parser-internal.h
@@ -597,7 +597,7 @@ std::optional<LexResult> keyword(std::string_view in) {
   if (ctx.empty()) {
     return {};
   }
-  char start = ctx.next()[0];
+  uint8_t start = ctx.peek();
   if ('a' <= start && start <= 'z') {
     ctx.take(1);
   } else {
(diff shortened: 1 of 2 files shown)


tlively:
**Graphite Merge Job**

Current status: ⏳ Queued to merge

This pull request is currently queued to be merged as part of a stack.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4688?utm_source=stack-comment).


Job Reference: xUqRAsHM2tUWW7RRZ6sh

src/wasm/wat-parser-internal.h
```diff
@@ -585,6 +585,30 @@ std::optional<LexStrResult> str(std::string_view in) {
   return ctx.lexed();
 }
 
+// keyword ::= ( 'a' | ... | 'z' ) idchar* (if literal terminal in grammar)
+// reserved ::= idchar+
```
aheejin, line 589: What are "literal terminal" and "reserved"? My guess is the first is the reserved keywords in wasm text grammar like `module`, `type`, or `func`, and the latter is just any other random sequence of chars, is that correct? If so it's ironical that the latter's name is "reserved".

tlively: Your guess is correct and yep, it is a little ironic. Here's where the spec talks about this: https://webassembly.github.io/spec/core/text/lexical.html#text-reserved.

PR #4690, opened 2022-05-25
[Wasm GC] Fix CFG traversal of call_ref and add missing validation check

### LINKED ISSUE #4689: Live code removed inside `catch` block
Optimizing [this module](https://github.com/WebAssembly/binaryen/files/8771181/nic2.wasm.gz) with `wasm-opt -all -O1 --nominal -g` results in the removal of an object allocation whose result is used later.

The offending function looks like this unoptimized:
```wat
 (func $GestureRecognizer.invokeCallback (type $type$38) (param $0 (ref $type$8)) (param $1 (ref null $type$3)) (param $2 (ref $type$28)) (result (ref null $type$28))
  (local $3 (ref null $type$28))
  (local $4 (ref null $type$26))
  (local $5 (ref null $type$8))
  (local $6 (ref null $type$28))
  (local $7 (ref null $type$28))
  (local $8 (ref null $type$8))
  (local $9 (ref null $type$39))
  (local $10 ((ref null $type$28) (ref null $type$8)))
  (local $11 (ref null $type$28))
  (local $12 (ref null $type$39))
  (block $label$2
   (try $label$4
    (do
     (block $label$1
      (local.set $3
       (call_ref
        (struct.get $type$26 2
         (local.tee $4
          (ref.cast_static $type$26
           (local.get $2)
          )
         )
        )
        (struct.get $type$26 3
         (local.get $4)
        )
       )
      )
      (br $label$1)
     )
    )
    (catch $tag$0
     (local.set $10
      (pop (ref $type$28) (ref $type$8))
     )
     (local.set $6
       (local.set $11
        (ref.as_non_null
         (tuple.extract 0
          (local.get $10)
         )
        )
       )
       (local.set $5
        (ref.as_non_null
         (tuple.extract 1
          (local.get $10)
         )
        )
       )
       (ref.as_non_null
        (local.get $11)
       )
      )
     )
     (block $label$3
      (local.set $7
       (local.get $6)
      )
      (local.set $8
       (local.get $5)
      )
      (call $FlutterErrorDetails.
       (ref.as_non_null
        (block (result (ref null $type$39))
         (local.set $12
          (local.tee $9
           (struct.new_default $type$39)
          )
         )
         (struct.set $type$39 0
          (local.get $9)
          (i32.const 4)
         )
         (local.get $12)
        )
       )
       (ref.as_non_null
        (local.get $7)
       )
      )
      (call $FlutterError.reportError
       (ref.as_non_null
        (local.get $9)
       )
      )
      (br $label$2)
     )
    )
   )
  )
  (return
   (local.get $3)
  )
 )
```
and like this optimized:
```wat
 (func $GestureRecognizer.invokeCallback (type $type$38) (param $0 (ref $type$7)) (param $1 (ref null $type$3)) (param $2 (ref $type$28)) (result (ref null $type$28))
  (local $3 (ref null $type$28))
  (local $4 (ref null $type$39))
  (local $5 (ref null $type$26))
  (local $6 ((ref null $type$28) (ref null $type$7)))
  (block $label$2
   (try
    (do
     (local.set $3
      (call_ref
       (struct.get $type$26 2
        (local.tee $5
         (ref.cast_static $type$26
          (local.get $2)
         )
        )
       )
       (struct.get $type$26 3
        (local.get $5)
       )
      )
     )
    )
    (catch $tag$0
     (drop
     )
     (drop
      (ref.as_non_null
       (tuple.extract 0
        (local.get $6)
       )
      )
     )
     (drop
      (ref.as_non_null
       (tuple.extract 1
        (local.get $6)
       )
      )
     )
     (struct.set $type$39 0
      (local.get $4)
      (i32.const 4)
     )
     (call $FlutterErrorDetails.
      (ref.as_non_null
       (local.get $4)
      )
      (ref.as_non_null
       (local.get $3)
      )
     )
     (call $FlutterError.reportError
      (ref.as_non_null
       (local.get $4)
      )
     )
     (br $label$2)
    )
   )
  )
  (local.get $3)
 )
```
The `struct.set $type$39 0` originally operates on the result of the `struct.new_default $type$39`, stored into local `$9`. But in the optimized code, the `struct.new_default` is missing, and the `struct.set` operates on the uninitialized local `$4`.

With `-O2` or higher and `--enable-gc-nn-locals`, Binaryen produces an error message:
```
Fatal: Non-nullable local accessing the default value in GestureRecognizer.invokeCallback (1)
```

kripken:
We were missing `CallRef` in the CFG traversal code in a place where we
note possible exceptions. As a result we thought `CallRef` cannot throw, and
were missing some control flow edges.

To actually detect the problem, we need to validate non-nullable locals
properly, which we were not doing. This adds that as well.

This could be split into two PRs if we prefer, one to add validation, and one
to add the CFG fix, but then full testing would only arrive with the second.
Happy to do that if we prefer, but OTOH this PR is fairly small.

Fixes #4689

cc @askeksa-google

### DIFF
--- a/src/wasm/wasm-validator.cpp
+++ b/src/wasm/wasm-validator.cpp
@@ -2798,14 +2798,15 @@ void FunctionValidator::visitFunction(Function* curr) {
     if (hasNNLocals) {
       LocalGraph graph(curr);
       for (auto& [get, sets] : graph.getSetses) {
-        if (curr->getLocalType(get->index).isNonNullable()) {
-          for (auto* set : sets) {
-            // If this is a null, then it is the value of a param or the default
-            // value of a var. The latter case is an error.
-            shouldBeTrue(!!set || curr->isParam(get->index),
-                         get->index,
-                         "non-nullable local must not read null");
-          }
+        auto index = get->index;
+        // It is always ok to read nullable locals, and it is always ok to read
+        // params even if they are non-nullable.
+        if (!curr->getLocalType(index).isNonNullable() ||
+            curr->isParam(index)) {
+          continue;
+        }
+        for (auto* set : sets) {
+          shouldBeTrue(!!set, index, "non-nullable local must not read null");
         }
       }
     }
--- a/src/wasm/wasm-validator.cpp
+++ b/src/wasm/wasm-validator.cpp
@@ -23,6 +23,7 @@
 #include "ir/features.h"
 #include "ir/global-utils.h"
 #include "ir/intrinsics.h"
+#include "ir/local-graph.h"
 #include "ir/module-utils.h"
 #include "ir/stack-utils.h"
 #include "ir/utils.h"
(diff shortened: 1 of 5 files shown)


src/wasm/wasm-validator.cpp
```diff
@@ -2779,6 +2780,36 @@ void FunctionValidator::visitFunction(Function* curr) {
...
+      }
+    }
+    if (hasNNLocals) {
+      LocalGraph graph(curr);
+      for (auto& [get, sets] : graph.getSetses) {
+        if (curr->getLocalType(get->index).isNonNullable()) {
+          for (auto* set : sets) {
+            // If this is a null, then it is the value of a param or the default
+            // value of a var. The latter case is an error.
+            shouldBeTrue(!!set || curr->isParam(get->index),
```
tlively, line 2805: We could hoist the check for whether this local is a param out of this loop to avoid some unnecessary work.

tlively (approved):

src/wasm/wasm-validator.cpp
kripken: Good idea, simplified now.

PR #4692, opened 2022-05-26
Add C API for printing stack IR

ApsarasX:
I want to write stack IR text form of wasm module to a specific file, but it can only be written to stdout for the lack of function in the existing C API.

So I add the following functions into C API for my needs.

- `BinaryenModulePrintStackIR`: similar to `BinaryenModulePrint`
- `BinaryenModuleWriteStackIR`: similar to `BinaryenModuleWriteText`
- `BinaryenModuleAllocateAndWriteStackIR`: similar to `BinaryenModuleAllocateAndWriteText`

I know my code is not so perfect. If there is a better way, please let me know.

### DIFF
--- a/src/binaryen-c.h
+++ b/src/binaryen-c.h
@@ -2471,8 +2471,8 @@ BINARYEN_API size_t BinaryenModuleWriteText(BinaryenModuleRef module,
 // @return how many bytes were written. This will be less than or equal to
 //         outputSize
 BINARYEN_API size_t BinaryenModuleWriteStackIR(BinaryenModuleRef module,
-                                            char* output,
-                                            size_t outputSize);
+                                               char* output,
+                                               size_t outputSize);
 
 typedef struct BinaryenBufferSizes {
   size_t outputBytes;
@@ -2517,8 +2517,8 @@ BINARYEN_API char* BinaryenModuleAllocateAndWriteText(BinaryenModuleRef module);
 // Serialize a module in stack IR form. Implicitly allocates the returned
 // char* with malloc(), and expects the user to free() them manually
 // once not needed anymore.
-BINARYEN_API char* BinaryenModuleAllocateAndWriteStackIR(BinaryenModuleRef
-                                                           module);
+BINARYEN_API char*
+BinaryenModuleAllocateAndWriteStackIR(BinaryenModuleRef module);
 
 // Deserialize a module from binary form.
 BINARYEN_API BinaryenModuleRef BinaryenModuleRead(char* input,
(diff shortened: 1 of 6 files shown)


src/binaryen-c.h
```diff
@@ -2504,6 +2514,12 @@ BinaryenModuleAllocateAndWrite(BinaryenModuleRef module,
 // once not needed anymore.
 BINARYEN_API char* BinaryenModuleAllocateAndWriteText(BinaryenModuleRef module);
 
+// Serialize a module in stack IR form. Implicity allocates the returned
```
kripken, line 2517:
```suggestion
// Serialize a module in stack IR form. Implicitly allocates the returned
```

src/pass.h
```diff
@@ -443,6 +444,8 @@ class WalkerPass : public Pass, public WalkerType {
   void setPassRunner(PassRunner* runner_) { runner = runner_; }
 };
 
+std::ostream& printStackIR(std::ostream& o, Module* module);
```
kripken, line 447: This doesn't seem like the right place for this - it doesn't fit in with anything around it, AFAICT.

I'm not sure offhand what the right place is, but I'd look for similar functions and where they are defined.

ApsarasX: Yes, the location of this function is really inappropriate.

But its definition will be used by the C API, and it will also access `GenerateStackIR` and `PrintStackIR`, I can't find a better place for it.

Hope you can find a more suitable location for it.

kripken: Hmm, how about at the very end of `wasm-stack.h`? We have similar things at the very end of `wasm.h` for non-stack IR, for comparison.

ApsarasX: Good idea! I have moved printStackIR's declaration into `wasm-stack.h`.

src/pass.h
```diff
@@ -20,6 +20,7 @@
 #include <functional>
 
 #include "mixed_arena.h"
+#include "passes/passes.h"
```
kripken, line 23: Why is this here?

src/passes/Print.cpp
```diff
@@ -3442,6 +3442,15 @@ printStackIR(StackIR* ir, std::ostream& o, Function* func) {
   return o;
 }
 
+std::ostream& printStackIR(std::ostream& o, Module* module) {
+  wasm::PassRunner runner(module);
+  Pass* generatePass = createGenerateStackIRPass();
```
kripken, line 3447: Instead of this, you can do `runner.add(..)` to create and add the pass, then do `runner.run()`. That would avoid manual memory management here.

src/pass.h
ApsarasX: This was originally intended to access `createGenerateStackIRPass`, but is no longer needed.

src/passes/Print.cpp
ApsarasX: Have modified as you suggested

kripken (approved):

kripken:
Looks good to land aside from a lint error, see the CI log there. Running `clang-format` should fix it, or you can manually apply the patch from the log.

PR #4693, opened 2022-05-26
[Parser] Lex floating point values

tlively:
Rather than trying to actually implement the parsing of float values, which
cannot be done naively due to precision concerns, just parse the float grammar
then postprocess the parsed text into a form we can pass to `strtod` to do the
actual parsing of the value.

Since the float grammar reuses `num` and `hexnum` from the integer grammar but
does not care about overflow, add a mode to `LexIntCtx`, `num`, and `hexnum` to
allow parsing overflowing numbers.

For NaNs, store the payload as a separate value rather than as part of the
parsed double. The payload will be injected into the NaN at a higher level of
the parser once we know whether we are parsing an f64 or an f32 and therefore
know what the allowable payload values are.

### DIFF
--- a/src/wasm/wat-parser-internal.h
+++ b/src/wasm/wat-parser-internal.h
@@ -242,7 +242,7 @@ struct LexFloatCtx : LexCtx {
   LexFloatCtx(std::string_view in) : LexCtx(in) {}
 
   std::optional<LexFloatResult> lexed() {
-    static_assert(!std::signbit(NAN), "Expected NAN to be positive");
+    assert(!std::signbit(NAN) && "Expected NAN to be positive");
     auto basic = LexCtx::lexed();
     if (!basic) {
       return {};
--- a/src/wasm/wat-parser-internal.h
+++ b/src/wasm/wat-parser-internal.h
@@ -242,6 +242,7 @@ struct LexFloatCtx : LexCtx {
   LexFloatCtx(std::string_view in) : LexCtx(in) {}
 
   std::optional<LexFloatResult> lexed() {
+    static_assert(!std::signbit(NAN), "Expected NAN to be positive");
     auto basic = LexCtx::lexed();
     if (!basic) {
       return {};
@@ -249,6 +250,10 @@ struct LexFloatCtx : LexCtx {
     if (nanPayload) {
       double nan = basic->span[0] == '-' ? -NAN : NAN;
       return LexFloatResult{*basic, nanPayload, nan};
+    }
+    // strtod does not return -NAN for "-nan" on all platforms.
+    if (basic->span == "-nan"sv) {
+      return LexFloatResult{*basic, nanPayload, -NAN};
     }
     // Do not try to implement fully general and precise float parsing
     // ourselves. Instead, call out to std::strtod to do our parsing. This means
@@ -594,7 +599,7 @@ std::optional<LexFloatResult> float_(std::string_view in) {
     if (ctx.takePrefix(":0x"sv)) {
       if (auto lexed = hexnum(ctx.next())) {
         ctx.take(*lexed);
-        if (1 <= lexed->n && lexed->n < (1ul << 52)) {
+        if (1 <= lexed->n && lexed->n < (1ull << 52)) {
           ctx.nanPayload = lexed->n;
         } else {
           // TODO: Add error production for invalid NaN payload.
--- a/src/wasm/wat-parser-internal.h
+++ b/src/wasm/wat-parser-internal.h
@@ -26,6 +26,7 @@
 
 #include <cassert>
 #include <cctype>
+#include <cmath>
 #include <iostream>
 #include <optional>
 #include <sstream>
@@ -107,6 +108,28 @@ public:
 
 enum Signedness { Unsigned, Signed };
 
+enum OverflowBehavior { DisallowOverflow, IgnoreOverflow };
+
+std::optional<int> getDigit(char c) {
+  if ('0' <= c && c <= '9') {
+    return {c - '0'};
+  }
+  return std::nullopt;
+}
+
+std::optional<int> getHexDigit(char c) {
+  if ('0' <= c && c <= '9') {
+    return {c - '0'};
+  }
+  if ('A' <= c && c <= 'F') {
+    return {10 + c - 'A'};
+  }
+  if ('a' <= c && c <= 'f') {
+    return {10 + c - 'a'};
+  }
+  return std::nullopt;
+}
+
 // The result of lexing an integer token fragment.
 struct LexIntResult : LexResult {
   uint64_t n;
@@ -124,28 +147,16 @@ private:
   bool negative = false;
   bool overflow = false;
 
-  std::optional<int> getDigit(char c) {
-    if ('0' <= c && c <= '9') {
-      return {c - '0'};
-    }
-    return std::nullopt;
-  }
-
-  std::optional<int> getHexDigit(char c) {
-    if ('0' <= c && c <= '9') {
-      return {c - '0'};
-    }
-    if ('A' <= c && c <= 'F') {
-      return {10 + c - 'A'};
-    }
-    if ('a' <= c && c <= 'f') {
-      return {10 + c - 'a'};
-    }
-    return std::nullopt;
-  }
-
 public:
   explicit LexIntCtx(std::string_view in) : LexCtx(in) {}
+
+  // Lex only the underlying span, ignoring the overflow and value.
+  std::optional<LexIntResult> lexedRaw() {
+    if (auto basic = LexCtx::lexed()) {
+      return LexIntResult{*basic, 0, Unsigned};
+    }
+    return {};
+  }
 
   std::optional<LexIntResult> lexed() {
     // Check most significant bit for overflow of signed numbers.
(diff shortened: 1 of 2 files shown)


tlively:
* main
  * **PR #4693** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4693" target="_blank">Graphite</a>  👈
    * **PR #4694** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4694" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4693?utm_source=stack-comment).

src/wasm/wat-parser-internal.h
```diff
@@ -217,6 +228,49 @@ struct LexIntCtx : LexCtx {
...
+                    *end = curr + basic->span.size();
+         curr != end;
+         ++curr) {
+      if (*curr != '_') {
+        ss << *curr;
+      }
+    }
+    std::string str = ss.str();
+    char* last;
+    double d = std::strtod(str.data(), &last);
```
kripken, line 268: The existing code uses `strtof` for an `f32`, but I'm not sure if that's necessary...

test/gtest/wat-parser.cpp
```diff
@@ -347,6 +356,483 @@ TEST(ParserTest, LexHexInt) {
...
+  }
+  {
+    Lexer lexer("NAN"sv);
+    EXPECT_EQ(lexer, lexer.end());
+  }
+  {
+    Lexer lexer("NaN"sv);
+    EXPECT_EQ(lexer, lexer.end());
+  }
+}
```
kripken, line 834: I'd say the spec tests may be good enough coverage for corner cases. But if you prefer to write out unit tests like this I'm not opposed.

src/wasm/wat-parser-internal.h
tlively: There's no way to tell whether or not we are lexing an f32 or an f64 without higher level parser context, so we have to conservatively use double precision here. I don't think there are any problems that can arise from that.

test/gtest/wat-parser.cpp
tlively: We're a while off from being able to run spec tests with this new parser and the unit tests have been helpful for catching bugs.

src/wasm/wat-parser-internal.h
```diff
@@ -107,6 +108,28 @@ struct LexCtx {
...
+std::optional<int> getDigit(char c) {
+  if ('0' <= c && c <= '9') {
+    return {c - '0'};
+  }
+  return std::nullopt;
+}
+
+std::optional<int> getHexDigit(char c) {
+  if ('0' <= c && c <= '9') {
+    return {c - '0'};
```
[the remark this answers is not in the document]
tlively: I'll remove these and other unnecessary braces in a separate NFC PR.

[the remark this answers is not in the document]
kripken: Sounds good.

test/gtest/wat-parser.cpp
```diff
@@ -347,6 +356,483 @@ TEST(ParserTest, LexHexInt) {
...
+  {
+    Lexer lexer("-nan"sv);
+    ASSERT_NE(lexer, lexer.end());
+    Token expected{"-nan"sv, FloatTok{{}, -NAN}};
+    EXPECT_EQ(*lexer, expected);
+  }
+  {
+    Lexer lexer("nan:0x01"sv);
+    ASSERT_NE(lexer, lexer.end());
+    Token expected{"nan:0x01"sv, FloatTok{{1}, NAN}};
```
kripken, line 744: Do we not need these NaN payloads for some spec tests?

kripken: I see, sounds good.

kripken (approved):

tlively:
**Graphite Merge Job**

Current status: 🔄 Waiting for checks to pass

This pull request is currently waiting for CI and mergeability checks to pass before being merged.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4693?utm_source=stack-comment).


Job Reference: A9XO0urDPXHWLrhQjA6D

src/wasm/wat-parser-internal.h
```diff
@@ -217,6 +228,54 @@ struct LexIntCtx : LexCtx {
...
+
+  std::optional<LexFloatResult> lexed() {
+    assert(!std::signbit(NAN) && "Expected NAN to be positive");
+    auto basic = LexCtx::lexed();
+    if (!basic) {
+      return {};
+    }
+    if (nanPayload) {
+      double nan = basic->span[0] == '-' ? -NAN : NAN;
+      return LexFloatResult{*basic, nanPayload, nan};
```
aheejin, line 252: Does this initialize base class members in order and then the child class members? Didn't know this was possible..

src/wasm/wat-parser-internal.h
```diff
@@ -217,6 +228,54 @@ struct LexIntCtx : LexCtx {
...
+    }
+    // strtod does not return -NAN for "-nan" on all platforms.
+    if (basic->span == "-nan"sv) {
+      return LexFloatResult{*basic, nanPayload, -NAN};
```
aheejin, line 256:
```suggestion
      return LexFloatResult{*basic, {}, -NAN};
```
Can `nanPayload` ever be non-null here, given that the `if` above does an early return?

src/wasm/wat-parser-internal.h
```diff
@@ -642,6 +816,31 @@ struct IntTok {
   }
 };
 
+struct FloatTok {
+  // The payload if we lexed a nan with payload. We cannot store the payload
+  // directly in `d` because we do not know at this point whether we are parsing
+  // an f32 or f64 and therefore we do not know what the allowable payloads are.
```
aheejin, line 822: For non-NaN numbers, you store as a double conservatively and fix it later. Can't we do the same for NaNs?

test/gtest/wat-parser.cpp
```diff
@@ -347,6 +356,483 @@ TEST(ParserTest, LexHexInt) {
...
+    ASSERT_NE(lexer, lexer.end());
+    Token expected{"+nan"sv, FloatTok{{}, NAN}};
+    EXPECT_EQ(*lexer, expected);
+  }
+  {
+    Lexer lexer("-nan"sv);
+    ASSERT_NE(lexer, lexer.end());
+    Token expected{"-nan"sv, FloatTok{{}, -NAN}};
+    EXPECT_EQ(*lexer, expected);
+  }
```
aheejin, line 740: Do we not need payload for `nan`? The spec says `2^(signif(N)-1)`
```
'nan' => nan(2^(signif(N)-1))
```

test/gtest/wat-parser.cpp
```diff
@@ -347,6 +356,483 @@ TEST(ParserTest, LexHexInt) {
...
+    ASSERT_NE(lexer, lexer.end());
+    Token expected{"nan:0x10_0000_0000_0000"sv, KeywordTok{}};
+    EXPECT_EQ(*lexer, expected);
+  }
+  {
+    Lexer lexer("nan:0x1_0000_0000_0000_0000"sv);
+    ASSERT_NE(lexer, lexer.end());
+    Token expected{"nan:0x1_0000_0000_0000_0000"sv, KeywordTok{}};
+    EXPECT_EQ(*lexer, expected);
+  }
```
aheejin, line 825: Aren't these `'nan:0x'n:hexnum` pattern?

test/gtest/wat-parser.cpp
```diff
@@ -347,6 +356,483 @@ TEST(ParserTest, LexHexInt) {
...
+  {
+    Lexer lexer("Inf"sv);
+    EXPECT_EQ(lexer, lexer.end());
+  }
+  {
+    Lexer lexer("INF"sv);
+    EXPECT_EQ(lexer, lexer.end());
+  }
```
aheejin, line 713: We don't recognize these as keywords? The same for `NaN`/`NAN`.

aheejin (commented):
I guess most of the questions stem from my lack of knowledge about floating point representation, but anyway..

src/wasm/wat-parser-internal.h
tlively: Yep, that's my understanding. I wouldn't have thought that this would work either, but apparently it does :shrug:

src/wasm/wat-parser-internal.h
tlively: No, it will always be nullopt. I can make this change in a follow-up.

src/wasm/wat-parser-internal.h
tlively: I thought about that, but the problem is that the default payload depends on the size of the float we are parsing. It's (2^52)-1 for f64 and (2^23)-1 for f32. If we stored the payload directly in the double to be fixed up later, then we wouldn't be able to tell the difference between a custom payload of (2^52)-1 that is out-of-bounds for an f32 and a default payload that was conservatively set to (2^53)-1 and can be fixed up. We could fix that in different ways, like having a bool `isDefaultPayload`, but this seems simpler overall.

test/gtest/wat-parser.cpp
tlively: Good observation. We will interpret a null payload to mean the default payload when `d` is NaN and inject that default payload in the same part of the parser (not implemented yet) where we will inject non-default payloads.

test/gtest/wat-parser.cpp
tlively: They would be, but `n` is outside the allowable bounds for f64 NaN payloads, so these fail to parse as floats. The lexer then tries to lex them as keywords, which succeeds. In the future once we have better error handling here, we should produce an error about the payload being out of bounds instead of falling back to parsing it as a keyword.

test/gtest/wat-parser.cpp
tlively: No, keywords can only start with lowercase letters :laughing:

test/gtest/wat-parser.cpp
aheejin: I see. Might be helpful for reading to add a comment after this block:
https://github.com/WebAssembly/binaryen/blob/590af8639ab4f6cfd82452b5ff2b431b470c64a5/src/wasm/wat-parser-internal.h#L602-L615

test/gtest/wat-parser.cpp
aheejin: We don't even allow `0x0`... 0 is also outside the allowable bounds?

tlively: That's right. Apparently if all the payload bits are zero, then the value is +-infinity rather than a NaN.

PR #4698, opened 2022-05-27
[Parser] Replace Signedness with ternary Sign

tlively:
Previously we were tracking whether integer tokens were signed but we did not
differentiate between positive and negative signs. Unfortunately, without
differentiating them, there's no way to tell the difference between an in-bounds
negative integer and a wildly out-of-bounds positive integer when trying to
perform bounds checks for s32 tokens. Fix the problem by tracking not only
whether there is a sign on an integer token, but also what the sign is.

### DIFF
--- a/src/wat-lexer.h
+++ b/src/wat-lexer.h
@@ -55,11 +55,11 @@ struct IdTok {
   friend std::ostream& operator<<(std::ostream&, const IdTok&);
 };
 
-enum Signedness { Unsigned, Signed };
+enum Sign { NoSign, Pos, Neg };
 
 struct IntTok {
   uint64_t n;
-  Signedness signedness;
+  Sign sign;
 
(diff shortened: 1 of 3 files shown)


kripken (approved):

src/wat-lexer.h
```diff
@@ -55,11 +55,11 @@ struct IdTok {
   friend std::ostream& operator<<(std::ostream&, const IdTok&);
 };
 
-enum Signedness { Unsigned, Signed };
+enum Sign { NoSign, Pos, Neg };
```
aheejin, line 58: Is there any reason we need to have both `NoSign` and `Pos`? Can't we make the default `Pos`, and change it to `Neg` only when we take `-`?

aheejin: Hmm nevermind. We need to distinguish u32 vs. s32...

PR #4703, opened 2022-05-31
[WasmGC] Fix non-nullable locals and try-catch in SimplifyLocals

### LINKED ISSUE #4702: Moving local initialization into `try` block breaks flow-sensitive nn local validation
Optimizing [this module](https://github.com/WebAssembly/binaryen/files/8800049/cur.wasm.gz) with `wasm-opt -all -O1 --nominal --enable-gc-nn-locals -g` results in a module that fails validation using the [experimental flow-sensitive non-defaultable local validation in V8](https://chromium-review.googlesource.com/c/v8/v8/+/3650926) (the original module passes validation with this patch).

The culprit is a method where the downcast of the receiver is moved inside a `try` block by Binaryen, but the local containing the type-precise receiver is used inside the `catch` block. The proposed validation scheme requires that such a local is initialized before the `try` block (as it is in the original module).

The method in question is optimized from
```wat
 (func $B.test (type $type$39) (param $0 (ref $type$7)) (result (ref $type$7))
  (local $1 (ref $type$37))
  (local $2 (ref $type$37))
  (local $3 (ref $type$7))
  (local $4 (ref $type$28))
  (local $5 (ref $type$28))
  (local $6 ((ref $type$28) (ref $type$7)))
  (local $7 (ref $type$28))
  (local.set $1
   (ref.cast_static $type$37
    (local.get $0)
   )
  )
  (try $label$4
   (do
    (call_indirect $0 (type $type$41)
     (local.tee $2
      (local.get $1)
     )
     (i32.add
      (struct.get $type$37 0
       (local.get $2)
      )
      (i32.const 37)
     )
    )
    (return
     (ref.as_non_null
      (struct.get $type$37 2
       (local.get $1)
      )
     )
    )
   )
   (catch $tag$0
    (local.set $6
    )
    (local.set $4
      (local.set $7
       (tuple.extract 0
        (local.get $6)
       )
      )
      (local.set $3
       (tuple.extract 1
        (local.get $6)
       )
      )
      (local.get $7)
     )
    )
    (block $label$3
     (local.set $5
      (local.get $4)
     )
     (return
      (call $B.transform
       (ref.as_non_null
        (struct.get $type$37 2
         (local.get $1)
        )
       )
      )
     )
    )
   )
  )
 )
```
to
```wat
 (func $B.test (type $type$38) (param $0 (ref $type$6)) (result (ref $type$6))
  (local $1 (ref $type$36))
  (try
   (do
    (call_indirect $0 (type $type$40)
     (local.tee $1
      (ref.cast_static $type$36
       (local.get $0)
      )
     )
     (i32.add
      (struct.get $type$36 0
       (local.get $1)
      )
      (i32.const 37)
     )
    )
    (return
     (ref.as_non_null
      (struct.get $type$36 2
       (local.get $1)
      )
     )
    )
   )
   (catch $tag$0
    (drop
     (pop (ref $type$27) (ref $type$6))
    )
    (return
     (call $B.transform
      (ref.as_non_null
       (struct.get $type$36 2
        (local.get $1)
       )
      )
     )
    )
   )
  )
 )
```

kripken:
Binaryen will not change dominance in SimplifyLocals, however, the current spec's
notion of dominance is simpler than ours, and we must not optimize certain cases in
order to still validate. See details in the comment and test.
 
Fixes #4702

cc @askeksa-google


tlively (approved):
LGTM! The explanatory comment was very helpful.

kripken:
@askeksa-google Hmm, not sure, I'll look into that testcase. At first glance, I don't think nesting is the issue. It could be a problem in another pass entirely.

In general I am a little less optimistic about option 1c given this type of issue... since binaryen has more precise dominance info, we'll need to modify various passes, almost certainly. Less than for options 1a and 1b, but it's not clear to me how much work will be needed here.

kripken:
Oh, wait, it is the nesting, good catch... the nesting requires us to look at not just the local index being moved but its contents.

kripken:
Landing this as a first step while I figure out how to fix the second issue.

PR #4704, opened 2022-05-31
[NFC] Refactor EHUtils::findPops() method

kripken:
This moves it out of the validator so it can be used elsewhere. It will be
used in https://github.com/WebAssembly/binaryen/pull/4685#discussion_r883983340

### DIFF
--- a/src/ir/eh-utils.cpp
+++ b/src/ir/eh-utils.cpp
@@ -162,6 +162,12 @@ void handleBlockNestedPops(Function* func, Module& wasm) {
   TypeUpdating::handleNonDefaultableLocals(func, wasm);
 }
 
+Pop* findPop(Expression* expr) {
+  auto pops = findPops(expr);
+  assert(pops.size() == 1);
+  return pops[0];
+}
+
 SmallVector<Pop*, 1> findPops(Expression* expr) {
   SmallVector<Pop*, 1> pops;
   SmallVector<Expression*, 8> work;
--- a/src/ir/eh-utils.cpp
+++ b/src/ir/eh-utils.cpp
@@ -162,6 +162,28 @@ void handleBlockNestedPops(Function* func, Module& wasm) {
   TypeUpdating::handleNonDefaultableLocals(func, wasm);
 }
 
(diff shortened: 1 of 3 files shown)


kripken:
Ah, very good point...

Ok, refactored to have both `findPop` which assumes a valid module, and `findPops` which does not.

src/ir/eh-utils.h
```diff
@@ -40,6 +41,22 @@ void handleBlockNestedPop(Try* try_, Function* func, Module& wasm);
...
+//
+//  (catch
+//    (pop) ;; we want this for the outer catch
+//    (try
+//      (catch
+//        (pop) ;; but we do not want this for the outer catch
+Pop* findPop(Expression* expr);
+
+// Like findPop(), but it does *not* assume that the module validates. A catch
+// might therefore have 0 or more than 1 pop. This is primarily useful in the
```
aheejin, line 56: Having 0 pop can be valid when a tag does not have params: https://github.com/WebAssembly/binaryen/blob/6257a5857a42fec6b86352b03c1e92dcb7a490ea/src/wasm/wasm-validator.cpp#L2199-L2204

This comment sounds like having 0 pop is invalid. Sorry I should've been more clear about this in my earlier comment.

src/ir/eh-utils.cpp
```diff
@@ -162,6 +162,34 @@ void handleBlockNestedPops(Function* func, Module& wasm) {
   TypeUpdating::handleNonDefaultableLocals(func, wasm);
 }
 
+Pop* findPop(Expression* expr) {
+  auto pops = findPops(expr);
+  assert(pops.size() == 1);
```
aheejin, line 167: Sorry I should've been more clear about this. There can be no pop when a tag doesn't have any params. So we should allow `pops.size()` to either 0 or 1, and should return nullptr in case of 0.

aheejin (approved):
LGTM % comments

src/ir/eh-utils.cpp
kripken: Ah, thanks... I should have known that about exception handling.

PR #4705, opened 2022-05-31
[WasmGC] Fix try-catch and non-nullable local sets in values in SimplifyLocals

### LINKED ISSUE #4702: Moving local initialization into `try` block breaks flow-sensitive nn local validation

kripken:
Followup to #4703, this also handles the case where there is a non-
nullable local.set in the value of a nullable one, which we also cannot
optimize.

Fixes #4702 

cc @askeksa-google

### DIFF
--- a/src/passes/SimplifyLocals.cpp
+++ b/src/passes/SimplifyLocals.cpp
@@ -306,8 +306,12 @@ struct SimplifyLocals
       for (auto& [index, info] : self->sinkables) {
         // Expressions that may throw cannot be moved into a try (which might
         // catch them, unlike before the move).
-        //
-        // Also, non-nullable local.sets cannot be moved into a try, as that may
+        if (info.effects.throws()) {
+          invalidated.push_back(index);
+          continue;
+        }
+
+        // Non-nullable local.sets cannot be moved into a try, as that may
         // change dominance from the perspective of the spec
         //
         //  (local.set $x X)
@@ -338,9 +342,15 @@ struct SimplifyLocals
         // dominance). To stay compliant with the spec, however, we must not
         // move code regardless of whether ".." can throw - we must simply keep
         // the set outside of the try.
-        if (info.effects.throws() ||
-            self->getFunction()->getLocalType(index).isNonNullable()) {
-          invalidated.push_back(index);
+        //
+        // The problem described can also occur on the *value* and not the set
+        // itself. For example, |X| above could be a local.set of a non-nullable
+        // local. For that reason we must scan it all.
+        for (auto* set : FindAll<LocalSet>(*info.item).list) {
+          if (self->getFunction()->getLocalType(set->index).isNonNullable()) {
+            invalidated.push_back(index);
+            break;
+          }
         }
       }
       for (auto index : invalidated) {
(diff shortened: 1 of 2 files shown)


tlively (approved):
LGTM, but is this inner `FindAll` going to cause unacceptable quadratic complexity?

kripken:
In the worst case it is `O(|# Try| |# sets|)`. So potentially quadratic, but the number of Try instructions tends to be reasonable...

kripken:
I added a check on the gc-nn-locals feature, so this time won't be spent without that enabled, at least.

tlively:
Not just the number of sets, but the total number of subexpressions, since `FindAll` has to traverse the full tree to find the sets. The check for gc-nn-locals sounds good for now, though.

kripken:
Ah yes, it's the number of tries times the number of sets (which is the maximum number of sinkable items, that we loop on) times the number of subexpressions in the traversal. So the worst worst case is cubic...

PR #4706, opened 2022-05-31
Updating spec test readme

ashleynh:
The old readme pointed users to run.py which is 1) not executing the correct command for wasm-shell and 2) incompatible with the current recommended workflow.


tlively (approved):
Nice!

PR #4708, opened 2022-06-01
Make WasmBinaryWriter writes and WasmBinaryBuilder reads more explicit

TheGreatRambler:
This PR makes reads and writes more explicit, replacing calls to `o << U32LEB(val);` with `putValue<ValueWritten::SomeType>(val);` and `getInt8();` with `getValue<ValueWritten::SomeType, uint8_t>();`. In addition to making reads and writes more explicit, the code becomes more self documenting and easier to modify in the future. Reads and writes are also constrained to a small number of functions that a developer could implement on their own to change how binaryen imports or exports modules. The only functions responsible for reading and writing are now:

Reading:
```
getValue();
getInlineString();
getResizableLimits();
getByteView();
```
Writing:
```
putValue();
writeInlineString();
writeResizableLimits();
writeData();
```

The following code was used to verify this code acts identically to the old code using `wasm-smith` (the three disabled extensions crash the original implementation, a separate PR would be required to fix them):
```c++
auto GenerateSeededRandomString = [](int seed, int len) {
	static constexpr auto chars = "0123456789"
								  "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
								  "abcdefghijklmnopqrstuvwxyz";
	std::mt19937 rng(seed);
	auto dist   = std::uniform_int_distribution { {}, std::strlen(chars) - 1 };
	auto result = std::string(len, '\0');
	std::generate_n(begin(result), len, [&]() { return chars[dist(rng)]; });
	return result;
};

bool READ_COMPARE = std::filesystem::exists("moduleCompare.bin");
std::fstream moduleCompare("moduleCompare.bin",
	READ_COMPARE ? (std::ios::in | std::ios::binary) : (std::ios::out | std::ios::trunc | std::ios::binary));

for(int i = 0; i < 32767; i++) {
	std::string commandString
		= std::string("echo ") + GenerateSeededRandomString(i, 1000)
		  + std::string(
			  " | wasm-tools smith -o test.wasm --sign-extension-ops false --saturating-float-to-int false --multi-value false");
	system(commandString.c_str());

	try {
		wasm::Module wasm;
		std::ifstream testFile("test.wasm", std::ios::binary);
		std::vector<char> fileContents((std::istreambuf_iterator<char>(testFile)), std::istreambuf_iterator<char>());
		wasm::WasmBinaryBuilder parser(wasm, wasm.features, fileContents);
		parser.read();

		wasm::BufferWithRandomAccess buffer;
		wasm::WasmBinaryWriter writer(&wasm, buffer);
		writer.setEmitModuleName(false);
		writer.setNamesSection(false);
		writer.write();

		if(READ_COMPARE) {
			int expectedSize;
			moduleCompare.read((char*)&expectedSize, sizeof(expectedSize));
			std::vector<uint8_t> expectedOutput(expectedSize);
			moduleCompare.read((char*)expectedOutput.data(), expectedSize);
			if(buffer != expectedOutput) {
				std::cout << "Module produces incorrect output:" << std::endl;
				system("wasm2wat test.wasm");
			}
		} else {
			int size = buffer.size();
			moduleCompare.write((char*)&size, sizeof(size));
			moduleCompare.write((char*)buffer.data(), size);
		}
	} catch(std::exception e) {
		std::cout << "Exception: " << e.what() << std::endl;
		system("wasm2wat test.wasm");
	}

	if(i % 100 == 0) {
		std::cout << "Checked " << i << " modules" << std::endl;
	}
}

moduleCompare.close();
```


kripken:
Interesting! I think this looks like a worthwhile improvement, but curious what others think.

For `getValue/putValue`, perhaps `readValue/writeValue`? Not sure what's best here.

This should not affect speed, but please do some measurements to be sure.

Also, in addition to the `wasm-smith` verification, please run the binaryen fuzzer on this PR, using `./scripts/fuzz_opts.py` for a while.

TheGreatRambler:
Sure, I'm running some performance measurements right now. Surprised I missed `./scripts/fuzz_opt.py` but I'll run it as well

TheGreatRambler:
After running the following code I got an average time for reading + writing a module

Results:
Original: 490,040,604 byte binary (debug build), 454.617 microseconds
Explicit reading+writing: 490,110,086 byte binary (debug build), 457.464 microseconds

```c++
auto GenerateSeededRandomString = [](int seed, int len) {
	static constexpr auto chars = "0123456789"
								  "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
								  "abcdefghijklmnopqrstuvwxyz";
	std::mt19937 rng(seed);
	auto dist   = std::uniform_int_distribution { {}, std::strlen(chars) - 1 };
	auto result = std::string(len, '\0');
	std::generate_n(begin(result), len, [&]() { return chars[dist(rng)]; });
	return result;
};

bool READ_COMPARE = std::filesystem::exists("moduleCompare.bin");
std::fstream moduleCompare("moduleCompare.bin",
	READ_COMPARE ? (std::ios::in | std::ios::binary) : (std::ios::out | std::ios::trunc | std::ios::binary));

int64_t cumulativeTime = 0;

for(int i = 0; i < 32767; i++) {
	std::string commandString
		= std::string("echo ") + GenerateSeededRandomString(i, 1000)
		  + std::string(
			  " | wasm-tools smith -o test.wasm --sign-extension-ops false --saturating-float-to-int false --multi-value false");
	system(commandString.c_str());

	try {
		wasm::Module wasm;
		std::ifstream testFile("test.wasm", std::ios::binary);
		std::vector<char> fileContents((std::istreambuf_iterator<char>(testFile)), std::istreambuf_iterator<char>());

		auto start = std::chrono::high_resolution_clock::now();
		wasm::WasmBinaryBuilder parser(wasm, wasm.features, fileContents);
		parser.read();

		wasm::BufferWithRandomAccess buffer;
		wasm::WasmBinaryWriter writer(&wasm, buffer);
		writer.setEmitModuleName(false);
		writer.setNamesSection(false);
		writer.write();
		auto stop = std::chrono::high_resolution_clock::now();

		cumulativeTime += std::chrono::duration_cast<std::chrono::microseconds>(stop - start).count();

		if(READ_COMPARE) {
			int expectedSize;
			moduleCompare.read((char*)&expectedSize, sizeof(expectedSize));
			std::vector<uint8_t> expectedOutput(expectedSize);
			moduleCompare.read((char*)expectedOutput.data(), expectedSize);
			if(buffer != expectedOutput) {
				std::cout << "Module produces incorrect output:" << std::endl;
				system("wasm2wat test.wasm");
			}
		} else {
			int size = buffer.size();
			moduleCompare.write((char*)&size, sizeof(size));
			moduleCompare.write((char*)buffer.data(), size);
		}
	} catch(std::exception e) {
		std::cout << "Exception: " << e.what() << std::endl;
		system("wasm2wat test.wasm");
	}

	if(i % 100 == 0) {
		std::cout << "Checked " << i << " modules" << std::endl;
	}
}

std::cout << "Average is " << (double)cumulativeTime / 32767 << " microseconds" << std::endl;

moduleCompare.close();
```

kripken:
Sounds good about performance!

About wabt's `wasm2wat`, I suspect it's GC support is not up to date. We should probably disable wabt fuzzing when GC is enabled.

Overall I think this is worth doing. @aheejin @tlively what are your thoughts on this direction?

tlively:
Between the extra verbosity and the extra indirection making it more difficult to find out exactly what is being written, I'm not sure this change is worth it. @TheGreatRambler, can you elaborate on the benefits?

TheGreatRambler:
I wrote it to change how certain values are written, mostly ASTNodes, for a personal project to make highly compressed modules. I'd say it's a benefit that it's a lot clearer what exactly is being written and changes to the format (once again, if someone wanted to make changes to a fork for their own project) are propagated to all uses of the type.

It's true that it makes large sweeping changes to these 3 files that have stayed relatively the same since they were written, but I wanted to see if the changes were considered useful to the project

TheGreatRambler:
Some uses I could see:
* read or write modules from alternate sources like sockets
* Identify bloated parts to cut down on filesize
* Create alternate formats for research

tlively:
I do think the use case of modifying the way modules are read or emitted is interesting and this patch definitely makes that easier, but even with this patch, you would still need to modify the Binaryen source to actually do the customization. I don't think the benefits here are worth the extra indirection.

TheGreatRambler:
The idea is you create a class that extends WasmBinaryBuilder or WasmBinaryWriter and overrides the 4 functions listed in the first comment (frankly I should have made it 1 but the three I didn't include have sufficiently complex outputs that I thought them best as separate functions like they are now).

An example would be
```c++
class WasmBinaryBuilderNew : public wasm::WasmBinaryBuilder {
public:
	template <wasm::ValueWritten VT, typename T = void, typename U = int> T getValue(U inputValue = 0) {

	}

	wasm::Name getInlineString() {

	}

	void getResizableLimits(
		wasm::Address& initial, wasm::Address& max, bool& shared, wasm::Type& indexType, wasm::Address defaultIfNoMax) {

	}

	std::pair<const char*, const char*> getByteView(size_t size) {

	}
}
```

The big issue is how changes to the internals of Binaryen would break code depending on this feature. A solution involving callbacks even could work but that seemed a bit much for this PR

aheejin:
I'm not sure about this approach. I agree the new form may provide slightly better readability in some cases, but I'm not sure if it's a large enough improvement to offset the verbosity. For example,
```cpp
// Current
uint8_t sectionCode = getInt8();

// This PR
uint8_t sectionCode = getValue<ValueWritten::SectionStart, uint8_t>();
```

I'm not sure if the current code is harder to read, because it's clear from the variable name that we are reading a section code.

Also this PR couples values with types, which can be convenient if we don't need to remember type for each of the values. But it doesn't seem to be the case here; it looks this makes changing type type for a value harder. For example, the current `getInt8()` is fully generic, so if we want to change the section code to `uint8_t` to `uint16_t` (which is very unlikely, but as an example), in the current code we only need to change all callsites. But this PR requires changing all callsites, `getValue` function, and `writeValue` function.

I mostly skimmed the PR so I might be mistaken, in which case please let me know. 😅

aheejin:
> Tbh I'm new at templates, if I omit the second template argument will the return value be the type of the destination variable? If so that second template argument is not necessary

I'm not sure, because `getValue` wouldn't have a way of knowing what the return type would be right? But anyway, I think some duplication is inevitable in this design, because callsites (= users) of these functions have to know what the type is anyway, even if they use `auto`.

TheGreatRambler:
Because most callsites themselves use `auto` there's actually less duplication than you would expect

aheejin:
I mean, you can use `auto` in callsites to make it concise, but you still have to know the type to actually use the data. For example, after you get the value you should use the value for something right? Like, passing it to some function or writing it to memory or something.

TheGreatRambler:
Yeah but at that point you can modify the argument types of functions receiving the values from callsites if you want. In any case with `getValue` returning auto now, less types have to be modified than the original code.

aheejin:
By the way, unrelated to this particular PR but in general, it would be more convenient for review if you don't force push commits; that way it's really hard to figure out what has changed since the last review.

TheGreatRambler:
Ah thank you, my bad.

Also, I forgot about virtual functions! Definitely questioning the usefulness of this PR with all these small invasive changes.

PR #4712, opened 2022-06-05
[Parser][WIP] Start parser

tlively:


tlively:
@kripken, PTAL at the overall structure of the parser. No need to look at every parsing function right now, since most are incomplete. I would focus on understanding the parsing context types (`ParseDeclsCtx`, `ParseTypesCtx`, and `ParseDefsCtx`) and how they interact with the individual parsing functions. For a top-to-bottom slice of relatively complete functions, check out `parseModule`, `module`, `modulefield`, `global`, `globaltype`, `valtype`, `reftype`, `heaptype`, and `typeidx`.

The parser can parse this wat and produce the same output as the existing parser:

```
(module $parse
 (type $vv (func (param $x i32) (param i64 f32) (result f64) (result)))

 (global $g1 (export "g1") (export "g1.1") (import "mod" "g1") (ref null $vv))
 (global $g2 (import "mod" "g2") (mut (ref $vv)))
)
```

Notable missing features include the ability to parse instructions and proper error handling. The vast majority of places where the parser currently has `return {};` will be replaced with something like `return Err("msg");` eventually.

src/wasm/wat-parser.cpp
```diff
@@ -0,0 +1,1809 @@
...
+  Name name;
+  std::string_view span;
+};
+
+struct Limits {
+  uint32_t min;
+  std::optional<uint32_t> max;
+};
+
+struct DefinedTableType {
```
kripken, line 316: Slightly odd to have `DefSpan` but `DefinedTableType`, that is, to abbreviate Def sometimes but not others. Though I guess one is "definition" and one is "defined"? But, how about `TableInfo` here?

src/wasm/wat-parser.cpp
```diff
@@ -0,0 +1,1809 @@
...
+  Name nm;
+};
+
+// ===============
+// Parser Contexts
+// ===============
+
+using IndexMap = std::unordered_map<Name, Index>;
+
+// A parsing context used while parsing top-level declarations for a module.
```
kripken, line 347: Maybe comment that this is pass 1 out of 4 (or however many)?

src/wasm/wat-parser.cpp
```diff
@@ -0,0 +1,1809 @@
...
+  // Counters used for generating names for module elements.
+  int globalCounter = 0;
+
+  // Used to verify that all imports come before all non-imports.
+  bool hasNonImport = false;
+
+  ParseDeclsCtx(Module& wasm) : wasm(wasm) {}
+};
+
+// A parsing context used while parsing explicit type definitions.
```
kripken, line 394: (maybe mention this is phase 2?)

src/wasm/wat-parser.cpp
```diff
@@ -0,0 +1,1809 @@
...
+}
+
+// TODO
+template<typename Ctx> Result<typename Ctx::Instrs> instrs(Ctx&, ParseInput&) {
+  return {};
+}
+
+// expr ::= (in:instr)* => in* end
+template<typename Ctx> Result<typename Ctx::Instrs> expr(Ctx&, ParseInput&) {
+  // TODO
```
kripken, line 835: Is this where `scripts/gen-s-parser.py` integration will be?

src/wasm/wat-parser.cpp
```diff
@@ -0,0 +1,1809 @@
...
+    if (*x > ctx.wasm.functions.size()) {
+      return {};
+    }
+    return ctx.wasm.functions[*x]->name;
+  }
+  if (auto id = in.takeID()) {
+    if (auto func = ctx.wasm.getFunctionOrNull(*id)) {
+      return func->name;
+    }
+    return {};
```
kripken, line 881:
```suggestion
```

src/wasm/wat-parser.cpp
```diff
@@ -0,0 +1,1809 @@
...
+} // anonymous namespace
+
+Result<> parseModule(Module& wasm, std::string_view input) {
+  // Parse module-level declarations and implicit type definitions.
+  ParseDeclsCtx decls(wasm);
+  {
+    ParseInput in(input);
+    if (!module(decls, in)) {
+      return {};
+    }
```
kripken, line 1679: How will error handling work? For example here, shouldn't we somehow return the source location on which we failed to find a module? Or should the caller somehow handle that?

kripken (commented):
Nice! Direction and style lgtm.

src/wasm/wat-parser.cpp
tlively: Yeah, I would have preferred to call this just `TableType`, but then the compiler complains about the `using TableType = Ok;` below.

I'll solve this by changing the types in the `Ctx` types to be e.g. `using TableTypeT = Ok;`, `using TableTypeT = TableType`, etc.

src/wasm/wat-parser.cpp
tlively: Yes, we will use that generated code to generate function calls for each individual instruction given the keyword (i.e. instruction name) in this function or a child.

src/wasm/wat-parser.cpp
tlively: I'm in the process of updating `Result<T>` to be a variant that either holds a `T` or holds an `Err` type that contains a message and a location. Everywhere we currently have `return {};`, we will either propagate an error or generate a new one instead.

tlively: Just pushed an update with error handling--PTAL! I'm next going to introduce a `MaybeResult<T>` as a more ergonomic return value for all the parser functions that are currently returning either bool or `std::optional<T>` depending on the parsing stage.

src/wasm/wat-parser.cpp
```diff
@@ -250,6 +269,17 @@ struct ParseInput {
...
     return std::string_view(prev.pos, curr.pos - prev.pos);
   }
+
+  [[nodiscard]] Err err(std::string reason) {
+    std::stringstream msg;
+    if (auto t = peek()) {
+      msg << lexer.position(*t);
+    } else {
+      msg << lexer.position(lexer.next());
+    }
```
kripken, line 279: I'm having trouble understanding what this if-else does. `peek` returns a value if there is a value left, so in that case we return the position of that value, but what is the else doing? What is "next" if there is nothing to peek at?

Overall the general shape of error handling looks very nice!

kripken: I see, thanks!

tlively:
Closing this because I have started opening separate (and smaller) non-draft PRs for this.

PR #4714, opened 2022-06-07
GlobalStructInference: Handle >2 globals if values coincide

kripken:
In GSI we look for a read of a global in a situation like this:

```wat
$global1: value1
$global2: value2

(struct.get $Type (ref))
```
If global inference shows this get must be of either `$global1` or `$global2`, then we
can optimize to this:
```
(ref) == $global1 ? value1 : value2
```

We focus on the case of two values because 1 is handled by other passes, and >2
makes the tradeoffs less clear.

However, a simple extension is the case where there are more than 2 globals, but
there are only two values, and one value is unique to one global:

```wat
$global1: valueA
$global2: valueB
$global3: valueA

=>

(ref) == $global2 ? valueB : valueA
```
We can still use a single comparison here, on the global that has the
unique value. Then the else will handle all the other globals.

This increases the cases that GSI can optimize J2Wasm output by over 50%.

### DIFF
--- a/src/passes/GlobalStructInference.cpp
+++ b/src/passes/GlobalStructInference.cpp
@@ -263,10 +263,10 @@ struct GlobalStructInference : public Pass {
         // value. And we have already exited if we have more than 2, so that
         // only leaves 1 and 2. We are looking for the case of 2 here, since
         // other passes (ConstantFieldPropagation) can handle 1.
-        assert(values.size() >= 1 && values.size() <= 2);
-        if (values.size() != 2) {
+        if (values.size() == 1) {
           return;
         }
+        assert(values.size() == 2);
 
         // We have two values. Check that we can pick between them using a
         // single comparison. While doing so, ensure that the index we can check
--- a/src/passes/GlobalStructInference.cpp
+++ b/src/passes/GlobalStructInference.cpp
@@ -259,7 +259,8 @@ struct GlobalStructInference : public Pass {
           }
         }
 
-        // We have at globals (at least 2), and so must have at least one value.
+        // We have some globals (at least 2), and so must have at least one
+        // value.
         assert(values.size() >= 1);
         if (values.size() != 2) {
           // More than two values to pick from, so we'd need more than one
--- a/src/passes/GlobalStructInference.cpp
+++ b/src/passes/GlobalStructInference.cpp
@@ -269,13 +269,13 @@ struct GlobalStructInference : public Pass {
         }
 
         // We have two values. Check that we can pick between them using a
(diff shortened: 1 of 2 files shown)


src/passes/GlobalStructInference.cpp
```diff
@@ -187,46 +187,113 @@ struct GlobalStructInference : public Pass {
...
         }
 
+        // We have some globals (at least 2), and so must have at least one
+        // value.
+        assert(values.size() >= 1);
+        if (values.size() != 2) {
+          // More than two values to pick from, so we'd need more than one
+          // comparison. Give up.
+          return;
+        }
```
tlively, line 269: This is already handled up on line 249, isn't it? Maybe this is necessary to catch `values.size() == 1`, in which case the condition and comment should be updated. Does another pass handle the case where there are multiple globals but just one value? If not, maybe we should handle that here.

src/passes/GlobalStructInference.cpp
```diff
@@ -187,46 +187,113 @@ struct GlobalStructInference : public Pass {
...
+
+        // We have two values. Check that we can pick between them using a
+        // single comparison. Find the index that we can check on, which is the
+        // index that has a single global.
+        Index checkIndex;
+        if (globalsForValue[0].size() == 1) {
+          checkIndex = 0;
+        } else if (globalsForValue[1].size() == 1) {
+          checkIndex = 1;
+        } else {
```
tlively, line 279: Instead of setting an index, it might be simpler to `std::swap` the checkable global to be the first one.

tlively (commented):
Nice!

src/passes/GlobalStructInference.cpp
kripken: Updating the comments and conditions, good point.

Yes, CFP should handle the case of multiple globals and one value, I'll mention that.

src/passes/GlobalStructInference.cpp
```diff
@@ -187,44 +187,110 @@ struct GlobalStructInference : public Pass {
...
         }
 
+        // We have some globals (at least 2), and so must have at least one
+        // value. And we have already exited if we have more than 2, so that
+        // only leaves 1 and 2. We are looking for the case of 2 here, since
+        // other passes (ConstantFieldPropagation) can handle 1.
+        assert(values.size() >= 1 && values.size() <= 2);
+        if (values.size() != 2) {
+          return;
+        }
```
tlively, line 269:
```suggestion
        if (values.size() == 1) {
          return;
        }
        assert(values.size() == 2);
```

tlively (approved):

PR #4722, opened 2022-06-14
[NFC] Optimize non-equirecursive LUB calculations

### LINKED ISSUE #4719: Floating issue with UNREACHABLE executed report
Hi.

I am facing an unstable behaviour of self-built [main](https://github.com/WebAssembly/binaryen/commit/093bf06e1115e84c4300eca91197dc31336ee3e5) branch binaryen for MacOs x86-64.

Half of the runs on **same** wasm file I got successful result but on other runs its `UNREACHABLE executed at binaryen/src/wasm/wasm-type.cpp:598!` . Sometimes (rare case) it is just a dump wasm text on stdout. I printed out this kind and it seems it change its value every run. I could share my .wasm file if needed for future investigation.

Ran it with next line:
wasm-opt --enable-nontrapping-float-to-int --enable-typed-function-references --enable-gc --enable-reference-types --enable-exception-handling --hybrid -O3 --inline-functions-with-loops --traps-never-happen --fast-math ./x.wasm -o ./xx.wasm

Thank you.

tlively:
Equirecursive LUB calculations potentially require building new recursive heap
types that did not already exist in the system, so they have a complicated code
path that uses a TypeBuilder to construct a LUB from the ground up. In contrast,
nominal and isorecursive LUB calculations never introduce new heap types, so
computing their LUBs is much simpler. Previously we were using the same code
path with the TypeBuilder for all type systems out of convenience, but this
commit factors out the LUB calculations for nominal and isorecursive types into
a separate code path that does not use a TypeBuilder.

Not only should this make LUB calculations faster for GC workloads, it also
avoids a mysterious race condition during parallel LUB calculations with
isorecursive types that resulted in a temporary type escaping from one thread
and being used-after-free from another thread. It would be good to fix that bug
properly, but it is very difficult to investigate. Sweeping it under the rug
instead is the best trade off for now.

Fixes #4719.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -187,8 +187,6 @@ private:
   // used directly.
   std::optional<Type> lub(Type a, Type b);
   HeapType lub(HeapType a, HeapType b);
-  HeapType::BasicHeapType lub(HeapType::BasicHeapType a,
-                              HeapType::BasicHeapType b);
   std::optional<Tuple> lub(const Tuple& a, const Tuple& b);
   std::optional<Field> lub(const Field& a, const Field& b);
   std::optional<Signature> lub(const Signature& a, const Signature& b);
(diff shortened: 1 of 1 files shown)


tlively:
* main
  * **PR #4722** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4722" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4722?utm_source=stack-comment).

kripken (approved):

PR #4733, opened 2022-06-16
First class Data Segments

ashleynh:
This PR changes struct Memory::Segment to class DataSegment, putting data segments on the same level as other module elements (functions, tables, globals, memory, etc.). Like the other module-level elements, instances of the DataSegment class must now always have an internal name.

DataSegments are now accessed via a vector member variable on the Module. All of the helper methods that are used to print, parse, encode, and decode the module-level classes have been added.

Many tests were updated because of a change made to Printing.cpp, that outputs the DataSegments after defined memory.


ashleynh:
Hmm, there's an issue with test/extended-names.wast.fromBinary, as the names were dropped from the updated test file! I'll look into this.

Also, just a FYI, data segments are supposed to be output after the defined memory, but the output looks a little weird for our tests with only memory imports, as the data segments show up after global instead of after the memory.

ashleynh:
Fixed test/extended-names.wast.fromBinary. This PR is ready for review

src/wasm-traversal.h
```diff
@@ -204,12 +205,14 @@ struct Walker : public VisitorType {
...
-      }
+  void walkDataSegment(DataSegment* segment) {
+    if (!segment->isPassive) {
+      walk(segment->offset);
     }
+    static_cast<SubType*>(this)->visitDataSegment(segment);
+  }
+
+  void walkMemory(Memory* memory) {
     static_cast<SubType*>(this)->visitMemory(memory);
```
kripken, line 216: This matches `walkTable`, but reading the old code now, it looks wrong. `walk` should visit all children, not just `visit` the parent and stop. We should fix that eventually, I guess. For now maybe let's add a TODO so we don't forget,
```suggestion
    // TODO: This method and walkTable should walk children too, or be renamed.
    static_cast<SubType*>(this)->visitMemory(memory);
```

src/passes/Metrics.cpp
```diff
@@ -62,16 +61,19 @@ struct Metrics
...
 
+    // add memory
+    walkMemory(&module->memory);
+    for (auto& segment : module->dataSegments) {
+      walkDataSegment(segment.get());
+    }
     Index size = 0;
+    ModuleUtils::iterActiveDataSegments(
+      *module, [&](DataSegment* segment) { size += segment->data.size(); });
+    counts["[memory-data]"] = size;
```
kripken, line 73: Looks like the table code below does
```cpp
    if (!module->tables.empty()) {
      counts["[table-data]"] = size;
    }
```
Doing the same here makes sense I think. That would also reduce the size of the test diff I believe.

src/wasm-binary.h
```diff
@@ -1255,7 +1261,7 @@ class WasmBinaryWriter {
   void writeFunctions();
   void writeGlobals();
   void writeExports();
-  void writeDataCount();
+  void writeDataSegmentCount();
```
tlively, line 1264: The name of the section this writes is the "data count section," so I don't think the name needs to change.

src/wasm-binary.h
```diff
@@ -1590,7 +1599,7 @@ class WasmBinaryBuilder {
   bool hasDataCount = false;
 
   void readDataSegments();
-  void readDataCount();
+  void readDataSegmentCount();
```
tlively, line 1602: Same here.

src/ir/memory-utils.cpp
```diff
@@ -37,34 +37,34 @@ bool flatten(Module& wasm) {
...
     }
-    std::copy(segment.data.begin(), segment.data.end(), data.begin() + start);
+    std::copy(segment->data.begin(), segment->data.end(), data.begin() + start);
   }
-  memory.segments.resize(1);
-  memory.segments[0].offset->cast<Const>()->value = Literal(int32_t(0));
-  memory.segments[0].data.swap(data);
+  dataSegments.resize(1);
+  dataSegments[0]->offset->cast<Const>()->value = Literal(int32_t(0));
+  dataSegments[0]->data.swap(data);
```
tlively, line 67: Does this need to update `dataSegmentsMap` as well?

src/ir/memory-utils.h
```diff
@@ -54,25 +54,23 @@ inline bool ensureLimitedSegments(Module& module) {
...
   };
 
   Index numConstant = 0, numDynamic = 0;
   bool hasPassiveSegments = false;
-  for (auto& segment : memory.segments) {
-    if (!isEmpty(segment)) {
-      if (isConstantOffset(segment)) {
+  for (auto& segment : dataSegments) {
+    if (!isEmpty(segment.get())) {
+      if (isConstantOffset(segment.get())) {
```
tlively, line 67: We could use `*segment` instead of `segment.get()` here as well.

src/ir/module-utils.h
```diff
@@ -106,6 +106,24 @@ inline Table* copyTable(const Table* table, Module& out) {
...
+    ret->name = segment->name;
+    ret->hasExplicitName = segment->hasExplicitName;
+    ret->isPassive = segment->isPassive;
+    ret->data = segment->data;
+
+    return out.addDataSegment(std::move(ret));
+  };
+
+  if (segment->isPassive) {
+    return copy(std::make_unique<DataSegment>());
```
tlively, line 120: It would be slightly simpler to unconditionally create and fill out the new `DataSegment` then conditionally set its `offset`.

src/wasm/wasm-binary.cpp
```diff
@@ -913,8 +913,8 @@ void WasmBinaryWriter::writeNames() {
   // data segment names
   if (wasm->memory.exists) {
     Index count = 0;
-    for (auto& seg : wasm->memory.segments) {
-      if (seg.name.is()) {
+    for (auto& seg : wasm->dataSegments) {
+      if (seg->name.is() && seg->hasExplicitName) {
```
tlively, line 917: `seg->name.is()` should always be true.

tlively (commented):
LGTM besides some minor comments!

src/passes/Metrics.cpp
```diff
@@ -62,16 +61,19 @@ struct Metrics
...
-    }
 
+    // add memory
+    walkMemory(&module->memory);
+    for (auto& segment : module->dataSegments) {
+      walkDataSegment(segment.get());
+    }
     Index size = 0;
+    ModuleUtils::iterActiveDataSegments(
+      *module, [&](DataSegment* segment) { size += segment->data.size(); });
```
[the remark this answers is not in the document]
tlively: I guess we already have this pattern established for element segments. It might be nice to do a drive-by cleanup here and use just a single loop each for element segments and data segments.

src/passes/Metrics.cpp
ashleynh: Good call out! The metrics related tests were changed because memory-data was showing with a 0 and will revert back 👍🏽

kripken (approved):
I read the tests now too.

Lgtm (aside from the unnecessary new `[memory-data]` stuff).

src/ir/memory-utils.h
ashleynh: Thanks! Updated all of the .get() spots I added in this file.

ashleynh:
Please note the removal of `[memory-data] : 0` from `test/passes/O3_low-memory-unused_metrics.txt` created by introducing the condition `if (!module->dataSegments.empty()) {
      counts["[memory-data]"] = size;
    }` to `src/passes/Metrics.cpp`

src/passes/Metrics.cpp
```diff
@@ -62,23 +61,26 @@ struct Metrics
...
-      }
+
+    // add memory
+    walkMemory(&module->memory);
+    Index size = 0;
+    for (auto& segment : module->dataSegments) {
+      walkDataSegment(segment.get());
+      size += segment->data.size();
+    }
+    if (!module->dataSegments.empty()) {
```
kripken, line 72:
```suggestion
    if (module->memory.exists) {
```
I think this might remove the diff on the metrics test?

kripken: (And after adding multiple memory support, the line could be
```cpp
    if (!module->memories.empty()) {
```
which would be symmetric to the table code below.)

ashleynh: Sweet, thanks for the idea. I changed the if-check per your suggestion but the file `test/passes/O3_low-memory-unused_metrics.txt` didn't revert back.

kripken: Ok, I guess I'm missing something here... let's leave this as-is then (with multi-memory it'll change anyhow).

tlively (approved):
LGTM besides those two last comments!

src/ir/module-utils.h
```diff
@@ -106,6 +106,24 @@ inline Table* copyTable(const Table* table, Module& out) {
...
+
+    return out.addDataSegment(std::move(ret));
+  };
+
+  auto curr = Builder::makeDataSegment();
+  if (!segment->isPassive) {
+    auto offset = ExpressionManipulator::copy(segment->offset, out);
+    curr->offset = offset;
+  }
+  return copy(std::move(curr));
```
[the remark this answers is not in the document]
ashleynh: Totally agree, thanks!

src/passes/MemoryPacking.cpp
```diff
@@ -151,17 +153,18 @@ void MemoryPacking::run(PassRunner* runner, Module* module) {
 
     std::vector<Range> ranges;
 
-    if (canSplit(segment, currReferrers)) {
-      calculateRanges(segment, currReferrers, ranges);
+    if (canSplit(segment.get(), currReferrers)) {
+      calculateRanges(segment.get(), currReferrers, ranges);
```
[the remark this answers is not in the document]
ashleynh: Because canSplit and calculateRanges takes a const DataSegment* segment, the get() call is needed. We'd need to remove the const.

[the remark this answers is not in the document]
tlively: I was thinking that we could change those function to take `const std::unique_ptr<DataSegment>&`.

PR #4745, opened 2022-06-22
[Parser] Parse struct and array types

tlively:
Parse struct and array type definitions along with field names. Only the most
basic definitions are parsed for now; subtype definitions (both nominal
prototype and standard formats) and recursion groups are left to follow-on PRs.

Since there is no official standard for the text format for GC type definitions,
attempt to define a grammar that allows abbreviations that we already use
widely, such as making `(field ... )` optional except for named fields.

### DIFF
--- a/src/wasm/wat-parser.cpp
+++ b/src/wasm/wat-parser.cpp
@@ -301,6 +301,10 @@ struct ParseDeclsCtx {
   using ParamsT = Ok;
   using ResultsT = Ok;
   using SignatureT = Ok;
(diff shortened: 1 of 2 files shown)


tlively:
* main
  * **PR #4745** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4745" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4745?utm_source=stack-comment).

PR #4748, opened 2022-06-23
[Wasm GC] OptimizeInstructions: remove casts leading to comparisons

kripken:
Comparing references does not depend on the cast, so if we are ignoring
traps we can remove them.

Noticed by @jakobkummerow


tlively (approved):
Nice!

kripken:
Hmm, I think I realized a bug. This has to at least keep the type an `eqref`, that is, this would be bad:
```
(ref.eq
  (ref.as_data
    (ref.func $foo) ;; not eqref! without the ref.as we won't validate
```
Of course this will trap at runtime, but we shouldn't break validation. I'll look into it...

kripken:
Fixed. However, I don't think this could have been a problem in practice as other opts will modify such trapping expressions anyhow...

tlively (approved):

kripken:
Fuzzing found no issues.

PR #4749, opened 2022-06-23
[Wasm GC] OptimizeInstructions: Optimize ref.eq on equal inputs with a tee

kripken:
```wat
(ref.eq
  (local.tee $x (..))
  (local.get $x)
)
```
That will definitely return 1. Before this PR the side effects of `tee` stopped us
from optimizing.

This "steals" `drop.h` from #4598, which it turns out would be useful here too.

Noticed by @jakobkummerow


src/ir/drop.h
```diff
@@ -0,0 +1,66 @@
...
+
+namespace wasm {
+
+// Given an expression, returns a new expression that drops the given
+// expression's children that cannot be removed outright due to their side
+// effects. This is useful if we know the node is not needed but may need to
+// keep the children around; this utility will automatically remove any children
+// we do not actually need to keep, based on their effects.
+//
+// The caller can also pass in an optional last item to add to the output.
```
tlively, line 33: I think it would make sense to keep this utility more focused by removing this optional last parameter and making the caller responsible for adding whatever additional instructions it wants to the result.

kripken: Fair point, but it would have downsides that way that I don't see a nice solution for. The caller would need to check if the output is a block or not, in particular, which adds complexity there. The caller could call a helper for that, I see we have the `blockify()` method which can append, but a downside is that then we'd be finalizing the block there. Whereas the current code can finalize exactly once in this method.

tlively: I would prefer to use `blockify` in the caller, but it's not a strong preference. LGTM either way.

kripken: I prefer the current way, but I would like to make sure we're talking about the same thing here. The current caller code looks like `foo(a, b, c)` and with the change you propose I think it would become `bar(foo(a, b), c)`. That adds complexity on each caller of the API, so I'd rather have a little more complexity in single place that `foo` is implemented. And it's consistent with other APIs we have (like `blockify` itself btw - it's last parameter is an optional thing to append).

tlively: Yeah, that's what I had in mind, where `bar` is `blockify` :)

(FWIW, the primary reason the second argument to the non-template blockify is optional is so that the base case of the variadic blockify works and the core function of blockify is to insert things into blocks, so I don't think it's very comparable to getDroppedChildren. Do we have other APIs that do this?)

kripken: `blockifyWithName` also has an `append` optional last argument, and there is no variadic function there, so maybe it's more clear there that it really is a normal optional item to be appended. Though, none of the callers in the codebase use that argument :smile: 

Would you like this more if it was `getDroppedChildrenAndAppend`, that is, if appending was an explicit part of the API?

tlively: `getDroppedChildrenAndAppend` would be great if the argument were non-optional. It looks like the only caller does do the append, so it looks like that wouldn't complicate things too much.

PR #4750, opened 2022-06-24
fix: Avoid trying to read offset if segment is passive

phated:
When trying to implement some tests in Binaryen.ml, @ospencer encountered an issue where `getMemorySegmentInfoByIndex` throws if the segment is passive. This happens because the helper function was trying to *always* read the offset, even if it isn't a valid operation.

This just re-orders the calls and defaults `offset` to null for passive segments.

I want to make a follow-up PR that allows Binaryen.ml to use these utilities individually, but this is the quickest fix.

### DIFF
--- a/src/js/binaryen.js-post.js
+++ b/src/js/binaryen.js-post.js
@@ -2519,8 +2519,13 @@ function wrapModule(module, self = {}) {
     return Module['_BinaryenGetNumMemorySegments'](module);
   };
   self['getMemorySegmentInfoByIndex'] = function(id) {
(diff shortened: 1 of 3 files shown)


kripken (commented):
Code looks good, thanks. Please add a test, maybe around the related code in `test/binaryen.js/kitchen-sink.js`?

phated:
> Please add a test, maybe around the related code in test/binaryen.js/kitchen-sink.js?

Should be good!

kripken (approved):

PR #4751, opened 2022-06-24
feat: Extract memorySegment functions into individual functions

phated:
This is a draft that builds upon #4750 

Binaryen.ml provides separate functions for each of these utilities, but using the `getMemorySegmentInfoByIndex` helper function does a lot of extra work on each function call. By providing these under a namespace, the individual functions can be provided and then used in the "info" helper.

I'd like to do this to the other "info" helpers in the JS wrapper, but wanted to gather feedback first.


dcodeIO:
Reminds me of the existing classes for `Expression`, `Block`, `Function` etc., which provide both namespaced static functions like `Function.getName` and convenient wrapper classes with getters for `Function#name` etc. through `deriveWrapperInstanceMembers`. Not all of the elements have a class yet, but perhaps here there could be a `MemorySegment` class (perhaps on top of a `MemorySegmentRef`) utilizing the existing concept?

```
MemorySegment
 .getOffset
 .setOffset
 .getData
 .setData
 .isPassive
 .setPassive
 #offset [get/set]
 #data [get/set]
 #passive [get/set]
```

Here's what's done for functions, for example:

https://github.com/WebAssembly/binaryen/blob/98016931d9de23c0b6a4b5b0df5d4c3954da2f39/src/js/binaryen.js-post.js#L4766-L4818

phated:
Yeah, I was wondering if doing something similar to Expressions was desirable. I'm fine implementing like that, but it'll add more work, which I'll have to find some time to do.

Is there a way to move forward with the straightforward "instance logic" that I did or would the PR only be accepted if the static/class implementation were completed?

tlively:
cc @ashleynh, who has worked on the internal representation of data segments in support of multiple memory. @dcodeIO's idea to reuse that class utility sounds like a good goal to me, but intermediate progress might be useful to land as well.

src/js/binaryen.js-post.js
```diff
@@ -709,6 +709,23 @@ function wrapModule(module, self = {}) {
...
+    'initial'() {
+      return Module['_BinaryenMemoryGetInitial'](module);
+    },
+    'shared'() {
+      return Boolean(Module['_BinaryenMemoryIsShared'](module));
+    },
+    'max'() {
+      if (Module['_BinaryenMemoryHasMax'](module)) {
+        return Module['_BinaryenMemoryGetMax'](module);
+      }
```
phated, line 728: Wasn't sure if this should have the `if` condition inside the function.

phated:
Thanks for the guidance folks! I tried to replicate patterns I had seen elsewhere in `binaryen.js-post.js`. Please take a look and let me know if I should be doing anything differently.

I've updated the `getSomethingInfo` functions that we want to remove in Grain but there are still more. Ideally, this PR can guide others in updating the rest (and of course I'll be updating things as we use them in Grain).

src/js/binaryen.js-post.js
kripken: I think if there is no maximum then we should probably either error or return `Infinity`. Doing it here in JS makes the most sense to me (we couldn't return `Infinity` from C).

src/js/binaryen.js-post.js
```diff
@@ -709,6 +709,23 @@ function wrapModule(module, self = {}) {
       'wait64'(ptr, expected, timeout) {
         return Module['_BinaryenAtomicWait'](module, ptr, expected, timeout, Module['i64']);
       }
+    },
+    'module'() {
+      return UTF8ToString(Module['_BinaryenMemoryImportGetModule'](module));
+    },
+    'base'() {
+      return UTF8ToString(Module['_BinaryenMemoryImportGetBase'](module));
+    },
```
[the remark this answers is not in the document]
kripken: I think it's ok as it is.

src/js/binaryen.js-post.js
```diff
@@ -4768,6 +4796,146 @@ Module['I31Get'] = makeExpressionWrapper({
   }
 });
 
+// ElementSegment wrapper
+Module['ElementSegment'] = (() => {
+  // Closure compiler doesn't allow multiple `Function`s at top-level, so:
+  function Function(func) {
```
kripken, line 4802: Should all these `Function` be `ElementSegment`?

phated: Probably. I wasn't sure if these constructors should match the key or if Function was meant to be generic (due to the comment). I'll change them.

phated:
@kripken thanks for the feedback! I've made the updates you suggested

src/js/binaryen.js-post.js
```diff
@@ -2505,36 +2524,48 @@ function wrapModule(module, self = {}) {
...
-      'shared': Boolean(Module['_BinaryenMemoryIsShared'](module))
+      'module': self['memory']['module'](),
+      'base': self['memory']['base'](),
+      'initial': self['memory']['initial'](),
+      'shared': self['memory']['shared']()
     };
-    if (Module['_BinaryenMemoryHasMax'](module)) {
-      memoryInfo['max'] = Module['_BinaryenMemoryGetMax'](module);
+    var max = self['memory']['max']();
+    if (max) {
```
kripken, line 2533: What is this checking? If the max is 0?

I'd expect this to check if the max is Infinity, but maybe I'm missing something...

phated: @kripken I'm not sure, I copied this from the existing code. Who wrote that?

src/js/binaryen.js-post.js
```diff
@@ -709,6 +709,25 @@ function wrapModule(module, self = {}) {
...
+    },
+    'shared'() {
+      return Boolean(Module['_BinaryenMemoryIsShared'](module));
+    },
+    'max'() {
+      if (Module['_BinaryenMemoryHasMax'](module)) {
+        return Module['_BinaryenMemoryGetMax'](module);
+      } else {
+        return Infinity;
+      }
```
dcodeIO, line 730: Is it intended that these are mixed with the instructions like `memory.fill`? Seems that it's quite easy to mistake them for instructions like `memory.shared`. Also, in light of multi memory, there'll likely need to be a memory wrapper anyhow to deal with any memory, so I wouldn't expect these to last for very long.

phated: Yes, see where Alon said "I think it's ok as it is." above.

I'm just making these easier to use, if this hasn't been updated for multiple memory, than someone else will need to help make those changes.

dcodeIO: Hmm, has the following variant been considered?

```
Module#getMemoryModule
Module#getMemoryBase
Module#getMemoryInitial
Module#getMemoryShared
Module#getMemoryMax
```

Those would about match `Module#setMemory` for now, which is similarly subject to change, with all of them not conflicting with instructions.

phated: I've not considered anything (maybe others have) because I'm just converting what is here from "expensive grab-bag" to "individual functions". Further improvements can be done later, as @tlively mentioned in the primary thread.

src/js/binaryen.js-post.js
phated: @kripken how about this?

```suggestion
    if (max !== Infinity) {
```

src/js/binaryen.js-post.js
dcodeIO: Perhaps this is a misunderstanding. I am not suggesting any significant changes in the above variant, just to move / rename these from `memory` (which so far covers exclusively instructions) to the `Module` instance (where methods of this kind without wrappers would typically go). Basically next to `setMemory`.

phated: I'm going to leave that up to @kripken since he said it was fine to have these here when I asked.

src/js/binaryen.js-post.js
```diff
@@ -4768,8 +4798,145 @@ Module['I31Get'] = makeExpressionWrapper({
...
+  deriveWrapperInstanceMembers(Export.prototype, Export);
+  Export.prototype['valueOf'] = function() {
+    return this[thisPtr];
+  };
+  return Export;
+})();
 
+// GlobalImport wrapper
+Module['GlobalImport'] = (() => {
+  // Closure compiler doesn't allow multiple `GlobalImport`s at top-level, so:
```
dcodeIO, line 4863: The IIFE here is not necessary. Only affects JS built-ins like `Function`.

src/js/binaryen.js-post.js
```diff
@@ -4768,8 +4798,145 @@ Module['I31Get'] = makeExpressionWrapper({
...
+  deriveWrapperInstanceMembers(GlobalImport.prototype, GlobalImport);
+  GlobalImport.prototype['valueOf'] = function() {
+    return this[thisPtr];
+  };
+  return GlobalImport;
+})();
+
+// Global wrapper
+Module['Global'] = (() => {
+  // Closure compiler doesn't allow multiple `Global`s at top-level, so:
```
dcodeIO, line 4887: Similarly, the IIFE here is not necessary. Only affects `Function`.

src/js/binaryen.js-post.js
```diff
@@ -4768,8 +4798,145 @@ Module['I31Get'] = makeExpressionWrapper({
...
+  deriveWrapperInstanceMembers(Global.prototype, Global);
+  Global.prototype['valueOf'] = function() {
+    return this[thisPtr];
+  };
+  return Global;
+})();
+
+// FunctionImport wrapper
+Module['FunctionImport'] = (() => {
+  // Closure compiler doesn't allow multiple `FunctionImport`s at top-level, so:
```
dcodeIO, line 4917: As well: The IIFE here is not necessary. Only affects JS built-ins like `Function`.

src/js/binaryen.js-post.js
```diff
@@ -4768,8 +4798,145 @@ Module['I31Get'] = makeExpressionWrapper({
...
+  deriveWrapperInstanceMembers(ElementSegment.prototype, ElementSegment);
+  ElementSegment.prototype['valueOf'] = function() {
+    return this[thisPtr];
+  };
+  return ElementSegment;
+})();
+
+// Export wrapper
+Module['Export'] = (() => {
+  // Closure compiler doesn't allow multiple `Export`s at top-level, so:
```
dcodeIO, line 4836: Same

src/js/binaryen.js-post.js
```diff
@@ -4768,8 +4798,145 @@ Module['I31Get'] = makeExpressionWrapper({
   }
 });
 
-// Function wrapper
+// ElementSegment wrapper
+Module['ElementSegment'] = (() => {
+  // Closure compiler doesn't allow multiple `ElementSegment`s at top-level, so:
```
dcodeIO, line 4803: Same :)

src/js/binaryen.js-post.js
```diff
@@ -2505,36 +2524,48 @@ function wrapModule(module, self = {}) {
...
+      Module['_BinaryenCopyMemorySegmentData'](module, id, ptr);
+      const res = new Uint8Array(size);
+      res.set(new Uint8Array(buffer, ptr, size));
+      _free(ptr);
+      return res.buffer;
+    },
+    'passive'(id) {
+      return Boolean(Module['_BinaryenGetMemorySegmentPassive'](module, id));
+    }
+  }
```
dcodeIO, line 2558: Another suggestion to make this more uniform with other code. Move / rename these to

```
Module#getMemorySegmentByteOffset
Module#getMemorySegmentData
Module#isMemorySegmentPassive
```

Then becomes

```js
myModule.getMemorySegmentByteOffset(segmentIndex)
```

instead of

```js
myModule.memorySegment.offset(id)
```

Apart from adhering to surrounding code style, this also has the advantage that these won't conflict with eventual setters.

src/js/binaryen.js-post.js
dcodeIO: FWIW, I guess the NFC value here would be `undefined`. Apart from this, if my other suggestions are agreed upon to be considered, there will be a `HasMax` function around that can be used instead, closely matching the previous code.

src/js/binaryen.js-post.js
kripken: Hmm, good point @dcodeIO , you're right that this would mix instructions like `memory.fill()` with helpers. We've put the helpers on the relevant JS objects using `makeExpressionWrapper`. So I agree that adding new `getX`/`setX` functions alongside those is more consistent. That seems best.

@phated Sorry for not reading this more in detail the first time! (I am not as up to date on the JS API as I used to be, since I've been busy with other things...)

phated: Can someone else pick up these changes? We really need these changes in Grain and I don't have time to continue to make these large changes, as I'm actively needing to interview for jobs.

PR #4754, opened 2022-06-27
[NFC] Refactor and clarify conditions for removing casts

kripken:
This just moves code around and adds comments.

I ended up writing this because in some future cast removal opts the
choices are less obvious, so I wanted to document the logic behind such
things. Hopefully it sounds like the right thing to do?

### DIFF
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -1430,7 +1430,7 @@ struct OptimizeInstructions
   //    having casts in the right places we can help GUFA optimize:
   //
   //      (if
-  //        (ref == 1)
+  //        (tag == 1)
   //        (struct.get $A 0
   //          (ref.cast $B ..))
   //


src/passes/OptimizeInstructions.cpp
```diff
@@ -1447,6 +1473,40 @@ struct OptimizeInstructions
...
   }
 
+  void visitRefEq(RefEq* curr) {
+    // Equality does not depend on the type, so casts may be removable.
+    //
+    // This is safe to do first because nothing else here cares about the type,
+    // and we consume the two input references, so removing a cast could not
+    // help our parents (see "notes on removing casts").
+    skipCast(curr->left, Type::eqref);
+    skipCast(curr->right, Type::eqref);
```
aheejin, line 1483: Do we need the condition of subtypes of `eqref`? If either of them is not a subtype of `eqref`, does this even validate?

src/passes/OptimizeInstructions.cpp
```diff
@@ -1426,6 +1450,8 @@ struct OptimizeInstructions
   // subtype of it. We will not remove a cast that would leave something that
   // would break that. If |requiredType| is not provided we will accept any type
   // there.
+  //
+  // See "notes on removing casts", above, for when this is safe to do.
   void skipCast(Expression*& input, Type requiredType = Type::anyref) {
```
aheejin, line 1455: Do we need the `requiredType=anyref`? If the input is not a reference, it's not even valid, no?

src/passes/OptimizeInstructions.cpp
```diff
@@ -1880,17 +1940,8 @@ struct OptimizeInstructions
...
-        //
-        //  (ref.is_func
-        //    (ref.as_func
-        //      (local.get $anyref)))
-        //
-        // The local has no useful type info. The cast forces it to be a
-        // function, so we can replace the ref.is with 1. But if we removed the
-        // ref.as first then we'd not have the type info we need to optimize
-        // that way.
+        // cast may remove info that we could have used to optimize, see
```
aheejin, line 1943: Does this apply to `ref.is_null`? Becuase this is within `if (curr->op == RefIsNull)`. The example for this uses `ref.is_func`.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1372,39 +1372,63 @@ struct OptimizeInstructions
...
+  // definitely be 1. But if we removed the ref.as first (which we can do in
+  // traps-never-happen mode) then we'd not have the type info we need to
+  // optimize that way.
+  //
+  // To avoid such risks we should keep in mind the following:
+  //
+  //  * Before removing a cast we should use its type information in the best
+  //    way we can. Only after doing so should a cast be removed. In the exmaple
+  //    above, that means first seeing that the ref.is must return 1, and only
+  //    then possibly removing the ref.as.
```
aheejin, line 1394: What does this mean? If we see that `ref.is` 1, and the remove `ref.as`, it still removes the info so we cannot optimize the remaining `ref.is` to 1, no?

src/passes/OptimizeInstructions.cpp
```diff
@@ -1372,39 +1372,63 @@ struct OptimizeInstructions
...
+  //  * Do not remove a cast if removing it might remove useful information for
+  //    others. For example,
+  //
+  //      (ref.cast $A
+  //        (ref.as_non_null ..))
+  //
+  //    If we remove the inner cast then the outer cast becomes nullable. That
```
aheejin, line 1401: Doesn't `ref.cast $A`'s type depend solely on `$A` type? Does whether we have `ref.as_non_null` inside determine whether the outer cast `ref.cast` becomes nullable or not?

aheejin (commented):
Mostly questions:

src/passes/OptimizeInstructions.cpp
```diff
@@ -1372,39 +1372,63 @@ struct OptimizeInstructions
...
+  //    it tells GUFA that we are reading from a $B here, and not the supertype
+  //    $A. If $B may contain fewer values in field 0 than $A, then GUFA might
+  //    be able to optimize better with this cast. But in traps-never-happen
+  //    mode we can assume that only $B can arrive here, which means GUFA should
+  //    be able to infer that even without the cast. Thus, in traps-never-happen
+  //    mode we should optimize out such a cast (assuming the code still
+  //    validates, of course), but not otherwise. (This may not always be the
+  //    optimal decision, say if GUFA hits a limitation and does not manage to
+  //    make the inference that only $B can arrive, but as it is important to
+  //    remove such redundant casts, we have to draw a line somewhere.)
```
aheejin, line 1423: Has this implemented?

src/passes/OptimizeInstructions.cpp
kripken: Yes, but it's ok to remove the info after it's already been used.
```wat
(ref.is_func
  (ref.as_func
    (local.get $anyref)))
=> ;; optimize ref.is to 1 (and drop child)
(block
  (drop
    (ref.as_func
      (local.get $anyref)))
  (i32.const 1))
=> ;; remove ref.as and the drop (in vacuum)
(i32.const 1)
```
The opposite order is less good:
```wat
(ref.is_func
  (ref.as_func
    (local.get $anyref)))
=> ;; remove ref.as
(ref.is_func
  (local.get $anyref))
```
And we get stuck there.

src/passes/OptimizeInstructions.cpp
kripken: Yes, the heap type of `(ref.cast $A ..)` is `$A`, correct, while it can be nullable or non-nullable (which it copies from the child - if the child is nullable, it is too).

src/passes/OptimizeInstructions.cpp
kripken: In GUFA? Yes, to some extent. It may see that `$B` is the only thing that reaches a location, say if it comes from a `struct.new $B`. But it will hit limitations if more than one type is possible, etc., at least for now.

src/passes/OptimizeInstructions.cpp
kripken: Yes, it wouldn't be valid. So `anyref` basically has no effect. That's what we want here - the default value means "no requirement on the type / accept any type there".

src/passes/OptimizeInstructions.cpp
kripken: The tricky situation is something like this:
```wat
(ref.eq
  (ref.cast $A
    (ref.func $func)
  ..
)
```
That will trap at runtime. But at compile time we can't remove the cast, as it would fail to validate - but it does validate before the removal.

src/passes/OptimizeInstructions.cpp
kripken: Correct, this isn't really an issue for `ref.is_null`. We have limited options there anyhow, and the order could be either way. But for the other code path we must do `skipCast` at the end (new line 1945). We do it the same here for consistency.

Ideally this comment would be on new line 1945... but that's lower down. I could move it and refer to it if that's clearer?

src/passes/OptimizeInstructions.cpp
```diff
@@ -1372,39 +1372,63 @@ struct OptimizeInstructions
...
+  //    without the cast. Only if we saw that the parents of the outer cast
+  //    cannot benefit from non-nullability should we remove it.
+  //    Another example:
+  //
+  //      (struct.get $A 0
+  //        (ref.cast $B ..))
+  //
+  //    The cast only changes the type of the reference, which is consumed in
+  //    this expression and so we don't have more parents to consider. But it is
+  //    risky to remove this cast, since e.g. GUFA does benefit from such info:
```
tlively, line 1413:
```suggestion
  //    risky to remove this cast, since e.g. GUFA benefits from such info:
```
I initially misread this as "GUFA does not benfit" and this would clarify it.

src/passes/OptimizeInstructions.cpp
tlively: It seems like it would probably be worth introducing a notion of "late" OptimizeInstructions in which we aggressively remove as many casts as possible without worrying about preserving information for follow-on optimizations.

src/passes/OptimizeInstructions.cpp
aheejin: Yeah that can be easier to understand. But it would be good to keep "Do this right before returning because removing a cast may remove info that we could have used to optimize" part then.

aheejin (approved):
Thanks for the explanation!

src/passes/OptimizeInstructions.cpp
kripken: Sounds good, changing.

src/passes/OptimizeInstructions.cpp
kripken: Yes, a "late" pass may be useful. The risk is the module becomes less optimizable later. But I can't see a way around that...

kripken:
Updated with feedback. Also I made the text more careful, since I realize we should probably *not* remove casts like `(struct.get (cast))`.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1410,17 +1410,20 @@ struct OptimizeInstructions
...
-  //    make the inference that only $B can arrive, but as it is important to
-  //    remove such redundant casts, we have to draw a line somewhere.)
+  //    be able to optimize better with this cast. Now, in traps-never-happen
+  //    mode we can assume that only $B can arrive here, which means GUFA might
+  //    be able to infer that even without the cast - but it might not, if we
+  //    hit a limitation of GUFA. Some code patterns simply cannot be expected
+  //    to be always inferred, say if a data structure has a tagged variant,
+  //    where a property is always in sync with whether a reference is say a
+  //    function - we can't expect GUFA to always detect such invariant
+  //    properties of programs. Given the large amount of potential benefit we
```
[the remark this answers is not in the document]
kripken: Sure, I added an example now with more details. Is that clearer?

src/passes/OptimizeInstructions.cpp
```diff
@@ -1417,13 +1417,29 @@ struct OptimizeInstructions
...
+  //        ref: anyref
+  //      }
+  //
+  //    Imagine that if tag == 0 then the reference always contains struct $A,
+  //    and if tag == 1 then it always contains a struct $B, and so forth. We
+  //    can't expect GUFA to figure out such invariants in general. But by
+  //    having casts in the right places we can help GUFA optimize:
+  //
+  //      (if
+  //        (ref == 1)
```
aheejin, line 1433:
```suggestion
  //        (tag == 1)
```

aheejin (commented):
Thanks! So tagged variants mean something like C++ `variant` but with an additional field that tells which type is stored? I haven't heard the term so I was confused.

kripken:
Sorry, yeah, I'm not sure where I've heard the term but I think other languages might use it. Or maybe I'm mixing up terminology...

Anyhow, yes, just a C++ variant + a tag field that says what is stored.

tlively:
(Just to clarify, C++ `variant` already has such a field internally. Tagged variant, variant type, discriminated union, enum with payload, and sum type all basically describe the same thing.)

PR #4755, opened 2022-06-27
[Strings] Add string proposal types

kripken:
This starts to implement the Wasm Strings proposal

https://github.com/WebAssembly/stringref/blob/main/proposals/stringref/Overview.md

This just adds the types.


src/wasm-binary.h
```diff
@@ -375,8 +375,12 @@ enum EncodedType {
   // run-time type info type, with depth index n
   rtt_n = -0x17, // 0x69
   // run-time type info type, without depth index n
-  rtt = -0x18,     // 0x68
-  dataref = -0x19, // 0x67
+  rtt = -0x18,              // 0x68
+  dataref = -0x19,          // 0x67
```
aheejin, line 379: Add some comments between `rtt` and `dataref`? This looks all these ref types belong to "run-time type info type, without depth index n" category.

src/wasm-type.h
```diff
@@ -109,8 +109,13 @@ class Type {
...
     anyref,
     eqref,
+    stringref,
+    stringview_wtf8,
+    stringview_wtf16,
+    stringview_iter,
+    // From here types are non-nullable.
     i31ref,
-    dataref,
+    dataref
```
kripken, line 118: Looks like this reordering breaks some tests in the C and JS APIs, where we care about the indexes here. So maybe this is not best. But it seems nicer to not have the nullable types split into sub-ranges... in particular it makes `isDefaultable` faster. Thoughts?

src/wasm-type.h
```diff
@@ -137,28 +142,32 @@ class Type {
...
+  // ├─ Aliases ────────╫───┼───┼───┼───┤───────┤
+  // │ funcref          ║ x │   │ x │ x │ f  n  │ ┐ Ref
+  // │ anyref           ║ x │   │ x │ x │ f? n  │ │  f_unc
+  // │ eqref            ║ x │   │ x │ x │    n  │ │  n_ullable
+  // │ i31ref           ║ x │   │ x │ x │       │ │
+  // │ dataref          ║ x │   │ x │ x │       │ │
+  // │ stringref        ║ x │   │ x │ x │    n  │ │
+  // │ stringview_wtf8  ║ x │   │ x │ x │    n  │ │
+  // │ stringview_wtf16 ║ x │   │ x │ x │    n  │ │
+  // │ stringview_iter  ║ x │   │ x │ x │    n  │ │
```
aheejin, line 165: So all `stringref` and `stringview`s are nullable? ([The spec](https://github.com/WebAssembly/stringref/blob/main/proposals/stringref/Overview.md) says `stringref` is, but no mention of `stringview`s.

src/wasm-type.h
```diff
@@ -137,28 +142,32 @@ class Type {
...
-  // │ dataref     ║ x │   │ x │ x │       │ │
-  // ├─ Compound ──╫───┼───┼───┼───┤───────┤ │
-  // │ Ref         ║   │ x │ x │ x │ f? n? │◄┘
-  // │ Tuple       ║   │ x │   │ x │       │
-  // │ Rtt         ║   │ x │ x │ x │       │
-  // └─────────────╨───┴───┴───┴───┴───────┘
+  //                      Compound Concrete
+  //   Type             Basic │ Single│
+  // ╒══════════════════╦═│═╤═│═╤═│═╤═│═╤═══════╕
+  // │ none             ║ x │   │   │   │       │
```
aheejin, line 148: Nit: I know it's preexisting, but what do you think about changing `x` to `o` in this table? We know what this means, but `x` can mean a negative, so..

src/wasm-type.h
aheejin: Do people rely on the hardcoded integer values? That seems bad in the first place. I think they should only get the value by calling functions like `BinaryenFeatureAtomics()` and use it..?

aheejin (approved):

src/wasm-type.h
tlively: Since `stringref` and the various views are proposed to be reftypes, it almost certainly makes sense to add the new types as `HeapType::BasicHeapType` instead. I would like to remove all the reference types from `Type::BasicType` at some point as a cleanup, since using them directly (instead of as a HeapType + nullability) is almost always incorrect.

kripken: @aheejin Yeah, I agree they should be doing that. So maybe it's fine to update the tests then.

@tlively I don't follow. This does add them to `HeapType::BasicHeapType`?

https://github.com/WebAssembly/binaryen/pull/4755/files#diff-afe568d2a9abd5777b21b5bd481afe1a51da75559098eb8feafe601be27c89c9R348-R360

Currently all ref types are added in both places.

I do like the idea to remove refs from `Type::BasicType`, but am not sure how that would work.

src/wasm-type.h
kripken: Yes, I can't find that in the spec either. I think nullable is more likely, but we can adjust later. The spec is still very early draft.

src/wasm-type.h
kripken: I'd be fine either way. Could also be a checkmark ✓

src/wasm-type.h
tlively: Ah right. In that case I think you can remove them from here and everything should just work? It doesn't look like you've implemented the part in wasm-type.cpp where you would canonicalize a nullable reference to the BasicHeapType to use the BasicType instead, and I think that's the only thing that you would have had to remove. Basically everywhere you produce or use a `Type::stringref` you would produce or use a `Type(HeapType::string, Nullable)` instead.

kripken: Thanks, I think I see now how we could remove all the types from here in general (though I'm not sure I see how it would work in all places, yet). For this PR, though, wouldn't it be oddly asymmetric to remove them from here? I'm just trying to match all other types in this PR.

(The original question in this thread was what to do with the constants. I buy @aheejin 's point that it's ok to just change the constants, as users shouldn't depend on them. Also it's just the constants for i31 and data, which is in the experimental GC proposal. I've updated the constants in the PR and the tests.)

tlively: Yes, there would be some assymmetry in keeping the BasicTypes here, but if you do that, you should also update `TypeInfo::getCanonical` in wasm-type.cpp so that `Type::stringref` and `Type(HeapType::string, Nullable)` are treated as the same type. Currently they would incorrectly be distinct types. Overall I think the patch would be much simpler if you just leave out the `BasicType`s, even though that would be (temporarily) inconsistent with the other types.

src/wasm-type.h
tlively: We are going to transition all the shorthands to be nullable, so this is a safe assumption.

src/wasm-type.h
tlively: The whole box is already made of non-alphanumeric chars, so checkmark sgtm ✔️

src/wasm-type.h
kripken: Ok, sounds good. I tried that, see the last commit (https://github.com/WebAssembly/binaryen/pull/4755/commits/bac57be4b45d75079e584b6ce6bec59ef058e79a). Definitely it makes the PR smaller, but it crashes in an infinite loop in `getCanonical / doInsert` which I'm not sure how to fix? Am I canonicalizing incorrectly there?

src/wasm/wasm-type.cpp
```diff
@@ -680,6 +684,14 @@ std::optional<Type> TypeInfo::getCanonical() const {
...
           case HeapType::data:
             break;
+          case HeapType::string:
+            return Type(HeapType::string, Nullable);
+          case HeapType::stringview_wtf8:
+            return Type(HeapType::stringview_wtf8, Nullable);
+          case HeapType::stringview_wtf16:
+            return Type(HeapType::stringview_wtf16, Nullable);
+          case HeapType::stringview_iter:
+            return Type(HeapType::stringview_iter, Nullable);
```
tlively, line 694: Just `break` here so that `getCanonical` ends up returning `{}` in these cases and you should be good to go.

kripken: Thanks! Right, so `{}` means it doesn't emit a canonical thing here, I guess, which is what we want... works for me locally, updated.

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ full changeset diff at the end of each section.
 
 Current Trunk
 -------------
+- Change constant values of some reference types in the C and JS APIs. This is
+  only observable if you hardcode specific values instead of calling the
+  relevant methods (like `BinaryenTypeDataref()`). (#4755)
```
tlively, line 19: Nit: I don't think this needs to be called out in the changelog since users shouldn't be using the raw values. For example, the numbers changed when we made `extern` an alias for `any`, but we didn't add that to the changelog.

tlively (approved):

CHANGELOG.md
kripken: Yeah, it's not strictly needed I guess. I imagine though maybe someone is using the raw values even though they shouldn't - maybe they don't know they shouldn't. Then if they see these docs maybe that gets them to fix it before running into any problems. Overall I think such "ABI" changes can't hurt in the changelog.

src/wasm-type.h
aheejin: So we add them only to `HeapType` and not `Type`? But we still have `dataref` or `funcref` as basic types. What's the criteria for the basic types?

tlively: I would like to remove `dataref` and `funcref` from `BasicType` at some point as well so that we don't have two different ways to represent the same type (e.g. heap type + nullability or BasicType).

tlively: +1, agree it would be good to use the shorthands when printing.

PR #4758, opened 2022-06-29
Disallow --nominal with GC

### LINKED ISSUE #4756: Fuzz bug on type emitting
```
$ v8 a.wasm.js --wasm-staging --experimental-wasm-eh --experimental-wasm-compilation-hints \
   --experimental-wasm-gc --experimental-wasm-typed-funcref --experimental-wasm-memory64 \
  --experimental-wasm-extended-const -- w.wasm

$ wasm-opt w.wasm -o z.wasm -all --disable-gc --nominal
warning: no passes specified, not doing any work

$ v8 a.wasm.js --wasm-staging --experimental-wasm-eh --experimental-wasm-compilation-hints \
   --experimental-wasm-gc --experimental-wasm-typed-funcref --experimental-wasm-memory64 \
  --experimental-wasm-extended-const -- z.wasm
a.wasm.js:38: CompileError: WebAssembly.Module(): Type index 1 is out of bounds @+14
var instance = new WebAssembly.Instance(new WebAssembly.Module(binary), {
                                        ^
CompileError: WebAssembly.Module(): Type index 1 is out of bounds @+14
    at a.wasm.js:38:41
```

That is, round-tripping the wasm using `-all --disable-gc --nominal` results in a binary that v8 cannot read.
[bug.zip](https://github.com/WebAssembly/binaryen/files/8996050/bug.zip)

cc @tlively - this sounds like a recent issue we had that also related to nominal mode without GC enabled?

tlively:
Nominal types don't make much sense without GC, and in particular trying to emit
them with typed function references but not GC enabled can result in invalid
binaries because nominal types do not respect the type ordering constraints
required by the typed function references proposal. Making this change was
mostly straightforward, but required fixing the fuzzer to use --nominal only
when GC is enabled and required exiting early from nominal-only optimizations
when GC was not enabled.

Fixes #4756.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -120,7 +120,7 @@ def randomize_feature_opts():
                     FEATURE_OPTS.extend(IMPLIED_FEATURE_OPTS[possible])
     print('randomized feature opts:', '\n  ' + '\n  '.join(FEATURE_OPTS))
     # Type system flags only make sense when GC is enabled
-    if "--disable-gc" not in FEATURE_OPTS:
+    if '--disable-gc' not in FEATURE_OPTS:
         FEATURE_OPTS.append(TYPE_SYSTEM_FLAG)
 
 
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -39,7 +39,6 @@ TYPE_SYSTEM_FLAG = '--nominal'
 
 # feature options that are always passed to the tools.
 CONSTANT_FEATURE_OPTS = ['--all-features']
-CONSTANT_FEATURE_OPTS.append(TYPE_SYSTEM_FLAG)
 
 INPUT_SIZE_MIN = 1024
 INPUT_SIZE_MEAN = 40 * 1024
@@ -120,6 +119,9 @@ def randomize_feature_opts():
                 if possible in IMPLIED_FEATURE_OPTS:
                     FEATURE_OPTS.extend(IMPLIED_FEATURE_OPTS[possible])
     print('randomized feature opts:', '\n  ' + '\n  '.join(FEATURE_OPTS))
+    # Type system flags only make sense when GC is enabled
+    if "--disable-gc" not in FEATURE_OPTS:
+        FEATURE_OPTS.append(TYPE_SYSTEM_FLAG)
 
 
 ALL_FEATURE_OPTS = ['--all-features', '-all', '--mvp-features', '-mvp']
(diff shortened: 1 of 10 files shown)


tlively:
* main
  * **PR #4758** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4758" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4758?utm_source=stack-comment).

scripts/fuzz_opt.py
```diff
@@ -120,6 +119,9 @@ def randomize_feature_opts():
                 if possible in IMPLIED_FEATURE_OPTS:
                     FEATURE_OPTS.extend(IMPLIED_FEATURE_OPTS[possible])
     print('randomized feature opts:', '\n  ' + '\n  '.join(FEATURE_OPTS))
+    # Type system flags only make sense when GC is enabled
+    if "--disable-gc" not in FEATURE_OPTS:
```
kripken, line 123:
```suggestion
    if '--disable-gc' not in FEATURE_OPTS:
```

src/passes/ConstantFieldPropagation.cpp
```diff
@@ -176,6 +176,9 @@ struct PCVScanner
 
 struct ConstantFieldPropagation : public Pass {
   void run(PassRunner* runner, Module* module) override {
+    if (!module->features.hasGC()) {
+      return;
+    }
```
kripken, line 181: I guess this is needed now because the fuzzer may call this pass without `--nominal`, and we want to avoid the Fatal below? It is odd though to silently return on not having GC but error on nominal. How about making the fuzzer not run these passes? We could move the relevant passes from `opt_choices` to `gc_opt_choices` and only pick from the latter when gc (+ nominal) is enabled?

kripken (commented):
lgtm with those comments addressed

src/passes/ConstantFieldPropagation.cpp
tlively: Yeah that would be another good solution. I slightly prefer keeping these early returns because they seem like what we will want to do in the long term; if running GC optimization passes on non-GC modules gracefully does nothing, then we can run GC optimization passes by default without a problem. I hope that the fatal errors on nonsupported type systems are more temporary and that we eventually update the passes to do what they can with whatever type systems we choose to permanently maintain.

kripken: I guess both are reasonable. I'd prefer to not run GC passes by default, though, as they add clutter - for example in debug mode they'd get printed out in the list. Seems nicer to only print out the relevant passes, which we do now. Maybe we can add a mechanism to avoid that, though, like a hook "canRun" that would avoid even adding the pass - that would avoid needing to skip the pass both in pass.cpp and in the fuzzer.

kripken: Anyhow, this lgtm + filing an issue to figure out a better solution later.

PR #4761, opened 2022-06-29
[Strings] Add string.new* instructions

kripken:
This is the first instruction from the Strings proposal.

This includes everything but interpreter support.


aheejin:
In `(string.new_wtf8 $memory $wtf8_policy ptr:address bytes:i32)`, can `$memory` be omitted? Can't see handling for that here, so..

kripken:
I think that's for future multimemory integration, but I could be wrong. That is, I made the instructions behave like `load` and `store` do right now, which don't mention a memory.

PR #4763, opened 2022-06-29
[Strings] Print shorthand types where possible

kripken:
As suggested in #4755

It doesn't look like there's an obvious way to share this code. Perhaps `TypePrinter`
could be in a header and used from `Print`?


tlively (approved):
Yeah, I refactored type printing but never got around to making Print.cpp use it properly...

PR #4768, opened 2022-06-30
[Strings] Add string.const

kripken:
This is more work than a typical instruction because it also adds a new section:
all the `(string.const "foo")` strings are put in a new "strings" section in the wasm, and
the instructions refer to them by index.


src/wasm-binary.h
```diff
@@ -326,7 +326,8 @@ enum Section {
   Code = 10,
   Data = 11,
   DataCount = 12,
-  Tag = 13
+  Tag = 13,
+  Strings = 14,
```
aheejin, line 330: Is it a regular section? [The string literal section](https://github.com/WebAssembly/stringref/blob/main/proposals/stringref/Overview.md#string-literal-section) in the explainer says
> The string.const section indicates the literal as an i32 index into a new custom section: a string table, encoded as a `vec(vec(u8))` of valid WTF-8 strings.

But [the binary encoding section](https://github.com/WebAssembly/stringref/blob/main/proposals/stringref/Overview.md#binary-encoding) says
```
stringrefs ::= section_14(0x00 vec(vec(u8)))
```
Not sure what 14 means here.

And [the requirement section](element) says:
> 5. Allow string literals in element sections

Not sure which one is correct, or are they talking about the same thing?

aheejin (commented):
Is there a text format representation of the string section, like `elem` or `data`? I can't find one in the explainer though.

kripken:
> Is there a text format representation of the string section, like elem or data? I can't find one in the explainer though.

My reading of the spec is that it isn't needed. Since we can just write `(string.const "foo")` anywhere. The section is just a binary format detail I guess. It is kind of similar but different to other things I guess.

src/wasm-binary.h
kripken: > Is it a regular section?

Good question... I hadn't noticed the word "custom" there.

I took "14" to mean it is a normal section of index 14. Which fits since it's right after Tag=13?

> "Allow string literals in element sections"

Maybe that means to allow stuff like
```wat
(elem (i32.const 0) (string.const "foo") (string.const "bar"))
```
? Not sure.

cc @wingo - maybe you can clarify your intention in the spec? Thanks!

kripken: Oh, scrolling all the way down I see backrefs showing that @dcodeIO raised these issues on the repo already. Sorry for the noise! And thanks @dcodeIO

aheejin (approved):
I think we can land this and change the section encoding part later if necessary.

src/wasm-binary.h
wingo: Yes thanks @dcodeIO, I merged those two clarifications -- normal section, and the intent was that `string.const` would be a constant expression, and that appearing in an element section is just a detail.

The text format has a couple open questions which I see you've run into already: https://github.com/WebAssembly/stringref/issues/23.

Specifically as regards the string literals section, just for info, I think you have the right minimum-viable strategy for now: just pull from `(string.const "str")` instructions and don't support an explicit section.  Things may get more complicated if we add support for a "late" string table for use in non-constant instructions; https://github.com/WebAssembly/stringref/issues/11#issuecomment-1130132626.  But that would be a post-MVP thing I think.

kripken: Thanks @wingo ! Ok, sounds like this PR is in the right direction. Let's land it for now (next I'll have more instructions PRs).

Not sure I understood this part:

> just pull from (string.const "str") instructions and don't support an explicit section

Is that regarding the text format specifically?

Issue #4739, opened 2022-06-17
Missing nan on wasm2js


willcohen:
When running the wasm2js's javascript version of the emscripten-compiled WASM of https://github.com/OSGeo/PROJ/blob/c0e2a9a49e01a8167944092a825044a3c5f468b9/src/geodesic.c#L83, that line fails with the following error:

```
Uncaught TypeError: nan is not a function
    at geod_init (projjs.js:378084:10)
    at pj_obj_create_28pj_ctx__2c_20dropbox__oxygen__nn_std____2__shared_ptr_osgeo__proj__common__IdentifiedObject__20__20const__29 (projjs.js:820529:7)
    at proj_create (projjs.js:820699:11)
    at proj_create_crs_to_crs (projjs.js:707136:12)
    at projjs.js:848414:22
    at ccall (projjs.js:847846:18)
    at <anonymous>:1:6
```

The js for the function in question:
```
 function geod_init($0, $1, $2) {
...
   $3 = +nan(258340 | 0);
...
```

Is there an issue with a collision between math.h's nan function and javascript's, leading to that function not being present?

willcohen:
As a followup, `sed -i 's/nan("0")/NAN/g' ../src/geodesic.c` on the source code gets the transpiled JS working, though I haven't checked enough in this case about what could eventually break with a logical change from C's nan().

kripken:
Looks like we emit this:

https://github.com/WebAssembly/binaryen/blob/bab21052c3e12df7ad2fd3711d0fe6e69e05f449/src/tools/wasm2js.cpp#L789-L794

Fixing that TODO should fix this issue. That is, since we don't need to obey the old asm.js rules, we can instead just use `NaN` normally from the global JS scope. The fix should remove that quoted code and replace our codegen for `nan` to use `NaN`. Hopefully that's all that is needed.

architsinghal-mriirs:
Hi ! I'll take this up !

kripken:
Great, thanks @architsinghal-mriirs !

architsinghal-mriirs:
Hi @kripken As per my understanding, I have to remove the above mentioned code ( line 791 to 794 ) & refactor nan with NaN in the JS files. Is this correct ?

PR #4770, opened 2022-07-03
Removed nan & infinity variables

architsinghal-mriirs:
Issue #4739
Hi @kripken I've made some changes as per the suggestions. I searched for nan in the entire codebase and found alot of instances of the same code. I've removed the code from few other files. Please review and let me know if this is this is correct.

Thanks and Regards
Archit Singhal


kripken:
In general I think this might be good, but let's get the test failures fixed first. Please run `./auto_update_tests.py wasm2js` and commit those changes.

PR #4777, opened 2022-07-07
[Strings] string.concat

kripken:


kripken:
Good question. I followed the example of `RefEq`, but all of `RefEq, StringConcat, StringEq` could all be under `Binary`...

Some reasons to have separate classes:
* Faster, since we constantly do a switch over all expression classes, and if we have an opcode inside them like `Binary` does then we'd have two switches basically.
* Shorter checking code: `curr->is<RefEq>()` versus checking if it's a binary + checking the op.
* Atm `Binary` is basically a simple MVP operation on numeric inputs. It can't trap, has no control flow, etc. When you read something is a `Binary` in the code that's simpler cognitively.

Some reasons to group into fewer classes with opcodes inside them:
* Less work when writing a PR like this. Especially with something like SIMD that has a billion operations.
* When the grouping matters it's simpler cognitively: `RefAs` is a family of similar operations, and often they are handled in the same way.

Overall my intuition is that grouping is nice when they are very similar, but maybe not otherwise.

If that makes sense, then perhaps we could have a `StringBinary` or `RefBinary`. But even `StringConcat` and `StringEq` feel "different" to me... only one can trap, for example. I could see an argument for combining `RefEq` and `StringEq` in theory...?

PR #4778, opened 2022-07-08
[Parser] Parse standard subtype declarations

tlively:
Parse type definitions with the format `(type $t (sub $super ...))`. Update the
test to use hybrid types so that the subtypes are reflected in the test output.


tlively:
* main
  * **PR #4778** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4778" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4778?utm_source=stack-comment).

src/wasm/wat-parser.cpp
```diff
@@ -925,18 +947,61 @@ template<typename Ctx> MaybeResult<> strtype(Ctx& ctx, ParseInput& in) {
...
-    ctx.typeDefs.push_back({name, start});
+    return name;
+  } else {
+    return Ok{};
   }
-  return Ok{};
 }
 
 // subtype ::= '(' 'sub' typeidx? strtype ')'
 //           | strtype
```
kripken, line 957: I wasn't familiar with this notation of `(sub ..)`. Which spec is that in?

test/lit/wat-kitchen-sink.wast
```diff
@@ -1,43 +1,47 @@
...
+ ;; CHECK:      (type $void (func_subtype func))
  (type $void (func))
- ;; CHECK:      (type $many (func (param i32 i64 f32 f64) (result anyref (ref func))))
+
+ ;; CHECK:      (type $many (func_subtype (param i32 i64 f32 f64) (result anyref (ref func)) func))
  (type $many (func (param $x i32) (param i64 f32) (param) (param $y f64)
                    (result anyref (ref func))))
 
- (type $s0 (struct))
+ (sub (type $s0 (struct)))
```
kripken, line 16: Reading this, I'm not actually sure what "sub" does or means..?

src/wasm/wat-parser.cpp
tlively: This is how subtypes are declared in the upstream GC spec: https://github.com/WebAssembly/gc/blob/main/test/core/gc/type-subtyping.wast. Now that I'm actually looking at the examples, though, I see that I have the order slightly wrong and that the `sub` should be inside the `type`. I'll fix the ordering.

test/lit/wat-kitchen-sink.wast
tlively: The reason there is no supertype here is that syntactically the supertypes are a vector, so it is allowed for there to be zero supertypes.

kripken (approved):

PR #4780, opened 2022-07-08
[Wasm GC] Check if ref.eq inputs can possibly be the same

kripken:
For them to be the same we must have a value that can appear on both
sides. If the heap types disallow that, then only null is possible, and if
that is impossible as well then the result must be 0.

### DIFF
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -1500,24 +1500,16 @@ struct OptimizeInstructions
       // Leave this for DCE.
       return;
     }
-    auto leftIsHeapSubtype = HeapType::isSubType(leftType, rightType);
-    auto rightIsHeapSubtype = HeapType::isSubType(rightType, leftType);
-    if (!leftIsHeapSubtype && !rightIsHeapSubtype) {
+    auto leftHeapType = leftType.getHeapType();
+    auto rightHeapType = rightType.getHeapType();
+    auto leftIsHeapSubtype = HeapType::isSubType(leftHeapType, rightHeapType);
+    auto rightIsHeapSubtype = HeapType::isSubType(rightHeapType, leftHeapType);
+    if (!leftIsHeapSubtype && !rightIsHeapSubtype &&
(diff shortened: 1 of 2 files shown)


src/passes/OptimizeInstructions.cpp
```diff
@@ -1493,9 +1493,33 @@ struct OptimizeInstructions
...
+        Builder(*getModule()).makeConst(Literal::makeZero(Type::i32));
+      replaceCurrent(getDroppedChildrenAndAppend(
+        curr, *getModule(), getPassOptions(), result));
+      return;
+    }
+
     // Equality does not depend on the type, so casts may be removable.
     //
-    // This is safe to do first because nothing else here cares about the type,
+    // This is safe to do first because nothing later down cares about the type,
```
tlively, line 1522:
```suggestion
    // This is safe to do first because nothing farther down cares about the type,
```

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -1690,6 +1688,299 @@
...
+  ;; NOMNL-NEXT:    )
+  ;; NOMNL-NEXT:    (ref.as_non_null
+  ;; NOMNL-NEXT:     (ref.cast_static $A
+  ;; NOMNL-NEXT:      (local.get $y)
+  ;; NOMNL-NEXT:     )
+  ;; NOMNL-NEXT:    )
+  ;; NOMNL-NEXT:   )
+  ;; NOMNL-NEXT:  )
+  ;; NOMNL-NEXT: )
+  (func $ref-eq-impossible (param $x eqref) (param $y eqref)
```
tlively, line 1889: Can we split up this function to make the output appear closer to the input?

tlively (approved):
LGTM modulo comments!

test/lit/passes/optimize-instructions-gc.wast
kripken: Sure, done.

tlively (approved):
Thanks!

PR #4781, opened 2022-07-08
[Strings] string.eq

kripken:
I considered maybe making this a shared class with `RefEq`. They are quite similar. But the internal opcode would need to be something like "compare references" versus "compare data", which seems slightly awkward. Overall the amount of boilerplate per new class is fairly small now, so I lean towards that.

### DIFF
--- a/scripts/gen-s-parser.py
+++ b/scripts/gen-s-parser.py
@@ -622,6 +622,7 @@ instructions = [
     ("string.encode_wtf8",   "makeStringEncode(s, StringEncodeWTF8)"),
     ("string.encode_wtf16",  "makeStringEncode(s, StringEncodeWTF16)"),
     ("string.concat",        "makeStringConcat(s)"),
+    ("string.eq",        "makeStringEq(s)"),
 ]
 
 
--- a/src/ir/cost.h
+++ b/src/ir/cost.h
@@ -684,6 +684,10 @@ struct CostAnalyzer : public OverriddenVisitor<CostAnalyzer, CostType> {
   CostType visitStringConcat(StringConcat* curr) {
     return 10 + visit(curr->left) + visit(curr->right);
   }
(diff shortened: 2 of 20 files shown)


scripts/gen-s-parser.py
```diff
@@ -622,6 +622,7 @@
     ("string.encode_wtf8",   "makeStringEncode(s, StringEncodeWTF8)"),
     ("string.encode_wtf16",  "makeStringEncode(s, StringEncodeWTF16)"),
     ("string.concat",        "makeStringConcat(s)"),
+    ("string.eq",        "makeStringEq(s)"),
```
tlively, line 625:
```suggestion
    ("string.eq",            "makeStringEq(s)"),
```

src/ir/cost.h
```diff
@@ -684,6 +684,10 @@ struct CostAnalyzer : public OverriddenVisitor<CostAnalyzer, CostType> {
   CostType visitStringConcat(StringConcat* curr) {
     return 10 + visit(curr->left) + visit(curr->right);
   }
+  CostType visitStringEq(StringEq* curr) {
+    // This assumes strings are interned in the engine.
```
tlively, line 688: I would be surprised if all strings were interned! This seems like it should be modeled as being very expensive since it's cost is unbounded.

tlively (approved):
LGTM modulo comments

src/ir/cost.h
kripken: I think the VM will intern them if they are used enough, at least in some cases, but I don't really know. JS string implementations are quite flexible. I guess we shouldn't strongly assume anything here, but like with concat, I don't think this can be a problem in general - there is no equivalent code that would be compared to this.

kripken: I adjusted it to 3 (in between 1 and 6 for interned and not), and added a comment.

PR #4786, opened 2022-07-08
[Wasm GC] RefIs / RefEq / RefTest all return a boolean

kripken:
This marks all reference operations that return 0/1 as doing so. This
allows various bitwise operations to be optimized on them.

This also marks StringEq as a boolean, though we can't test that fully yet
as Strings support is wip (no interpreter or other stuff yet).

As a driveby this moves `emitsBoolean` to its own file, and uses it
in `getMaxBits` to avoid redundancy (the redundant code paths now have
a `WASM_UNREACHABLE`). I could do that in a prior PR if you prefer but it
seems trivial?

This removes 14% of all `select` operations on j2wasm. Those selects
are `x ? 1 : 0` which we turn into `x` if the input is already a boolean.

### DIFF
--- a/src/ir/bits.h
+++ b/src/ir/bits.h
@@ -17,10 +17,11 @@
 #ifndef wasm_ir_bits_h
 #define wasm_ir_bits_h
 
+#include "ir/boolean.h"
 #include "ir/literal-utils.h"
+#include "ir/load-utils.h"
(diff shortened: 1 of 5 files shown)


kripken:
(Also 5% of selects on Dart.)

tlively (approved):

PR #4787, opened 2022-07-09
[GUFA] Simplify routines for dropping children (NFC)

aheejin:
I think we can actually move all the checking routines to
`getDroppedUnconditionalChildrenAndAppend` and remove `canRemove` and
`canRemoveStructurally`. Checking things in three different places was a
little hard to understand, and I think
`getDroppedUnconditionalChildrenAndAppend` actually needs to exclude
more things that `canRemoveStructurally` covers, if it is to be used in
other places as well.

Additionally, I think it'd better to make these two dropping-children
functions (`getDroppedChildrenAndAppend` and
`getDroppedUnconditionalChildrenAndAppend`) to a cpp file and make
`getDroppedChildrenAndAppend` an internal function, inaccessible to
outer passes, given that it looks all places should use
`getDroppedUnconditionalChildrenAndAppend` instead. But this can be a
follow-up.


src/passes/GUFA.cpp
```diff
@@ -143,23 +97,11 @@ struct GUFAOptimizer
...
-      if (canReplaceWithUnreachable(curr)) {
-        replaceCurrent(getDroppedUnconditionalChildrenAndAppend(
-          curr, wasm, options, builder.makeUnreachable()));
-      } else {
-        // We can't remove this, but we can at least put an unreachable
-        // right after it.
-        replaceCurrent(builder.makeSequence(builder.makeDrop(curr),
-                                            builder.makeUnreachable()));
-      }
-      optimized = true;
```
kripken, line 156: I think maybe the removal of this line is why the test results changed. We are not doing a refinalize like we used to?

I think there might be cases where we need it, so it would be safer to set `optimized=true` in all places we change things.

kripken (approved):
Nice! Definitely simpler.

src/passes/GUFA.cpp
aheejin: Thanks, done.

kripken (approved):

kripken:
@aheejin I think we can merge this into the `gufa` branch? (the CI error can be ignored)

PR #4789, opened 2022-07-11
[Parser] Start to parse instructions

tlively:
Update gen-s-parser.py to produce a second version of its parsing code that
works with the new wat parser. The new version automatically replaces the `s`
element argument in the existing parser with the `ctx` and `in` arguments used
by the new parser, so adding new instructions will not require any additional
work in gen-s-parser.py after this change.

Also add stub `make***` functions to the new wat parser, with a few filled out,
namely `makeNop`, `makeUnreachable`, `makeConst`, and `makeRefNull`. Update the
`global` parser to parse global initializer instructions and update
wat-kitchen-sink.wast to demonstrate that the instructions are parsed correctly.

Adding new instruction classes will require adding a new `make***` function to
wat-parser.cpp in additional to wasm-s-parser.{h,cpp} after this change, but
adding a trivial failing implementation is good enough for the time being, so I
don't expect this to appreciably increase our maintenance burden in the near
term.

The infrastructure for parsing folded instructions, instructions with operands,
and control flow instructions will be implemented in future PRs.

### DIFF
--- a/src/wasm/wat-parser.cpp
+++ b/src/wasm/wat-parser.cpp
@@ -48,8 +48,8 @@
 // before we parse instructions because they determine the types of instructions
 // such as global.get and ref.func.
 //
-// In the fifth and final phase, not yet implemented, parses the remaining
-// contents of all module elements, including instructions.
+// The fifth and final phase parses the remaining contents of all module
+// elements, including instructions.
 //
 // Each phase of parsing gets its own context type that is passed to the
 // individual parsing functions. There is a parsing function for each element of
@@ -122,6 +122,16 @@ struct ParseInput {
         ++lexer;
         // See comment on takeName.
         return Name(std::string(*id));
+      }
+    }
+    return {};
+  }
+
+  std::optional<std::string_view> takeKeyword() {
+    if (auto t = peek()) {
+      if (auto keyword = t->getKeyword()) {
+        ++lexer;
+        return *keyword;
       }
     }
     return {};
@@ -308,6 +318,10 @@ struct ParseDeclsCtx {
   using ArrayT = Ok;
   using GlobalTypeT = Ok;
 
+  using InstrT = Ok;
+  using InstrsT = Ok;
+  using ExprT = Ok;
+
   // Declared module elements are inserted into the module, but their bodies are
   // not filled out until later parsing phases.
   Module& wasm;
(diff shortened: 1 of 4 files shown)


tlively:
* main
  * **PR #4789** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4789" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4789?utm_source=stack-comment).

src/wasm/wat-parser.cpp
```diff
@@ -421,6 +474,205 @@ template<typename Ctx>
 MaybeResult<typename Ctx::StructT> structtype(Ctx&, ParseInput&);
 template<typename Ctx>
 MaybeResult<typename Ctx::ArrayT> arraytype(Ctx&, ParseInput&);
+template<typename Ctx>
+Result<typename Ctx::GlobalTypeT> globaltype(Ctx&, ParseInput&);
+
+// Instructions
+template<typename Ctx>
```
kripken, line 481: These declarations could be autogenerated using an include of `wasm-delegations.def`

src/wasm/wat-parser.cpp
```diff
@@ -813,6 +1065,602 @@ Result<typename Ctx::GlobalTypeT> globaltype(Ctx& ctx, ParseInput& in) {
...
+  }
+}
+
+template<typename Ctx> Result<typename Ctx::InstrT> makeNop(Ctx& ctx) {
+  if constexpr (parsingDefs<Ctx>) {
+    return ctx.builder.makeNop();
+  } else {
+    return Ok{};
+  }
+}
```
kripken, line 1135: Does SFINAE or such let us write a template def more than once? if so we could emit a failing stub for all the instructions (using `wasm-delegations.def`) and then just add the working ones after that.

src/wasm/wat-parser.cpp
tlively: This would be tricky because some of them have extra arguments, such as the calls with their `isReturn` arguments, i31 and struct get operations with their `signed_` argument, and various other instructions with operand arguments.

kripken: Could we get the `isReturn` from the name, if we read it again? Or is it lost at that point?

kripken: (In the old parser that was possible, though not sure if we actually used that ability or not)

PR #4790, opened 2022-07-11
[C-API] Add type system C-API

dcodeIO:
Enables toying around with type system modes using the C-API.


kripken (commented):
Thanks!

Looks good, although I do hope we can remove all the typesystem modes soon... the spec process has agreed on one. @tlively do you think that is possible? If so we might not need this.

tlively:
It makes sense to make the standard isorecursive type system and the nominal type system available via the C API. I think we should land this since it's a relatively small and simple change and removing type systems from Binaryen isn't a top priority right now.

kripken (approved):
Ok, sgtm to land then.

(looks like this has a conflict atm)

PR #4791, opened 2022-07-11
[C-API] Add packed type constants for use with GC types

dcodeIO:
Adds the packed typed constants for `i8` and `i16` to the C-API that will be useful to construct respectively observe certain GC types.


kripken (approved):

kripken:
CI error should be fixed once https://github.com/WebAssembly/binaryen/pull/4793 lands.

PR #4795, opened 2022-07-11
Remove usage of emscripten's deprecated allocate runtime function

sbc100:


src/js/binaryen.js-post.js
```diff
@@ -3323,7 +3324,7 @@ Module['emitText'] = function(expr) {
 Object.defineProperty(Module, 'readBinary', { writable: true });
 
 Module['readBinary'] = function(data) {
-  const buffer = allocate(data, ALLOC_NORMAL);
+  const buffer = allocateUTF8(data);
```
dcodeIO, line 3327: Afaik `data` is [expected to be an `Uint8Array`](https://github.com/AssemblyScript/binaryen.js/blob/3c60b3d48f78a0026afb25692e2d8dc611a68af7/index.d.ts#L2084) here. Does this work with `allocateUTF8`?

sbc100: I think readBinary takes a fileanme in string form.   At least looking the emscripten sources it looks like that is the case:

https://github.com/emscripten-core/emscripten/blob/97b7e2ea200eca744d07e95dd3ee5bb778166890/src/shell.js#L293-L296

https://github.com/emscripten-core/emscripten/blob/97b7e2ea200eca744d07e95dd3ee5bb778166890/src/web_or_worker_shell_read.js#L27

https://github.com/emscripten-core/emscripten/blob/97b7e2ea200eca744d07e95dd3ee5bb778166890/tools/lz4-compress.js#L30-L32

What else would it take other than filename?

dcodeIO: Ah, I think these are different `readBinary`s. The one here is the one introduced by the JS glue specifically for Binaryen.js, reading a binary in from a buffer.

sbc100: Ah!

kripken (approved):
This can also revert #4793

src/js/binaryen.js-post.js
```diff
@@ -3323,7 +3324,8 @@ Module['emitText'] = function(expr) {
 Object.defineProperty(Module, 'readBinary', { writable: true });
 
 Module['readBinary'] = function(data) {
-  const buffer = allocate(data, ALLOC_NORMAL);
+  const buffer = _malloc(data.length);
+  HEAP8.set(buffer, data);
```
dcodeIO, line 3328:
```suggestion
  HEAP8.set(data, buffer);
```

src/js/binaryen.js-post.js
```diff
@@ -2484,7 +2484,8 @@ function wrapModule(module, self = {}) {
       const segmentOffset = new Array(segmentsLen);
       for (let i = 0; i < segmentsLen; i++) {
         const { data, offset, passive } = segments[i];
-        segmentData[i] = allocate(data, ALLOC_STACK);
+        segmentData[i] = stackAlloc(data.length);
+        HEAP8.set(segmentData[i], data);
```
dcodeIO, line 2488:
```suggestion
        HEAP8.set(data, segmentData[i]);
```

PR #4798, opened 2022-07-12
[Strings] stringview access operations

kripken:
`stringview`s have a variety of access types based on what they are: wtf8 can
only advance forward as it reads, wtf16 only random access, and `iter` can iterate in any
direction. I think it's best to merge all these into a single class for simplicity.

### DIFF
--- a/src/ir/effects.h
+++ b/src/ir/effects.h
@@ -762,11 +762,11 @@ private:
     void visitStringIterNext(StringIterNext* curr) {
       // traps when ref is null.
       parent.implicitTrap = true;
-      // modifies state in the iterator. we model that as modifying heap memory
+      // modifies state in the iterator. we model that as accessing heap memory
       // in an array atm TODO consider adding a new effect type for this (we
-      // added one for arrays because struct/array operations often interleave
-      // like with vtable accesses, but it's not clear adding overhead to this
-      // class is worth it for string iters.
+      // added one for arrays because struct/array operations often interleave,
+      // say with vtable accesses, but it's not clear adding overhead to this
+      // class is worth it for string iters)
       parent.readsArray = true;
       parent.writesArray = true;
     }
--- a/src/ir/effects.h
+++ b/src/ir/effects.h
@@ -751,9 +751,31 @@ private:
       // traps when ref is null.
       parent.implicitTrap = true;
     }
-    void visitStringViewAccess(StringViewAccess* curr) {
+    void visitStringWTF8Advance(StringWTF8Advance* curr) {
       // traps when ref is null.
       parent.implicitTrap = true;
+    }
+    void visitStringWTF16Get(StringWTF16Get* curr) {
+      // traps when ref is null.
+      parent.implicitTrap = true;
+    }
+    void visitStringIterNext(StringIterNext* curr) {
+      // traps when ref is null.
+      parent.implicitTrap = true;
+      // modifies state in the iterator. we model that as modifying heap memory
+      // in an array atm TODO consider adding a new effect type for this (we
+      // added one for arrays because struct/array operations often interleave
+      // like with vtable accesses, but it's not clear adding overhead to this
+      // class is worth it for string iters.
+      parent.readsArray = true;
+      parent.writesArray = true;
+    }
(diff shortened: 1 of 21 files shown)


aheejin (approved):

src/ir/effects.h
```diff
@@ -751,6 +751,10 @@ class EffectAnalyzer {
       // traps when ref is null.
       parent.implicitTrap = true;
     }
+    void visitStringViewAccess(StringViewAccess* curr) {
+      // traps when ref is null.
+      parent.implicitTrap = true;
```
tlively, line 756: `stringview_iter.next` also has the side effect of advancing the iterator. We should probably represent that here somehow.

tlively (approved):

src/ir/effects.h
kripken: I ended up splitting this into 4 classes because of the differences in argument count and side effects, PTAL

src/ir/effects.h
```diff
@@ -751,10 +751,32 @@ class EffectAnalyzer {
...
+      parent.readsArray = true;
+      parent.writesArray = true;
+    }
+    void visitStringIterMove(StringIterMove* curr) {
+      // traps when ref is null.
+      parent.implicitTrap = true;
+      // see StringIterNext.
+      parent.readsArray = true;
+      parent.writesArray = true;
+    }
```
aheejin, line 779: If they have the same side effects, can't we merge `next` with `move`, with an opcode, and  if the opcode is `next`, artificially set `num` to `(i32.const 1)`? We shouldn't print the `(i32.const 1)` when emitting text or binary, but having it a real `Expression` can spare us of many `if`-`else`es, like, `if (op == move) do something; else do something else`

src/wasm.h
```diff
@@ -1766,18 +1764,48 @@ class StringAs : public SpecificExpression<Expression::StringAsId> {
...
 
-  StringViewAccessOp op;
+  Expression* ref;
+  Expression* pos;
+  Expression* bytes;
+
+  void finalize();
+};
+
+class StringWTF16Get : public SpecificExpression<Expression::StringWTF16GetId> {
```
aheejin, line 1779: Nit: I still think `GetCodeUnit` is clearer because everything else matches the instruction name, but I'm not very opinionated. I repost this just in case you dropped it by mistake 😅 In case you intentionally did it nevermind.

src/ir/effects.h
kripken: Hmm, I don't think we have an example where a child exists but isn't printed in some cases? Maybe I forgot something.

Another option is for `num` to be `nullptr` for `next`. But I'm not sure how much I like that either.

Maybe we can revisit this when we get to optimizations? If the if-else pattern you mention becomes a problem then we can refactor.

src/wasm.h
kripken: Oh, sorry, I thought you meant in `wasm-binary.h`. I did make sure to match the binary format there in our own binary constants. But otherwise we have some flexibility I think, both in how we split into classes (and sometimes we invent names for them), and in the names we give. And `StringWTF16GetCodeUnit` is a lot to type :smile:

src/ir/effects.h
aheejin: > Hmm, I don't think we have an example where a child exists but isn't printed in some cases? Maybe I forgot something.

Yeah this would be an exception.. But we already do many tweaks in wast printer (for example, omitting block printing), so I thought that wouldn't be too big an issue.

> Another option is for `num` to be `nullptr` for `next`. But I'm not sure how much I like that either.

That's an option too. I thought having if-else only when printing would be better than having it in all other computational places... Maybe I'm too optimistic?

> Maybe we can revisit this when we get to optimizations? If the if-else pattern you mention becomes a problem then we can refactor.

Yeah that sounds fine too.

aheejin (approved):

src/ir/effects.h
kripken: Ok, cool, let's land this for now. I agree that the wast printer workaround would be ok, so it really depends on how the optimizer code ends up looking. We'll see when we get there I guess...

PR #4799, opened 2022-07-12
[Parser][NFC] Refactor to use context callbacks

tlively:
The parser functions previously both parsed the input and controlled what was
done with the results using `constexpr` if-else chains. As the number of parsing
contexts grew, these if-else chains became increasingly complex and distracting
from the core parsing logic of the parsing functions.

To simplify the code, refactor the parsing functions to replace the `constexpr`
if-else chains with unconditional calls to methods on the context. To avoid
duplicating most method definitions for multiple parsing contexts, introduce new
utility contexts that implement common methods and (ab)use inheritance and
multiple inheritance to reuse their methods from the main parsing contexts.

This change will also make it easier to reuse the parser code for entirely
different purposes in the future by providing new context implementations. For
example, V8 could reuse the code and provide different parser contexts that
construct V8-internal data structures rather than Binaryen data structures.

### DIFF
--- a/src/wasm/wat-parser.cpp
+++ b/src/wasm/wat-parser.cpp
@@ -57,13 +57,6 @@
 // may be passed the appropriate context type and return the correct result type
 // for each phase.
 
-#define RETURN_OR_OK(val)                                                      \
-  if constexpr (parsingDecls<Ctx>) {                                           \
-    return Ok{};                                                               \
-  } else {                                                                     \
-    return val;                                                                \
-  }
-
 #define CHECK_ERR(val)                                                         \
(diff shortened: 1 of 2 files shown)


tlively:
* main
  * **PR #4799** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4799" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4799?utm_source=stack-comment).

tlively:
cc @jakobkummerow

kripken (approved):

PR #4803, opened 2022-07-14
[C-API] Add type builder C-API

dcodeIO:
Introduces the necessary APIs to use the type builder from C. Enables construction of compound heap types (arrays, structs and signatures) that may be recursive, including assigning concrete names to the built types and, in case of structs, their fields.

---

Some minor observations:

- From a producer's perspective, it would be nice if the type builder was a bit more integrated, so one can already work with temporary types during code generation and populate the concrete types once compilation is otherwise complete.

- Generated names of the created types (e.g. `${mut:ref?|...0|}`) seem off as these do not match the printing logic I have glimpsed at that seems to be supposed to make, say, `array.0`s. Not visible in the test fixture because concrete names are set.


kripken (commented):
In general this looks good to me, but this is a large API surface to expose to the C API - @tlively is there some other approach we are missing, or is this the right thing to do? And, is that API stable enough do you think?

tlively:
API LGTM. It should be fairly stable, I think.

> * From a producer's perspective, it would be nice if the type builder was a bit more integrated, so one can already work with temporary types during code generation and populate the concrete types once compilation is otherwise complete.

This would be possible today if you're very careful to replace the temporary types everywhere they are used without calling any methods on them as soon as the permanent types are built. The existence of temporary types is the least safe part of this API, though, so I would be hesitant to encourage them to be widely used.

> * Generated names of the created types (e.g. `${mut:ref?|...0|}`) seem off as these do not match the printing logic I have glimpsed at that seems to be supposed to make, say, `array.0`s. Not visible in the test fixture because concrete names are set.

Yeah, we're unfortunately half way through migrating to a new, consolidated type printing system. The new system that emits things like `array.0` exists, but Print.cpp hasn't been updated to use it yet.

tlively (approved):

PR #4805, opened 2022-07-14
[Strings] stringview_*.slice

kripken:
Unfortunately one slice is the same as python `[start:end]`, using 2 params,
and the other slice is one param, `[CURR:CURR+num]` (where CURR is implied
by the current state in the iter). So we can't use a single class here. Perhaps
a different name would be good, like `slice` vs `substring` (like JS does), but
I picked names to match the current spec.

### DIFF
--- a/src/passes/Print.cpp
+++ b/src/passes/Print.cpp
@@ -2315,6 +2315,21 @@ struct PrintExpressionContents
         WASM_UNREACHABLE("invalid string.move*");
     }
   }
+  void visitStringSliceWTF(StringSliceWTF* curr) {
+    switch (curr->op) {
+      case StringSliceWTF8:
+        printMedium(o, "stringview_wtf8.slice");
+        break;
+      case StringSliceWTF16:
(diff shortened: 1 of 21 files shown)


src/passes/Print.cpp
```diff
@@ -2315,6 +2315,21 @@ struct PrintExpressionContents
...
+  void visitStringSliceWTF(StringSliceWTF* curr) {
+    switch (curr->op) {
+      case StringSliceWTF8:
+        printMedium(o, "stringview_wtf8.slice");
+        break;
+      case StringSliceWTF16:
+        printMedium(o, "stringview_wtf16.slice");
+        break;
+      default:
+        WASM_UNREACHABLE("invalid string.move*");
```
aheejin, line 2327:
```suggestion
        WASM_UNREACHABLE("invalid string.slice*");
```

aheejin (approved):

src/passes/Print.cpp
kripken: Oops, thanks!

PR #4809, opened 2022-07-15
[Strings] stringview_wtf16.length

kripken:
This measures the length of a view, so it seems simplest to make it a
sub-operation of the existing measure instruction.

### DIFF
--- a/src/wasm.h
+++ b/src/wasm.h
@@ -595,7 +595,7 @@ enum StringMeasureOp {
   StringMeasureWTF8,
   StringMeasureWTF16,
   StringMeasureIsUSV,
-  StringMeasureView,
+  StringMeasureWTF16View,
 };
 
 enum StringEncodeOp {
--- a/src/wasm.h
+++ b/src/wasm.h
@@ -595,6 +595,7 @@ enum StringMeasureOp {
   StringMeasureWTF8,
   StringMeasureWTF16,
   StringMeasureIsUSV,
+  StringMeasureView,
 };
 
 enum StringEncodeOp {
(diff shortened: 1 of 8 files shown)


kripken:
They are separate types. Maybe that justifies separate instructions, but given the operation is identical I thought it's not worth the extra code. (When I add full validation support we'll of course enforce that the right type is being received.)

src/wasm.h
```diff
@@ -595,6 +595,7 @@ enum StringMeasureOp {
   StringMeasureWTF8,
   StringMeasureWTF16,
   StringMeasureIsUSV,
+  StringMeasureView,
```
aheejin, line 598:
```suggestion
  StringIsUSVSequence,
  StringViewLengthWTF16,
```
Don't we need to make the op name include it is for wtf16? I understand there's only one `length` op, but I think it's better for readability. Also not sure if we need `Measure` in the name.. `

The above is only my suggestion and I'm not very opinionated, but I think it'd be good to at least include `WTF16` in the last name..

aheejin (approved):

src/wasm.h
kripken: I'd prefer to keep the `StringMeasure` prefix since all the ops of instructions have that convention atm. It makes it easy to know which class an op belongs to. But good point on the other thing, being more explicit here might be helpful. How about `StringMeasureViewWTF16` or `StringMeasureWTF16View`? The last one just adds "View" to an existing one, which might be the most consistent.

aheejin: Yeah both sounds good.

PR #4810, opened 2022-07-16
[C/JS API] Add string reference types

dcodeIO:
Adds the `stringref`, `stringview_wtf8`, `stringview_wtf16` and `stringview_iter` reference types to the C and JS APIs.


test/example/c-api-kitchen-sink.c
```diff
@@ -270,6 +270,30 @@ void test_types() {
   BinaryenTypeExpand(dataref, &valueType);
   assert(valueType == dataref);
 
+  BinaryenType stringref = BinaryenTypeStringref();
+  printf("BinaryenTypeStringref: (ptr)\n", stringref);
+  assert(BinaryenTypeArity(stringref) == 1);
+  BinaryenTypeExpand(stringref, &valueType);
+  assert(valueType == stringref);
+
+  BinaryenType stringview_wtf8_ = BinaryenTypeStringviewWtf8();
```
dcodeIO, line 279: Note: The `_` is there to substitute `ref` so these don't conflict with the heap types of the same name that another PR adds tests for below.

kripken (approved):

PR #4811, opened 2022-07-18
Mutli-Memories Support in IR

ashleynh:
[Multiple Memories for Wasm Proposal](https://github.com/WebAssembly/multi-memory/blob/main/proposals/multi-memory/Overview.md)

This PR removes the single memory restriction in IR, adding support for a single module to reference multiple memories. To support this change, a new memory name field was added to 13 memory instructions in order to identify the memory for the instruction.

It is a goal of this PR to maintain backwards compatibility with existing text and binary wasm modules, so memory indexes remain optional for memory instructions. Similarly, the JS API makes assumptions about which memory is intended when only one memory is present in the module. Another goal of this PR is that existing tests behavior be unaffected. That said, tests must now explicitly define a memory before invoking memory instructions or exporting a memory, and memory names are now printed for each memory instruction in the text format.

There remain quite a few places where a hardcoded reference to the first memory persist (memory flattening, for example, will return early if more than one memory is present in the module). Many of these call-sites, particularly within passes, will require us to rethink how the optimization works in a multi-memories world. Other call-sites may necessitate more invasive code restructuring to fully convert away from relying on a globally available, single memory pointer.


juntao:
Would love to see this feature! We can really use it at WasmEdge. 😂 Please let us know how we can be of assistance.

tlively:
🎉 🎉 🎉

src/binaryen-c.cpp
```diff
@@ -3852,28 +3944,35 @@ void BinaryenSetMemory(BinaryenModuleRef module,
...
+  memory->max = int32_t(maximum); // Make sure -1 extends.
+  memory->shared = shared;
   if (exportName) {
     auto memoryExport = make_unique<Export>();
     memoryExport->name = exportName;
-    memoryExport->value = Name::fromInt(0);
+    memoryExport->value = memory->name;
     memoryExport->kind = ExternalKind::Memory;
-    wasm->addExport(memoryExport.release());
+    ((Module*)module)->addExport(memoryExport.release());
```
tlively, line 3959: It's more efficient to use the existing unique ptr rather than releasing it just for a new one to be created inside `addExport`.
```suggestion
    ((Module*)module)->addExport(std::move(memoryExport));
```

src/binaryen-c.cpp
```diff
@@ -3852,28 +3944,35 @@ void BinaryenSetMemory(BinaryenModuleRef module,
...
+                                         memory->name,
                                          segmentPassive[i],
                                          (Expression*)segmentOffsets[i],
                                          segments[i],
                                          segmentSizes[i]);
     curr->hasExplicitName = false;
-    wasm->dataSegments.push_back(std::move(curr));
+    ((Module*)module)->addDataSegment(std::move(curr));
   }
+  ((Module*)module)->removeMemories([&](Memory* curr) { return true; });
```
tlively, line 3974: We should also do something with any existing memory exports. Probably deleting them makes the most sense, but that would be a functional change, since previously they would have implicitly switched to exporting the new memory.

src/binaryen-c.cpp
```diff
@@ -3914,35 +4013,84 @@ uint32_t BinaryenGetMemorySegmentByteOffset(BinaryenModuleRef module,
...
+BinaryenIndex BinaryenMemoryGetInitial(BinaryenModuleRef module,
+                                       const char* name) {
+  // Maintaining compatibility for instructions with a single memory
+  if (name == nullptr && module->memories.size() == 1) {
+    name = module->memories[0]->name.c_str();
+  }
+  auto* memory = ((Module*)module)->getMemoryOrNull(name);
+  if (memory == nullptr) {
+    Fatal() << "invalid memory '" << name << "'.";
+  }
```
tlively, line 4027: Since we are producing a fatal error anyway, might as well call `getMemory` instead of `getMemoryOrNull` and skip the error checking here.

src/ir/memory-utils.cpp
```diff
@@ -62,11 +66,11 @@ bool flatten(Module& wasm) {
     }
     std::copy(segment->data.begin(), segment->data.end(), data.begin() + start);
   }
-  dataSegments.resize(1);
   dataSegments[0]->offset->cast<Const>()->value = Literal(int32_t(0));
   dataSegments[0]->data.swap(data);
   wasm.removeDataSegments(
     [&](DataSegment* curr) { return curr->name != dataSegments[0]->name; });
+  dataSegments.resize(1);
```
tlively, line 73: It looks like moving this would cause problems if there are no segments to begin with. Then `dataSegments[0]` would be out of bounds.

src/ir/memory-utils.h
```diff
@@ -30,19 +30,25 @@ namespace wasm::MemoryUtils {
...
   }
 }
 
 // Try to merge segments until they fit into web limitations.
 // Return true if successful.
+// Does not yet support multi-memories
 inline bool ensureLimitedSegments(Module& module) {
+  if (module.memories.size() > 1) {
+    return false;
+  }
```
tlively, line 51: I wonder if we should produce a fatal error here (and in similar places) so we don't forget to come back and fix it up later. @kripken, wdyt?

src/js/binaryen.js-post.js
```diff
@@ -689,30 +689,30 @@ function wrapModule(module, self = {}) {
   }
 
   self['memory'] = {
-    'size'() {
-      return Module['_BinaryenMemorySize'](module);
+    'size'(name) {
+      return Module['_BinaryenMemorySize'](module, strToStack(name));
```
tlively, line 693: I see we have existing bugs here, but every `strToStack` call needs to be wrapped in a `preserveStack` call to avoid permanently leaking stack space.

tlively (commented):
Here are some comments on everything before `src/passes` in GitHub's review order. Taking a look at everything under `src/passes` next.

src/wasm/wasm-debug.cpp
```diff
@@ -1065,7 +1065,8 @@ void writeDWARFSections(Module& wasm, const BinaryLocations& newLocations) {
 
   updateDebugLines(data, locationUpdater);
 
-  updateCompileUnits(info, data, locationUpdater, wasm.memory.is64());
+  bool is64 = wasm.memories.size() > 0 ? wasm.memories[0]->is64() : false;
```
kripken, line 1068: Worth a comment here that we assume all memories have the same type (all 32 or all 64).

src/wasm/wasm-s-parser.cpp
```diff
@@ -1939,7 +1999,17 @@ SExpressionWasmBuilder::makeLoad(Element& s, Type type, bool isAtomic) {
...
-  size_t i = parseMemAttributes(s, ret->offset, ret->align, ret->bytes);
+  Index i = 1;
+  Name memory;
+  // Check to make sure there are more than the default args & this str isn't
+  // the mem attributes
+  if (hasMemoryIdx(s, 2, i)) {
+    memory = getMemoryName(*s[i++]);
+  } else {
+    memory = getMemoryNameAtIdx(0);
+  }
```
kripken, line 2010: These 5 lines seem to repeat a lot - maybe a helper function in this file?

src/wasm/wasm-validator.cpp
```diff
@@ -3036,25 +3045,28 @@ static void validateGlobals(Module& module, ValidationInfo& info) {
 }
 
 static void validateMemory(Module& module, ValidationInfo& info) {
-  auto& curr = module.memory;
+  if (module.memories.empty()) {
+    return;
+  }
+  auto& curr = module.memories[0];
```
kripken, line 3051: maybe a TODO comment for looping over the memories?

src/wasm/wasm.cpp
```diff
@@ -1590,6 +1608,13 @@ void Module::removeTags(std::function<bool(Tag*)> pred) {
   removeModuleElements(tags, tagsMap, pred);
 }
 
+void Module::updateDataSegmentsMap() {
+  dataSegmentsMap.clear();
+  for (auto& curr : dataSegments) {
+    dataSegmentsMap[curr->name] = curr.get();
+  }
+}
```
kripken, line 1616: We don't seem to have this for ElementSegments? Why is this not symmetric to that?

test/atomics-unshared.wast.from-wast
```diff
@@ -3,7 +3,7 @@
  (memory $0 1 1)
  (func $foo
   (drop
-   (i32.atomic.rmw.cmpxchg
+   (i32.atomic.rmw.cmpxchg $0
```
kripken, line 6: To reduce the test diff, can we avoid printing the name when there is a single memory?

kripken (commented):
Really nice work! Relative to the massive amount of code here I have very few comments. (So far I read the source files from `src/shell-interface.h` - about the middle - in the github order here, to try to avoid overlapping with @tlively 's comments.)

Also none of my comments are really serious. So I think perhaps this can land without addressing all of them, and the rest could be fixed in followups. On a massive PR like this that might be the most practical thing. @tlively what do you think?

One critical thing before landing would be fuzzing, though - we shouldn't break that.

src/wasm-interpreter.h
```diff
@@ -23,17 +23,16 @@
...
 #include "ir/module-utils.h"
 #include "support/bits.h"
 #include "support/safe_integer.h"
 #include "wasm-builder.h"
 #include "wasm-traversal.h"
 #include "wasm.h"
+#include <cmath>
+#include <limits.h>
+#include <sstream>
+#include <variant>
```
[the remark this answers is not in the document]
ashleynh: Yes, clang-format.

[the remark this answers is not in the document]
kripken: Hmm, it looks wrong though...

https://github.com/WebAssembly/binaryen/blob/c81a4f6004950f3c29c067a689a3c0c8a0065454/src/wasm-interpreter.h#L25-L36

I would expect this:

```cpp
  
 #include <cmath> 
 #include <limits.h> 
 #include <sstream> 
 #include <variant> 

 #include "ir/module-utils.h" 
 #include "support/bits.h" 
 #include "support/safe_integer.h" 
 #include "wasm-builder.h" 
 #include "wasm-traversal.h" 
 #include "wasm.h" 

  ```

We should have the general C++ includes first and then our own stuff later, and after a blank line. That's how all other files are I think. Does changing it to be like other files get undone by clang-format if you rerun it? If so we should look into that, though I guess it's not urgent.

**Comment by L-jasmine** [comment]
I find a bug.
`(i32.load $1 offset=4 )` will be compiled into `28H 42H 01H 04H`.
but in other wasm project. it is `28H 42H 04H 01H`.

File: src/wasm/wasm-stack.cpp [hunk 13]
```diff
@@ -2516,8 +2519,19 @@ void BinaryInstWriter::setScratchLocals() {
...
+  uint32_t memoryIdx = parent.getMemoryIndex(memory);
+  if (memoryIdx > 0) {
+    // Set bit 6 in the alignment to indicate a memory index is present per:
+    // https://github.com/WebAssembly/multi-memory/blob/main/proposals/multi-memory/Overview.md
+    alignmentBits = alignmentBits | 1 << 6;
+  }
+  o << U32LEB(alignmentBits);
+  if (memoryIdx > 0) {
+    o << U32LEB(memoryIdx);
+  }
```
L-jasmine, line 2534: memoryIdx after offset

src/wasm/wasm-stack.cpp
```diff
@@ -2516,8 +2519,19 @@ void BinaryInstWriter::setScratchLocals() {
...
   o << U32LEB(offset);
```
L-jasmine, line 2535: `memoryIdx` should after `offset`

tlively (commented):
Finished looking at src/passes

tlively:
I would be ok with landing this as-is and applying fixups in follow-on PRs after the fuzzer passes :+1:

kripken (commented):
Looks like there are some new test files here, `test/multi-memories*`? It would be good to make those `wast` files instead of `wasm`, unless I'm missing a reason not to? As text they are easier to review, and look at if a test fails in the future.

test/lit/passes/O4_disable-bulk-memory.wast
```diff
@@ -80,8 +80,8 @@
  (global $global$6 i32 (i32.const 100))
  ;; CHECK:      (elem (i32.const 0) $null)
 
- ;; CHECK:      (export "memory" (memory $0))
- (export "memory" (memory $0))
+ ;; CHECK:      (export "memory" (memory $1))
+ (export "memory" (memory $1))
```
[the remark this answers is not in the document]
ashleynh: This is because of what occurs on line 57: `(import "env" "memory" (memory $1 1))`. We import a memory with the name $1. Before multi-memories, wasm-validator would verify the memory export by checking whether the name was "0" (as it is in this test) or the name of the single memory (which would have failed as the memory is named $1) [(link)](https://sourcegraph.com/github.com/WebAssembly/binaryen@main/-/blob/src/wasm/wasm-validator.cpp?L2943&subtree=true). 

In a multi-memories world, wasm-validator now looks for the exported memory name (like for function/global/table exports) in our list of memories. I changed the name of the export in this test to actually match the name of the import.

ashleynh:
> Looks like there are some new test files here, test/multi-memories*? It would be good to make those wast files instead of wasm, unless I'm missing a reason not to? As text they are easier to review, and look at if a test fails in the future.

@kripken I added the multi-memories wasm test files to verify correct parsing of binary format and printing to text format. There are additional ".wasm.fromBinary" text files added for each binary to verify the wast is as expected. Happy to make any change needed here.

test/example/relooper-fuzz2.c
```diff
@@ -23,7 +23,8 @@ int main() {
                    0,
                    0,
                    BinaryenTypeInt32(),
-                   BinaryenConst(module, BinaryenLiteralInt32(4))),
+                   BinaryenConst(module, BinaryenLiteralInt32(4)),
+                   "0"),
```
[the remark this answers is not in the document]
ashleynh: For the JS API, null is appended for the memory name and this is handled inside the different calls (BinaryenLoad, BinaryenStore, etc.). For C calls, we can set a default parameter of null for the memName to avoid updating these tests.

test/lit/passes/O4_disable-bulk-memory.wast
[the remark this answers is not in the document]
kripken: I see, thanks!

test/example/relooper-fuzz2.c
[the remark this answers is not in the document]
kripken: I see, right... a null is the best we could do, since we must still change the C API to add the new param either way. Sounds ok to leave null for later then.

kripken:
@ashleynh 

> I added the multi-memories wasm test files to verify correct parsing of binary format and printing to text format.

Makes sense, we do want that coverage. But adding a `wast` file in the same directory should get the same test coverage (it will also emit `.fromBinary` outputs etc. - it will write to binary, read, then print that). And text tests are generally easier to handle.

We do have some actual `.wasm` files in the test suite, but mainly to handle code that can't be directly represented in the text format for some reason.

kripken (approved):

tlively (approved):

src/ir/memory-utils.cpp
ashleynh: Moving this was okay because there is a check above on line 46 to return early `if (dataSegments.size() == 0)`. Removed the resize line and the asan tests are passing 👍🏾

src/wasm-interpreter.h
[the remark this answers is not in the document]
ashleynh: Looks like this was fixed back 🎉  and is as you expected above [link](https://sourcegraph.com/github.com/WebAssembly/binaryen@multi-memories/-/blob/src/wasm-interpreter.h?subtree=true)

src/passes/Print.cpp
```diff
@@ -540,6 +540,8 @@ struct PrintExpressionContents
       o << (curr->signed_ ? "_s" : "_u");
     }
     restoreNormalColor(o);
+    o << ' ';
+    printName(curr->memory, o);
```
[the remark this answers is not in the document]
ashleynh: Done!

test/atomics-unshared.wast.from-wast
ashleynh: Done! 🎉

src/wasm/wasm.cpp
ashleynh: It's a bug that I never updated the maps when writing the PR for First Class Data Segments. In the future, I'd make this a PR that comes before multi-memories, and it still can!

This is not symmetric to ElementSegments, because we don't do vector swaps for ElementSegments like we do for DataSegments: [Ex1](https://sourcegraph.com/github.com/WebAssembly/binaryen@multi-memories/-/blob/src/ir/memory-utils.h?L165:15&subtree=true) [Ex2](https://sourcegraph.com/github.com/WebAssembly/binaryen@multi-memories/-/blob/src/passes/MemoryPacking.cpp?L516:11&subtree=true)

I thought it might be wasteful to call updateMaps() just for DataSegments, but there is precedent for doing so with [Functions](https://sourcegraph.com/github.com/WebAssembly/binaryen@multi-memories/-/blob/src/passes/opt-utils.h?L39:6&subtree=true). Happy to remove this function and just call updateMaps(), wdyt?

test/lit/wasm-split/instrument-in-memory.wast
```diff
@@ -1,88 +1,46 @@
...
-;; CHECK-NEXT:  (i32.atomic.store8 offset=1
-;; CHECK-NEXT:   (i32.const 0)
-;; CHECK-NEXT:   (i32.const 1)
-;; CHECK-NEXT:  )
-;; CHECK-NEXT:  (local.get $0)
-;; CHECK-NEXT: )
 
 ;; Check that the profiling function is correct.
 
-;; CHECK:      (func $__write_profile (param $addr i32) (param $size i32) (result i32)
```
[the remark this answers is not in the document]
ashleynh: I reverted this test file back to main, because it was manually created and just added the memory line I needed for the test to pass. We're all good now 👍🏾

src/wasm/wasm.cpp
kripken: Hmm, I get 404s on both the first 2 links?

ashleynh: Makes sense, the multi-memories branch is deleted now. Fixed the links to point to main now. [Ex1](https://sourcegraph.com/github.com/WebAssembly/binaryen@main/-/blob/src/ir/memory-utils.h?L165&subtree=true) [Ex2](https://sourcegraph.com/github.com/WebAssembly/binaryen@main/-/blob/src/passes/MemoryPacking.cpp?L176&subtree=true)

kripken: I see, thanks! Yeah, maybe it's worth having a function for it, if we have multiple places that don't need to update all the maps. lgtm as it is then.

PR #4815, opened 2022-07-19
[Test] Refactor C API kitchen sink test's Memory usage

kripken:
This moves the memory creation higher up, so that it is before any of
the instructions that use a memory. That order makes more sense
anyhow I think, and also it's future compatible with multimemory
(where memories need to be explicitly referenced from instructions,
so the memory must appear before).

The output changes slightly since by creating the memory earlier
we also create its export earlier, and before the other export.


ashleynh (approved):
Looks great. Tested locally 👍🏾

PR #4819, opened 2022-07-22
[Wasm GC] Properly represent nulls in i31

kripken:
This was wrong all along and I'm not sure why the fuzzer never noticed
that...

The proposed encoding here is simple: we store `i31` values in the `literal.i32`
field. The top bit says if a value exists, which means `literal.i32 == 0` is the
same as null.


test/lit/exec/i31.wast
```diff
@@ -0,0 +1,114 @@
...
+;; RUN: wasm-opt %s -all --fuzz-exec -q -o /dev/null 2>&1 | filecheck %s
+
+(module
+  ;; CHECK:      [fuzz-exec] calling null-local
+  ;; CHECK-NEXT: waka i31ref(null) : (ref null i31)
+  ;; CHECK-NEXT: [fuzz-exec] note result: null-local => 1
+  (func "null-local" (result i32)
+    (local $ref (ref null i31))
+    (ref.is_null
+      (ref.null i31)
```
tlively, line 12: Is this supposed to be a local.get?

tlively (approved):
Oops! I guess maybe we weren't testing with nullable i31 references until recently?

test/lit/exec/i31.wast
kripken: Oops, yes, thanks...

PR #4820, opened 2022-07-22
Add TypeBuilder API to group and order isorecursive types

### LINKED ISSUE #4816: Fuzz bug on hybrid typing + SignatureRefining
Testcase:
```wat
(module
 (type ${} (struct_subtype  data))
 (type $funcref_=>_none (func_subtype (param funcref) func))
 (type $v128_=>_i32 (func_subtype (param v128) (result i32) func))
 (type $none_=>_none (func_subtype func))
 (type $anyref_v128_f64_=>_ref?|data| (func_subtype (param anyref v128 f64) (result (ref null data)) func))
 (type $eqref_v128_f32_=>_v128 (func_subtype (param eqref v128 f32) (result v128) func))
 (memory $0 (shared 16 17))
 (elem declare func $0 $1)
 (export "func_86" (func $3))
 (export "func_90" (func $4))
 (export "func_16_invoker" (func $2))
 (func $3 (type $funcref_=>_none) (param $0 funcref)
  (i32.store8
   (i32.const 22)
   (i32.const 130)
  )
 )
 (func $4 (type $v128_=>_i32) (param $0 v128) (result i32)
  (global.set $global$0
   (i32.const -1)
  )
  (call $3
   (ref.func $1)
  )
  (i32.const -32767)
 )
 (func $2 (type $none_=>_none)
  (drop
   (call $1
    (ref.func $0)
    (v128.const i32x4 0xffffffff 0x1fffffff 0x0000ff95 0x00000000)
    (f64.const 17179869184)
   )
  )
 )
 (func $1 (type $anyref_v128_f64_=>_ref?|data|) (param $0 anyref) (param $1 v128) (param $2 f64) (result (ref null data))
  (if
   (global.get $global$0)
   (unreachable)
  )
  (struct.new_default ${})
 )
 (func $0 (type $eqref_v128_f32_=>_v128) (param $0 eqref) (param $1 v128) (param $2 f32) (result v128)
  (unreachable)
 )
)
```
STR:
```
$ wasm-opt a.wat -all --hybrid --signature-refining
Fatal: Internal GlobalTypeRewriter build error: Heap type has an undeclared child at index 4
```
Works ok with nominal types.

To reproduce you need to allow hybrid typing to work on that pass. You can grab the `hybrid` branch which adds that if condition.

tlively:
The existing TypeBuilder::build API requires isorecursive types to have been
inserted into the TypeBuilder in a valid order and for the correct recursion
groups to have been explicitly created. This makes sense when parsing Wasm
modules, since invalid type sections need to be handled as errors, but it is
overly restrictive for optimizations that introduce new types.

For example, it is possible for optimizations like signature refining to create
new mutually recursive types that did not previously exist in the module. Under
isorecursive typing, these new types need to be placed into their own recursion
group separate from any existing recursion group. It would be prohibitively
complex, however, for each optimization to be responsible for detecting these
mutual recursions and setting up the correct recursion groups in the correct
order.

Instead, provide a new API, TypeBuilder::groupAndBuild, that uses Tarjan's
strongly connected components algorithm to automatically group and order types
before building them. Any optimization that produces new types should use this
new API so that it doesn't have to worry about correctly grouping and ordering
isorecursive types.

Fixes #4816.


tlively:
* main
  * **PR #4820** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4820" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4820?utm_source=stack-comment).

tlively:
This is essentially ready for review, but I marked it draft because I still need to add tests for the ordering based on supertype constraints.

src/passes/SignatureRefining.cpp
```diff
@@ -51,8 +51,8 @@ struct SignatureRefining : public Pass {
     if (!module->features.hasGC()) {
       return;
     }
-    if (getTypeSystem() != TypeSystem::Nominal) {
-      Fatal() << "SignatureRefining requires nominal typing";
+    if (getTypeSystem() == TypeSystem::Equirecursive) {
+      Fatal() << "SignatureRefining requires nominal or isorecursive typing";
```
kripken, line 55: In the `hybrid` branch this and a bunch of other places are updated. The code there does checks for "must be nominal or isorecursive" which I think is more explicit? Sgtm to add one of those checks in this PR but it would be good to have the same code to avoid conflicts later.

src/wasm/wasm-type.cpp
```diff
@@ -3968,6 +3968,188 @@ TypeBuilder::BuildResult TypeBuilder::build() {
...
+    for (auto pred : succ.getReferencedHeapTypes()) {
+      auto predIndexIt = typeIndices.find(pred);
+      if (predIndexIt == typeIndices.end()) {
+        // This is a basic or canonical child, so it is a predecessor of all the
+        // types we are building and won't affect their ordering or grouping.
+        continue;
+      }
+      // Record the successor if we have not already recorded it.
+      auto& succIndices = nodes[predIndexIt->second].succIndices;
+      if (succIndices.empty() || succIndices.back() != succIndex) {
```
kripken, line 4018: Why is it enough to check `back()` here and not look through the entire vector?

tlively: We are recording the successor indices in order from zero to `numTypes`, so the highest successor index is always at the end of the vector and is never greater than the current successor index. That means that if there would be a repeat, it would be at the end of the vector.

tlively:
I realized that similar types were being collapsed because `GlobalTypeRewriter` was not setting up any recursion groups for the replacement types. This is a problem because we need to maintain the property that when two original types are distinct, their rewritten types are distinct as well. Otherwise we might accidentally change semantics around casts. To fix this I will:

 - Update `GlobalTypeRewriter` to set up recursion groups
 -  Allow recursion groups in the input to `TypeBuilder::groupAndBuild`
 - Create synthetic edges in the graph input to Tarjan's SCC that will ensure that input rec groups are not split up in the output. (But groups might be merged!)

src/wasm/wasm-type.cpp
kripken: I see, thanks!

kripken:
Interesting... and also sounds complicated. This is just a fuzz bug, and I don't think any major user of wasm GC will hit it any time soon. If it's not a quick fix then it might not be urgent.

tlively:
@kripken, this should be good to go now.

src/wasm-type.h
```diff
@@ -672,6 +672,11 @@ struct TypeBuilder {
   // relationships are not valid.
   BuildResult build();
 
+  // Like `build`, but first automatically groups types into rec groups and
+  // sorts the groups to create a valid type ordering, if possible. Explicitly
+  // set up rec groups will not be split up, but they may be merged.
+  BuildResult groupAndBuild();
```
kripken, line 678: How does this automatic grouping work? I'm assuming that's only relevant if the user never called `createRecGroup` - if not, how does that interact?

test/lit/passes/signature-refining-isorecursive.wast
```diff
@@ -0,0 +1,207 @@
...
+ )
+
+ ;; CHECK:      (elem declare func $foo)
+
+ ;; CHECK:      (func $foo (type $0) (param $0 (ref $0))
+ ;; CHECK-NEXT:  (call $foo
+ ;; CHECK-NEXT:   (ref.func $foo)
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT: )
+ (func $foo (type $0)
```
kripken, line 100: where is the param declaration for this?

tlively: There is none, since the parameters are given in the definition of `type $0` and don't need to be repeated. I could add them here if you want, but IMO they just clutter things.

src/wasm-type.h
tlively: That's what Tarjan's algorithm is for. The strongly connected components in the type definition graph are turned into rec groups. Explicitly defined rec groups are represented using synthetic edges in the input to Tarjan's to make sure all the types in an input rec group are in the same strongly connected component and therefore in the same output rec group.

test/lit/passes/signature-refining-isorecursive.wast
kripken: Ah, does the text format allow skipping params that are not used (and so they don't need names, which aren't in the type)?

I wasn't aware of that, if so. Seems less surprising to write them out, and also it helps compare the before and after states without reading above to the type definition. But I don't feel strongly.

tlively: Correct. You can even still refer to them by index in that case. I'll add them here.

src/wasm-type.h
kripken: I guess I just don't understand how this works. The existing types are all in rec groups, correct? (Looks like the PR changes to in `type-updating.cpp` assume that, in fact.) It seems odd to me that there is a way to infer grouping for new types added after them - that's a new problem that didn't exist before, and it's not obvious to me how it is solved. I get (sort of :laughing: ) what Tarjan does, but it makes the assumption that it *can* find a grouping without the user providing one. Why is that?

test/lit/passes/type-refining-isorecursive.wast
```diff
@@ -0,0 +1,141 @@
...
+    (local.get $x)
+   )
+  )
+ )
+)
+
+(module
+ ;; The types will all be mutually recursive because they all reference and are
+ ;; referenced by $all, but now we need to worry about ordering supertypes
+ ;; correctly.
```
kripken, line 62: What should I be looking for in terms of the right order of supertypes? Since they are all in one rec group, doesn't the order not matter - they can all refer to each other?

src/wasm-type.h
kripken: (IOW, what I'm trying to ask is - why is the result here guaranteed to be what the user intended? Since the user did not provide more information, the problem seems too ambiguous to me. I'm definitely not looking at this right...)

tlively:
While talking with @kripken offline, I realized that this general solution is incorrect. It's not just types within the same rec group in the input but _all_ the input types that are distinct, so the only correct solution here is to put them all in a single large rec group. Otherwise independent types in distinct rec groups could be refined to the same types and have their rec groups incorrectly decuplicated. I'll open a new PR with that drastically simpler fix.

PR #4823, opened 2022-07-25
Move method implementations to drop.cpp

aheejin:
Now that we have `getDroppedUnconditionalChildrenAndAppend`, all passes
need to use it rather than the old `getDroppedChildrenAndAppend`, which
is used within the new method
`getDroppedUnconditionalChildrenAndAppend`.

This creates `drop.cpp` and move method implementations in drop.h there,
and merge two methods given that the old method is not supposed to be
used from outside anyway, and drops `Unconditional` from the new method
name because this is the only method and doesn't have to be specific
about that.

### DIFF
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -1514,7 +1514,7 @@ struct OptimizeInstructions
       // reference can appear on both sides.
       auto* result =
         Builder(*getModule()).makeConst(Literal::makeZero(Type::i32));
-      replaceCurrent(getDroppedChildrenAndAppend(
+      replaceCurrent(getDroppedUnconditionalChildrenAndAppend(
         curr, *getModule(), getPassOptions(), result));
       return;
     }
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -1536,7 +1536,7 @@ struct OptimizeInstructions
     if (areConsecutiveInputsEqualAndFoldable(curr->left, curr->right)) {
       auto* result =
         Builder(*getModule()).makeConst(Literal::makeOne(Type::i32));
-      replaceCurrent(getDroppedChildrenAndAppend(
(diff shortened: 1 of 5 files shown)


src/passes/OptimizeInstructions.cpp
```diff
@@ -1514,7 +1514,7 @@ struct OptimizeInstructions
       // reference can appear on both sides.
       auto* result =
         Builder(*getModule()).makeConst(Literal::makeZero(Type::i32));
-      replaceCurrent(getDroppedChildrenAndAppend(
+      replaceCurrent(getDroppedUnconditionalChildrenAndAppend(
```
kripken, line 1517: I don't think this is strictly needed, as `curr` is `RefAs` whose shape is simple. Given that, we know we won't need the extra logic in `getDroppedUnconditionalChildrenAndAppend` vs `getDroppedChildrenAndAppend`. The old code is also slightly faster to run, and shorter to write.

I guess on the other hand it's less to think about to just have a single 'safe' API for this?

aheejin: > I guess on the other hand it's less to think about to just have a single 'safe' API for this?

Yeah this is the rationale behind this PR. Sometimes you can use the old method but you have to be careful. Using the new method can add a little unnecessary work but safer. And the additional work doesn't involve checking children or anything so I didn't think it mattered much. But your preference may be different, in which case I'm OK either way.

kripken: Good point that the additional work is pretty cheap, yeah.

I think this is probably worthwhile. How about changing the name, in this case? If there's just a single public API then I think the name doesn't need "unconditional" (just the comment is enough).

aheejin: Hmm, then I wonder we need two different methods for this..? How about merging the two, given that the old method is not supposed to be used? I'll try that; PTAL.

kripken (approved):
Nice!

PR #4825, opened 2022-07-25
[Wasm GC] i31get can trap

kripken:
Another fuzz bug found now that the fuzzer has more i31 code to fuzz on...


src/ir/effects.h
```diff
@@ -653,7 +653,12 @@ class EffectAnalyzer {
     void visitTupleMake(TupleMake* curr) {}
     void visitTupleExtract(TupleExtract* curr) {}
     void visitI31New(I31New* curr) {}
-    void visitI31Get(I31Get* curr) {}
+    void visitI31Get(I31Get* curr) {
+      // traps when the ref is null
+      if (curr->i31->type.isNullable()) {
+        parent.implicitTrap = true;
+      }
```
tlively, line 660: As a follow-up, perhaps we should do similar checks in `visitCallRef`. I see we already do it for the struct and array instructions.

tlively (approved):

PR #4827, opened 2022-07-25
[GUFA] Ignore trapping children when unreachable

aheejin:
In `getDroppedChildrenAndAppend`, when the type of the given last
expression is unreachable, we [ignore](https://github.com/WebAssembly/binaryen/blob/c40b402aaffa24c641cbb0e8efd0019c5c4a68b6/src/ir/drop.cpp#L46-L49) trapping effects when deciding
whether to drop the current expression because the last expression will
trap anyway.

We can do the same thing when dropping children. This will make GUFA's
generated code a little simpler in some cases. In GUFA's case these are
supposed to be optimized away by DCE anyway, but I think it's good to
handle this in this utility function too.

### DIFF
--- a/src/ir/drop.cpp
+++ b/src/ir/drop.cpp
@@ -64,7 +64,12 @@ Expression* getDroppedChildrenAndAppend(Expression* curr,
 
   std::vector<Expression*> contents;
   for (auto* child : ChildIterator(curr)) {
-    if (!EffectAnalyzer(options, wasm, child).hasUnremovableSideEffects()) {
+    EffectAnalyzer effects(options, wasm, child);
+    // Ignore a trap, as the unreachable replacement would trap too.
+    if (last->type == Type::unreachable) {
+      effects.trap = false;
+    }
+    if (!effects.hasUnremovableSideEffects()) {
       continue;
     }
     if (child->type.isConcrete()) {
--- a/test/lit/passes/gufa-refs.wast
+++ b/test/lit/passes/gufa-refs.wast
@@ -35,10 +35,6 @@
   ;; CHECK-NEXT:   (block
   ;; CHECK-NEXT:    (drop
   ;; CHECK-NEXT:     (loop $loop
-  ;; CHECK-NEXT:      (block
-  ;; CHECK-NEXT:       (unreachable)
-  ;; CHECK-NEXT:       (unreachable)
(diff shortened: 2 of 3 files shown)


src/ir/drop.cpp
```diff
@@ -64,7 +64,12 @@ Expression* getDroppedChildrenAndAppend(Expression* curr,
 
   std::vector<Expression*> contents;
   for (auto* child : ChildIterator(curr)) {
-    if (!EffectAnalyzer(options, wasm, child).hasUnremovableSideEffects()) {
+    EffectAnalyzer effects(options, wasm, child);
+    // Ignore a trap, as the unreachable replacement would trap too.
+    if (last->type == Type::unreachable) {
```
kripken, line 69: Interesting idea, but I think there are some corner cases here:

* We should check `last->is<Unreachable>()` as e.g. `(br $foo)` would have type unreachable but not trap.
* We can reorder traps but not with other effects. For example if child0 has a possible trap, child1 does a call, and then we append a trap, we can't remove child0 - we need to run that trap before doing the call, which might have noticeable effects.

I think this can still be done, but I'm not sure how much more complex handling that last situation is?

aheejin: Hmm I see, it's more complicated that I expected..

> * We should check `last->is<Unreachable>()` as e.g. `(br $foo)` would have type unreachable but not trap.

The is what we are already doing, which checks only if the type is unreachable, fine? https://github.com/WebAssembly/binaryen/blob/c40b402aaffa24c641cbb0e8efd0019c5c4a68b6/src/ir/drop.cpp#L46-L49

> * We can reorder traps but not with other effects. For example if child0 has a possible trap, child1 does a call, and then we append a trap, we can't remove child0 - we need to run that trap before doing the call, which might have noticeable effects.
> I think this can still be done, but I'm not sure how much more complex handling that last situation is?

I see, this is tricky... This probably is not worth it then. Thanks.

kripken: > The is what we are already doing, which checks only if the type is unreachable, fine? 

Good point, yes, that code doesn't look correct. It should check `is<Unreachable>()`. Do you want to open a PR?

aheejin: Opened #4834.

PR #4829, opened 2022-07-26
Make `GlobalTypeRewriter` work for isorecursive types

tlively:
There are two new potential problems that `GlobalTypeRewriter` can run into when
working with isorecursive types instead of nominal types. First, the refined
types may have replaced generic references with references to specific other
types, potentially creating new recursions and making the existing recursion
groups insufficient. Second, distinct types may be refined to structurally
identical types and those distinct input types may map the same output type,
potentially changing cast behavior.

Both of these problems are solved by putting all the new types in a single large
recursion group.

We do not currently account for the fact that types may be used in the external
interface of the module, but when we do, externalized types will be excluded
from optimizations and will not be affected by the creation of this single large
rec group.

Fixes https://github.com/WebAssembly/binaryen/issues/4816.


tlively:
* main
  * **PR #4829** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4829" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4829?utm_source=stack-comment).

tlively:
It's literally a one-line fix :cry:

kripken (approved):

PR #4830, opened 2022-07-26
Changing ref maps in wasm-binary to use a value of a vector of Name*

ashleynh:
wasm-binary.h/.cpp has maps of Index to Expression* to link already parsed global/function/table expression indexes with the corresponding name for the global/function/table. This structure assumes that each expression type will need only a single reference to the corresponding global/function/table, but this breaks down with the introduction of multi-memories as the MemoryCopy instruction has a pointer to the source memory and a pointer to the destination memory.

This change provides a mechanism for a single expression to have multiple Name* pointers set. That said, we lose the possibility of hitting WASM_UNREACHABLE, as we won't know the Expression type. In practice, this seems a difficult condition to reach?

### DIFF
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -4284,8 +4284,8 @@ void WasmBinaryBuilder::visitCall(Call* curr) {
     curr->operands[num - i - 1] = popNonVoidExpression();
   }
   curr->type = sig.results;
-  functionRefs[index].push_back(
-    &curr->target); // we don't know function names yet
+  // We don't know function names yet.
+  functionRefs[index].push_back(&curr->target);
   curr->finalize();
 }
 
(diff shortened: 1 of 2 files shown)


src/wasm/wasm-binary.cpp
```diff
@@ -4308,7 +4284,8 @@ void WasmBinaryBuilder::visitCall(Call* curr) {
     curr->operands[num - i - 1] = popNonVoidExpression();
   }
   curr->type = sig.results;
-  functionRefs[index].push_back(curr); // we don't know function names yet
+  functionRefs[index].push_back(
+    &curr->target); // we don't know function names yet
```
tlively, line 4288:
```suggestion
  // We don't know function names yet.
  functionRefs[index].push_back(&curr->target);
```

tlively (approved):
Nice simplification!

PR #4834, opened 2022-07-26
Fix unreachable handling in getDroppedChildrenAndAppend

aheejin:
The previous code assumes if `last`'s type is unreachable it traps. But
it's not always the case because it can be other instructions like `br`
whose type is unreachable but doesn't necessarily trap.
Context:
https://github.com/WebAssembly/binaryen/pull/4827#discussion_r929395477

Not sure how I can add a test for this, because the `last` argument of
this function is currently only used in GUFA, and we only add
`unreachable` for it, so to add a test case, I think we need a separate
test pass or something to test this. I'm not sure if it's worth it, but
please let me know if you think it does.


kripken (approved):
Good point about testing, yeah, we can't really test it now. And that's probably why the bug wasn't noticed. Still very good to fix for later.

PR #4838, opened 2022-07-27
Fuzzing: Add support for hybrid mode

kripken:
The fuzzer looks stable on hybrid after the recent fix. However, do not switch it over
yet, until most users use it (so that we keep testing the actually-used mode for now).


scripts/fuzz_opt.py
```diff
@@ -35,6 +35,7 @@
 
 # parameters
 
+# TODO: switch to --hybrid when that becomes the most common mode for users
 TYPE_SYSTEM_FLAG = '--nominal'
```
tlively, line 39: Can we make the fuzzer randomize this so we get coverage of both modes right away? That would be useful in the longer term as well if we keep the `--nominal` option around.

src/tools/execution-results.h
```diff
@@ -151,10 +151,11 @@ struct ExecutionResults {
   bool areEqual(Literal a, Literal b) {
     // We allow nulls to have different types (as they compare equal regardless)
     // but anything else must have an identical type.
-    // We cannot do this in nominal typing, however, as different modules will
+    // We cannot do this in nominal/hybrid, however, as different modules will
```
tlively, line 154: I don't think this comment applies to isorecursive typing. If you round trip with isorecursive types, you get the same types out the other end.

kripken: In that case there are more bugs to fix in hybrid mode. Here is one:
```wat
(module
 (type $f32_i64_v128_=>_dataref (func_subtype (param f32 i64 v128) (result dataref) func))
 (type $[i8] (array_subtype i8 data))
 (type $f32_=>_f32 (func_subtype (param f32) (result f32) func))
 (type $f64_=>_f64 (func_subtype (param f64) (result f64) func))
 (export "func_9" (func $0))
 (func $0 (type $f32_i64_v128_=>_dataref) (param $0 f32) (param $1 i64) (param $2 v128) (result dataref)
  (local.set $0
   (local.get $0)
  )
  (array.init_static $[i8])
 )
)
```
STR:
```
wasm-opt a.wat  --gto  -all --hybrid --fuzz-exec
[fuzz-exec] calling func_9
[fuzz-exec] note result: func_9 => dataref
[fuzz-exec] calling func_9
[fuzz-exec] note result: func_9 => dataref
[fuzz-exec] comparing func_9
types not identical! [ref [rtt (rtt 0 $array.0)] ()] != [ref [rtt (rtt 0 $array.0)] ()]
[fuzz-exec] optimization passes changed results
```

tlively: That's not a bug due to `--gto` changing the types. Since we do want to fuzz with various optimizations that will change the types, I don't think we need to make any changes to the code here, but we could be more precise in the comment.

PR #4839, opened 2022-07-27
[GUFA] Handle GUFA + Intrinsics

kripken:
Like RemoveUnusedModuleElements, places that build graphs of function
reachability must special-case the call-without-effects intrinsic. Without that,
it looks like a call to an import. Normally a call to an import is fine - it makes us
be super-pessimistic, as we think things escape all the way out - but in GC
for now we are assuming a closed world, and so we end up broken. To fix that,
properly handle the intrinsic case.

### DIFF
--- a/src/ir/possible-contents.cpp
+++ b/src/ir/possible-contents.cpp
@@ -554,8 +554,11 @@ struct InfoCollector
       handleDirectCall(curr, refFunc->func);
     } else {
       // We can't see where this goes. We must be pessimistic and assume it
-      // can call anything of the proper type, the same as a CallRef. Do that
-      // by reusing the CallRef code.
+      // can call anything of the proper type, the same as a CallRef. (We could
+      // look at the possible contents of |target| during the flow, but that
+      // would require special logic like we have for RefCast etc., and the
+      // intrinsics will be lowered away anyhow, so just running after that is
+      // a workaround.)
       handleIndirectCall(curr, target->type);
     }
   }
--- a/src/ir/possible-contents.cpp
+++ b/src/ir/possible-contents.cpp
@@ -544,7 +544,9 @@ struct InfoCollector
     // handle this automatically by the reference flowing out to an import,
     // which is what binaryen intrinsics look like. For now, to support use
     // cases of a closed world but that also use this intrinsic, handle the
-    // intrinsic specifically here.
+    // intrinsic specifically here. (Without that, the closed world assumption
+    // makes us ignore the function ref that flows to an import, so we are not
+    // aware that it is actually called.)
     auto* target = curr->operands.back();
     if (auto* refFunc = target->dynCast<RefFunc>()) {
       // We can see exactly where this goes.
--- a/src/ir/possible-contents.cpp
+++ b/src/ir/possible-contents.cpp
@@ -503,7 +503,41 @@ struct InfoCollector
   // Calls send values to params in their possible targets, and receive
   // results.
   void visitCall(Call* curr) {
-    auto* target = getModule()->getFunction(curr->target);
+    auto& wasm = *getModule();
+    Name targetName;
+    if (!Intrinsics(wasm).isCallWithoutEffects(curr)) {
+      // This is just a normal call.
+      targetName = curr->target;
+    } else {
+      // A call-without-effects receives a function reference and calls it, the
+      // same as a CallRef. When we have a flag for non-closed-world, we should
+      // handle this automatically by the reference flowing out to an import,
+      // which is what binaryen intrinsics look like. For now, to support use
+      // cases of a closed world but that also use this intrinsic, handle the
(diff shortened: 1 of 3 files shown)


src/ir/possible-contents.cpp
```diff
@@ -525,21 +524,53 @@ struct InfoCollector
...
+      // This is just a normal call.
+      handleDirectCall(curr, curr->target);
+      return;
+    }
+    // A call-without-effects receives a function reference and calls it, the
+    // same as a CallRef. When we have a flag for non-closed-world, we should
+    // handle this automatically by the reference flowing out to an import,
+    // which is what binaryen intrinsics look like. For now, to support use
+    // cases of a closed world but that also use this intrinsic, handle the
+    // intrinsic specifically here.
```
aheejin, line 547: > When we have a flag for non-closed-world,

What is a flag for non-closed world?

> we should handle this automatically by the reference flowing out to an import,

What does "by the reference flowing out to an import"? Does this mean you call an imported function? Why is that "flowing out"?

> which is what binaryen intrinsics look like.

What does this mean? Aren't we handling binaryen intrinsics here..? Are there two kinds of intrinsics?

src/ir/possible-contents.cpp
```diff
@@ -525,21 +524,53 @@ struct InfoCollector
...
+    if (auto* refFunc = target->dynCast<RefFunc>()) {
+      // We can see exactly where this goes.
+      handleDirectCall(curr, refFunc->func);
+    } else {
+      // We can't see where this goes. We must be pessimistic and assume it
+      // can call anything of the proper type, the same as a CallRef. (We could
+      // look at the possible contents of |target| during the flow, but that
+      // would require special logic like we have for RefCast etc., and the
+      // intrinsics will be lowered away anyhow, so just running after that is
+      // a workaround.)
```
aheejin, line 558: > the intrinsics will be lowered away anyhow,

Where are the intrinsics lowered away?

test/lit/passes/gufa.wast
```diff
@@ -920,3 +920,149 @@
...
+    ;; reference, so $target-keep's code is reachable. We should leave it alone,
+    ;; but we can put an unreachable in $target-drop.
+    (call $call-without-effects
+     (i32.const 1)
+     (ref.func $target-keep)
+    )
+    ;; The other import is not enough to keep $target-drop alive.
+    (call $other-import
+      (ref.func $target-drop)
+    )
```
aheejin, line 968: Can't `other-import` call `target-drop` using `call_ref`?

aheejin (commented):
Sorry I think I am missing some important context... 😢

src/ir/possible-contents.cpp
kripken: > What is a flag for non-closed world?

Right now in GC we assume a closed world, which is what all producers assume anyhow. That assumption makes it ok in e.g. GlobalTypeOptimization to optimize a type (remove a field, say) even if that type "escapes" to the outside. If the module were linked with another, such optimizations could be noticeable, so they are technically unsafe. We do them for now, but eventually we'll want to have a flag to opt-in to a closed world assumption, so the default behavior is safe. That will take work to find which types "escape" to the outside.

> What does "by the reference flowing out to an import"? Does this mean you call an imported function? Why is that "flowing out"?

Sorry, that wasn't clear. I meant, if a `ref.func` can reach the outside of the module - by being returned from an export, for example - then the function might be called from the outside. Normally we assume that just seeing a `ref.func` doesn't mean the contents of that function are used - we also need to see a call that might operate on it.

In a closed world, which we assume for now, not seeing such a call means it cannot be called. The problem is that a binaryen intrinsic is exactly that, a `ref.func` is sent to the outside of the module (sent to an import), which does the call there. We need to be aware that the function is called somewhere.

> Are there two kinds of intrinsics?

Sorry, that wasn't clear either I think. This is all about binaryen intrinsics, which is just the `call.without.effects` intrinsic really.

src/ir/possible-contents.cpp
kripken: The `--intrinsic-lowering` lowering pass does that.

test/lit/passes/gufa.wast
kripken: In theory, yes, but this is where the closed world assumption is used. We assume nothing actually happens outside the module. This is unsafe, but good enough for all producers for now (and we'll add a flag for it later).

src/ir/possible-contents.cpp
tlively: I think the key thing missing from the comment that would make it clear is an explanation of what goes wrong without this special handling.

tlively (approved):
LGTM with comment clarified.

kripken:
Sounds good, I improved the comment here and the existing one this is copied from.

PR #4848, opened 2022-07-29
Remove RTTs

tlively:
RTTs were removed from the GC spec and if they are added back in in the future,
they will be heap types rather than value types as in our implementation.
Updating our implementation to have RTTs be heap types would have been more work
than deleting them for questionable benefit since we don't know how long it will
be before they are specced again.

Work in progress because all the tests still need to be updated.


tlively:
* main
  * **PR #4848** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4848" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4848?utm_source=stack-comment).

scripts/fuzz_opt.py
```diff
@@ -1225,7 +1225,7 @@ def randomize_opt_flags():
                 print('avoiding --flatten due to multivalue + reference types not supporting it (spilling of non-nullable tuples)')
                 continue
             if '--gc' not in FEATURE_OPTS:
-                print('avoiding --flatten due to GC not supporting it (spilling of RTTs)')
+                print('avoiding --flatten due to GC not supporting it (spilling of non-nullable locals)')
```
kripken, line 1228: I don't think this will be true once we support 1a. Might be worth a TODO here for that.

src/ir/literal-utils.h
```diff
@@ -29,30 +29,8 @@ inline Expression* makeFromInt32(int32_t x, Type type, Module& wasm) {
   return ret;
 }
 
-inline bool canMakeZero(Type type) {
```
kripken, line 32: I think it's best to keep this function as it shows a clear intent. We have pairs of "can make zero" and "make zero" in the code which look nice.

Also, we may add nondefaultable types to wasm, and this would help refactor for that.

src/literal.h
```diff
@@ -94,9 +78,9 @@ class Literal {
   explicit Literal(const std::array<Literal, 8>&);
   explicit Literal(const std::array<Literal, 4>&);
   explicit Literal(const std::array<Literal, 2>&);
-  explicit Literal(Name func, Type type) : func(func), type(type) {}
+  explicit Literal(Name func, HeapType type)
```
kripken, line 81: the heap type change here could be split out perhaps?

src/tools/wasm-ctor-eval.cpp
```diff
@@ -633,7 +632,7 @@ EvalCtorOutcome evalCtor(EvallingModuleRunner& instance,
   Literals params;
   for (Index i = 0; i < func->getNumParams(); i++) {
     auto type = func->getLocalType(i);
-    if (!LiteralUtils::canMakeZero(type)) {
+    if (LiteralUtils::canMakeZero(type)) {
```
kripken, line 635: This looks wrong?

test/let.wasm.fromBinary
```diff
@@ -10,15 +10,14 @@
   (drop
```
kripken, line 10: Separately from this PR, since let was removed from the spec we could remove this test and the binary reading code for it.

test/lit/passes/dae-gc.wast
```diff
@@ -69,30 +69,29 @@
    (unreachable)
   )
  )
- ;; a function that gets an rtt that is never used. we cannot create a local for
+ ;; a function that gets an non-nullable reference that is never used. we cannot create a local for
```
kripken, line 72:
```suggestion
 ;; a function that gets a non-nullable reference that is never used. we cannot create a local for
```

test/lit/passes/inlining_splitting.wast
```diff
@@ -203,18 +201,8 @@
...
-  ;; CHECK-NEXT:  )
-  ;; CHECK-NEXT:  (loop $l
-  ;; CHECK-NEXT:   (call $import)
-  ;; CHECK-NEXT:   (br $l)
-  ;; CHECK-NEXT:  )
-  ;; CHECK-NEXT: )
-  (func $nondefaultable-param (param $x i32) (param $y (rtt $struct))
-    ;; The RTT param here prevents us from even being inlined, even with
+  (func $nondefaultable-param (param $x i32) (param $y (ref $struct))
+    ;; The non-nullable param here prevents us from even being inlined, even with
```
kripken, line 205: Looks like it is inlined just below... which is ok if we fix up the type to be nullable, which we do. So just the comment should be fixed here. Maybe check if there is already a test for non-nullable locals though in this file, this might be deletable.

test/passes/Oz_fuzz-exec_all-features.wast
```diff
@@ -77,91 +72,23 @@
...
         ;; first, try to cast our simple $struct to an extended, which will fail
-        (br_on_cast $extendedblock
-         (local.get $any) (rtt.canon $extendedstruct)
+        (br_on_cast_static $extendedblock $extendedstruct
+         (local.get $any)
         )
-        (rtt.canon $struct)
        )
       )
       (call $log (i32.const -1)) ;; we should never get here
```
kripken, line 94: This line is reached now based on the updated output.

kripken (commented):
Github seems unable to open the `heap2local.wat` test (!) in any browser I try, so I can't add a comment there. But reading `func $nondefaultable` there, we *do* optimize that case, so the new comment is incorrect. We do fix up locals to be nullable, which is fine.

tlively:
FWIW I fuzzed this and after about 20k iterations the fuzzer had some issue with a data segment being out of bounds, which didn't seem related to this at all.

src/passes/StackIR.cpp
```diff
@@ -90,15 +90,17 @@ class StackIROptimizer {
...
     //       (nop)
     //       (local.get)
     //     )
     //     (i32)
     //   )
     //
-    // However, that is not valid as an rtt cannot be set to a local.
+    // However, that is not valid as an non-nullable reference cannot be set to
+    // a local. TODO: double check that this is still true now that we don't
+    // have RTTs.
```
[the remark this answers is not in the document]
tlively: Sounds good. Tracking in #4862, but this should be an easy follow-up as well.

src/tools/wasm-ctor-eval.cpp
tlively: Oops! A bug from restoring `canMakeZero`.

test/let.wasm.fromBinary
tlively: Ah, I actually did remove the other files for this test in this PR because they used RTTs and deleting them was easier than regenerating them. I missed this one, but I'll remove it, too.

test/lit/passes/inlining_splitting.wast
tlively: Good catch. It looks like this gets inlined to use a `(ref null $struct)` local, but I think we could even use a non-nullable `(ref $struct)` local once we handle those well. Does that need to be tracked separately or will it be handled by your work?

kripken:
Can you get a testcase for that fuzz bug? Data segment out of bounds doesn't sound like a known issue.

test/lit/passes/inlining_splitting.wast
kripken: Yes, that will work automatically once we have 1a support. We'll see that the local is in proper 1a position and keep it non-nullable.

tlively:
Fuzz bug investigated and tracked in #4865.

scripts/fuzz_shell.js
```diff
@@ -174,8 +174,12 @@ var imports = {
 
 imports = Asyncify.instrumentImports(imports);
 
+var module = new WebAssembly.Module(binary);
+
+console.log('module ok');
```
kripken, line 179: Is this meant to be committed?

src/passes/OptimizeInstructions.cpp
```diff
@@ -1767,9 +1766,6 @@ struct OptimizeInstructions
       // the type received by our parent.
       std::vector<Expression*> items;
       items.push_back(builder.makeDrop(curr->ref));
-      if (curr->rtt) {
-        items.push_back(builder.makeDrop(curr->rtt));
-      }
       items.push_back(builder.makeRefNull(intendedType));
```
kripken, line 1769: We can just create a block here directly. In fact since it's just 2 items we can use `builder.makeSequence`.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1797,9 +1792,6 @@ struct OptimizeInstructions
         // types for other passes.
         std::vector<Expression*> items;
         items.push_back(builder.makeDrop(curr->ref));
-        if (curr->rtt) {
-          items.push_back(builder.makeDrop(curr->rtt));
-        }
```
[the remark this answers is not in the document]
kripken: ah, sorry, like above this can use `makeSequence`

test/lit/passes/dae-gc.wast
```diff
@@ -69,30 +69,29 @@
    (unreachable)
   )
  )
- ;; a function that gets an rtt that is never used. we cannot create a local for
+ ;; a function that gets a non-nullable reference that is never used. we cannot create a local for
  ;; that parameter, as it is not defaultable, so do not remove the parameter.
- ;; CHECK:      (func $get-rtt (param $0 (rtt ${}))
+ ;; TODO: We should be able to create a local for it after #4824 is resolved.
```
kripken, line 74: it looks from the output like we do optimize here. the argument is removed and a local is added, as nullable.

test/passes/Oz_fuzz-exec_all-features.txt
```diff
@@ -530,10 +429,6 @@
 [LoggingExternalInterface logging 99]
 [LoggingExternalInterface logging 0]
 [LoggingExternalInterface logging 10]
-[fuzz-exec] calling rtt_Fresh
-[LoggingExternalInterface logging 1]
-[LoggingExternalInterface logging 0]
-[LoggingExternalInterface logging 1]
 [fuzz-exec] calling array.init
 [LoggingExternalInterface logging 2]
 [LoggingExternalInterface logging 42]
```
kripken, line 434: Hmm, nice, the last result in `static-casts` was, before this PR, 0 that become 1 after opts - which was an RTT bug I guess. But after the change to the test it is the same after opts as it should be...

kripken (approved):
lgtm % comments

scripts/fuzz_shell.js
tlively: Oh, no, this was from my investigations of that fuzz bug...

test/passes/Oz_fuzz-exec_all-features.txt
tlively: We should make this a lit test to make the output easier to match to the input :grimacing:

scripts/gen-s-parser.py
```diff
@@ -584,22 +582,14 @@
...
     ("struct.new_default",   "makeStructNewStatic(s, true)"),
     ("struct.get",           "makeStructGet(s)"),
     ("struct.get_s",         "makeStructGet(s, true)"),
     ("struct.get_u",         "makeStructGet(s, false)"),
     ("struct.set",           "makeStructSet(s)"),
-    ("array.new_with_rtt",   "makeArrayNew(s, false)"),
-    ("array.new_default_with_rtt", "makeArrayNew(s, true)"),
     ("array.new",            "makeArrayNewStatic(s, false)"),
     ("array.new_default",    "makeArrayNewStatic(s, true)"),
-    ("array.init",           "makeArrayInit(s)"),
```
MaxGraey, line 602: Hmm `array.init` was remved from parser but still exists in IR and C/JS-API. Maybe it was removed by accident?

tlively: It remains in the IR because array.init_static still exists, but maybe the APIs could be cleaned up more.

MaxGraey: got it, thanks!

PR #4851, opened 2022-07-29
Add interpreter support for intrinsics

kripken:
Adding that is trivial (just get the function target and call it), but it did require
some small refactoring to get an instance reference to call. `callImport` now
mirrors `callTable` which also receives an instance reference.

This can give us some chance to catch bugs like #4839 in the fuzzer.


src/tools/execution-results.h
```diff
@@ -75,7 +78,17 @@ struct LoggingExternalInterface : public ShellExternalInterface {
       } else if (import->base == "getTempRet0") {
         return {Literal(state.tempRet0)};
       }
+    } else if (import->module == Intrinsics::BinaryenIntrinsics) {
```
tlively, line 81: Should this be handled in the core interpreter instead of in the external interface? Since it's intrinsic, I would expect the core interpreter to know about it and be able to handle it.

kripken: Yeah, as an intrinsic that seems to make sense. OTOH the mechanism is an import call, and those are not handled in the core interpreter. For now I think avoiding import handling in the core interpreter is best. Later if we refactor intrinsics to something nicer without an import then the core interpreter would be the right place, I agree.

src/ir/intrinsics.h
```diff
@@ -35,6 +35,9 @@ class Intrinsics {
   Module& module;
 
 public:
+  static Name BinaryenIntrinsics;
```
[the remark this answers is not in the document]
kripken: Good idea, done.

src/tools/execution-results.h
tlively: Agree we shouldn't try to handle real imports in the core interpreter, but I don't think we should consider this a real import. Given we need to special case intrinsics either way, is there any benefit to not handling them in the core interpreter? I don't know if we have any optimizations that interpret across call boundaries, but in principle handling this in core would make those optimizations more robust.

kripken: You're right that they could benefit in theory, but no existing pass (or planned one AFAIK) does that.

The benefit to not adding import handling in the core interpreter is just to keep it a little simpler for now.

tlively: I still think it's simpler to handle this in the core interpreter, but if you feel strongly about it that's ok with me.

kripken: I don't feel strongly. Changed to how you prefer.

src/wasm-interpreter.h
```diff
@@ -2758,7 +2759,17 @@ class ModuleRunnerBase : public ExpressionRunner<SubType> {
     auto* func = wasm.getFunction(curr->target);
     Flow ret;
     if (func->imported()) {
-      ret.values = externalInterface->callImport(func, arguments);
+      // The call.without.effects intrinsic is a call to an import that actually
+      // calls the given function reference that is the final argument.
+      if (func->module == Intrinsics::BinaryenIntrinsicsModule &&
+          func->base == Intrinsics::CallWithoutEffects) {
```
tlively, line 2765: It looks like we could use `Intrinsics::isCallWithoutEffects` for this check, and we could do it even before checking `if (func->imported())`. That would let us leave the `func->imported()` case unmodified.

tlively (approved):

src/wasm-interpreter.h
kripken: Sounds good, done.

PR #4852, opened 2022-07-29
Change stack pass to use existing global variables.

eyebrowsoffire:
This partially fixes https://github.com/emscripten-core/emscripten/issues/16496

With this change, toolchains can specify `__stack_base` and `__stack_end`
global variables which the binaryen stack pass uses directly to check against
overflows, instead of the pass emitting its own variables and setter (which
previously was `__set_stack_limits`, and toolchains had to invoke this setter).

If the globals are not declared, the stack pass errors out.

I am leaving `__set_stack_limits` in as a no-op in order to not break emscripten. Once emscripten gets rid of its invocations of `__set_stack_limits` I can follow up and remove that function completely.


src/passes/StackCheck.cpp
```diff
@@ -130,9 +119,19 @@ struct StackCheck : public Pass {
       return;
     }
 
-    // Pick appropriate names.
-    auto stackBaseName = Names::getValidGlobalName(*module, "__stack_base");
-    auto stackLimitName = Names::getValidGlobalName(*module, "__stack_limit");
+    // __stack_base and __stack_end are created by the toolchain (emscripten)
+    // If we don't find these globals, skip this pass.
```
kripken, line 123: Skipping seems risky. How about erroring?

eyebrowsoffire: Sounds good, I'll change that.

test/lit/passes/stack-check-memory64.wast
```diff
@@ -1,33 +1,32 @@
 ;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
 ;; NOTE: This test was ported using port_test.py and could be cleaned up.
 
-;; RUN: foreach %s %t wasm-opt --stack-check --enable-memory64 -S -o - | filecheck %s
+;; RUN: wasm-opt %s --stack-check --enable-memory64 -S -o - | filecheck %s --dump-input fail
```
kripken, line 4: what is `dump-input-fail` and why do we need it now?

kripken (commented):
In general this seems ok to me, but I am not sure enough about the LLVM side to know if it always emits the name section, and if we expect that to continue in the future as well. I also am not sure what to look for in the lld tests here. @sbc100 PTAL at that question and the lld tests, the rest lgtm

src/passes/StackCheck.cpp
```diff
@@ -130,9 +119,17 @@ struct StackCheck : public Pass {
...
+    // If we don't find these globals, error out.
+    auto stackBase = module->getGlobalOrNull("__stack_base");
+    if (!stackBase) {
+      Fatal() << "cannot do stack checking without __stack_base variable\n";
+    }
+
+    auto stackEnd = module->getGlobalOrNull("__stack_end");
+    if (!stackEnd) {
+      Fatal() << "cannot do stack checking without __stack_end variable\n";
+    }
```
sbc100, line 132: IIRC we decided that we should not add dependency the name section, but instead using the import/export names to find these?

I think we need to check for both exports (in the case of normal/main modules) and imports (in the case of side modules).

e.g.
```
auto stackEnd = module->getExportOrNull("__stack_end");
if (!stackEnd) {
   // somehow look for a global that was imported using the name `__stack_end`
}
```

Out of interest, does this current approach pass all the emscripten tests? i.e. are we already always including a name section?

test/lit/passes/stack-check_enable-mutable-globals-fail.wast
```diff
@@ -0,0 +1,15 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
```
sbc100, line 1: Shouldn't this test be called something like `stack-check-missing-globals`?

sbc100, line 1: Is this first line of the comment true for this test?

test/lit/passes/stack-check_enable-mutable-globals.wast
```diff
@@ -0,0 +1,45 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
```
sbc100, line 1: The name of this file doesn't need to include the full string `enable-mutable-globals` anymore.   Perhaps just call it `stack-check-mutable-globals`?

src/passes/StackCheck.cpp
sbc100: It looks like we don't even generate a name section under all circumstances: https://github.com/emscripten-core/emscripten/blob/0e75ea8f916f9b4baaa35a17824606836ffc0dc1/tools/building.py#L286-L292

We could add STACK_CHECK to this condition, but maybe better to use import/export names?

src/passes/StackCheck.cpp
```diff
@@ -145,34 +142,23 @@ struct StackCheck : public Pass {
...
     PassRunner innerRunner(module);
-    EnforceStackLimits(stackPointer, stackBase, stackLimit, builder, handler)
+    EnforceStackLimits(stackPointer, stackBase, stackEnd, builder, handler)
       .run(&innerRunner, module);
 
-    // Generate the exported function.
+    // TODO: Remove this function once emscripten stops calling it.
+    // This function is now a no-op, but it still needs to be here until
+    // emscripten is changed.
+    wasm::Name limitsFuncName = "__set_stack_limits";
```
sbc100, line 153: Can you put this back up the top to avoid churn?  (i.e. just add the comment to existing static?)

src/passes/StackCheck.cpp
```diff
@@ -145,34 +142,23 @@ struct StackCheck : public Pass {
...
-    auto* getBase = builder.makeLocalGet(0, stackPointer->type);
-    auto* storeBase = builder.makeGlobalSet(stackBaseName, getBase);
-    auto* getLimit = builder.makeLocalGet(1, stackPointer->type);
-    auto* storeLimit = builder.makeGlobalSet(stackLimitName, getLimit);
-    limitsFunc->body = builder.makeBlock({storeBase, storeLimit});
-    addExportedFunction(*module, std::move(limitsFunc));
+    limitsFunc->body = builder.makeBlock({});
+    module->addFunction(std::move(limitsFunc));
+    module->addExport(Builder::makeExport(
+      limitsFuncName, limitsFuncName, ExternalKind::Function));
```
sbc100, line 161: Can we just keep the existing `addExportedFunction` for now?

sbc100 (commented):
This is looking great!  .. we need to start exporting those new globals from emscripten before we can land it though right?

Issue #4772, opened 2022-07-06
asyncify - option to use multiple memories


wanderer:
There are a few implementation of the nascent multiple memory proposal now. Asyncify could benefit from using a dedicated memory, which would simplify the api since you wouldn't need to specify the memory offset where unwind data lives.

tlively:
cc @ashleynh

L-jasmine:
I did it https://github.com/WebAssembly/binaryen/pull/4853

PR #4853, opened 2022-07-30
asyncify-option with multi-memories

L-jasmine:
Hi @ashleynh . I implement the issue https://github.com/WebAssembly/binaryen/issues/4772 . 
Please review.

### DIFF
--- a/src/ir/memory-utils.h
+++ b/src/ir/memory-utils.h
@@ -42,6 +42,14 @@ inline void ensureExists(Module* wasm) {
   }
 }
 
+inline void newOne(Module* wasm) {
+  auto memoriesSize = wasm->memories.size();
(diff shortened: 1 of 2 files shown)


kripken:
Good start, thanks for the PR @L-jasmine !

This looks like the right direction. One thing to think about here though is multiple buffers. Current Asyncify allows you to pause more than once and swap between them, and that is how Fibers support in Emscripten works today. Each buffer has its own stack, and multiple of them can be alive at once. IIUC in this PR there is a single buffer in the new memory, so only one can be alive at once. To handle more we'd need some kind of memory management in the new memory. I'm not sure how best to do that.

L-jasmine:
OK. We'll wait for you.

PR #4854, opened 2022-07-31
Add DevContainer

cyw3:
With DevContainer, developers can join the project without deploy the build env.


tlively:
Is this a GitHub feature or some external service?

cyw3:
> Is this a GitHub feature or some external service?

This is the function of a plugin `ms-vscode-remote.remote-containers` for VSCode.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,36 @@
+
+FROM centos:centos7.9.2009
```
tlively, line 2: Why centos rather than something like Ubuntu?

tlively (commented):
I'd be fine landing something like this. I don't see a `pip3 install -r requirements-dev.txt` run line, though. Is this environment enough to both develop and run tests successfully?

PR #4856, opened 2022-08-01
[C API] Add BinaryenGetGlobalIndex method

ApsarasX:


kripken (commented):
Comparing to functions, we have `GetFunction` for the name and `GetFunctionByIndex` for the index. So I think we don't need `ByName` here, that is the default.

kripken:
Also this should add some testing (probably in `test/example/kitchen-sink` alongside similar methods.

dcodeIO:
I also think that `BinaryenGetGlobalIndexByName` might not be necessary. There are both `BinaryenGetGlobal` (by name) and `BinaryenGetGlobalByIndex` already, where the latter is typically used along `BinaryenGetNumGlobals` to iterate. Seems risky to depend on the index otherwise?

kripken:
@dcodeIO Good point, yes, it does seem like GetGlobalIndex could be implemented using the existing APIs (do a loop on the number of globals, get name by index, return if it's that one). I tend to agree that we don't need a new API, then. @ApsarasX does that make sense, or are we missing something?

ApsarasX:
@kripken You are right, I will close this PR.

PR #4860, opened 2022-08-02
Fix a new compiler warning

kripken:
Avoids a "may fall through" warning.


tlively (approved):
Should the assertion just be a `WASM_UNREACHABLE`?

kripken:
It does feel like an assert though, as it checks an internal invariant... the only problem is that we need to tell the compiler there is no fallthrough there. In any case, in a release build this should be emitted optimally I think.

PR #4861, opened 2022-08-02
Re-run scripts/test/generate_lld_tests.py. NFC

sbc100:


scripts/test/generate_lld_tests.py
```diff
@@ -57,13 +57,12 @@ def generate_wat_files(llvm_bin, emscripten_root):
...
             '-Xclang', '-nobuiltininc',
             '-Xclang', '-nostdsysteminc',
-            '-Xclang', '-I%s/system/include' % emscripten_root,
+            '-Xclang', '-I%s/include' % emscripten_sysroot,
             '-O1',
         ]
 
         link_cmd = [
             os.path.join(llvm_bin, 'wasm-ld'), '-flavor', 'wasm',
-            '-z', '-stack-size=1048576',
```
sbc100, line 66: There was a typo on this line. The leading `-` meant that the linker was ignoring this option, but newer versions of wasm-ld fail with an error.    

I'm not sure why this was added, but removing it keeps the test expectations unchanged in their stack size.  Fixing the option by removing the leading `-` will cause all the tests to have larger stack sizes..

tlively (approved):
Are there any specific functional changes we're looking for here, or is it generally just to keep things up to date?

sbc100:
> Are there any specific functional changes we're looking for here, or is it generally just to keep things up to date?

Just keeping things up-to-date.  (I want to make some other changes and an up-to-date basline is useful).

PR #4869, opened 2022-08-03
[NFC] Mark modifiesBinaryenIR=false in more places

kripken:
Atm this saves only a super-tiny amount of work. However, with 1a validation
it will save a lot more, since then we will be able to skip any 1a fixup work if
a pass does not modify the IR at all.

### DIFF
--- a/src/passes/OnceReduction.cpp
+++ b/src/passes/OnceReduction.cpp
@@ -84,6 +84,8 @@ struct OptInfo {
 struct Scanner : public WalkerPass<PostWalker<Scanner>> {
   bool isFunctionParallel() override { return true; }
 
+  bool modifiesBinaryenIR() override { return false; }
+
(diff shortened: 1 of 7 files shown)


src/passes/OnceReduction.cpp
```diff
@@ -84,6 +84,8 @@ struct OptInfo {
 struct Scanner : public WalkerPass<PostWalker<Scanner>> {
   bool isFunctionParallel() override { return true; }
 
+  bool modifiesBinaryenIR() override { return false; }
```
aheejin, line 87: Doesn't `OnceReduction` modify IR?

aheejin: On nevermind, this is `Scanner`...

aheejin (approved):

PR #4871, opened 2022-08-03
Cleanup em_asm/em_js strings as part of PostEmscripten

sbc100:
Rather than doing it as a side effect of dumping the metadata in
wasm-emscripten-finalize.


sbc100:
Depends on #4870

kripken:
To make sure I understand, this might regress some non-optimized builds, but that's probably ok? I mean, if we ran finalize but not the optimizer then previously we'd remove these, but after this PR we no longer would. That seems ok as before we wouldn't run finalize some of the time anyhow, and these are non-optimized builds, but I just want to make sure I understand the implications.

sbc100:
Yes, but remember that we often don't run finalize anyway these days, because we don't need it for metadata extraction anymore.

Basically this moves an optimization that was previously a side effect of metadata extraction in the PostEmscripten pass. 

For build that don't run wasm-emscripten-finalize this (i.e. builds that use WASM_BIGINT in emscripten) then I think this should actually be a win... will confirm that.

sbc100:
Indeed I confirmed that this is code size win in some cases.
Without this change:
```
$ ./emcc -O2 -sWASM_BIGINT test/hello_world_em_asm.c
$ ls -l a.out.wasm 
-rwxr-xr-x 1 sbc primarygroup 400 Aug  3 12:18 a.out.wasm
```

With this change:

```
$ ./emcc -O2 -sWASM_BIGINT test/hello_world_em_asm.c
$ ls -l a.out.wasm 
-rwxr-xr-x 1 sbc primarygroup 318 Aug  3 12:18 a.out.wasm
```

The reason is that with `-sWASM_BIGINT` we completely skip finalize, but we don't skip wasm-opt.

kripken (approved):
Great!

It looks like this adds more lines than it removes. Is that due to duplicating code from wasm-emscripten-finalize? (I guess the duplication would go away once we remove wasm-emscripten-finalize?)

Can we add a test for `--post-emscripten`'s new behavior here?

sbc100:
Yes, sorry I forgot to mention this duplicates some code, but that will be removed once #4863 lands

sbc100:
Added a test

test/lld/em_asm_pthread.wasm.out
```diff
@@ -146,27 +152,15 @@
...
-     (drop
-      (i32.const 1588)
-     )
-     (drop
-      (i32.const 0)
-     )
-     (drop
-      (i32.const 70)
-     )
+    (memory.init 2
```
kripken, line 155: why did these change?

sbc100: Because this tests wasm-emscripten-finalize, and this tools no longer does the removal the em_js/em_asm data which means it no longer modifies this part of the code and just leaves it as is.

The `--post-emscripten` pass now does this part.

PR #4874, opened 2022-08-04
Allow `BINARYEN_DEBUG` environment variable to be used in place of `--debug`. NFC

sbc100:
For example I found it useful to able to do something like this:

```
$ BINARYEN_DEBUG=post-emscripten ./test/runner sometest
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,7 +387,11 @@ the [`name-types` pass](https://github.com/WebAssembly/binaryen/blob/main/src/pa
 Some more notes:
 
  * See `bin/wasm-opt --help` for the full list of options and passes.
- * Passing `--debug` will emit some debugging info.
+ * Passing `--debug` will emit some debugging info.  Individual debug channels
+   (defined in the source code via `#define DEBUG_TYPE xxx`) can be enabled by
+   passing them as list of comma-separated strings.  For example: `bin/wasm-opt
+   --debug=binary`.  These debug channels can also be enabled via the
+   `BINARYEN_DEBUG` environment variable.
 
 ### wasm2js
 
--- a/src/support/command-line.cpp
+++ b/src/support/command-line.cpp
@@ -55,6 +55,10 @@ Options::Options(const std::string& command, const std::string& description)
   : debug(false), positional(Arguments::Zero) {
   std::string GeneralOption = "General options";
 
+  if (getenv("BINARYEN_DEBUG")) {
+    setDebugEnabled(getenv("BINARYEN_DEBUG"));
+  }
+
   add("--version",
       "",
       "Output version information and exit",


README.md
```diff
@@ -387,7 +387,11 @@ the [`name-types` pass](https://github.com/WebAssembly/binaryen/blob/main/src/pa
 Some more notes:
 
  * See `bin/wasm-opt --help` for the full list of options and passes.
- * Passing `--debug` will emit some debugging info.
+ * Passing `--debug` will emit some debugging info.  Individual debug channels
+   (defined in the source code via `#define DEBUG_TYPE xxx` can be enabled by
```
kripken, line 391:
```suggestion
   (defined in the source code via `#define DEBUG_TYPE xxx`) can be enabled by
```

README.md
```diff
@@ -387,7 +387,11 @@ the [`name-types` pass](https://github.com/WebAssembly/binaryen/blob/main/src/pa
...
+   passing them as list of comma separateed strings.  For example: `bin/wasm-opt
```
kripken, line 392:
```suggestion
   passing them as list of comma separated strings.  For example: `bin/wasm-opt
```

kripken (approved):

tlively (commented):
Huh, I didn't actually know about `--debug`. How does it differ from `BINARYEN_PASS_DEBUG`? Is there concern about that environment variable and this new one being confused?

sbc100:
> Huh, I didn't actually know about `--debug`. How does it differ from `BINARYEN_PASS_DEBUG`? Is there concern about that environment variable and this new one being confused?

`BINARYEN_PASS_DEBUG` looks like its limited in scope the pass manager itself where as `DEBUG_TYPE` and `BYN_TRACE` macros work like the `LLVM_DEBUG` in llvm.   They can by used to trace on any number of different channels.

We could probably make `BINARYEN_PASS_DEBUG` into a `BYN_TRACE` channel called passes such and just map `BINARYEN_PASS_DEBUG` to `BINARYEN_DEBUG=passes`?

sbc100:
> Huh, I didn't actually know about `--debug`. How does it differ from `BINARYEN_PASS_DEBUG`? Is there concern about that environment variable and this new one being confused?

`BINARYEN_PASS_DEBUG` looks like its limited in scope the pass manager itself where as `DEBUG_TYPE` and `BYN_TRACE` macros work like the `LLVM_DEBUG` in llvm.   They can by used to trace on any number of different channels.

We could probably make `BINARYEN_PASS_DEBUG` into a `BYN_TRACE` channel called passes such and just map `BINARYEN_PASS_DEBUG` to `BINARYEN_DEBUG=passes`?

sbc100:
OK to land this thomas?  We can look at trying to merge the two debug mechanism later perhaps?

tlively:
yep, lgtm.

PR #4876, opened 2022-08-05
wasm-emscripten-finalize: Remove em_js/em_asm start/stop symbols when stripping data segments.

sbc100:
This avoid a fatal crash in `--post-emscripten` where it tries to remove
data that is no longer part of the file.

This fixes bug introduced by #4871 that causes emscripten tests to fail.

### DIFF
--- a/src/wasm/wasm-emscripten.cpp
+++ b/src/wasm/wasm-emscripten.cpp
@@ -445,6 +445,15 @@ void EmscriptenGlueGenerator::separateDataSegments(Output* outfile,
     lastEnd = offset + seg->data.size();
   }
   wasm.dataSegments.clear();
+  // Remove the start/stop symbols that the PostEmscripten uses to remove
+  // em_asm/em_js data.  Since we just removed all the data segments from the
+  // file there is nothing more for that pass to do.
+  // TODO(sbc): Fix the ordering so that the removal the EM_ASM/EM_JS data comes
+  // before this pass.
+  wasm.removeExport("__start_em_asm");
+  wasm.removeExport("__stop_em_asm");
+  wasm.removeExport("__start_em_js");
+  wasm.removeExport("__stop_em_js");
 }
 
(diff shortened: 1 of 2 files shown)


sbc100:
I'm not quite sure how to go about fixing this in a better way.

Stripping of the em_asm/em_js is now an optional optimization, but writing of the memfile happens before any optimizations are run.

Perhaps the memfile creation should run after optimizations?   For now it runs as part of finalize before any optmizations.

kripken:
Can you explain a little more what's happening here? Sorry, I'm not understanding it yet.

> This avoid a fatal crash in --post-emscripten where it tries to remove data that is no longer part of the file.

What is the specific data it tries to remove? The segments? (If it's them, could we just check if they exist before trying to remove them?)

> Perhaps the memfile creation should run after optimizations?

How does the memfile tie into this?

sbc100:
> Can you explain a little more what's happening here? Sorry, I'm not understanding it yet.
> 
> > This avoid a fatal crash in --post-emscripten where it tries to remove data that is no longer part of the file.
> 
> What is the specific data it tries to remove? The segments? (If it's them, could we just check if they exist before trying to remove them?)

PostEmscripten assumes that if these symbol exist there is a corresponding data segment to match.  If this is not true it errors out.  We would make it more sloppy and just ignore the error when these symbols point to data that is missing... but I'd rather keep the error there as it seems more precise and would catch possible issues in the algorithm for mapping addresses to segments.

I'd rather have this pass remove these symbols, since it is removing the data they point to.

> 
> > Perhaps the memfile creation should run after optimizations?
> 
> How does the memfile tie into this?


This code implements the memfile stuff, it splits the data out of the wasm in into a file.   The problem is that that happens during wasm-emscripten-finalize, which occurs before other optimization passes (such as --post-emscripten).

kripken (approved):

PR #4878, opened 2022-08-05
SimplifyLocals: ReFinalize when needed

kripken:
Similar to #4877

### DIFF
--- a/src/passes/SimplifyLocals.cpp
+++ b/src/passes/SimplifyLocals.cpp
@@ -262,7 +262,7 @@ struct SimplifyLocals
         // We are replacing a local.get with the value of the local.set. That
         // may require a refinalize in certain cases, like this:
         //
-        //  (struct.get $X
+        //  (struct.get $X 0
         //    (local.get $x)
         //  )
         //
--- a/src/passes/SimplifyLocals.cpp
+++ b/src/passes/SimplifyLocals.cpp
@@ -53,6 +53,7 @@
 #include <ir/linear-execution.h>
 #include <ir/local-utils.h>
 #include <ir/manipulation.h>
+#include <ir/utils.h>
 #include <pass.h>
 #include <wasm-builder.h>
 #include <wasm-traversal.h>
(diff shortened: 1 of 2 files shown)


test/lit/passes/simplify-locals-gc.wast
```diff
@@ -214,4 +225,31 @@
...
+  ;; CHECK-NEXT: )
+  ;; NOMNL:      (func $needs-refinalize (type $ref|$B|_=>_anyref) (param $b (ref $B)) (result anyref)
+  ;; NOMNL-NEXT:  (local $a (ref null $A))
+  ;; NOMNL-NEXT:  (nop)
+  ;; NOMNL-NEXT:  (struct.get $B 0
+  ;; NOMNL-NEXT:   (local.get $b)
+  ;; NOMNL-NEXT:  )
+  ;; NOMNL-NEXT: )
+  (func $needs-refinalize (param $b (ref $B)) (result anyref)
+    (local $a (ref null $A))
```
aheejin, line 244:
```suggestion
    (local $a (ref $A))
```
Given that `$A` is already a supertype of `$B`, do we need this?

src/passes/SimplifyLocals.cpp
```diff
@@ -254,6 +258,23 @@ struct SimplifyLocals
       if (oneUse) {
         // with just one use, we can sink just the value
         this->replaceCurrent(set->value);
+
+        // We are replacing a local.get with the value of the local.set. That
+        // may require a refinalize in certain cases, like this:
+        //
+        //  (struct.get $X
```
aheejin, line 265:
```suggestion
        //  (struct.get $X 0
```

aheejin (approved):

test/lit/passes/simplify-locals-gc.wast
kripken: You're right that it isn't needed for the purpose of the test. But until we have 1a validation support, locals can't be non-nullable.

PR #4879, opened 2022-08-05
Move `auto_update_tests.py` into scripts directory. NFC

sbc100:
So it lives alongside `update_lit_checks.py` and `update_help_checks.py`

### DIFF
--- a/README.md
+++ b/README.md
@@ -566,9 +566,18 @@ The `check.py` script supports some options:
  * We have tests from upstream in `tests/spec`, in git submodules. Running
    `./check.py` should update those.
 
-Note that we are trying to gradually port the legacy wasm-opt tests to use `lit` and `filecheck` as we modify them.
-For `passes` tests that output wast, this can be done automatically with `scripts/port_passes_tests_to_lit.py` and for non-`passes` tests that output wast, see
-https://github.com/WebAssembly/binaryen/pull/4779 for an example of how to do a simple manual port.
+Note that we are trying to gradually port the legacy wasm-opt tests to use `lit`
+and `filecheck` as we modify them.  For `passes` tests that output wast, this
+can be done automatically with `scripts/port_passes_tests_to_lit.py` and for
+non-`passes` tests that output wast, see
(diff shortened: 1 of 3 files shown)


kripken:
Hmm, makes sense, but then maybe `check.py` also makes sense to move - then we'd have no scripts in the top level.

The current state before this PR also makes sense to me, of having just the 2 main scripts there: check and update (and all their helpers are in subdirs).

But I don't feel strongly here.

sbc100:
I think its ok to keep check.py at the top level for now.. but I also don't feel strongly.   @tlively what do you think?  Ok to land this as is or should I move check too?

tlively:
I don't feel too strongly. I think it would be fine to move auto_update_tests.py to scripts/ and keep check.py where it is. It would be great to document auto_update_tests.py in the README, though. Now that it's not in the top directory, it's less discoverable.

sbc100:
Updated the README to point to the two different auto-update scripts.

sbc100:
OK to land this now, with the documentations changes?

README.md
```diff
@@ -566,9 +566,18 @@ The `check.py` script supports some options:
...
-For `passes` tests that output wast, this can be done automatically with `scripts/port_passes_tests_to_lit.py` and for non-`passes` tests that output wast, see
-https://github.com/WebAssembly/binaryen/pull/4779 for an example of how to do a simple manual port.
+Note that we are trying to gradually port the legacy wasm-opt tests to use `lit`
+and `filecheck` as we modify them.  For `passes` tests that output wast, this
+can be done automatically with `scripts/port_passes_tests_to_lit.py` and for
+non-`passes` tests that output wast, see
+https://github.com/WebAssembly/binaryen/pull/4779 for an example of how to do a
+simple manual port.
+
+For lit test the test expectations (the CHECK lines) can often be automatically
```
tlively, line 576:
```suggestion
For lit tests the test expectations (the CHECK lines) can often be automatically
```

tlively (approved):
LGTM, thanks!

PR #4880, opened 2022-08-05
Validator: Validate intrinsics

kripken:
`call.without.effects` has a specific form, where the last parameter is a
function reference, and that function reference must have the right type
for the other parameters if called with them:

```wat
(call $call.without.effects
  (..i32..)
  (..f64..)
  (..function reference, which takes params i32 and f64..)
)
````

### DIFF
--- a/src/wasm/wasm-validator.cpp
+++ b/src/wasm/wasm-validator.cpp
@@ -820,7 +820,7 @@ void FunctionValidator::visitCall(Call* curr) {
     if (!curr->operands.empty()) {
       auto* target = curr->operands.back();
       // Ignore unreachable here, and leave other validation issues here for the
-      // global scope which already does this..
+      // global scope which already does this.
       if (target->type.isFunction()) {
         // Copy the original call and remove the reference. It must then match
         // the expected signature.
--- a/src/wasm/wasm-validator.cpp
+++ b/src/wasm/wasm-validator.cpp
@@ -443,15 +443,17 @@ private:
                  "return_call* requires tail calls to be enabled");
   }
 
+  // |printable| is the expression to print in case of an error. That may differ
+  // from |curr| which we are validating.
   template<typename T>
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/wasm/wasm-validator.cpp [hunk 1]
```diff
@@ -796,6 +807,37 @@ void FunctionValidator::visitCall(Call* curr) {
...
+  if (Intrinsics(*getModule()).isCallWithoutEffects(curr)) {
+    // call.without.effects has the specific form of the last argument being a
+    // function reference, which will be called with all the previous arguments.
+    // The type must be consistent with that. This, for example, is not:
+    //
+    //  (call $call.without.effects
+    //    (i32.const 1)
+    //    (.. some function reference that expects an f64 param and not i32 ..)
+    //  )
+    if (!curr->operands.empty()) {
```
aheejin, line 820: Is this valid to be empty?

src/wasm/wasm-validator.cpp
```diff
@@ -796,6 +807,37 @@ void FunctionValidator::visitCall(Call* curr) {
...
+      auto* target = curr->operands.back();
+      // Ignore unreachable here, and leave other validation issues here for the
+      // global scope which already does this.
```
aheejin, line 823: Not sure what this means. What are the other validation issues? And what do we need to handle if this is unreachable?

aheejin (approved):

src/wasm/wasm-validator.cpp
kripken: Good point, I added a error for that now.

src/wasm/wasm-validator.cpp
kripken: Yes, this wasn't very clear... I rewrote it not.

If it's unreachable we can ignore it. If it's not a function, then that's a problem with the signature of the import which has the wrong form, and we have validation for that already.

PR #4881, opened 2022-08-05
LegalizeJSInterface: Look for get/setTempRet0 as exports

sbc100:
This allows emscripten to move these helper functions from JS library
imports to native wasm exports.

This change continues to support `getTempRet0/setTempRet0` as imports
while we transition emscripten.  See https://github.com/emscripten-core/emscripten/pull/17581

See https://github.com/emscripten-core/emscripten/issues/7273


sbc100:
I tested this change with the current emscripten, running all of core2 and other suites.

sbc100:
PTAL, I think this is good to land.   Once it rolls in we can look at landing the emscripten side of this, followed by a tidy up here once that is in.

scripts/fuzz_shell.js
```diff
@@ -166,10 +165,6 @@ var imports = {
     'log-f32': logValue,
     'log-f64': logValue,
   },
-  'env': {
-    'setTempRet0': function(x) { tempRet0 = x },
-    'getTempRet0': function() { return tempRet0 },
-  },
```
kripken, line 172: Don't we need to keep providing the imports for now, to avoid fuzzer breakage?

If we don't, then who is responsible for creating and providing those functions after this PR>

sbc100:
Actually I'm working on writing some tests.. not quite ready for review yet.

sbc100:
This change is less intrusive now.  A new option was added that will cause the pass to use existing exports instead of imports.    This will likely only ever be used by emscripten.

src/passes/LegalizeJSInterface.cpp
```diff
@@ -185,6 +208,32 @@ struct LegalizeJSInterface : public Pass {
...
+      if (exportedHelpers) {
+        auto* ex = module->getExport(GET_TEMP_RET_EXPORT);
+        getTempRet0 = module->getFunction(ex->value);
+      } else {
+        getTempRet0 = getFunctionOrImport(
+          module, GET_TEMP_RET_IMPORT, Type::none, Type::i32);
+      }
+    }
+    return getTempRet0;
+  }
```
kripken, line 235: Can use a function called by both of these to reduce the code duplication perhaps?

test/lit/passes/legalize-js-interface.wast
```diff
@@ -0,0 +1,52 @@
...
+  (func $__get_temp_ret (result i32)
+    (i32.const 0)
+  )
+)
+;; CHECK:      (func $legalstub$get_i64 (result i32)
+;; CHECK-NEXT:  (local $0 i64)
+;; CHECK-NEXT:  (local.set $0
+;; CHECK-NEXT:   (call $get_i64)
+;; CHECK-NEXT:  )
+;; CHECK-NEXT:  (call $__set_temp_ret
```
kripken, line 41: It would be good to also have a test that uses `__get_temp_ret`.

kripken (approved):

src/passes/LegalizeJSInterface.cpp
sbc100: The function have different signatures .. so its not trivial to use the same getting here.

kripken: Yeah, I guess it's best as it is.

PR #4884, opened 2022-08-06
Avoid SmallSetBase heap allocation in default constructor

Squareys:
Hi @kripken & @sbc100 !

As described in https://github.com/WebAssembly/binaryen/issues/4165#issuecomment-1176486775, this is the change that wraps the `std::set` in the `SmallSetBase` into a std::optional to avoid the heap allocation that creates the red-black-tree root on MSVC STL.

I made one drive-by change in a separate commit that avoids a copy per element in the for-each of the `std::initializer_list` constructor of SmallSetBase.

Best,
Jonathan from [Wonderland Engine](https://wonderlandengine.com).


src/ir/effects.h
```diff
@@ -318,8 +319,8 @@ class EffectAnalyzer {
     return hasAnything();
   }
 
-  std::set<Name> breakTargets;
-  std::set<Name> delegateTargets;
+  SmallSet<Name, 10> breakTargets;
+  SmallSet<Name, 10> delegateTargets;
```
kripken, line 323: The changes to this file might be worth doing in a separate PR, to allow separate benchmarking.

Offhand I'm not sure if 10 is the best value here. I'd guess lower, since this makes the size of the class much larger and often we don't use any of that. But let's just benchmark various numbers here and see (I can do some of that too).

src/support/small_set.h
```diff
@@ -116,23 +121,23 @@ template<typename T, size_t N, typename FlexibleSet> class SmallSetBase {
...
+      return flexible->size();
     }
   }
 
   bool empty() const { return size() == 0; }
 
   void clear() {
     usedFixed = 0;
-    flexible.clear();
+    if (flexible) flexible->clear();
```
kripken, line 140:
```suggestion
    if (flexible) {
      flexible->clear();
    }
```

src/support/small_set.h
```diff
@@ -39,16 +40,19 @@ template<typename T, size_t N, typename FlexibleSet> class SmallSetBase {
...
 
   bool usingFixed() const {
-    // If the flexible storage contains something, then we are using it.
+    // If the flexible optional has a value, we are using it.
     // Otherwise we use the fixed storage. Note that if we grow and shrink then
     // we will stay in flexible mode until we reach a size of zero, at which
     // point we return to fixed mode. This is intentional, to avoid a lot of
     // movement in switching between fixed and flexible mode.
-    return flexible.empty();
+    return !flexible.has_value();
```
kripken, line 55: I'm not an expert on `std::optional`, but wouldn't this work?
```suggestion
    return !flexible;
```

kripken (commented):
Thanks for the PR!

This sounds good to me. Please let's split out the `effects.h` part though, as suggested in the comments, and then let's benchmark this - I'd like to verify it isn't a significant regression on non-MSVC (but I'm pretty sure it's not).

src/support/small_set.h
Squareys: Yes, that works, too 👍

src/ir/effects.h
Squareys: Sure, can split it off 👍 Though I believe these were the original reason I found the allocation in std::set.
Good point on the "10".

Squareys:
Thanks for the review! I'll have time to apply the requested changes and split the changes from effects.h tomorrow evening CEST.

Squareys:
@kripken Please consider the benchmark and analysis [here, on the other PR](https://github.com/WebAssembly/binaryen/pull/4885#issuecomment-1209527026) before merging this.

While it's a net-positive, it does pessimize the code without the EffectsAnalyzer change, which seems to be a special case that benefits massively from avoiding the allocation, possibly allocated especially often.

kripken (approved):
lgtm % comment

src/support/small_set.h
```diff
@@ -39,16 +40,19 @@ template<typename T, size_t N, typename FlexibleSet> class SmallSetBase {
   size_t usedFixed = 0;
   std::array<T, N> fixed;
 
-  // flexible additional storage
-  FlexibleSet flexible;
+  // flexible additional storage.
+  // Because std::set allocates the root node of its red-black tree on the
+  // heap, we wrap it std::optional to defer that allocation until we start
+  // using the flexible storage.
```
kripken, line 46: This comment is a little specific to the MSVC STL I think. How about this?
```
Flexible additional storage. We use std::optional here to reduce the overhead in the
common case, where the flexible storage is never needed. With std::optional we just
have a pointer taking up space, as opposed to a full std::set which is larger. Also, on
MSVC's STL at least the std::set constructor does work, which we want to avoid here.
```

Squareys:
@kripken Changed the comment and rebased. Sorry for the crazy long delay!

I'll be rebasing #4885 next, which is waiting for this to land.

kripken:
Looks like there are some errors on CI.

Squareys:
Rebased onto latest `main` and fixed the failing test case.

Sorry for the crazy slow iteration cycle times on this, this project was done during a once-per-month "Workflow Improvement Day".

Squareys:
Note, while rebasing #4885, I noticed, that `SmallSet::erase()` was returning `void` [instead of `size_t` like std::set does](https://en.cppreference.com/w/cpp/container/set/erase). I added a commit to update that, such that the code which depends on it in effects analyzer compiles.

kripken (approved):

kripken:
Code lgtm, also ran a quick local benchmark and looks good. Once tests pass this will merge.

kripken:
Looks like there is a test error though.

Squareys:
> Looks like there is a test error though.

Indeed, will find time to have another look before end of the year.

Squareys:
>  end of the year.

Oh well... sorry this is taking so long.

It's not a test failure, but likely a clang or GCC compiler difference, so I'd appreciate if you could approve the workflow as the previous logs expired.
I just rebased to trigger the workflow run.

kripken:
I re-ran the workflow now. (It seems like it needs to be approved each time for first-time contributors, unfortunately.)

Squareys (commented):
@kripken I fixed the test and commented on the two changes I made in the process.

src/support/small_set.h
```diff
@@ -132,16 +135,20 @@ class SmallSetBase {
   // fixed-space storage
   FixedStorage fixed;
 
-  // flexible additional storage
-  FlexibleSet flexible;
+  // Flexible additional storage. We use std::optional here to reduce the
+  // overhead in the common case, where the flexible storage is never needed.
+  // With std::optional we just have a pointer taking up space, as opposed to a
```
kripken, line 140: Rereading this now, this doesn't seem true at least on linux, as `sizeof` tells me that the full storage is used. That is, the gnu C++ STL seems to implement an optional as enough space for the item, and a tag whether it is present or not.

Given that, I think we might have been wrong in our previous discussion here. That is, using `std::optional` here increases memory size, so it may have downsides, and we should probably be more careful.

If our goal is to avoid a problem with the MSVC STL, then a workaround specific to there might make more sense actually.

What I'd suggest immediately though is to perhaps go look at what LLVM's SmallSet does, as a lot of thought has gone into that. We should probably do what they are doing.

Squareys: Unfortunately, [LLVM's SmallSet](https://llvm.org/doxygen/SmallSet_8h_source.html) doesn't seem to address the heap allocation on MSVC. 

I believe the [measurements you had made](https://github.com/WebAssembly/binaryen/pull/4885#issuecomment-1209693350) were on Linux? It seemed benefits from using avoiding the std::set construction outweighed the size increase there, too? (At least when used in EffectsAnalyzer, and way less drastic as on windows... I suppose it's not making heap allocations in the STL implementation there).

Effectively, we could, of course, avoid the entire PR, if we'd use another STL, [as you had suggested in this comment](https://github.com/WebAssembly/binaryen/issues/4165#issuecomment-1176486775).

PR #4889, opened 2022-08-09
[Wasm GC] GC-prefixed opcodes are int8, not LEBs

kripken:
This starts to matter with strings, it turns out. This change should make us
runnable in v8.

https://github.com/WebAssembly/gc/blob/main/proposals/gc/MVP.md#instructions-1

### DIFF
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -3830,7 +3830,7 @@ BinaryConsts::ASTNodes WasmBinaryBuilder::readExpression(Expression*& curr) {
       break;
     }
     case BinaryConsts::GCPrefix: {
-      auto opcode = getU32LEB();
+      uint8_t opcode = getInt8();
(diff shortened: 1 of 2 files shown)


aheejin (approved):

tlively:
Please revert this. It must be a bug in V8.

kripken:
Happy to do it either way, but the spec doc linked above seems to be aligned with v8?

kripken:
To be clear, the part of the spec I am reading is stuff like `0xfb01` which I assume means those literal bytes..? Maybe that's wrong...

tlively:
Talked to Jakob this morning about this and we're going to switch over to using LEBs for now. It's been an established convention for years that prefixed opcodes are LEBs and if we're going to break with that convention, we will require more stakeholder buy-in first. You can see that this is how the binary format for SIMD instructions and all other prefixed instructions is specified: https://webassembly.github.io/spec/core/binary/instructions.html#vector-instructions.

PR #4890, opened 2022-08-09
Fuzzer: Emit absolute paths in commands

kripken:
Any filename that shows up in a command that we log for the user should be
an absolute path, so that the user can just copy and paste the command. Otherwise,
the fuzzer runs in `out/test/` which the user has to constantly type.

I looked for a better/general way to do this, but can't find anything good other than
manually annotating all the relevant places...

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -33,6 +33,7 @@ import re
 import sys
 import time
 import traceback
+from os.path import abspath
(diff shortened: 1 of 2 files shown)


ashleynh (approved):
Awesome, great improvement!

tlively:
:tada:

It would be great to fix this in check.py as well.

PR #4895, opened 2022-08-10
Revert "[Wasm GC] GC-prefixed opcodes are int8, not LEBs"

kripken:
Reverts WebAssembly/binaryen#4889

The spec is unclear on this, and that PR moved us to do what V8 does. But
it sounds like we should clarify the spec to do things the other way, so this
goes back to that.


tlively (approved):
Great, thanks!

PR #4897, opened 2022-08-10
Function-level pass-debug mode 2 validation

kripken:
In `BINARYEN_PASS_DEBUG=2` we save the module before each pass, and if
validation fails afterwards, we print the module before. This PR does the same for
function-parallel passes - in that case, we can actually show the specific function
that broke validation, as opposed to the whole module.

This has been quite useful in debugging some tricky wasm GC bugs that were
uncovered on J2Wasm by GUFA - getting a single function instead of the entire
huge module.


src/passes/pass.cpp
```diff
@@ -891,18 +891,48 @@ void PassRunner::runPass(Pass* pass) {
...
   }
   instance->runOnFunction(this, wasm, func);
   handleAfterEffects(pass, func);
-  if (getPassDebug()) {
+  if (passDebug) {
     checker->check();
   }
+
+  if (extraFunctionValidation) {
+    if (!WasmValidator().validate(func, *wasm, WasmValidator::Minimal)) {
```
aheejin, line 928: What do we do when the validation mode is `WasmValidator::Minimal`?

src/passes/pass.cpp
```diff
@@ -891,18 +891,48 @@ void PassRunner::runPass(Pass* pass) {
...
+      Fatal() << "Last nested function-parallel pass (" << pass->name
+              << ") broke validation of function " << func->name
+              << ". Here is the function body before:\n"
+              << bodyBefore.str() << "\n\nAnd here it is now:\n"
+              << *func->body << '\n';
```
aheejin, line 933: When multiple functions break, does this printing logic run in parallel so printing can be mixed? It's fine bc multiple failings don't occur often and it's mostly for our debugging?

tlively (commented):
LGTM beyond @aheejin's questions

src/passes/pass.cpp
kripken: In minimal mode we don't do any global checks. Like we don't check if a call's target exists. Maybe "minimal" could be "local" or "function-level"...?

src/passes/pass.cpp
kripken: Hmm, good point... yes, this is really just for internal debugging, and in practice this works well, but this is nondeterministic if there are multiple errors. I hadn't thought of that.

Handling that deterministically would take some work (save all the results globally, then process them later) so maybe we can leave that for later if we ever need it?

src/passes/pass.cpp
aheejin: I asked because we don't currently seem to use this `Minimal` value anywhere in the toolchain. (I grepped with `Minimal`)

And why do we skip the `extraFunctionValidation` if it's `MInimal`, given that this is a function validation?

src/passes/pass.cpp
aheejin: Yeah that sounds good.

aheejin (approved):

src/passes/pass.cpp
kripken: Yeah, I guess Minimal is just 0, so it's "nothing specified" (no global validation, on web validation, no quiet validation). Maybe we can just delete it and use 0,  if we think that's clearer.

I don't understand the second question? We only do this extra work if we decided to do `extraFunctionValidation`. That the work is just a non-global validation is separate from that?

aheejin: Oh, I was just wondering why we have that `if` clause here, if `extraFunctionValidation` and minimal validation are orthogonal things.

kripken: Oh, just that I thought to do this extra minimal validation only when pass-debug=2 (which makes `extraFunctionValidation` true). Mode 2 is much slower than mode 1, and I didn't want to slow down mode 1 here.

PR #4898, opened 2022-08-11
Restore the `extern` heap type

tlively:
The GC proposal has split `any` and `extern` back into two separate types, so
reintroduce `HeapType::ext` to represent `extern`. Before it was originally
removed in #4633, externref was a subtype of anyref, but now it is not. Now that
we have separate heaptype type hierarchies, make `HeapType::getLeastUpperBound`
fallible as well.


tlively:
* main
  * **PR #4898** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4898" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4898?utm_source=stack-comment).

src/ir/possible-constant.h
```diff
@@ -114,8 +114,9 @@ struct PossibleConstantValues {
       auto type = getConstantLiteral().type.getHeapType();
       auto otherType = other.getConstantLiteral().type.getHeapType();
       auto lub = HeapType::getLeastUpperBound(type, otherType);
-      if (lub != type) {
-        value = Literal::makeNull(lub);
+      assert(lub && "TODO: Handle case without LUB");
```
kripken, line 117: I think we should handle this in this PR, or else we'll get fuzzer failures. To handle it, if there is no LUB let's turn the result into `Many()`.

src/ir/possible-contents.cpp
```diff
@@ -56,7 +56,8 @@ void PossibleContents::combine(const PossibleContents& other) {
       assert(other.isNull());
       auto lub = HeapType::getLeastUpperBound(type.getHeapType(),
                                               otherType.getHeapType());
-      value = Literal::makeNull(lub);
+      assert(lub && "TODO: handle case where there is no LUB");
+      value = Literal::makeNull(*lub);
```
kripken, line 60: Same as above, this can make a `Many()`.

kripken (commented):
Why are the tests changed here?

tlively:
The tests changed because their input used "externref" or "(ref extern)", which previously was an alias for "anyref" and "(ref any)" but now is not.

src/ir/possible-constant.h
tlively: Shouldn't we turn it into a `None`, since the types are incompatible? Although once we implement bottom types, we'll be able to make it a `Literal` of that bottom type instead.

kripken: `None` means "no contents". `Many` means "many things are possible here", which can include many incompatible things.

kripken:
I see about the tests, thanks.

kripken (approved):
lgtm if this passes fuzzing.

src/binaryen-c.cpp
```diff
@@ -58,10 +58,11 @@ BinaryenLiteral toBinaryenLiteral(Literal x) {
         case HeapType::func:
           ret.func = x.isNull() ? nullptr : x.getFunc().c_str();
           break;
-        case HeapType::any:
+        case HeapType::ext:
         case HeapType::eq:
           assert(x.isNull() && "unexpected non-null reference type literal");
           break;
+        case HeapType::any:
```
aheejin, line 65: Why is `any`'s behavior here different from that of `ext`? Is `any` not nullable?

src/passes/InstrumentLocals.cpp
```diff
@@ -57,15 +57,15 @@ Name get_f32("get_f32");
 Name get_f64("get_f64");
 Name get_v128("get_v128");
 Name get_funcref("get_funcref");
-Name get_anyref("get_anyref");
+Name get_extref("get_extref");
```
aheejin, line 60: We don't abbreviate `externref` to `extref` in other places, so how about just `get_externref`/`set_externref` here as well?

aheejin (commented):
Some places this didn't just add `externref` but replaced `anyref` with it. We don't need `anyref` in those places anymore?

tlively:
Note to self: I should also update the changelog.

tlively:
> Some places this didn't just add `externref` but replaced `anyref` with it. We don't need `anyref` in those places anymore?

Right, these were places that were logically using externref all along, but because we had externref == anyref, they had to use anyref in the code.

src/binaryen-c.cpp
tlively: We only support the standardized reference types `extern` and `func` in this API and all the GC types are meant to hit that WASM_UNREACHABLE. Previously `HeapType::any` represented the standardized `extern`, but now it only represents the sepate `any` GC type.

tlively:
Fuzzer looks good when I hack it to use v8-10.5.117, which still has func <: any. The next PR will make it work with up-to-date v8.

kripken:
`./scripts/fuzz_opt.py 5710490634504070365` fails on `main` now, which I am guessing is related to this as the error is:
```
binaryen/src/tools/fuzzing/fuzzing.cpp:1972: wasm::Expression* wasm::TranslateToFuzzReader::makeConstBasicRef(wasm::Type): Assertion `type.isNullable() && "Cannot handle non-nullable externref"' failed.
```

kripken:
Bisected to confirm, the breakage indeed starts here.

This type of error seems to happen every 1,000 or so fuzzer iterations now, but sometimes it's 100 or 10,000, so this type of rare issue is easy to miss...

PR #4900, opened 2022-08-11
[Wasm GC] Fix TypeRefining on fallthrough values via tee

kripken:
A rather tricky corner case: we normally look at fallthrough values for copies of
fields, so when we try to refine a field, we ignore stuff like this:
```js
a.x = b.x;
```
That copies the same field on the same type to itself, so refining is not limited by
it. But if we have something else in the middle, and that thing cannot change
type, then it is a problem, like this:
```wat
(struct.set
  (..ref..)
  (local.tee $temp
    (struct.get)))
```
`tee` has the type of the local, which does not change in this pass. So we can't
look at just the fallthrough here and skip the `tee`: after refining the field, the
`tee`'s old type might not fit in the field's new type.

We could perhaps add casts to fix things up, but those may have too big a
cost. For now, just ignore the fallthrough.


tlively:
This is why we should have fixed the return type of `tee` :cry:

tlively (commented):
It would be more precise to add a `includeTees` parameter to `getFallthrough`, since AFAICT that's the only instruction that would exhibit this problem. Pass-through control flow structures would have to have their types updated, but we must already be handling that.

kripken:
Thinking about that, I'm not sure if `tee` is the only hard case here. It's true that we can handle block fallthroughs:
```wat
(struct.set
  ..
  (block (result $A)
    (struct.get)))
```
But we don't have a guarantee in general of handling other stuff, I think:
```wat
(struct.set
  ..
  (if (result $A)
    ..
    (return)
    (struct.get)))
```
Do we handle that today? I wasn't sure, but reading the code I think we happen to.

What about `br_if`'s fallthrough - is that the same as the input value, or does it match the block's type? Seems similar in a way to the `tee` situation, where the type is pessimized to match the place it'll be written.

tlively:
Oh right, `br_if` is the same as `tee`. Those are definitely the only two, though.

kripken:
Hmm, I think we don't follow the spec then, since finalize for `br_if` will return the type of the value. But (as I always end up saying when `br_if`'s return value comes up), `br_if`'s return value is not used in practice and is arguably a spec blunder...

Ok, if it's just those two then I'll look into adding logic in `getFallthrough`.

kripken:
@tlively done.

src/ir/properties.h
```diff
@@ -240,17 +240,28 @@ inline Index getZeroExtBits(Expression* curr) {
...
+// not the value, so if we returned the fallthrough of the tee we'd have a
+// possible difference between the type in the IR and the type of the value,
+// which some cases care about).
+//
 // TODO: Receive a Module instead of FeatureSet, to pass to EffectAnalyzer?
-inline Expression* getImmediateFallthrough(Expression* curr,
-                                           const PassOptions& passOptions,
-                                           Module& module) {
+
+enum class FallthroughBehavior { AllowTee, NoTee };
```
tlively, line 251: Maybe `AllowTypeNarrowing` and `NoAllowTypeNarrowing` to be more descriptive and to generalize to br_if as well?

tlively (commented):
I think it would be good to include `br_if` here so we can avoid some miscompilations, even if we don't catch all of them due to our incorrect refinalization.

src/ir/properties.h
kripken: Hmm, I'm not sure `AllowTypeNarrowing` is quite right. A block can also narrow a type, so the real issue is allowing certain easily-fixable narrowings. A block's narrowing can always be fixed with a refinalize (I think?), but a tee's can't. A tee's can always be fixed in other ways though, like adding a cast, so even "AllowFixableTypeNarrowing" isn't quite right.

Since I can't see a good way to do that, I'd still prefer to have this focus on tee. If we are worried about br_if return values, then separately we could just ignore them in this function (since they are never used in practice, that would be fine).

tlively: I do think it would be good to handle br_if correctly as well. Presumably we will eventually see fuzzer failures if we don't(?). I can't think of a better name for this property, so how about keeping `AllowTee` and `NoTee`, but also having it apply to br_if?

tlively (approved):

src/ir/properties.h
kripken: Fair enough, maybe we'll get fuzzer failures some day. Ok, I generalized the name to `AllowTeeBrIf`, a little verbose but it's accurate and the best name I can think of.

PR #4901, opened 2022-08-11
[EH] Pop should be supertype of tag type

aheejin:
`pop`s type should be a supertype, not a subtype, of the tag's type
within `catch`.

Thanks @tlively for catching this!


tlively (approved):
Thanks!

PR #4902, opened 2022-08-12
Fix name of port_passes_tests_to_lit.py script. NFC

sbc100:
I was reading these tests and failing to find the names script.


tlively (approved):
Oops! Thanks for the fix

PR #4904, opened 2022-08-12
ReorderGlobals pass

kripken:
This sorts globals by their usage (and respecting dependencies). If the module
has very many globals then using smaller LEBs can matter.

On most MVP files this won't matter, since there are only a few globals at
most (the stack pointer, maybe dynamic linking?). With GC there is more reason
to use globals for things like vtables. This shrinks J2Wasm by 7% (!), though only
0.5% after gzip. This doesn't help all GC code, though - on Dart it just helps
by 0.2% both before and after gzip.

Given the low benefit in general, I wonder if this should not be turned on by
default. Perhaps it could be skipped if we have fewer than 128 globals, for
example? (though that would make testing less simple, as small testcases
would be skipped...)

Credit to @jakobkummerow for the idea.


kripken:
I made it not do any work if there are less than 128 globals. That should avoid any slowdown on existing MVP wasm files.

src/passes/ReorderGlobals.cpp
```diff
@@ -0,0 +1,159 @@
...
+    for (auto& global : module->globals) {
+      counts[global->name];
+    }
+
+    // Count uses.
+    UseCountScanner scanner(counts);
+    scanner.run(getPassRunner(), module);
+    scanner.runOnModuleCode(getPassRunner(), module);
+
+    // To sort, we must first find dependencies, since if $b's definition
```
tlively, line 86: I think this could be much simpler if you use the `TopologicalSort` utility. If you sort the initial items from least used to most used, then the topological sort will generally order them with the most used items first. We use this strategy to sort types by decreasing use here: https://github.com/WebAssembly/binaryen/blob/main/src/ir/module-utils.cpp#L297-L300

src/passes/pass.cpp
```diff
@@ -356,6 +356,12 @@ void PassRegistry::registerPasses() {
   registerPass("reorder-functions",
                "sorts functions by access frequency",
                createReorderFunctionsPass);
+  registerPass("reorder-globals",
+               "sorts globals by access frequency",
+               createReorderGlobalsPass);
+  registerPass("reorder-globals-always",
```
tlively, line 362: This will keep the pass from showing up in the `--help` output.
```suggestion
  registerTestPass("reorder-globals-always",
```

test/lit/passes/reorder-globals.wast
```diff
@@ -0,0 +1,376 @@
...
+(module
+  ;; CHECK:      (global $a i32 (i32.const 10))
+  ;; RTRIP:      (global $a i32 (i32.const 10))
+  (global $a i32 (i32.const 10))
+  ;; CHECK:      (global $b i32 (global.get $a))
+  ;; RTRIP:      (global $b i32 (global.get $a))
+  (global $b i32 (global.get $a))
+  ;; CHECK:      (global $c i32 (global.get $b))
+  ;; RTRIP:      (global $c i32 (global.get $b))
+  (global $c i32 (global.get $b))
```
tlively, line 163: This is not actually valid Wasm because constant expressions can only refer to imported globals. Is that a problem?

test/lit/passes/reorder-globals.wast
```diff
@@ -0,0 +1,376 @@
...
+      (global.get $c)
+    )
+  )
+)
+
+;; $b has more uses, but $a is an import and must remain first.
+(module
+  ;; CHECK:      (import "a" "b" (global $a i32))
+  ;; RTRIP:      (import "a" "b" (global $a i32))
+  (import "a" "b" (global $a i32))
```
tlively, line 356: I don't see where in the code this is enforced. Is it first here because of the tie breaking with the names?

src/passes/ReorderGlobals.cpp
kripken: Good idea, done. It's not less code, it turns out, but it is simpler and avoids slowdowns.

test/lit/passes/reorder-globals.wast
kripken: Right, I forgot that spec limitation... that does make the problem simpler I guess, and the pass is more general than it needs to be. But there's no harm in supporting more things, and I'm sure the spec will eventually allow this, at which point this code will just work.

test/lit/passes/reorder-globals.wast
kripken: There is no need to handle this in the pass, as both text and binary writing first emit all imports and then all non-imports. So this part of the test really just tests other parts of the codebase, I guess.

tlively: Hmm, I thought we had an invariant that imported items came first in the IR as well. Maybe I'm wrong about that, though. Looking at `ReorderFunctions.cpp`, it doesn't look like it tries to keep imports first.

(I also noticed that `ReorderFunctions` does not take `ref.func` into account!)

test/lit/passes/reorder-globals.wast
```diff
@@ -0,0 +1,404 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
+;; RUN: foreach %s %t wasm-opt --reorder-globals-always -S -o - | filecheck %s
+;; RUN: foreach %s %t wasm-opt --reorder-globals-always --roundtrip -S -o - | filecheck %s --check-prefix RTRIP
```
tlively, line 3: Unless I missed something, the roundtrip output is the same as the normal output, so there's no need for a separate prefix.

```suggestion
;; RUN: foreach %s %t wasm-opt --reorder-globals-always --roundtrip -S -o - | filecheck %s
```

tlively (approved):

test/lit/passes/reorder-globals.wast
kripken: It doesn't seem like we do have that invariant. Though in practice that might be the case.

Good catch about ReorderFunctions and `ref.func` - good thing we don't run that pass yet :smile:

test/lit/passes/reorder-globals.wast
kripken: The output is identical, but I was thinking this might prevent us from somehow emitting invalid binaries. Like if we had an invariant on the order in the IR (which apparently we don't, as mentioned above, but if we did, and we broke it).

(see comment on lines 5-6)

tlively: Yeah, testing roundtripping sounds good, but it can reuse the same `CHECK` output as the previous `RUN` line.

kripken: Oh, ok, makes sense.

PR #4908, opened 2022-08-15
Fuzzer: When typed-function-references is disabled, disable GC too

kripken:


tlively:
Does this just work around bugs where we emit GC things without properly checking that typed function references is enabled as well? If so, maybe we should fix those bugs or just combine the two feature flags into a single feature, since there is no use case for typed function references on their own.

kripken:
I'm not sure. I realize I didn't fully understand your answer to the question I asked you offline that led to this PR. But the issue is one of the recent V8 changes somehow.

This change looks right to me since it is consistent with the existing reference type code right above it. I can investigate more if we want, but regardless I think this looks correct?

tlively (approved):
Sure, this LGTM and maybe I can look at combining these features separately.

PR #4909, opened 2022-08-15
Validator: Validate unreachable calls more carefully

kripken:
Normally the validator will find stale types properly, by just running refinalize and seeing
if the type has changed (if so, then some code forgot to refinalize). However, refinalize
is a local operation, so it does not apply to calls: a call's proper type is determined by
the global information of the function we are calling. As a result, we would not notice
errors like this:
```wat
(call $foo) ;; type: unreachable
```
Refinalizing that would not change the type from unreachable to the proper type, since
that is global information.

To validate this properly, validate that a call whose type is unreachable actually has
an unreachable child. That rules out an invalid unreachable type here, which leaves
concrete types, that we already have proper global validation for.


tlively (commented):
Would it be worth doing this more generally rather than just for calls? We could check that every unreachable instruction is in a specific list of unreachable instructions or has an unreachable child. That would catch other cases such as `global.get` and future cases as well.

kripken:
Good idea! Yes, I think that makes sense, done.

src/wasm/wasm-validator.cpp
```diff
@@ -327,6 +327,57 @@ struct FunctionValidator : public WalkerPass<PostWalker<FunctionValidator>> {
...
       }
     }
+
+    // Also verify that only allowed expressions end up in the situation where
+    // the expression has type unreachable but there is no unreachable child.
+    // For example a Call with no unreachable child cannot be unreachable, but a
+    // Break can be.
+    if (curr->type == Type::unreachable) {
+      switch (curr->_id) {
+        case Expression::BreakId:
```
tlively, line 337: If the break has a value, it should fall through to the general logic, right? (same with br_table)

tlively (approved):

src/wasm/wasm-validator.cpp
kripken: I don't think a value changes things here? The type is still unreachable even with a value. A condition can change the type, but we already have full validation logic for that in `visitBreak`. I added a comment clarifying all that now.

tlively: Or right, I was thinking of `br_if` but forgot that it's a separate expression class. For br_table I was just confused.

tlively: Oh wait, it's not a separate class. But the comment you added was exactly what I needed 👍

PR #4913, opened 2022-08-17
Try to fix Alpine breakage on CI

kripken:
```
 "/src/scripts/test/node-esm-loader.mjs 'resolve'" did not call the next hook in its chain and did not explicitly signal a short circuit. If this is intentional, include `shortCircuit: true` in the hook's return.
```
Seems like it might be due to a node update in Alpine?

Hopefully this PR fixes that, but I have no idea...


aheejin (approved):
I have no idea either but yeah why not 🤷🏻

kripken:
Looks like the Alpine build passed... I'll wait for the rest before landing.

PR #4933, opened 2022-08-18
Fix Memory64 binary parsing after #4811

kripken:
Due to missing test coverage, we missed in #4811 that some memory operations
needed to get `make64()` called on them.

This should unbreak the Binaryen roller.


src/wasm/wasm-binary.cpp
```diff
@@ -2220,6 +2220,16 @@ Name WasmBinaryBuilder::getTagName(Index index) {
...
+Memory* WasmBinaryBuilder::getMemory(Index index) {
+  Index numMemoryImports = memoryImports.size();
+  if (index < numMemoryImports) {
+    return memoryImports[index];
+  } else if (index - numMemoryImports < memories.size()) {
+    return memories[index - numMemoryImports].get();
+  }
+  throwError("Memory index out of range.");
+}
+
```
kripken, line 2232: This code is just moved.

src/passes/Memory64Lowering.cpp
```diff
@@ -29,10 +29,6 @@ namespace wasm {
 
 struct Memory64Lowering : public WalkerPass<PostWalker<Memory64Lowering>> {
 
-  void run(PassRunner* runner, Module* module) override {
```
ashleynh, line 32: Are you disabling this pass for now?

ashleynh (approved):
I see how I missed this. Thanks for fixing Alon!

src/passes/Memory64Lowering.cpp
kripken: Oh, this is just a no-op change. The function was just calling the super, and without this code the super is just called normally.

(If the pass were disabled then the relevant test would break - I verified we have such a test now, `test/passes/memory64-lowering_enable-memory64_enable-bulk-memory_enable-threads.wast`)

kripken:
(Landing to try to unbreak the roller, but we can continue discussion here and have followups if necessary.)

PR #4936, opened 2022-08-18
[Fuzzer] Fuzz TrapsNeverHappen mode

kripken:
This mode is tricky to fuzz because the mode is basically "assume traps never
happen; if a trap *does* happen, that is undefined behavior". So if any trap occurs
in the random fuzz testcase, we can't optimize with `-tnh` and assume the results
stay to same. To avoid that, we ignore all functions from the first one that traps,
that is, we only compare the code that ran without trapping. That often is a small
subset of the total functions, sadly, but I do see that this ends up with some useful
coverage despite the drawback.

This also requires some fixes to comparing of references, specifically, funcrefs
are printed with the function name/index, but that can change during opts, so
ignore that. This wasn't noticed before because this new fuzzer mode does
comparisons of `--fuzz-exec-before` output, instead of doing `--fuzz-exec` which
runs it before and after and compares it internally in `wasm-opt`. Here we are
comparing the output externally, which we didn't do before.

Fuzzing with this for a few days, it seems stable.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -508,8 +508,14 @@ def fix_output(out):
             x = str(float(x))
         return 'f64.const ' + x
     out = re.sub(r'f64\.const (-?[nanN:abcdefxIity\d+-.]+)', fix_double, out)
+
     # mark traps from wasm-opt as exceptions, even though they didn't run in a vm
     out = out.replace(TRAP_PREFIX, 'exception: ' + TRAP_PREFIX)
+
+    # funcref(0) has the index of the function in it, and optimizations can
+    # change that index, so ignore it
+    out = re.sub(r'funcref\(\d+\)', 'funcref()', out)
+
     lines = out.splitlines()
     for i in range(len(lines)):
         line = lines[i]
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -1039,7 +1039,7 @@ class TrapsNeverHappen(TestCaseHandler):
         before = run_bynterp(before_wasm, ['--fuzz-exec-before'])
         after_wasm_tnh = after_wasm + '.tnh.wasm'
         run([in_bin('wasm-opt'), before_wasm, '-o', after_wasm_tnh, '-tnh'] + opts + FEATURE_OPTS)
-        after = 'waka\n' + run_bynterp(after_wasm_tnh, ['--fuzz-exec-before'])
+        after = run_bynterp(after_wasm_tnh, ['--fuzz-exec-before'])
 
         # if a trap happened, we must stop comparing from that.
         if TRAP_PREFIX in before:
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -1061,7 +1061,7 @@ class TrapsNeverHappen(TestCaseHandler):
             after_index = after.index(call_line)
             after = after[:after_index]
 
-        compare_between_vms(before, after, 'TrapsNeverHappen')
+        compare_between_vms(fix_output(before), fix_output(after), 'TrapsNeverHappen')
 
 
 # Check that the text format round-trips without error.
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -1028,6 +1028,42 @@ class Asyncify(TestCaseHandler):
         return all_disallowed(['exception-handling', 'simd', 'tail-call', 'reference-types', 'multivalue', 'gc'])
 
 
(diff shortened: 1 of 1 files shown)


aheejin:
When do we use `--fuzz-exec` and when to use `--fuzz-exec-before`? I think we have a similar logic of comparison within binaryen itself when using `--fuzz-exec`... Can we use that instead?

scripts/fuzz_opt.py
```diff
@@ -1028,6 +1036,65 @@ def can_run_on_feature_opts(self, feature_opts):
...
+            after = after[:after_index]
+
+        # some results cannot be compared, so we must filter them out here.
+        def ignore_references(out):
+            ret = []
+            for line in out.splitlines():
+                # only result lines are relevant here, which look like
+                # [fuzz-exec] note result: foo => [...]
+                if FUZZ_EXEC_NOTE_RESULT in line:
+                    # we want to filter out things like "anyref(null)" or
```
tlively, line 1085: Why are we filtering out nulls, again?

tlively (approved):

kripken:
@aheejin 

> When do we use --fuzz-exec and when to use --fuzz-exec-before? I think we have a similar logic of comparison within binaryen itself when using --fuzz-exec... Can we use that instead?

`--fuzz-exec` will print the execution results before and after opts, and it compares them and errors if they do not match. `--fuzz-exec-before` just prints results before and nothing else.

`--fuzz-exec` isn't good for us here since we need to stop comparing once a trap happens. So we need to add logic to find if a trap happens and to ignore everything from there.

This PR adds that kind of logic in the python driver, by parsing and modifying `--fuzz-exec-before` output. I guess another option could be to add that logic in `execution-results.h`. I suspect that would take more work but I could look into it if that sounds better for some reason.

scripts/fuzz_opt.py
kripken: All nulls have the same value, so we want to ignore their type, basically. This changes them so that all nulls end up with the same string and they compare equal (but non-nulls would not).

kripken:
Actually, wait, most of the logic is already written in `execution-results.h`... maybe 90% of it :rofl: Not sure when that happened... anyhow I'll look into that before doing anything more here.

kripken:
I was overly optimistic about the other approach. Existing code was doing maybe 10% of the work, not 90%... here is a diff that does 90%, I think:

https://github.com/WebAssembly/binaryen/compare/fuzz_tnh_erh?expand=1

It is more complex a diff, I think, so I'm not sure it's actually a better approach. I lean towards this PR.

scripts/fuzz_opt.py
tlively: I wonder if we should be omitting types when we print null values... But I guess this will be moot as soon as I implement the new bottom types, since then null values will always have a bottom type and will not be interchangeable.

tlively (approved):

scripts/fuzz_opt.py
```diff
@@ -1028,6 +1036,65 @@ def can_run_on_feature_opts(self, feature_opts):
...
+                # during startup (the start function, or memory segment
+                # operations, etc.). in that case there is nothing for us to
+                # compare here; just leave.
+                return
+            call_end = before.index('\n', call_start)
+            call_line = before[call_start:call_end]
+            before_index = before.index(call_line)
+            lines_pre = before.count(os.linesep)
+            before = before[:before_index]
+            lines_post = before.count(os.linesep)
```
aheejin, line 1071: Without knowing the print format, it's kind of hard to figure out what all these variables will contain.. but this is a script anyway, so if this works I don't want to hold this up...

aheejin (approved):

scripts/fuzz_opt.py
kripken: Makes sense to add more details here, I added some comments now.

scripts/fuzz_opt.py
```diff
@@ -1028,6 +1036,80 @@ def can_run_on_feature_opts(self, feature_opts):
...
+                # operations, etc.). in that case there is nothing for us to
+                # compare here; just leave.
+                return
+            call_end = before.index('\n', call_start)
+            # we now know the contents of the call line after which the trap
+            # happens, which is something like "[fuzz-exec] calling bar", and
+            # it is unique since it contains the function being called.
+            call_line = before[call_start:call_end]
+            # find that call line, and remove everything from it onward.
+            before_index = before.index(call_line)
```
aheejin, line 1083: Isn't `before_index` the same as `call_start`?

kripken: Hmm, good point... I had it like this to be more symmetric to the "after" case, but you're right that maybe it's confusing. I'll open a followup.

PR #4937, opened 2022-08-18
Only look at the relevant parameter in param-utils:removeParameter

kripken:
Followup to #4910.

### DIFF
--- a/src/passes/param-utils.cpp
+++ b/src/passes/param-utils.cpp
@@ -81,9 +81,7 @@ bool removeParameter(const std::vector<Function*>& funcs,
         .hasUnremovableSideEffects();
     bool wouldChangeType =
       call->type == Type::unreachable && !call->isReturn &&
-      std::any_of(operands.begin(), operands.end(), [](Expression* operand) {
-        return operand->type == Type::unreachable;
-      });
+      operands[index]->type == Type::unreachable;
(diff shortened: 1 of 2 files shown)


src/passes/param-utils.cpp
```diff
@@ -79,11 +79,8 @@ bool removeParameter(const std::vector<Function*>& funcs,
     bool hasUnremovable =
       EffectAnalyzer(runner->options, *module, operands[index])
         .hasUnremovableSideEffects();
-    bool wouldChangeType =
-      call->type == Type::unreachable && !call->isReturn &&
-      std::any_of(operands.begin(), operands.end(), [](Expression* operand) {
-        return operand->type == Type::unreachable;
-      });
+    bool wouldChangeType = call->type == Type::unreachable && !call->isReturn &&
+                           operands[index]->type == Type::unreachable;
```
tlively, line 83: We could be even more precise and additionally check that none of the _other_ operands is unreachable. If one of them is, then removing this one still won't change the type. Of course if this DAE is run after DCE, that won't matter, so maybe it's not worth doing.

tlively (approved):

src/passes/param-utils.cpp
kripken: Yes, I think that could help but is not likely to in the general case. For now let's not add that extra complexity.

PR #4939, opened 2022-08-19
Fixing print-profile.wast test

ashleynh:
In a multi-memories world, memory must be explicitly defined before being exported. Fix CI for 6b3f3af606f2cbe8da29996e5600d174c5653d05.

/cc @sps-gold

### DIFF
--- a/test/lit/wasm-split/print-profile.wast
+++ b/test/lit/wasm-split/print-profile.wast
@@ -15,6 +15,7 @@
 ;; UNESCAPED: - bar(double[3])
 
 (module
+  (memory 0 0)
   (export "memory" (memory 0 0))
   (export "foo" (func $foo))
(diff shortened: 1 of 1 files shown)


tlively (approved):
I’m surprised this works without giving the memory a name. A name (or at least index) on the export statement would say which memory is being exported. Maybe one of the 0s in the export declaration is being interpreted as an index? In that case, the other 0 seems like it shouldn’t be there.

But anyway lgtm to fix CI. Parser issues will eventually be resolved by switching to the new parser.

kripken (approved):

kripken:
@tlively - good points, it seems like something is wrong in the s-parser logic there... the export declaration should only take a name or an index, and error on having an extra param.

kripken:
Landing this to unbreak CI and PRs.

PR #4943, opened 2022-08-19
Avoid adding new unneeded names to blocks in text roundtripping 

kripken:
Previously the wat parser would turn this input:
```wat
(block
  (nop)
)
```
into something like this:
```wat
(block $block17
  (nop)
)
```
It just added a name all the time, in case the block is referred to by an index
later even though it doesn't have a name.

This PR makes us rountrip more precisely by not adding such names: if there
was no name before, and there is no break by index, then do not add a name.

In addition, this will be useful for non-nullable locals since whether a block has
a name or not matters there. Like #4912, this makes us more regular in our
usage of block names.

Unfortunately the diff here is huge. But 99% of it is in the very first commit,
which aside from the small code change is 100% automatic. The small commits later
are where I had to add some manual changes (as some tests actually depended
on the old behavior, and I wanted to keep the tests doing the same as much as
possible). Reading them might be more useful (but they are relative to the big
initial comment, and undo changes there, so looking at the same file in the total
diff might be interesting).


src/wasm/wasm-s-parser.cpp
```diff
@@ -1562,8 +1573,17 @@ Expression* SExpressionWasmBuilder::makeBlock(Element& s) {
...
     curr->finalize(curr->type);
+    // If the block never had a name, and one was not needed in practice (even
+    // if one did not exist, perhaps a break targeted it by index), then we can
+    // remove the name. Note that we only do this if it never had a name: if it
+    // did, we don't want to change anything; we just want to be the same as
+    // the code we are loading - if there was no name before, we don't want one
+    // now, so that we roundtrip text precisely.
+    if (!hadName && !BranchUtils::BranchSeeker::has(curr, curr->name)) {
+      curr->name = Name();
+    }
```
tlively, line 1584: I think it would be fairly simple to avoid the quadratic behavior of ``BranchSeeker` by keeping a set of referenced names as we build branch expressions.

kripken: Yeah, that could be done. It would add complexity though, and this is not a hot code path - we only seek here if there never was a name. Blocks without a name tend to be rare workarounds for stacky code etc., and they do not occur in the actually dangerous situation of a tall tower of blocks for a switch - all the blocks there are named. Also, this is in the s-parser, which is not heavily optimized anyhow.

test/passes/remove-unused-brs_enable-multivalue.txt
```diff
@@ -19,10 +19,11 @@
  )
  (func $b1 (param $i1 i32)
   (block $topmost
-   (block $block
+   (block
     (drop
      (i32.const 0)
     )
+    (nop)
```
tlively, line 26: Do you know where these extra `(nop)`s are coming from?

tlively (approved):

test/passes/remove-unused-brs_enable-multivalue.txt
kripken: Good point, those shouldn't be there. This looks like a minor issue uncovered by this PR. It's not a regression on real-world code since this PR just affects text inputs, so I'll land this, but I'll fix it as a followup.

PR #4944, opened 2022-08-20
Converting multi-memories wasm tests to wast tests

ashleynh:
Removing the .wasm multi-memories tests as they can be easily represented in .wast format, which is easier to read/handle.


tlively (approved):

kripken (approved):

kripken:
I think we can delete `test/multi-memories-atomics64.wasm.fromBinary` etc. now?

ashleynh:
@kripken Those files were generated by auto_update_tests.py. ./check.py won't pass without them.

kripken:
Ah, I think the names might be confusing because they are so similar. `test/multi-memories-atomics64.wasm.fromBinary` is no longer needed, but `test/multi-memories-atomics64.wast.fromBinary` is. The only difference is `wasm / wast` in the middle. I verified that the former is no longer needed and `check.py` passes without it (but removing the latter would break it).

PR #4951, opened 2022-08-22
[TNH Fuzzing] Fix some equality checks that ignored effects

kripken:
Fuzzing with TrapsNeverHappen found a bug, and then reading similar code
I found another, where we check structural equality but ignored effects. Some
things look equal but may have different values at runtime:
```wat
(foo
  (call $x)
  (call $y)
)
```
The arms are identical structurally but due to side effects may not be identical
in value.

### DIFF
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -1019,7 +1019,7 @@ struct OptimizeInstructions
       if (auto* binary = curr->value->dynCast<Binary>()) {
         if ((binary->op == Abstract::getBinary(binary->type, Abstract::Mul) ||
              binary->op == Abstract::getBinary(binary->type, Abstract::DivS)) &&
-            ExpressionAnalyzer::equal(binary->left, binary->right)) {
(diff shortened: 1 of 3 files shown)


src/passes/OptimizeInstructions.cpp
```diff
@@ -2074,6 +2079,13 @@ struct OptimizeInstructions
     return areConsecutiveInputsEqualAndRemovable(left, right);
   }
 
+  // Similar to areConsecutiveInputsEqualAndFoldable, but only checks that they
+  // are equal (and not that they are foldable).
+  bool areConsecutiveInputsEqual(Expression* left, Expression* right) {
+    // TODO: optimize cases that must be equal but are *not* foldable.
```
tlively, line 2085: It would be helpful to provide an example of such a pattern here.

tlively (approved):

src/passes/OptimizeInstructions.cpp
kripken: I don't have one :smile: at least not a believable one. e.g. `(x++, x-1), (x--, x)` are equal (both return the original value of x) but we can't remove either one.

PR #4952, opened 2022-08-22
Materialize non-null externrefs in the fuzzer

tlively:
Some fuzzer initial contents contain non-nullable externrefs that cause the
fuzzer to try to materialize non-nullable externref values. Perviously the
fuzzer did not support this and crashed with an assertion failure. Fix the
assertion failure by instead returning a null cast to non-null, which will trap
at runtime but at least produce a valid module.


tlively:
* main
  * **PR #4952** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4952" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4952?utm_source=stack-comment).

kripken (approved):

PR #4955, opened 2022-08-22
Separate `func` into a separate type hierarchy

tlively:
Just like `extern` is no longer a subtype of `any` in the new GC type system,
`func` is no longer a subtype of `any`, either. Make that change in our type
system implementation and update tests and fuzzers accordingly.


tlively:
* main
  * **PR #4955** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4955" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4955?utm_source=stack-comment).

kripken (approved):
lgtm with fuzzing.

aheejin:
Question about the decision itself... What was the reason we decided to this and what is the purpose of `anyref` given that it is not _any ref_ anymore now?

tlively:
The reason we did this was because no one could come up with any concrete use case to have `func <: any`, so it was simpler to keep the status quo of having completely separate types like we had in the reference types proposal. Also, since we were going to split `extern` out into a separate type hierarchy anyways, folks were happy with the consistency of having `func` be separate.

tlively:
I ran the fuzzer for 3892 iterations before updating that test, so if there are no further functional changes necessary, I think it's ready to go.

tlively:
Oh, and to finish answering @aheejin's question: `any` is still the top type of all non-function wasm references like `eq`, `data`, `i31`, and all the struct and array types. Also, `extern` can be converted (i.e. "internalized") into `any` so that external references can be treated uniformly with internal references as well (but possibly with some explicit performance overhead on the conversion).

PR #4957, opened 2022-08-23
Simplify TrapsNeverHappen fuzzing

kripken:


kripken:
Actually I noticed a tiny possible bug here, fixing that while I have this open: we should include the line separator in the call line, as otherwise function names might be prefixes of each other and we could find the wrong one.

scripts/fuzz_opt.py
```diff
@@ -1082,15 +1082,16 @@ def handle_pair(self, input, before_wasm, after_wasm, opts):
                 # operations, etc.). in that case there is nothing for us to
                 # compare here; just leave.
                 return
-            call_end = before.index('\n', call_start)
+            # include the line separator in the index, as function names may
+            # be prefixes of each other
```
aheejin, line 1086: Not sure if I understand... What's an example of this case?

aheejin (approved):

scripts/fuzz_opt.py
kripken: Something like `$foo` and $foo2`. If we just search for `calling $foo` then we might end up finding the first characters of `calling $foo2` instead of `calling $foo`, since one is a prefix of the other.

PR #4959, opened 2022-08-23
[Wasm GC] Support non-nullable locals in the "1a" form

### LINKED ISSUE #4824: Non-nullable locals, option 1a
Now that "1a" is the final implementation of non-nullable locals in the spec, we should implement it. That approach does have the downside of extra work for tools like Binaryen, though, so it's not trivial to do. The problem is that a set allows gets only until the end of the current block, which means many natural code transforms "break" 1a, such as:
```wat
(local.set $x ..)
(block
  (local.get $x)
)
(local.get $x)

=>

(block
  (local.tee $x ..) ;; set+get => tee here
)
(local.get $x)
```

```wat
(local.set $x ..)
(local.get $x)

=>

(block
  (local.set $x ..)
  ..new code.. ;; new code added here, wrapped in a block
)
(local.get $x)
```

However we handle this, we need to add fixup logic for 1a, as well as specific workarounds in relevant passes (we don't want to add a local that will end up fixed up later anyhow, if that is worse, which some passes know). Aside from that, some possible approaches:

## No 1a validation in Binaryen IR

In this approach binaryen IR allows more things than normal wasm, in particular, non-nullable locals can be used arbitrarily and not just in the 1a form. To emit valid wasm binaries:

* Fix up 1a during binary writing. That is, when we find a local that does not validate, make it nullable and add `ref.as_non_null` on its gets (to keep the types of gets identical).
  * Extra `ref.as_non_null` may be removed, but this is too late to run the pass that does so.
* Fix up 1a as part of the pipeline, probably pretty late, and with the opt pass to remove `ref.as_non_null` after it.
  * Just running an opt pass or two may not validate - the user must remember to run the fixup pass.
* Do both of the above: fix up late in the pass pipeline, and also in binary writing. The binary writing fixups would hopefully never happen if the normal pass pipeline is run.

A downside is that we never validate that we are emitting a valid wasm (again, since we don't validate 1a). Also, intermediate states during the pass pipeline - before we get to the fixups - may not validate as 1a, which can be annoying during debugging.

We'd also need to decide what to do in the text format. Perhaps we wouldn't do fixups there, which would keep it 1:1 with Binaryen IR for testing.

## 1a validation in Binaryen IR

Here our validator adds 1a checks. Each pass must therefore emit valid 1a. Some options for that:

* Fix up after each pass we run, automatically, in the pass runner.
* Passes that might break validation must run a fixup.
  * This adds a few lines to most of our passes.

The extra work is something like 7% slower compilation overall. It's not a huge cost since it's basically 1a validation, which is simple and fast, and the data is in cache already since we just worked on it. Still, it's a noticeable slowdown. (Just on GC builds, though: when GC is off we can avoid any extra work.)

A downside here is that forcing 1a limits what we can do in intermediate steps. In theory allowing more there might help out even if we end up undoing it later (the other effects might remain). On J2Wasm and Dart at least this seems to be negligible, though it does happen.

Thoughts?

kripken:
An overview of this is in the README in the diff here (conveniently, it is near the
top of the diff). Basically, we fix up nn locals after each pass, by default. This keeps
things easy to reason about - what validates is what is valid wasm - but there are
some minor nuances as mentioned there, in particular, we ignore nameless blocks
(which are commonly added by various passes; ignoring them means we can keep
more locals non-nullable).

The key addition here is `LocalStructuralDominance` which checks which local
indexes have the "structural dominance" property of 1a, that is, that each get has
a set in its block or an outer block that precedes it. I optimized that function quite
a lot to reduce the overhead of running that logic after each pass. The overhead
is something like 2% on J2Wasm and 0% on Dart (0%, because in this mode we
shrink code size, so there is less work actually, and it balances out).

Since we run fixups after each pass, this PR removes logic to manually call the
fixup code from various places we used to call it (like eh-utils and various passes).

Various passes are now marked as `requiresNonNullableLocalFixups => false`.
That lets us skip running the fixups after them, which we normally do automatically.
This helps avoid overhead. Most passes still need the fixups, though - any pass
that adds a local, or a named block, or moves code around, likely does.

This removes a hack in SimplifyLocals that is no longer needed. Before we
worked to avoid moving a set into a try, as it might not validate. Now, we just do it
and let fixups happen automatically if they need to: in the common code they
probably don't, so the extra complexity seems not worth it.

Also removes a hack from StackIR. That hack tried to avoid roundtrip adding a
nondefaultable local. But we have the logic to fix that up now, and opts will
likely keep it non-nullable as well.

Various tests end up updated here because now a local can be non-nullable -
previous fixups are no longer needed.

Note that this doesn't remove the gc-nn-locals feature. That has been useful for
testing, and may still be useful in the future - it basically just allows nn locals in
all positions (that can't read the null default value at the entry). We can consider
removing it separately.

Fixes #4824


src/ir/type-updating.cpp
```diff
@@ -307,17 +330,18 @@ void handleNonDefaultableLocals(Function* func, Module& wasm) {
     if (!set->isTee() || set->type == Type::unreachable) {
       continue;
     }
-    auto type = func->getLocalType(set->index);
-    if (type.isNonNullable()) {
+    if (badIndexes.count(set->index)) {
```
tlively, line 333: It looks like we could remove the `if (!func->isVar(...` case above since it is implicit in this new check.

src/ir/type-updating.cpp
```diff
@@ -307,17 +330,18 @@ void handleNonDefaultableLocals(Function* func, Module& wasm) {
...
     }
   }
 
   // Rewrite the types of the function's vars (which we can do now, after we
   // are done using them to know which local.gets etc to fix).
-  for (auto& type : func->vars) {
-    type = getValidLocalType(type, wasm.features);
+  for (auto index : badIndexes) {
+    func->vars[index - func->getNumParams()] =
+      getValidLocalType(func->getLocalType(index), wasm.features);
```
tlively, line 344: It looks like this needs to be updated to handle tuple-typed locals that contain non-nullable references.

src/pass.h
```diff
@@ -386,6 +386,15 @@ class Pass {
...
   virtual bool invalidatesDWARF() { return false; }
 
+  // Whether this pass modifies Binaryen IR in ways that may require fixups for
+  // the "1a" form of non-nullable locals which is in the wasm spec. If the pass
+  // adds locals not in that form, or moves code around in ways that might break
+  // that validation, this must return true. In that case the pass runner will
+  // automatically run the necessary fixups afterwards.
+  //
+  // For more details see the LocalStructuralDominance class.
+  virtual bool requiresNonNullableLocalFixups() { return true; }
```
tlively, line 396: An alternative would be to default to `false` and depend on the fuzzer to find mistakes. That would make for a more annoying developer experience, but might be worth it to avoid introducing unnecessary performance overhead that would be hard to discover and eliminate.

tlively (commented):
Initial comments after reviewing everything before `src/passes`.

README.md
```diff
@@ -117,6 +117,22 @@ There are a few differences between Binaryen IR and the WebAssembly language:
     `(elem declare func $..)`. Binaryen will emit that data when necessary, but
     it does not represent it in IR. That is, IR can be worked on without needing
     to think about declaring function references.
+  * Binaryen IR allows non-nullable locals in the form that the wasm spec does,
+    nicknamed "1a", in which a `local.get` must be structurally dominated by a
```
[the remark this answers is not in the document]
kripken: Good point. Removed & edited.

README.md
```diff
@@ -117,6 +117,22 @@ There are a few differences between Binaryen IR and the WebAssembly language:
...
+    `local.set` in order to validate (that ensures we do not read the default
+    value of null). Despite being aligned with the wasm spec, there are some
+    minor details that you may notice:
+    * A nameless `Block` in Binaryen IR does not interfere with 1a validation.
+      Nameless blocks are never emitted into the binary format (we just emit
+      their contents), so we ignore them for purposes of non-nullable locals. As
+      a result you may see what seems to be code that should not validate per
+      the spec, but that difference will not exist in the binary format.
```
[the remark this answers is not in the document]
kripken: Hmm, we do have several other such issues - our text format is just not 100% compatible as our IR is not. Only the binary format is. I'll add a note though, fair enough.

src/ir/LocalStructuralDominance.cpp
```diff
@@ -0,0 +1,231 @@
...
+      }
+
+      // Note that we do not need to start a scope for the function body.
+      // Logically there is a scope there, but there is no code after it, so
+      // there is nothing to clean up when that scope exits, so we may as well
+      // not even create a scope. Just start walking the body now.
+      walk(func->body);
+    }
+
+    using Locals = SmallVector<Index, 5>;
```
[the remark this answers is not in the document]
kripken: I did some benchmarking and 5 seemed to work well (I just benchmarked stuff like 1, 2, 5, 10, 20, etc.).

IIUC, your second idea is to store the depth at which a local was set, as opposed to a stack of lists of cleanups? I explored approaches like that at some point, but it seemed more complex. I don't remember if I benchmarked it though.

src/ir/LocalStructuralDominance.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+    static void doLocalSet(Scanner* self, Expression** currp) {
+      auto index = (*currp)->cast<LocalSet>()->index;
+      if (!self->localsSet[index]) {
+        // This local is now set until the end of this scope.
+        self->localsSet[index] = true;
+        // If we are not in the topmost scope, note this for later cleanup.
+        if (!self->cleanupStack.empty()) {
+          self->cleanupStack.back().push_back(index);
+        }
```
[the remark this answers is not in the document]
kripken: Yes, this seemed to help in my measurements. Small functions with only toplevel content are pretty common, and adding the index requires loading info about `cleanupStack` anyhow, so the if is very cheap.

src/ir/LocalStructuralDominance.cpp
```diff
@@ -0,0 +1,231 @@
...
+            currp = &curr->cast<Try>()->body;
+            continue;
+          }
+
+          default: {
+            // Control flow structures have been handled. This is an expression,
+            // which we scan normally.
+            assert(!Properties::isControlFlowStructure(curr));
+            PostWalker<Scanner>::scan(self, currp);
+            return;
```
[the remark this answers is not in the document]
kripken: Not sure what you mean? `Unary` is not needed in this processing. Or do you mean `LocalSet` which has a single child? - that is already handled in the loop here.

src/ir/LocalStructuralDominance.cpp
```diff
@@ -0,0 +1,231 @@
...
+          }
+        }
+      }
+    }
+
+    // Only local.set needs to be visited.
+    void pushTask(TaskFunc func, Expression** currp) {
+      // Visits to anything but a set can be ignored, so only very specific
+      // tasks need to actually be pushed here.
```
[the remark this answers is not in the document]
kripken: I'll add a comment, good point.

I think we do want to inherit from PostWalker because we call `PostWalker<Scanner>::scan` which has a lot of useful logic we need - the normal scanning of content to walk children.

src/ir/local-structural-dominance.h
```diff
@@ -0,0 +1,95 @@
...
+// nicely, it can be validated in linear time.
+//
+// Concretely, this class finds which local indexes lack the structural
+// dominance property. It only looks at reference types and not anything else.
+// It can look at both nullable and non-nullable references, though, as it can
+// be used to validate non-nullable ones, and also to check if a nullable one
+// could become non-nullable and still validate.
+//
+// The property of "structural dominance" means that the set dominates the gets
+// using wasm's structured control flow constructs, like this:
```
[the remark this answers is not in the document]
kripken: Hmm, a quick grep shows we have 10x more places with lowercase than upper. If the upper is more correct, let's discuss that separately?

src/ir/local-structural-dominance.h
```diff
@@ -0,0 +1,95 @@
...
+    IgnoreNullable,
+  };
+
+  LocalStructuralDominance(Function* func,
+                           Module& wasm,
+                           Mode mode = ProcessNullable);
+
+  // Local indexes for whom a local.get exists that is not structurally
+  // dominated.
+  std::set<Index> nonDominatingIndexes;
```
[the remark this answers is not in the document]
kripken: It can work either way, I guess, but a benefit of a class is that there's a natural place to put the enum `Mode` - inside the class. Otherwise we'd want to add a new namespace or such, which seems clunky to me.

Spelling-wise, looks like an almost equal split in our existing code, 100 vs 120... Googling this, I get
```
"Indices" is used when referring to mathematical, scientific and statistical contexts.
It is used to refer to a numbers, symbols, and figures comparing a value to a
standard. "Indexes" is usually used in reference to written documents, such as
bibliographical or citation listings`
```
I don't know what to make of that :laughing: arguably our indexes are numbers, but
often they are references to things which is *sort of* like a citation listing...

src/ir/type-updating.cpp
```diff
@@ -280,18 +285,36 @@ void handleNonDefaultableLocals(Function* func, Module& wasm) {
...
   if (!hasNonNullable) {
+    // No non-nullable types exist in practice.
+    return;
+  }
+
+  // Non-nullable locals exist, which we may need to fix up. See if they
+  // validate as they are, that is, if they fall within the "1a" validation
+  // rules of the wasm spec. We do not need to modify such locals.
+  LocalStructuralDominance info(
+    func, wasm, LocalStructuralDominance::IgnoreNullable);
```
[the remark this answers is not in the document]
kripken: Good point, updated to `NonNullableOnly`.

src/ir/type-updating.cpp
```diff
@@ -280,18 +285,36 @@ void handleNonDefaultableLocals(Function* func, Module& wasm) {
...
+  std::unordered_set<Index> badIndexes;
+  for (auto index : info.nonDominatingIndexes) {
+    badIndexes.insert(index);
+
+    // LocalStructuralDominance should have only looked at non-nullable indexes
+    // since we told it to ignore nullable ones. Also, params always dominate
+    // and should not appear here.
+    assert(func->getLocalType(index).isNonNullable());
+    assert(!func->isParam(index));
```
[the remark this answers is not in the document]
kripken: > These assertions seem like they could be replaced with unit tests

https://twitter.com/johnregehr/status/1560717134065414144

Jokes aside I think readability is only improved - without this, the reader may need to go read that other file to see what the exact output from that function must be.

I think a default of an ordered set makes sense as it avoids possible nondeterminism issues by default (like the order of reporting validation errors, for example).

src/ir/local-structural-dominance.h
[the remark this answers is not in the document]
tlively: Sounds good 👍

src/ir/type-updating.cpp
kripken: Interesting idea. I think you are right that it would work 100% correctly that way, and is shorter code. OTOH it may be slower, since `isVar` is basically `index >= numParams` - faster than a set lookup (ordered or not).

kripken: Perhaps if we used `std::vector<bool>` instead of a set then this would be fast either way. OTOH the number of set bits would be 0 in almost all cases.

src/ir/local-structural-dominance.h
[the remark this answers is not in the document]
tlively: I would say we're firmly on the "technical" side of that technical-textual divide.

src/ir/type-updating.cpp
kripken: Good point, this looks like an existing bug. I wrote myself a note to do a followup.

src/pass.h
kripken: Good question. I'm not sure what's best.

Another option might be to make this pure virtual, so it forces all passes to define it, so we need to think about it at least.

src/ir/type-updating.cpp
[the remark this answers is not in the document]
tlively: Sounds good 👍

src/pass.h
tlively: I don't have a strong opinion here, and if I weren't thinking about performance so much I would definitely prefer this conservative default. The pure virtual idea seems nice.

src/ir/local-structural-dominance.h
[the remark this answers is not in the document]
kripken: Sounds good, renaming.

src/pass.h
kripken: Thinking on this more, given the overhead is as low as it is, I think the higher priority is to avoid correctness bugs for now. We can look into making this faster later on if that becomes relevant, which might mean flipping the default or going pure virtual.

src/ir/LocalStructuralDominance.cpp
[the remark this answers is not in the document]
tlively: Right, the second idea is basically to flatten the the stack of stacks into just a single stack, then keep a second stack to track the beginning index of each inner stack. My hope would be that having two stacks would reduce the number of allocations, although using `SmallVector` might do that well enough already.

src/ir/LocalStructuralDominance.cpp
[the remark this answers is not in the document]
tlively: Specifically, we could add this:

```
case Expression::Id::UnaryId: {
  // Immediately continue in the loop.
  currp = curr->cast<Unary>()->value;
  continue;
}
```

We could do this with any expression type that only has a single operand.

src/ir/type-updating.cpp
tlively: I would prefer the shorter code if we don't _strongly_ suspect (or measure) that it is faster this way. (It doesn't seem clear cut to me and my guess is that it doesn't make much of a difference).

But LGTM whichever way you decide.

src/ir/LocalStructuralDominance.cpp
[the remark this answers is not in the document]
kripken: I see, yes, that could be faster. I think it would add a bunch of extra code though, and need to be updated when we add new instructions, etc. I am also unsure how much it would help. I did measure something related, btw - I checked whether we pushed a single child on the stack (by measuring the size before and after) and if so, continued in the loop on that. That does what you said, but it does write and read from the stack, so it just saves the call. I saw no significant benefit there.

src/ir/type-updating.cpp
kripken: I do strongly suspect the current way is faster, just because a set lookup is fairly costly, isn't it? It's either using a tree and traversing it, or it's doing math to do hashing.

test/lit/non-nullable-locals.wast
```diff
@@ -0,0 +1,177 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
+
+;; RUN: foreach %s %t wasm-opt -all -S -o - | filecheck %s
+
+;; Tests for the "1a" form of non-nullable locals. This will likely be the final
```
tlively, line 5: Stray "1a" here.

test/lit/passes/coalesce-locals-gc.wast
```diff
@@ -1,7 +1,11 @@
 ;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
-;; RUN: wasm-opt %s --coalesce-locals -all -S -o - \
+;; RUN: wasm-opt %s --remove-unused-names --coalesce-locals -all -S -o - \
 ;; RUN:   | filecheck %s
 
+;; --remove-unused-names is run to avoid adding names to blocks. Block names
+;; can prevent 1a validation (we emit named blocks in the binary format, if we
```
tlively, line 6: Stray "1a"

test/lit/passes/heap2local.wast
```diff
@@ -2826,4 +2808,59 @@
...
+  ;; NOMNL-NEXT:  (ref.as_non_null
+  ;; NOMNL-NEXT:   (local.get $0)
+  ;; NOMNL-NEXT:  )
+  ;; NOMNL-NEXT: )
+  (func $non-nullable-local (result anyref)
+   (local $0 (ref $struct.A))
+   ;; The local.get here is in unreachable code, which means we won't do
+   ;; anything to it. But when we remove the local.set during optimization (we
+   ;; can replace it with new locals for the fields of $struct.A), we must make
+   ;; sure that validation of 1a still passes, that is, since the local.get is
```
tlively, line 2857: "1a"

test/lit/passes/local-subtyping.wast
```diff
@@ -1,7 +1,11 @@
 ;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
-;; RUN: wasm-opt %s --local-subtyping -all -S -o - \
+;; RUN: wasm-opt %s --remove-unused-names --local-subtyping -all -S -o - \
 ;; RUN:   | filecheck %s
 
+;; --remove-unused-names is run to avoid adding names to blocks. Block names
+;; can prevent 1a validation (we emit named blocks in the binary format, if we
```
tlively, line 6: "1a"

test/lit/passes/roundtrip-gc.wast
```diff
@@ -8,33 +8,38 @@
...
  ;; NOMNL-NEXT:  (call $help
- ;; NOMNL-NEXT:   (struct.new_default $\7bi32\7d)
- ;; NOMNL-NEXT:   (block $label$1 (result i32)
+ ;; NOMNL-NEXT:   (block (result (ref $\7bi32\7d))
+ ;; NOMNL-NEXT:    (local.set $0
+ ;; NOMNL-NEXT:     (struct.new_default $\7bi32\7d)
+ ;; NOMNL-NEXT:    )
  ;; NOMNL-NEXT:    (nop)
- ;; NOMNL-NEXT:    (i32.const 1)
+ ;; NOMNL-NEXT:    (local.get $0)
```
tlively, line 31: Is it expected that the code got worse after we re-enabled this optimization?

test/lit/passes/simplify-locals-gc-nn-1a.wast
```diff
@@ -0,0 +1,48 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
+;; RUN: wasm-opt %s --simplify-locals -all -S -o - | filecheck %s
+
+;; Tests for the "1a" form of non-nullable locals. In this form a local.set
```
tlively, line 4: "1a" here and in the file name.

test/lit/validation/bad-non-nullable-locals.wast
```diff
@@ -0,0 +1,64 @@
+;; Tests for the "1a" form of non-nullable locals. This will likely be the final
```
tlively, line 1: "1a"

test/lit/validation/nn-tuples.wast
```diff
@@ -1,17 +0,0 @@
-;; Test for non-nullable types in tuples
```
tlively, line 1: Why was this deleted?

src/ir/type-updating.cpp
tlively: I guess I would expect the math to be pretty fast, but `std::unordered_map` is still notoriously slow, so I guess you're probably right.

src/ir/LocalStructuralDominance.cpp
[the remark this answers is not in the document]
kripken: Ok, I implemented this and measured it. I don't see a noticeable difference in speed, it's basically the same. This approach is a little more code, and I think harder to understand, so I think I prefer the old way. See the commit before last for the code (last commit is a revert).

Amusingly, this work uncovered a bug in our SmallVector impl :fearful: PR shortly...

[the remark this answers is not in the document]
tlively: Oh wow, that's a good find!

src/ir/LocalStructuralDominance.cpp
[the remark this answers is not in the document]
tlively: If the loop doesn't help much, would be simpler to do without it entirely?

src/ir/LocalStructuralDominance.cpp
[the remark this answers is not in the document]
tlively: Sounds good, especially if it's not worth using the loop mechanism more.

test/lit/non-nullable-locals.wast
kripken: Ah, thanks, I removed "1a" from the src but not the tests... fixing now.

test/lit/passes/roundtrip-gc.wast
kripken: Yes, this is kind of a corner case. Before we disabled a bunch of stack IR opts to avoid such situations. Now we no longer disable them, which helps normally, but in corner cases like this, we end up with needing a local due to stack code. This doesn't matter in practice except for roundtripping a stacky binary without opts. This is not new to non-nullable locals - it has been the case for a long time in general, but now we just treat non-nullable locals like other locals.

test/lit/validation/nn-tuples.wast
kripken: This tested a case that now validates fine (non-nullable entries in tuples).

I guess it could be manually rewritten to show the opposite result, if we think that's important? (the auto-updater script can't do it for us here, and I'm not sure it's worth the work)

src/ir/LocalStructuralDominance.cpp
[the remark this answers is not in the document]
kripken: The loop does help in itself, in my measurements. Just adding those particular additional uses of the loop didn't help.

One reason the loop helps in the current code is that not only do we avoid a call, but also we can avoid pushing a task and later popping it and doing an indirect call. (see e.g. how Loop just calls `doBeginScope` rather than push it)

test/lit/passes/roundtrip-gc.wast
tlively: Makes sense, thanks!

src/ir/LocalStructuralDominance.cpp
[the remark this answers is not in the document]
tlively: I'm confused about why avoiding that task push, pop, and indirect call doesn't help as much when applied to other expression types. The existing code lgtm, though.

test/lit/validation/nn-tuples.wast
tlively: It would be nice to test that non-nullable tuples are ok now with the proper feature set enabled. If we only have one RUN line, this could be handled by the auto update script.

src/ir/LocalStructuralDominance.cpp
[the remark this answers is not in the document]
kripken: Yeah, it's not totally obvious to me either. It could depend on the frequency of the instructions or their size (which affects caching) for example. Overall, the current code has some added optimization complexity but not much, and each piece seemed worthwhile when I measured it.

test/lit/validation/nn-tuples.wast
kripken: Ok, I restored the test and made it just check validation.

tlively (approved):
Excellent, LGTM!

PR #4965, opened 2022-08-24
Update fuzzer to newer GC spec regarding JS interop

kripken:
Do not export functions that have types not allowed in the rules for
JS interop. Only very few GC types can be on the JS boundary atm.

### DIFF
--- a/src/tools/fuzzing/fuzzing.cpp
+++ b/src/tools/fuzzing/fuzzing.cpp
@@ -549,22 +549,19 @@ Function* TranslateToFuzzReader::addFunction() {
   // at least one, though, to keep each testcase interesting. Only functions
   // with valid params and returns can be exported because the trap fuzzer
   // depends on that (TODO: fix this).
+  bool validExportType = [](Type t) {
+    if (!t.isRef()) {
+      return true;
+    }
+    auto heapType = t.getHeapType();
+    return heapType == HeapType::ext || heapType ==HeapType::func || heapType == HeapType::string;
+  };
   bool validExportParams =
     std::all_of(paramType.begin(), paramType.end(), [](Type t) {
-      if (t.isRef() && t.getHeapType() == HeapType::any) {
-        // Anyref is not allowed in JS interop in the current GC spec.
-        return false;
-      }
-      return t.isDefaultable();
+      return validExportType(t) && t.isDefaultable();
     });
   bool validExportResults =
-    std::all_of(resultType.begin(), resultType.end(), [](Type t) {
-      if (t.isRef() && t.getHeapType() == HeapType::any) {
-        // Anyref is not allowed in JS interop in the current GC spec.
-        return false;
-      }
-      return true;
-    });
+    std::all_of(resultType.begin(), resultType.end(), validExportType);
   if (validExportParams && validExportResults &&
       (numAddedFunctions == 0 || oneIn(2)) &&
       !wasm.getExportOrNull(func->name)) {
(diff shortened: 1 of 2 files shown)


src/tools/fuzzing/fuzzing.cpp
```diff
@@ -545,13 +546,24 @@ Function* TranslateToFuzzReader::addFunction() {
...
+    return heapType == HeapType::ext || heapType == HeapType::func ||
+           heapType == HeapType::string;
+  };
+  bool validExportParams =
+    std::all_of(paramType.begin(), paramType.end(), [&](Type t) {
+      return validExportType(t) && t.isDefaultable();
     });
-  if (defaultableParams && (numAddedFunctions == 0 || oneIn(2)) &&
+  bool validExportResults =
+    std::all_of(resultType.begin(), resultType.end(), validExportType);
```
tlively, line 564: I don't think non-nullable values are allowed in the return position, either.

kripken: Hmm, they do work in the fuzzer itself (we don't mind receiving such values, the problem is manufacturing them in order to send them), and I don't see errors in V8? It accepts e.g. `(result (ref func))` on an export, and I'm not sure why it shouldn't?

tlively: This contradicts the draft document, which states, "Function parameters and results: only externref + funcref allowed." @jakobkummerow am I reading the intention there incorrectly?

jakobkummerow: Ha, good question!

We've only partially implemented this (because we've allowed more so far, dating back to older guesses about what the JS API might allow, and wanted to be careful about causing breakage), so the fact that you're not getting errors currently isn't a reliable datapoint.

The quoted phrasing disallows non-nullable refs. The _intention_ is that there will be no or absolutely minimal implicit checking at the boundary (for `funcref` we can't get around checking whether it's a function, but that ship has sailed; `stringref` can't get around a string check but is considered important enough to allow). To achieve that goal, it's sufficient if the _result_ types of exported Wasm functions are restricted to being _subtypes_ of `funcref` or `externref` or `stringref`, which in particular includes the non-nullable variants of these types. I don't feel strongly about how to resolve this:
(1) most strict option: require exactly `funcref` or `externref` or `stringref`, nothing else, for both parameters and results.
(2) slightly relaxed: only `funcref` or `externref` or `stringref` for parameters, optionally non-nullable results.
(3) maximally relaxed: parameters must be exactly `funcref` or `externref` or `stringref`, results must be subtypes of `funcref` or `externref` or `stringref` (which for the latter two is almost the same as option (2): it would also allow the respective `none`-type). 

Notably none of these options has an impact on run-time performance, they're just minor details of static type checking behavior. I assume that for module producers it doesn't make a practical difference, in particular because of implicit upcasting for any return values (e.g. a function's return type could be relaxed from `(result (ref $my_custom_func))` to `(result funcref)` without changing _anything_ in the function's body).

I think I'm weakly leaning towards option (1) because I don't yet see compelling reasons for anything else, but I could be convinced otherwise. WDYT?

Note 1: option (3) has some analogy to function subtyping: parameters must be "contravariant", i.e. must be supertypes of `funcref` or `externref`. Since these are top types, their only supertypes are themselves. Results must be "covariant", i.e. subtypes of F/E. Stringrefs break this analogy though: we don't want to allow `anyref` parameters, even though that's ([most likely](https://github.com/WebAssembly/stringref/issues/40)) a supertype of `stringref`.
Note 2: in all statements about required types, there's an implicit "...or one of the traditional Wasm types `i32` etc." that's left out for brevity.

kripken:
There may be more to discuss here, but what do you think about landing this to fix fuzzing @tlively ?

tlively (approved):
Let's at least add a comment on `validExportResults` mentioning that the intended behavior isn't quite settled.

PR #4968, opened 2022-08-25
Adding Multi-Memories Wasm Feature

ashleynh:
Adding multi-memories to the the list of wasm-features.


kripken (approved):
Code looks great!

Just so I'm clear on the direction, the idea is that when this feature is enabled, we may create new memories, but not otherwise? If so, then we may need someday to add a multi-table feature if we want to create new tables (but we just haven't needed that so far)?

tlively:
Support for multiple tables is included in the reference types feature, so we don't need a separate feature for it. Multiple memories is its own proposal/feature, though.

kripken:
@tlively Thanks, I didn't realize that.

PR #4969, opened 2022-08-25
[NFC] Simplify binary reading logic for functions

kripken:
We do a call to `updateMaps()` at the end of `processNames` anyhow, and so we
may as well call `addFunction` immediately (and the names will get fixed up in that
`updateMaps` later). The old code for some reason did that for function imports, but
not normal functions. It also stored them separately in temporary storage for some
unclear reason...

I happened to notice this while starting to add support for tag names.


ashleynh (approved):
Great work simplifying this code!

PR #4970, opened 2022-08-25
Add support for Tag names in the Names section

kripken:
This is the latest spec I believe? https://github.com/WebAssembly/exception-handling/pull/213/files

Existing tests look like they are enough here.


src/wasm/wasm-binary.cpp
```diff
@@ -3464,6 +3490,22 @@ void WasmBinaryBuilder::readNames(size_t payloadLen) {
...
       }
+    } else if (nameType == BinaryConsts::UserSections::Subsection::NameTag) {
+      auto num = getU32LEB();
+      NameProcessor processor;
+      for (size_t i = 0; i < num; i++) {
+        auto index = getU32LEB();
+        auto rawName = getInlineString();
+        auto name = processor.process(rawName);
+        if (index < wasm.tags.size()) {
+          wasm.tags[index]->setExplicitName(name);
```
aheejin, line 3501: We separately keep `globals` and `globalImports` for globals.. and others too like memories: https://github.com/WebAssembly/binaryen/blob/8b81405778eb85fccbfdbe789beeba5108cf1021/src/wasm-binary.h#L1552-L1556
https://github.com/WebAssembly/binaryen/blob/8b81405778eb85fccbfdbe789beeba5108cf1021/src/wasm/wasm-binary.cpp#L3436-L3439 

Why are we doing this, and do we need to do that for tags too?

kripken: Yeah, I noticed that too when I read this code. It's silly and unneeded, it turns out :smile: I opened a PR earlier #4969 for functions and I'll clean up the others as well in separate PRs (the PR deletes [that code](https://github.com/WebAssembly/binaryen/pull/4969/files#diff-e1f361a2d97280e82d323166f8f3528a4ba66bc5d68853759a12764b5b87dd5d)).

aheejin (approved):

PR #4972, opened 2022-08-26
[Parser] Parse functions and implicit types

tlively:
Implement function parsing, including parsing of locals and type uses. Also add
a new phase of parsing that iterates through type uses that do not have explicit
types to create any implicitly defined types and append them to the type index
space. This is important because the implicitly defined types may be referred to
by index elsewhere and because the legacy parser does not handle these
implicitly defined types correctly. Finally, maintain a map of implicit type use
locations to corresponding types for use in later phases of parsing.

### DIFF
--- a/src/wasm/wat-parser.cpp
+++ b/src/wasm/wat-parser.cpp
@@ -281,9 +281,11 @@ struct GlobalType {
   Type type;
 };
 
+// A signature type and parameter names (possibly empty), used for parsing
+// function types.
 struct TypeUse {
   HeapType type;
-  std::vector<NameType> ids;
+  std::vector<Name> names;
 };
 
 struct ImportNames {
@@ -864,9 +866,12 @@ struct ParseModuleTypesCtx : TypeParserCtx<ParseModuleTypesCtx>,
                                ParamsT* params,
                                ResultsT* results,
                                ParseInput& in) {
-    std::vector<NameType> ids;
+    std::vector<Name> ids;
     if (params) {
-      ids = *params;
+      ids.reserve(params->size());
+      for (auto& p : *params) {
+        ids.push_back(p.name);
+      }
     }
 
     if (type) {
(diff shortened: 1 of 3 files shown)


tlively:
* main
  * **PR #4972** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4972" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4972?utm_source=stack-comment).

src/wasm/wat-parser.cpp
```diff
@@ -282,6 +281,11 @@ struct GlobalType {
   Type type;
 };
 
+struct TypeUse {
```
kripken, line 284: Maybe add a comment here? Puzzling over this I don't have a good guess as to what a "type use" is that contains a type and a vector of name/type pairs.

src/wasm/wat-parser.cpp
```diff
@@ -573,10 +588,15 @@ struct ParseDeclsCtx : NullTypeParserCtx, NullInstrParserCtx {
   // The module element definitions we are parsing in this phase.
   std::vector<DefPos> typeDefs;
   std::vector<DefPos> subtypeDefs;
+  std::vector<DefPos> funcDefs;
   std::vector<DefPos> globalDefs;
 
+  // Positions of typeuses that might implicitly define new types.
```
kripken, line 594: Maybe give an example here? I assume an example could be something like the implicit type that we give to functions with an explicit one?

src/wasm/wat-parser.cpp
tlively: Will add a comment about how it is used, but it comes from here: https://webassembly.github.io/spec/core/text/modules.html#type-uses

src/wasm/wat-parser.cpp
tlively: For example, block signatures are syntactically `typeuse`s, so if the [abbreviation](https://webassembly.github.io/spec/core/text/modules.html#abbreviations) is used, the block signature may implicitly define a new type. That type doesn't have a name, but it can be referred to by index, so it is important to get its position in the type index space correct. The current parser does not handle this correctly.

src/wasm/wat-parser.cpp
kripken: As, so the vector is the parameter names? Makes sense.

src/wasm/wat-parser.cpp
kripken: Interesting.. I didn't know that was possible.

tlively:
I updated `TypeUse` to have a comment and to contain a vector of `Name` rather than `NameType`, since the parameter types are already encoded in the signature heap type.

kripken (approved):

PR #4974, opened 2022-08-26
[NFC] Simplify binary reading logic for tables

kripken:
Similar to #4969 but for tables.

(After this, we have these to clean up: globals, memories, and segments.)


ashleynh (approved):
Much simpler, well done!

PR #4975, opened 2022-08-26
Implement `extern.externalize` and `extern.internalize`

tlively:
These new GC instructions infallibly convert between `extern` and `any`
references now that those types are not in the same hierarchy.


tlively:
* main
  * **PR #4975** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/4975" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/4975?utm_source=stack-comment).

tlively:
cc @askeska-google

src/wasm.h
```diff
@@ -570,6 +570,11 @@ enum RefAsOp {
   RefAsI31,
 };
 
+enum ExternConversionOp {
+  Externalize,
+  Internalize,
+};
```
kripken, line 576: Another option might be to make this part of `RefAs`. This does receive a reference and return a reference, so it has the same shape, and is really quite similar aside from the name?

I don't feel strongly, though, and there is obviously some benefit to following the spec names.

test/lit/extern-conversions.wast
```diff
@@ -0,0 +1,33 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
+
+;; Check that string types are emitted properly in the binary format.
```
kripken, line 3:
```suggestion
;; Check that extern conversions are emitted properly in the binary format.
```

src/wasm-interpreter.h
```diff
@@ -1879,6 +1879,9 @@ class ExpressionRunner : public OverriddenVisitor<SubType, Flow> {
     }
     return value;
   }
+  Flow visitExternConversion(ExternConversion* curr) {
+    WASM_UNREACHABLE("unimp");
+  }
```
kripken, line 1884: You also need to implement the other place here in this file that returns a nonconstant flow (search for the other `visitStringNew`).

test/lit/extern-conversions.wast
```diff
@@ -0,0 +1,33 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
```
kripken, line 1: This file will need to be added to `fuzz_opt`'s ignore-list, like `strings.wast`, until the interpreter can execute this stuff.

kripken (commented):
lgtm % comments + fuzzing

test/lit/extern-conversions.wast
tlively: Heh, oops.

src/wasm.h
tlively: Yeah, actually I think I will just make these `RefAs` operations. They're similar enough and I don't know that the extra class is pulling its weight.

tlively:
So the fuzzer doesn't show any errors even when I add the new test to the list of important initial contents. I think all the externrefs make the fuzzer bail out before it tries to execute the instructions. I'll add the test to the list of skipped initial contents anyway, though.

src/wasm-interpreter.h
tlively: Why is this? None of the `ref.as_*` instructions are nonconstant. Is it just because these new instructions aren't implemented in the interpreter yet?

kripken: Yes, an instruction must either be fully implemented, or return "nonconstant" (which makes the interpreter give up on it, but not fatally error).

tlively:
@kripken, this should be ready for another look now that it's rewritten to use `RefAs`.

src/wasm-interpreter.h
```diff
@@ -1874,6 +1874,9 @@ class ExpressionRunner : public OverriddenVisitor<SubType, Flow> {
           trap("not an i31");
         }
         break;
+      case ExternInternalize:
+      case ExternExternalize:
+        WASM_UNREACHABLE("unimplemented extern conversion");
```
kripken, line 1879: I think we can just do a no-op here?
```suggestion
        // In a production VM interesting things happen here, but all references
        // are the same for us.
        break;
```

src/wasm-interpreter.h
```diff
@@ -2226,6 +2229,13 @@ class ConstantExpressionRunner : public ExpressionRunner<SubType> {
   Flow visitStringSliceIter(StringSliceIter* curr) {
     return Flow(NONCONSTANT_FLOW);
   }
+  Flow visitRefAs(RefAs* curr) {
+    // TODO: Remove this once interpretation is implemented.
+    if (curr->op == ExternInternalize || curr->op == ExternExternalize) {
+      return Flow(NONCONSTANT_FLOW);
+    }
+    return ExpressionRunner<SubType>::visitRefAs(curr);
+  }
```
kripken, line 2238: This diff can be removed if we agree on the above.

scripts/fuzz_opt.py
```diff
@@ -270,8 +270,10 @@ def is_git_repo():
 INITIAL_CONTENTS_IGNORE = [
     # not all relaxed SIMD instructions are implemented in the interpreter
     'relaxed-simd.wast',
-    # TODO fuzzer and interpreter support for strings
+    # TODO: fuzzer and interpreter support for strings
     'strings.wast',
+    # TODO: fuzzer and interpreter support for extern conversions
+    'extern-conversions.wast',
```
kripken, line 276: no need for this if we agree to the above

kripken (commented):
lgtm % that question and its implications.

src/wasm-interpreter.h
tlively: We don't currently have a way of representing externref values, and unlike the real `ref.as_*`, this is a conversion rather than a downcast, so I think we'll need to at least change the type of the `Literal` value we return.

kripken (approved):
Other options are less simple than I'd thought, so lgtm as it is following our offline discussion.

tlively:
We talked about this offline, and supporting externref values would be nontrivial (although hopefully still simple), so I'll merge this as-is. In particular, GCData contains a heap type, so we would be able to have an `externref` literal refer to a GCData without losing any type information, but i31 values are currently stored in the i32 field, so we would have to figure out an alternative way to store them in externrefs. This can probably be as simple as using a GCData with a single i32 field.

PR #4977, opened 2022-08-26
Multi-Memories wasm-split

ashleynh:
Adds an --in-secondary-memory switch to wasm-split tool that allows profile data to be stored in a separate memory from module main memory. With this option, users do not need to reserve the initial memory region for profile data and the data can be shared between multiple threads.

### DIFF
--- a/src/tools/wasm-split/instrumenter.cpp
+++ b/src/tools/wasm-split/instrumenter.cpp
@@ -78,7 +78,7 @@ void Instrumenter::addSecondaryMemory(size_t numFuncs) {
   secondaryMemory = Names::getValidMemoryName(*wasm, "split_data");
   // Create a memory with enough pages to write into
   size_t pages = (numFuncs + Memory::kPageSize - 1) / Memory::kPageSize;
-  wasm->addMemory(Builder::makeMemory(secondaryMemory, pages, pages));
+  wasm->addMemory(Builder::makeMemory(secondaryMemory, pages, pages, true));
 }
 
 void Instrumenter::instrumentFuncs() {
--- a/src/tools/wasm-split/instrumenter.cpp
+++ b/src/tools/wasm-split/instrumenter.cpp
@@ -75,7 +75,8 @@ void Instrumenter::addSecondaryMemory(size_t profileSize) {
     return;
   }
   if (!wasm->features.hasMultiMemories()) {
-    Fatal() << "error: --in-secondary-memory requires multi-memories to be enabled";
+    Fatal()
+      << "error: --in-secondary-memory requires multi-memories to be enabled";
   }
 
   secondaryMemory = Names::getValidMemoryName(*wasm, "split_data");
@@ -125,14 +126,18 @@ void Instrumenter::instrumentFuncs() {
       break;
     }
     case WasmSplitOptions::StorageKind::InMemory:
-    case WasmSplitOptions::StorageKind::InSecondaryMemory:{
+    case WasmSplitOptions::StorageKind::InSecondaryMemory: {
       if (!wasm->features.hasAtomics()) {
-        Fatal() << "error: --in-memory and --in-secondary-memory requires atomics to be enabled";
+        Fatal() << "error: --in-memory and --in-secondary-memory requires "
+                   "atomics to be enabled";
       }
       // (i32.atomic.store8 offset=funcidx (i32.const 0) (i32.const 1))
       Index funcIdx = 0;
       assert(!wasm->memories.empty());
-      Name memoryName = options.storageKind == WasmSplitOptions::StorageKind::InMemory ? wasm->memories[0]->name : secondaryMemory;
+      Name memoryName =
+        options.storageKind == WasmSplitOptions::StorageKind::InMemory
+          ? wasm->memories[0]->name
+          : secondaryMemory;
       ModuleUtils::iterDefinedFunctions(*wasm, [&](Function* func) {
         func->body = builder.makeSequence(
           builder.makeAtomicStore(1,
@@ -278,9 +283,13 @@ void Instrumenter::addProfileExport(size_t numFuncs, size_t profileSize) {
     case WasmSplitOptions::StorageKind::InSecondaryMemory: {
       // Copy the secondary memory into main memory for exporting the profile to
       // the user provided buffer
-      writeData = builder.blockify(
-          writeData,
-          builder.makeMemoryCopy(builder.makeConst(8), builder.makeConst(0), builder.makeConst(numFuncs), wasm->memories[0]->name, secondaryMemory));
+      writeData =
+        builder.blockify(writeData,
+                         builder.makeMemoryCopy(builder.makeConst(8),
+                                                builder.makeConst(0),
+                                                builder.makeConst(numFuncs),
+                                                wasm->memories[0]->name,
+                                                secondaryMemory));
       break;
     }
   }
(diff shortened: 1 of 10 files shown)


src/tools/wasm-split/instrumenter.cpp
```diff
@@ -62,6 +69,22 @@ void Instrumenter::addGlobals() {
...
+    return;
+  }
+  if (!wasm->features.hasMultiMemories()) {
+    Fatal()
+      << "error: --in-secondary-memory requires multi-memories to be enabled";
+  }
+
+  secondaryMemory = Names::getValidMemoryName(*wasm, "split_data");
+  // Create a memory with enough pages to write into
+  size_t pages = (profileSize + Memory::kPageSize - 1) / Memory::kPageSize;
```
ashleynh, line 84: I could change the pages to: (numFuncs + Memory::kPageSize - 1) / Memory::kPageSize.

src/tools/wasm-split/instrumenter.cpp
```diff
@@ -102,21 +125,27 @@ void Instrumenter::instrumentFuncs() {
       });
       break;
     }
-    case WasmSplitOptions::StorageKind::InMemory: {
+    case WasmSplitOptions::StorageKind::InMemory:
+    case WasmSplitOptions::StorageKind::InSecondaryMemory: {
       if (!wasm->features.hasAtomics()) {
-        Fatal() << "error: --in-memory requires atomics to be enabled";
+        Fatal() << "error: --in-memory and --in-secondary-memory requires "
```
tlively, line 131:
```suggestion
        Fatal() << "error: --in-memory and --in-secondary-memory require "
```

Or alternatively we could specialize the error message to use only the flag the user passed, which would be nicer IMO.

tlively (commented):
Excellent! LGTM modulo comments.

It also occurred to me that we will need to update Emscripten to actually send secondary shared memories to the WebWorkers it starts up. Otherwise there's no way for the different threads to write into the same memory. Does V8 support multiple memories behind a flag? If not, it will be difficult to write tests for that.

src/tools/wasm-split/instrumenter.cpp
ashleynh: Specializing is much nicer 👍🏾

src/tools/wasm-split/instrumenter.cpp
```diff
@@ -28,19 +28,22 @@ Instrumenter::Instrumenter(const WasmSplitOptions& options, uint64_t moduleHash)
 void Instrumenter::run(PassRunner* runner, Module* wasm) {
   this->runner = runner;
   this->wasm = wasm;
-  addGlobals();
+  size_t numFuncs = 0;
+  ModuleUtils::iterDefinedFunctions(*wasm, [&](Function*) { ++numFuncs; });
```
tlively, line 32: It would be nice to give this some space to hint visually that the two lines work together.
```suggestion

  size_t numFuncs = 0;
  ModuleUtils::iterDefinedFunctions(*wasm, [&](Function*) { ++numFuncs; });
  
```

src/tools/wasm-split/instrumenter.cpp
```diff
@@ -62,6 +65,22 @@ void Instrumenter::addGlobals() {
...
+  }
+  if (!wasm->features.hasMultiMemories()) {
+    Fatal()
+      << "error: --in-secondary-memory requires multi-memories to be enabled";
+  }
+
+  secondaryMemory = Names::getValidMemoryName(*wasm, "split_data");
+  // Create a memory with enough pages to write into
+  size_t pages = (numFuncs + Memory::kPageSize - 1) / Memory::kPageSize;
+  wasm->addMemory(Builder::makeMemory(secondaryMemory, pages, pages, true));
```
tlively, line 81: Here's how you can make the secondary module imported. We should choose some reasonable default for the import module and base name, but ideally make them each configurable via the command line.

```suggestion
  auto mem = Builder::makeMemory(secondaryMemory, pages, pages, true)
  mem->module = "???";
  mem->base = "???";
  wasm->addMemory(std::move(mem));
```

test/lit/wasm-split/invalid-options.wast
```diff
@@ -17,14 +17,10 @@
...
 
-;; --instrument cannot be used with --import-namespace
-;; RUN: not wasm-split %s --instrument --import-namespace=foo 2>&1 \
-;; RUN:   | filecheck %s --check-prefix INSTRUMENT-IMPORT-NS
-
 ;; --instrument cannot be used with --placeholder-namespace
 ;; RUN: not wasm-split %s --instrument --placeholder-namespace=foo 2>&1 \
 ;; RUN:   | filecheck %s --check-prefix INSTRUMENT-PLACEHOLDER-NS
-
+
```
ashleynh, line 23: I'll remove this before submitting

src/tools/wasm-split/split-options.cpp
```diff
@@ -185,10 +185,12 @@ WasmSplitOptions::WasmSplitOptions()
       [&](Options* o, const std::string& argument) { placeholderMap = true; })
     .add("--import-namespace",
          "",
-         "The namespace from which to import objects from the primary "
-         "module into the secondary module.",
+         "When provided as an option for module splitting, the namespace from "
+         "which to import objects from the primary "
+         "module into the secondary module. In instrument mode, refers to the "
+         "namespace from which to import the secondary memory.",
```
tlively, line 191:
```suggestion
         "namespace from which to import the secondary memory, if any.",
```

test/lit/wasm-split/instrument-in-secondary-memory.wast
```diff
@@ -0,0 +1,92 @@
...
+  (func $bar
+    (call $foo)
+  )
+  (func $baz (param i32) (result i32)
+    (local.get 0)
+  )
+)
+
+;; Check that a memory import has been added for secondary memory
+;; CHECK: (import "env" "profile-data" (memory $profile-data (shared 1 1)))
```
tlively, line 20: It would be good to additionally test that the options to change these import names also work.

tlively (approved):
LGTM modulo two small comments 👍

test/lit/wasm-split/instrument-in-secondary-memory.wast
ashleynh: Great catch, added another test file.

test/lit/wasm-split/instrument-in-secondary-memory-custom-names.wast
```diff
@@ -0,0 +1,92 @@
...
+  )
+  (func $baz (param i32) (result i32)
+    (local.get 0)
+  )
+)
+
+;; Check that a memory import has been added for secondary memory
+;; CHECK: (import "custom_env" "custom_name" (memory $custom_name (shared 1 1)))
+
+;; And the profiling function exported
```
tlively, line 22: Everything from here and below is the same as in the other test. Since we update these tests by hand, it's probably best to minimize the toil and remove the extra checks here.

tlively (approved):

test/lit/wasm-split/instrument-in-secondary-memory-custom-names.wast
ashleynh: Just want to call out that the "custom_name" memory name is used in three spots below on lines 31, 39, and 76.

ashleynh: I deleted everything that was duplicated in instrument-in-secondary-memory.wast and kept three lines prepended with a "CHECK:" to handle 31, 39, and 76.

PR #4978, opened 2022-08-26
Fix ModuleUtils::renameFunctions on RefFunc

kripken:
It had hardcoded handling of the table, but was missing RefFunc (and maybe more?)

Also rewrite it to be clearer (avoid special-casing the table) and faster (parallel).

### DIFF
--- a/src/ir/module-utils.h
+++ b/src/ir/module-utils.h
@@ -204,13 +204,9 @@ template<typename T> inline void renameFunctions(Module& wasm, T& map) {
 
     Updater* create() override { return new Updater(map); }
 
-    void visitCall(Call* curr) {
-      maybeUpdate(curr->target);
-    }
+    void visitCall(Call* curr) { maybeUpdate(curr->target); }
 
-    void visitRefFunc(RefFunc* curr) {
-      maybeUpdate(curr->func);
-    }
+    void visitRefFunc(RefFunc* curr) { maybeUpdate(curr->func); }
   };
 
   Updater updater(map);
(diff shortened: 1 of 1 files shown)


kripken:
Not really possible to test this atm as it's only used in Asyncify, and on functions that can't appear in the table (the special "runtime" functions in the Asyncify API). But if we use this in the future, we might hit the bug.

src/ir/module-utils.h
```diff
@@ -186,30 +186,34 @@ template<typename T> inline void renameFunctions(Module& wasm, T& map) {
...
+
+  // Update all references to it.
+  struct Updater : public WalkerPass<PostWalker<Updater>> {
+    bool isFunctionParallel() override { return true; }
+
+    T& map;
+
+    void maybeUpdate(Name& name) {
+      auto iter = map.find(name);
+      if (iter != map.end()) {
```
tlively, line 198: We can use C++17 fanciness to make this more pithy:
```suggestion
      if (auto iter = map.find(name); iter != map.end()) {
```
(Also we typically use `it` for iterators like this)

tlively (approved):
Can't comment inline, but there is a "to" missing in the last line of the comment above `renameFunctions`. Also the variable `F` should be renamed to `f`.

If we really want to test this, we could in principle write a gtest for it. But I'd be ok landing this without a test.

src/ir/module-utils.h
kripken: About `it` vs `iter`, either sgtm, but seems like `iter` is the current majority:
```
$ grep "auto iter =" `find -name "*.cpp"` | wc -l
94
$ grep "auto it =" `find -name "*.cpp"` | wc -l
55
```

tlively: oh interesting. I guess it's only the standard in code I frequently look at 😬

PR #4981, opened 2022-08-26
Fix test spacing after recent update

kripken:
A recent test update seems to have added or removed a space here. Running
the auto-updater script keeps showing this diff.


kripken:
Looks like `(rec` always has an extra space added, that we could probably remove in `Print.cpp`. I changed this PR to do that now.

tlively (approved):

tlively:
Thanks!

PR #4983, opened 2022-08-27
Adds assert to RemoveUnusedModuleElements

ashleynh:
Pass RemoveUnusedModuleElements does not have multi-memories support. This adds an assert to ensure the pass is not run if the module has multiple memories.


src/passes/RemoveUnusedModuleElements.cpp
```diff
@@ -236,6 +236,8 @@ struct RemoveUnusedModuleElements : public Pass {
     : rootAllFunctions(rootAllFunctions) {}
 
   void run(PassRunner* runner, Module* module) override {
+    assert(module->memories.size() <= 1);
```
kripken, line 239:
```suggestion
    // TODO: support multiple memories in this pass
    assert(module->memories.size() <= 1);
```

kripken (approved):

PR #4987, opened 2022-08-29
[NFC] Simplify binary reading logic for memories

kripken:
Similar to #4969 but for memories.


tlively (approved):
It looks like we follow a similar pattern with `elementSegments` and `dataSegments`. My guess is that those could be simplified as well.

kripken:
Correct, those are the last two things to fix here, after #4974 #4969 . I'll open PRs soon.

PR #5000, opened 2022-08-31
[Wasm GC] Fix evaluateKindCheck() on new externalize/internalize operations

### LINKED ISSUE #4999: UNREACHABLE when optimizing module with `extern` conversions
When optimizing [this module](https://github.com/WebAssembly/binaryen/files/9462694/m.wasm.gz) with `wasm-opt -all -O1 --hybrid m.wasm -o mo.wasm` (using da24ef6ed7055f64299fce22a051ba0e85284c23), I get the message:
```
unhandled BrOn
UNREACHABLE executed at ~/checkout/binaryen/src/ir/gc-type-utils.h:121!
Aborted
```
Unrelatedly, it looks like there's a copy-paste mistake in the error message here. It's the same for `BrOn`, `RefIs` and `RefAs`.

kripken:
Also add testing that they pass through the full optimizer.

Fixes #4999

Drive-by fixes to the text in the assertions, which was copy-pasted.


src/ir/gc-type-utils.h
```diff
@@ -117,8 +117,12 @@ inline EvaluationResult evaluateKindCheck(Expression* curr) {
       case RefAsI31:
         expected = I31;
         break;
+      case ExternInternalize:
+      case ExternExternalize:
+        // These instructions can never be removed.
+        return Unknown;
       default:
-        WASM_UNREACHABLE("unhandled BrOn");
+        WASM_UNREACHABLE("unhandled RefAs");
```
tlively, line 125: Let's get rid of the `default` case here so the compiler will flag this next time we add a new op.

test/lit/extern-conversions.wast
```diff
@@ -1,20 +1,40 @@
 ;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
 
 ;; Check that extern conversion instructions are emitted properly in the binary format.
+;; Also check that the optimizer does not break on this code.
 
 ;; RUN: wasm-opt %s -all --roundtrip -S -o - | filecheck %s
+;; RUN: wasm-opt %s -all -O1 --roundtrip -S -o - | filecheck %s --check-prefix OPTIM
```
tlively, line 7: We can probably just have the single `RUN` line with `-O1`.

tlively (approved):
LGTM % comments.

src/ir/gc-type-utils.h
kripken: Yeah, that might make sense. But then the compiler can't tell that `expected` is always set in all control flow paths - it doesn't know we treat warnings as errors. I can avoid that warning by setting a default that isn't needed, but that's a little less clean, requires a comment, and also may be slower to execute (the optimizer can't remove it; though that's probably trivial). Overall, I'm not sure which is better.

test/lit/extern-conversions.wast
kripken: Coverage would be the same, yeah, since nothing gets optimized here. I thought it might be more clear to explicitly test both cases, but fair enough, I'll remove.

src/ir/gc-type-utils.h
kripken: Which way do you prefer @tlively ?

tlively: Keeping the unreachable default seems fine. I think we do that in many places.

Issue #5015, opened 2022-09-04
ConstantExpressionRunner cannot handle `drop` case


HerrCai0907:
when I use `ConstantExpressionRunner` in `FlagValues::DEFAULT` mode to calculate code like
```wasm
(block $label$4 (result i32)
 (drop
  (local.tee $7
   (local.get $8)
  )
 )
 (i32.const 0)
)
```
current implement will return `NONCONSTANT_FLOW` instead of `i32.const 0` if local $8 is unknown.

HerrCai0907:
It can be optimized now.

PR #5016, opened 2022-09-04
Fix/constant expression runner visit drop

HerrCai0907:
fix: #5015

- Add a override of `visitDrop` in `ConstantExpressionRunner`
- Testcase result changed due to this code can be infer to `i32.const 0`
```wasm
  (block $label$4 (result i32)
   (drop
    (local.tee $7
     (local.get $8)
    )
   )
   (i32.const 0)
  )
```

### DIFF
--- a/src/wasm-interpreter.h
+++ b/src/wasm-interpreter.h
@@ -48,7 +48,7 @@ using namespace cashew;
 
 // Utilities
 
-extern Name WASM, RETURN_FLOW, NONCONSTANT_FLOW;
+extern Name WASM, RETURN_FLOW, NONCONSTANT_FLOW, RECOVERABLE_NONCONSTANT_FLOW;
 
 // Stuff that flows around during executing expressions: a literal, or a change
 // in control flow.
@@ -2010,7 +2010,7 @@ public:
     NOTE_ENTER("Drop");
     Flow value = ExpressionRunner<SubType>::visit(curr->value);
     if (!(flags & FlagValues::PRESERVE_SIDEEFFECTS)) {
-      value.clearIf(NONCONSTANT_FLOW);
+      value.clearIf(RECOVERABLE_NONCONSTANT_FLOW);
     }
     if (value.breaking()) {
       return value;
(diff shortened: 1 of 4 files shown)


src/wasm-interpreter.h
```diff
@@ -2006,6 +2006,17 @@ class ConstantExpressionRunner : public ExpressionRunner<SubType> {
     globalValues[name] = values;
   }
 
+  Flow visitDrop(Drop* curr) {
+    NOTE_ENTER("Drop");
+    Flow value = ExpressionRunner<SubType>::visit(curr->value);
+    if (!(flags & FlagValues::PRESERVE_SIDEEFFECTS)) {
+      value.clearIf(NONCONSTANT_FLOW);
```
kripken, line 2013: I don't remember this code well, but can't various things lead to `NONCONSTANT_FLOW`? I'm not sure they are all ok to ignore here, even if we ignore side effects.

test/lit/passes/dae-optimizing.wast
```diff
@@ -16,29 +16,19 @@
...
- ;; CHECK-NEXT:      (local.get $1)
- ;; CHECK-NEXT:     )
- ;; CHECK-NEXT:     (local.set $1
- ;; CHECK-NEXT:      (i32.const 0)
- ;; CHECK-NEXT:     )
- ;; CHECK-NEXT:     (br_if $label$2
- ;; CHECK-NEXT:      (local.get $0)
- ;; CHECK-NEXT:     )
- ;; CHECK-NEXT:     (f32.const 1)
+ ;; CHECK-NEXT:     (f32.const 0)
```
kripken, line 31: This seems odd - it used to return 1, and now returns 0?

HerrCai0907: Because this value is drop, so return whatever is Ok, just for keep align in true and false branch.
I have update the testcase and received this value by a global. The return value is 1 now.

src/wasm-interpreter.h
HerrCai0907: I think changing to a new error is better and can avoid this issue. So this behavior will be more robust and not be changed by some behavior in base class .

PR #5019, opened 2022-09-06
Fix build on Emscripten tip of tree

kripken:


tlively (approved):

CMakeLists.txt
```diff
@@ -447,6 +448,7 @@ if(EMSCRIPTEN)
   else()
     target_link_libraries(binaryen_js "-sEXPORT_ES6=1")
   endif()
+  target_link_libraries(binaryen_js "-sDEFAULT_LIBRARY_FUNCS_TO_INCLUDE=[$allocateUTF8OnStack]")
```
sbc100, line 451: you don't need the square brackets here or above.

kripken: I tried without the square brackets, and it hits an escaping issue with the `$`.

kripken: Oh, actually my local testing must have been wrong... it's still failing on CI due to escaping :cry:

kripken: I'm trying `$$` now for the escaping. I have to try on CI as for some reason my local CMake behaves differently.

Is there a simpler way to fix the escaping that I'm missing perhaps?

kripken: Looks like I've hit this problem before https://github.com/WebAssembly/binaryen/pull/4793#issuecomment-1180671654

sbc100: You can also use `-sEXPORTED_RUNTIME_METHODS=allocateUTF8OnStack` which doesn't use $-prefixing.

sbc100: Oh I see that you did that :)

PR #5024, opened 2022-09-07
[Strings] Begin support for String/StringView literals

kripken:
This adds enough to create null values of those types. That is enough for passes
that assume they can create a null for every type.

### DIFF
--- a/src/literal.h
+++ b/src/literal.h
@@ -32,6 +32,7 @@ namespace wasm {
 
 class Literals;
 struct GCData;
+using StringData = std::vector<uint8_t>;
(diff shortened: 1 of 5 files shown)


kripken:
friendly ping @tlively

tlively (approved):
Ideally I would have finished my PR changing how nulls work by now, but this LGTM in the meantime.

kripken:
Oh, sorry, was this PR waiting for that? I forgot that context, and just noticed it was an open PR...

It's not urgent, let's wait for your PR.

PR #5028, opened 2022-09-08
[Exceptions] Optimize in CodePushing even with exceptions thrown

kripken:
We had some concerns about this not working in the past, but thinking about it
now, I believe it is safe to do. Specifically, a throw is either like a break or a return -
either it jumps out to an outer scope (like a break) or it jumps out of the function
(like a return), and both breaks and returns have already been handled here.

This change has some nice effects on J2Wasm output, where there are quite a
lot of throws, which we can now optimize around.

### DIFF
--- a/src/passes/CodePushing.cpp
+++ b/src/passes/CodePushing.cpp
@@ -161,11 +161,22 @@ private:
     // everything that matters if you want to be pushed past the pushPoint
     EffectAnalyzer cumulativeEffects(passOptions, module);
     cumulativeEffects.walk(list[pushPoint]);
-    // it is ok to ignore the branching here, that is the crucial point of this
-    // opt
-    // TODO: it would be ok to ignore thrown exceptions here, if we know they
-    //       could not be caught and must go outside of the function
+    // It is ok to ignore branching out of the block here, that is the crucial
+    // point of this optimization. That is, we are in a situation like this:
+    //
+    // {
+    //   x = value;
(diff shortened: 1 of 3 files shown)


src/passes/CodePushing.cpp
```diff
@@ -161,11 +161,24 @@ class Pusher {
...
+    //
+    // If the branch is taken, then that's fine, it will jump out of this block
+    // and reach some outer scope, and in that case we never need x at all
+    // (since we've proven before that x is not used outside of this block, see
+    // numGetsSoFar which we use for that). Similarly, the break could be a
+    // return and that would be ok as well, and also an exception (if the
+    // exception is caught in the function, it must be outside the block, which
+    // is just like a break; and if it is not caught, it is just like a return).
     cumulativeEffects.ignoreBranches();
+    cumulativeEffects.throws_ = false;
```
aheejin, line 181: This is the only location `EffectAnalyzer.ignoreBranches` is used. Should we change this to `ignoreControlFlow` or something and make it include `throws_`?

aheejin (approved):
Other that that LGTM. Sorry for the late reply!

src/passes/CodePushing.cpp
kripken: Good idea, thanks! Done.

I also made it clear `delegateTargets`. After that, this method parallels `transfersControlFlow()`.

PR #5029, opened 2022-09-09
Fix compilation errors with C++20

ilya-biryukov:
In C++20 makes operator calls become ambiguous wrt to a rewritten version of itself with swapped candidates.

See https://gcc.godbolt.org/z/5q7GTr3bM for an example.

This commit fixes an occurence of this breakage in `parent_index_iterator.h`. Note that the actual breakage is in `passes/Poppify.cpp` that exposes the constraints of comparison operator for `std::reverse_iteratot<Iterator>`.

### DIFF
--- a/src/support/parent_index_iterator.h
+++ b/src/support/parent_index_iterator.h
@@ -49,10 +49,10 @@ template<typename Parent, typename Iterator> struct ParentIndexIterator {
 
   Iterator& self() { return *static_cast<Iterator*>(this); }
 
-  bool operator==(const Iterator& other) const {
+  bool operator==(const ParentIndexIterator& other) const {
     return index == other.index && parent == other.parent;
   }
-  bool operator!=(const Iterator& other) const { return !(*this == other); }
+  bool operator!=(const ParentIndexIterator& other) const { return !(*this == other); }
   Iterator& operator++() {
     ++index;
     return self();


kripken:
Thanks for the PR!

Hopefully someone that knows C++ better than me can take a look here, though this does seem safe to land.

tlively:
+1 seems safe to land, but I would also like to understand this more. @ilya-biryukov, do you have a link to cppreference or even the spec that explains what's going wrong here?

ilya-biryukov:
The issue is that C++20 adds [rewritten candidates](https://en.cppreference.com/w/cpp/language/overload_resolution) with reversed parameters. These sometimes happen to be ambiguous with themselves.
In particular, in this case for `operator ==(const Base*/*this*/, const Derived&)` we additionally get a rewritten `operator==(const Derived&, const Base*/*this*/)`. Since none of the two is more specialized if the arguments types are `(Derived, Derived)`, the call happens to be ambiguous.

However, when sending the patch I was not aware that this is considered to be a defect in the C++ standard and there is a recently accepted [proposal](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2468r2.html) that fixes the issue for that particular code pattern.

The [fix for Clang](https://reviews.llvm.org/D134529) is on the way, so there is no need for this change. Although, I would still suggest to apply the same diff (maybe with a different description) as a best practice to make sure the types of both parameters of `operator==` are the same. That does not cause any immediate issues, though, so it's a purely stylistic suggestion.

tlively:
Thanks for the additional context, @ilya-biryukov! I think using `Iterator` here is helpful for conveying that this is meant to be used with the CRTP subtype, so I'll close this PR, but thanks for the suggestion!

tlively:
* main
  * **PR #5029** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5029" target="_blank">Graphite</a>  👈
    * **PR #5229** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5229" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5029?utm_source=stack-comment).

tlively:
Landed as #5567

PR #5030, opened 2022-09-09
Remove typed-function-references feature

tlively:
In practice typed function references will not ship before GC and is not
independently useful, so it's not necessary to have a separate feature for it.
Roll the functionality previously enabled by --enable-typed-function-references
into --enable-gc instead.

This also avoids a problem with the ongoing implementation of the new GC bottom
heap types. That change will make all ref.null instructions in Binaryen IR refer
to one of the bottom heap types. But since those bottom types are introduced in
GC, it's not valid to emit them in binaries unless unless GC is enabled. The fix
if only reference types is enabled is to emit (ref.null func) instead
of (ref.null nofunc), but that doesn't always work if typed function references
are enabled because a function type more specific than func may be required.
Getting rid of typed function references as a separate feature makes this a
nonissue.


tlively:
* main
  * **PR #5030** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5030" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5030?utm_source=stack-comment).

kripken (approved):
lgtm with fuzzing.

tlively:
Fuzzed for almost 5k iteration with https://github.com/WebAssembly/binaryen/pull/5031 applied.

kripken:
@tlively

A fuzz failure started with this landing:

```
$ wasm-opt input.dat -ttf  --all-features --nominal --disable-multivalue --initial-fuzz=input.wat --roundtrip
[parse exception: bad heap type: expected (array_subtype (mut i8) data) but found i32 (at 0:3155)]
Fatal: error in parsing wasm binary
```

[inputs.zip](https://github.com/WebAssembly/binaryen/files/9558647/inputs.zip)

tlively:
That looks like it's possibly fixed by https://github.com/WebAssembly/binaryen/pull/5031. Did you try applying that? Will investigate locally as well.

tlively:
If you apply the fuzzing.cpp changes from this PR to previous commits (to get the same ttf output) the problem reproduces, so this isn't a new bug from this commit. Unfortunately the bug reproduces on 8e1abd1 and carrying fuzzer patches back beyond that to try to find the real source of the bug doesn't seem likely to work.

Poking around in a debugger shows that we are in the `skipUnreachableCode` mode of the binary builder and that we are trying to parse an array.get_u for which we have read a `ref` of type `i32` and an `index` of type `unreachable`. The program only has a single array.get_u and it had proper types when it was originally written, so I think there's a bug somewhere in our handling of unreachables in the binary builder.

PR #5037, opened 2022-09-13
[C API] Add getters and setters for various expressions

dcodeIO:
Covers `CallRef`, `RefTest`, `RefCast`, `BrOn`, `StructNew`, `StructGet`, `StructSet`, `ArrayNew`, `ArrayInit`, `ArrayGet`, `ArraySet`, `ArrayLen`, `ArrayCopy`, `StringNew`, `StringConst`, `StringMeasure`, `StringEncode`, `StringConcat`, `StringEq`, `StringAs`, `StringWTF8Advance`, `StringWTF16Get`, `StringIterNext`, `StringIterMove`, `StringSliceWTF`, `StringSliceIter`.

### DIFF
--- a/src/binaryen-c.cpp
+++ b/src/binaryen-c.cpp
@@ -1606,6 +1606,20 @@ BinaryenExpressionRef BinaryenI31Get(BinaryenModuleRef module,
   return static_cast<Expression*>(
     Builder(*(Module*)module).makeI31Get((Expression*)i31, signed_ != 0));
 }
+BinaryenExpressionRef BinaryenCallRef(BinaryenModuleRef module,
+                                      BinaryenExpressionRef target,
+                                      BinaryenExpressionRef* operands,
+                                      BinaryenIndex numOperands,
+                                      BinaryenType type,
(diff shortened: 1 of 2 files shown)


dcodeIO:
Random observation: `StringNew` and `StringEncode` utilize `ptr` expressions in their linear memory variants, but have no attached memory name for multi-memory support like other such expressions.

kripken (approved):
Thanks!

kripken:
I still think it would be a good idea to autogenerate these, but I haven't found the time to look into that. After this PR, perhaps there won't be a big number of new instructions for a while... so autogenerating might be less urgent. Though it could help on the JS API side.

kripken:
> StringNew and StringEncode utilize ptr expressions in their linear memory variants, but have no attached memory name for multi-memory support like other such expressions.

Ah, I think what happened is I added these around when @ashleynh was working on multi-memories, and because they landed around the same time they don't have multi-memory support yet. That would be good to add, though not urgent I suppose.

PR #5039, opened 2022-09-14
Effects: Clarify trap effect meaning, and consider infinite loops to trap due to timeout

kripken:
I think this simplifies the logic behind what we consider to trap. Before we had kind of
a hack in `visitLoop` that now has a more clear reasoning behind it.

OTOH this does prevent us from optimizing out infinite loops with no other effects,
```js
while (1) {}
```
I'm not sure if that matters or not. Also, in traps-never-happens mode we can still
optimize out such loops - but it's not clear if we'll ever be able to use that mode on
LLVM output (see related discussion here: https://github.com/emscripten-core/emscripten/issues/17732).

Thoughts?


tlively:
"but even they will error before the heat death of the universe, which is a kind of trap." 😂 

I agree that treating infinite looping as a branching side effect doesn't make much sense, but I don't think treating it as trapping makes more sense. In particular it makes the issue I'm worried about on https://github.com/emscripten-core/emscripten/issues/17732 worse; if infinite loops are traps, then traps-never-happen also assumes infinite loops never happen, which does not seem like what we want. I firmly believe that we should be able to run with traps-never-happen on LLVM output since LLVM output only traps on UB (or in other niche cases we can ignore like calls to __builtin_trap) and that anything that prevents us from being able to safely do that should be considered a bug in our treatment of traps-never-happen. Infinite loops without any other side effects [are UB](https://en.cppreference.com/w/cpp/language/memory_model#:~:text=the%20cache%20line)-,Progress%20guarantee,-In%20a%20valid) in C/C++, but we shouldn't mark legal infinite loops as trapping.

Could we add a new kind of effect for infinite loops instead?

kripken:
Interesting points. Yeah, maybe a new effect type is best. However, I'm not sure yet.

Getting to the "webserver infinite loop" mentioned in that other issue, there is an infinite loop there, but it has side effects in it - it's a webserver, it's doing stuff. So TNH might ignore the "effect" of the infinite loop, but the other effects would keep the loop alive. The problem only happens if TNH tries to infer from the trap after the loop that the loop isn't even reached.

Thinking about that now, I think the right rule might be **TNH can never remove an import call**, not even this:
```wat
(call $import) ;; imported function - call cannot be removed, but anything else here could
(unreachable) ;; assumed to never be reached in TNH
```
This solves the webserver case in a logical way: the webserver doesn't actually infinite loop, rather eventually it will call an import that exits the process. That's why we can't remove all the code.

It's possible we'd want to remove *some* import calls some day, however. But perhaps we can annotate "removable imports" if so.

Another thing to consider, if we add a new effect type "infinite loops" then I think we'd want a flag "infinite loops never happen" just like "traps never happen" (for example, ILNH would help remove calls to functions whose only effect is to run a loop to do some pure computation). I'd prefer to have a single flag for both if we can.

tlively:
> I think the right rule might be **TNH can never remove an import call**

Yes, that sounds appropriate, independent of whether the import call is in a loop. Note that this rule means TNH can't remove normal calls either unless it has first proved that the normal call cannot transitively call an import.

For loops that may not exit, I would additionally restrict TNH from removing any atomic write. In a multithreaded context it's entirely possible to have an infinite loop doing useful work in memory without calling imports and the C/C++ memory model explicitly allows for this. It's also possible that a program may be communicating with the host via shared memory rather than through imports.

I would still allow atomic writes that always precede a trap to be removed, though.

kripken:
Ok, this should now be ready for review. It also documents the recent discussions about trapsNeverHappens mode in comments, so that with all these comment changes I think we have the full plan moving forward here.

test/lit/passes/vacuum-tnh.wast
```diff
@@ -117,12 +190,41 @@
...
+  ;; NO_TNH-NEXT:   )
+  ;; NO_TNH-NEXT:  )
+  ;; NO_TNH-NEXT: )
+  (func $drop-loop
+    ;; A loop has effects, since it might infinite loop (and hit a timeout trap
+    ;; eventually), so we do not vacuum it out unless we are ignoring traps.
+    (drop
+      (loop $loop (result i32)
+        (br_if $loop
+          (i32.const 1)
```
tlively, line 224: What happens if the loop contains atomic operations?

test/wasm2js/br_table_to_loop.2asm.js.opt
```diff
@@ -15,7 +15,7 @@ function asmFunc(importObject) {
  var nan = NaN;
  var infinity = Infinity;
  function $0() {
-  while (1) continue;
+  abort();
```
tlively, line 18: Is this what we want to happen here?

tlively (approved):
LGTM % comments.

test/lit/passes/vacuum-tnh.wast
kripken: Atomics are an additional effect, so it wouldn't optimize out. I added a testcase now.

test/wasm2js/br_table_to_loop.2asm.js.opt
kripken: Yes, it's ok to do this since the only effect of an infinite loop is "trap" now, and we can replace traps. This is a smaller replacement so it does it.

Practically this is also fine since it replaces hanging forever with an immediate error, so it's not breaking useful code, it's just eagerly erroring basically.

tlively (approved):

PR #15, opened 2015-12-19
Initial start on CMakeLists.txt.

waywardmonkeys:
This is not yet complete as it doesn't perform the emscripten
build stage, nor does it attempt to run tests.

This is work in progress on issue #2.

Issue #14, opened 2015-12-19
Warning due to uninitialized value


waywardmonkeys:
I get this on OS X when I build with my cmake build:

```
../src/s2wasm.h:957:65: warning: reference 'o' is not yet bound to a value when used here [-Wuninitialized]
      AsmConstWalker(S2WasmBuilder* parent) : parent(parent), o(o) {}
```

kripken:
Thanks, that was partially-removed debug code.

Issue #13, opened 2015-12-18
Implement binary format


kripken:
There is a proposal for a wasm binary format, links in the design repo.

We should implement it in a new header (`wasm-binary.h` perhaps), and add tests. Might use the tests from the v8-native repo, although that's been merged to upstream v8, so maybe we'd need to look there. Or, maybe we'd just use our existing tests and see we can translate them to binary and back?

JSStats:
The V8 binary format might be a waste of time, and they already have a wast to binary encoder for testing, so perhaps just check that it can encode the asm2wasm output.

If you are working on the binary format then here's some background that I absorbed looking at publicly released prototypes and a little experimentation.

The polyfill-prototype-1 still seems a useful guide and it used a number of techniques to compress the opcode encoding but it is now obsolete because it was designed for a fixed set of opcodes and the current design appears to require supporting an arbitrary set of opcodes.

It used separate opcode tables for a child branch of the AST based on the immediate parent opcode, using the type I32, F32, and F64 to distinguish between opcode tables. This would appear to generalize to an adaptive tree based compression approach.

It used a byte size encoding which might be very significant for another 'string' based compression layer to do well with the output. However it does appear that a variable size Huffman encoding could do much better, and it's not clear if a general 'string' based compressor could predict the patterns nearly as well from a linearised AST.

It uses a one byte encoding for a few high frequency opcodes that includes an short immediate within this byte. Bit 7 encodes between the immediate format and the non-immediate format. The immediate format includes a 2 bit opcode field and a 5 bit immediate field. The non-immediate format includes a 7 bit opcode field.

Some quick checking shows that considering both the parent opcode plus the argument position can give much better prediction for some high frequency opcodes. A simple Huffman encoding could compress these child opcodes very well above. Using a general 'string' compression algorithm would miss some of these patterns because there might be other branches between a parent opcode and a child opcode breaking up the 'string'.

kripken:
I think it's still worth doing the current binary proposal. Then our binary output can be tested with the other tools. And we can track the spec as it changes, continuously over time. It's likely the final spec will not be radically different, and so that way we'll move towards it along with everything else.

kripken:
I did a bunch of work on this, it's in `src/wasm-binary.h`. It should be most of the code to write the binary format, but it's not complete, and it's written but not compiled or tested.

I'll get back to this when I can, but if someone wants to work on this, that would be great. Getting it to write the binary format shouldn't be much more work.

[...]

PR #12, opened 2015-12-18
fix typo in readme

loganchien:
It seems that

"compiling asm.js to WebAssembly (the opposite direction of `asm2wasm`)"

is a typo of

"compiling WebAssembly to asm.js (the opposite direction of `asm2wasm`)"

Please have a look.  Thanks!

PR #5043, opened 2022-09-15
[opt] let OptimizeAddedConstants handle memory64 well

dongAxis:
hi all,

    When program is compile with wasm64, the addr and offset of load and store should be i64.
   But OptimizeAddedConstants dose not handle well, so I use this patch to fix that.
   Sample wasm will attached.

Best Regards
Yifeng


dongAxis:
Since it is wip MR, so i do not write unit test. If this patch accept, I will write more unit test

dongAxis:
attached sample file.
When we use binaryen to optimize this file, we will got an assert failed.
The backtrace will look like:

`
➜  ~ lldb "/Users/bytedance/Desktop/Code/navi_compiler/upstream/bin/wasm-opt" -- --strip-dwarf --post-emscripten -O3 --low-memory-unused --zero-filled-memory --strip-producers a.out.wasm -o a.out.wasm.opt --detect-features
(lldb) target create "/Users/bytedance/Desktop/Code/navi_compiler/upstream/bin/wasm-opt"
Current executable set to '/Users/bytedance/Desktop/Code/navi_compiler/upstream/bin/wasm-opt' (arm64).
(lldb) settings set -- target.run-args  "--strip-dwarf" "--post-emscripten" "-O3" "--low-memory-unused" "--zero-filled-memory" "--strip-producers" "a.out.wasm" "-o" "a.out.wasm.opt" "--detect-features"
(lldb) r
Process 62776 launched: '/Users/bytedance/Desktop/Code/navi_compiler/upstream/bin/wasm-opt' (arm64)
Assertion failed: (type == Type::i32), function geti32, file literal.h, line 259.
Process 62776 stopped
* thread #4, stop reason = hit program assert
    frame #4: 0x000000010249630c libbinaryen.dylib`wasm::MemoryAccessOptimizer<wasm::OptimizeAddedConstants, wasm::Load>::optimizeConstantPointer() (.cold.1) + 40
libbinaryen.dylib`wasm::MemoryAccessOptimizer<wasm::OptimizeAddedConstants, wasm::Load>::optimizeConstantPointer() (.cold.2):
->  0x10249630c <+0>:  stp    x29, x30, [sp, #-0x10]!
    0x102496310 <+4>:  mov    x29, sp
    0x102496314 <+8>:  adrp   x0, 496
    0x102496318 <+12>: add    x0, x0, #0x803            ; "cast"
Target 0: (wasm-opt) stopped.
(lldb) bt
* thread #4, stop reason = hit program assert
    frame #0: 0x000000019fcced98 libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x000000019fd03ee0 libsystem_pthread.dylib`pthread_kill + 288
    frame #2: 0x000000019fc3e340 libsystem_c.dylib`abort + 168
    frame #3: 0x000000019fc3d754 libsystem_c.dylib`__assert_rtn + 272
  * frame #4: 0x000000010249630c libbinaryen.dylib`wasm::MemoryAccessOptimizer<wasm::OptimizeAddedConstants, wasm::Load>::optimizeConstantPointer() (.cold.1) + 40
    frame #5: 0x0000000101eb40e0 libbinaryen.dylib`wasm::MemoryAccessOptimizer<wasm::OptimizeAddedConstants, wasm::Load>::optimizeConstantPointer() + 168
    frame #6: 0x0000000101eb3f14 libbinaryen.dylib`wasm::MemoryAccessOptimizer<wasm::OptimizeAddedConstants, wasm::Load>::optimize() + 60
    frame #7: 0x0000000101eb34a8 libbinaryen.dylib`wasm::Walker<wasm::OptimizeAddedConstants, wasm::UnifiedExpressionVisitor<wasm::OptimizeAddedConstants, void> >::doVisitLoad(wasm::OptimizeAddedConstants*, wasm::Expression**) + 60
    frame #8: 0x0000000101eb101c libbinaryen.dylib`wasm::Walker<wasm::OptimizeAddedConstants, wasm::UnifiedExpressionVisitor<wasm::OptimizeAddedConstants, void> >::walk(wasm::Expression*&) + 160
    frame #9: 0x0000000101eb4a28 libbinaryen.dylib`wasm::OptimizeAddedConstants::doWalkFunction(wasm::Function*) + 216
    frame #10: 0x0000000101eb0dd0 libbinaryen.dylib`wasm::WalkerPass<wasm::PostWalker<wasm::OptimizeAddedConstants, wasm::UnifiedExpressionVisitor<wasm::OptimizeAddedConstants, void> > >::runOnFunction(wasm::PassRunner*, wasm::Module*, wasm::Function*) + 32
    frame #11: 0x0000000101ceeab4 libbinaryen.dylib`wasm::PassRunner::runPassOnFunction(wasm::Pass*, wasm::Function*) + 552
    frame #12: 0x0000000101cef890 libbinaryen.dylib`std::__1::__function::__func<wasm::PassRunner::run()::$_1::operator()() const::'lambda'(), std::__1::allocator<wasm::PassRunner::run()::$_1::operator()() const::'lambda'()>, wasm::ThreadWorkState ()>::operator()() + 108
    frame #13: 0x000000010229665c libbinaryen.dylib`wasm::Thread::mainLoop(void*) + 116
    frame #14: 0x0000000102297398 libbinaryen.dylib`void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void (*)(void*), wasm::Thread*> >(void*) + 44
    frame #15: 0x000000019fd0426c libsystem_pthread.dylib`_pthread_start + 148
(lldb)
`

src/passes/OptimizeAddedConstants.cpp
```diff
@@ -222,7 +242,7 @@ template<typename P, typename T> class MemoryAccessOptimizer {
 
   // Sees if we can optimize a particular constant.
   Result canOptimizeConstant(Literal literal) {
-    auto value = literal.geti32();
+    uint64_t value = memory64 ? literal.geti64() : literal.geti32();
```
kripken, line 245:
```suggestion
    uint64_t value = literal.getInteger();
```
There is a helper for this pattern.

src/passes/OptimizeAddedConstants.cpp
```diff
@@ -123,12 +126,29 @@ template<typename P, typename T> class MemoryAccessOptimizer {
...
+          uint32_t base = c->value.geti32();
+          uint32_t offset = curr->offset;
+          if (uint64_t(base) + uint64_t(offset) < (uint64_t(1) << 32)) {
+            c->value = c->value.add(Literal(uint32_t(curr->offset)));
+            curr->offset = 0;
+          }
+        }
+      };
+
+      optIfSatisfied(c);
```
kripken, line 151: I don't think we need to add this lambda function? The code inside it can just be here normally, without being wrapped in the lambda.

src/passes/OptimizeAddedConstants.cpp
```diff
@@ -123,12 +126,29 @@ template<typename P, typename T> class MemoryAccessOptimizer {
...
-        curr->offset = 0;
-      }
+
+      auto optIfSatisfied = [&](Const* c) {
+        if (memory64) {
+          uint64_t base = c->value.geti64();
+          uint64_t offset = curr->offset;
+
+          uint64_t max = std::numeric_limits<uint64_t>::max();
+          bool overflow = (base > max - offset) || (offset > max - base);
```
kripken, line 136: Are both the checks needed?

kripken (commented):
Thanks!

In addition to the comments below please also add a test. Perhaps in a new file, `test/lit/passes/optimize-added-constants-memory64.wast`.

src/passes/OptimizeAddedConstants.cpp
dongAxis: the result might be overflow?

src/passes/OptimizeAddedConstants.cpp
dongAxis: address comment, thanks

src/passes/OptimizeAddedConstants.cpp
dongAxis: address comment, thanks

src/passes/OptimizeAddedConstants.cpp
dongAxis: @kripken the code just check overflow, dose it right?

kripken: The code just needs to check overflow, correct. But I think one of the two is enough. It looks a little weird that it's not symmetric, that's true, but it's still correct I believe. Perhaps I'm missing something, though - can you find some values where both checks are needed to find the overflow?

dongAxis: thanks, after the `#5038` is accepted. I will write more tests for this case, because now the parser of binaryen can not accept the number that is larger than int32_max.

Yifeng

dongAxis: > The code just needs to check overflow, correct. But I think one of the two is enough. It looks a little weird that it's not symmetric, that's true, but it's still correct I believe. Perhaps I'm missing something, though - can you find some values where both checks are needed to find the overflow?

thanks, after the #5038 is accepted. I will write more tests for this case, because now the parser of binaryen can not accept the number that is larger than int32_max.
Yifeng

kripken: Sorry, maybe I wasn't clear. This line can be simplified like this:
```cpp
bool overflow = base > max - offset;
```
It may be surprising that only one of the two checks you wrote is needed, but mathematically it works out. For simplicity, consider the 8-bit case, and that
```
base = 100
offset = 200
max = 255
```
Then
```
base > max - offset
100 > 255 - 200
100 > 55
true
```
and
```
offset > max - base
200 > 255 - 100
200 > 155
true
```
So the checks are redundant. We only need one of them.

Mathematically, `base > max - offset` and `offset > max - base` are equivalent. That is true both in normal math, where there are no overflows (just add offset and subtract base from the first equation to get the second), and also in wrapping math (both equations imply an overflow in `base + offset`, and an overflow there implies each of them, so we have equivalence).

MATRIXKOO (approved):
LGTM

MATRIXKOO (approved):

MATRIXKOO (commented):
LGTM

MATRIXKOO:
```cpp
// /src/src/passes/OptimizeAddedConstants.cpp:242:15 looks like `value >=0` always true.
 uint64_t value = literal.getInteger();
 // Avoid uninteresting corner cases with peculiar offsets.
 if (value >= 0 && value < PassOptions::LowMemoryBound) {
```

src/passes/OptimizeAddedConstants.cpp
dongAxis: thanks, I add new test for the overflow check.

dongAxis:
@kripken In addition, could you give me a write permission? Because we will focus on binaryen and wasm, I would love to write more code for binaryen. Thanks in advance.

dongAxis:
> @kripken In addition, could you give me a write permission? Because we will focus on binaryen and wasm, I would love to write more code for binaryen. Thanks in advance.

Hi kripken, I found some of unit test failed, could you tell me how to run those tests in my local environment?  thanks

kripken:
Hmm, the CI error looks unrelated to this PR. It's a closure error,
```
building:ERROR: Closure compiler completed with warnings and -Werror=closure enabled, aborting!
```
Likely that is due to @sbc100's recent changes on emscripten (CI here uses latest unstable emscripten). Does that sound likely @sbc100 ?

@dongAxis In general though to run those tests you can look at `.github/workflows/ci.yml`, and I think it just installs the emsdk and runs `./scripts/emcc-tests.sh`. But in this case we can ignore that failure.

About write access, I think the main issue was that github wouldn't run CI for you without someone approving it? That seems to have been fixed after your earlier commits that landed - I didn't need to manually run CI here today. I believe github requires manual approval for CI only for people that haven't committed to a repo.

test/lit/passes/optimize-added-constants-memory64.wast
```diff
@@ -0,0 +1,54 @@
+;; RUN: wasm-opt %s -all --roundtrip --optimize-added-constants --low-memory-unused -S -o - | filecheck %s
```
kripken, line 1:
```suggestion
;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.

;; RUN: wasm-opt %s -all --roundtrip --optimize-added-constants --low-memory-unused -S -o - | filecheck %s
```
Adding that first magical line will make `./scripts/auto_update_tests.py lit` auto-generate the CHECK lines. It will also place each CHECK before the relevant function etc. in a more readable way.

sbc100:
> Hmm, the CI error looks unrelated to this PR. It's a closure error,
> 
> ```
> building:ERROR: Closure compiler completed with warnings and -Werror=closure enabled, aborting!
> ```
> 
> Likely that is due to @sbc100's recent changes on emscripten (CI here uses latest unstable emscripten). Does that sound likely @sbc100 ?

Yup that sounds expected.   I guess we should  either fix the warnings of build with `-Wno-closure` to disable them.

test/lit/passes/optimize-added-constants-memory64.wast
dongAxis: address the comment.

dongAxis:
> 

ok, thanks

dongAxis:
> Hmm, the CI error looks unrelated to this PR. It's a closure error,
> 
> ```
> building:ERROR: Closure compiler completed with warnings and -Werror=closure enabled, aborting!
> ```
> 
> Likely that is due to @sbc100's recent changes on emscripten (CI here uses latest unstable emscripten). Does that sound likely @sbc100 ?
> 
> @dongAxis In general though to run those tests you can look at `.github/workflows/ci.yml`, and I think it just installs the emsdk and runs `./scripts/emcc-tests.sh`. But in this case we can ignore that failure.
> 
> About write access, I think the main issue was that github wouldn't run CI for you without someone approving it? That seems to have been fixed after your earlier commits that landed - I didn't need to manually run CI here today. I believe github requires manual approval for CI only for people that haven't committed to a repo.

ok thanks

test/lit/passes/optimize-added-constants-memory64.wast
```diff
@@ -0,0 +1,56 @@
...
+;; CHECK-NEXT:   (i64.const 579)
+;; CHECK-NEXT:   (i64.const 123)
+;; CHECK-NEXT:  )
+;; CHECK-NEXT: )
+;; CHECK-NEXT: (func $store_overflow
+;; CHECK-NEXT:  (i64.store offset=32
+;; CHECK-NEXT:   (i64.const -16)
+;; CHECK-NEXT:   (i64.const 123)
+;; CHECK-NEXT:  )
+;; CHECK-NEXT: )
```
kripken, line 29: @tlively , these automatically-generated checks don't look right - they aren't right before each function etc. Any idea?

tlively: The functions looks like they should be handled ok. @dongAxis, did you actually run `./scripts/auto_update_tests.py lit`? If not, it would be good to run that and push the fixed checks so we don't get unnecessary churn in later PRs.

dongAxis: address the comment, thanks

kripken:
About memory64 performance, this PR fixes one important issue, and there are likely more places we just support 32-bit integers but not 64. I bet in OptimizeInstructions for example we have such issues. One way to find them might be to compile the same code in wasm32 and wasm64 and just look at the code side by side - anywhere the wasm64 code is larger is worth investigating.

cc @sbc100

PR #5044, opened 2022-09-15
JPSI - Support re-entering a suspended Wasm module.

brendandahl:
Fixes: https://github.com/emscripten-core/emscripten/issues/17846

More detailed explanation of the issue from Thibaud:
- A promising export is entered, generating a suspender s1, which is stored in the global
- The wasm code calls a wrapped import, passing it the value in the global (s1) and suspends
- Another export is entered, generating suspender s2, which is stored in the global
- We call another wrapped import, which suspends s2 (so far so good)
- We return to the event loop and s1 is resumed

And now we are in an inconsistent state: the active suspender is "s1", but the object in the global is "s2". So the next time we call a wrapped import, there is a mismatch, which is what this runtime error reports.

### DIFF
--- a/src/passes/JSPI.cpp
+++ b/src/passes/JSPI.cpp
@@ -155,9 +155,33 @@ private:
       params.push_back(param);
       ++i;
     }
-
+    auto* block = builder.makeBlock();
+    // Store the suspender so it can be restored after the call in case it is
+    // modified by another entry into a Wasm export.
+    auto supsenderCopyIndex = Builder::addVar(stub.get(), externref);
+    // If there's a return value we need to store it so it can be returned
+    // after restoring the suspender.
+    std::optional<Index> returnIndex;
+    if (stub->getResults().isConcrete()) {
+      returnIndex = Builder::addVar(stub.get(), stub->getResults());
+    }
+    block->list.push_back(builder.makeLocalSet(
+      supsenderCopyIndex, builder.makeGlobalGet(suspender, externref)));
+    if (returnIndex) {
+      block->list.push_back(builder.makeLocalSet(*returnIndex, call));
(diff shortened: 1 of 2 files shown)


tlively (approved):
Looks good, although since we don't save and restore the stack pointer (or allocate new stacks), we can still get problems.

If A suspends, then B gets called and suspends, then A resumes, then we will be in an inconsistent state because then B's stack frames will still be on the stack. On the other hand, if A suspends, then B gets called and returns, then A resumes, everything will be fine because B's stack frames will have been popped off the stack.

We could detect that situation by saving the stack pointer locally when suspending, then checking that it has the same value when we resume, trapping or calling an import to report the error otherwise. I do think that's worth doing, but it doesn't have to be in this PR.

kripken (approved):

kripken:
@tlively 

> Looks good, although since we don't save and restore the stack pointer (or allocate new stacks), we can still get problems.

True. Atm the C stack is managed by JS runtime code for Asyncify Fibers, where it allocates new stacks as needed etc. We might be able to use the same code from there. But Fibers might be low priority.

tlively:
To be clear, I'm not suggesting that we immediately implement fully fiber support. I'm suggesting that we detect and error out eagerly on cases where full fiber support would have been necessary rather than continuing execution with a silently corrupted stack.

kripken:
@tlively Ah, got it, thanks. Yes, that makes sense.

I believe the existing assertions in the JS runtime on the emscripten side handle that. But if there is a good way to add them here that would be better.

brendandahl:
I should have mentioned this is just to fix the current case where things resolve in a well behaved order. I've filed #5052 for adding the check.

PR #5045, opened 2022-09-15
[C API] Make TypeBuilderSetSubType take a heap type

### LINKED ISSUE #5041: TypeBuilder woes
Say a compiler uses Binaryen as its backend, and the compiler reuses as much as possible from Binaryen, incrementally generating IR nodes and types while compiling a program. Now say there is the following simple program:

```ts
class A {}
class B extends A {}
new A();
new B();
```

Upon encountering `new A()`, the compiler generates a `struct.new` instruction and uses a `TypeBuilder` to build the respective heap type of `A` for use with the instruction. This one is simple, since there's just a single slot with the structure of `A`, and `A` has no super type. It's not yet known whether `B` will be used, and the compiler doesn't register a heap type for `B` eagerly.

Upon encountering `new B()` later, the compiler again generates a `struct.new` instruction and uses another `TypeBuilder` to build the respective heap type of `B`. This time, it populates the `TypeBuilder` with both the supertype `A` and the subtype `B` to establish the subtyping relationship.

In isorecursive mode, incrementally building types like this works fine, since if types are structurally identical, a canonical heap type is returned (iiuc). In nominal mode, however, incrementally building types like this behaves differently, in that upon each `build` fresh types are created, so there are both an `A` and a `B <: A'`, with mismatching `A`s.

To avoid this problem, I figured that I might well need an entire separate IR, so code can be generated before a final `TypeBuilder` is used to build the full nominal hierarchy in one go, only then translating the separate IR to Binaryen IR. Now, I'd like to use nominal mode but creating a separate IR seems unfortunate, as it would also require reinventing lots of wheels like an interpreter and so on, with the result that using Binaryen as a backend doesn't make a lot of sense in the first place anymore. 

Questions: Is it actually a valid use case that I want to use nominal instead of isorecursive mode? And if so, any ideas how nominal mode can be supported in incremental codegen like this as well, without reinventing half of Binaryen? :)

(I guess this didn't come up earlier since, if Binaryen is solely used as an optimizer, this is a non-issue.)

Some ideas:
* When creating a `TypeBuilder`, allow adding already built types to it if a fresh type is not desired
* Allow reusing the same `TypeBuilder` forever, not destroying it when types are built
* Integrate type building more closely and have a finalization step at the end (but that's probably much more work)

dcodeIO:
Fixes #5041

### DIFF
--- a/src/ir/type-updating.cpp
+++ b/src/ir/type-updating.cpp
@@ -80,7 +80,8 @@ void GlobalTypeRewriter::update() {
 
     // Apply a super, if there is one
     if (auto super = type.getSuperType()) {
-      typeBuilder.setSubType(i, typeBuilder.getTempHeapType(indexedTypes.indices[*super]));
+      typeBuilder.setSubType(
+        i, typeBuilder.getTempHeapType(indexedTypes.indices[*super]));
     }
   }
 
--- a/src/ir/type-updating.cpp
+++ b/src/ir/type-updating.cpp
@@ -80,7 +80,7 @@ void GlobalTypeRewriter::update() {
 
     // Apply a super, if there is one
     if (auto super = type.getSuperType()) {
-      typeBuilder.setSubType(i, *super);
+      typeBuilder.setSubType(i, typeBuilder.getTempHeapType(indexedTypes.indices[*super]));
     }
   }
 
--- a/src/ir/type-updating.cpp
+++ b/src/ir/type-updating.cpp
@@ -80,7 +80,7 @@ void GlobalTypeRewriter::update() {
 
     // Apply a super, if there is one
     if (auto super = type.getSuperType()) {
-      typeBuilder.setSubType(i, indexedTypes.indices[*super]);
+      typeBuilder.setSubType(i, *super);
     }
   }
 
--- a/src/wasm-type.h
+++ b/src/wasm-type.h
@@ -602,6 +602,10 @@ struct TypeBuilder {
   // index `i` to be an immediate subtype of the HeapType being built at index
   // `j`. Does nothing for equirecursive types.
   void setSubType(size_t i, size_t j);
+  // In nominal mode, or for nominal types, declare the HeapType being built at
+  // index `i` to be an immediate subtype of the given HeapType. Does nothing
+  // for equirecursive types.
+  void setSubType(size_t i, HeapType super);
 
   // Create a new recursion group covering slots [i, i + length). Groups must
   // not overlap or go out of bounds.
--- a/src/wasm-type.h
+++ b/src/wasm-type.h
@@ -599,10 +599,6 @@ struct TypeBuilder {
   Type getTempRefType(HeapType heapType, Nullability nullable);
 
   // In nominal mode, or for nominal types, declare the HeapType being built at
-  // index `i` to be an immediate subtype of the HeapType being built at index
-  // `j`. Does nothing for equirecursive types.
-  void setSubType(size_t i, size_t j);
-  // In nominal mode, or for nominal types, declare the HeapType being built at
   // index `i` to be an immediate subtype of the given HeapType. Does nothing
   // for equirecursive types.
   void setSubType(size_t i, HeapType super);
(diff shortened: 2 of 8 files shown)


dcodeIO:
Alternative names:
* `TypeBuilderSetSubTypeByIndex` (for the old)
* `TypeBuilderSetSubTypeFromType` (for the new)

src/wasm-type.h
```diff
@@ -602,6 +602,10 @@ struct TypeBuilder {
   // index `i` to be an immediate subtype of the HeapType being built at index
   // `j`. Does nothing for equirecursive types.
   void setSubType(size_t i, size_t j);
+  // In nominal mode, or for nominal types, declare the HeapType being built at
+  // index `i` to be an immediate subtype of the given HeapType. Does nothing
+  // for equirecursive types.
+  void setSubType(size_t i, HeapType super);
```
tlively, line 608: I don't think we need the previous `setSubType` at all. Would you mind trying to replace it? It would also be good to add some tests showing that your use case of setting the supertype to a previously existing type works, perhaps in test/gtest/type-builder.cpp.

test/gtest/type-builder.cpp
```diff
@@ -533,3 +533,35 @@ TEST_F(NominalTest, TestSubTypes) {
...
+  }
+
+  // Build a type B <: A using a new builder
+  Type B;
+  {
+    TypeBuilder builder(1);
+    builder[0] = Struct();
+    builder.setSubType(0, A.getHeapType());
+    auto result = builder.build();
+    ASSERT_TRUE(result); // FIXME: ErrorReason::InvalidSupertype
```
dcodeIO, line 557: I might be doing something wrong, or the use case doesn't work. Ideas?

dcodeIO: Correcting myself, this seems to be pretty broken now and I have no idea what I'm doing :)

src/ir/type-updating.cpp
```diff
@@ -80,7 +80,8 @@ void GlobalTypeRewriter::update() {
 
     // Apply a super, if there is one
     if (auto super = type.getSuperType()) {
-      typeBuilder.setSubType(i, indexedTypes.indices[*super]);
+      typeBuilder.setSubType(
+        i, typeBuilder.getTempHeapType(indexedTypes.indices[*super]));
```
dcodeIO, line 84: This was the problematic part, where the new temp heap type must be used instead of the original type.

test/gtest/type-builder.cpp
```diff
@@ -533,3 +533,35 @@ TEST_F(NominalTest, TestSubTypes) {
...
+    auto built = *result;
+    B = Type(built[0], Nullable);
+  }
+
+  // Test that B <: A where A is the initial type A
+  auto superOfB = B.getHeapType().getSuperType();
+  ASSERT_TRUE(superOfB);
+  EXPECT_EQ(*superOfB, A.getHeapType());
+  EXPECT_NE(B.getHeapType(), A.getHeapType());
+}
```
tlively, line 567: Could you add one more test in isorecursive mode checking that if we build `A`, then separately we build `A'`, `B <: A`, and `B' <: A'`, then at the end we still have `A == A'` and `B == B'`? That will test that there is no difference between existing and new supertypes as far as isorecursive canonicalization goes.

tlively (approved):
LGTM pending one more test. Thanks!

test/gtest/type-builder.cpp
dcodeIO: Currently looking into the additional test. So far, when I make a type `A1` using its own builder, an identical type `A2` using its own builder, and then a type `B1 <: A1` using its own builder, building fails with `"Heap type has an undeclared supertype" at index 0`. So I guess there's something else missing. Ideas?

tlively: Ah right, good find. This code needs to be updated to differentiate between canonical and temporary supertypes that haven't been seen yet. The canonical types should be allowed, but the temporary types should continue producing this error. There's an `isTemp(HeapType)` function you can use to check this.

tlively (approved):
Thanks!

PR #5046, opened 2022-09-15
Vacuum trivial trys

kripken:
A try whose body throws, and does nothing else, and the try catches that
exception, can be removed.


aheejin (approved):

kripken:
Spec test found a bug, we can't do this if a result is returned from the try.

PR #5049, opened 2022-09-16
wasm2js: Don't assume that `env.abort` can always be impored.

sbc100:


sbc100:
I ran the entire wasm2js1 test suite in emscripten an verified that this doesn't break anyting.

src/wasm2js.h
```diff
@@ -2811,10 +2805,19 @@ void Wasm2JSGlue::emitSpecialSupport() {
   // The special support functions are emitted as part of the JS glue, if we
   // need them.
   bool need = false;
+  bool need_scratch = false;
```
kripken, line 2808:
```suggestion
  bool needScratch = false;
```

kripken (commented):
lgtm % style nit

PR #5051, opened 2022-09-16
wasm2js: Generalize to allow 2-level imports

sbc100:
This removes any hard-coded references to `env` (the mangic module name used by llvm and emscripten) and instead imports from whatever module name is used in import declaration.


sbc100:
We can't land this until emscripten starts passing the full imports object: https://github.com/emscripten-core/emscripten/pull/17806

kripken (approved):
Nice!

sbc100:
Replaces by #5114

PR #5053, opened 2022-09-16
Vacuum: Ignore effects at the entire function scope when possible

kripken:
Recently we added logic to ignore effects that don't "escape" past the function call.
That is, e.g.  `local.set` only affects the current function scope, and once the call stack
is unwound it no longer matters as an effect. This moves that logic to a shared place,
and uses it in the core Vacuum logic.

The new constructor in `EffectAnalyzer` receives a function and then scans it as
a whole. This works just like e.g. scanning a Block as a whole (if we see a break in
the block, that has an effect only inside it, and the Block + children doesn't have a
branch effect).

Various tests are updated so they don't optimize away trivially, by adding new
return values for them.


test/lit/passes/vacuum-func.wast
```diff
@@ -0,0 +1,60 @@
...
+  ;; CHECK-NEXT:   (i32.const 2)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (return)
+  ;; CHECK-NEXT: )
+  (func $partial (param $x i32)
+    (local $y i32)
+
+    ;; As above, but with this if added which adds extra possible effects. This
+    ;; prevents optimization.
+    ;; We could, however, optimize away the local operations TODO (we can also
```
tlively, line 45: Don't we already have a pass that will remove unused `local.set`s? What more is there for vacuum to do here?

tlively (approved):
LGTM.

A similar change I've thought about is allowing functions to have `unreachable` as their result type if they can never return. That would allow unreachability to propagate much farther than it currently does. I think the only change to module writing would be to translate unreachable signature results to empty signature results. After that our usual unreachable-handling logic should just work (I think).

test/lit/passes/vacuum-func.wast
kripken: We do, yeah, but there is actually a tricky case I'm worried about, which is why I wrote this (but I can delete it). Consider this code:
```js
function foo(i) {
  while (i < 1000) {
    i += 22;
    i *= 3;
  }
}
```
It seems obvious that code is not needed, since the result of the computation does not go anywhere. But none of our existing passes remove it, since the local `i` is both read and written. SSA form wouldn't help here AFAICT, we'd have a phi that has a use - itself - and so it seems used. What I suppose is necessary is a reachability analysis to see if the result of a computation "escapes" from the function (into either a return value or global state). Do you happen to know if that's what LLVM does, or if I'm missing something simpler?

kripken:
Nice idea about `unreachable`! That would be kind of like annotating "noreturn" - which is what unreachable means I guess. Sounds simple and useful, and a nice use of the type system.

aheejin (approved):

test/lit/passes/vacuum-func.wast
tlively: Makes sense, thanks. It would be good to update the comment to mention that it's more complicated cases where values locally appear used that needs the additional optimization.

I don't know specifically what LLVM does here, but my guess is that they have a more sophisticated data flow analysis that shows that the values computed here don't end up going anywhere. I hope that if we integrated the interpreter into GUFA we would be able to discover the same thing.

PR #5057, opened 2022-09-19
[Strings] Add missing String effects + tests

kripken:
Also fix some formatting issue in the file.


test/lit/passes/simplify-locals-strings.wast
```diff
@@ -0,0 +1,544 @@
...
+  ;; CHECK-NEXT:   (i32.const 3)
+  ;; CHECK-NEXT:   (i32.const 4)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (local.get $temp)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $no-new-past-store-gc (param $array (ref $array)) (param $array16 (ref $array16))
+    (local $temp stringref)
+    ;; A string.new cannot be moved past a GC store.
```
aheejin, line 224:
```suggestion
    ;; A string.new_***_array cannot be moved past a GC store.
```

test/lit/passes/simplify-locals-strings.wast
```diff
@@ -0,0 +1,544 @@
...
+  ;; CHECK-NEXT:   (string.encode_wtf16
+  ;; CHECK-NEXT:    (local.get $ref)
+  ;; CHECK-NEXT:    (i32.const 30)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (local.get $temp)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $no-load-past-encode (param $ref stringref)
```
aheejin, line 332: Why can't this go past the encode? Does `encode` operation only read memory or write to memory as well? It'd be helpful to have one-line comments in this function and the functions below as we did for previous functions.

kripken: Encode takes an existing String and writes it out into either linear memory or an array. So it has write effects.

Added comments now.

aheejin (approved):

PR #5058, opened 2022-09-19
[Strings] UTF-8 support

kripken:
This takes @tlively 's initial patch, makes it a little worse by removing some performance optimizations (that I couldn't figure out how to get to compile :rofl: - specifically the vector-of-references stuff), and pokes at it until it compiles and passes the test suite.

Also I simplified `printEscapedString`. That was necessary to avoid an assert on previously-existing tests. It seems like an optional style thing? But I'm not actually sure what the extra code did, I'm not that familiar with UTF8 stuff.

Also adds a test of a UTF-8 string.


kripken:
Chatting with @tlively , this PR isn't quite right, and the better approach is probably to make `Name` support all utf-8 values. That would be useful for other things, and would also make String literals just work with utf-8 since they use `Name`.

PR #5063, opened 2022-09-20
Make closure errors into warnings in the Emscripten builds

tlively:
This should make the CI green again. Also fix one of the errors. I haven't fixed
the other errors because I don't know how.


tlively:
* main
  * **PR #5063** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5063" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5063?utm_source=stack-comment).

kripken (approved):

src/js/binaryen.js-post.js
```diff
@@ -2567,7 +2567,7 @@ function wrapModule(module, self = {}) {
   self['removeExport'] = function(externalName) {
     return preserveStack(() => Module['_BinaryenRemoveExport'](module, strToStack(externalName)));
   };
-  self['setMemory'] = function(initial, maximum, exportName, segments = [], shared = false, memory64 = false, internalName) {
+  self['setMemory'] = function(initial, maximum, exportName, segments = [], shared = false, memory64 = false, internalName = 0) {
```
MaxGraey, line 2570: Isn't this should be `internalName = null` or `internalName = "0"`?

kripken: I think 0 or null work, as the call to `strToStack` will treat them as "nothing". "0" however would be used as a name, which is not good.

Maybe null or undefined are best. `undefined` is the old value actually, so that might be best.

tlively: `null` seems clearest to me, since we really are trying to get a null ptr as the default here.

PR #5067, opened 2022-09-21
[GUFA] Optimize ref.test

kripken:
Similar to ref.cast slightly, but simpler.

Also update some TODO text as I go through them.


kripken:
(I verified this helps j2wasm, though it seems rare)

tlively (approved):

PR #5068, opened 2022-09-21
Add a type annotation to return_call_ref

tlively:
The GC spec has been updated to have heap type annotations on call_ref and
return_call_ref. To avoid breaking users, we will have a graceful, multi-step
upgrade to the annotated version of call_ref, but since return_call_ref has no
users yet, update it in a single step.

### DIFF
--- a/src/passes/Print.cpp
+++ b/src/passes/Print.cpp
@@ -2014,7 +2014,13 @@ struct PrintExpressionContents
   }
   void visitCallRef(CallRef* curr) {
     if (curr->isReturn) {
-      printMedium(o, "return_call_ref");
+      if (printUnreachableReplacement(curr->target)) {
+        return;
+      }
+      printMedium(o, "return_call_ref ");
+      assert(curr->target->type != Type::unreachable);
+      // TODO: Workaround if target has bottom type.
+      printHeapType(o, curr->target->type.getHeapType(), wasm);
     } else {
       printMedium(o, "call_ref");
     }
--- a/src/passes/Print.cpp
+++ b/src/passes/Print.cpp
@@ -2012,6 +2012,22 @@ struct PrintExpressionContents
   void visitI31Get(I31Get* curr) {
     printMedium(o, curr->signed_ ? "i31.get_s" : "i31.get_u");
   }
+
+  // If we cannot print a valid unreachable instruction (say, a struct.get,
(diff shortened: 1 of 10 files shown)


tlively:
* main
  * **PR #5068** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5068" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5068?utm_source=stack-comment).

src/passes/Print.cpp
```diff
@@ -2014,7 +2014,13 @@ struct PrintExpressionContents
   }
   void visitCallRef(CallRef* curr) {
     if (curr->isReturn) {
-      printMedium(o, "return_call_ref");
+      if (printUnreachableReplacement(curr->target)) {
```
kripken, line 2017: return calls are always unreachable aren't they? this would always happen then, unless I'm missing something.

tlively: We're passing `curr->target`, so we only print the replacement if the _target_ is unreachable.

test/lit/passes/dae-gc-refine-return.wast
```diff
@@ -654,7 +658,9 @@
  ;; CHECK-NEXT:  (drop
  ;; CHECK-NEXT:   (call $tail-caller-call_ref-no)
  ;; CHECK-NEXT:  )
- ;; CHECK-NEXT:  (call $tail-caller-call_ref-unreachable)
+ ;; CHECK-NEXT:  (drop
+ ;; CHECK-NEXT:   (call $tail-caller-call_ref-unreachable)
+ ;; CHECK-NEXT:  )
```
kripken, line 663: Why were drops added here?

kripken (approved):
lgtm % question

test/lit/passes/dae-gc-refine-return.wast
tlively: Previously the text parser used `ValidatingBuilder::validateAndMakeCallRef` for return calls and that would produce an unreachable block expression rather than an actual call inside `tail-caller-call_ref-unreachable`. Now the parser actually produces a tail call `CallRef` expression there that prevents DAE from removing the return value, so that return value needs to be dropped here.

PR #5069, opened 2022-09-21
[OptimizeInstructions] Optimize comparisons of a shifted value to a constant

kripken:
Handles https://github.com/WebAssembly/binaryen/issues/5008#issuecomment-1235824896


kripken:
This looks correct, but is quite complex, and though it was a top superoptimizer suggestion I see little benefit. So I am not sure it is worth it. Leaving as draft for now.

PR #5070, opened 2022-09-22
Correctly handle escapes in string constants

tlively:
Previously when we parsed `string.const` payloads in the text format we were
using the text strings directly instead of un-escaping them. Fix that parsing,
and while we're editing the code, also add support for the `\r` escape allowed
by the spec. Remove a spurious nested anonymous namespace and spurious `static`s
in Print.cpp as well.


tlively:
* main
  * **PR #5070** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5070" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5070?utm_source=stack-comment).

test/lit/strings.wast
```diff
@@ -31,8 +31,8 @@
 
   ;; CHECK:      (type $stringref_ref|$array|_ref|$array16|_=>_none (func (param stringref (ref $array) (ref $array16))))
 
-  ;; CHECK:      (global $string-const stringref (string.const "string in a global"))
-  (global $string-const stringref (string.const "string in a global"))
+  ;; CHECK:      (global $string-const stringref (string.const "string in a global \01\ff\t\n\r\"\'\\"))
+  (global $string-const stringref (string.const "string in a global \01\ff\09\0a\0d\22\27\5c"))
```
kripken, line 35: Maybe also add a string with the output, to make sure we parse that? That is, a string that looks like what is in the `CHECK`.

kripken (approved):
lgtm % comment

PR #5073, opened 2022-09-22
Fix binaryen.js tests after require() changes on emscripten

kripken:
See https://github.com/WebAssembly/binaryen/issues/5071


tlively (approved):

kripken:
Not sure if we should land this now to fix the breakage, as discussion is still ongoing on the emscripten issue... if the broken tests prevent stuff from landing maybe we should.

kripken:
We went the other way.

PR #5074, opened 2022-09-22
Memory64Lowering: Ignore data segments with non-const iniital offset

sbc100:
This is the case for dynamic linking where the segment offset are derived from he `__memory_base` import.

### DIFF
--- a/src/passes/Memory64Lowering.cpp
+++ b/src/passes/Memory64Lowering.cpp
@@ -21,11 +21,15 @@
 // TODO(wvo): make this run in parallel if needed.
 
 #include "ir/bits.h"
+#include "ir/import-utils.h"
 #include "pass.h"
 #include "wasm-builder.h"
 #include "wasm.h"
 
 namespace wasm {
+
+static Name MEMORY_BASE("__memory_base");
+static Name MEMORY_BASE32("__memory_base32");
 
 struct Memory64Lowering : public WalkerPass<PostWalker<Memory64Lowering>> {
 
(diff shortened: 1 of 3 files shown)


src/passes/Memory64Lowering.cpp
```diff
@@ -126,9 +131,31 @@ struct Memory64Lowering : public WalkerPass<PostWalker<Memory64Lowering>> {
...
     if (!segment->isPassive) {
-      auto* c = segment->offset->cast<Const>();
-      c->value = Literal(static_cast<uint32_t>(c->value.geti64()));
-      c->type = Type::i32;
+      if (auto* c = segment->offset->dynCast<Const>()) {
+        c->value = Literal(static_cast<uint32_t>(c->value.geti64()));
+        c->type = Type::i32;
+      } else if (auto* get = segment->offset->dynCast<GlobalGet>()) {
+        auto& module = *getModule();
+        Global* g = module.getGlobal(get->name);
```
kripken, line 139:
```suggestion
        auto* g = module.getGlobal(get->name);
```

src/passes/Memory64Lowering.cpp
```diff
@@ -126,9 +131,31 @@ struct Memory64Lowering : public WalkerPass<PostWalker<Memory64Lowering>> {
...
+        if (g->imported() && g->base == MEMORY_BASE) {
+          ImportInfo info(module);
+          Global* memoryBase32 =
+            info.getImportedGlobal(g->module, MEMORY_BASE32);
+          if (!memoryBase32) {
+            memoryBase32 = new Global();
```
kripken, line 145: Please use https://github.com/WebAssembly/binaryen/blob/e750809428e447396c1568139bf970ce5a1b5aed/src/wasm-builder.h#L153

kripken (approved):

PR #5080, opened 2022-09-23
Remove the GCNNLocals feature

tlively:
Now that the WasmGC spec has settled on a way of validating non-nullable locals,
we no longer need this experimental feature that allowed nonstandard uses of
non-nullable locals.


tlively:
* main
  * **PR #5080** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5080" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5080?utm_source=stack-comment).

tlively:
Marked draft because I still have to go through and look for stale comments and useless tests.

tlively:
This should be ready to go. I tried finding stale comments and tests by grepping for "non-nullable" and looking at changed test files, but didn't find much.

src/wasm/wasm-binary.cpp
```diff
@@ -2963,19 +2961,18 @@ void WasmBinaryReader::pushExpression(Expression* curr) {
...
-          t = Type(t.getHeapType(), Nullable);
-        }
-        finalTypes.push_back(t);
-      }
-      localType = Type(Tuple(finalTypes));
-    }
+    // Non-nullable types require special handling as they cannot be stored to a
+    // local, so we may need to use a different local type than the original.
+    // TODO: Now that the spec supports limited non-nullable locals, is this
+    // still true?
```
kripken, line 2967: I think if it is then we have a bug with tuple handling...

test/lit/passes/optimize-instructions-gc-tnh-nn.wast
```diff
@@ -1,18 +1,16 @@
...
+  ;; CHECK-NEXT:   (local.get $x)
   ;; CHECK-NEXT:  )
   ;; CHECK-NEXT:  (drop
-  ;; CHECK-NEXT:   (local.tee $x
-  ;; CHECK-NEXT:    (ref.as_non_null
+  ;; CHECK-NEXT:   (ref.as_non_null
+  ;; CHECK-NEXT:    (local.tee $x
   ;; CHECK-NEXT:     (local.get $x)
   ;; CHECK-NEXT:    )
   ;; CHECK-NEXT:   )
```
kripken, line 16: Comment on line 21 looks like it needs updating.

kripken (approved):

src/wasm/wasm-binary.cpp
tlively: I tried deleting this special handling and everything seemed to work fine.

test/lit/passes/optimize-instructions-gc-tnh-nn.wast
tlively: Looks like I had accidentally changed the behavior of this in OptimizeInstructions. I fixed that, then updated all the comments on these tests.

tlively:
Please take another look at the latest changes.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1153,46 +1153,6 @@ struct OptimizeInstructions
...
-    // Interactions between local.set/tee and ref.as_non_null can be optimized
-    // in some cases, by removing or moving the ref.as_non_null operation. In
-    // all cases, we only do this when we do *not* allow non-nullable locals. If
-    // we do allow such locals, then (1) this local might be non-nullable, so we
-    // can't remove or move a ref.as_non_null flowing into a local.set/tee, and
-    // (2) even if the local were nullable, if we change things we might prevent
-    // the LocalSubtyping pass from turning it into a non-nullable local later.
-    // Note that we must also check if this local is nullable regardless, as a
-    // parameter might be non-nullable even if nullable locals are disallowed
-    // (as that just affects vars, and not params).
```
kripken, line 1166: Why remove this logic? The comment is state, but the code is still useful?

src/wasm/wasm-binary.cpp
```diff
@@ -2963,29 +2961,16 @@ void WasmBinaryReader::pushExpression(Expression* curr) {
...
-    for (Index i = 0; i < localType.size(); ++i) {
+    for (Index i = 0; i < type.size(); ++i) {
       Expression* value =
-        builder.makeTupleExtract(builder.makeLocalGet(tuple, localType), i);
-      if (localType[i] != type[i]) {
-        // We modified this to be nullable; undo that.
-        value = builder.makeRefAs(RefAsNonNull, value);
-      }
+        builder.makeTupleExtract(builder.makeLocalGet(tuple, type), i);
       expressionStack.push_back(value);
```
kripken, line 2974: I would fuzz this change carefully, as I'm not sure we always treat a tuple with a NN field as a NN "thing". That is, such a tuple local should be assigned before it is read, like a NN local.

test/lit/passes/roundtrip.wast
```diff
@@ -23,10 +23,8 @@
...
  ;; CHECK-NEXT:    )
  ;; CHECK-NEXT:    (drop
- ;; CHECK-NEXT:     (ref.as_non_null
- ;; CHECK-NEXT:      (tuple.extract 1
- ;; CHECK-NEXT:       (local.get $0)
- ;; CHECK-NEXT:      )
+ ;; CHECK-NEXT:     (tuple.extract 1
+ ;; CHECK-NEXT:      (local.get $0)
  ;; CHECK-NEXT:     )
  ;; CHECK-NEXT:    )
```
kripken, line 29: Maybe worth adding a test like this that checks we do not validate if such a tuple is read before it is written to?

src/passes/OptimizeInstructions.cpp
tlively: Note the `!getModule()->features.hasGCNNLocals()` below. None of this code is supposed to run when non-nullable locals are enabled, so it should all be dead now.

kripken: Am I reading it wrong? It looks the opposite to me.

When running normally before, all of our users had `!getModule()->features.hasGCNNLocals()` as true (because none of them set that special flag), and so that can be removed from the `&&`-ed conditions in the `if` without changing anything. But if the code is removed, our users could notice.

tlively: Ugh, yes, I confused myself by forgetting what --enable-gc-nn-locals even did. Ignore all this for now and I'll ping you once I have it cleaned up.

src/wasm-features.h
```diff
@@ -40,21 +40,15 @@ struct FeatureSet {
...
     MVP = None,
     // Keep in sync with llvm default features:
     // https://github.com/llvm/llvm-project/blob/c7576cb89d6c95f03968076e902d3adfd1996577/clang/lib/Basic/Targets/WebAssembly.cpp#L150-L153
     Default = SignExt | MutableGlobals,
-    // GCNNLocals are opt-in: merely asking for "All" does not apply them. To
-    // get all possible values use AllPossible. See setAll() below for more
-    // details.
-    All = ((1 << 17) - 1) & ~GCNNLocals,
-    AllPossible = (1 << 17) - 1,
+    All = (1 << 17) - 1,
```
kripken, line 51: Shouldn't this be 16 now that you removed one?

tlively: Yes 😢

PR #5084, opened 2022-09-23
[GUFA] Fix RefTest logic

kripken:
See https://github.com/WebAssembly/binaryen/pull/5081#issuecomment-1256737179 - this is the same as that fix but for RefTest

### DIFF
--- a/src/ir/possible-contents.cpp
+++ b/src/ir/possible-contents.cpp
@@ -421,8 +421,7 @@ struct InfoCollector
     addChildParentLink(curr->ref, curr);
   }
   void visitRefTest(RefTest* curr) {
-    // We will handle this similarly to RefCast.
-    addChildParentLink(curr->ref, curr);
+    addRoot(curr);
   }
   void visitBrOn(BrOn* curr) {
     // TODO: optimize when possible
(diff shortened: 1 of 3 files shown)


tlively (approved):
Do I understand correctly that this means the GUFA analysis will not be able to propagate the 0 or 1 internally?

kripken:
Yes. We have to first finish the flow until it converges. Only when we know all the possible contents arriving into the RefTest can we determine that the test will definitely fail.

We could in theory converge, then apply such things, then iterate some more in GUFA from that position. That would give no more optimization power but it would save work.

tlively:
I would have expected the propagated 0 or 1 to be joined with all other possible-contents values that could reach that location as part of the iteration to a fixed point, but if GUFA doesn't support that, this seems fine for now.

kripken:
It's tricky since the result may not be monotonic - that only more values can appear at each location. Say for RefEq, if the only possible value is a null on each side, then the only possible result is 1. But if we do it incrementally, after the first value arrives on one side we have null and nothing, which has no intersection, so we infer a 0. Later the second value arrives on the other side and now we see the possible result is 1. So the previous 0 was incorrect breaking the monotonic property.

With RefTest that isn't technically possible atm since there is just one input, and we won't do anything until the first value arrives, which I think happens to work out. But relying on that technicality seems risky.

In general getting such monotonic/incremental logic right is not easy. Here is the reasoning for why it works for GC reads:

https://github.com/WebAssembly/binaryen/blob/e09fb5656d5e791051393bd08e6522f068ec7fe4/src/ir/possible-contents.cpp#L1569-L1581

tlively:
We talked about this offline a bit and I believe the resulting consensus was that something here should be fixed so that the operation is monotonic. (And in fact that it should be the case that all operations are monotonic, under the intuition that having more possible inputs can only ever produce more possible outputs).

kripken:
As in RefEq, this could be fixed in the flow with some more work. But this PR is a strict simplification of the code that makes it shorter, so I think it's the better approach.

PR #5086, opened 2022-09-26
[GUFA] Prepare for cone types [NFC]

kripken:
This does not actually add cone types, but it does NFC refactoring towards that.
Specifically it replaces the internal `ExactType` with `ConeType`, and the latter
has a depth, so a cone type of depth 0 is the old exact type.

Cone types with `depth > 0` are not possible yet, keeping this NFC.

I believe this design with a depth for cone types has little overhead. It does add
to the size of `ConeType`, but the variant there is larger anyhow (it contains a
`Literal`). And things like `isSubType` need to loop anyhow, so looping up to the
depth etc. in checks won't make things slower. Does that sound right?

### DIFF
--- a/src/ir/possible-contents.h
+++ b/src/ir/possible-contents.h
@@ -44,15 +44,16 @@ namespace wasm {
 //                     to that global. Typically we can only infer this for
 //                     immutable globals.
 //
-//  * ExactType:       Any possible value of a specific exact type - *not*
-//                     including subtypes. For example, (struct.new $Foo) has
-//                     ExactType contents of type $Foo.
+//  * ConeType:        Any possible value of a particular type, and a possible
+//                     "cone" of a certain depth below it. If the depth is 0
+//                     then only the exact type is possible; if the depth is 1
+//                     then either that type or its immediate subtypes, and so
+//                     forth.
 //                     If the type here is nullable then null is also allowed.
-//                     TODO: Add ConeType, which would include subtypes.
-//                     TODO: Add ExactTypePlusContents or such, which would be
+//                     TODO: Add ConeTypePlusContents or such, which would be
 //                           used on e.g. a struct.new with an immutable field
 //                           to which we assign a constant: not only do we know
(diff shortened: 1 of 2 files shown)


aheejin:
Wasn't able to participate in the discussion about `ConeType`... Is there anything to know about?

src/ir/possible-contents.h
```diff
@@ -199,8 +221,10 @@ class PossibleContents {
       return size_t(1) | (std::hash<Literal>()(getLiteral()) << 3);
     } else if (isGlobal()) {
       return size_t(2) | (std::hash<Name>()(getGlobal()) << 3);
-    } else if (isExactType()) {
-      return size_t(3) | (std::hash<Type>()(getType()) << 3);
+    } else if (auto* coneType = std::get_if<ConeType>(&value)) {
+      return size_t(3) | ((std::hash<std::pair<Type, Index>>{}(
+                            {coneType->type, coneType->depth}))
+                          << 3);
```
tlively, line 227: Instead of `|` and the shift, this should probably be using `wasm::rehash`. If the distribution of content kinds is not uniform, the low two bits of the hash will have less than two bits of entropy with this strategy.

tlively (approved):
LGTM % some comments on comment wording and a separable suggestion for improving the hashing.

kripken:
@aheejin 

> Is there anything to know about?

Hmm, there wasn't much background before this PR actually. I've been meaning to get to cone types eventually but there's no design doc or such. I think at some point @tlively might have mentioned using a depth is an option there, and I've come to think that's a good idea, since it makes exact types just a special case of depth=0, instead of having separate cone type and exact type. So this PR goes in that direction.

src/ir/possible-contents.h
kripken: Makes sense, yeah, this is worth a separate improvement. I'll add a TODO for myself.

aheejin (approved):

Issue #4938, opened 2022-08-18
file.cpp calls `exit` instead of throwing exception


brson:
I am calling binaryen's ModuleReader and ModuleWriter APIs from Rust and discovered that there are three cases in file.cpp where it calls `exit`. This isn't suitable for embedding in other applications. Can it be changed to throw an exception instead? Callers already need to deal with ParseException and MapParseException so this doesn't seem too burdensome a change.

If it's ok to change this to throw exceptions, would following the same pattern as MapException be acceptable? I am willing to make this change myself.

kripken:
I think this makes sense. My one concern is that in general we are trying to move to have less exception-handling code in Binaryen, in particular so that the wasm port can be built with `-fno-exceptions`. But I think this change would not impact that, as in `-fno-exceptions` builds the exceptions would just `exit` or `abort` I guess. And there should be no other downside to using exceptions for ModuleReader/Writer errors AFAICT. But perhaps people that know more about C++ than me can confirm if there's really no downside though.

brson:
Not using exceptions is generally better for rust too, so if there's a different error handling pattern you'd like I'm happy to do that instead.

kripken:
It doesn't seem like anyone else has strong opinions here, so I think using C++ exceptions for `file.cpp` is fine.

tlively:
We have all kinds of error handling patterns in Binaryen. Another to consider would be something like `BuildResult` from wasm-type.h. https://github.com/WebAssembly/binaryen/blob/main/src/wasm-type.h#L610-L642. Alternatively, we could move `Result` and `BuildResult` (https://github.com/WebAssembly/binaryen/blob/main/src/wat-parser.h#L26-L67) to a more general header and use those, or we could pull in a `std::expected` implementation from C++23 (https://en.cppreference.com/w/cpp/header/expected).

I would mildly prefer any of those options to using exceptions, but also using exceptions here is certainly the simplest short-term fix. It would be nice to go through and standardize our error handling as a code health project at some point.

brson:
Pulling the Result and MaybeResult types out of wat-parser.h and reusing them looks like something I can handle. I don't quite understand BuildResult on first read though.

brson:
After spending a day working on this I discovered the `Fatal` type, which is used in quite a few places, some of which also are possibly places it might be desirable for the process to not abort in an embedded context.

[...]

PR #5087, opened 2022-09-27
Change `exit()` to `Fatal()`; make it possible to throw on `Fatal()`.

brson:
This patch makes binaryen easier to call from other applications by making more errors recoverable instead of early-exiting.

The main thing it does is change three calls to `exit` on I/O errors into calls to `Fatal()`, which is an existing custom abstraction for handling unrecoverable errors. Currently `Fatal`'s destructor calls `_Exit(1)`.

My intent is to make it possible for `Fatal` to not exit, but to throw, allowing an embedding application to catch the exception.

Because the previous early exits were exiting with error code `EXIT_FAILURE`, I also changed `Fatal` to exit with `EXIT_FAILURE`. The test suite continues to pass so I assume this is ok.

Next I changed `Fatal` to buffer its error message until the destructor instead of immediately printing it to stderr. This is for ease of patching `Fatal` to throw instead.

Finally, I also included the patch I need to make `Fatal` throw when `THROW_ON_FATAL` is defined at compile time. I can carry this patch out of tree, but it is a small patch, so perhaps you will be willing to take it. I am happy to remove it.

The aforementioned patch throws in a dtor, which is so strongly discouraged that it requires a special `noexcept(false)` annotation in c++17, the pitfall being that it is easy to throw during unwinding, which causes an abort. This is a special type though that only exists on error paths, and this solution is the smallest solution that might work for my use case. I have accounted for double throws by inserting an explicit abort in that case.

I have run the test suite successfully, but have not manually tested that fatal errors print the same as previously. I have tested that the patch throws exceptions that can be recovered in my project https://github.com/brson/wasm-opt-rs

Fixes https://github.com/WebAssembly/binaryen/issues/4938

### DIFF
--- a/src/support/utilities.h
+++ b/src/support/utilities.h
@@ -24,6 +24,7 @@
 #include <cstring>
 #include <iostream>
 #include <memory>
+#include <sstream>
 #include <type_traits>
 
 #include "support/bits.h"
@@ -62,14 +63,16 @@ std::unique_ptr<T> make_unique(Args&&... args) {
 
 // For fatal errors which could arise from input (i.e. not assertion failures)
 class Fatal {
+private:
+  std::stringstream buffer;
 public:
-  Fatal() { std::cerr << "Fatal: "; }
+  Fatal() { buffer << "Fatal: "; }
   template<typename T> Fatal& operator<<(T arg) {
-    std::cerr << arg;
+    buffer << arg;
     return *this;
   }
   [[noreturn]] ~Fatal() {
-    std::cerr << "\n";
+    std::cerr << buffer.str() << std::endl;
     // Use _Exit here to avoid calling static destructors. This avoids deadlocks
     // in (for example) the thread worker pool, where workers hold a lock while
     // performing their work.
--- a/src/support/utilities.h
+++ b/src/support/utilities.h
@@ -73,7 +73,7 @@ public:
     // Use _Exit here to avoid calling static destructors. This avoids deadlocks
     // in (for example) the thread worker pool, where workers hold a lock while
     // performing their work.
-    _Exit(1);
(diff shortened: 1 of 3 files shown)


src/support/utilities.h
```diff
@@ -62,19 +63,33 @@ std::unique_ptr<T> make_unique(Args&&... args) {
...
     // in (for example) the thread worker pool, where workers hold a lock while
     // performing their work.
-    _Exit(1);
+    _Exit(EXIT_FAILURE);
   }
+#else
+  // This variation is a best-effort attempt to make fatal errors recoverable
+  // for embedders of Binaryen as a library, namely wasm-opt-rs.
+  [[noreturn]] ~Fatal() noexcept(false) {
+    if (std::uncaught_exceptions()) {
```
kripken, line 86: Is this part necessary? What bad thing happens if you just run the else arm unconditionally?

kripken (commented):
Sounds good to me!

We should document that flag perhaps. At least in the changelog, but maybe there could be a section in the readme for emdedding? That section might also link to the rust bindings actually.

kripken:
I see, thanks. That goes beyond my understanding of C++ exceptions :smile: I'd be ok either with removing it or a comment that explains it.

brson:
I've updated the THROW_ON_FATAL patch with comments about throwing in destructors and removed the explicit abort; and added a commit that mentions it in the changelog.

I have contacted you on discord about how to best document it.

I think there's a lint error yet that I am waiting on ci again to look at.

kripken (approved):
This seems reasonable for now. Perhaps we can improve things further later but I don't have an idea how atm, and there is no harm in adding this now to help the Rust bindings.

brson:
Updated for the lint.

PR #5089, opened 2022-09-28
[GUFA] Fix haveIntersection on comparing nullable with non-nullable

kripken:
We compared types and not heap types, so a difference in nullability
confused us. But at that point in the code, we've ruled out nulls, so we
should focus on heap types only.

This should unblock the binaryen roll into J2Wasm.


kripken:
In a cruel coincidence, I debugged this manually on j2wasm over several hours but the fuzzer just now found the exact same bug...

tlively (approved):

PR #5091, opened 2022-09-28
[GUFA] Improve hashing

kripken:
Avoid manually doing bitshifts etc. - leave combining to the core hash
logic, which can do a better job.

### DIFF
--- a/src/ir/possible-contents.h
+++ b/src/ir/possible-contents.h
@@ -21,6 +21,7 @@
 
 #include "ir/possible-constant.h"
 #include "ir/subtypes.h"
+#include "support/hash.h"
 #include "support/small_vector.h"
 #include "wasm-builder.h"
 #include "wasm.h"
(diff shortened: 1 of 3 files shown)


src/ir/possible-contents.h
```diff
@@ -213,23 +214,22 @@ class PossibleContents {
...
-    // contents.
-    if (isNone()) {
-      return 0;
+    // First hash the index of the variant, then add the internals for each.
+    size_t ret = std::hash<size_t>()(value.index());
+    if (isNone() || isMany()) {
+      // Nothing to add.
     } else if (isLiteral()) {
-      return size_t(1) | (std::hash<Literal>()(getLiteral()) << 3);
+      hash_combine(ret, std::hash<Literal>()(getLiteral()));
```
tlively, line 222: I think it would be slightly simpler to use `rehash(ret, getLiteral())` and similar elsewhere. This could also be implemented via `std::visit([&](auto val) { rehash(ret, val); }, value)`, but that would require hash implementations for none and many as well.

kripken: Oh nice, I'd forgotten about `rehash` I guess... done.

tlively (approved):

PR #5092, opened 2022-09-28
Fix nondeterminism in GlobalStructInference

kripken:
We append to vectors of globals in a nondeterministically-ordered loop, which can lead to
different orderings of the vectors. This happens quite frequently in very large J2Wasm
files it turns out. As a solution, simply sort them after the nondeterministic stage.

### DIFF
--- a/src/passes/GlobalStructInference.cpp
+++ b/src/passes/GlobalStructInference.cpp
@@ -176,12 +176,8 @@ struct GlobalStructInference : public Pass {
     // The above loop on typeGlobalsCopy is on an unsorted data structure, and
     // that can lead to nondeterminism in typeGlobals. Sort the vectors there to
     // ensure determinism.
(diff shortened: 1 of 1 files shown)


tlively (commented):
I see how `typeGlobals` can be nondeterministic, but how does that affect the output? I tried following it through, but this code is somewhat complicated.

kripken:
Basically, each entry in `typeGlobals` matches a type to the list of relevant globals. We then pick from that list a global to compare to. If the list is shuffled, we can end up picking another global that is equivalent to it. Without getting into all the details you can just see that `globals` vector is used in a way where its order matters, in `visitStructGet`.

kripken:
(I've confirmed the nondeterminism no longer happens after this PR, at least with very high likelihood... used to happen every other time, now it didn't happen once in 20 times.)

tlively (approved):
Thanks!

PR #5093, opened 2022-09-28
Refactor interaction between Pass and PassRunner

tlively:
Previously only WalkerPasses had access to the `getPassRunner` and
`getPassOptions` methods. Move those methods to `Pass` so all passes can use
them. As a result, the `PassRunner` passed to `Pass::run` and
`Pass::runOnFunction` is no longer necessary, so remove it.

Also update `Pass::create` to return a unique_ptr, which is more efficient than
having it return a raw pointer only to have the `PassRunner` wrap that raw
pointer in a `unique_ptr`.

Delete the unused template `PassRunner::getLast()`, which looks like it was
intended to enable retrieving previous analyses and has been in the code base
since 2015 but is not implemented anywhere.

This is nearly an NFC change, but I fixed a few places where pass options were
not being passed to nested `PassRunner`s and it looks like that had a very small
effect on the O3_converge test.

### DIFF
--- a/src/pass.h
+++ b/src/pass.h
@@ -450,7 +450,10 @@ public:
     assert(getPassRunner());
     // Parallel pass running is implemented in the PassRunner.
     if (isFunctionParallel()) {
-      PassRunner runner(module, getPassOptions());
+      // TODO: We should almost certainly be propagating pass options here, but
+      // that is a widespread change, so make sure it doesn't unacceptably
+      // regress compile times.
+      PassRunner runner(module /*, getPassOptions()*/);
       runner.setIsNested(true);
       runner.add(create());
       runner.run();
--- a/src/pass.h
+++ b/src/pass.h
@@ -425,7 +425,10 @@ public:
   std::string name;
 
   PassRunner* getPassRunner() { return runner; }
-  void setPassRunner(PassRunner* runner_) { runner = runner_; }
+  void setPassRunner(PassRunner* runner_) {
+    assert((!runner || runner == runner_) && "Pass already had a runner");
+    runner = runner_;
+  }
 
   PassOptions& getPassOptions() { return runner->options; }
 
--- a/src/pass.h
+++ b/src/pass.h
@@ -271,9 +271,7 @@ struct PassRunner {
   }
 
   // Add a pass given an instance.
-  template<class P> void add(std::unique_ptr<P> pass) {
-    doAdd(std::move(pass));
-  }
+  void add(std::unique_ptr<Pass> pass) { doAdd(std::move(pass)); }
 
   // Adds the pass if there are no DWARF-related issues. There is an issue if
   // there is DWARF and if the pass does not support DWARF (as defined by the
@@ -311,9 +309,6 @@ struct PassRunner {
 
   // Run the passes on a specific function
   void runOnFunction(Function* func);
-
-  // Get the last pass that was already executed of a certain type.
-  template<class P> P* getLast();
 
   // When running a pass runner within another pass runner, this
   // flag should be set. This influences how pass debugging works,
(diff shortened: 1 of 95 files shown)


tlively:
* main
  * **PR #5093** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5093" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5093?utm_source=stack-comment).

src/passes/pass.cpp
```diff
@@ -896,7 +896,9 @@ void PassRunner::runPass(Pass* pass) {
     checker = std::unique_ptr<AfterEffectModuleChecker>(
       new AfterEffectModuleChecker(wasm));
   }
-  pass->run(this, wasm);
+  assert(!pass->getPassRunner());
+  pass->setPassRunner(this);
+  pass->run(wasm);
```
kripken, line 901: How about resetting the runner to nullptr afterwards? Or is it the idea that once a pass has run it can't run again anyhow? That seems right to me, but I'm not 100% sure. Might be worth a comment here as to why we don't reset to null.

src/passes/pass.cpp
```diff
@@ -925,15 +927,18 @@ void PassRunner::runPassOnFunction(Pass* pass, Function* func) {
...
   if (passDebug) {
-    checker = std::unique_ptr<AfterEffectFunctionChecker>(
-      new AfterEffectFunctionChecker(func));
+    checker = std::make_unique<AfterEffectFunctionChecker>(func);
   }
-  instance->runOnFunction(this, wasm, func);
+
+  // Function-parallel passes get a new instance per function
+  auto instance = pass->create();
+  assert(instance);
```
kripken, line 937:
```suggestion
```
(next line would crash anyhow)

test/passes/converge_O3_metrics.bin.txt
```diff
@@ -164,19 +164,17 @@ total
...
-   (local.get $0)
+  (i32.store
+   (i32.const 40)
    (i32.const 2)
   )
   (drop
    (call $import$0
     (i32.const 146)
-    (local.get $0)
+    (i32.const 32)
```
kripken, line 177: Yes, looks like `precompute-propagate` did its thing here, then `coalese-locals` and `optimize-instructions` did the rest... so this proves `--optimize-level=3` was noticed in more places, as it runs `precompute-propagate`.

In general this seems good, but we should probably investigate more here, since `-O3` is much slower than even `-O2`. And in theory it can have downsides.

I would guess the place the change happens is `optimizeAfterInlining` if it now gets the options from its parent but it didn't before. In that case, it might be better to keep it at `--optimize-level=0`, which would make this NFC. Changing that to 3 may well make sense but we should benchmark both compile time and throughput, and that seems like it could be done separately later.

kripken (commented):
Nice simplification!

Overall sgtm, but let's figure out the NFC issue.

src/passes/pass.cpp
tlively: Will add a comment.

aheejin (commented):
Haven't checked every single line, but looks nice!

src/pass.h
```diff
@@ -367,18 +362,18 @@ struct PassRunner {
...
+  friend PassRunner;
+
 public:
   virtual ~Pass() = default;
 
   // Implement this with code to run the pass on the whole module
-  virtual void run(PassRunner* runner, Module* module) {
-    WASM_UNREACHABLE("unimplemented");
-  }
+  virtual void run(Module* module) { WASM_UNREACHABLE("unimplemented"); }
```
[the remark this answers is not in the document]
tlively: I don't think so, unfortunately. `Pass`es are supposed to implement either `run` or `runOnFunction`, but not both. If we made them pure virtual functions, then all passes would have to implement both instead.

src/pass.h
```diff
@@ -429,6 +424,11 @@ class Pass {
 
   std::string name;
 
+  PassRunner* getPassRunner() { return runner; }
+  void setPassRunner(PassRunner* runner_) { runner = runner_; }
```
kripken, line 428: Can maybe assert here that the old value was `nullptr`.

kripken (approved):

src/pass.h
tlively: There's a pattern where a function-parallel pass is run on the functions and then run on module code, and both code paths set the runner, so just checking for nullptr won't work. I can check for null or that the old runner is the same as the new runner, though.

kripken (approved):

PR #5094, opened 2022-09-29
[GUFA] Test that combined things intersect with the combination.

kripken:
Just an automatic test that `A` has an intersection with `A + B` for the contents
in our unit test.

### DIFF
--- a/test/gtest/possible-contents.cpp
+++ b/test/gtest/possible-contents.cpp
@@ -329,9 +329,9 @@ TEST_F(PossibleContentsTest, TestIntersectWithCombinations) {
   auto doTest = [](std::unordered_set<PossibleContents> set) {
     std::vector<PossibleContents> vec(set.begin(), set.end());
 
-    // Go over all permutations up to a certain size. 4 takes ~100ms, while 5
-    // takes several seconds, so don't go that high.
-    size_t max = 4;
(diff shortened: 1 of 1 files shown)


tlively (commented):
Since we have a fixed max number of elements anyway, how about doing this with three nested for loops instead? Also, I'm not sure how much extra confidence testing with sets of three elements gives beyond testing with sets of two elements. If you define `<` over possible contents, then you could really check that we're correctly implementing a lattice. In particular, you could check that for all `a` and `b` we have:

 -  `a < combine(a, b)`
 -  `b < combine(a, b)`
 -  `a < b => combine(a, b) = b`
 -  `b < a => combine(a, b) = a`
 -  `a < b, a != None => hasIntersection(a, b)`
 -  `b < a, b != None => hasIntersection(a, b)`
 -  `a < b, b < a => a = b`

kripken:
I found it useful to test with 4 and 5 locally, even if I decided to reduce to 3 for the PR. There is some chance of finding bugs.

Yes, we can add more checks here - sounds good. But note that that we don't have `<`, we only have "hasIntersection" - which limits us. But I want to add at least things like the order not mattering (`A+B+C = C+B+A` etc.).

tlively (approved):

PR #5096, opened 2022-09-30
BinaryenFunctionImportToFunction

volovyks:
- added `BinaryenFunctionImportToFunction` function
- fixed linter error


kripken (commented):
This is to turn an imported function into a non-imported one?

Another option might be to have a generic function to set the import module and base names. Setting them to nullptr could turn an import into a non-import. We could do this with a single function for all things - functions, globals, tables, memories, etc. - since they all inherit from `Importable`. Perhaps `BinaryenImportableSetModuleAndBase`? Or `BinaryenImportableSetNames` or `BinaryenImportableSetImportNames`...

volovyks:
Thanks for the suggestions @kripken. Here is what we want to achieve: https://github.com/near/wasi-stub/blob/main/wasi-stub.c. It is working, but we do not want to keep a fork just for 1 function.
If you think it is possible to do it with standard APIs - I can try to use Binaryen JS, since we are doing our work in JS env.

kripken:
Makes sense, yes, let's avoid you needing a fork for this. We should find the best way and add it.

Thinking on this more, how about making `BinaryenFunctionSetBody` unset the module and base fields, if the body is set to a non-null value? Adding a body means it is not imported, so it is the right behavior. I think that's simpler than adding a new API here, and it avoids possible bugs otherwise. If that makes sense, would like to change this PR to that?

PR #5101, opened 2022-09-30
[Fuzzing] Allow recombine() to replace with a subtype

kripken:
Previously it would randomly replace an expression with another one with the
exact same type. Allowing a subtype may give us more coverage.

### DIFF
--- a/src/tools/fuzzing/fuzzing.cpp
+++ b/src/tools/fuzzing/fuzzing.cpp
@@ -689,6 +689,7 @@ void TranslateToFuzzReader::recombine(Function* func) {
     Module& wasm;
     Scanner& scanner;
     TranslateToFuzzReader& parent;
+    bool needRefinalize = false;
 
     Modder(Module& wasm, Scanner& scanner, TranslateToFuzzReader& parent)
       : wasm(wasm), scanner(scanner), parent(parent) {}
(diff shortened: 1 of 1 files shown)


src/tools/fuzzing/fuzzing.cpp
```diff
@@ -623,8 +623,38 @@ void TranslateToFuzzReader::recombine(Function* func) {
...
+      auto heapType = type.getHeapType();
+      auto nullability = type.getNullability();
+
+      if (nullability == NonNullable) {
+        ret = getRelevantTypes(Type(heapType, Nullable));
+      }
+
+      while (1) {
+        ret.push_back(Type(heapType, nullability));
+        auto super = heapType.getSuperType();
```
tlively, line 650: Note that this won't capture basic heap types like `any` (although maybe it should).

tlively (approved):

src/tools/fuzzing/fuzzing.cpp
kripken: Hmm, this is probably good enough for now but I'll add a TODO. Unless there's a simple API to get basic supertypes too?

tlively: No, I don't think we have an API for that.

PR #5102, opened 2022-09-30
[Wasm GC] Fix .depth() and add testing

kripken:


src/wasm/wasm-type.cpp
```diff
@@ -1392,6 +1392,36 @@ size_t HeapType::getDepth() const {
...
     ++depth;
   }
+  // In addition to the explicit supertypes we just traversed over, there is
+  // implicit supertyping wrt basic types. A signature type always has one more
+  // super, HeapType::func, etc.
+  if (!isBasic()) {
+    if (isFunction()) {
+      depth++;
+    } else if (isStruct() || isArray()) {
+      // any < eq < data < specific struct types
```
tlively, line 1402:
```suggestion
      // specific struct types <: data <: eq <: any
```

src/wasm/wasm-type.cpp
```diff
@@ -1392,6 +1392,36 @@ size_t HeapType::getDepth() const {
   for (auto curr = *this; (super = curr.getSuperType()); curr = *super) {
     ++depth;
   }
+  // In addition to the explicit supertypes we just traversed over, there is
+  // implicit supertyping wrt basic types. A signature type always has one more
+  // super, HeapType::func, etc.
+  if (!isBasic()) {
+    if (isFunction()) {
+      depth++;
+    } else if (isStruct() || isArray()) {
```
tlively, line 1401:
```suggestion
    } else if (isData()) {
```

tlively (approved):
Another way to implement this fix would be to make `getSuperType()` return basic supertypes as well, which would be useful for your recent fuzzer PR.

kripken:
I tried the other approach with this:
```cpp
std::optional<HeapType> HeapType::getSuperType() const {
  if (isBasic()) {
    switch (getBasic()) {
      case HeapType::ext:
      case HeapType::func:
      case HeapType::any:
        return {};
      case HeapType::eq:
        return HeapType::any;
      case HeapType::i31:
      case HeapType::data:
      case HeapType::string:
      case HeapType::stringview_wtf8:
      case HeapType::stringview_wtf16:
      case HeapType::stringview_iter:
        return HeapType::eq;
    }
  }

  HeapTypeInfo* super = getHeapTypeInfo(*this)->supertype;
  if (super != nullptr) {
    return HeapType(uintptr_t(super));
  }

  if (isFunction()) {
    return HeapType::func;
  } else if (isData()) {
    return HeapType::data;
  }
  return {};
}
```
However this fails because some places in the code appear to assume that `getSuperType()` returns a non-basic type. The first error I got is in ModuleUtils which assumes it can call `getRecGroup` on the result of `getSuperType`. So it looks like that would need a larger refactoring. It does sound like a good idea, but for now let's start by landing this.

PR #5103, opened 2022-09-30
Avoid specifying multiple opt level in cflags

sbc100:
CMake defaults to -O3 to release builds.  We were overriding that to -O2 in general and re-overriding back to -O3 for emscripten.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -290,9 +290,7 @@ else()
   elseif(NOT EMSCRIPTEN)
     add_compile_flag("-fPIC")
   endif()
-  add_debug_compile_flag("-O0")
   add_debug_compile_flag("-g3")
-  add_nondebug_compile_flag("-O2")
   if(BYN_ENABLE_ASSERTIONS)
     # On non-Debug builds cmake automatically defines NDEBUG, so we
     # explicitly undefine it:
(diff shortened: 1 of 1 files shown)


sbc100:
Is there some reason we don't want -O3 by default on all platforms?

kripken (commented):
CMake's default is probably fine. Is `-O3` fixed though in CMake, or does it change with the versions now and then?

sbc100:
> CMake's default is probably fine. Is `-O3` fixed though in CMake, or does it change with the versions now and then?

Seems pretty fixed to me.  I guess maybe a specific toolchain could override it.

Here is where cmake itself defines this stuff:
https://github.com/Kitware/CMake/blob/c7170651f48bac9e6b80c58f5e97f687d9f6c302/Modules/Compiler/GNU.cmake#L55-L60

The blame for those lines is 6 years old.. but even before that the values where the same.

kripken:
I'd be happier if this was mentioned in CMake's docs somewhere, as something one can rely on.

For emscripten builds as you know the difference between `-O2` and `-O3` is quite large, so I'd be worried about any change in CMake that could affect that. For that reason I'd prefer to keep the emscripten one fixed to `-O3`. But I don't feel strongly.

sbc100:
> I'd be happier if this was mentioned in CMake's docs somewhere, as something one can rely on.
> 
> For emscripten builds as you know the difference between `-O2` and `-O3` is quite large, so I'd be worried about any change in CMake that could affect that. For that reason I'd prefer to keep the emscripten one fixed to `-O3`. But I don't feel strongly.

OK, I added an assertion to ensure that cmake does set `-O3` as expected.

kripken (approved):

PR #5105, opened 2022-10-01
Make Asyncify work with wasm64

sbc100:
The emscripten side is a little tricky but I've got some tests passing. Currently blocked on:
https://github.com/emscripten-core/emscripten/issues/17969


src/passes/Asyncify.cpp
```diff
@@ -795,30 +802,35 @@ static bool doesCall(Expression* curr) {
...
+                    makeGlobalGet(ASYNCIFY_DATA, pointerType),
+                    pointerType,
                     wasm.memories[0]->name);
   }
 
   Expression* makeIncStackPos(int32_t by) {
     if (by == 0) {
       return makeNop();
     }
+    Literal literal = is64 ? Literal(int64_t{by}) : Literal(int32_t{by});
```
kripken, line 826:
```suggestion
    auto literal = Literal::makeFromInt64(by, pointerType);
```
I think with this we can avoid the `bool is64` on the class.

src/passes/Asyncify.cpp
```diff
@@ -795,30 +802,35 @@ static bool doesCall(Expression* curr) {
...
     return makeStore(
       4,
       int32_t(DataOffset::BStackPos),
       4,
-      makeGlobalGet(ASYNCIFY_DATA, Type::i32),
-      makeBinary(AddInt32, makeGetStackPos(), makeConst(Literal(by))),
-      Type::i32,
+      makeGlobalGet(ASYNCIFY_DATA, pointerType),
+      makeBinary(is64 ? AddInt64 : AddInt32, makeGetStackPos(), makeConst(literal)),
```
kripken, line 832:
```suggestion
      makeBinary(Abstract::getBinary(pointerType, Abstract::Add), makeGetStackPos(), makeConst(literal)),
```

src/passes/Asyncify.cpp
kripken: (I mean with changes like this and the one below - then all we need is `pointerType`)

src/passes/Asyncify.cpp
```diff
@@ -795,31 +802,36 @@ static bool doesCall(Expression* curr) {
 class AsyncifyBuilder : public Builder {
 public:
   Module& wasm;
+  Type pointerType;
 
-  AsyncifyBuilder(Module& wasm) : Builder(wasm), wasm(wasm) {}
+  AsyncifyBuilder(Module& wasm)
+    : Builder(wasm), wasm(wasm),
+      pointerType(wasm.memories[0]->is64() ? Type::i64 : Type::i32) {}
```
kripken, line 809:
```suggestion
      pointerType(wasm.memories[0]->indexType) {}
```

src/passes/Asyncify.cpp
```diff
@@ -1497,6 +1510,8 @@ struct Asyncify : public Pass {
   void run(Module* module) override {
     auto& options = getPassOptions();
     bool optimize = options.optimizeLevel > 0;
+    is64 = module->memories.size() && module->memories[0]->is64();
+    pointerType = is64 ? Type::i64 : Type::i32;
```
kripken, line 1514:
```suggestion
    pointerType = module->memories[0]->indexType;
```

src/passes/Asyncify.cpp
```diff
@@ -795,31 +802,36 @@ static bool doesCall(Expression* curr) {
...
   Module& wasm;
+  Type pointerType;
 
-  AsyncifyBuilder(Module& wasm) : Builder(wasm), wasm(wasm) {}
+  AsyncifyBuilder(Module& wasm)
+    : Builder(wasm), wasm(wasm),
+      pointerType(wasm.memories[0]->is64() ? Type::i64 : Type::i32) {}
 
   Expression* makeGetStackPos() {
     return makeLoad(4,
```
kripken, line 812: Doesn't the load size need to be 8 now? This is loading a pointer, which is 8 bytes.

src/passes/Asyncify.cpp
sbc100: IIRC the problem here is that module->memories might be empty.   I guess asyncify should probably error out in that case?    But that seems like a bigger change.

kripken: Oh, right, sorry, my suggested diff here isn't right then. Still, using `indexType` might simplify things.

src/passes/Asyncify.cpp
sbc100: Yes, you are right, updated here and elsewhere.

kripken (approved):

PR #5107, opened 2022-10-03
Multi-Memories Lowering Pass

ashleynh:
Adds a multi-memories lowering pass that will create a single combined memory from the memories added to the module. This pass assumes that each memory is configured the same (type, shared).

This pass also:
- replaces existing memory.size instructions with a custom function that returns the size of each memory as if they existed independently
- replaces existing memory.grow instructions with a custom function, using global offsets to track the page size of each memory so data doesn't overlap in the singled combined memory
- adjusts the offsets of active data segments
- adjusts the offsets of Loads/Stores


test/lit/passes/multi-memory-lowering.wast
```diff
@@ -0,0 +1,65 @@
...
+
+(module
+  (memory $memory1 1)
+  (memory $memory2 2)
+  (data (memory $memory1) (i32.const 0) "a")
+  (data (memory $memory2) (i32.const 1) "123")
+)
+
+;; CHECK: (module
+ ;; CHECK-NEXT:(type $none_=>_i32 (func (result i32)))
```
tlively, line 12: Did update_lit_checks.py emit these unaligned checks? This seems like a bug in the script if so. It looks like CHECK prefix and the whitespace that is supposed to come after it have been swapped.

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,345 @@
...
+        Index idx = iter->second - 1;
+        curr->ptr = builder.makeGlobalGet(parent.offsetGlobalNames[idx],
+                                          parent.pointerType);
+      }
+    };
+    Replacer(*this, *wasm).run(runner, wasm);
+  }
+
+  void addCombinedMemory() {
+    // We are assuming that each memory is configured the same as the first
```
kripken, line 144: Perhaps we should assert on them all having the same index type?

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,345 @@
...
+  }
+
+  void addOffsetGlobals() {
+    auto addGlobal = [&](Name name, size_t offset) {
+      auto global = Builder::makeGlobal(
+        name,
+        pointerType,
+        Builder(*wasm).makeConst(pointerType == Type::i32
+                                   ? Literal(int32_t(offset))
+                                   : Literal(int64_t(offset))),
```
kripken, line 169: This can use `Literal::makeFromInt64(offset, pointerType)`

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,345 @@
...
+      auto iter = memoryIdxMap.find(dataSegment->memory);
+      assert(iter != memoryIdxMap.end());
+      dataSegment->memory = combinedMemory;
+      // No need to update the offset of data segments for the first memory
+      if (iter->second != 0) {
+        // We need to subtract 1 here because there is no offsetGlobal for the
+        // first memory
+        auto offsetGlobalName = offsetGlobalNames[iter->second - 1];
+        dataSegment->offset = builder.makeBinary(
+          pointerType == Type::i32 ? AddInt32 : AddInt64,
```
kripken, line 205: This can use `Abstract::getBinary(pointerType, Abstract::Add)`

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,345 @@
...
+  }
+
+  // This function replaces memory.size instructions with a function that can
+  // return the size of each memory as if each was discrete and separate.
+  std::unique_ptr<Function> memorySize(Index memIdx) {
+    Builder builder(*wasm);
+    Name functionName = Names::getValidFunctionName(*wasm, "multi_memory_size");
+    auto function = Builder::makeFunction(
+      functionName, Signature(Type::none, pointerType), {});
+    function->hasExplicitName = true;
```
kripken, line 302: Why is this needed? I suspect it can be removed.

kripken (commented):
Nice work! Some comments on the code (haven't gotten to the test yet).

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,345 @@
...
+        addGlobal(name, offsetRunningTotal);
+      }
+      offsetRunningTotal += memory->initial;
+    }
+  }
+
+  void adjustActiveDataSegmentOffsets() {
+    Builder builder(*wasm);
+    ModuleUtils::iterActiveDataSegments(*wasm, [&](DataSegment* dataSegment) {
+      assert(dataSegment->offset->is<Const>());
```
tlively, line 195:
```suggestion
      assert(dataSegment->offset->is<Const>() && "TODO: handle non-const segment offsets");
```

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,345 @@
...
+      adjustOffsetNames.push_back(function->name);
+      wasm->addFunction(std::move(function));
+    }
+  }
+
+  // This function replaces memory.grow instruction calls in the wasm module.
+  // Because the multiple discrete memories are lowered into a single memory,
+  // we need to adjust offsets as a particular memory receives an
+  // instruction to grow.
+  std::unique_ptr<Function> adjustOffsetGlobals(Index memIdx) {
```
tlively, line 236: How about calling this `memoryGrow`, like how you call the helper function for memory.size `memorySize`?

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,345 @@
...
+    auto getPageDelta = [&]() { return builder.makeLocalGet(0, pointerType); };
+    Expression* functionBody;
+
+    Index returnLocal = Builder::addVar(function.get(), pointerType);
+    functionBody = builder.blockify(builder.makeLocalSet(
+      returnLocal, builder.makeCall(memorySizeNames[memIdx], {}, pointerType)));
+
+    // If we are not growing the last memory, then we need to copy data,
+    // shifting it over to accomodate the increase from page_delta
+    if (memIdx != offsetGlobalNames.size()) {
```
tlively, line 253: Since this also depends on the detail that `offsetGlobalNames` contains N-1 memories, it would be good to make a helper function like `isLastMemory` for this.

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,345 @@
...
+    function->hasExplicitName = true;
+    function->setLocalName(0, "page_delta");
+    auto getPageDelta = [&]() { return builder.makeLocalGet(0, pointerType); };
+    Expression* functionBody;
+
+    Index returnLocal = Builder::addVar(function.get(), pointerType);
+    functionBody = builder.blockify(builder.makeLocalSet(
+      returnLocal, builder.makeCall(memorySizeNames[memIdx], {}, pointerType)));
+
+    // If we are not growing the last memory, then we need to copy data,
```
tlively, line 251: I would expect to see the real `memory.grow` before copying the data.

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,345 @@
...
+  }
+
+  // This function replaces memory.grow instruction calls in the wasm module.
+  // Because the multiple discrete memories are lowered into a single memory,
+  // we need to adjust offsets as a particular memory receives an
+  // instruction to grow.
+  std::unique_ptr<Function> adjustOffsetGlobals(Index memIdx) {
+    Builder builder(*wasm);
+    Name functionName =
+      Names::getValidFunctionName(*wasm, "adjust_memory_offsets");
```
tlively, line 239: It would be good to explicitly put the memory index into the function name. Right now the auto-generated indices appended to the names are off by one.

tlively:
Oops, I hadn't seen @kripken's comments in the meantime. It looks like there's a good amount of overlap. Sorry about that!

test/lit/passes/multi-memory-lowering.wast
ashleynh: Talked offline

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,345 @@
...
+        assert(iter != parent.memoryIdxMap.end());
+        // No need to change the offset for the first memory
+        if (iter->second == 0) {
+          return;
+        }
+        // There is no offset global for the first memory, so we need to
+        // subtract one when indexing into the offsetGlobalName vector
+        Index idx = iter->second - 1;
+        curr->ptr = builder.makeGlobalGet(parent.offsetGlobalNames[idx],
+                                          parent.pointerType);
```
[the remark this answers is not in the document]
ashleynh: Great catch, thanks

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,345 @@
...
+        replaceCurrent(builder.makeCall(funcName, {}, curr->type));
+      }
+
+      // We diverge from the spec here and are not trapping if the offset + type
+      // / 8 is larger than the length of the memory's data Warning,
+      // out-of-bounds loads and stores can read junk out of or corrupt other
+      // memories instead of trapping
+      void visitLoad(Load* curr) {
+        curr->memory = parent.combinedMemory;
+        auto iter = parent.memoryIdxMap.find(curr->memory);
```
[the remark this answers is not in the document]
ashleynh: Great catch, thanks!

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,345 @@
...
+          pointerType == Type::i32 ? AddInt32 : AddInt64,
+          builder.makeGlobalGet(offsetGlobalName, pointerType),
+          dataSegment->offset);
+      }
+    });
+  }
+
+  void createMemorySizeFunctions() {
+    // Don't create a memory size function for the last memory in the vector as
+    // its the combinedMemory we just added
```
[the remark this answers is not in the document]
ashleynh: Used helper functions to avoid subtracting 1 in various places

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,394 @@
...
+  // for each memory
+  std::vector<Name> memorySizeNames;
+  // A vector of the memory grow functions that were created proactively for
+  // each memory
+  std::vector<Name> memoryGrowNames;
+
+  void run(Module* module) override {
+    // Disabling multi-memories so that the target features esction in the
+    // emitted module does not report the use of MultiMemories. Also prevents
+    // later passes from adding additional memories.
```
kripken, line 63: Maybe move this comment to the top comment at the start of the file? It's important for people using the pass to know it I think.

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,394 @@
...
+        auto idx = parent.memoryIdxMap.at(curr->memory);
+        Name funcName = parent.memoryGrowNames[idx];
+        replaceCurrent(builder.makeCall(funcName, {}, curr->type));
+      }
+
+      // We diverge from the spec here and are not trapping if the offset +
+      // write size is larger than the length of the memory's data. Warning:
+      // out-of-bounds loads and stores can read junk out of or corrupt other
+      // memories instead of trapping.
+      // TODO: Add an option to add bounds checks.
```
kripken, line 118: Also might be best in the top comment at the start.

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,394 @@
...
+
+  void addOffsetGlobals() {
+    auto addGlobal = [&](Name name, size_t offset) {
+      auto global =
+        Builder::makeGlobal(name,
+                            pointerType,
+                            Builder(*wasm).makeConst(
+                              pointerType == Type::i32
+                                ? Literal::makeFromInt32(offset, pointerType)
+                                : Literal::makeFromInt64(offset, pointerType)),
```
kripken, line 213:
```suggestion
                              Literal::makeFromInt64(offset, pointerType)),
```
This works out: we create something of the right type, and if the type is 32, the offset is truncated to 32 bits anyhow, either way. That is, `makeFromInt64` is safe to use to create 32-bit numbers.

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,394 @@
...
+                            Builder::Mutable);
+      global->hasExplicitName = true;
```
kripken, line 215: I think there was another discussion about this for functions - I think we don't need it for globals either.

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,394 @@
...
+                         getPageDelta(), combinedMemory, memoryInfo)));
+
+    // If we are not growing the last memory, then we need to copy data,
+    // shifting it over to accomodate the increase from page_delta
+    if (!isLastMemory(memIdx)) {
+      // This offset is the starting pt for copying
+      auto& offsetGlobalName = offsetGlobalNames[memIdx];
+      functionBody = builder.blockify(
+        functionBody,
+        builder.makeMemoryCopy(
```
kripken, line 303: Really nice how `MemoryCopy` makes this relatively easy... without that this would be much less fun.

kripken (commented):
Nice!

Looks correct to me, remaining comments are minor.

It might be useful to manually check this on a wasm-split codebase, for a real-world testcase. That is, using a second memory and then lowering it back down should not cause any runtime errors.

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,394 @@
...
+        builder.makeMemorySize(combinedMemory, memoryInfo),
+        builder.makeGlobalGet(offsetGlobalName, pointerType)));
+    } else {
+      auto offsetGlobalName = getOffsetGlobal(memIdx);
+      auto nextOffsetGlobalName = getOffsetGlobal(memIdx + 1);
+      functionBody = builder.makeReturn(builder.makeBinary(
+        Abstract::getBinary(pointerType, Abstract::Sub),
+        builder.makeGlobalGet(nextOffsetGlobalName, pointerType),
+        builder.makeGlobalGet(offsetGlobalName, pointerType)));
+    }
```
kripken, line 372: There are three cases here, but the test only has two memories so I think we are missing test coverage for one of them.

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,394 @@
...
+  }
+
+  // This function replaces memory.grow instruction calls in the wasm module.
+  // Because the multiple discrete memories are lowered into a single memory,
+  // we need to adjust offsets as a particular memory receives an
+  // instruction to grow.
+  std::unique_ptr<Function> memoryGrow(Index memIdx) {
+    Builder builder(*wasm);
+    Name name = "custom_memory_grow_" + std::to_string(memIdx);
```
tlively, line 277: I don't think the "custom_" prefix adds much here. Maybe we can put the original memory name into the name instead like we do with the offset globals?
```suggestion
    Name name = "memory_grow_" + std::to_string(memIdx);
```

test/lit/passes/multi-memory-lowering.wast
```diff
@@ -0,0 +1,82 @@
...
+;; CHECK-NEXT:  (i32.const 1)
+;; CHECK-NEXT: ) "123")
+
+;; CHECK:      (func $custom_memory_size_0 (result i32)
+;; CHECK-NEXT:  (return
+;; CHECK-NEXT:   (global.get $memory2_page_offset)
+;; CHECK-NEXT:  )
+;; CHECK-NEXT: )
+
+;; CHECK:      (func $custom_memory_size_1 (result i32)
```
tlively, line 32: It would be good to add a middle memory so we test all three cases for memory_size.

test/lit/passes/multi-memory-lowering.wast
```diff
@@ -0,0 +1,82 @@
...
+;; CHECK-NEXT:   )
+;; CHECK-NEXT:  )
+;; CHECK-NEXT:  (memory.copy
+;; CHECK-NEXT:   (i32.add
+;; CHECK-NEXT:    (global.get $memory2_page_offset)
+;; CHECK-NEXT:    (local.get $page_delta)
+;; CHECK-NEXT:   )
+;; CHECK-NEXT:   (global.get $memory2_page_offset)
+;; CHECK-NEXT:   (i32.sub
+;; CHECK-NEXT:    (memory.size)
```
tlively, line 58: This size isn't correct anymore since we have already grown the underlying memory.

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,394 @@
...
+             "TODO: handle non-const segment offsets");
+      auto idx = memoryIdxMap.at(dataSegment->memory);
+      dataSegment->memory = combinedMemory;
+      // No need to update the offset of data segments for the first memory
+      if (idx != 0) {
+        // We need to subtract 1 here because there is no offsetGlobal for the
+        // first memory
+        auto offsetGlobalName = getOffsetGlobal(idx);
+        dataSegment->offset = builder.makeBinary(
+          Abstract::getBinary(pointerType, Abstract::Add),
```
kripken, line 248: I believe an add in a data segment offset requires the ExtendedConst feature, so we should error if `!wasm->features.hasExtendedConst()`

src/passes/MultiMemoryLowering.cpp
ashleynh: Good idea, thanks!

src/passes/MultiMemoryLowering.cpp
ashleynh: Played around with a few combinations for name, I think memoryName + "_grow" is the most readable. dropping the memidx. Doing the same naming scheme for the replacement memory size function names.

test/lit/passes/multi-memory-lowering.wast
ashleynh: Great catch, thanks!

test/lit/passes/multi-memory-lowering.wast
```diff
@@ -0,0 +1,214 @@
...
+;; CHECK-NEXT:  (return
+;; CHECK-NEXT:   (global.get $memory2_page_offset)
+;; CHECK-NEXT:  )
+;; CHECK-NEXT: )
+
+;; CHECK:      (func $memory2_size (result i32)
+;; CHECK-NEXT:  (return
+;; CHECK-NEXT:   (i32.sub
+;; CHECK-NEXT:    (global.get $memory3_page_offset)
+;; CHECK-NEXT:    (global.get $memory2_page_offset)
```
kripken, line 115: This looks wrong - `memory.size` returns a value in page size units, so I think this needs to be divided by the page size.

`memory.grow` also takes page sizes. I think the only place that needs the actual byte offsets are loads and stores.

kripken (commented):
Great! lgtm % the issue in the comment

test/lit/passes/multi-memory-lowering.wast
ashleynh: Spoke with Alon offline and explained that the (memoryName)_page_offset globals created for each memory are in page units. As a result, the load/store offsets need to change, with the page_offset being multiplied by the page size to get the byte offset that will be added to the existing load/store offset.

Given that we expect there to be far more load/store instructions than memory.grow/ memory.size, I’ll change the (memoryName)_page_offset globals to instead store the byte offset so that we’re able to just conduct a single add operation in each load/store instruction. For memory.grow/memory.size, I’ll divide the byte offset by the page size to get the page units needed for these instructions.

tlively (commented):
LGTM % that issue as well 👍

src/passes/MultiMemoryLowering.cpp
```diff
@@ -0,0 +1,431 @@
...
+    Expression* functionBody;
+    Index sizeLocal = -1;
+
+    Index returnLocal =
+      Builder::addVar(function.get(), "return_size", pointerType);
+    functionBody = builder.blockify(builder.makeLocalSet(
+      returnLocal, builder.makeCall(memorySizeNames[memIdx], {}, pointerType)));
+
+    if (!isLastMemory(memIdx)) {
+      pageSizeLocal = Builder::addVar(function.get(), "page_size", Type::i32);
```
kripken, line 301: A page size local can be avoided, as the `makeConst` can just be done wherever the `local.get` of the local is atm. (There isn't much cost to having multiple appearances of it, too, as it is just a constant. Other things might be different if they take work at runtime though, like an add.)

kripken (approved):
Great!

lgtm % last comment

PR #5108, opened 2022-10-03
Fix ordering of visit() in MemoryGrow interpretation

kripken:
This is a pretty subtle point that we all missed in #4811 - we need to first visit the
child, then compute the size, as the child may alter that size.

Found by the fuzzer.


tlively (approved):
Oh wow, that is subtle.

ashleynh (approved):
Nice fix!

PR #5079, opened 2022-09-22
Emit call_ref with a type annotation

tlively:
Emit call_ref instructions with type annotations and a temporary opcode. Also
implement support for parsing optional type annotations on call_ref in the text
and binary formats. This is part of a multi-part graceful update to switch
Binaryen and all of its users over to using the type-annotated version of
call_ref without there being any breakage.

PR #5110, opened 2022-10-04
Simplify and fix heap type counting

tlively:
Annotations on array.get and array.set were not being counted and the code could
generally be simplified since `count` already ignores types that don't need to
be counted.


tlively:
* main
  * **PR #5110** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5110" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5110?utm_source=stack-comment).

src/ir/module-utils.cpp
```diff
@@ -54,17 +54,15 @@ struct CodeScanner
     if (auto* call = curr->dynCast<CallIndirect>()) {
       counts.note(call->heapType);
     } else if (auto* call = curr->dynCast<CallRef>()) {
-      if (call->isReturn && call->target->type.isFunction()) {
```
kripken, line 57: The `isReturn` change makes us count in the way that is correct in the spec but we haven't yet moved to, I think? But the idea is that there's no harm to doing that for now, and eventually this code will become right in binaryen as well?

kripken (approved):

src/ir/module-utils.cpp
tlively: We emit a type annotation for non-return `call_ref` as well since #5079, so this is necessary for correctness.

PR #5114, opened 2022-10-05
wasm2js: Support for flexible module import naming.

sbc100:
The previous code was making emscripten-specific assumptions about imports basically all coming from the `env` module.

I can't find a way to make this backwards compatible so may do a combined roll with the emscripten-side change:
https://github.com/emscripten-core/emscripten/pull/17806

### DIFF
--- a/src/wasm2js.h
+++ b/src/wasm2js.h
@@ -53,7 +53,7 @@ namespace wasm {
 
 using namespace cashew;
 
-static IString importObject("importObject");
+static IString importObject("imports");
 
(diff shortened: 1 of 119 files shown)


kripken (approved):

kripken:
This seems to have broken fuzzing, e.g. `./scripts/fuzz_opt.py --auto-initial-contents 7044408155933374954`
```
var fimport$0 = fuzzing_support["log-i32"];                                                   
                                ^                                                                                                                                                                                                                                                         
TypeError: Cannot read properties of undefined (reading 'log-i32')
```

PR #5112, opened 2022-10-05
Pin v8 10.8.104

tlively:
More recent version of V8 include a change from `dataref` to `structref` that
prevent WasmGC modules produced by the fuzzer from validating. Use the last
compatible v8 version if it is in the path until we can update Binaryen to be
compatible with newer v8.

PR #4843, opened 2022-07-29
Make `i31ref` and `dataref` nullable

tlively:
Match the latest version of the GC spec. This change does not depend on V8
changing its interpretation of the shorthands because we are still temporarily
not emitting the binary shorthands, but all Binaryen users will have to update
their interpretations along with this change if they use the text or binary
shorthands.

PR #5115, opened 2022-10-05
Implement bottom heap types

tlively:
These types, `none`, `nofunc`, and `noextern` are uninhabited, so references to
them can only possibly be null. To simplify the IR and increase type precision,
introduce new invariants that all `ref.null` instructions must be typed with one
of these new bottom types and that `Literals` have a bottom type iff they
represent null values. These new invariants requires several additional changes.

First, it is now possible that the `ref` or `target` child of a `StructGet`,
`StructSet`, `ArrayGet`, `ArraySet`, or `CallRef` instruction has a bottom
reference type, so it is not possible to determine what heap type annotation to
emit in the binary or text formats. (The bottom types are not valid type
annotations since they do not have indices in the type section.)

To fix that problem, update the printer and binary emitter to emit unreachables
instead of the instruction with undetermined type annotation. This is a valid
transformation because the only possible value that could flow into those
instructions in that case is null, and all of those instructions trap on nulls.

That fix uncovered a latent bug in the binary parser in which new unreachables
within unreachable code were handled incorrectly. This bug was not previously
found by the fuzzer because we generally stop emitting code once we encounter an
instruction with type `unreachable`. Now, however, it is possible to emit an
`unreachable` for instructions that do not have type `unreachable` (but are
known to trap at runtime), so we will continue emitting code. See the new
test/lit/parse-double-unreachable.wast for details.

Update other miscellaneous code that creates `RefNull` expressions and null
`Literals` to maintain the new invariants as well.


tlively:
* main
  * **PR #5110** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5110" target="_blank">Graphite</a> 
    * **PR #5112** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5112" target="_blank">Graphite</a> 
      * **PR #5115** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5115" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5115?utm_source=stack-comment).

src/wasm/wasm.cpp
```diff
@@ -796,7 +796,10 @@ void MemoryGrow::finalize() {
   }
 }
 
-void RefNull::finalize(HeapType heapType) { type = Type(heapType, Nullable); }
+void RefNull::finalize(HeapType heapType) {
+  assert(heapType.isBottom());
+  type = Type(heapType, Nullable);
```
kripken, line 801:
```suggestion
  type = Type(heapType, Nullable);
  assert(type.isNull());
```
(maybe excessive though?)

kripken (commented):
Nice!

Comments on the code so far.

Please update the changelog.

src/ir/effects.h
```diff
@@ -701,6 +701,10 @@ class EffectAnalyzer {
       }
     }
     void visitCallRef(CallRef* curr) {
+      if (curr->target->type.isNull()) {
+        parent.trap = true;
```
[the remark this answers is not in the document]
tlively: Are you thinking that we should be setting `implicitTrap` when the target is null rather than setting `trap`? I think it's more consistent to set `trap`, just like `visitUnreachable` does. In particular, the `CallRef` will be emitted as `(unreachable)` in this case, so it would be strange if its effects were different from `Unreachable`'s effects.

You're probably right that optimizations would take care of this anyway, but the code seems more correct to me as-is.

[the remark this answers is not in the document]
kripken: I was thinking we can set `implicitTrap` when the target is nullable, which would include null. You're right that the current code is more precise in a way, but I don't think we gain from the extra verbosity.

I guess it's fine, though, if we make sure that it doesn't limit us. Specifically, with trapsNeverHappen and the new code then a callRef of a null would no longer be removed by vacuum (like it doesn't remove an unreachable). We should have a test that optimize-instructions turns the callRef into an unreachable (which dce would then remove). (There might already be a test, sorry if so - I haven't gotten to those yet.)

src/wasm-builder.h
```diff
@@ -699,10 +699,11 @@ class Builder {
   }
   RefNull* makeRefNull(HeapType type) {
     auto* ret = wasm.allocator.alloc<RefNull>();
-    ret->finalize(Type(type, Nullable));
+    ret->finalize(Type(type.getBottom(), Nullable));
     return ret;
   }
   RefNull* makeRefNull(Type type) {
+    assert(type.isNullable() && type.getHeapType().isBottom());
```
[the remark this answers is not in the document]
tlively: This isn't quite the same, since `isNull` also includes non-nullable references to bottom types (which are not inhabitable). I'll change this to `type.isNullable() && type.isNull()`.

src/wasm-type.h
```diff
@@ -421,6 +430,8 @@ class HeapType {
   std::string toString() const;
 };
 
+inline bool Type::isNull() const { return isRef() && getHeapType().isBottom(); }
```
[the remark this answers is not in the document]
tlively: Unfortunately we can't put the definition in the normal place because it uses `HeapType` by value and therefore has to come after the `HeapType` definition.

We don't want to require any nullability here because most users do not need to differentiate between nullable and non-nullable references to bottom types. For example, in the printer and binary emitter, we need the same behavior when the `ref` child is a reference to a bottom type no matter whether it is nullable or not.

In principle we could optimize even more when we see a non-nullable reference to a bottom type, but we don't do that yet.

src/wasm-builder.h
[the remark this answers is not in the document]
kripken: Hmm, how can we have a non-nullable reference to a bottom type? (what does that look like in wasm?)

src/ir/effects.h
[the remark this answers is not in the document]
tlively: It seems odd that `trapsNeverHappen` doesn't remove an `unreachable`, but I guess that's a separate topic.

Will make sure I add specific tests for the OptimizeInstructions changes.

src/wasm-builder.h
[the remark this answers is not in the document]
tlively: `(ref none)` is a perfectly valid type, it's just not possible to create a value that inhabits it. So I could have a function that returns `(ref none)`, for instance, and know that it will never return.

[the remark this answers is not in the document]
kripken: Oh, really.. I would have expected the wasm spec to disallow that. Just me I guess...

src/wasm/literal.cpp
```diff
@@ -125,7 +142,7 @@ Literal::~Literal() {
   if (type.isBasic()) {
     return;
   }
-  if (isData()) {
+  if (isNull() || isData()) {
```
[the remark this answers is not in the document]
tlively: It seems simpler to keep the 1:1 correspondence between `Literal::isData` and `Type::isData`. Now that Literals representing null values have reference-to-bottom types, they no longer have types satisfying `Type::isData`.

src/wasm/wasm-binary.cpp
```diff
@@ -2844,7 +2898,12 @@ void WasmBinaryBuilder::skipUnreachableCode() {
       expressionStack = savedStack;
       return;
     }
-    pushExpression(curr);
+    if (curr->type == Type::unreachable) {
+      // Don't want to pop back past the unreachable.
+      expressionStack.clear();
```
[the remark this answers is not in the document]
tlively: Sure, will expand the comment.

src/wasm/wasm-validator.cpp
```diff
@@ -2079,6 +2079,10 @@ void FunctionValidator::visitSelect(Select* curr) {
 }
 
 void FunctionValidator::visitDrop(Drop* curr) {
+  if (!curr->value->type.isConcrete() &&
+      curr->value->type != Type::unreachable) {
+    assert(false);
```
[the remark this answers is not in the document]
tlively: Looks like this is left over from some debugging. Will remove.

src/wasm/wasm-validator.cpp
```diff
@@ -2110,13 +2114,13 @@ void FunctionValidator::visitRefNull(RefNull* curr) {
...
-  shouldBeTrue(curr->type.getFeatures() <= getModule()->features,
-               curr->type,
-               "ref.null type should be allowed");
+  if (!shouldBeTrue(
+        curr->type.isNullable(), curr, "ref.null types must be nullable")) {
+    return;
+  }
+  shouldBeTrue(curr->type.getHeapType().isBottom(),
+               curr,
+               "ref.null must have a bottom heap type");
```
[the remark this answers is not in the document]
tlively: Sure, since we already checked that it's nullable, this can be `isNull()`.

src/wasm/wasm.cpp
tlively: Yeah, this seems excessive since we already checked that `heapType.isBottom()`.

src/wasm/literal.cpp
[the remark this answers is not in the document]
kripken: I see. Seems reasonable. Maybe update the comments on `isData` to clarify that.

I'd also suggest auditing all places with `isData` in the codebase, though, as they all assume nulls there as well atm.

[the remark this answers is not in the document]
tlively: Nice, looking now and this is letting me delete a bunch of code 👍

test/lit/passes/remove-unused-module-elements-refs.wast
```diff
@@ -18,10 +18,16 @@
...
+  ;; CHECK-NEXT:  (block
+  ;; CHECK-NEXT:   (drop
+  ;; CHECK-NEXT:    (ref.null nofunc)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:   (unreachable)
   ;; CHECK-NEXT:  )
   ;; CHECK-NEXT:  (block
+  ;; CHECK-NEXT:   (drop
+  ;; CHECK-NEXT:    (unreachable)
+  ;; CHECK-NEXT:   )
```
kripken, line 30: I think this PR makes the pass no longer test what it should,

https://github.com/WebAssembly/binaryen/blob/ddfebfa9b9014b80072ca77232181bd8e22c28ee/test/lit/passes/remove-unused-module-elements-refs.wast#L61-L62

We need an actual `call_ref` of the relevant type, but with this PR no such `call_ref` exists in IR even though there is one in the text.

If null doesn't work anymore, we could perhaps do a `ref.cast` of a null to that type? (Or we could use a `ref.func` to get that type, but that would take some more careful reading.)

test/lit/passes/optimize-instructions-gc-tnh.wast
```diff
@@ -53,17 +53,7 @@
...
-  ;; TNH-NEXT:    (block (result (ref null $struct))
-  ;; TNH-NEXT:     (drop
-  ;; TNH-NEXT:      (ref.null any)
-  ;; TNH-NEXT:     )
-  ;; TNH-NEXT:     (ref.null $struct)
-  ;; TNH-NEXT:    )
-  ;; TNH-NEXT:    (ref.as_data
-  ;; TNH-NEXT:     (ref.null any)
-  ;; TNH-NEXT:    )
-  ;; TNH-NEXT:   )
```
kripken, line 66: I guess the additional type info about nulls makes it optimizable now?

Comments on line 120-121 looks stale now though.

test/lit/passes/merge-blocks.wast
```diff
@@ -71,11 +73,13 @@
...
- ;; CHECK-NEXT:  )
- ;; CHECK-NEXT:  (drop
  ;; CHECK-NEXT:   (struct.get $struct 0
- ;; CHECK-NEXT:    (ref.null $struct)
+ ;; CHECK-NEXT:    (block (result (ref null $struct))
+ ;; CHECK-NEXT:     (drop
+ ;; CHECK-NEXT:      (i32.const 1234)
+ ;; CHECK-NEXT:     )
+ ;; CHECK-NEXT:     (ref.null none)
+ ;; CHECK-NEXT:    )
```
kripken, line 82: This is no longer being optimized properly. MergeBlocks should have moved the nested drop to the outer block. Perhaps the pass is confused by the type of the null?

kripken (commented):
Some comments after perhaps half the tests.

test/lit/passes/remove-unused-module-elements-refs.wast
tlively: I'm going through and replacing a bunch of `ref.null`s with `local.get`s of new parameters to avoid introducing new `;; replaces something unreachable we can't emit` blocks.

tlively: Pushed these changes now. Will look at the specific comments tomorrow.

test/lit/passes/dae-gc-refine-params.wast
```diff
@@ -218,55 +218,43 @@
...
+ ;; CHECK-NEXT:   (call $get_null_{i32_i64})
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT:  (drop
+ ;; CHECK-NEXT:   (local.get $y)
  ;; CHECK-NEXT:  )
  ;; CHECK-NEXT: )
  ;; NOMNL:      (func $various-params-set (type $ref?|${i32}|_ref?|${i32}|_=>_none) (param $x (ref null ${i32})) (param $y (ref null ${i32}))
- ;; NOMNL-NEXT:  (local $2 (ref null ${}))
- ;; NOMNL-NEXT:  (local.set $2
+ ;; NOMNL-NEXT:  (drop
```
kripken, line 241: See comment on new line 264, it is important to see a new local appear here for this test, so I guess we need a value other than null so subtyping is a problem (that forces the local to be added).

test/lit/passes/gufa.wast
```diff
@@ -1046,7 +1046,7 @@
 
   ;; CHECK:      (func $target-keep (param $x i32)
   ;; CHECK-NEXT:  (drop
-  ;; CHECK-NEXT:   (i32.const 1)
+  ;; CHECK-NEXT:   (unreachable)
```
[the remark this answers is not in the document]
kripken: (or `local.get` as I see you've been doing)

test/lit/passes/cfp.wast
```diff
@@ -70,24 +69,22 @@
...
-  ;; CHECK-NEXT:   (block (result f32)
+  ;; CHECK-NEXT:   (block
   ;; CHECK-NEXT:    (drop
-  ;; CHECK-NEXT:     (ref.as_non_null
-  ;; CHECK-NEXT:      (ref.null $struct)
-  ;; CHECK-NEXT:     )
+  ;; CHECK-NEXT:     (ref.null none)
   ;; CHECK-NEXT:    )
-  ;; CHECK-NEXT:    (f32.const 42)
+  ;; CHECK-NEXT:    (unreachable)
```
kripken, line 87: This looks wrong.

test/lit/parse-double-unreachable.wast
```diff
@@ -0,0 +1,44 @@
...
+  (drop
+   ;; This gets emitted as an unreachable, but it doesn't have type
+   ;; unreachable, so we continue emitting instructions afterward. When
+   ;; parsing, this will put us into unreachable mode.
+   (call_ref $func
+    (ref.null nofunc)
+   )
+  )
+
+  ;; Now we are in unreachable parsing mode. Due to the bug, we would have
```
kripken, line 31: sentence cut off?

test/lit/heap-types.wast
```diff
@@ -10,21 +10,20 @@
...
   ;; CHECK:      (type $struct.A (struct (field i32)))
-  ;; NOMNL:      (type $struct.A (struct_subtype (field i32) data))
   (type $struct.A (struct i32))
   ;; NOMNL:      (type $struct.B (struct_subtype (field i32) data))
   (type $struct.B (struct i32))
   ;; CHECK:      (func $test
   ;; CHECK-NEXT:  (drop
   ;; CHECK-NEXT:   (ref.test_static $struct.A
-  ;; CHECK-NEXT:    (ref.null $struct.A)
+  ;; CHECK-NEXT:    (ref.null none)
```
kripken, line 19: The binary outputs for this test changed a lot, I think because the null now makes everything after it unreachable and not emitted. Might be better to make this read a local.

test/lit/passes/optimize-instructions-gc-tnh.wast
tlively: Yeah, the new types on nulls help. This whole second case seems stale, so I'll delete it.

test/example/typeinfo.txt
```diff
@@ -9,8 +9,8 @@ eq
 eqref
 (ref eq)
 i31
-(ref null i31)
 i31ref
+(ref i31)
```
kripken, line 13: This change seems odd... did `i31ref` turn from nullable to non-nullable?

test/example/cpp-unit.cpp
```diff
@@ -602,31 +602,6 @@ void test_effects() {
...
-    EffectAnalyzer effects(options, module);
-    effects.visit(&arrayCopy);
-    assert_equal(effects.trap, true);
-    assert_equal(effects.readsArray, true);
-    assert_equal(effects.writesArray, true);
-    assert_equal(effects.readsMutableStruct, false);
-    assert_equal(effects.writesStruct, false);
-  }
-}
-
```
kripken, line 618: why remove the above and the below?

test/example/typeinfo.txt
tlively: It switched from non-nullable to nullable, but that was back in #4843. It looks like I made a drive-by fix to the printing in this PR when I added printing support for the new bottom types: https://github.com/WebAssembly/binaryen/pull/5115/files#diff-bff8588536e048ccdcbed7cb98bd8d3b564439785901aaa60bdfcdd8d0ac355eR2120-R2121.

kripken: I see, thanks. Might be better in an earlier PR, though there isn't really much risk, so I don't feel strongly.

test/example/cpp-unit.cpp
tlively: This test didn't initialize the `ArrayCopy`'s children, so it ended up reading junk data. Rather than fix it by building a complete expression here, it seemed simpler to delete it. This is not a comprehensive test of the effects API, so deleting this one test didn't seem too bad.

Similarly, I deleted the test below because it's output changed but it didn't seem like it was pulling its weight.

kripken: Good point about junk data. But it's likely the only place we test this, so I think it's worth fixing. I can do that in a tiny initial PR if you want. (I mean, just to create some kind of valid data for children - doesn't need to pass the validator even, but just be scannable.)

The test below I do agree is not worthwhile since lots of tests print i31s.

tlively: It's ok, I can fix it.

tlively:
I believe all the feedback so far has been addressed.

src/wasm-builder.h
```diff
@@ -1262,7 +1266,7 @@ class Builder {
     if (curr->type.isTuple() && curr->type.isDefaultable()) {
       return makeConstantExpression(Literal::makeZeros(curr->type));
     }
-    if (curr->type.isNullable()) {
+    if (curr->type.isNullable() && curr->type.isNull()) {
```
kripken, line 1269: Why can't we just have the 2nd condition? (How can this have a null type, but that type is not nullable?)

kripken (approved):
Great! Code and tests lgtm. Please fuzz this carefully before landing though.

src/wasm-builder.h
tlively: For example this could be a `Call` of a function that returns `(ref none)`. At runtime that call will never finish, but it is still a valid instruction with type `(ref none)`. In that case we couldn't replace it with a `ref.null` because that would require changing the type to `(ref null none)`.

kripken: Interesting...

So `(ref none)` is kind of like an unreachable type in wasm, that can actually be used as a function's return type..? Perhaps we could use that.

tlively:
Fuzzed 44k iterations overnight with no problems.

src/wasm-builder.h
tlively: Yeah exactly. Although there are three of them (`(ref none)`, `(ref nofunc)`, and `(ref noextern)`), they don't actually put validation into unreachable mode, and the rest of the code is already set up to handle `Type::unreachable` correctly, so I think the best way to use them would be to translate them to `unreachable` or place an `(unreachable)` after any instruction that produces one.

PR #5117, opened 2022-10-05
Fix fuzzer after #5114

sbc100:
Fixes `./scripts/fuzz_opt.py --auto-initial-contents 7044408155933374954`

### DIFF
--- a/scripts/wasm2js.js
+++ b/scripts/wasm2js.js
@@ -64,9 +64,9 @@ var WebAssembly = {
     };
     var atob = decodeBase64;
     // Additional imports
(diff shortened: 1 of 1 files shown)


scripts/wasm2js.js
```diff
@@ -85,7 +85,7 @@ var WebAssembly = {
 
 var tempRet0 = 0;
 
-var asmLibraryArg = {
+var env = {
```
sbc100, line 88: I'm not sure this var used anymore .. but I don't know how to tell for sure.  I don't normally run the fuzzer.

kripken (approved):

PR #5119, opened 2022-10-06
Warn on too many parameters for Web VMs

kripken:
Similar to our existing handling of memory segment size:

https://github.com/WebAssembly/binaryen/blob/e6be381b6f5adbd07af080a4e1f74ba04c8eda82/src/wasm/wasm-binary.cpp#L600-L604

Fixes https://github.com/emscripten-core/emscripten/issues/17988

### DIFF
--- a/test/unit/test_web_limitations.py
+++ b/test/unit/test_web_limitations.py
@@ -1,0 +1,21 @@
+import os
+
+from scripts.test import shared
+from . import utils
+
+"""Test that we warn on large numbers of parameters, which Web VMs disallow."""
+
+
+class WebLimitations(utils.BinaryenTestCase):
+    def test_many_params(self):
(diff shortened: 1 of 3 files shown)


tlively (approved):

test/unit/test_web_limitations.py
```diff
@@ -0,0 +1,23 @@
...
+        params = '(param i32) ' * 1001
+        module = '''
+        (module
+         (func $foo %s
+         )
+        )
+        ''' % params
+        p = shared.run_process(shared.WASM_OPT + ['-o', os.devnull],
+                               input=module, check=False, capture_output=True)
+        self.assertEqual(p.returncode, 0)
```
sbc100, line 21: Drop the assert and use check=True instead?

PR #5121, opened 2022-10-06
GUFA: Use SSA-style information

kripken:
Previously we treated each local index as a location, and every `local.set` to
that index could be read by every `local.get`. With this we connect only
relevant sets to gets.

Practically speaking, this removes `LocalLocation` which is what was just
described, and instead there is `ParamLocation` for incoming parameter
values. And `local.get/set` use normal `ExpressionLocations` to connect a
set to a get.

I was worried this would be slow, since computing `LocalGraph` takes time,
but it actually more than makes up for itself on J2Wasm and we are faster
actually :rocket: I guess since we do less updating after `local.set`s.

This makes a noticeable change on the J2Wasm binary, and perhaps will
help with benchmarks.


test/lit/passes/gufa-refs.wast
```diff
@@ -1422,14 +1424,26 @@
...
+      (local.set $parent
+        (local.tee $child
+          (struct.new $child
+            (i32.const 20)
+            (i32.const 30)
+          )
+        )
+      )
+    )
     ;; This get cannot be optimized because later down the local is written a
```
tlively, line 1447: "later down" looks stale here.

tlively (approved):
Very nice!

PR #5124, opened 2022-10-07
[Parser] Parse folded instructions

tlively:
Parse folded expressions as described in the spec:
https://webassembly.github.io/spec/core/text/instructions.html#folded-instructions.
The old binaryen parser _only_ parses folded expressions, and furthermore
requires them to be folded such that a parent instruction consumes the values
produced by its children and only those values. The standard format is much more
general and allows folded instructions to have an arbitrary number of children
independent of dataflow.

To prevent the rest of the parser from having to know or care about the
difference between folded and unfolded instructions, parse folded instructions
after their children have been parsed. This means that a sequence of
instructions is always parsed in the order they would appear in a binary no
matter how they are folded (or not folded).

### DIFF
--- a/src/wasm/wat-parser.cpp
+++ b/src/wasm/wat-parser.cpp
@@ -81,6 +81,10 @@ struct ParseInput {
     lexer.setIndex(index);
   }
 
+  ParseInput(const ParseInput& other, size_t index) : lexer(other.lexer) {
+    lexer.setIndex(index);
+  }
+
   bool empty() { return lexer.empty(); }
 
   std::optional<Token> peek() {
@@ -106,6 +110,19 @@ struct ParseInput {
     }
     ++lexer;
     return true;
+  }
+
+  bool takeUntilParen() {
+    while (true) {
+      auto t = peek();
+      if (!t) {
+        return false;
+      }
+      if (t->isLParen() || t->isRParen()) {
+        return true;
+      }
+      ++lexer;
+    }
   }
 
   std::optional<Name> takeID() {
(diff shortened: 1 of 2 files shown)


tlively:
* main
  * **PR #5124** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5124" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5124?utm_source=stack-comment).

test/lit/wat-kitchen-sink.wast
```diff
@@ -121,6 +121,66 @@
...
+    (nop)
+    (nop)
+    (nop
+     (nop)
+    )
+   )
+   (nop)
+  )
+  (nop)
+ )
```
kripken, line 161: Rather than nops perhaps these could be drops of constants, which would make it possible to see the order is correct?

src/wasm/wat-parser.cpp
```diff
@@ -1610,10 +1627,65 @@ MaybeResult<typename Ctx::InstrT> instr(Ctx& ctx, ParseInput& in) {
...
+        // one.
+        if (in.takeLParen()) {
+          foldedInstrs.push_back({in.getPos(), {}});
+        } else if (in.takeRParen()) {
+          auto [start, end] = foldedInstrs.back();
+          assert(end && "Should have found end of instruction");
+          foldedInstrs.pop_back();
+
+          ParseInput foldedIn(in, start);
+          if (auto inst = instr(ctx, foldedIn)) {
```
kripken, line 1663: I'm probably missing something, but I'd expect to see a code path that handles non-folded children in a folded instruction around here, which should call `instrs()` on a substring of the folded instruction. Is that right? If so where is it?

What I mean is this:
```
(block ;; after parsing the shape of the block, where is the code that will parse the nops?
  nop
  nop
  nop
)
```

tlively: Blocks and other control flow structures will definitely need to be special cased in this code once we support them at all. The two locations here that check for left parens will have to be updated to check for control flow structures first. Since we don't support control flow structures yet, it's not possible to have non-folded instructions as children of folded instructions. This surprised me when I was reading the spec, but it definitely makes the parsing problem simpler.

test/lit/wat-kitchen-sink.wast
tlively: I'll have to drop my funny (probably to me only) function names, but that's ok.

tlively: Ah, I forgot that we don't support drop yet. Adding support would require new functionality to hook up children and parents, which I would rather do as a separate PR. I'll sprinkle in some `unreachable`s instead to try to achieve the same thing.

kripken (approved):

PR #5129, opened 2022-10-12
[Parser] Parse instructions with children

tlively:
Parse unary, binary, drop, and select instructions, properly fixing up stacky
code, unreachable code, and multivalue code so it can be represented in Binaryen
IR.

### DIFF
--- a/src/wasm/wat-lexer.cpp
+++ b/src/wasm/wat-lexer.cpp
@@ -943,7 +943,7 @@ void Lexer::lexToken() {
 }
 
 TextPos Lexer::position(const char* c) const {
-  assert(size_t(c - buffer.data()) < buffer.size());
+  assert(size_t(c - buffer.data()) <= buffer.size());
   TextPos pos{1, 0};
   for (const char* p = buffer.data(); p != c; ++p) {
     if (*p == '\n') {
--- a/src/wasm/wat-parser.cpp
+++ b/src/wasm/wat-parser.cpp
@@ -276,15 +276,13 @@ struct ParseInput {
     return lexer.getIndex();
   }
 
-  [[nodiscard]] Err err(std::string reason) {
+  [[nodiscard]] Err err(Index pos, std::string reason) {
     std::stringstream msg;
-    msg << lexer.position(lexer.getIndex()) << ": error: " << reason;
+    msg << lexer.position(pos) << ": error: " << reason;
     return Err{msg.str()};
   }
 
-  [[nodiscard]] Err err(Index pos, std::string reason) {
-    return ParseInput(*this, pos).err(reason);
-  }
+  [[nodiscard]] Err err(std::string reason) { return err(getPos(), reason); }
 };
 
 // =========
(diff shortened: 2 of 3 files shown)


tlively:
* main
  * **PR #5129** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5129" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5129?utm_source=stack-comment).

src/wasm/wat-lexer.cpp
```diff
@@ -943,7 +943,7 @@ void Lexer::lexToken() {
 }
 
 TextPos Lexer::position(const char* c) const {
-  assert(size_t(c - buffer.data()) < buffer.size());
+  assert(size_t(c - buffer.data()) <= buffer.size());
```
kripken, line 946: If the buffer is of size 1 then `&buffer[1] - buffer.data()` is equal to 1, so this check passes, but that's out of bounds isn't it? Or is there a padding byte..?

src/wasm/wat-parser.cpp
```diff
@@ -596,16 +709,55 @@ template<typename Ctx> struct InstrParserCtx : TypeParserCtx<Ctx> {
...
-  InstrT makeF64Const(double c) { return builder.makeConst(Literal(c)); }
+  Result<> makeUnreachable(Index pos) {
+    return push(pos, builder.makeUnreachable());
+  }
+  Result<> makeNop(Index pos) { return push(pos, builder.makeNop()); }
+  Result<> makeBinary(Index pos, BinaryOp op) {
+    auto rhs = pop(pos);
+    CHECK_ERR(rhs);
+    auto lhs = pop(pos);
+    CHECK_ERR(lhs);
```
kripken, line 720: I think we can autogenerate this using `wasm-delegations-fields`?

test/lit/wat-kitchen-sink.wast
```diff
@@ -181,6 +185,455 @@
...
+ )
+
+ ;; CHECK:      (func $add-twice-unreachable-2 (type $ret2) (result i32 i32)
+ ;; CHECK-NEXT:  (drop
+ ;; CHECK-NEXT:   (i32.add
+ ;; CHECK-NEXT:    (i32.const 1)
+ ;; CHECK-NEXT:    (i32.const 2)
+ ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT:  (tuple.make
```
kripken, line 454: Why does this make a tuple? A block would be simpler (in particular I'm not sure how the optimizer handles tuples).

src/wasm/wat-parser.cpp
tlively: For some instructions yes, but not for others. For example, `select` needs to do some extra validation for the type parameter, and I don't see a good way of integrating that with `wasm-delegations-fields`. Note that the validation can't happen entirely in the general parsing function because general `ResultsT` might not have the necessary information. Other expressions like blocks and calls with variable numbers of children won't work with `wasm-delegations-fields` because those children aren't attached to the expression yet.

test/lit/wat-kitchen-sink.wast
tlively: This is a multivalue function, so it has to return a tuple.

The optimizer (along with everything else) handles tuples very poorly!

src/wasm/wat-lexer.cpp
tlively: There's no padding byte, but `c` is never dereferenced, so that's ok.

src/wasm/wat-parser.cpp
kripken: Ah, I see, thanks...

src/wasm/wat-lexer.cpp
kripken: I see now, thanks.

test/lit/wat-kitchen-sink.wast
kripken: I missed the function signature... makes sense now.

kripken (approved):

PR #5132, opened 2022-10-12
[Parser][NFC] Move `ParseInput` into the parser context

tlively:
Rather than passing both a `Ctx` and a `ParseInput` to every parsing function,
pass only a `Ctx` with a `ParseInput` inside of it. This significantly reduces
verbosity in the parser. To handle cases where parsing needs to happen at
specific locations, which used to be handled by constructing a new `ParseInput`
independent from the ctx, introduce a new RAII utility for temporarily changing
the location of the `ParseInput` inside a context.

Also add a utility for generating an error at a particular location to avoid
having to construct new `ParseInput` objects just for that purpose. This
resolves a few TODOs about correcting error locations, but since we don't test
those yet, I still consider this NFC.

### DIFF
--- a/scripts/gen-s-parser.py
+++ b/scripts/gen-s-parser.py
@@ -721,7 +721,7 @@ def instruction_parser(new_parser=False):
     def print_leaf(expr, inst):
         if new_parser:
             expr = expr.replace("()", "(ctx)")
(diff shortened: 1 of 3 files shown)


kripken (approved):
Looks simpler!

PR #5134, opened 2022-10-12
[Wasm GC] Add a getMaxDepths() helper for heap types

kripken:
This computes how deep the children of a heap type are. This will be useful in
cone type optimizations, since we want to "normalize" cones: a cone of depth
infinity can just be a cone of the actual maximum depth of existing children, etc.,
and it's simpler to have a single canonical representation to avoid extra work.


src/ir/subtypes.h
```diff
@@ -73,6 +73,50 @@ struct SubTypes {
...
+    std::unordered_set<HeapType> work(types.begin(), types.end());
+
+    while (!work.empty()) {
+      auto iter = work.begin();
+      auto type = *iter;
+      work.erase(iter);
+      if (auto super = type.getSuperType()) {
+        auto depth = depths[type];
+        auto& superDepth = depths[*super];
+        if (depth + 1 >= superDepth) {
```
tlively, line 100: If the `superDepth` doesn't change, then we shouldn't need to keep working.
 ```suggestion
        if (depth + 1 > superDepth) {
```

File: src/ir/subtypes.h [hunk 2]
```diff
@@ -73,6 +73,50 @@ struct SubTypes {
...
+    std::unordered_map<HeapType, Index> depths;
+
+    // Begin with depth 0.
+    for (auto type : types) {
+      depths[type] = 0;
+    }
+
+    // Begin with a plan to work on all the types. When we visit an item, we'll
+    // update our super type based on our current depth.
+    std::unordered_set<HeapType> work(types.begin(), types.end());
```
tlively, line 91: This algorithm is O(n^2), but I would expect that the other O(n^2) algorithm using two nested for loops and no work list would perform better.

An O(n) algorithm would be to topologically sort the types then collect the max depths in a single pass. We have a topo sort utility here: https://github.com/WebAssembly/binaryen/blob/main/src/support/topological_sort.h. You can see how it is used here: https://github.com/WebAssembly/binaryen/blob/main/src/support/topological_sort.h.

kripken: Good idea! Done.

PR #5135, opened 2022-10-12
[Wasm GC] Add GUFA tests for struct reads from cones

kripken:
We had cone tests for `ref.eq` and `ref.cast` etc. but not for `struct.get`.

### DIFF
--- a/test/lit/passes/gufa-refs.wast
+++ b/test/lit/passes/gufa-refs.wast
@@ -4564,3 +4564,230 @@
     )
   )
 )
+
+;;      A(42)
+;;     /     \
+;; A1(100) A2(100)
+;;
+;; A disagrees on the value of the field with its subtypes.
+(module
+  ;; CHECK:      (type $A (struct_subtype (field i32) data))
+  (type $A (struct_subtype (field i32) data))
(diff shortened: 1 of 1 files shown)


kripken:
Oops, I pushed the wrong tests a few minutes ago. Should be correct now, sorry.

test/lit/passes/gufa-refs.wast
```diff
@@ -4564,3 +4564,634 @@
...
+    (struct.set $A 0
+      (select
+        (local.get $A)
+        (local.get $B)
+        (local.get $x)
+      )
+      (i32.const 10)
+    )
+    ;; Read from all the locals. We can optimize them all, to 10, 10, 20. The
+    ;; last requires more cone opts, however. TODO
```
tlively, line 4982: Alternatively, we could boost the precision here by only flowing written values out to type locations when those values escape a function.

tlively (approved):

test/lit/passes/gufa-refs.wast
kripken: Interesting. Yeah, if we tracked GC allocations by reference and not just by type, we could do a lot more. Maybe even just be escaping or not could help, as you say. Worth thinking on.

PR #5143, opened 2022-10-16
Support try-catch blocks in Asyncify.

### LINKED ISSUE #4470: Can't build web-gphoto2 with -fwasm-exceptions instead of -fexceptions
web-gphoto2 builds with pthreads, Asyncify, exceptions and LTO, and I'm not sure which combo is specifically problematic here, but the end result is that it crashes when trying to switch from `-fexceptions` to `-fwasm-exceptions`:

```
unexpected expression type
UNREACHABLE executed at /b/s/w/ir/cache/builder/emscripten-releases/binaryen/src/passes/Asyncify.cpp:998!
emcc: error: '/emsdk/upstream/bin/wasm-opt --strip-dwarf --post-emscripten -Os --low-memory-unused --asyncify --pass-arg=asyncify-imports@env.invoke_*,env.__call_main,env.emscripten_sleep,env.emscripten_wget,env.emscripten_wget_data,env.emscripten_idb_load,env.emscripten_idb_store,env.emscripten_idb_delete,env.emscripten_idb_exists,env.emscripten_idb_load_blob,env.emscripten_idb_store_blob,env.SDL_Delay,env.emscripten_scan_registers,env.emscripten_lazy_load_code,env.emscripten_fiber_swap,wasi_snapshot_preview1.fd_sync,env.__wasi_fd_sync,env._emval_await,env._dlopen_js,env.__asyncjs__* --zero-filled-memory --strip-debug --strip-producers ui/libapi.wasm -o ui/libapi.wasm --mvp-features --enable-threads --enable-mutable-globals --enable-bulk-memory --enable-sign-ext --enable-exception-handling' failed (received SIGABRT (-6))
```

A branch that's failing can be checked out from here: https://github.com/GoogleChromeLabs/web-gphoto2/tree/fwasm-exceptions. It should be pretty reproducible as it uses Docker, so all you need to do is run `./build.sh` in the checkout folder.

jashug:
Previously, when the Asyncify pass ran on a wasm file that includes try blocks (in a position such that asyncify has to un/re-wind through those try blocks), an error would be produced:
`WASM_UNREACHABLE("unexpected expression type");`

This PR adds support for the try case in Asyncify.

Fixes #4470

I was not sure how to add tests for this feature to the repository; it runs without crashing and produces reasonable-looking output on one sample file I have. This feature deserves tests: if someone will explain how to add them I will.

This should have no effect on previously successful applications of Asyncify (the behavior change here is from a crash to producing code), but if the implementation is wrong could produce broken code.

Convert
```
(try
  (...)
 catch $i
  (...))
```
into
```
(try
  (...)
 catch $i
  (set tag 1))
(if (or (eq tag 1) (rewinding))
 (...))
```

### DIFF
--- a/src/ir/eh-utils.cpp
+++ b/src/ir/eh-utils.cpp
@@ -32,7 +32,7 @@ namespace EHUtils {
 // no way to get the given expression's address. But that's fine because pop's
 // pointer is only necessary (in handleBlockNestedPops) to fix it up when it is
 // nested, and if 'catchBody' itself is a pop, we don't need to fix it up.
-static Expression*
+Expression*
 getFirstPop(Expression* catchBody, bool& isPopNested, Expression**& popPtr) {
   Expression* firstChild = catchBody;
   isPopNested = false;
--- a/src/ir/eh-utils.h
+++ b/src/ir/eh-utils.h
@@ -24,6 +24,18 @@ namespace wasm {
 
 namespace EHUtils {
 
+// This returns three values, some of them as output parameters:
+// - Return value: 'pop' expression (Expression*), when there is one in
+//   first-descendant line. If there's no such pop, it returns null.
+// - isPopNested: Whether the discovered 'pop' is nested within a block
+// - popPtr: 'pop' expression's pointer (Expression**), when there is one found
+//
+// When 'catchBody' itself is a 'pop', 'pop''s pointer is null, because there is
+// no way to get the given expression's address. But that's fine because pop's
+// pointer is only necessary (in handleBlockNestedPops) to fix it up when it is
+// nested, and if 'catchBody' itself is a pop, we don't need to fix it up.
+Expression* getFirstPop(Expression* catchBody, bool& isPopNested, Expression**& popPtr);
+
 // Returns true if a 'pop' instruction exists in a valid location, which means
 // right after a 'catch' instruction in binary writing order.
 // - This assumes there should be at least a single pop. So given a catch body
--- a/src/passes/Asyncify.cpp
+++ b/src/passes/Asyncify.cpp
@@ -1087,8 +1087,9 @@ private:
           for (size_t i = catchBodies.size(); i > 0; i--) {
             // TODO: Can optimize if !analyzer->canChangeState(child, func)
             auto* child = catchBodies[i - 1];
+            Type type = Type::none;
             if (i - 1 < catchTags.size()) {
-              auto type = module->getTag(catchTags[i - 1])->sig.params;
+              type = module->getTag(catchTags[i - 1])->sig.params;
               if (type != Type::none) {
                 bool isPopNested = false;
                 Expression** popPtr = nullptr;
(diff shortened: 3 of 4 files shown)


kripken:
Thanks for the PR @jashug !

I think this almost works, but I believe in a recent discussion with @aheejin and @tlively it came up that it isn't quite enough, since moving code out of the catch block means that rethrow would work differently.

jashug:
Good point. I think converting all the `rethrow` instructions targeting a converted catch block into `throw` instructions would fix this? Extra data on the exception like a stack trace will be lost, but I don't see any way of preserving that in Asyncify.

tlively:
Every throw would have to throw the original value and also a full set of Asyncify unwind data describing how to get back to the throw (including the data for any previous throws). This data would have to be collected eagerly, since there would be no way to go back and collect it at the suspension point. I think it would be simpler to lower away EH entirely before applying the Asyncify transformation.

jashug:
For `catch_all`, the tricky case is when an exception is thrown without a wasm tag to be caught, such as an exception thrown from an imported Javascript function. In that case, Javascript has to be the one to save the exception, since it might not be representable/savable in wasm. If all the exceptions come from wasm, they have exception tags to be caught, and we could convert `catch_all` into `forall i. catch i`.

If we could save `catch_all` exceptions on a Javascript stack data structure, we could "just" pass around the pair of Asyncify unwind data and that Javascript stack, and avoid having to get back to the throw. "Exception reference" types look like they might solve this problem, but I'm having trouble finding their specification.

tlively:
Yeah, exception references would have made this simpler, but they were removed from the final EH spec and do not exist.

jashug:
I pushed a sketch dealing with rethrows in catch but not catch_all blocks, but I don't see a way to handle the full exception handling WebAssembly feature, so I'm closing this PR.

kripken:
Makes sense. This is very hard to do in Asyncify. I guess the solutions are either

1. Wait for wasm stack switching, which can replace Asyncify. That's testable now but will take time to be standardized.
2. As @tlively said, `I think it would be simpler to lower away EH entirely before applying the Asyncify transformation` - we could write an EH lowering pass to wasm MVP.

aheejin:
How do you lower away EH? Do you mean removing landingpads like `-fignore-exceptions` so if there's an exception we just crash, or do you mean something else?

tlively:
No, I mean a binaryen pass to lower away EH but preserve semantics as much as possible, probably by instrumenting functions to return an extra value signaling an exception and instrumenting all function calls to check that value. Exception payloads could go on a stack in a second memory or could be stored in GC data. This lowering could not possibly do the right thing for exceptions entering the module from the outside, but that might be ok for some use cases where the only exceptions are expected to come from the Wasm module itself.

aheejin:
You think we can handle `rethrow` too?

tlively:
Yeah, I think so. If we take all the engine-internal state from the EH proposal and put it in a secondary memory or in GC objects, then we should have everything we need to implement `rethrow` correctly as well.

aheejin:
How can we access the engine-internal state from user code?

tlively:
If we lower away EH in a Binaryen pass, there won't be any EH-internal state at runtime. It will all be stored in data locations introduced by the Binaryen pass instead.

aheejin:
To lower away EH in Binaryen, we need to access the engine-internal state to make `rethrow` work. I think it's what you said, no?

kripken:
Actually we should have an issue open for this, not a PR. Closing this, will open an issue.

PR #5149, opened 2022-10-18
[Wasm GC] Filter GUFA expression locations by their type

kripken:
Now that we have a cone type, we are able to represent in PossibleContents the
natural content of a wasm location: a type or any of its subtypes. This allows us to
enforce the wasm typing rules, that is, to filter the data arriving at a location by the
wasm type of the location.

Technically this could be unnecessary if we had full implementations of flowFoo
and so forth, that is, tailored code for each wasm expression that makes sure we
only contain and flow content that fits in the wasm type. Atm we don't have that,
and until the wasm spec stabilizes it's probably not worth the effort. Instead,
simply filter based on the type, which gives the same result (though it does take
a little more work; I measured it at 3% or so of runtime).

While doing so normalize cones to their actual maximum depth, which simplifies
things and will help more later as well.

### DIFF
--- a/src/ir/possible-contents.cpp
+++ b/src/ir/possible-contents.cpp
@@ -1872,19 +1872,21 @@ void Flower::filterExpressionContents(PossibleContents& contents,
   //
   // Note that the intersection may also not be a cone type, if it is a global
   // or literal. In that case we don't have anything more to do here.
-  if (contents.isConeType()) {
-    normalizeConeType(contents);
+  if (!contents.isConeType()) {
+    return;
+  }
 
-    // There is a chance that the intersection is equal to the maximal contents,
-    // which would mean nothing more can arrive here. (Note that we can't
-    // normalize |maximalContents| before the intersection as
-    // intersectWithFullCone assumes a full/infinite cone.)
-    normalizeConeType(maximalContents);
+  normalizeConeType(contents);
 
-    if (contents == maximalContents) {
-      // We already contain everything possible, so this is the worst case.
-      worthSendingMore = false;
-    }
+  // There is a chance that the intersection is equal to the maximal contents,
+  // which would mean nothing more can arrive here. (Note that we can't
+  // normalize |maximalContents| before the intersection as
+  // intersectWithFullCone assumes a full/infinite cone.)
+  normalizeConeType(maximalContents);
+
+  if (contents == maximalContents) {
+    // We already contain everything possible, so this is the worst case.
+    worthSendingMore = false;
   }
 }
 
--- a/src/ir/possible-contents.cpp
+++ b/src/ir/possible-contents.cpp
@@ -1696,9 +1696,7 @@ bool Flower::updateContents(LocationIndex locationIndex,
     filtered = true;
   }
 
-  // If we think more might arrive, and we filtered, the filtering might have
-  // changed that, so check.
-  if (worthSendingMore && filtered && contents == oldContents &&
+  if (filtered && contents == oldContents &&
       contents.getType() == oldContents.getType()) {
     // Nothing actually changed after filtering, so just return.
     return worthSendingMore;
--- a/src/ir/possible-contents.cpp
+++ b/src/ir/possible-contents.cpp
@@ -1840,8 +1840,9 @@ void Flower::filterExpressionContents(PossibleContents& contents,
   if (!type.isRef()) {
     return;
   }
-  // The maximal contents here are the declared type and all subtypes.
-  // Nothing else can pass through, so filter such things out.
+
+  // The maximal contents here are the declared type and all subtypes. Nothing
+  // else can pass through, so filter such things out.
   auto maximalContents = PossibleContents::fullConeType(type);
   contents.intersectWithFullCone(maximalContents);
   if (contents.isNone()) {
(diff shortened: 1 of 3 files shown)


src/ir/possible-contents.cpp
```diff
@@ -1359,8 +1359,13 @@ struct Flower {
...
 
-  // Contents sent to a global location can be filtered in a special way during
-  // the flow, which is handled in this helper.
+  // Contents sent to certain locations can be filtered in a special way during
+  // the flow, which is handled in these helpers. These may update
+  // |worthSendingMore| which is whether it is worth sending any more content to
+  // this location in the future.
+  void filterExpressionContents(PossibleContents& contents,
+                                const ExpressionLocation& exprLoc,
+                                bool& worthSendingMore);
```
tlively, line 1368: Why not return the bool?

src/ir/possible-contents.cpp
```diff
@@ -1384,6 +1389,34 @@ struct Flower {
...
+  // subtypes, 1 if it has subtypes but none of those have subtypes themselves,
+  // and so forth.
+  std::unordered_map<HeapType, Index> maxDepths;
+
+  // Given a ConeType, normalize it, that is, make its depth the canonical
+  // depth given the actual children it has. If this is a full cone, then we can
+  // always pick the actual maximal depth and use that instead of FullDepth==-1.
+  // For a non-full cone, we also reduce the depth as much as possible, so it is
+  // equal to the maximum depth of an existing subtype.
+  void normalizeConeType(Type type, Index& depth) {
```
tlively, line 1402: It seems simpler to return the normalized depth.

src/ir/possible-contents.cpp
```diff
@@ -1639,16 +1682,30 @@ bool Flower::updateContents(LocationIndex locationIndex,
...
-    if (contents == oldContents &&
-        contents.getType() == oldContents.getType()) {
-      // Nothing actually changed after filtering, so just return.
-      return worthSendingMore;
-    }
+    filtered = true;
   }
 
+  if (filtered && contents == oldContents &&
+      contents.getType() == oldContents.getType()) {
```
tlively, line 1700: Why do we need to compare the types if we've already compared the contents?

src/ir/possible-contents.cpp
```diff
@@ -1776,6 +1833,63 @@ void Flower::connectDuringFlow(Location from, Location to) {
...
+  // Consider the case where |maximalContents| is Cone(B, Infinity) and the
+  // original |contents| was Cone(A, 2) (which is normalized). The naive
+  // intersection is Cone(B, 1), since the core intersection logic makes no
+  // assumptions about the rest of the types. That is then normalized to
+  // Cone(B, 0) since there happens to be no subtypes for B.
+  //
+  // Note that the intersection may also not be a cone type, if it is a global
+  // or literal. In that case we don't have anything more to do here.
+  if (!contents.isConeType()) {
+    return;
```
tlively, line 1876: Is it `worthSendingMore` in this case?

test/lit/passes/gufa-refs.wast
```diff
@@ -5324,4 +5326,80 @@
...
+  (func $filtering
+    ;; Check for filtering of values by the declared type in the wasm. We do not
+    ;; have specific filtering or flowing for br_on_* yet, so it will always
+    ;; send the value to the branch target. But the target has a declared type
+    ;; of $B, which means the exact $A gets filtered out, and nothing remains,
+    ;; so we can append an unreachable.
+    ;;
+    ;; When we add filtering/flowing for br_on_* this test should continue to
+    ;; pass and only the comment will need to be updated, so if you are reading
+    ;; this and it is stale, please fix that :)
```
tlively, line 5375: This is an excellent friendly comment for our future selves 👍

tlively (commented):
Mostly LGTM, but I have a few questions.

src/ir/possible-contents.cpp
kripken: Keeping it as an out param keeps the name. And the return value would not be related to the function's name.

src/ir/possible-contents.cpp
kripken: This is copied from before. Hmm, maybe we don't need it now? I think when we had different nulls that compare equal this was an issue, but that's gone now, correct? If so I can fix that here (maybe as a followup).

src/ir/possible-contents.cpp
kripken: It might be (e.g. if the intersection is a literal, another literal might arrive later). Now that you mention it I think it always has to be worth sending more in these cases. I added an assert for that here, and a comment.

src/ir/possible-contents.cpp
kripken: Good idea, done. Renamed to `getNormalizedConeDepth`.

src/ir/possible-contents.cpp
tlively: Ah, that explanation about nulls needing this in the past makes sense. Removing this here or in a follow-up sgtm.

src/ir/possible-contents.cpp
tlively: Oh I see, `worthSendingMore` is some global-ish state that all these helper functions essentially close over. I would probably still prefer a purer approach, but this makes sense.

tlively (approved):

tlively:
Thanks for the explanations!

PR #5153, opened 2022-10-18
[C API] Align I31ref and Dataref to be nullable

dcodeIO:
The C API still returned non nullable types for `dataref` (`ref data` instead of `ref null data`) and `i31ref` (`ref i31` instead of `ref null i31`). This PR aligns with the current state of the GC proposal, making them nullable when obtained via the C API.


tlively (approved):
👍

PR #5154, opened 2022-10-18
[Wasm GC] [NFC] Remove .type checks from GUFA that are not needed with modern nulls

kripken:
Modern nulls never compare equal unless they have the same type too.


tlively (approved):
LGTM, but interested to see what the CI failure is.

kripken:
Looks like a random python flake... but let's wait and see if the others fail too.

tlively:
Oh yeah, I guess I should have clicked it to check. Curious that that happens.

kripken:
Yep, was a flake.

PR #5155, opened 2022-10-18
[Parser] Parse global.get and global.set

tlively:
Also add some missing error checking for the similar local instructions and make
some neighboring styling more consistent.


kripken (approved):
lgtm % comment

PR #5156, opened 2022-10-18
[Parser] Parse SIMD lane manipulation instructions

tlively:
Including all `SIMDExtract`, `SIMDReplace`, `SIMDShuffle` expressions.

### DIFF
--- a/test/lit/wat-kitchen-sink.wast
+++ b/test/lit/wat-kitchen-sink.wast
@@ -25,6 +25,12 @@
 
   ;; CHECK:      (type $i32_i32_i32_=>_none (func_subtype (param i32 i32 i32) func))
 
+  ;; CHECK:      (type $v128_=>_i32 (func_subtype (param v128) (result i32) func))
+
+  ;; CHECK:      (type $v128_i32_=>_v128 (func_subtype (param v128 i32) (result v128) func))
+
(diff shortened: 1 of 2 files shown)


tlively:
* main
  * **PR #5155** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5155" target="_blank">Graphite</a> 
    * **PR #5156** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5156" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5156?utm_source=stack-comment).

test/lit/wat-kitchen-sink.wast
```diff
@@ -25,6 +25,12 @@
```
kripken, line 25: Could we add a text roundtrip test in this file perhaps? (to also test the nested versions in the new parser)

kripken (approved):

test/lit/wat-kitchen-sink.wast
tlively: The new parser does not support the non-standard `func_subtype`, `struct_subtype`, and `array_subtype` type definition format that we emit by default, so it can't currently round trip the file. My plan was to make our printer emit the standard format just before switching the default parser over, but it's probably worth doing that sooner rather than later.

PR #5161, opened 2022-10-19
[NFC] Add and use hashing for 3-tuples

kripken:
See https://github.com/WebAssembly/binaryen/pull/5157#discussion_r999657514

Apparently there is a clever recursive approach, sgtm to do that instead. (It seems like recursion would need to slice into a tuple somehow..?)


kripken:
Replaced by #5162

PR #5162, opened 2022-10-19
[NFC] Add a generic hash implementation for tuples

tlively:
We already provided a specialization of `std::hash` for arbitrary pairs, so add
one for `std::tuple` as well. Use the new specialization where we were
previously using nested pairs just to be able to use the pair specialization.

### DIFF
--- a/src/support/hash.h
+++ b/src/support/hash.h
@@ -61,6 +61,24 @@ template<typename T1, typename T2> struct hash<pair<T1, T2>> {
   }
 };
 
+// And hashing tuples is useful, too.
+template<typename T, typename... Ts> struct hash<tuple<T, Ts...>> {
+private:
+  template<size_t... I>
+  static void rehash(const tuple<T, Ts...>& tup,
+                     size_t& digest,
+                     std::index_sequence<I...>) {
+    (wasm::rehash(digest, std::get<1 + I>(tup)), ...);
+  }
+
+public:
+  size_t operator()(const tuple<T, Ts...>& tup) const {
+    auto digest = wasm::hash(std::get<0>(tup));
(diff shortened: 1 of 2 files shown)


tlively:
* main
  * **PR #5162** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5162" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5162?utm_source=stack-comment).

src/support/hash.h
```diff
@@ -61,6 +61,24 @@ template<typename T1, typename T2> struct hash<pair<T1, T2>> {
...
 };
 
+// And hashing tuples is useful, too.
+template<typename T, typename... Ts> struct hash<tuple<T, Ts...>> {
+private:
+  template<size_t... I>
+  static void rehash(const tuple<T, Ts...>& tup,
+                     size_t& digest,
+                     std::index_sequence<I...>) {
+    (wasm::rehash(digest, std::get<1 + I>(tup)), ...);
```
kripken, line 71: Interesting, how does this work? I see this `(CODE, ...)` pattern used in https://en.cppreference.com/w/cpp/utility/integer_sequence but I don't see an explanation for it.

tlively: This is a [fold expression](https://en.cppreference.com/w/cpp/language/fold) that folds the comma operator over the parameter pack, so it just ends up calling `wasm::rehash` with every value in `I`.

kripken: TIL, thanks!

kripken (approved):

PR #5163, opened 2022-10-19
SimplifyGlobals: Optimize math in init expressions

kripken:
E.g.
```wat
(global $foo i32
  (i32.add
    (i32.const 9)
    (i32.const 1)
  )
)

=>

(global $foo i32
  (i32.const 10)
)
```

For this to help most we also need to run `ConstantGlobalApplier` on global code, which
this PR adds. That is, we want to apply known global values before optimizing math
on constants (in the example above, `9` could have been generated from a `global.get` of
some other global).

With this we can optimize patterns that can be generated from things like https://github.com/WebAssembly/binaryen/pull/5107
It may also help with dynamic linking perhaps? cc @sbc100 

Two trivial drive-by fixes:
* Missing `nullptr` init of elem offsets. This may not be needed if all callsites do it, but it's safer and more consistent.
* Walk data offsets in walkModule code. I guess we've only used this for finding funcrefs in the past.


sbc100:
I don't think the linker will ever generate such expressions (IIUC it would only generate addition of imported global + cost).  But its not impossible to imagine the situation arising after other transforms.

kripken:
Hmm, the fuzzer found an interesting problem here:
```wat
(memory 1 1)
(data (i32.add
  (i32.const -1)
  (i32.const 0)
) ..)
```
That will trap at runtime, as offset -1 is not valid. But if we optimize this, we'll get a simple const offset of -1, which is actually a validation error - the wasm spec has additional validation rules for such things.

Replacing a trap with a validation error might be ok, but it does confuse the fuzzer at least. Not sure what to do about this offhand...

sbc100:
> Hmm, the fuzzer found an interesting problem here:
> 
> ```webassembly
> (memory 1 1)
> (data (i32.add
>   (i32.const -1)
>   (i32.const 0)
> ) ..)
> ```
> 
> That will trap at runtime, as offset -1 is not valid. But if we optimize this, we'll get a simple const offset of -1, which is actually a validation error - the wasm spec has additional validation rules for such things.
> 
> Replacing a trap with a validation error might be ok, but it does confuse the fuzzer at least. Not sure what to do about this offhand...

I guess it might be considered a feature to promote a guaranteed trap on instantiate to a built time validation error?

kripken:
Yeah, it might be a good thing. It would require some work in the fuzzer though, as it doesn't expect validation to change - any new validation failure is a bug in its mind atm.

tlively:
That being a validation error sounds like an engine bug. Anything that involves interpreting instructions (even constants) to determine validity should produce traps, not validation errors. IIRC, bulk memory similarly made invalid segment offsets produce traps instead of validation errors.

tlively:
Yep, not a validation error: https://webassembly.github.io/spec/core/valid/modules.html#data-segments

kripken:
@tlively Thanks! I'll experiment with what happens when those validation checks are removed. The spec tests at least pass locally.

kripken:
Fuzzer found that V8 complains about this:
```wat
 (global $global$0 (mut i32) (i32.const -2))
 (global $global$1 (mut i32) (global.get $global$0))
```
```
CompileError: WebAssembly.Module(): mutable globals cannot be used in constant expressions @+42
```
That restriction seems very odd, but it does seem to be [in the spec](https://webassembly.github.io/spec/core/valid/instructions.html#valid-constant).

Does anyone know why?

kripken:
Globals cannot be set during global evaluation - you can't put `global.set` there - so they are all effectively immutable during that phase, aren't they? I guess the only theoretical issue is an imported mutable global, but it would have to be modified on another thread, which again isn't an issue.

(Maybe I missed it, but I didn't see a requirement for globals in initializers to be imported?)

tlively:
Check out the note on constant expression validation: https://webassembly.github.io/spec/core/valid/instructions.html#valid-const

kripken:
Thanks... yikes, that seems arbitrary and kind of annoying. This pass will need to be careful to not do "partial" optimization then - only improve when the final result actually validates. Not sure offhand how best to do that.

tlively:
We could also get this fixed in the spec. There was a [recent issue](https://github.com/WebAssembly/spec/issues/1522) about this where different implementations had different behavior and the resolution was somewhat arbitrary. Relaxing this might not need to go through the full proposal process.

PR #5164, opened 2022-10-20
[Parser] Parse memories

tlively:
Parse 32-bit and 64-bit memories, including their initial and max sizes. Shared
memories are left to a follow-up PR. The memory abbreviation that includes
inline data is parsed, but the associated data segment is not yet created. Also
do some minor simplifications in neighboring helper functions for other kinds of
module elements.

### DIFF
--- a/src/wasm/wat-parser.cpp
+++ b/src/wasm/wat-parser.cpp
@@ -324,6 +324,16 @@ struct ImportNames {
   Name nm;
 };
 
+struct Limits {
+  uint64_t initial;
+  uint64_t max;
+};
+
+struct MemType {
+  Type type;
+  Limits limits;
+};
+
 // RAII utility for temporarily changing the parsing position of a parsing
 // context.
 template<typename Ctx> struct WithPosition {
@@ -342,8 +352,7 @@ template<typename Ctx> WithPosition(Ctx& ctx, Index) -> WithPosition<Ctx>;
 
 using IndexMap = std::unordered_map<Name, Index>;
 
-void applyImportNames(Importable& item,
-                      const std::optional<ImportNames>& names) {
+void applyImportNames(Importable& item, ImportNames* names) {
   if (names) {
     item.module = names->mod;
     item.base = names->nm;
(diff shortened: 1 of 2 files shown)


tlively:
* main
  * **PR #5164** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5164" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5164?utm_source=stack-comment).

src/wasm/wat-parser.cpp
```diff
@@ -562,6 +586,22 @@ template<typename Ctx> struct TypeParserCtx {
...
+  Limits makeLimits(uint64_t n, std::optional<uint64_t> m) {
+    return m ? Limits{n, *m} : Limits{n, Memory::kUnlimitedSize};
+  }
+  Limits getLimitsFromData(const std::vector<char>& data) {
+    uint64_t size = (data.size() + Memory::kPageSize - 1) / Memory::kPageSize;
+    return {size, size};
+  }
+
+  MemType makeMemType32(Limits limits) { return {Type::i32, limits}; }
+  MemType makeMemType64(Limits limits) { return {Type::i64, limits}; }
```
kripken, line 604: Perhaps a single function that takes the type as a parameter? I don't feel strongly though.

test/lit/wat-kitchen-sink.wast
```diff
@@ -103,13 +105,31 @@
...
+ (memory 0)
+ ;; CHECK:      (memory $0 0)
+
+ ;; CHECK:      (memory $mem-i32 0 1)
+ (memory $mem-i32 i32 0 1)
+ ;; CHECK:      (memory $mem-i64 i64 2)
+ (memory $mem-i64 i64 2)
+ (memory (export "mem") (export "mem2") (import "" "mem") 0)
+ ;; CHECK:      (memory $mem-init 1 1)
+ (memory $mem-init (data "hello, world!"))
```
kripken, line 120: The data segment not being added in the output is what the description mentioned as a TODO, I guess?

kripken (approved):

test/lit/wat-kitchen-sink.wast
tlively: Yeah, exactly. There's a TODO in the code for this as well.

src/wasm/wat-parser.cpp
tlively: I don't want the core parser templates to have to create `Type`s (since they should ideally be reusable outside Binaryen), so the only alternative would be to use a `bool` or a new enum. But these functions are so simple that just having two functions seems simpler and more readable than either of those options.

PR #5167, opened 2022-10-20
Remove excessive validation that is not in the wasm spec

kripken:
Specifically if a segment offset was a const, we checked that it made sense. But the
wasm spec doesn't do that, and it actually causes some issues (https://github.com/WebAssembly/binaryen/pull/5163).

In theory this extra validation might be useful - compile-time error rather than runtime -
but if we want this it should probably be an optional thing, like an opt-in flag or a `--lint`
pass or such. I'm not sure it's worth it.

### DIFF
--- a/src/wasm/wasm-validator.cpp
+++ b/src/wasm/wasm-validator.cpp
@@ -2909,27 +2909,7 @@ static bool checkSegmentOffset(Expression* curr,
                                Address add,
                                Address max,
                                FeatureSet features) {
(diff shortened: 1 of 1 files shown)


tlively (approved):
LGTM. Alternatively, we could perform these checks only if bulk memory is not enabled to match engines that don't support the bulk memory proposal, but I don't think that's necessary. (We would also have to go back and double check what the pre-bulk-memory behavior was.)

PR #5170, opened 2022-10-20
Add a way to run a pass on module-level code

kripken:
This is basically similar to `runOnFunction` which runs on a function.

This will be useful in #5163

### DIFF
--- a/src/pass.h
+++ b/src/pass.h
@@ -304,11 +304,14 @@ struct PassRunner {
   // afterwards.
   void addDefaultGlobalOptimizationPostPasses();
 
-  // Run the passes on the module
+  // Run the passes on the module.
   void run();
 
-  // Run the passes on a specific function
+  // Run the passes on a specific function.
   void runOnFunction(Function* func);
+
+  // Run the passes on module-level code.
+  void runOnModuleCode(Module* module);
 
   // When running a pass runner within another pass runner, this
   // flag should be set. This influences how pass debugging works,
@@ -346,6 +349,7 @@ private:
 
   void runPass(Pass* pass);
   void runPassOnFunction(Pass* pass, Function* func);
+  void runPassOnModuleCode(Pass* pass, Module* module);
 
   // After running a pass, handle any changes due to
   // how the pass is defined, such as clearing away any
@@ -374,6 +378,11 @@ public:
   // Implement this with code to run the pass on a single function, for
   // a function-parallel pass
   virtual void runOnFunction(Module* module, Function* function) {
+    WASM_UNREACHABLE("unimplemented");
+  }
+
+  // Implement this with code to run the pass on module-level code.
+  virtual void runOnModuleCode(Module* module) {
     WASM_UNREACHABLE("unimplemented");
   }
 
@@ -483,9 +492,15 @@ public:
     runOnFunction(module, func);
   }
 
+  void runOnModuleCode(Module* module) override {
+    assert(getPassRunner());
+    WalkerType::walkModuleCode(module);
+  }
+
+  // Utility for ad-hoc running.
   void runOnModuleCode(PassRunner* runner, Module* module) {
     setPassRunner(runner);
-    WalkerType::walkModuleCode(module);
+    runOnModuleCode(module);
   }
 };
 
--- a/src/passes/pass.cpp
+++ b/src/passes/pass.cpp
@@ -785,6 +785,15 @@ void PassRunner::runOnFunction(Function* func) {
   }
 }
 
+void PassRunner::runOnModuleCode(Module* module) {
+  if (options.debug) {
+    std::cerr << "[PassRunner] running passes on module code\n";
+  }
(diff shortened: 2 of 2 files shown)


tlively (commented):
Instead of providing `{assRunner::runOnModuleCode`, would it make sense to automatically call `Pass::runOnModuleCode` for all parallel passes from `PassRunner::run`? That way passes could be parallel and still get a chance to run on module code. I don't see a good use case for running a pass on only the module-level code and not the function code as well.

kripken:
I considered a few options here. Maybe there is something better. But:

1. Running all parallel passes on module code has some downsides. First, it is more work for no benefit atm. Second, it is odd to see a function-parallel pass modify global code. We could perhaps rethink the name "function-parallel", but I'm not sure it's worthwhile. Third, it may actually be incorrect, as what is valid in module-level code is pretty limited (even with extended-const).
2. The use case for running only on module-level code is in #5163 : We are actually only working on module-level code there, and want to improve it. We know it may likely be improved thanks to the work we've already done, which sets up such optimizations on precisely that code. Note also that that pass may end up doing multiple internal cycles of work, so we really don't want to be running all of Precompute on all the function bodies each time.

tlively:
Ok, I can buy that there's not much use for running general passes on module code right now. Can we therefore keep `Pass::runOnModuleCode` without any `PassRunner` integration at all? This intermediate state where the `PassRunner` has the ability to run a pipeline on module code, but that's not built into `run` seems weird to me.

kripken:
> Can we therefore keep Pass::runOnModuleCode without any PassRunner integration at all? 

Not sure I follow. Do you mean to remove the additions to the PassRunner class?

If so, how would this code work?

https://github.com/WebAssembly/binaryen/pull/5163/files#diff-7ffdb96cfe3f5e24b626eea4118f2a6b11c6723f7aee53d77b75ef6641c8069cR657-R659

In that PR we want to run some passes on module-level code. I think we need a PassRunner for that - I can't think of good alternative that doesn't involve hacks, at least. (The issue is that we want to run a generic pass here - not an instance of a Walker that we wrote in the same file, in which case we could do a manual call.)

> This intermediate state where the PassRunner has the ability to run a pipeline on module code, but that's not built into run seems weird to me.

It is how function-parallel passes always work, though. We run the function stuff automatically, and in the rare cases where we also want to scan module code we do that manually.

Perhaps this code could assert that the pass is function-parallel, to avoid the risk of used elsewhere where that might be odd?

tlively:
I guess that code could do `PassRegistry::get()->createPass("precompute")->runOnModuleCode(wasm)`, but I admit it's strange to access the pass registry directly like that.

Thinking about this more, this seems similar to `PassRunner::runOnFunction` in that it uses the pass runner to run on a more limited amount of code. I guess the difference that makes me unhappy is that `PassRunner::runOnFunction` does a subset of what `PassRunner::run` would do, whereas `PassRunner::runOnModuleCode` is not included in `PassRunner::run`.

Can we add a TODO comment about integrating `PassRunner::runOnModuleCode` into `PassRunner::run` if it ever becomes more widely useful?

kripken:
> `PassRegistry::get()->createPass("precompute")->runOnModuleCode(wasm)`

I don't think that can work as-is because before this PR only WalkerPasses have that method, but not generic passes. And we don't have a dynamic cast to a WalkerPass. To make this work I think we'd need to add a header for the Precompute pass basically, which seems awkward.

> Can we add a TODO comment about integrating PassRunner::runOnModuleCode into PassRunner::run if it ever becomes more widely useful?

Maybe we'd want an option for that, but I don't see how we'd integrate it as the normal workflow. For a function-parallel pass, the norm is to run only on function code, with non-normal workflows running it manually on module code. An example of that pattern:

https://github.com/WebAssembly/binaryen/blob/7144c922f1cd19a4c55c603c7f57f224e9ec4975/src/passes/ConstantFieldPropagation.cpp#L197-L200

This PR is consistent with that, I think. I'm not opposed to changing this if we have better ideas, but it would mean changing the existing model for all those places too.

I do agree though as I said that this is specific to function-parallel passes, and maybe that should be explicit in the docs + an assertion?

kripken:
> I don't think that can work as-is

To clarify what I meant, I was wrong as `runOnModuleCode(module*)` is virtual. So that can work. But it doesn't handle setting the pass runner, for which we have `runOnModuleCode(runner*, module*)` which is only in `WalkerPass`. So we'd need to either set it manually ourselves, or make that one virtual as well in the Pass class. That would break symmetry with `runOnFunction`, so it seems somewhat awkward.

Another option might be as discussed offline to run `runOnModuleCode` automatically for function-parallel passes, and make it do nothing by default. Thinking on that, it is kind of odd that the override would need to be calling `walkModuleCode` - the obvious thing to hook it up to, making the default seem confusing I think.

After more thought I still don't have a great idea here. I pushed a commit with some assertions and comments that might help clarify the current state of the PR. But I do agree it would be nice to do better somehow.

PR #5171, opened 2022-10-20
[NFC] Avoid re-parsing instruction names

tlively:
Since gen-s-parser.py is essentially a giant table mapping instruction names to
the information necessary to construct the corresponding IR nodes, there should
be no need to further parse instruction names after the code generated by
gen-s-parser.py runs. However, memory instruction parsing still parsed
instruction names to get information such as size and default alignment. The new
parser does not have the ability to parse that information out of instruction
names, so put it in the gen-s-parser.py table instead.


tlively:
* main
  * **PR #5171** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5171" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5171?utm_source=stack-comment).

kripken (approved):

PR #5172, opened 2022-10-20
[WasmGC] DCE impossible types

kripken:
`(ref nofunc)` is an "unreachable"/impossible type since the heap type is a null
but the type is non-nullable. We can emit an unreachable after such things as they
are definitely not reached.

We could use this in the return types of functions to indicate they are "noreturn",
in the future, even in MVP wasm (but we'd need to not emit the GC type in the
binary).


tlively:
> We could use this in the return types of functions to indicate they are "noreturn", in the future, even in MVP wasm (but we'd need to not emit the GC type in the binary).

I do think `Type::unreachable` would be a better choice here. It's the true bottom type in the full hierarchy, so there's not multiple versions of it to choose between or handle, and the entire codebase already uses handles `unreachable` as the canonical type that is impossible to produce. The problem of making sure the type does not escape into binaries is almost the same either way as well.

test/lit/passes/dce-gc.wast
```diff
@@ -0,0 +1,104 @@
...
+  ;; CHECK-NEXT:   (local.get $ignore)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (block
+  ;; CHECK-NEXT:   (drop
+  ;; CHECK-NEXT:    (local.get $no)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:   (unreachable)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $param1 (param $ignore (ref null nofunc)) (param $no (ref nofunc))
```
tlively, line 16: We could be even more aggressive (and save some compile time!) by replacing the entire body of any function that takes an impossible argument with `unreachable`.

tlively (approved):

kripken:
> The problem of making sure the type does not escape into binaries is almost the same either way as well.

Good point, yeah, it's really the same. And using `unreachable` is nicer. So I guess that's a better plan.

test/lit/passes/dce-gc.wast
kripken: Makes sense.

I'm rethinking this though, given we should use `unreachable` in the long term. Maybe this isn't worth the work to scan the entire function another time. It might make more sense as a peephole optimization actually.

tlively: Yeah, I think this general optimization could make sense as a peephole optimization as well. The specific optimization for the impossible parameters still makes sense as its own pass, I think.

PR #5181, opened 2022-10-24
[Parser] Parse `return`

tlively:
Unlike in the legacy parser, we cannot depend on the folded text format to
determine how many values to return, so we determine that solely based on the
current function context.

To handle multivalue return correctly, fix a bug in which we could synthesize
new `unreachable`s and place them before existing unreachable instructions (like
returns) at the end of instruction sequences.

### DIFF
--- a/test/lit/wat-kitchen-sink.wast
+++ b/test/lit/wat-kitchen-sink.wast
@@ -39,6 +39,16 @@
 
   ;; CHECK:      (type $i32_i32_i64_i64_=>_none (func_subtype (param i32 i32 i64 i64) func))
 
+  ;; CHECK:      (type $i32_=>_i32 (func_subtype (param i32) (result i32) func))
+
+  ;; CHECK:      (type $i32_i64_=>_i32_i64 (func_subtype (param i32 i64) (result i32 i64) func))
+
+  ;; CHECK:      (type $i64_=>_i32_i64 (func_subtype (param i64) (result i32 i64) func))
+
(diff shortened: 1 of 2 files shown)


tlively:
* main
  * **PR #5181** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5181" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5181?utm_source=stack-comment).

test/lit/wat-kitchen-sink.wast
```diff
@@ -466,15 +476,9 @@
...
- ;; CHECK-NEXT:    (unreachable)
- ;; CHECK-NEXT:    (i32.const 2)
- ;; CHECK-NEXT:   )
- ;; CHECK-NEXT:   (i32.add
- ;; CHECK-NEXT:    (i32.const 3)
- ;; CHECK-NEXT:    (i32.const 4)
- ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:  (i32.add
+ ;; CHECK-NEXT:   (unreachable)
+ ;; CHECK-NEXT:   (i32.const 2)
```
kripken, line 481: Looks correct, but is this still testing all we wanted it to test earlier?

kripken (approved):

test/lit/wat-kitchen-sink.wast
tlively: Yep, these tests are only trying to check that the parse is correct in the sense that it has the same semantics as the input.

PR #5182, opened 2022-10-24
[NFC] Inherit from Visitor in OverriddenVisitor

kripken:
Doing so shortens the code by removing duplicate logic.

Also this will avoid a compile error in a future PR, as by inheriting from
Visitor we include functions like `visitFunction` which were otherwise
missing from OverriddenVisitor. We could duplicate those like we
duplicated the expression logic, but just removing all the duplication
seems best.

I manually verified OverriddenVisitor still provides the same error messages
as before.


tlively (approved):
Huh, I'm surprised this wasn't always the case.

PR #5184, opened 2022-10-24
Fix a fuzz issue with scanning heap read types

### LINKED ISSUE #5180: Fuzz error with null types
```wat
(module
 (type $type$2 (struct_subtype (field (mut (ref null $type$2))) data))
 (type $type$0 (func_subtype func))
 (type $type$1 (func_subtype (param (ref $type$2) (ref null $type$2)) func))
 (type $type$3 (func_subtype (param funcref i32) (result funcref) func))
 (elem declare func $0)
 (func $0 (type $type$0)
  (nop)
 )
 (func $1 (type $type$1) (param $0 (ref $type$2)) (param $1 (ref null $type$2))
  (drop
   (struct.get $type$2 0
    (local.get $1)
   )
  )
 )
 (func $2 (type $type$3) (param $0 funcref) (param $1 i32) (result funcref)
  (call $1
   (ref.as_non_null
    (ref.null none)
   )
   (ref.as_non_null
    (ref.null none)
   )
  )
  (ref.func $0)
 )
)
```

```wat
$ wasm-opt a.wat -all --nominal --signature-refining --print --gto

(module
 (type $type$0 (func_subtype func))
 (type $type$1 (func_subtype (param (ref none) (ref none)) func))
 (type $type$3 (func_subtype (param funcref i32) (result (ref $type$0)) func))
 (elem declare func $0)
 (func $0 (type $type$0)
  (nop)
 )
 (func $1 (type $type$1) (param $0 (ref none)) (param $1 (ref none))
  (drop
   (block ;; (replaces something unreachable we can't emit)
    (drop
     (local.get $1)
    )
    (unreachable)
   )
  )
 )
 (func $2 (type $type$3) (param $0 funcref) (param $1 i32) (result (ref $type$0))
  (call $1
   (ref.as_non_null
    (ref.null none)
   )
   (ref.as_non_null
    (ref.null none)
   )
  )
  (ref.func $0)
 )
)

src/ir/type-updating.cpp:140: wasm::HeapType wasm::GlobalTypeRewriter::update()::CodeUpdater::getNew(wasm::HeapType): Assertion `oldToNewTypes.count(type)' failed.
```

The problem appears to be that `struct.get` that is printed as unreachable. It is printed as unreachable, but it isn't actually unreachable. `BINARYEN_PRINT_FULL=1` shows this:

```wat
 (func $1 (type $type$1) (param $0 (ref none)) (param $1 (ref none))
  [none] (drop
   [(ref null $struct.0)] (block ;; (replaces something unreachable we can't emit)
    [none] (drop
     [(ref none)] (local.get $1)
    )
    [unreachable] (unreachable)
   )
  )
 )
```

That is, signature refining turned the `struct.get` input into a null type. So we print it as unreachable. But the type of the expression was not updated, and still refers to the type that used to be returned here. When we run GTO, that ends up being the only place that refers to that type, and so we hit that assertion on not knowing how to map it.

It does look like if the type is null then we do not update the expression's type:

https://github.com/WebAssembly/binaryen/blob/899ffddd109710d0e8794aa7f0ce36aa0410407f/src/wasm/wasm.cpp#L1036-L1042

Perhaps we should update it to unreachable @tlively ? Do you remember why we don't do that already?

kripken:
If a heap type only ever appears as the result of a read, we must include it in
the analysis in ModuleUtils, even though it isn't written in the binary format.
Otherwise analyses using ModuleUtils can error on not finding all types in the
list of types.

Fixes #5180


tlively (commented):
I hadn't realized we were talking about the type `struct.get` _returns_, which doesn't correspond to any type annotation at all. Only including the return type of `struct.get` seems awfully ad hoc. Why is `struct.get` special? We should either include the returns types of every expression or we should fix this another way.

When GTO optimizes a `struct.get` ref to be nullref, could we special-case that and replace the struct.ref with an unreachable?

We should also make sure the optimized type ordering used to emit binaries does not contain types with count 0.

kripken:
> Why is struct.get special?

I think it's exactly that property that is the issue here, that we don't refinalize them automatically, so a "stale" type can remain. I agree that feels like a hack.

> When GTO optimizes a struct.get ref to be nullref, could we special-case that and replace the struct.ref with an unreachable?

I think it would be better to do it in refinalize so it happens in all possible locations, not just that particular pass. But I don't have a full sense of how big the problem you mentioned in the issue was, about unreachability in the parsers.

kripken:
> I think it would be better to do it in refinalize so it happens in all possible locations, not just that particular pass.

To clarify that, I said that because I think this can happen in any of our various refining passes.

tlively:
I'd be ok with this if we do the same thing for `array.get`, which in principle is in the same position as `struct.get`, and if we add tests demonstrating that we don't end up emitting the unnecessary type into the type section.

kripken:
ArrayGet should already be handled:

https://github.com/WebAssembly/binaryen/pull/5184/files#diff-a7d5babfd44d7b59a2fc7d828bce7eac17bce1a4669bab7bbf0ab1bb5461945cR91

How do you suggest testing the contents of the type section? Some kind of byte-by-byte comparison against a reference file that has the optimal order..? Or to do tons of gets and see they don't affect the order? I'm not too happy with either idea but probably I'm missing something...

This also isn't urgent. If we think we can fix it in the parsers eventually that might be fine too?

tlively:
We can test the type section by printing the text module and `CHECK-NOT` asserting that the type is not printed. (The text format and binary format calculate the types to include in the same way.)

kripken:
Sounds good, `CHECK-NOT` added.

PR #5185, opened 2022-10-24
[Wasm GC] Remove unused signature results

kripken:
If a signature type is always dropped after being called/callRefed then
we can remove the result from the signature and then remove the drops.

This affects some 1.5% of total calls in J2WASM - not a lot, but more than I would
have expected actually.


src/passes/SignatureRefining.cpp
```diff
@@ -50,6 +51,28 @@ struct SignatureRefining : public Pass {
...
+
+    // A possibly improved LUB for the results.
+    LUBFinder resultsLUB;
+
+    // Normally we can optimize, but some cases prevent a particular signature
+    // type from being changed at all, see below.
+    bool canModify = true;
+
+    // Sometimes we can modify the signature as a whole, but not the results.
+    bool canModifyResults = true;
```
kripken, line 73: (This is the only change (addition) to this structure, aside from moving it to the outer scope so we can use it in a helper function.)

src/passes/DeadArgumentElimination.cpp
```diff
@@ -317,25 +317,7 @@ struct DAE : public Pass {
...
-        curr->value = nullptr;
-        Builder builder(*module);
-        replaceCurrent(builder.makeSequence(builder.makeDrop(value), curr));
-      }
-    } returnUpdater(func, module);
-    // Remove any value flowing out.
-    if (func->body->type.isConcrete()) {
-      func->body = builder.makeDrop(func->body);
-    }
+    ParamUtils::removeReturns(func, module);
```
kripken, line 320: (This was moved so we can share it.)

PR #5186, opened 2022-10-25
[Wasm GC] Fix the depth of the new array heap type

kripken:
Found by the fuzzer. Add some automatic testing that would have found this,
and also a dedicated test.

### DIFF
--- a/src/ir/subtypes.h
+++ b/src/ir/subtypes.h
@@ -116,13 +116,12 @@ struct SubTypes {
     // TODO: update when we get structtype
     for (auto type : types) {
       HeapType basic;
-      if (type.isData()) {
-        if (type.isArray()) {
-          basic = HeapType::array;
-        } else {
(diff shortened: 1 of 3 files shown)


src/ir/subtypes.h
```diff
@@ -113,12 +113,23 @@ struct SubTypes {
...
+      HeapType basic;
+      if (type.isData()) {
+        if (type.isArray()) {
+          basic = HeapType::array;
+        } else {
+          basic = HeapType::data;
+        }
+      } else {
+        basic = HeapType::func;
+      }
```
tlively, line 127:
```suggestion
      if (type.isStruct()) {
        basic = HeapType::data;
      } else if (type.isArray()) {
        basic = HeapType::array;
      } else {
        assert(type.isSignature());
        basic = HeapType::func;
      }
```

tlively (approved):

PR #5187, opened 2022-10-25
Move removable code in CodePushing

kripken:
This is safe since we "partially remove" it: we don't move it to a place it might
execute more, but make it possibly execute less. See the new comment for
more details.

Motivated by wasm GC but this can help wasm MVP as well. In both cases
loads from memory can trap, which limits what the VM can do to optimize
them past conditions, but in trapsNeverHappens we can do that at the
toolchain level:

```js
x = read();
if () { .. }
use(x);

=>

if () { .. }
x = read(); // moved to here, and might not execute if the if did a break/return
use(x);
```


src/passes/CodePushing.cpp
```diff
@@ -124,11 +124,26 @@ class Pusher {
...
+    // move them, because of stuff like this:
+    //
+    //   if (x != 0) foo(1 / x);
+    //
+    // If we move 1 / x to execute unconditionally then it may trap, but it
+    // would be fine to remove it. This pass does not move code to places where
+    // it might execute more, but *less*: we keep the code behind any conditions
+    // it was already behind, and potentially put it behind further ones. In
+    // effect, we "partially remove" the code, making it not execute some of the
+    // time, which is fine.
```
aheejin, line 142: Code LGTM, but not sure if I understand this... in `trapsNeverHappen` mode, we simply assume traps never happen, so it doesn't matter if we execute possibly trapping code more of less.. no?

kripken: It's a little tricky. Given `if (x != 0) foo(1/x)` we assume it never traps as it executes normally. But the only reason it doesn't trap might be that the `if` prevents it (that's what "executes normally" means, sort of). So moving the `1/x` out of the if is risky. (But in this pass we only execute code less, not more, so it's ok.)

The old "ignore implicit traps" option was basically "nothing can trap (even if you move it)", but the movement turned out to be a problem, so we added trapsNeverHappen which is more subtle.

Docs for this should be here: (but it is just a tricky topic, unfortunately - more complex than I'd like)

https://github.com/WebAssembly/binaryen/blob/79e032ac1124f82729bb21684f370044612c7124/src/pass.h#L119-L167

aheejin: Thanks for the pointer. I had read it before but apparently forgot it... 😅

aheejin (approved):

PR #5191, opened 2022-10-27
CodePushing: Push into If arms

kripken:
Previously the pass only pushed past an `if` or a `br_if`. This does the same but into an
`if` arm. On Wasm GC for example this can perform allocation sinking:
```js 
function foo() {
  x = new A();
  if (..) {
    use(x);
  }
}

=>

function foo() {
  if (..) {
    x = new A(); // this moved
    use(x);
  }
}
```
The allocation won't happen if we never enter the if. This helps wasm MVP too,
and in fact some existing tests benefit.


src/passes/CodePushing.cpp
```diff
@@ -237,6 +244,178 @@ class Pusher {
...
+    // Here we will try to push y first, and then x. Note that if we push y
+    // then we can immediately try to push x after it, as it will remain in
+    // order with x if we do. If we do *not* push y we can still try to push x
+    // but we must move it past y, which means we need to check for interference
+    // between them (which we do by adding y's effects to cumulativeEffects).
+    //
+    // Decrement at the top of the loop for simplicity, so start with i at one
+    // past the first thing we can push (which is right before the push point).
+    Index i = pushPoint;
+    while (1) {
```
tlively, line 311: I was going to suggest using `while (true)` here, but it looks like `while (1)` is much more common in the code base (just not in any code I wrote 😅).

src/passes/CodePushing.cpp
```diff
@@ -237,6 +244,178 @@ class Pusher {
...
+      }
+
+      auto index = pushable->index;
+
+      auto iter = pushableEffects.find(pushable);
+      if (iter == pushableEffects.end()) {
+        iter = pushableEffects
+                 .emplace(std::piecewise_construct,
+                          std::forward_as_tuple(pushable),
+                          std::forward_as_tuple(passOptions, module, pushable))
```
tlively, line 334: Is all this verbosity worth it? We could use `emplace({pushable, EffectAnalyzer(passOptions, module, pushable)})` and it would be much more readable. This could also be pulled out into a helper function.

src/passes/CodePushing.cpp
```diff
@@ -237,6 +244,178 @@ class Pusher {
...
+          // in the other arm which as mentioned above is the only thing that
+          // could have made it work out for us.
+          return false;
+        }
+
+        // We can do it! Push into one of the if's arms, and put a nop where it
+        // used to be.
+        Builder builder(module);
+        auto* block = builder.blockify(arm);
+        arm = block;
```
tlively, line 399: It doesn't look like the separate `block` variable adds much here.
```suggestion
        arm = builder.blockify(arm);
```

test/lit/passes/code-pushing_into_if.wast
```diff
@@ -0,0 +1,540 @@
...
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $unpushed-interference (param $p i32)
+    (local $x i32)
+    (local $y i32)
+    (local.set $x (i32.const 1))
+    ;; This set is not pushed (as it is not used in the if) and it will then
+    ;; prevent the previous set of $x from being pushed, since we can't push a
+    ;; set of $x past a get of it.
+    (local.set $y (local.get $x))
```
tlively, line 147: How about a test case where $y is pushed into one arm and $x would be pushed into the other arm except for the get of $x that went along with $y?

test/lit/passes/code-pushing_into_if.wast
```diff
@@ -0,0 +1,540 @@
...
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $if-condition-return (param $p i32)
+    (local $x i32)
+    (local.set $x (i32.const 1))
+    (if
+      (block (result i32)
+        (return) ;; This return does not prevent us from optimizing; if it
+                 ;; happens then we don't need the local.set to execute
+                 ;; anyhow.
```
tlively, line 534: What if this is not a return but a branch to an outer block after which the local is used?

src/passes/CodePushing.cpp
kripken: I think `while (1)` might be a JS habit of mine :smile: though I guess even there it could be `true`. I'd be fine with a bulk replace in the entire codebase if we have a preference one way or the other.

src/passes/CodePushing.cpp
kripken: Hmm, I copied this from the other function. Trying it your way, I get
```cpp
/home/azakai/Dev/2-binaryen/src/passes/CodePushing.cpp:331:39: error: no matching function for call to ‘std::unordered_map<wasm::LocalSet*, wasm::EffectAnalyzer>::emplace(<brace-enclosed initializer list>)’
  331 |         iter = pushableEffects.emplace({pushable, EffectAnalyzer(passOptions, module, pushable)}).first;
      |                ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I guess that's why the other function does it... That's more C++ than I understand, to be honest, not sure why it's unhappy. But I can do a helper function, that sounds good.

src/passes/CodePushing.cpp
kripken: We need the block on the next line,
```cpp
        ExpressionManipulator::spliceIntoBlock(block, 0, pushable);
```
That function takes a `Block` as its input.

src/passes/CodePushing.cpp
```diff
@@ -272,9 +451,8 @@ struct CodePushing : public WalkerPass<PostWalker<CodePushing>> {
 
   void visitBlock(Block* curr) {
     // Pushing code only makes sense if we are size 3 or above: we need
```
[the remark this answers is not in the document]
kripken: Good catch, yes, stale comment.

test/lit/passes/code-pushing_into_if.wast
kripken: Good catch! Looks like that's a bug in the current logic.

test/lit/passes/code-pushing_into_if.wast
kripken: That's handled by the shared logic in `isPushable`, which rules out such cases. But good to test it here as well, I added a test now.

test/lit/passes/code-pushing_into_if.wast
kripken: Fixed.

src/passes/CodePushing.cpp
tlively: Ugh. I wonder if using `std::make_pair` instead of `{}` would work.

tlively: It also looks like there is a method [`try_emplace`](https://en.cppreference.com/w/cpp/container/unordered_map/try_emplace) that builds in all this `piecewise_construct` and `forward_as_tuple` junk so you don't have to think about it.

tlively (approved):

src/passes/CodePushing.cpp
kripken: Nice, looks like it builds with `try_emplace`...

PR #5192, opened 2022-10-27
[NFC] Rewrite PossibleContents::combine to be static

kripken:
As suggested by @tlively in the past. This makes the method symmetric.

Measuring speed, this seems identical to before, so that concern seems fine.

However, after writing this I'm not sure this is worth it. Perhaps the `combine()` method itself is a little easier to read, but using the API gets more cumbersome in the common case:

```diff
-contents.combine(newContents);
+contents = PossibleContents::combine(contents, newContents);
```

(There is also some risk of using it without realizing it is static, `contents.combine(.., ..)` can be written, but it does not modify `contents`. I guess `[[nodiscard]]` could help there though.)

### DIFF
--- a/src/ir/possible-contents.cpp
+++ b/src/ir/possible-contents.cpp
@@ -1629,7 +1629,7 @@ bool Flower::updateContents(LocationIndex locationIndex,
   std::cout << '\n';
 #endif
 
-  contents = PossibleContents::combine(contents, newContents);
+  contents.combine(newContents);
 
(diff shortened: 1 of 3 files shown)


tlively (approved):
I do think this is easier to read, thanks! For keeping client code shorter, I'd be ok with having a method like this:

```
void combine(const PossibleContents& other) {
  *this = PossibleContents::combine(*this, other);
}
```

PR #5194, opened 2022-10-28
[Wasm GC] SimplifyLocals: Switch local.get to use a more refined type when possible

kripken:
```wat
(local.set $refined (cast (local.get $plain)))
.. (local.get $plain) .. ;; we can change this to read from $refined
```
By using the more refined type we may be able to eliminate casts later. I see a
handlful that are eliminated by this in the big J2Wasm binary.

To do this, look at the fallthrough value (so we can look through a cast or a block
value - this is the reason for the small wasm2js improvements in tests), and also
extend the code that picks which local index to read to look at types (previously
we just ignored any pairs of locals with different types).


src/passes/SimplifyLocals.cpp
```diff
@@ -1067,35 +1060,70 @@ struct SimplifyLocals
...
+              refinalize = true;
+            }
           }
         }
       }
     };
 
     EquivalentOptimizer eqOpter;
     eqOpter.module = this->getModule();
+    eqOpter.passOptions = this->getPassOptions();
```
tlively, line 1120: Is the `this->` necessary here?

tlively (approved):
Is it worth testing with more interesting interactions between subtype relationships and live ranges?

src/passes/SimplifyLocals.cpp
kripken: Unfortunately yes,
```cpp
SimplifyLocals.cpp:1120:27: error: there are no arguments to ‘getPassOptions’ that depend on a template parameter, so a declaration of ‘getPassOptions’ must be available [-fpermissive]
 1120 |     eqOpter.passOptions = getPassOptions();
      |                           ^~~~~~~~~~~~~~
SimplifyLocals.cpp:1120:27: note: (if you use ‘-fpermissive’, G++ will accept your code, but allowing the use of an undeclared name is deprecated)
```
Or, from clang,
```cpp
SimplifyLocals.cpp:1120:27: error: use of undeclared identifier 'getPassOptions'; did you mean 'PassOptions'?
    eqOpter.passOptions = getPassOptions();
                          ^
```
Something to do with all this being inside a templated struct, I think.

kripken:
What kind of tests did you have in mind? The new tests here check what changed, which is that we may now pick a local of another type, so long as we go to a subtype. (Otherwise the handling of separate live ranges etc. should be identical to the old code which is already tested.)

kripken:
I added a non-gc test for the fallthrough value, which I realize helps non-gc too (as shown by the non-lit tests, but good to have a lit test as well).

tlively (approved):

PR #5195, opened 2022-10-29
[C API] Add APIs to inspect compound heap types

dcodeIO:
Adds C APIs to inspect compound struct, array and signature heap types:

Obtain field types, field packed types and field mutabilities of struct types:

* BinaryenStructTypeGetNumFields (to iterate)
  * BinaryenStructTypeGetFieldType
  * BinaryenStructTypeGetFieldPackedType
  * BinaryenStructTypeIsFieldMutable

Obtain element type, element packed type and element mutability of array types:

* BinaryenArrayTypeGetElementType
* BinaryenArrayTypeGetElementPackedType
* BinaryenArrayTypeIsElementMutable

Obtain parameter and result types of signature types:

* BinaryenSignatureTypeGetParams
* BinaryenSignatureTypeGetResults

### DIFF
--- a/src/binaryen-c.h
+++ b/src/binaryen-c.h
@@ -165,6 +165,23 @@ BINARYEN_API BinaryenHeapType
 BinaryenHeapTypeGetBottom(BinaryenHeapType heapType);
 BINARYEN_API bool BinaryenHeapTypeIsSubType(BinaryenHeapType left,
                                             BinaryenHeapType right);
+BINARYEN_API BinaryenIndex
+BinaryenStructTypeGetNumFields(BinaryenHeapType heapType);
+BINARYEN_API BinaryenType
+BinaryenStructTypeGetFieldType(BinaryenHeapType heapType, BinaryenIndex index);
+BINARYEN_API BinaryenPackedType BinaryenStructTypeGetFieldPackedType(
+  BinaryenHeapType heapType, BinaryenIndex index);
+BINARYEN_API bool BinaryenStructTypeIsFieldMutable(BinaryenHeapType heapType,
+                                                   BinaryenIndex index);
+BINARYEN_API BinaryenType
+BinaryenArrayTypeGetElementType(BinaryenHeapType heapType);
+BINARYEN_API BinaryenPackedType
+BinaryenArrayTypeGetElementPackedType(BinaryenHeapType heapType);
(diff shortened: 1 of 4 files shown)


src/binaryen-c.h
```diff
@@ -165,6 +165,23 @@ BINARYEN_API BinaryenHeapType
...
+                                                   BinaryenIndex index);
+BINARYEN_API BinaryenType
+BinaryenArrayTypeGetElementType(BinaryenHeapType heapType);
+BINARYEN_API BinaryenPackedType
+BinaryenArrayTypeGetElementPackedType(BinaryenHeapType heapType);
+BINARYEN_API bool BinaryenArrayTypeIsElementMutable(BinaryenHeapType heapType);
+BINARYEN_API BinaryenType
+BinaryenSignatureTypeGetParams(BinaryenHeapType heapType);
+BINARYEN_API BinaryenType
+BinaryenSignatureTypeGetResults(BinaryenHeapType heapType);
```
kripken, line 184: How about `FooType` => `Foo`? That is, `BinaryenStructTypeGetNumFields => BinaryenStructGetNumFields`.

We do need `HeapType` in full since `Heap` is meaningless and `HeapType` is a proper term, but `Struct` is a proper term by itself (it appears in `wasm-type.h`).

I don't feel strongly though. @tlively ?

dcodeIO: Have also been thinking about how to best name these. If we were strict, these should probably be `BinaryenHeapTypeXY` because the first argument is a heap type, but this introduces surface area for naming collisions in case there are more kinds of types that have "params", "results", an "element" or "fields" eventually. Right now it should work without collisions. Using `StructType` instead of `HeapType` was a bit of a compromise on my end, since a "struct type" implies a more specific form of "heap type", somewhat. Could also shorten to "Struct" I guess. Not feeling strongly either way as well.

1. BinaryenHeapType because first argument is a heap type
2. BinaryenStructType to hint that this is some sort of special type
3. BinaryenStruct to stay brief

tlively: I think `StructType` makes sense, mostly to differentiate from functions operating on `struct.*` instructions.

kripken: Sounds good.

PR #5202, opened 2022-10-31
[Wasm GC] Enable various passes in hybrid mode, not just nominal

kripken:


tlively (approved):

kripken:
@askeksa-google This should fix any differences between nominal and hybrid mode. Let us know if you find any regressions we've missed.

askeksa-google:
I tested this change on barista3. The hybrid mode module is 6% bigger than the nominal mode module, but the difference seems to be solely due to more types in the type section. I assume that in nominal mode, Binaryen can remove all unused types, whereas in hybrid mode, it can only remove whole recursion groups, so if the input contains one big recursion group, no types can be eliminated. Does this sound right? It would explain the remaining difference.

tlively:
Yes, that sounds correct 👍

PR #5205, opened 2022-11-01
[NFC] Add a TODO in ReorderFunctions

kripken:


tlively (approved):

tlively:
Might be worth noting that we should also remember to add one to the counts of functions used with `ref.func` to account for the declarative segment.

PR #5209, opened 2022-11-02
Fix includes when using Binaryen together with Wasmtime

TheGreatRambler:
When using Binaryen and Wasmtime in the same C/C++ project the two have conflicting headers. Binaryen exposes most functionality through `wasm.h` but Wasmtime also does so through [wasm-c-api](https://github.com/WebAssembly/wasm-c-api/blob/master/include/wasm.h). It is possible it ignore Wasmtime's header by exclusively using `wasmtime.h` but whenever it makes a reference to `wasm.h` it does so through the include path, not relatively because it would be quite messy (`../../wasm-c-api/include/wasm.h`). A possible fix is what I implemented.

If I put Wasmtime before Binaryen in my include path Wasmtime will resolve to the correct header. To enforce this order I use:

```
include_directories(${WASMTIME_INCLUDES})
include_directories(${BINARYEN_DIR}/src)
```

When using Binaryen I never refer to `wasm.h` and instead use `wasm-binary.h`, `wasm-io.h` etc. Binaryen however refers to `wasm.h` a lot throughout the codebase. This PR makes every internal reference to `wasm.h` relative so both Binaryen and Wasmtime are happy.

If a test being added would be appreciated or any normalization of other headers I can implement that.


tlively:
So the problem is that both projects have a header called `wasm.h` and they conflict? Is there a less invasive way to solve this problem that doesn't require modifying either project?

src/passes/AvoidReinterprets.cpp
```diff
@@ -20,11 +20,11 @@
...
-#include <ir/local-graph.h>
-#include <ir/properties.h>
-#include <pass.h>
-#include <wasm-builder.h>
-#include <wasm.h>
+#include "ir/local-graph.h"
+#include "ir/properties.h"
+#include "pass.h"
+#include "wasm-builder.h"
+#include "../wasm.h"
```
kripken, line 27: I think changes from `<` to `"` definitely make sense to do.

But I hope we can avoid the `../` changes. For those, wouldn't it work if the binaryen sources are built with different `-I` paths than the wasmtime sources?

TheGreatRambler: If every path was made relative throughout the codebase yeah binaryen could just be included differently but it would be as `src/wasm.h` not `binaryen/wasm.h`

TheGreatRambler:
> So the problem is that both projects have a header called `wasm.h` and they conflict? Is there a less invasive way to solve this problem that doesn't require modifying either project?

Both projects currently expect their `wasm.h` to be first on the include path so unfortunately not really

src/passes/AvoidReinterprets.cpp
kripken: Sorry, I don't follow. By "codebase" which codebase do you mean?

TheGreatRambler: In binaryen sorry, making every single path in binaryen `../[something]` and using no `<something.h>` would make that possible, but that would just create even more `../` changes. Binaryen can't be given its own personal include path so one project would have to refer to only relative paths. There might be another fix but I don't know of it, maybe with CMake.

kripken: > Binaryen can't be given its own personal include path so one project would have to refer to only relative paths

There should be a way to do this in most build systems I hope. For example in CMake if you put Binaryen in one directory you could add include paths in the CMakeLists.txt in that directory which only take effect there.

TheGreatRambler: Are you saying I use binaryen and wasmtime in separate projects and just link one to the other? I definitely could do that, that would fix the include clashing. I can close this PR then

kripken: Yes, not building their sources all at once and just linking the compiled code is a good way to avoid such issues. Different build systems help do that in different ways (I mentioned how CMake does it; I'm less familiar with others).

TheGreatRambler:
A proper solution to this that does not involve modifying either library exists (using each library in different libraries and linking them together later) so I'll close this for now

PR #5211, opened 2022-11-02
JSPI - Support options to control what imports/exports are async.

brendandahl:
Extract and reuse ModuleAnalyzer and argument parsing code from Asyncify in JSPI.

Note: There's some shared code that really only needs to be in Asyncify, but I didn't want to change too much for now.


brendandahl:
After talking with Francis we're not going to do auto async for now, but explicitly specify async imports/exports.

PR #5208, opened 2022-11-02
Fix binary parsing of data segment memory

tlively:
The binary parser was eagerly getting the name of memories to set the `memory`
field of data segments, but that meant that when the memory names were updated
later while parsing the names section, the data segment memory fields would
become out of date. Update the issue by deferring setting the `memory` fields
like we do for other parts of IR that reference memories.

Also fix a segfault in the validator that was triggered by the reproducer for
this bug before the bug was fixed.

Fixes #5204.

PR #5214, opened 2022-11-03
Implement `array.new_data` and `array.new_elem`

tlively:
In order to test them, fix the binary and text parsers to accept passive data
segments even if a module has no memory. In addition to parsing and emitting the
new instructions, also implement their validation and interpretation. Test the
interpretation directly with wasm-shell tests adapted from the upstream spec
tests. Running the upstream spec tests directly would require fixing too many
bugs in the legacy text parser, so it will have to wait for the new text parser
to be ready.


tlively:
* main
  * **PR #5208** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5208" target="_blank">Graphite</a> 
    * **PR #5214** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5214" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5214?utm_source=stack-comment).

src/ir/possible-contents.cpp
```diff
@@ -887,6 +887,14 @@ struct InfoCollector
     }
     addRoot(curr, PossibleContents::exactType(curr->type));
   }
+  void visitArrayNewSeg(ArrayNewSeg* curr) {
+    if (curr->type == Type::unreachable) {
+      return;
+    }
+    // TODO: How to add a connection between data/elem segments and the
+    // allocated value / its type?
+    WASM_UNREACHABLE("What is this supposed to do???");
```
tlively, line 896: @kripken, guidance on what to do here would be much appreciated.

src/ir/effects.h
```diff
@@ -755,6 +755,7 @@ class EffectAnalyzer {
       }
     }
     void visitArrayNew(ArrayNew* curr) {}
+    void visitArrayNewSeg(ArrayNewSeg* curr) {}
```
kripken, line 758: Can this not trap if the segment has been dropped or something like that?

src/ir/possible-contents.cpp
kripken: As the segments are not immutable I guess the values here must be roots. So something like
```cpp
addRoot(DataLocation{type, 0}, PossibleContents::exactType(type from the segment?));
```

src/literal.h
```diff
@@ -197,6 +197,40 @@ class Literal {
         WASM_UNREACHABLE("unexpected type");
     }
   }
+
+  static Literal makeFromMemory(void* p, Type type) {
```
kripken, line 201: Maybe the body could be in `literal.cpp`? Doesn't seem like it needs to be super-fast.

src/wasm-interpreter.h
```diff
@@ -3494,6 +3499,83 @@ class ModuleRunnerBase : public ExpressionRunner<SubType> {
...
+
+    Literals contents;
+    contents.reserve(size);
+
+    switch (curr->op) {
+      case NewData: {
+        assert(curr->segment < wasm.dataSegments.size());
+        assert(elemType.isNumber());
+        const auto& seg = *wasm.dataSegments[curr->segment];
+        auto elemBytes =
```
kripken, line 3528: Maybe `Field` could have `getByteSize` that does this?

src/wasm-interpreter.h
```diff
@@ -3494,6 +3499,83 @@ class ModuleRunnerBase : public ExpressionRunner<SubType> {
...
+              memcpy(&val, addr, 1);
+              contents.push_back(Literal(int32_t(val)));
+              break;
+            }
+            case Field::i16: {
+              uint16_t val;
+              memcpy(&val, addr, 2);
+              contents.push_back(Literal(int32_t(val)));
+              break;
+            }
```
kripken, line 3555: Maybe these memcpys could also be in `Literal::makeFromMemory`? If that were in `literal.cpp` (as separately suggested above) then we could just forward-declare `Field` I think and it would work.

src/wasm/wasm-binary.cpp
```diff
@@ -3122,12 +3122,15 @@ void WasmBinaryBuilder::readDataSegments() {
...
-    if (!curr->isPassive) {
+    if (curr->isPassive) {
+      curr->memory = Name();
+      curr->offset = nullptr;
+    } else {
+      Index memIdx = 0;
+      if (flags & BinaryConsts::HasIndex) {
+        memIdx = getU32LEB();
+      }
+      memoryRefs[memIdx].push_back(&curr->memory);
```
kripken, line 3133: Is this a separate change?

src/ir/effects.h
tlively: Oh yes, good point. Also for out-of-bounds accesses.

src/wasm-interpreter.h
tlively: It turns out it already does!

src/wasm/wasm-binary.cpp
tlively: No, the test I added depends on being able to have passive segments without any memories. The previous code incorrectly segfaulted because it unconditionally assumed there was a memory available.

kripken: I see, thanks!

tlively:
I have no idea why the test output changed. Trying to track that down now...

src/ir/effects.h
```diff
@@ -754,7 +754,12 @@ class EffectAnalyzer {
         parent.implicitTrap = true;
       }
     }
-    void visitArrayNew(ArrayNew* curr) {}
+    void visitArrayNew(ArrayNew* curr) {
+      // Traps on out of bounds access to segments or access to dropped
+      // segments.
+      parent.implicitTrap = true;
```
tlively, line 760: Oh, this is why the tests changed 😬

src/ir/possible-contents.cpp
```diff
@@ -896,6 +896,27 @@ struct InfoCollector
...
+  void visitArrayNewSeg(ArrayNewSeg* curr) {
+    if (curr->type == Type::unreachable) {
+      return;
+    }
+    auto heapType = curr->type.getHeapType();
+    switch (curr->op) {
+      case NewData: {
+        Type elemType = heapType.getArray().element.type;
+        addRoot(DataLocation{heapType, 0},
+                PossibleContents::exactType(elemType));
```
kripken, line 908:
```suggestion
                PossibleContents::fromType(elemType));
```

src/ir/possible-contents.cpp
```diff
@@ -896,6 +896,27 @@ struct InfoCollector
...
+        return;
+      }
+      case NewElem: {
+        Type segType = getModule()->elementSegments[curr->segment]->type;
+        addRoot(DataLocation{heapType, 0},
+                PossibleContents::fullConeType(segType));
```
kripken, line 914:
```suggestion
                PossibleContents::fromType(segType));
```

src/ir/possible-contents.cpp
kripken: (we don't know this is an exact type, do we? regardless the `fromType` helper is meant for this: a situation where all you know is the wasm type)

kripken (approved):

kripken:
Please fuzz this as well.

src/ir/possible-contents.cpp
kripken: (Ah, it's an exact type since it's coming from a data section... makes sense. Still, please use the helper.)

tlively:
Fuzzed over 3k iterations with no problems. (But of course we don't have specific fuzzer support for these instructions)

src/ir/possible-contents.cpp
```diff
@@ -896,6 +896,26 @@ struct InfoCollector
...
+    switch (curr->op) {
+      case NewData: {
+        Type elemType = heapType.getArray().element.type;
+        addRoot(DataLocation{heapType, 0},
+                PossibleContents::fromType(elemType));
+        return;
+      }
+      case NewElem: {
+        Type segType = getModule()->elementSegments[curr->segment]->type;
+        addRoot(DataLocation{heapType, 0}, PossibleContents::fromType(segType));
```
kripken, line 913: Sorry for missing this in review, this should be
```
info.links.push_back({DataLocation{heapType, 0}, PossibleContents::fromType(segType)});
addRoot(curr, PossibleContents::exactType(curr->type));
```

kripken: Actually the first line should be a root as well, sorry, since we know nothing about it but the type.

PR #5217, opened 2022-11-03
Add arguments to control which imports/exports are JSPI'd.

brendandahl:
Instead of automatically determining which exports will be async they will be explicitly set by the user. We'll rely on the runtime trapping if they are incorrectly set.

Two new arguments that behave similar to asyncify-imports:
 - jspi-imports
 - jspi-exports


brendandahl:
In the future we could use my work in #5211 to check that all the correct exports are specified.

src/passes/JSPI.cpp
```diff
@@ -22,26 +22,71 @@
...
+#include "support/file.h"
+#include "support/string.h"
 #include "wasm-builder.h"
 #include "wasm.h"
 #include <utility>
 
 //
 // Convert a module to be compatible with JavaScript promise integration (JSPI).
-// All exports will be wrapped with a function that will handle storing
+// Promising exports will be wrapped with a function that will handle storing
```
kripken, line 33: "An export of great promise" :wink:

kripken (approved):

PR #5220, opened 2022-11-04
[Wasm GC] Fix GUFA on externalize/internalize

### LINKED ISSUE #5219: Live code eliminated when using `--gufa`
When optimizing [hello.wasm.gz](https://github.com/WebAssembly/binaryen/files/9939464/hello.wasm.gz) with `wasm-opt -all -O3 --gufa -O3 -tnh --nominal -g`, the resulting module hits an `unreachable` instruction.

In the original module, I have:
```wat
 (func $$main (type $type$40)
  (drop
   (call $main)
  )
 )
```
```wat
 (func $main (type $type$41) (result (ref null $type$42))
  (call $print)
  (ref.null none)
 )
```
```wat
 (func $print (type $type$40)
  (local $0 (ref $type$14))
  (local $1 (ref null $type$16))
  (local.set $0
   (call $_StringBase._interpolate
    (struct.new $type$4
     (i32.const 230)
     (i32.const 0)
     (global.get $global$10)
     (i64.const 1)
     (array.init_static $type$3
      (global.get $global$1346)
     )
    )
   )
  )
  (local.set $1
   (global.get $global$1347)
  )
  (call $dart2wasm.printToConsole\20\28import\29
   (extern.externalize
    (local.get $0)
   )
  )
 )
```
which gets optimized to:
```wat
 (func $$main (type $type$21)
  (drop
   (call $_StringBase._interpolate
    (struct.new $type$1
     (i32.const 230)
     (i32.const 0)
     (global.get $global$10)
     (i64.const 1)
     (array.init_static $type$0
      (global.get $global$1346)
     )
    )
   )
  )
  (unreachable)
 )
```
where the final call to `$dart2wasm.printToConsole\20\28import\29` (an imported function) is apparently deemed unreachable and replaced by `(unreachable)`.

kripken:
These operations emit a completely different type than their input, so they must be
marked as roots, and not as things that flow values through them (because then
we filter everything out as the types are not compatible).

Fixes #5219

cc @askeksa-google


src/ir/possible-contents.cpp
```diff
@@ -670,7 +670,16 @@ struct InfoCollector
     receiveChildValue(curr->ref, curr);
   }
   void visitRefAs(RefAs* curr) {
-    // TODO: optimize when possible: like RefCast, not all values flow through.
+    if (curr->op == ExternExternalize || curr->op == ExternInternalize) {
```
tlively, line 673: Hmm, maybe these should be a separate expression class since they're materially different than the other `RefAs` cast instructions.

tlively (approved):

src/ir/possible-contents.cpp
kripken: Yeah, I think we considered that when you added those. Since then we've seen 2 places that needed special handling, so maybe it could have been separate I guess. Still if it's just those 2 places I think having less code is worthwhile.

PR #5221, opened 2022-11-04
[NFC] Avoid scanning SmallSets twice during insert

kripken:
As suggested in #5218


src/support/small_set.h
```diff
@@ -61,16 +82,24 @@ struct UnorderedFixedStorage : public FixedStorageBase<T, N> {
...
-    assert(this->used < N);
+  using InsertResult = typename FixedStorageBase<T, N>::InsertResult;
 
+  InsertResult insert(const T& x) {
     // Find the insertion point |i| where x should be placed.
     size_t i = 0;
     while (i < this->used && this->storage[i] <= x) {
+      if (this->storage[i] == x) {
+        return InsertResult::AlreadyExists;
+      }
```
tlively, line 93: It would be better to change the loop to use `<` instead of `<=` and to move this check to be after the loop.

src/support/small_set.h
```diff
@@ -130,16 +160,9 @@ class SmallSetBase {
...
-        // The item already exists, so there is nothing to do.
-        return;
-      }
-      // We must add a new item.
-      if (fixed.used < N) {
-        // Room remains in the fixed storage.
-        fixed.insert(x);
-      } else {
-        // No fixed storage remains. Switch to flexible.
+      if (fixed.insert(x) == FixedStorage::InsertResult::CouldNotInsert) {
```
tlively, line 163: If we only check for this one result, do we really need the enum with three options?

kripken: I guess we don't need both. I switched it to an enum with two values, "no error" and "failed to insert" (which is an error the caller must handle). Could also be a bool but I think the enum is more easy to read, and less likely to be confused.

tlively (approved):

PR #5222, opened 2022-11-04
Multi-Memories Asyncify

ashleynh:
Adds support for the Asyncify pass to use Multi-Memories. This is specified by passing flag `--asyncify-in-secondary-memory`. Another flag, `--asyncify-secondary-memory-size`, is used to specify the initial and max size of the secondary memory.


src/passes/Asyncify.cpp
```diff
@@ -1551,6 +1562,21 @@ struct Asyncify : public Pass {
     auto verbose = options.getArgumentOrDefault("asyncify-verbose", "") != "";
     auto relocatable =
       options.getArgumentOrDefault("asyncify-relocatable", "") != "";
+    auto secondaryMemory =
+      options.getArgumentOrDefault("in-secondary-memory", "") != "";
```
kripken, line 1566:
```suggestion
      options.getArgumentOrDefault("asyncify-in-secondary-memory", "") != "";
```

src/passes/Asyncify.cpp
```diff
@@ -1551,6 +1562,21 @@ struct Asyncify : public Pass {
...
+
+    // Ensure there is a memory, as we need it.
+    if (secondaryMemory) {
+      auto secondaryMemorySizeString =
+        options.getArgumentOrDefault("secondary-memory-size", "1");
```
kripken, line 1571:
```suggestion
        options.getArgumentOrDefault("asyncify-secondary-memory-size", "1");
```

kripken (commented):
Great!

Code lgtm aside from the suggested change to the option names. Looks like all the existing ones are prefixed so it seems best to follow that.

For testing, I think we can use a smaller file - perhaps just `func $liveness1` and just the first `local.get` of that. That would be enough to see the memory is created and used properly.

Also the last two tests could be in the same file, to avoid repeating the testcase. This is a nice feature of lit tests. For example, [this test file](https://github.com/WebAssembly/binaryen/blob/20115c8e2a038646858a0408beca7be319c31035/test/lit/passes/optimize-instructions-gc-iit.wast#L1-L8) runs three times on the same input, each with a different `--check-prefix`:

https://github.com/WebAssembly/binaryen/blob/20115c8e2a038646858a0408beca7be319c31035/test/lit/passes/optimize-instructions-gc-iit.wast#L1-L8

So adding a second `--check-prefix` for a different memory size would be simpler I think.

kripken (commented):
Sorry, looks like I was wrong - simplifying the tests that much removed all load/stores from the instrumented functions. We only have the new `$asyncify_*` methods using loads and stores now. I think it would be enough to restore `$liveness1` back to its original state. But we don't need the other functions, that one is enough.

kripken (approved):
Great!

PR #5227, opened 2022-11-07
OptimizeInstructions: Check for possible added-constant overflows

kripken:
Fix a regression from #5025 : we subtract constants there, and we need to be aware
that such subtraction can change a constant from signed to unsigned if the comparison
is signed, as

```
0x80000000 - 1 = 0x7fffffff
```

`0x8000000` is a negative number when seen as signed, but always positive after the
subtraction.

Also remove a duplicate `matches` that was there for no reason.


src/passes/OptimizeInstructions.cpp
```diff
@@ -3994,26 +3994,32 @@ struct OptimizeInstructions
...
-            c1->value = c1->value.sub(c2->value);
-            c2->value = zero;
-            return curr;
+          // Also check for an overflow when doing C2-C1. We could check this in
+          // a way that considers whether the comparison is signed or unsigned,
+          // but for simplicity for now rule out all overflows in a simple way.
+          auto typeMaxBits = getBitsForType(add->type);
+          auto c1Bits = Bits::getMaxBits(c1, this);
+          auto c2Bits = Bits::getMaxBits(c2, this);
+          if (c1Bits + c2Bits < typeMaxBits) {
```
tlively, line 4006: I think it would be pretty simple to perform this check only for signed comparisons to avoid regressing out optimizations of unsigned comparisons.

kripken: I am actually not 100% sure that is sufficient. This area is rife with with corner cases, and the common usage has small constants anyhow, so I'd prefer to keep this simple. Unless you can see a solution that is easy to prove is always correct?

tlively: I think the existing code was fairly well-reasoned. It always subtracted a smaller number from a larger number, so it never caused unsigned underflow. We definitely forgot to consider signed underflow, but I don't think there are other edge cases in the unsigned case.

kripken: Yeah, I guess you're right, I was being overly cautious. Done. I also can't think of another possible overflow issue here (but would not want to bet on it..)

kripken: friendly ping.

src/passes/OptimizeInstructions.cpp
```diff
@@ -3994,26 +4010,40 @@ struct OptimizeInstructions
...
+          // smaller from the larger, which avoids unsigned overflows, but if
+          // the comparison is signed we need to also consider the sign bit.
+          // E.g. 0x80000000 - 1 = 0x7fffffff is fine as unsigned, but as a
+          // signed number the -1 turns a negative number into a positive one.
+          bool constsOverflow = false;
+          if (isSignedOp(curr->op)) {
+            auto typeMaxBits = getBitsForType(add->type);
+            auto c1Bits = Bits::getMaxBits(c1, this);
+            auto c2Bits = Bits::getMaxBits(c2, this);
+            if (c1Bits + c2Bits >= typeMaxBits) {
```
tlively, line 4026: Why is this the correct condition to check? It seems that problems only arise when the greater unsigned C can be at least INT_MIN and the subtraction would take it below INT_MIN, i.e. its signed interpretation undeflows from negative to positive. Notably, if the greater unsigned C starts out less than INT_MIN, there can never be a problem since we know that the subtraction never wraps across zero.

Actually, since C1 and C2 are known constants, we can just directly check whether the subtraction would wrap across INT_MIN and don't need to generalize to consider max bits at all.

And if the subtraction C1 - C2 would wrap across INT_MIN, then we know that C1 is negative and C2 is positive, so we could have taken C2 to be our larger value and used the C2 - C1 code path instead.

We know the constant values and we know whether to interpret them as signed or unsigned, so we should always be able to choose the substraction that avoids the kind of underflow we care about avoiding.

kripken: Hmm, I'm not sure. If we always subtract the smaller from the larger (using the proper sign, and not the unsigned sign as we did before this PR), then given `0, 0x80000000` then for the signed case we'd be doing `0 - 0x80000000` which overflows for another reason: `0x80000000` is the weird nonzero value for whom `x == -x = 0 - x`; and in particular `0 - 0x7fffffff` is positive while `0 - 0x80000000` is negative, so this is nonlinear (I'm not sure if that's what you mean by "wraps across INT_MIN" or not?). And, specifically for the rule here,
```
;; this is always false
x + 0x80000000 > 0
;; apply the rule
x > 0 - 0x80000000 = 0x80000000
;; this is sometimes true, so we broke something
```

I spent some time thinking on this but just don't think it's worth trying to handle all these cases given the complexity and low benefit. Unless I'm missing a simple solution? Otherwise, I think it's fine if we don't optimize all the things; just handling the common cases is enough, which I think this PR does.

tlively: Sure, 0x80000000 is an edge case to consider. My high level point is that we have the constant values, so doing exhaustive case analysis is feasible and wouldn't be that complicated.

But being more conservative sounds fine as well. In that case, where does the idea to compare `c1Bits + c2Bits >= typeMaxBits` come from? It seems correct, but also somewhat arbitrary and not much simpler or obviously correct than the precise analysis.

tlively: Another conservative approach would be to keep the existing logic but restrict the optimization to only work on unsigned comparisons.

kripken: My idea was that if the total sum of bits can't even add up to the overflow area, then we're ok. But actually I guess that's wrong, as any signed number has max bits of the maximum...

I think it would be nice to also get small signed values here. If I can't figure out a good way though then I agree focusing on unsigned ones is enough.

kripken:
Ok, this is rewritten and should be ready for review now. The code has been generalized to check for nonlinearity in the subtraction of the constants.

kripken:
ping @tlively , perhaps we can get this in before the end of the year? :smile:

tlively (approved):
Looks great!

PR #5241, opened 2022-11-11
Handles memory.grow failure in MultiMemoryLowering Pass

ashleynh:
Per the [wasm spec](https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-instr-memory), memory.grow instructions should return -1 when there is a failure to allocate enough memory. This PR adds support for returning this error code.

### DIFF
--- a/src/passes/MultiMemoryLowering.cpp
+++ b/src/passes/MultiMemoryLowering.cpp
@@ -304,12 +304,12 @@ struct MultiMemoryLowering : public Pass {
           sizeLocal, builder.makeMemorySize(combinedMemory, memoryInfo)));
     }
 
-    // TODO: Check the result of makeMemoryGrow for errors and return the error
-    // instead
+    // Attempt to grow the combinedMemory. If -1 returns, enough memory could
+    // not be allocated, so return -1.
     functionBody = builder.blockify(
       functionBody,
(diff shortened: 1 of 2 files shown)


src/passes/MultiMemoryLowering.cpp
```diff
@@ -304,12 +304,18 @@ struct MultiMemoryLowering : public Pass {
...
-      builder.makeDrop(builder.makeMemoryGrow(
-        builder.makeLocalGet(0, pointerType), combinedMemory, memoryInfo)));
+      builder.makeIf(
+        builder.makeBinary(
+          EqInt32,
+          builder.makeMemoryGrow(
+            builder.makeLocalGet(0, pointerType), combinedMemory, memoryInfo),
+          builder.makeConst(-1)),
+        builder.makeReturn(builder.makeConst(-1)),
+        nullptr));
```
kripken, line 318:
```suggestion
        builder.makeReturn(builder.makeConst(-1))));
```
`makeIf` has a default value of `nullptr` there, so you can skip passing it,

https://github.com/WebAssembly/binaryen/blob/38c152e0d8bc593c2861c251cc2a875cfb1a21f0/src/wasm-builder.h#L224-L226

kripken (commented):
Code lgtm. I think this should show up in test output too after updating?

PR #5242, opened 2022-11-11
Fix two fuzz bugs with ArrayNewSeg

### LINKED ISSUE #5236: array.new_data fuzz error

tlively:
First, we forgot to note the type annotation on `ArrayNewSeg` instructions, so
in small modules where these are the only annotated instructions, the type
section would be incomplete.

Second, in the interpreter we were reserving space for the array before checking
that the segment access was valid. This could cause huge allocations that threw
bad_alloc exceptions before the interpreter could get around to trapping. Fix
the problem by reserving the array after validating the arguements.

Fixes #5236.


tlively:
* main
  * **PR #5242** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5242" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5242?utm_source=stack-comment).

kripken (approved):

PR #5243, opened 2022-11-11
[Wasm GC] Fix nondeterminism in GUFA due to ordering

kripken:
We don't actually have the transitive and distributive properties since our
PossibleContents representation is an approximation, and the fuzzer found
a case where that is noticeable. See more details in the new comment +
testcase.

I measured speed and memory usage and this actually causes almost no
noticeable change.

### DIFF
--- a/src/ir/possible-contents.cpp
+++ b/src/ir/possible-contents.cpp
@@ -1341,7 +1341,8 @@ private:
 #ifdef POSSIBLE_CONTENTS_INSERT_ORDERED
   InsertOrderedSet<LocationIndex> workQueue;
 #else
-  std::unordered_set<LocationIndex> workQueue;
+  //std::unordered_set<LocationIndex> workQueue;
+  UniqueDeferredQueue<LocationIndex> workQueue;
 #endif
 
   // All existing links in the graph. We keep this to know when a link we want
@@ -1634,9 +1635,7 @@ Flower::Flower(Module& wasm) : wasm(wasm) {
     }
 #endif
 
-    auto iter = workQueue.begin();
-    auto locationIndex = *iter;
-    workQueue.erase(iter);
+    auto locationIndex = workQueue.pop();
 
     flowAfterUpdate(locationIndex);
   }
@@ -1734,7 +1733,7 @@ bool Flower::updateContents(LocationIndex locationIndex,
 #endif
 
   // Add a work item if there isn't already.
-  workQueue.insert(locationIndex);
+  workQueue.push(locationIndex);
 
   return worthSendingMore;
 }
--- a/src/ir/possible-contents.cpp
+++ b/src/ir/possible-contents.cpp
@@ -22,13 +22,8 @@
 #include "ir/local-graph.h"
 #include "ir/module-utils.h"
 #include "ir/possible-contents.h"
-#include "wasm.h"
-
-#ifdef POSSIBLE_CONTENTS_INSERT_ORDERED
-// Use an insert-ordered set for easier debugging with deterministic queue
-// ordering.
 #include "support/insert_ordered.h"
-#endif
+#include "wasm.h"
 
 namespace std {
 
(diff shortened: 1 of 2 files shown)


src/ir/possible-contents.cpp
```diff
@@ -1338,11 +1333,19 @@ struct Flower {
...
   // their work item in the queue) will see the newer value, possibly avoiding
   // flowing an old value that would later be overwritten.
-#ifdef POSSIBLE_CONTENTS_INSERT_ORDERED
+  //
+  // This must be ordered to avoid nondeterminism. The problem is that our
+  // operations are imprecise and so the transitive property does not hold:
+  // (AvB)vC may differ from Av(BvC). Likewise (AvB)^C may differ from
+  // (A^C)v(B^C). An example of the latter is if a location is sent a null func
+  // and an i31, and the location can only contain funcref. If the null func
+  // arrives first, then later we'd merge null func + i31 which ends up as Many,
```
tlively, line 1342: i31 and `func` are on different type hierarchies so how could they possibly arrive at the same place?

It seems to me that the principled fix for this problem would be to fix our operations to have the algebraic properties we would expect, not to make the control flow deterministic. Is there a reason we can't or shouldn't do that?

test/lit/passes/gufa-refs.wast
```diff
@@ -358,6 +360,50 @@
...
+    ;;    a cone of funcref. We cannot optimize in that case, unlike before.
+    ;;
+    ;; Ideally we'd optimize here, but atm we do not since the order in
+    ;; practice is a less ideal one. At minimum we should be deterministic in
+    ;; how we handle this, which this test enforces at least.
+    ;;
+    ;; TODO: Find a way to actually optimize such cases, perhaps by filtering
+    ;;       when sending as well and not just when receiving (the br_on_func
+    ;;       here should not send anything, as what it sends should be first
+    ;;       intersected with funcref).
```
tlively, line 395: This TODO seems like the better solution. Would it be complicated to implement?

tlively (commented):
It seems to me that the principled fix for this problem would be to fix our operations to have the algebraic properties we would expect, not to make the control flow deterministic. Is there a reason we can't or shouldn't do that?

One thing I've been thinking about for the abstract interpretation internship is writing a general fuzzer to test the algebraic properties of lattices to build confidence taht at lattice implementation actually implements a lattice. That might be worth doing sooner rather than later.

kripken:
> Is there a reason we can't or shouldn't do that?

We have approximations to mathematical sets here. For example merging exact types A and B leads to a cone type that contains the subtypes of A and B rather than a union of exactly A and B. We can't really avoid such approximations since we have to keep our runtime reasonable. The example in the code and test is
```
(i31 ^ funcref) v (func.foo ^ funcref) = nothing v func.foo = func.foo
but
(i31 v func.foo) ^ funcref = many ^ funcref = funcref
```
which breaks the distributive property. (If, instead, `many` were a precise union of two things, this would not happen.)

test/lit/passes/gufa-refs.wast
kripken: It would add work, but even so it does not address the underlying issue I think (see below).

tlively:
Ah, looking a little more closely at the wikipedia page for lattices, `v` and `^` are supposed to be commutative and associative, but not necessarily distributive.

test/lit/passes/gufa-refs.wast
tlively: Wouldn't this address the underlying issue by making it so that GUFA never has to take any intersections after taking unions, so it would no longer depend on distributivity to make the results deterministic for any iteration order?

kripken:
Even if we found a way to fix the current specific bug by avoiding depending on the distributive property, I don't think that's the best path forward. Any custom filtering we add will need to be careful to avoid distributivity (is that a word?), and transitivity, etc. Life is quite hard when you don't have the basic math assumptions we normally expect :smile: but as I've said, we just can't have those assumptions here, the cost is too high. (Kind of how CPU integers aren't real integers, I suppose... so overflows break stuff.)

Making the flow deterministic as this PR does is a trivial change but it avoids a whole family of possible errors, and luckily it does so without noticeable runtime cost.

tlively (approved):
Code change LGTM, but I still want to understand whether we otherwise fundamentally depend on the distributive property for determinism or if that's just a quirk of the current implementation.

test/lit/passes/gufa-refs.wast
tlively: I still don't understand why this wouldn't solve the underlying issue.

kripken: Well, again, it can solve the current issue. But there is a fundamental problem when a system doesn't have nice things like distributivity and associativity (sorry if those aren't the right terms in English, but you know what I mean). It's very easy for us to implicitly depend on them later and cause nondeterminism.

I think the only true solutions for the underlying issue are:

1. Get those nice properties. But that's impossibly slow.
2. Order the flow so even without those properties we are fully deterministic.

kripken:
I'd say there's a chance distributivity is the only current issue. But I'd need to read all the rest of the code to check if we don't depend on it elsewhere. (I'd also want to see proofs of associativity - I think we might have that property, but I'm not sure...)

test/lit/passes/gufa-refs.wast
tlively: We should have commutativity and associativity (and if we don't, that can be treated as a bug by itself), and I believe that should be enough to make the output independent of the iteration order if we implement this TODO since the rest of the system would never depend on distributivity, which is the only property we don't have.

tlively: Acknowledged that this belief would be difficult to verify, though 👍

kripken: Yeah, I sort of think the same, but it's just hard to verify. That this even depends on distributivity is itself kind of subtle - that's only apparent when comparing multiple possible execution paths.

Doing this TODO might make sense. But I'd need to see if it doesn't slow us down. We'd be doing 2 intersections + one merge in each flow iteration instead of 1 of each, and adding the first intersection when I did that definitely added overhead, unfortunately (but it was necessary).

PR #5251, opened 2022-11-14
Fix arithmetic in interpretation of ArrayNewSeg

### LINKED ISSUE #5249: Fuzz error on array.new_elem:   what():  vector::reserve
```wat
(module
 (type $[funcref] (array_subtype funcref data))
 (type $none_=>_ref|[funcref]| (func_subtype (result (ref $[funcref])) func))
 (elem func)
 (export "new" (func $0))
 (func $0 (type $none_=>_ref|[funcref]|) (result (ref $[funcref]))
  (array.new_elem $[funcref] 0
   (i32.const 3)
   (i32.const -3)
  )
 )
)
```

```
$ bin/wasm-opt a.wat -all --nominal --fuzz-exec
[fuzz-exec] calling new
terminate called after throwing an instance of 'std::length_error'
  what():  vector::reserve
Aborted
```

tlively:
The offset and size were previously being sign extended from 32 to 64 bits,
which meant that negative sizes could make the bounds check pass and cause an
exception to be thrown by an overly large allocation. Switch to using uint64_t
from the start rather than mixing sizes and signs, and update the tests to
reproduce the error more robustly in the absence of the fix.

Fixes #5249.

### DIFF
--- a/src/passes/RemoveUnusedModuleElements.cpp
+++ b/src/passes/RemoveUnusedModuleElements.cpp
@@ -364,8 +364,7 @@ struct RemoveUnusedModuleElements : public Pass {
                ModuleElement(ModuleElementKind::Tag, curr->name)) == 0;
     });
     module->removeElementSegments([&](ElementSegment* curr) {
-      return curr->data.empty() ||
-             analyzer.reachable.count(ModuleElement(
+      return analyzer.reachable.count(ModuleElement(
                ModuleElementKind::ElementSegment, curr->name)) == 0;
     });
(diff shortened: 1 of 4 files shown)


tlively:
* main
  * **PR #5251** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5251" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5251?utm_source=stack-comment).

kripken (approved):

PR #5253, opened 2022-11-14
GlobalStructInference: Handle cases with just 1 global too

kripken:
Expand GlobalStructInference to operate on cases with a single possible global, and
not just 2 or more. Even the case of a single global is useful, it turns out, as we can
alter the reference in places like this:
```wat
(struct.get $type 0
  (..ref..)
)
```
No matter what `ref` is, if there is a single global it must refer to, we can switch to
this:
```wat
(struct.get $type 0
  (global.get $global)
)
```
That can unlock further opts later. Note that we can do this even if we don't know
what the value actually is - we may not know what the `struct.get` returns, but we
do know what it reads from.


test/lit/passes/gsi_vacuum_precompute.wast
```diff
@@ -0,0 +1,59 @@
...
+;;  * --gsi : Infers that a reference to a particular itable type must be a
+;;            global.get of the single itable defined using that type (since no
+;;            other object of that type is ever created).
+;;  * --vacuum : Cleans up some dropped stuff to enable the subsequent pass.
+;;  * --precompute : Represents immutable globals in a way that we can start
+;;                   from a global.get of an itable, get a vtable, and get a
+;;                   field in the vtable, and we end up optimizing to produce
+;;                   the final value in that vtable.
+
+(module
```
tlively, line 20: What happened to the test output?

tlively (approved):

PR #5256, opened 2022-11-15
Adds bounds checks to Load/Store in Multi-Memories Lowering Pass

ashleynh:
Following the wasm spec guidelines for [Load](https://webassembly.github.io/spec/core/exec/instructions.html#t-mathsf-xref-syntax-instructions-syntax-instr-memory-mathsf-store-xref-syntax-instructions-syntax-memarg-mathit-memarg-and-t-mathsf-xref-syntax-instructions-syntax-instr-memory-mathsf-store-n-xref-syntax-instructions-syntax-memarg-mathit-memarg) (rule 10) & [Store](https://webassembly.github.io/spec/core/exec/instructions.html#t-mathsf-xref-syntax-instructions-syntax-instr-memory-mathsf-store-xref-syntax-instructions-syntax-memarg-mathit-memarg-and-t-mathsf-xref-syntax-instructions-syntax-instr-memory-mathsf-store-n-xref-syntax-instructions-syntax-memarg-mathit-memarg) (rule 12), this PR adds a runtime error if the instruction exceeds the bounds of the particular memory within the combined memory.


src/passes/MultiMemoryLowering.cpp
```diff
@@ -118,35 +118,76 @@ struct MultiMemoryLowering : public Pass {
...
+              Abstract::getBinary(parent.pointerType, Abstract::Add),
+              builder.makeBinary(
+                Abstract::getBinary(parent.pointerType, Abstract::Add),
+                builder.makeLocalGet(ptrIdx, parent.pointerType),
+                builder.makeConstPtr(curr->offset, parent.pointerType)),
+              builder.makeConstPtr(curr->bytes, parent.pointerType)),
+            builder.makeCall(
+              parent.memorySizeNames[idx], {}, parent.pointerType)),
+          builder.makeUnreachable());
+        replacement = builder.makeBlock({ptrSet, boundsCheck, curr});
```
kripken, line 189: Perhaps we could have a `makeBoundsCheck` helper function that is called from both places, so we share the code?

src/passes/MultiMemoryLowering.cpp
```diff
@@ -118,35 +118,76 @@ struct MultiMemoryLowering : public Pass {
...
+            Abstract::getBinary(parent.pointerType, Abstract::GtU),
+            builder.makeBinary(
+              // ptr + offset (ea from wasm spec) + bit width
+              // two builder Adds, we'll add the first two operands in the first
+              // add and then add the third operand in the second add
+              Abstract::getBinary(parent.pointerType, Abstract::Add),
+              builder.makeBinary(
+                Abstract::getBinary(parent.pointerType, Abstract::Add),
+                builder.makeLocalGet(ptrIdx, parent.pointerType),
+                builder.makeConstPtr(curr->offset, parent.pointerType)),
```
kripken, line 184: Probably worth a comment that the offset computation is not precise according to the spec. In the spec offsets do not overflow as twos-complement, but `i32.add` does. Concretely, a load from address `1000` with offset `0xffffffff` should actually trap, as the combined number is greater than 32 bits. But with an add, `1000 + 0xffffffff = 999` due to overflow, which would not trap.

In theory we could compute this like the spec, by expanding the i32s to i64s and adding there (where we won't overflow). But I think for now a comment is enough, since it's not that important to keep trapping programs trapping (as opposed to working programs working).

tlively (commented):
Two high-level comments:

 1. Accesses to memory zero currently do not get bounds checks, but they probably should to be consistent with accesses to the other memories.
 2. It would be good to make bounds checks optional, since I expect they would have significant performance overhead. Maybe we can have two versions of the pass registered, one that does checks and another that does not?

kripken (approved):
Great!

PR #5260, opened 2022-11-15
Fix an unused var warning in some compilers

kripken:


src/wasm/wat-parser.cpp
```diff
@@ -1881,6 +1881,7 @@ struct ParseDefsCtx : TypeParserCtx<ParseDefsCtx> {
   Result<> makeStructSet(Index pos, HeapType type, Index field) {
     assert(type.isStruct());
     const auto& fields = type.getStruct().fields;
+    WASM_UNUSED(fields);
     assert(fields.size() > field);
```
tlively, line 1885: Since `fields` isn't otherwise used, lets fix this by doing

```
assert(type.getStruct().fields.size() > field);
``

and getting rid of the variable entirely.

tlively (approved):

PR #5262, opened 2022-11-16
[Wasm GC] Fix a GUFA bug on null call_ref targets

kripken:
If the target is a bottom type then it is a heap type but it is not a signature
type, and we should treat it as unreachable (and not crash).


test/lit/passes/monomorphize_gufa.wast
```diff
@@ -0,0 +1,53 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
+
+;; RUN: foreach %s %t wasm-opt --nominal --monomorphize-always --gufa -all -S -o - | filecheck %s
+
+;; Monomorphize will generate a version of the called function with a refined
+;; target type, to a null type. Gufa should not crash on that. (We must run both
+;; passes in sequence, as either text or binary parsing will turn a call_ref of
+;; null into an unreachable, which would avoid the issue.)
```
tlively, line 8: This is no longer true since we added type annotations to `call_ref`, so you should be able to write this test with just `--gufa` now.

tlively: You can use an arbitrary signature type as the annotation:

```(call_ref $some_valid_signature (ref.null nofunc))`

Unless the target type is important for this test? I thought it was only the target that mattered.

kripken: Ah, right, I thought I need to set the annotation, but I just need the ref value to be `nofunc`, thanks. Done.

tlively (approved):

PR #5263, opened 2022-11-16
[Wasm GC] Start an OptimizeCasts pass and reuse cast values there

kripken:
```wat
(some.operation
  (ref.cast .. (local.get $ref))
  (local.get $ref)
)

;; => ;;

(some.operation
  (local.tee $temp
    (ref.cast .. (local.get $ref))
  )
  (local.get $temp)
)
```

This can help cases where we cast for some reason but happen to not use the
cast value in all places. This occurs in j2wasm in itable calls sometimes: The
`this` pointer is is refined, but the itable may be done with an unrefined pointer,
which is less optimizable.

So far this is just inside basic blocks, but that is enough for the cast of itable
calls and other common patterns I see.


src/passes/OptimizeCasts.cpp
```diff
@@ -0,0 +1,219 @@
...
+namespace {
+
+// Find the best sources for local.gets: other locals with the same value, but
+// cast to a more refined type.
+struct BestSourceFinder : public LinearExecutionWalker<BestSourceFinder> {
+
+  PassOptions options;
+
+  // A map of the best local.get for a particular index: the local.get that has
+  // the most refined type.
```
tlively, line 98:
```suggestion
  // Map local indices to the most refined downcastings of local.gets from those indices.
```

src/passes/OptimizeCasts.cpp
```diff
@@ -0,0 +1,219 @@
...
+  std::unordered_map<Index, Expression*> bestSourceForIndexMap;
+
+  // For each expression, a vector of local.gets that would like to use its
+  // value instead of themselves (as it is more refined).
```
tlively, line 102:
```suggestion
  // For each most-downcasted local.get, a vector of other local.gets that could be replaced with gets of the downcasted value.
```

src/passes/OptimizeCasts.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+// Find the best sources for local.gets: other locals with the same value, but
+// cast to a more refined type.
+struct BestSourceFinder : public LinearExecutionWalker<BestSourceFinder> {
+
+  PassOptions options;
+
+  // A map of the best local.get for a particular index: the local.get that has
+  // the most refined type.
+  std::unordered_map<Index, Expression*> bestSourceForIndexMap;
```
tlively, line 99:
```suggestion
  std::unordered_map<Index, Expression*> mostCastedGets;
```

src/passes/OptimizeCasts.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+  // For each expression, a vector of local.gets that would like to use its
+  // value instead of themselves (as it is more refined).
+  std::unordered_map<Expression*, std::vector<LocalGet*>> requestMap;
```
tlively, line 103:
```suggestion
  std::unordered_map<Expression*, std::vector<LocalGet*>> lessCastedGets;
```

src/passes/OptimizeCasts.cpp
```diff
@@ -0,0 +1,219 @@
...
+  static void doNoteNonLinear(BestSourceFinder* self, Expression** currp) {
+    self->bestSourceForIndexMap.clear();
+  }
+
+  void visitLocalSet(LocalSet* curr) {
+    // Clear any information about this local; it has a new value here.
+    bestSourceForIndexMap.erase(curr->index);
+  }
+
+  void visitLocalGet(LocalGet* curr) {
```
tlively, line 114: I would expect that we could also look at `LocalSet` here as long as we were careful about not looking through interfering tees. Maybe worth a TODO?

tlively (commented):
Nice! I suggested some comment and name changes that I think make things clearer, but other than that, LGTM.

kripken:
Thanks for the feedback! Applied, plus some related renamings and comments.

src/passes/OptimizeCasts.cpp
```diff
@@ -88,37 +92,43 @@ namespace wasm {
...
+  // Map local indices to the most refined downcastings of local.gets from those
+  // indices.
+  //
+  // This is tracked in each basic block, and cleared between them.
+  std::unordered_map<Index, Expression*> mostCastedGets;
+
+  // For each most-downcasted local.get, a vector of other local.gets that could
+  // be replaced with gets of the downcasted value.
+  //
+  // This is tracked til the end of the entire function, and contains the
```
tlively, line 110:
```suggestion
  // This is tracked until the end of the entire function, and contains the
```

`'til` is an abbreviation of `until` (note the apostrophe), but not a word on its own.

src/passes/OptimizeCasts.cpp
```diff
@@ -88,37 +92,43 @@ namespace wasm {
 
 namespace {
 
-// Find the best sources for local.gets: other locals with the same value, but
-// cast to a more refined type.
-struct BestSourceFinder : public LinearExecutionWalker<BestSourceFinder> {
+// Find the best casted verisons of local.gets: other locals with the same
```
tlively, line 95:
```suggestion
// Find the best casted verisons of local.gets: other local.gets with the same
```

tlively (approved):
Looks good!

src/passes/OptimizeCasts.cpp
kripken: Thanks, good to know...

PR #5266, opened 2022-11-16
Reland "Make `call_ref` type annotations mandatory"

tlively:
Reverts WebAssembly/binaryen#5265

It turns out that this revert was due to a false alarm.


kripken (approved):
re-lgtm

PR #5273, opened 2022-11-17
Fix warnings that make build fail with -Werror,-Wheader-hygiene,-Wimplicit-const-int-float-conversion

mkustermann:
When `-Wheader-hygiene` is enabled, C compiler will warn when using namespace directive in global context in header file.

When `-Wimplicit-const-int-float-conversion` is enabled C compiler will warn on implicit integer to double conversions that change values.


kripken (approved):
Thanks!

PR #5274, opened 2022-11-17
Add post-emscripten-side-module pass argument

sbc100:
In this mode we don't remove the start/stop_em_asm symbols or data. This is because with side modules we read this information directly from the wasm binaryen at runtime.

See https://github.com/emscripten-core/emscripten/pull/18228

### DIFF
--- a/src/passes/PostEmscripten.cpp
+++ b/src/passes/PostEmscripten.cpp
@@ -214,12 +214,19 @@ struct PostEmscripten : public Pass {
     std::vector<Address> segmentOffsets; // segment index => address offset
     calcSegmentOffsets(module, segmentOffsets);
 
-    removeData(module, segmentOffsets, "__start_em_asm", "__stop_em_asm");
+    auto& options = getPassOptions();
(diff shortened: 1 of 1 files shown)


src/passes/Asyncify.cpp
```diff
@@ -1536,8 +1536,7 @@ struct Asyncify : public Pass {
     String::Split listedImports(stateChangingImports, ",");
     // TODO: consider renaming asyncify-ignore-indirect to
     //       asyncify-ignore-nondirect, but that could break users.
-    auto ignoreNonDirect =
-      options.getArgumentOrDefault("asyncify-ignore-indirect", "") == "";
```
sbc100, line 1540: This line seems wrong... aren't we setting ignoreNonDirect in cases where asyncify-ignore-indirect is not set?

kripken: Good point, this does look wrong... We do have tests for it, and this PR is flipping the value, so we should see an effect from that which could help explain what's going on.

kripken (commented):
lgtm aside from that Asyncify issue. Let's see what happens with tests.

PR #5275, opened 2022-11-17
Fix inverted logic bug with asyncify-ignore-indirect

sbc100:


sbc100:
I'm a little confused by the resulting test expectation changes.. @kripken since you know this code better than me to know what is going on here?   Why does the logic appear inverted here?

sbc100:
Oh I figured it out.. it was the variable name that was backwards, not the logic.

kripken (approved):
Thanks!

Maybe we renamed the variable in ModuleAnalyzer at some point but not the local in the main pass..? Not sure how we ended up in the weird state before this PR...

PR #5240, opened 2022-11-11
Remove equirecursive typing

tlively:
Equirecursive is no longer standards track and its implementation is extremely
complex. Remove it.

PR #5279, opened 2022-11-18
Validate heap types

tlively:
Validate that GC is enabled if struct and array types are used by the module and
validate that both GC and isorecursive types are enabled when nontrivial rec
groups are used.

This fixes a fuzz bug in #5239 where initial contents included a rec group but
the fuzzer disabled GC. Since the resulting module passed validation, the rec
groups made it into the binary output, making the type section malformed.

### DIFF
--- a/src/wasm/wasm-validator.cpp
+++ b/src/wasm/wasm-validator.cpp
@@ -3461,6 +3461,25 @@ static void validateTags(Module& module, ValidationInfo& info) {
   }
 }
 
+static void validateTypes(Module& module, ValidationInfo& info) {
+  for (auto type : ModuleUtils::collectHeapTypes(module)) {
+    if (type.getRecGroup().size() > 1) {
+      info.shouldBeTrue(module.features.hasGC() &&
+                          getTypeSystem() == TypeSystem::Isorecursive,
+                        type,
+                        "Recursion groups require GC [--enable-gc] and "
+                        "isorecursive types [--hybrid]");
+    }
+
+    if (!module.features.hasGC()) {
+      info.shouldBeTrue(
+        !type.isStruct(), type, "Struct types require GC [--enable-gc]");
+      info.shouldBeTrue(
+        !type.isArray(), type, "Array types require GC [--enable-gc]");
+    }
+  }
+}
+
 static void validateModule(Module& module, ValidationInfo& info) {
   // start
   if (module.start.is()) {
@@ -3505,6 +3524,7 @@ bool WasmValidator::validate(Module& module, Flags flags) {
     validateDataSegments(module, info);
     validateTables(module, info);
     validateTags(module, info);
+    validateTypes(module, info);
     validateModule(module, info);
     validateFeatures(module, info);
   }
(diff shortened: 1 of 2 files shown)


tlively:
* main
  * **PR #5279** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5279" target="_blank">Graphite</a>  👈
    * **PR #5239** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5239" target="_blank">Graphite</a> 
      * **PR #5240** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5240" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5279?utm_source=stack-comment).

src/wasm/wasm-validator.cpp
```diff
@@ -3461,6 +3461,25 @@ static void validateTags(Module& module, ValidationInfo& info) {
   }
 }
 
+static void validateTypes(Module& module, ValidationInfo& info) {
+  for (auto type : ModuleUtils::collectHeapTypes(module)) {
```
kripken, line 3465: I am somewhat worried about scanning the entire module for this validation - it will basically double validation times, I think.

We already validate GC instructions, which are most of the places GC types can appear. Perhaps we can add validation in the other places, like function params and locals?

tlively: Yeah, maybe. I'd be worried about missing places and duplicating the logic from `ModuleUtils::collectHeapTypes` (which we've had trouble remembering to update in just a single place), but for better or worse maybe the fuzzer will help find cases where we miss adding validation.

kripken: Yeah, definitely a tradeoff. Not sure what's best.

Can we at least not do the extra pass when GC is not enabled? That is, if GC is not enabled then I'd expect to get lots of other validation errors if the code contains GC types. And the new check added here seems to mostly be important to differentiate things in the context of GC being enabled anyhow (nominal/isorecursive differences).

tlively: Sure, that seems reasonable for now.

tlively: I pushed a better solution just now. PTAL.

src/wasm/wasm-validator.cpp
```diff
@@ -2903,6 +2903,9 @@ void FunctionValidator::visitFunction(Function* curr) {
                  "Multivalue function results (multivalue is not enabled)");
   }
   FeatureSet features;
+  // Check for things like having a rec group with GC enabled.
+  features |=
```
kripken, line 2907: Why are reference types ignored here?

kripken (approved):
Looks great aside from that question!

src/wasm/wasm-validator.cpp
tlively: Because a function reference type always requires `ReferenceTypes` as a baseline, but it's kind of a false positive in this situation. MVP functions will have some MVP signature type here, but `getFeatures()` will still return `ReferenceTypes`.

In practice the only thing this line checks for is rec groups.

kripken: Ah, right, the type itself is a reference type. A comment perhaps could help?

kripken (approved):

src/wasm/wasm-validator.cpp
tlively: Oh oops, I saw this comment just after merging. Will add a comment in a follow-up.

PR #5284, opened 2022-11-21
Code Pushing: Ignore unreachable sets

kripken:
Normally we ignore them anyhow (unreachability is an effect, either a trap or
a control flow switch), but in traps-never-happen mode we can ignore a trap, so
we need to check this manually.

Found by the fuzzer over the weekend after 500K iterations...

### DIFF
--- a/src/passes/CodePushing.cpp
+++ b/src/passes/CodePushing.cpp
@@ -323,6 +323,26 @@ private:
       assert(i > 0);
       i--;
       auto* pushable = isPushable(list[i]);
+      if (pushable && pushable->type == Type::unreachable) {
+        // Don't try to push something unreachable. If we did, then we'd need to
+        // refinalize the block we are moving it from:
+        //
+        //  (block $unreachable
+        //    (local.set $x (unreachable))
+        //    (if (..)
+        //      (.. (local.get $x))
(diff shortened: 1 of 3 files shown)


src/passes/CodePushing.cpp
```diff
@@ -323,6 +323,26 @@ class Pusher {
       assert(i > 0);
       i--;
       auto* pushable = isPushable(list[i]);
+      if (pushable && pushable->type == Type::unreachable) {
+        // Don't try to push something unreachable. If we did, then we'd need to
+        // refinalize the block we are moving it from:
```
aheejin, line 328: Is having to refinalize a problem? Are there any other cases in this pass that we need to refinalize?

tlively:
500K iterations 🤯

src/passes/CodePushing.cpp
kripken: It's possible to refinalize. In this pass though we don't even include the header for it - nothing else required it - so it would definitely be more work to add refinalizing logic than not (specially if we try to refinalize only when necessary, to be more efficient). And in general leaving work for DCE is reasonable, I think.

aheejin (approved):

Issue #4822, opened 2022-07-22
Optimize Asyncify to not flatten unnecessarily


kripken:
We could only flatten when we see we actually need to instrument a function. If most functions don't need instrumentation then it would avoid most of the build time overhead (but it just saves build work - it would not help runtime). In general that's not the case - indirect calls in a big project tend to mean everything needs instrumentation - but when it is true it could be a nice speedup.

To do this we'd need to refactor the pass quite a bit, though, to not just run the Flatten pass and the others immediately. We'd need to identify the functions, either by scanning them first or as we go. And we'd need to run all the followup passes only in the case that we do flatten.

kripken:
(Aside from flattening, we could avoid running the optimization pipeline as well. That is, we should do zero work on functions we do not instrument.)

PR #5293, opened 2022-11-25
#4822 optmize asyncify to not flatten/optimize unnecessarily

caiiiycuk:
I tried to do what is described in #4822. I'm not sure if everything is correct, but it seems to work for my case. Compilation time was reduced from 14 minutes to 3.5 minutes and all my tests passed.

It's more like proposal, if idea is correct I can refactor as needed.

--
Related to [discussion](https://groups.google.com/g/emscripten-discuss/c/tpKmW0jLlmw/m/pFB9DSEGBQAJ?utm_medium=email&utm_source=footer)

### DIFF
--- a/src/pass.h
+++ b/src/pass.h
@@ -251,6 +251,8 @@ struct PassRunner {
   PassRunner(const PassRunner&) = delete;
   PassRunner& operator=(const PassRunner&) = delete;
 
+  virtual ~PassRunner() = default;
+
   // But we can make it easy to create a nested runner
   // TODO: Go through and use this in more places
   explicit PassRunner(const PassRunner* runner)
@@ -333,6 +335,9 @@ struct PassRunner {
   // Returns whether a pass by that name will remove debug info.
   static bool passRemovesDebugInfo(const std::string& name);
 
+protected:
+  virtual void doAdd(std::unique_ptr<Pass> pass);
+
 private:
   // Whether this is a nested pass runner.
   bool isNested = false;
@@ -343,8 +348,6 @@ private:
 
   // Whether this pass runner has run. A pass runner should only be run once.
   bool ran = false;
-
-  void doAdd(std::unique_ptr<Pass> pass);
 
   void runPass(Pass* pass);
   void runPassOnFunction(Pass* pass, Function* func);
@@ -426,8 +429,8 @@ public:
 
   std::string name;
 
-  PassRunner* getPassRunner() { return runner; }
-  void setPassRunner(PassRunner* runner_) {
+  virtual PassRunner* getPassRunner() { return runner; }
+  virtual void setPassRunner(PassRunner* runner_) {
     assert((!runner || runner == runner_) && "Pass already had a runner");
     runner = runner_;
   }
--- a/src/pass.h
+++ b/src/pass.h
@@ -429,8 +429,8 @@ public:
 
   std::string name;
 
-  virtual PassRunner* getPassRunner() { return runner; }
-  virtual void setPassRunner(PassRunner* runner_) {
+  PassRunner* getPassRunner() { return runner; }
+  void setPassRunner(PassRunner* runner_) {
     assert((!runner || runner == runner_) && "Pass already had a runner");
     runner = runner_;
   }
--- a/test/lit/passes/asyncify_enable-multivalue.wast
+++ b/test/lit/passes/asyncify_enable-multivalue.wast
@@ -37,18 +37,10 @@
   ;; CHECK:      (export "asyncify_get_state" (func $asyncify_get_state))
 
   ;; CHECK:      (func $do_sleep
-  ;; CHECK-NEXT:  (local $0 i32)
-  ;; CHECK-NEXT:  (local $1 i32)
-  ;; CHECK-NEXT:  (local.set $0
-  ;; CHECK-NEXT:   (global.get $sleeping)
-  ;; CHECK-NEXT:  )
-  ;; CHECK-NEXT:  (local.set $1
-  ;; CHECK-NEXT:   (i32.eqz
-  ;; CHECK-NEXT:    (local.get $0)
-  ;; CHECK-NEXT:   )
-  ;; CHECK-NEXT:  )
   ;; CHECK-NEXT:  (if
-  ;; CHECK-NEXT:   (local.get $1)
+  ;; CHECK-NEXT:   (i32.eqz
+  ;; CHECK-NEXT:    (global.get $sleeping)
+  ;; CHECK-NEXT:   )
   ;; CHECK-NEXT:   (block
   ;; CHECK-NEXT:    (global.set $sleeping
   ;; CHECK-NEXT:     (i32.const 1)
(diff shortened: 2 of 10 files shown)


caiiiycuk:
> Maybe there is a way to "proxy" an entire pass runner.

I also thought about it, but PassRunner is not virtual struct. Don't have idea how to do this. Anyway even this single passes reduce build time for me on 80%.

caiiiycuk:
I added new commit with PassRunner proxy. I make it virtual to do this. The time reduced twice. Total number is:

1. dosbox-x from 14Min to 2 min
2. dosbox from 2min to 30 sec

src/pass.h
```diff
@@ -426,8 +429,8 @@ class Pass {
 
   std::string name;
 
-  PassRunner* getPassRunner() { return runner; }
-  void setPassRunner(PassRunner* runner_) {
+  virtual PassRunner* getPassRunner() { return runner; }
+  virtual void setPassRunner(PassRunner* runner_) {
```
kripken, line 433: Why do these two methods need to be virtual? I think we can use a fixed pass runner, even if we virtualize the passes and the addition of passes.

(If this is necessary, we'd need to benchmark it to see if this is not a general regression - it's possible that `getPassRunner` is called quite frequently during optimizations.)

kripken (commented):
Nice results! Excellent speedup.

Overall this looks good to me, aside from (1) the comment below and (2) the test failures on CI.

caiiiycuk:
Not clear for me what to do with CI:
```

['-g']
  file size: 5220
[]
  file size: 4899
['-O1']
  file size: 3288
['--optimize-level=1']
  file size: 3397
['-O3']
  file size: 3129
['-Os', '-g']
  file size: 3354
```

I need to fix action, to match new file sizes?

kripken:
For the `lit` tests, you can auto-update them, `./scripts/auto_update_tests.py lit`.

The `file size` messages are not an error, I think - they might happen to be last because of the order in the output, I think it mixes up stdout and stderr, I'm not sure. To make sure you can run `./check.py unit` to run the unit tests which include the test that mentions `file size`.

caiiiycuk:
`./scripts/auto_update_tests.py` lit to solved CI problem. It change tests, how to validate that they are correct?

Regarding to (1), trying to understand how to solve it. If I don't override them, then I have assertion error, that PassRunner is not set for pass.

kripken:
> then I have assertion error, that PassRunner is not set for pass.

Which assertion specifically? (full text or line)

kripken:
> It changed the tests, how to check if they are correct?

They just need to be manually reviewed :smile: I'm reading them now.

test/lit/passes/asyncify_enable-multivalue.wast
```diff
@@ -37,18 +37,10 @@
...
-  ;; CHECK-NEXT:  (local $1 i32)
-  ;; CHECK-NEXT:  (local.set $0
-  ;; CHECK-NEXT:   (global.get $sleeping)
-  ;; CHECK-NEXT:  )
-  ;; CHECK-NEXT:  (local.set $1
+  ;; CHECK-NEXT:  (if
   ;; CHECK-NEXT:   (i32.eqz
-  ;; CHECK-NEXT:    (local.get $0)
+  ;; CHECK-NEXT:    (global.get $sleeping)
   ;; CHECK-NEXT:   )
```
kripken, line 43: This diff looks like it is more optimized after the change. That's odd, I'd expect less.

That is, the optimizer is removing the unneeded locals (using simplify-locals + reorder-locals, which means it is running those passes here that it didn't before.

test/lit/passes/asyncify_optimize-level=1.wast
```diff
@@ -301,7 +301,11 @@
     (drop (call $import2))
   )
   ;; CHECK:      (func $calls-nothing
-  ;; CHECK-NEXT:  (nop)
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (i32.eqz
+  ;; CHECK-NEXT:    (i32.const 17)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
   ;; CHECK-NEXT: )
```
kripken, line 309: This is the kind of diff I'd expect: the function after asyncify is identical to before asyncify. We did not modify it at all, showing the optimizer never ran, which is good.

test/lit/passes/asyncify_pass-arg=asyncify-asserts_pass-arg=asyncify-onlylist@waka.wast
```diff
@@ -39,128 +39,34 @@
...
-  ;; CHECK-NEXT:   (call $import)
-  ;; CHECK-NEXT:   (if
-  ;; CHECK-NEXT:    (i32.ne
-  ;; CHECK-NEXT:     (global.get $__asyncify_state)
-  ;; CHECK-NEXT:     (local.get $0)
-  ;; CHECK-NEXT:    )
-  ;; CHECK-NEXT:    (unreachable)
-  ;; CHECK-NEXT:   )
-  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (call $import)
```
kripken, line 42: This seems wrong - this calls an import, so asyncify should instrument it, as it did before.

caiiiycuk: @kripken As far I undestand in ModuleAnalyzer.map this function stored as **'import'**, but when InstumentedProxy.runOnFunction is called the function name is **'calls-import'**, this is the reason why it is not instrumented.

src/pass.h
caiiiycuk: Without overriding set/get I have following assertion errors:
```
Assertion failed: (getPassRunnerAssertion failed: (getPassRunner()), function runOnFunction, fil()), function runOnFunction, file pass.h, line 481.
e pass.h, line 481.
Assertion failed: (getPassRunner()), function runOnFunction, file pass.h, line 481.
Assertion failed: (getPassRunner()), function runOnFunction, file pass.h, line 481.
```

caiiiycuk: However I think I can workround this, by calling setPassRunner just before invoking proxied runner in case if it's pass runner is null.

test/lit/passes/asyncify_pass-arg=asyncify-asserts_pass-arg=asyncify-onlylist@waka.wast
caiiiycuk: If you run it in verbose mode you will see following output:
```
[asyncify] import is an import that can change the state
[asyncify] import2 is an import that can change the state
[asyncify] import3 is an import that can change the state
[asyncify] calls-indirect can change the state due to initial scan
[asyncify] calls-import can change the state due to import
[asyncify] calls-import2-drop can change the state due to import2
[asyncify] returns can change the state due to import2
[asyncify] calls-import's state is set based on the only-list to 0
[asyncify] calls-import2-drop's state is set based on the only-list to 0
[asyncify] returns's state is set based on the only-list to 0
[asyncify] calls-indirect's state is set based on the only-list to 0
```

Initially it marked as instrumented, but becaouse of asyncify-only-list it's marked back to non instrumented.

```
[asyncify] calls-import can change the state due to import
[asyncify] calls-import's state is set based on the only-list to 0
```

Because of that InstrumentedProxy didn't run passes on it.

kripken: I see, thanks - I missed that there is an only-list in this test.

src/pass.h
kripken: That sounds good, yes, the proxied runner can call `setPassRunner` before each proxied operation, if necessary.

caiiiycuk:
Updated, I checked that pass runner is required only in runOnFunction, so I set it just before this method call. I didn't squash the commits yet, because you need a history view.

test/lit/passes/asyncify_enable-multivalue.wast
kripken: Ah, this is a top-most runtime function, that we do not instrument. So this PR just makes us now flatten it, which is why it looks better.

kripken (approved):
Great!

kripken:
Thanks, yeah, it is useful to not squash before landing.

Everything here looks great, landing!

PR #5295, opened 2022-11-28
[NFC] Refactor GlobalTypeRewriter to split out the type mapping logic

kripken:
The logic that is split out into `mapTypes` gets a map of old type => new type and then
updates the module to replace the old with the new.

This will be useful in a future pass to merge types. We will tell it to map the types to
be merged with the types we want to merge them into.

This removes an assert on all types being in the map to allow some of them not to be.
(the new pass that will use this will only want to map some types).

### DIFF
--- a/src/ir/type-updating.cpp
+++ b/src/ir/type-updating.cpp
@@ -97,11 +97,23 @@ void GlobalTypeRewriter::update() {
   // Map the old types to the new ones. This uses the fact that type indices
   // are the same in the old and new types, that is, we have not added or
   // removed types, just modified them.
-  using OldToNewTypes = std::unordered_map<HeapType, HeapType>;
-  OldToNewTypes oldToNewTypes;
+  TypeMap oldToNewTypes;
   for (Index i = 0; i < indexedTypes.types.size(); i++) {
     oldToNewTypes[indexedTypes.types[i]] = newTypes[i];
   }
+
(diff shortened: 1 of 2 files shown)


src/ir/type-updating.cpp
```diff
@@ -136,8 +148,10 @@ void GlobalTypeRewriter::update() {
         return type;
       }
       if (type.isFunction() || type.isData()) {
-        assert(oldToNewTypes.count(type));
-        return oldToNewTypes[type];
+        auto iter = oldToNewTypes.find(type);
```
tlively, line 151: Is the condition `type.isFunction() || type.isData()` doing anything here? I don't think that can ever be false, unless I'm missing something.

tlively (approved):

src/ir/type-updating.cpp
kripken: Good question, I'm not sure. Perhaps it guards against a null type? Or maybe if we added `isRef()` on heap types (and not just types) we could use that here?

tlively: Can we just remove it, though?

kripken: Sorry, I missed your point earlier. Yes, I think that's right, now that the assert is gone we can remove the if too. I'll open a PR.

PR #5297, opened 2022-11-28
Validator: Print the field number on subtyping errors

kripken:


src/wasm/wasm-validator.cpp
```diff
@@ -2586,10 +2586,12 @@ void FunctionValidator::visitStructNew(StructNew* curr) {
                       "struct.new must have the right number of operands")) {
       // All the fields must have the proper type.
       for (Index i = 0; i < fields.size(); i++) {
-        shouldBeSubType(curr->operands[i]->type,
-                        fields[i].type,
-                        curr,
-                        "struct.new operand must have proper type");
+        if (!Type::isSubType(curr->operands[i]->type, fields[i].type)) {
+          info.fail("struct.new operand " + std::to_string(i) +
```
aheejin, line 2590: What's the difference between the case you use `shouldBeSubType`'s error message and `info.fail`?

aheejin (approved):

src/wasm/wasm-validator.cpp
kripken: Just that `shouldBeSubType` executes the arguments, so it would do wasted work to allocate a new string each time if I just added to `std::to_string` there.

Issue #5292, opened 2022-11-23
Dead code elimination is too agressive on struct globals [Closed World]


chambart:
When using `wasm-opt` (with `--enable-gc --enable-reference-types`) on the following `.wast` with optimisation level `-01` and up, there seems to be some dead code elimination pass that consider the function `$f` unreachable.

I suppose that the criterion for reachability of a function code is: it is reachable if it is called by a reachable function, by start or part of a function table. But with the gc extension, functions can also be part of exported globals.
```WASM
(module
  (type $f
    (func (param (ref eq)) (result (ref eq))))
  (type $s
    (struct (field (ref $f))))

  (elem declare func $f)
  (func $f (param $x (ref eq)) (result (ref eq))
    (local.get $x))

  (global $glob (export "g") (ref $s) (struct.new $s (ref.func $f)))
)
```

The produced code with `-01` is
```WASM
(module
 (type $f (func (param (ref eq)) (result (ref eq))))
 (type $s (struct (field (ref $f))))
 (global $glob (ref $s) (struct.new $s
  (ref.func $f)
 ))
 (export "g" (global $glob))
 (func $f (type $f) (param $0 (ref eq)) (result (ref eq))
  (unreachable)
 )
)
```

This was tested on the current tree (https://github.com/WebAssembly/binaryen/commit/2c0fb8c513b3534576383d250c593a9bd28347e3)

kripken:
I believe this is due to our current closed-world assumption (assuming a closed world, without any `call_ref` to the type of `$f`, it must be unreachable code). We've been planning to add a flag to turn that on and off, but so far haven't had use cases which require it. If it's necessary now perhaps we should prioritize that.

Can you describe your use case here? (what are you compiling to wasm GC, and how the outside code - JS? - wants to interact with it)

chambart:
This is an hand written example, but this is a pattern expected to happen from compiling OCaml.
This would be just one module out of a whole program, and I don't plan for the compiler to require a whole program linking phase. The closed-world assumption would not hold in my use case.

kripken:
Interesting, thanks for the info!

How would the separate modules communicate? We don't have any wasm GC toolchains that do that atm, so I'd like to learn more about that, if you have specific ideas there. Were you thinking to link the wasm modules together using imports and exports, or something involving JS?

chambart:
[...]

PR #5298, opened 2022-11-28
Add a placeholder closed-world flag

kripken:
The flag does nothing so far. My thinking is that we'll do this:

* Land this no-op flag.
* Get partners to pass the flag in (which will not error as there is this placeholder).
* Implement the flag, that is, make us open-world by default and make the flag switch to closed-world. This will not regress partners' performance since they pass the flag and so they are opting in to closed-world, making this part NFC for them.

See #5292


tlively (approved):
Code and plan LGTM.

PR #5299, opened 2022-11-29
[Wasm GC] Add TypeSSA pass 

kripken:
This creates new nominal types for each (interesting) `struct.new`. That then allows
type-based optimizations to be more precise, as those optimizations will track
separate info for each `struct.new`, in effect. That is kind of like SSA, however, we
do not handle merges. For example:

```js
x = struct.new $A (5);
print(x.value);
y = struct.new $A (11);
print(y.value);

// => //

x = struct.new $A.x (5);
print(x.value);
y = struct.new $A.y (11);
print(y.value);
```

After the pass runs each of those `struct.new` creates a unique type, and type-based
analysis can see that 5 or 11 are the only values written in that type (if nothing else
writes there).

This bloats the type section with the new subtypes, so it is best used with a pass
to merge unneeded duplicate types, which a later PR will add. That later PR will
exactly merge back in the types created here, which are nominally different but
indistinguishable otherwise.

I measured results using something like `--type-ssa --gufa -O2 --type-merging -O2`
(that does TypeSSA to let GUFA do more, then optimizes, then merges, then
does another optimization after types are merged). It looks quite good on Dart,
removing 25% of `call_indirect` and 33% of `call_ref` (also 4% of total locals and 10%
of `local.set`). The results on Java are much less impressive, there are benefits but
all are less than 1%. Not sure yet why that is, but it might just be that Dart code
has a form that benefits more from this type of optimization; also possible there
is a bug somewhere and it is optimizing too much, so that needs to be investigated.

This pass is not enabled by default. It's not clear yet where is the best place to do it,
as it must be balanced by type merging, but it might be better to do multiple
rounds of optimization between the two. Needs more investigation.

cc @askeksa-google for the (possible) Dart benefits here


src/passes/TypeSSA.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+    ModuleUtils::ParallelFunctionAnalysis<News> analysis(
+      *module, [&](Function* func, News& news) {
+        if (func->imported()) {
+          return;
+        }
+
+        NewFinder finder;
+        finder.walk(func->body);
+        news = std::move(finder.news);
```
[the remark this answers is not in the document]
kripken: I suppose, but then we'd need to declare the vectors in more places (before passing in a reference to them), so I think the current code is shorter. But I don't feel strongly.

src/passes/TypeSSA.cpp
```diff
@@ -0,0 +1,204 @@
...
+    for (auto* curr : news.structNews) {
+      if (isInteresting(curr)) {
+        // This is interesting; create a new type and use it. The new type is
+        // identical to the existing one, and a subtype of it, so the only
+        // difference between them is nominal - which is enough for optimization
+        // passes to learn different things about the two.
+        auto oldType = curr->type.getHeapType();
+        TypeBuilder builder(1);
+        builder.setHeapType(0, oldType.getStruct());
+        builder.setSubType(0, oldType);
```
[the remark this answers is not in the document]
kripken: Oh, yikes, good catch. That's pretty bad actually - sounds like we'd need to accumulate all the relevant news and handle them all at once. Perhaps the pass could only work in nominal mode for now, though - do we have users of non-nominal mode, do you know?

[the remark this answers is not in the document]
tlively: I assume the ocaml folks are using the isorecursive system, and in general we currently have the nice property that you get pretty much the same optimizations in isorecursive and nominal modes. I think it would be worth supporting isorecursive types now.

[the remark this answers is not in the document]
tlively: > The problem is that we don't want to create a new rec group that happens to match an existing one...

It's possible that the new rec group matches an existing rec group, but in that case none of the types in the existing rec group could possible have been allocated. (if they had been allocated, they wouldn't be in the group of new subtypes of allocated types!) It's still possible that there exists some cast to the existing type, though, which could cause problems. I seriously doubt this would happen in practice, so I think it would be fine to detect whether one of the "new" types is already used by the module and bail out of the optimization if that's the case.

> a single use of a type from that huge new rec group would keep the entire thing alive, wasting space...

This could be a problem even without hacks where we make huge rec groups just to make them "nominal." In a closed world, we could be optimizing unused types out of rec groups as long as we can prove that removing the types won't unify groups that would otherwise be distinct.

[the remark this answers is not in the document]
kripken: > It's possible that the new rec group matches an existing rec group, but in that case none of the types in the existing rec group could possible have been allocated. (if they had been allocated, they wouldn't be in the group of new subtypes of allocated types!) 

I don't follow that, sorry. (What does "allocated" mean? `struct.new`? But even if so, I don't follow.)

To elaborate what I think is the problem, say we have existing an existing rec group:
```wat
(rec
  (type $existing (struct_subtype (..fields of $A, no changes..) $A))
)
```
That rec group just happens to exist right now. And lets say we have a `struct.new $A` and we want to create a new type for that new, so we create
```wat
(rec
  (type $intended.to.be.new (struct_subtype (..fields of $A, no changes..) $A))
)
```
That is structurally identical to the first rec group, so `$existing` and `$intended.to.be.new` would be equal, I think?

And that is regardless of how `$existing` is used in the program. We'd be switching that `struct.new $A` to `struct.new $existing`, and old uses of that type could include things that prevent optimizations that we could do if we used a new type here (`$existing` could be written something generic, while the `struct.new $A` might have only sets of something specific/refined).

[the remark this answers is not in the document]
tlively: Yes, that's another problem with creating one new type at a time. They will be duplicates of each other, but also potentially duplicates of other types that already exist. I suppose if you create new subtypes for types that are only ever allocated once you can end up with a similar problem.

To minimize the chances of getting a conflicting with existing types, all the new types this pass will add for the entire module could be part of a single large rec group. I think that's what I had in mind when I wrote "none of the types in the existing rec group could possibl[y] have been allocated. (if they had been allocated, they wouldn't be in the group of new subtypes of allocated types!)" Yes, by allocated I mean allocated with a struct.new instruction.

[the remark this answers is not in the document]
kripken: I believe the problem I described can still happen while generating a single big rec group for all types at once, since we don't create a new type for all `struct.new`s. I think I see now that you were assuming that. But the pass actually finds news that can't be optimized ("uninteresting" ones) and does nothing on them. So I might find I want to create 10 new fresh subtypes of `$A, $B` and maybe, by chance, those types already have 10 existing subtypes - none of which are interesting - and those existing subtypes are in a rec group with the exact same shape as the one the pass creates :scream: 

IIUC the only real solution is to actually manually verify no existing rec group can possibly be structurally identical. Like by make a rec group larger than all others, or with a new weird element with more fields than all others. Which are all quite horrible...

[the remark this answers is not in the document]
tlively: Ah, I see, thanks. I still think that bailing out of the optimization if any of the new types are already used by the module would be an acceptable solution since it should be sufficiently rare (I hope) in practice.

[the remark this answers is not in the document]
kripken: Ok, code rewritten to create a single big rec group. I decided to error if we fail to create new/private/fresh types, for now - seems safer, and hopefully very rare as you said.

src/passes/TypeSSA.cpp
```diff
@@ -0,0 +1,257 @@
...
+  std::unordered_set<HeapType> allSeenTypes;
+
+  // As we generate new names, use a consistent index.
+  Index nameCounter = 0;
+
+  void processNews(const News& news) {
+    for (auto* curr : news.structNews) {
+      if (isInteresting(curr)) {
+        newsToModify.structNews.push_back(curr);
+        allSeenTypes.insert(curr->type.getHeapType());
```
tlively, line 129: This may miss types that are used e.g. in casts but not in allocations. We should still error out in those cases, so it would be better to use `ModuleUtils::collectHeapTypes` to find all the types.

src/passes/TypeSSA.cpp
```diff
@@ -0,0 +1,257 @@
...
+    // also a possibility, and so for that reason this pass is likely mostly
+    // useful in the closed-world scenario.
+
+    auto& structNews = newsToModify.structNews;
+    TypeBuilder builder(structNews.size());
+    for (Index i = 0; i < structNews.size(); i++) {
+      auto* curr = structNews[i];
+      auto oldType = curr->type.getHeapType();
+      builder.setHeapType(i, oldType.getStruct());
+      builder.setSubType(i, oldType);
```
tlively, line 155:
```suggestion
      builder[i] = oldType.getStruct();
      builder[i].subTypeOf(oldType);
```

src/passes/TypeSSA.cpp
```diff
@@ -0,0 +1,257 @@
...
+    // existing types, the new types merged with them would now succeed on those
+    // casts). We could try to create a "weird" rec group to avoid this (e.g. we
+    // could make a rec group larger than any existing one, or with an initial
+    // member that is "random"), but hopefully this is rare, so just error for
+    // now.
+    for (auto newType : newTypes) {
+      if (allSeenTypes.count(newType)) {
+        Fatal() << "Rec group collision in TypeSSA! Please file a bug";
+      }
+    }
```
tlively, line 174: Once you switch to using `ModuleUtils::collectHeapTypes`, you will only need to check a single type here rather than checking all of them. Since the types are all in the same rec group, if one of them is collected, they will all be collected.

Also, this seems like it should produce a warning at most and then bail out of the optimization without changing anything rather than producing a fatal error. We could more easily test this case that way, too.

src/passes/TypeSSA.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+    // TODO: arrays
+  }
+
+  // An interesting StructNew, which we think is worth creating a new type for,
+  // is one that can be optimized better with a new type. That means it must
+  // have something interesting for optimizations to work with.
+  bool isInteresting(StructNew* curr) {
```
tlively, line 212: I realized that the reason I was assuming all `struct.new`s get new types is that they are all potentially interesting if we additionally get an optimization that can partition a type hierarchy where subtypes from the two partitions are never mixed up together. Once the hierarchy is partitioned, the two partitions can be optimized separately and with fewer constraints.

Maybe worth a TODO?

test/lit/passes/type-ssa.wast
```diff
@@ -0,0 +1,221 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
+;; RUN: foreach %s %t wasm-opt           --type-ssa -all -S -o - | filecheck %s
+;; RUN: foreach %s %t wasm-opt --nominal --type-ssa -all -S -o - | filecheck %s --check-prefix NOMINAL
```
tlively, line 3:
```suggestion
;; RUN: foreach %s %t wasm-opt --nominal --type-ssa -all -S -o - | filecheck %s --check-prefix NOMNL
```

We use `NOMNL` in a few tests to get the checks to line up nicely.

test/lit/passes/type-ssa.wast
```diff
@@ -0,0 +1,221 @@
...
+  ;; NOMINAL-NEXT:  (drop
+  ;; NOMINAL-NEXT:   (block ;; (replaces something unreachable we can't emit)
+  ;; NOMINAL-NEXT:    (drop
+  ;; NOMINAL-NEXT:     (unreachable)
+  ;; NOMINAL-NEXT:    )
+  ;; NOMINAL-NEXT:    (unreachable)
+  ;; NOMINAL-NEXT:   )
+  ;; NOMINAL-NEXT:  )
+  ;; NOMINAL-NEXT: )
+  (func $foo (param $any (ref null any)) (param $data (ref null data))
```
tlively, line 192: It would be great if you could use `array` or some other type instead of `data`, since I'm planning to remove `data` soon.

src/passes/TypeSSA.cpp
```diff
@@ -0,0 +1,257 @@
...
+    if (curr->type == Type::unreachable) {
+      // This is dead code anyhow.
+      return false;
+    }
+
+    if (curr->isWithDefault()) {
+      // This starts with all default values - zeros and nulls - and that might
+      // be useful.
```
tlively, line 220: This assumes the struct isn't entirely composed of references to bottom types, but that's probably a fine assumption to make.

askeksa-google:
Very interesting that it has such a huge effect on Dart. A large part of the benefit likely comes from two particularities in how dart2wasm assigns Wasm structs:
- The struct for a class is sometimes reused by a subclass if the subclass does not add any fields.
- The structs for closures are the same for all closures with the same number of parameters. Similarly for their vtables.

The type diversification here thus leads to much better devirtualization for method calls (`call_indirect`) and function calls (`call_ref`).

I did an experiment where I use different Wasm structs for all classes and create specialized closure (and vtable) subclasses for every function expression / tear-off. I then tried optimizing barista3 with `--nominal -tnh -O3 --gufa -O3` before and after this change.

| | `call_indirect` | `call_ref` |
| - | - | - |
| Original | 4939 | 428 |
| Optimized, before | 4704 | 341 |
| Optimized, after | 4661 | 210 |

It appears that this way of emitting more specialized types is even more effective than TypeSSA for closure devirtualization, but not nearly as effective for method call devirtualization. Probably TypeSSA also captures second-level effects where fields of objects of the same class refer to different types in different situations.

src/passes/TypeSSA.cpp
kripken: Good catch! Done.

src/passes/TypeSSA.cpp
kripken: Hmm, good point about just checking one. I'll add that.

I do think this should error. A warning + bailout will likely mean this ends up ignored by the user, and we silently miss optimization opportunities. If that's an important case we'd have a slowdown there that is very hard to diagnose if the warning is missed. With an error the user would report a bug and then we'd know we need to prioritize a proper hack for the isorecursive case.

src/passes/TypeSSA.cpp
kripken: Sounds good, adding TODO.

kripken:
@tlively one of your suggestions fails to compile:
```cpp
TypeSSA.cpp:154:28: error: cannot convert ‘wasm::HeapType’ to ‘wasm::TypeBuilder::Entry’
  154 |       builder[i].subTypeOf(oldType);
      |                            ^~~~~~~
      |                            |
      |                            wasm::HeapType
```
Should that work?

kripken:
(seems like I need an Entry for a type not in the group - I don't see how to do that)

src/passes/TypeSSA.cpp
kripken: Adding a comment.

test/lit/passes/type-ssa.wast
kripken: Done.

src/passes/TypeSSA.cpp
tlively: Ok, if our response to a bug report about this error would be to implement a workaround, then that sounds good to me.

kripken:
@askeksa-google Very interesting data, thanks! And yes, those two bullet points very much sound like they explain why this pass helps as much as it does.

tlively:
https://github.com/WebAssembly/binaryen/pull/5314 should fix it.

test/lit/passes/type-ssa.wast
```diff
@@ -0,0 +1,221 @@
...
+;; RUN: foreach %s %t wasm-opt --nominal --type-ssa -all -S -o - | filecheck %s --check-prefix NOMNL
+
+;; Test in both isorecursive and nominal modes to make sure we create the new
+;; types properly in both.
+
+;; Every struct.new here should get a new type.
+(module
+  ;; CHECK:      (type $struct (struct (field i32)))
+  ;; NOMNL:      (type $struct (struct (field i32)))
+  (type $struct (struct_subtype (field i32) array))
```
tlively, line 12:
```suggestion
  (type $struct (struct_subtype (field i32) data))
```

This still needs to be "data", since it's not a real type but rather placeholder syntax from before we had an official text format for supertypes. I'll fix it to accept "array" and "struct" though.

tlively: Oh, actually, a better solution would be to not use `struct_subtype` here at all. It can just be a normal `struct`.

kripken: Meanwhile I flipped it to `data`. Isn't a normal `struct` the old equirecursive notation? I thought `struct_subtype` was the modern/correct form...

tlively: No, `struct_subtype` is a nonstandard format that we invented to experiment with nominal types. At some point we'll have to port everything over to the standard format and it will be a huge pain :(. Meanwhile you can use plain `(struct ...)` when there is no nontrivial supertype to declare. We should actually prefer that, since that will be valid in the standard format.

tlively (approved):

PR #5301, opened 2022-11-29
Fix validation and inlining bugs

### LINKED ISSUE #5294: Failing testcase after changing the default type system to isorecursive
Bisection found #5239
```
853b31ec89416bef0014e06f2defaef74f47b81e is the first bad commit                               
commit 853b31ec89416bef0014e06f2defaef74f47b81e                                                
Author: Thomas Lively <tlively@google.com>                                                     
Date:   Tue Nov 22 20:48:58 2022 -0600                                                         
                                                                                               
    Change the default type system to isorecursive (#5239)                                     
                                                                                               
    This makes Binaryen's default type system match the WasmGC spec.                           
                                                                                               
    Update the way type definitions without supertypes are printed to reduce the               
    output diff for MVP tests that do not involve WasmGC. Also port some                       
    type-builder.cpp tests from test/example to test/gtest since they needed to be             
    rewritten to work with isorecursive type anyway.                                           
                                                                                               
    A follow-on PR will remove equirecursive types completely.                                 
```

Testcase:
```wat
(module
 (rec
  (type $none_=>_none (func))
  (type $[i8] (array i8))
  (type $ref|array|_=>_i32 (func (param (ref array)) (result i32)))
  (type $f32_=>_f32 (func (param f32) (result f32)))
  (type $f64_=>_f64 (func (param f64) (result f64)))
 )
 (func $0 (type $ref|array|_=>_i32) (param $0 (ref array)) (result i32)
  (global.set $global$0
   (i32.const 1)
  )
  (i32.const -113)
 )
 (func $1 (type $none_=>_none)
  (drop
   (call $0
    (array.init_static $[i8])
   )
  )
 )
 (func $2 (type $none_=>_none)
  (drop
   (call $3)
  )
 )
 (func $3 (type $none_=>_i32) (result i32)
  (return_call $0
   (select (result (ref array))
    (ref.as_non_null
     (ref.null none)
    )
    (array.init_static $[i8])
    (i32.const 0)
   )
  )
 )
)
```
STR:
```
$ bin/wasm-opt -all w.wat --gufa --inlining --monomorphize-always
wasm-opt: /usr/local/google/home/azakai/Dev/binaryen/src/wasm-type.h:482: void wasm::Tuple::validate(): Assertion `type.isSingle()' failed.
Aborted
```
(also happens with `--nominal`)

tlively:
Inlining had a bug where it gave return_calls in inlined callees concrete types
even when they should have remained unreachable. This bug flew under the radar
because validation had a bug where it allowed expressions to have concrete types
when they should have been unreachable. The fuzzer found this bug by adding
another pass after inlining where the unexpected types caused an assertion
failure.

Fix the bugs and add a test that would have triggered the inlining bug.
Unfortunately the test would have also passed before this change due to the
validation bug, but it's better than nothing.

Fixes #5294.


tlively:
* main
  * **PR #5301** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5301" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5301?utm_source=stack-comment).

kripken (approved):
Another option might be to ignore (not inline) calls with an unreachable operand, but this seems simpler actually.

tlively:
It's a call inside the inlined callee, not the inlined call, that has the unreachable operand. I briefly considered that we might not want to inline functions with unreachable bodies (that never return), though.

kripken:
Ah, right, thanks...

I agree it makes sense to inline such things. An unreachable body might be important to inline actually, as it can lead to more code removal.

tlively:
If we propagated unreachability to function return types, then we'd get the DCE benefits in the caller without having to do the inlining. Maybe that's something to keep in mind for the future.

PR #5303, opened 2022-11-30
Implement closed-world flag

### LINKED ISSUE #5292: Dead code elimination is too agressive on struct globals [Closed World]

kripken:
With this change we default to an open world, that is, we do the safe thing
by default: we no longer assume a closed world. Users that want a closed
world must pass `--closed-world`.

Atm we just do not run passes that assume a closed world. (We might later
refine them to find which types don't escape and only optimize those.) The
RemoveUnusedModuleElements is an exception in that the closed-world
flag influences one part of its operation, but not the rest.

Fixes #5292


test/lit/passes/Oz.wast
```diff
@@ -5,14 +5,13 @@
 
 (module
   (memory 100 100)
-  ;; CHECK:      (rec
-  ;; CHECK-NEXT:  (type $i32_i32_=>_i32 (func (param i32 i32) (result i32)))
```
kripken, line 9: I have no idea why this changed. @tlively is it something obvious perhaps? If not I can investigate.

test/lit/passes/gto_and_cfp_in_O.wast
```diff
@@ -33,6 +52,11 @@
   ;; CHECK:      (func $main (type $none_=>_i32) (; has Stack IR ;) (result i32)
   ;; CHECK-NEXT:  (i32.const 100)
   ;; CHECK-NEXT: )
+  ;; OPEN_WORLD:      (func $main (type $none_=>_i32) (; has Stack IR ;) (result i32)
+  ;; OPEN_WORLD-NEXT:  (struct.get $struct 1
+  ;; OPEN_WORLD-NEXT:   (global.get $glob)
+  ;; OPEN_WORLD-NEXT:  )
+  ;; OPEN_WORLD-NEXT: )
```
kripken, line 59: closed world optimized to 100 above; open fails to optimize.

test/lit/passes/remove-unused-module-elements-refs.wast
```diff
@@ -1,14 +1,26 @@
 ;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
-;; RUN: foreach %s %t wasm-opt --remove-unused-module-elements --nominal -all -S -o - | filecheck %s
+;; RUN: foreach %s %t wasm-opt --remove-unused-module-elements --closed-world --nominal -all -S -o - | filecheck %s
+;; RUN: foreach %s %t wasm-opt --remove-unused-module-elements                --nominal -all -S -o - | filecheck %s --check-prefix OPEN_WORLD
+
+;; Test both open world (default) and closed world. In a closed world we can do
+;; more with function refs, as we assume nothing calls them on the outside, so
+;; if no calls exist to the right type, the function is not reached.
```
kripken, line 7: Open world does the same as closed here, except it fails to turn some function bodies to unreachable (when closed world can see they are not called, open can't infer that).

test/lit/passes/Oz.wast
kripken: So what seems to be happening is that the old code would run GTO which would rewrite types (it modifies types "in place" by creating copies and modifying them, then switching everything to the new types). Doing that in isorecursive mode leads to the creation of a new rec group, and we end up printing it. But the new code avoids running that pass, so we don't add a rec group, and I'm guessing the printer will by default not emit a rec group at all if all we have is a single big one?

If so then this change seems expected and ok. Actually the old behavior of running some opt passes and getting a rec group printed out, when there wasn't one before, seems potentially confusing. (That makes me think again about the connection between closed-world and isorecursive types...)

tlively: > I'm guessing the printer will by default not emit a rec group at all if all we have is a single big one?

Not quite. The printer will not emit a rec group for singleton rec groups, i.e. rec groups with just one element. The input module only has singleton rec groups, but `TypeUpdater` always creates single large rec groups, which is why the output has a rec group.

kripken: Got it, thanks! So any use of TypeUpdater will go from many rec groups to a single one, which is then printed...

But isn't that unsafe in isorecursive mode? The many rec groups could have matched something in another module, but when we switch to a single big one they no longer do?

tlively: It's unsafe with an open world, but fine with a closed world + isorecursive types.

kripken: Ok, now I see. Then I think we will be doing the right thing after this PR, as we won't run those passes in an open world.

But is there any reason for someone to use closed world + isorecursive? What I mean is, if you are closed world, wouldn't you also want to use nominal? It's strictly better and has no downside AFAICT.

tlively: Talked offline and decided that supporting all combinations of nominal or isorecursive and open or closed world doesn't make sense. The solution will be to delete the nominal type system and determine which optimizations are safe to run just based on the open/closed world flags.

src/passes/RemoveUnusedModuleElements.cpp
```diff
@@ -313,28 +323,34 @@ struct RemoveUnusedModuleElements : public Pass {
...
+    ReachabilityAnalyzer analyzer(module, roots, closedWorld);
 
     // RefFuncs that are never called are a special case: We cannot remove the
     // function, since then (ref.func $foo) would not validate. But if we know
     // it is never called, at least the contents do not matter, so we can
     // empty it out.
+    //
+    // We can only do this in a closed world, as otherwise function references
+    // may be called outside of the module (if they escape, which we could in
+    // principle track, so the TODO earlier in this file).
```
tlively, line 336:
```suggestion
    // principle track, see the TODO earlier in this file).
```

src/passes/pass.cpp
```diff
@@ -594,18 +594,24 @@ void PassRunner::addDefaultGlobalOptimizationPrePasses() {
     addIfNoDWARFIssues("once-reduction");
   }
   if (wasm->features.hasGC() && options.optimizeLevel >= 2) {
-    addIfNoDWARFIssues("type-refining");
-    addIfNoDWARFIssues("signature-pruning");
-    addIfNoDWARFIssues("signature-refining");
+    if (options.closedWorld) {
```
tlively, line 597: What would you think about moving this logic to check `closedWorld` into each pass, just like we used to bail out of passes early under the equirecursive type system? That will be closer to how we want this to work eventually, where passes will still run under an open world but will be less aggressive.

test/lit/passes/remove-unused-module-elements-refs.wast
```diff
@@ -119,9 +179,13 @@
 ;; As above, but interleave CallRefs with RefFuncs.
 (module
   ;; CHECK:      (type $A (func))
+  ;; OPEN_WORLD:      (type $A (func))
   (type $A (func))
   (type $B (func))
 
+
+
+
```
tlively, line 188:
```suggestion
```

It also looks like there a number of other whitespace additions below. Are they intentional?

src/passes/pass.cpp
kripken: That's possible, but it would be harder to test. The current test checks if the passes are run at all. If instead the passes ran but exited early then we'd need specific tests for each pass to see if they have an effect or not. That's not unreasonable, but I'm not sure it's worth the extra work at this time?

tlively: Ok, doing this work later seems fine as well.

test/lit/passes/remove-unused-module-elements-refs.wast
kripken: Hmm, not sure what caused those... removed.

src/passes/RemoveUnusedModuleElements.cpp
```diff
@@ -313,28 +323,34 @@ struct RemoveUnusedModuleElements : public Pass {
...
+    //
+    // We can only do this in a closed world, as otherwise function references
+    // may be called outside of the module (if they escape, which we could in
+    // principle track, see the TODO earlier in this file).
     std::unordered_set<Name> uncalledRefFuncs;
-    for (auto& [type, targets] : analyzer.uncalledRefFuncMap) {
-      for (auto target : targets) {
-        uncalledRefFuncs.insert(target);
-      }
+    if (closedWorld) {
```
aheejin, line 338: Aren't we adding anything anyway if `closedWorld` is not true, in `ReachabilityAnalyzer ::visitCallRef`?

test/lit/passes/remove-unused-module-elements-refs.wast
tlively: I hope the update script doesn't put them back. Can you check?

test/lit/passes/remove-unused-module-elements-refs.wast
aheejin: Which functions? Can we specify them in comments to make this easier to read?

tlively (approved):

aheejin (approved):

test/lit/passes/remove-unused-module-elements-refs.wast
kripken: It doesn't, yeah. But it left them in while they were there, somehow.

test/lit/passes/remove-unused-module-elements-refs.wast
kripken: I added comments in the tests now, good idea.

Also when reading them carefully I noticed we introduced a bug in a test recently... we had a null for a call_ref input when it should have been a local.get, so it was not testing the right thing. I changed it to match the comments around it.

kripken: (that is the last commit)

src/passes/RemoveUnusedModuleElements.cpp
kripken: Good point, yeah, `uncalledRefFuncMap` will be empty then. I removed the `if` (and added an assert).

src/passes/RemoveUnusedModuleElements.cpp
```diff
@@ -333,24 +333,27 @@ struct RemoveUnusedModuleElements : public Pass {
     //
     // We can only do this in a closed world, as otherwise function references
     // may be called outside of the module (if they escape, which we could in
-    // principle track, see the TODO earlier in this file).
+    // principle track, see the TODO earlier in this file). So in the case of an
+    // open world we should not have noted anything in uncalledRefFuncMap
+    // earlier and not do any related optimizations there.
+    if (!closedWorld) {
+      assert(analyzer.uncalledRefFuncMap.empty());
+    }
```
aheejin, line 341:
```suggestion
    assert(closedWorld || analyzer.uncalledRefFuncMap.empty());
```
Super minor nit... Feel free to ignore if you see this after CI finishes.

test/lit/passes/remove-unused-module-elements-refs.wast
tlively: Great, that sounds like the expected behavior.

PR #5304, opened 2022-11-30
[NFC] Avoid unneeded work in GTO

kripken:
As noticed in #5303, the test changes here are because we did unnecessary work
which created a new rec group, which then led to a rec group being printed out.


kripken:
(I verified we have separate tests for both immutability and removals, so we have test coverage for both parts of the condition in the new `if`.)

tlively (approved):

PR #5307, opened 2022-12-01
Do not special case ref.null in `LUBFinder`

tlively:
Do not special case ref.null in LUBFinder

Before we implemented bottom heap types, `ref.null` had to be annotated with
specific types. The `LUBFinder` utility ignored these types so that it could
find the best LUB from all considered non-null expressions, then go back and
update the type annotations on the nulls to match that LUB. Now that we have
bottom types, however, none of that is necessary, and in fact ignoring nulls can
miss possible refinements to bottom types.

Update and simplify `LUBFinder` so that it is a simple wrapper around the
underlying `Type::getLeastUpperBound` utility with no additional logic. Update
tests to account for the more powerful optimizations.

no-op internal change

remove updateNulls

Remove noteNullDefault

note nullrefs

Remove noteUpdatableExpression

Rename getBestPossible to getLUB

Remove obsolete comment


tlively:
* main
  * **PR #5307** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5307" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5307?utm_source=stack-comment).

test/lit/passes/signature-refining.wast
```diff
@@ -508,16 +508,16 @@
   (type $sig-can-refine (func_subtype (result anyref) func))
 
   ;; Also a single function, but no refinement is possible.
-  ;; CHECK:      (type $sig-cannot-refine (func (result anyref)))
-  (type $sig-cannot-refine (func_subtype (result anyref) func))
+  ;; CHECK:      (type $sig-cannot-refine (func (result (ref func))))
+  (type $sig-cannot-refine (func_subtype (result funcref) func))
```
kripken, line 512: Comment says no refinement is possible, but the type is refined from `funcref` to `ref func` :thinking:

kripken (commented):
lgtm otherwise

test/lit/passes/signature-refining.wast
tlively: Ah oops, good catch.

PR #5308, opened 2022-12-02
Support `array` and `struct` types in the type fuzzer

tlively:
Since `data` has been removed from the upstream proposal and `struct` has been
added in its place, update the type fuzzer to be structured around `struct` and
`array` (which it had not previously been updated to support) rather than
`data`. A follow-on PR will make the broader change of removing `data` and
adding `struct`.


tlively:
* main
  * **PR #5308** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5308" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5308?utm_source=stack-comment).

PR #5309, opened 2022-12-02
Use C++17's [[maybe_unused]]. NFC

sbc100:


tlively (approved):
Nice, I like this better.

PR #5313, opened 2022-12-02
Work around possible unsafe MSVC NaN optimization

tlively:
MSVC 2019 is getting all parsed NaN signs backwards, even with `/fp:precise` or
`/fp:strict`. Those options only cause floats to conform to the source at
assignments, casts, and on function boundaries, but notably not in intermediate
results. My guess is that the double produced by the ternary statement is
considered an intermediate result, so MSVC considers itself free to perform
unsafe optimizations on it. Avoid the ternary statement to fix this problem if
that hypothesis is correct.

### DIFF
--- a/src/wasm/wat-lexer.cpp
+++ b/src/wasm/wat-lexer.cpp
@@ -226,8 +226,10 @@ struct LexFloatCtx : LexCtx {
       return {};
     }
     if (nanPayload) {
-      double nan = basic->span[0] == '-' ? negNan : posNan;
-      return LexFloatResult{*basic, nanPayload, nan};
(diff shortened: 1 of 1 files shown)


tlively:
* main
  * **PR #5313** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5313" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5313?utm_source=stack-comment).

kripken (approved):
Makes sense and lgtm if it works.

tlively:
This is superseded by #5316.

PR #5315, opened 2022-12-02
Fix two issues in --spill-pointers

juj:
Fix two issues in --spill-pointers: 1) properly generate the stack frame in a stack-grows-downwards manner, and 2) do not spill pointers in the functions stackAlloc()/stackSave()/stackRestore(), which should operate on the global stack pointer without a stack frame of their own.


src/passes/SpillPointers.cpp
```diff
@@ -76,7 +77,12 @@ struct SpillPointers
   Type pointerType;
 
   void spillPointers() {
-    pointerType = getModule()->memories[0]->indexType;
+     // TODO: Move these two lines somewhere earlier since it only needs to be done once (but where?)
```
kripken, line 80: A function-parallel pass like this doesn't have a "global" phase where you can look at imports once and then use that info. We could make the pass not parallel, but the number of imports is so small, and this pass is experimental, that it's probably fine to just leave it.

src/passes/SpillPointers.cpp
```diff
@@ -76,7 +77,12 @@ struct SpillPointers
...
+     const char *HANDLE_STACK_OVERFLOW = "__handle_stack_overflow"; // TODO: This should be passRunner.options.arguments["stack-check-handler"], but don't know how to access it here.
```
kripken, line 81:
```suggestion
     const char *HANDLE_STACK_OVERFLOW = getPassRunner().options.getArgumentOrDefault(...
```

hoodmane:
What is the status of this patch? Has anyone investigated the CI failures? Thanks for working on this @juj! It looks helpful.

juj:
Took quite a bit of time to get back to looking into GC on Emscripten..

I see that in current main branch, the issue 1) has already been fixed.

So I reset this branch now on top of latest main, and only kept the second fix

   "2) do not spill pointers in the functions stackAlloc()/stackSave()/stackRestore(), which should operate on the global stack pointer without a stack frame of their own."

from the original code.

I am developing a small toy garbage collector to test things out, and with this Binaryen PR landed, it is now working properly to scan the local stack and global static data section.

src/passes/SpillPointers.cpp
juj: Thanks, that worked.

juj:
I am testing this out against https://github.com/juj/emgc to prototype garbage collection against Emscripten. The plan is to experiment with multithreaded garbage collection from multiple managed threads, with cooperative stop-the-world signaling, and then try to transplant the whole thing back to Boehm, or something like that.

kripken:
@juj Are you aiming to get this approved and landed here in upstream, or is this just for experimentation at this time?

juj:
This is now for experimentation, though I'll try to work towards test cases on the stackAlloc issue to make sure I'm proposing a fix that patches that issue.

PR #5316, opened 2022-12-02
[Parser] Avoid calling `strtod` on NaNs entirely

tlively:
MSVC's implementation of `strtod` doesn't return a negative Nan for "-nan", so
we already had a workaround to explicitly handle that case without calling
`strtod`. Unfortunately the workaround was not used for negative NaNs with
payloads, so there were still bugs. Fix the problem and make the code even more
portable by avoiding `strtod` completely for any kind of nan, positive or
negative, with or without payload.

### DIFF
--- a/src/wasm/wat-lexer.cpp
+++ b/src/wasm/wat-lexer.cpp
@@ -227,12 +227,13 @@ struct LexFloatCtx : LexCtx {
     if (!basic) {
       return {};
     }
-    if (nanPayload) {
-      double nan = basic->span[0] == '-' ? negNan : posNan;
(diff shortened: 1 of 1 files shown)


tlively:
* main
  * **PR #5316** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5316" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5316?utm_source=stack-comment).

kripken (approved):

PR #5320, opened 2022-12-02
Update tests ahead of transition from `data` to `struct`

tlively:
The upstream WasmGC spec has removed `data` and introduced `struct`. To make the
migration easier, we have been supporting `struct` as an `alias` for `data` and
`structref` as an alias for `dataref`.

Update the tests to prefer the `struct` aliases over `data` for test input to
make the future migration easier. Also update some tests that had stale comments
about ref.null types being updated and remove some tests for instructions like
br_on_data and ref.as_data that do not make sense without a `data` type.

### DIFF
--- a/test/lit/passes/dae-gc.wast
+++ b/test/lit/passes/dae-gc.wast
@@ -211,15 +211,15 @@
  ;; NOMNL-NEXT:  )
  ;; NOMNL-NEXT: )
  (func $call-bar
-  ;; Call with nulls. Mixing nulls is fine as they all have the same value, and
-  ;; we can optimize (to the LUB of the nulls). However, mixing a null with a
-  ;; reference stops us in the second param.
+  ;; Call with nulls. Mixing nulls is fine as they all have the same type and
+  ;; value. However, mixing a null with a reference stops us in the second
+  ;; param.
   (call $bar
-   (ref.null i31)
-   (ref.null data)
+   (ref.null none)
+   (ref.null none)
   )
   (call $bar
-   (ref.null any)
+   (ref.null none)
    (i31.new (i32.const 0))
   )
  )
--- a/test/lit/passes/inlining_splitting.wast
+++ b/test/lit/passes/inlining_splitting.wast
@@ -9,10 +9,10 @@
 
   ;; CHECK:      (type $anyref_=>_anyref (func (param anyref) (result anyref)))
 
-  ;; CHECK:      (type $anyref_=>_none (func (param anyref)))
-
   ;; CHECK:      (type $struct (struct ))
   (type $struct (struct))
+
+  ;; CHECK:      (type $anyref_=>_none (func (param anyref)))
 
   ;; CHECK:      (type $i64_i32_f64_=>_none (func (param i64 i32 f64)))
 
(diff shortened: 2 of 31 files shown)


tlively:
* main
  * **PR #5320** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5320" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5320?utm_source=stack-comment).

test/lit/passes/dae-gc.wast
```diff
@@ -211,15 +211,15 @@
...
-  ;; we can optimize (to the LUB of the nulls). However, mixing a null with a
-  ;; reference stops us in the second param.
+  ;; Call with nulls. Mixing nulls is fine as they all have the same type and
+  ;; value. However, mixing a null with a reference stops us in the second
+  ;; param.
   (call $bar
-   (ref.null i31)
-   (ref.null data)
+   (ref.null none)
+   (ref.null none)
```
kripken, line 219: Is there no value in keeping tests for us treating all nulls the same? I guess we are assuming that we turn them into the same `none` during parsing - how is that tested?

test/lit/passes/inlining_splitting.wast
```diff
@@ -1192,96 +1189,29 @@
...
+  ;; CHECK-NEXT:      (local.get $0)
   ;; CHECK-NEXT:     )
   ;; CHECK-NEXT:    )
   ;; CHECK-NEXT:   )
   ;; CHECK-NEXT:  )
   ;; CHECK-NEXT: )
   (func $call-multi-if
-    (drop (call $multi-if (ref.null any)))
-    (drop (call $multi-if (ref.null data)))
+    (drop (call $multi-if (ref.null none)))
```
kripken, line 1212: maybe call it twice, like before, even with the same param? The diff will be smaller, and it would then obviously be testing the same thing. Otherwise the test output here changes in an interesting way...

test/lit/passes/dae-gc.wast
tlively: I guess we test the parsing in every test that uses a non-bottom type with a null. But in any particular test, it's better to use `none` so that the difference between input and output is limited to things that the test is trying to show. If we ever get to a point where most tests are using bottom types in their input, then we should add a dedicated test for the other cases.

kripken: Sounds good.

test/lit/passes/inlining_splitting.wast
```diff
@@ -1245,43 +1245,10 @@
...
-  ;; CHECK-NEXT:   (call $import)
-  ;; CHECK-NEXT:  )
-  ;; CHECK-NEXT:  (if
-  ;; CHECK-NEXT:   (ref.is_null
-  ;; CHECK-NEXT:    (local.get $x)
-  ;; CHECK-NEXT:   )
-  ;; CHECK-NEXT:   (call $import)
-  ;; CHECK-NEXT:  )
-  ;; CHECK-NEXT:  (local.get $x)
-  ;; CHECK-NEXT: )
```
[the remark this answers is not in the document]
tlively: Ah, this function got inlined and removed because I had reduced the number of calls below as well. Fixed now.

kripken (approved):

PR #5323, opened 2022-12-05
Fix an Inlining bug with a name collision in a br nested in a call param

kripken:
This bug has existed for many years, but the fuzzer only found it now...


src/ir/branch-utils.h
```diff
@@ -327,6 +327,12 @@ struct BranchAccumulator
     auto selfBranches = getUniqueTargets(curr);
     branches.insert(selfBranches.begin(), selfBranches.end());
   }
+
+  static NameSet get(Expression* tree) {
+    BranchAccumulator branches;
+    branches.walk(tree);
+    return branches.branches;
```
tlively, line 334: I would find this naming clearer.
```suggestion
    BranchAccumulator accumulator;
    accumulator.walk(tree);
    return accumulator.branches;
```

src/passes/Inlining.cpp
```diff
@@ -338,11 +338,25 @@ static Expression* doInlining(Module* module,
...
+  //      (call's first parameter
+  //        (br $X)         ;; nested br in call's first parameter
+  //      )
+  //    )
+  //
+  // (In this case we could use a second block and define the named block $X
+  // after the call's parameters, but that adds work for an extremely rare
+  // situation.)
+  if (BranchUtils::hasBranchTarget(from->body, block->name) ||
+      BranchUtils::BranchSeeker::has(call, block->name)) {
```
tlively, line 354: What is the difference between `BranchUtils::hasBranchTarget` and `BranchUtils::BranchSeeker::has`?

tlively: Got it, thanks. Does it save work to check for existence before gathering all the names below? It seems like gathering names would be about the same amount of work.

kripken: Gathering them requires us to build up a set (allocation+hashing) so it is a little more work at least. Probably not a lot in the general case. But it is extremely rare to have a collision here.

tlively: Makes sense 👍

PR #5325, opened 2022-12-06
Fix regression of assertions in asyncify pass

caiiiycuk:
Ohh, I am really sorry but my last PR #5293 have an issue. The code run passes only on functions that are instrumented, because of that injection of assertions is not working.

`AsyncfiyFlow` is not runned for non instrumented functions, but this pass also adds assert messages:

```c++
  void runOnFunction(Module* module_, Function* func_) override {
    module = module_;
    func = func_;
    builder =
      make_unique<AsyncifyBuilder>(*module, pointerType, asyncifyMemory);
    // If the function cannot change our state, we have nothing to do -
    // we will never unwind or rewind the stack here.
    if (!analyzer->needsInstrumentation(func)) {
      if (analyzer->asserts) {
        addAssertsInNonInstrumented(func);
      }
      return;
    }

```

I modified the code to run `AsyncifyFlow` on all functions as before. Maybe is 
better to extract this part of `AsyncifyFlow` to something like `AssyncifyAssertsPass`
to be more precies. What do you think?

--
Seems that asyncify-only lit test was correct :facepalm:

### DIFF
--- a/src/passes/Asyncify.cpp
+++ b/src/passes/Asyncify.cpp
@@ -886,6 +886,10 @@ struct InstrumentedPassRunner : public PassRunner {
   InstrumentedPassRunner(Module* wasm, ModuleAnalyzer* analyzer)
     : PassRunner(wasm), analyzer(analyzer) {}
 
+  void addForAll(std::unique_ptr<Pass> pass) {
+    PassRunner::doAdd(std::move(pass));
+  }
+
 protected:
   void doAdd(std::unique_ptr<Pass> pass) override {
     PassRunner::doAdd(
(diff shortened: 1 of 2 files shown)


kripken:
Good catch - I forgot we also add assertions in non-instrumented functions. The tests should have mentioned it more clearly so we didn't miss this :smile: Also, apparently we are missing an Emscripten test for that as the test suite there passed as well... I'll look into that.

> Maybe is better to extract this part of AsyncifyFlow to something like AssyncifyAssertsPass to be more precies. What do you think?

Yes, I think that's a good idea. How about just moving `addAssertsInNonInstrumented()` into a new pass, `AssertsInNonInstrumented`? We could then avoid adding that pass at all based on whether assertions are on or not. That would keep things at full speed in the normal case.

kripken:
> Also, apparently we are missing an Emscripten test for that as the test suite there passed as well... I'll look into that.

Oh, to reply to myself - the Emscripten test suite on Chromium CI passed, but that does not run browser tests. I think once this change reaches Emscripten on github (in the new emsdk tot build) we'll see failures. So this would not have gotten missed, but there is a delay unfortunately...

kripken:
Actually, I think I'll look into implementing that idea - I'm not sure which time zone you're in, but if you can only look at it tomorrow I'd rather spend a few minutes to fix this (should be simple) and avoid breakage on CI. Unless you respond in the next 10 minutes or so I'll start work.

kripken:
I opened #5328, let's close this in favor of that.

Thanks for noticing this problem so quickly!

PR #5327, opened 2022-12-06
[Wasm GC] Add array support to TypeSSA

kripken:
Previously it only handled structs.

### DIFF
--- a/src/passes/TypeSSA.cpp
+++ b/src/passes/TypeSSA.cpp
@@ -212,8 +212,6 @@ struct TypeSSA : public Pass {
         existingTypeNames.insert(newName);
       }
     }
-
-    // TODO: arrays
   }
 
   // An interesting *.new, which we think is worth creating a new type for, is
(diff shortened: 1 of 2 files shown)


src/passes/TypeSSA.cpp
```diff
@@ -210,41 +212,28 @@ struct TypeSSA : public Pass {
...
-    auto& fields = curr->type.getHeapType().getStruct().fields;
-    for (Index i = 0; i < fields.size(); i++) {
-      assert(i <= curr->operands.size());
-      auto* operand = curr->operands[i];
-      if (operand->type != fields[i].type) {
+    // Look for at least one interesting operand. We will consider each operand
+    // against the declared type, that is, the type declared for where it is
+    // stored. If it has more information than the declared type then it is
+    // interesting.
+    auto isInterestingRelevantTo = [&](Expression* operand, Type declaredType) {
```
tlively, line 235: I'm not sure this is worth pulling out into a lambda rather than just having `operand->type != declaredType` inline.

tlively (approved):
LGTM % comment.

src/passes/TypeSSA.cpp
kripken: The lambda also contains the `PossibleConstantValues` logic. The type might be the same, but if it is a constant or a global then we also consider it interesting.

PR #5328, opened 2022-12-06
Fix Asyncify assertions after #5293

kripken:
Followup to #5293, this fixes a small regression there regarding assertions. We do have
a need to visit non-instrumented functions if we want assertions, as we assert on some
things there, namely that such functions do not change the state (if they changed it,
we'd need to instrument them to handle that properly).

This moves that logic into a new pass. We run that pass when assertions are enabled.

Test diff basically undoes part the test diff from that earlier PR for that one file.

### DIFF
--- a/test/lit/passes/asyncify_pass-arg=asyncify-asserts_pass-arg=asyncify-onlylist@waka.wast
+++ b/test/lit/passes/asyncify_pass-arg=asyncify-asserts_pass-arg=asyncify-onlylist@waka.wast
@@ -3,6 +3,11 @@
 
 ;; RUN: foreach %s %t wasm-opt --asyncify --pass-arg=asyncify-asserts --pass-arg=asyncify-onlylist@waka -S -o - | filecheck %s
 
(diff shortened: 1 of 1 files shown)


src/passes/Asyncify.cpp
```diff
@@ -1689,6 +1724,16 @@ struct Asyncify : public Pass {
       runner.setValidateGlobally(false);
       runner.run();
     }
+    if (asserts) {
+      // Add asserts in non-instrumented code. Note we do not use an
+      // instrumented pass runner here as we do want to run on all functions.
```
tlively, line 1729: What is an "instrumented pass runner"?

tlively (approved):

src/passes/Asyncify.cpp
kripken: A pass runner with some instrumentation to decide on which functions to run. Specifically we just want to run on functions that need the asyncify instrumentation. Maybe the word is used in two slightly different manners here? Can improve that in a followup if there are better ideas.

PR #5330, opened 2022-12-07
Validate ref.as_* argument is a reference

kripken:
Without this we hit an assert with no line number info (or in a no-asserts build,
bad things can happen). With this:
```
$ bin/wasm-opt -all ~/Downloads/crash.wat  --nominal
[parse exception: Invalid ref for ref.as (at 155065:119)]
Fatal: error parsing wasm
```

(That can only happen for `ref.as_non_null`, as all the others do not have that
assert - their types do not depend on the child's type, so their finalize does not
error. Still, it is nice to validate earlier for them as well, so this PR handles them
all.)


tlively (commented):
This will conflict with #5331, where I delete `ValidatingBuilder`. If we can't leave this check to the validator, can you do it inline in the parser instead?

kripken:
Sure, I can do that. This is not urgent to land, also.

kripken:
Rewritten that way.

tlively (approved):
LGTM, but why can't this be left to the validator? We leave most similar checks to the validator.

kripken:
The validator is run after we build the IR. But this will error during building of the IR - we call finalize to set the type, and that errors because it has no valid type to work with.

PR #5331, opened 2022-12-07
Add standard versions of WasmGC casts

tlively:
We previously supported only the non-standard cast instructions introduced when
we were experimenting with nominal types. Parse the names and opcodes of their
standard counterparts and switch to emitting the standard names and opcodes.
Port all of the tests to use the standard instructions, but add additional tests
showing that the non-standard versions are still parsed correctly.


tlively:
* main
  * **PR #5331** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5331" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5331?utm_source=stack-comment).

kripken (approved):

tlively:
This isn't quite correct because it emits the new opcodes but still emits target type immediates as unsigned LEBs, when the new instructions expect signed LEBs. I have a follow-up PR almost ready to go that will fix this, but if you see strange errors in the meantime, that might explain them.

tlively:
The binary format issue should be fixed by #5332.

PR #5333, opened 2022-12-08
Fix opt/shrink levels when running the optimizer multiple times

kripken:
Previously `-O3 -O1` would run `-O1` twice since the last flag set the global opt level
to 1, and then all invocations of the optimizer pipeline read that. This makes each
pipeline define its own opt level.

This has been a long-standing annoyance, which wasn't much noticed except that
with wasm GC there is more of a need to run the optimization pipeline more than
once. And sometimes it is nice to run different levels.

### DIFF
--- a/src/tools/optimization-options.h
+++ b/src/tools/optimization-options.h
@@ -336,13 +336,13 @@ struct OptimizationOptions : public ToolOptions {
     }
     for (auto& pass : passes) {
       // We apply the pass's intended opt and shrink levels, if any.
-      auto oldOptimizeLevel = passOptions.optimizeLevel;
-      auto oldShrinkLevel = passOptions.shrinkLevel;
+      auto oldOptimizeLevel = passRunner.options.optimizeLevel;
+      auto oldShrinkLevel = passRunner.options.shrinkLevel;
       if (pass.optimizeLevel) {
-        passOptions.optimizeLevel = *pass.optimizeLevel;
+        passRunner.options.optimizeLevel = *pass.optimizeLevel;
       }
       if (pass.shrinkLevel) {
-        passOptions.shrinkLevel = *pass.shrinkLevel;
+        passRunner.options.shrinkLevel = *pass.shrinkLevel;
       }
 
       if (pass.name == DEFAULT_OPT_PASSES) {
@@ -352,8 +352,8 @@ struct OptimizationOptions : public ToolOptions {
       }
 
       // Revert back to the default levels, if we changed them.
-      passOptions.optimizeLevel = oldOptimizeLevel;
-      passOptions.shrinkLevel = oldShrinkLevel;
+      passRunner.options.optimizeLevel = oldOptimizeLevel;
+      passRunner.options.shrinkLevel = oldShrinkLevel;
     }
     passRunner.run();
   }
--- a/src/tools/optimization-options.h
+++ b/src/tools/optimization-options.h
@@ -336,14 +336,12 @@ struct OptimizationOptions : public ToolOptions {
     }
     for (auto& pass : passes) {
       // We apply the pass's intended opt and shrink levels, if any.
-      int oldOptimizeLevel;
-      int oldShrinkLevel;
+      auto oldOptimizeLevel = passOptions.optimizeLevel;
+      auto oldShrinkLevel = passOptions.shrinkLevel;
       if (pass.optimizeLevel) {
-        oldOptimizeLevel = passOptions.optimizeLevel;
         passOptions.optimizeLevel = *pass.optimizeLevel;
       }
       if (pass.shrinkLevel) {
-        oldShrinkLevel = passOptions.shrinkLevel;
         passOptions.shrinkLevel = *pass.shrinkLevel;
       }
 
@@ -354,12 +352,8 @@ struct OptimizationOptions : public ToolOptions {
       }
 
       // Revert back to the default levels, if we changed them.
-      if (pass.optimizeLevel) {
-        passOptions.optimizeLevel = oldOptimizeLevel;
-      }
-      if (pass.shrinkLevel) {
-        passOptions.shrinkLevel = oldShrinkLevel;
-      }
+      passOptions.optimizeLevel = oldOptimizeLevel;
+      passOptions.shrinkLevel = oldShrinkLevel;
     }
     passRunner.run();
   }
--- a/src/tools/optimization-options.h
+++ b/src/tools/optimization-options.h
@@ -42,7 +42,8 @@ struct OptimizationOptions : public ToolOptions {
 
     PassInfo(std::string name) : name(name) {}
     PassInfo(const char* name) : name(name) {}
-    PassInfo(std::string name, int optimizeLevel, int shrinkLevel) : name(name), optimizeLevel(optimizeLevel), shrinkLevel(shrinkLevel) {}
+    PassInfo(std::string name, int optimizeLevel, int shrinkLevel)
+      : name(name), optimizeLevel(optimizeLevel), shrinkLevel(shrinkLevel) {}
   };
 
   std::vector<PassInfo> passes;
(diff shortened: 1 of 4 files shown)


src/tools/optimization-options.h
```diff
@@ -30,7 +30,36 @@ struct OptimizationOptions : public ToolOptions {
...
+  // current opt and shrink levels specified, which are read from passOptions.
+  //
+  // Each caller to here sets the opt and shrink levels before, which provides
+  // the right values for us to read. That is, -Os etc. sets the default opt
+  // level, so that the last of -O3 -Os will override the previous default, but
+  // also we note the current opt level for when we run the pass, so that the
+  // sequence -O3 -Os will run -O3 and then -Os, and not -Os twice.
+  void addDefaultOptPasses() {
+    passes.push_back(PassInfo{
+      DEFAULT_OPT_PASSES, passOptions.optimizeLevel, passOptions.shrinkLevel});
```
tlively, line 61: Instead of this magic `DEFAULT_OPT_PASSES` name, would it make sense to just add each of the passes here?

src/tools/optimization-options.h
```diff
@@ -30,7 +30,36 @@ struct OptimizationOptions : public ToolOptions {
...
+      : name(name), optimizeLevel(optimizeLevel), shrinkLevel(shrinkLevel) {}
+  };
+
+  std::vector<PassInfo> passes;
+
+  // Add a request to run all the default opt passes. They are run with the
+  // current opt and shrink levels specified, which are read from passOptions.
+  //
+  // Each caller to here sets the opt and shrink levels before, which provides
+  // the right values for us to read. That is, -Os etc. sets the default opt
```
tlively, line 55: This would be simpler if the levels were arguments to `addDefaultOptPasses`. Is there a reason not to do that?

src/tools/optimization-options.h
```diff
@@ -308,11 +337,25 @@ struct OptimizationOptions : public ToolOptions {
...
+      if (pass.name == DEFAULT_OPT_PASSES) {
         passRunner.addDefaultOptimizationPasses();
       } else {
-        passRunner.add(pass);
+        passRunner.add(pass.name);
       }
+
+      // Revert back to the default levels, if we changed them.
+      passRunner.options.optimizeLevel = oldOptimizeLevel;
+      passRunner.options.shrinkLevel = oldShrinkLevel;
```
tlively, line 358: Can we skip restoring the old levels now that we have per-pass levels? It would be simpler to  not try to mix some concept of global levels with per-pass levels.

src/tools/optimization-options.h
kripken: Perhaps. That would mean refactoring the pass runner logic to be able to give us a list of passes, and then we'd add them. Perhaps that could be a separate refactoring, though? The magic name is a pre-existing thing.

src/tools/optimization-options.h
kripken: That would take more work, I think. That method would need to temporarily override the default opt levels like we do in this PR. But it would need to create a new pass runner to do so, I suspect. At minimum it would be a similar amount of code, if not more. And definitely more if we also need such a capability in a general pass and not just the default opt passes, which I think is possible in the future (then all `add*()` methods would need to get those arguments).

src/tools/optimization-options.h
kripken: I think the global level is unavoidable, see the note in the changelog, since we need `--precompute -O1` to still see the opt level of `1` in precompute. That reads a global opt level.

I suppose we could do 2 passes, first set up the pass names, and while doing so find the default, then go over the passes once more and set their opt levels. But I'm not sure that's simpler.

tlively: While we're changing how things work, can we make it so that `--precompute -O1` runs precompute with the default settings but `-O1 --precompute` runs it with the `-O1` settings?

kripken: I considered that, yeah - in a way it makes sense. But it would be a larger breaking change. I think this PR just fixes behavior that is obviously surprising and wrong, while what you suggest is more debatable. I'd suggest leaving that to a separate discussion, but I'm open to the idea.

aheejin: How many passes do have different behaviors according to their "global level"? I like the direction this PR is going and I think this itself LGTM, but I think that `--precompute` can do different things based on its location is confusing. Does clang have this kind of behavior too? (I don't think so, but it's possible that I'm not aware of it)

If there are not too many passes that have different behaviors depending on the global level, can we just make different options for those options? We already have several passes that have two options like `--inlining` and `--inlining-optimizing`.

aheejin (approved):
I think this PR itself LGTM but the concept of the global level is a little confusing to me (wrote more in https://github.com/WebAssembly/binaryen/pull/5333#discussion_r1044923141) But that should be separately addressed (if we are going to address it) anyway.

src/tools/optimization-options.h
tlively: IMO having the `-O*` flags be non-idempotent is already surprising. Given that repeating them is meaningful, I think it's a much simpler mental model if they only affect optimization levels for later passes. Otherwise they still have the kind of retroactive "time travelling" behavior that made the semantics this PR fixes so confusing.

tlively:
I would rather fix this behavior completely all at once, but if folks feel strongly that we should merge this and keep discussing, we can do that, too.

src/tools/optimization-options.h
kripken: > How many passes do have different behaviors according to their "global level"?

Looks like around 8, currently.

kripken:
I think there are good reasons both ways for the long-term solution here, and I'm not sure which is better. On the one hand, I agree that making the larger change of `-O3` affecting the global level from its position is logically consistent. On the other hand, it has the downside @aheejin mentioned which is that `-O1 --precompute -O3 --precompute` would actually do something different in each `precompute`. That's potentially surprising, but also potentially useful...

I am also a little worried about the larger change being a breaking change for users as it changes the meaning of flags. This PR is also technically a breaking change, but I see it as fixing an obvious bug (that hopefully no one depends on, but it is possible).

Overall I suggest we land this PR for now. With it, `-O3` means "set the global opt level to 3 (as it always has), and also run `-O3` in this position," which is not the ideal model perhaps but it is simple to reason about, and it fixes the largest existing weirdness, since before the PR it was "set the global opt level to 3, and also run opts in this position, the level of whom depends on later flags."

tlively:
Personally, I doubt any users would be broken by the larger change because my guess is that most users (who are not our direct partners) aren't doing anything more complicated than passing a single `-02` or such to wasm-opt.

Assuming there are users who could be broken by the larger change, is there a path toward implementing that change that you would be happy with @kripken? In principle I suppose we could have the new behavior be opt-in or opt-out for a while, but I'm not sure it would be worth it.

src/tools/optimization-options.h
kripken: Hmm, yeah, maybe we should add options for each of those. Like `--precompute-3` for opt level 3, or something nicer? Anyhow, yes, we can leave that for follow-up discussions. I feel that isn't urgent, but this PR is (since it fixes a common situation when optimizing more than once, which with GC gets more common).

kripken:
For the larger change, I like @aheejin 's idea above to just add variations on the 8 or so passes that look at the global opt level. Then we can just remove the global opt level entirely. I think that would address your concerns @tlively ?

But as I said there, I think this PR is somewhat urgent while those things could be left as a TODO for now. What do you think?

src/tools/optimization-options.h
aheejin: I would prefer more descriptive names, like `--dae` and `--dae-optimizing`. (This style of `--something` and `--something-optimizing` looks already being used by many passes anyway.) But yeah, they can be follow-ups.

kripken:
@tlively any objections to landing this?

tlively:
No objections!

PR #5335, opened 2022-12-09
Use non-nullable ref.cast for non-nullable input

tlively:
We switched from emitting the legacy `ref.cast_static` instruction to emitting
`ref.cast null` in #5331, but that wasn't quite correct. The legacy instruction
had polymorphic typing so that its output type was nullable if and only if its
input type was nullable. In contrast, `ref.cast null` always has a a nullable
output type.

Fix our output by instead emitting non-nullable `ref.cast` if the output should
be non-nullable. Parse `ref.cast` in binary and text forms as well. Since the IR
can only represent the legacy polymorphic semantics, disallow unsupported casts
from nullable to non-nullable references or vice versa for now.

### DIFF
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -6915,7 +6915,8 @@ bool WasmBinaryBuilder::maybeVisitRefCast(Expression*& out, uint32_t code) {
     if (ref->type.isRef()) {
       if (code == BinaryConsts::RefCast && ref->type.isNullable()) {
         throwError("ref.cast on nullable input not yet supported");
-      } else if (code == BinaryConsts::RefCastNull && !ref->type.isNullable()) {
+      } else if (code == BinaryConsts::RefCastNull &&
+                 ref->type.isNonNullable()) {
         throwError("ref.cast null on non-nullable input not yet supported");
       }
     }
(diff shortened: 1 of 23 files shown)


tlively:
* main
  * **PR #5335** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5335" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5335?utm_source=stack-comment).

src/wasm/wasm-binary.cpp
```diff
@@ -6904,12 +6904,21 @@ bool WasmBinaryBuilder::maybeVisitRefTest(Expression*& out, uint32_t code) {
...
     bool legacy =
       code == BinaryConsts::RefCastStatic || code == BinaryConsts::RefCastNop;
     auto intendedType = legacy ? getIndexedHeapType() : getHeapType();
     auto* ref = popNonVoidExpression();
+    // Even though we're parsing new instructions, we only support those that
+    // emulate the legacy polymorphic behavior for now.
+    if (ref->type.isRef()) {
+      if (code == BinaryConsts::RefCast && ref->type.isNullable()) {
+        throwError("ref.cast on nullable input not yet supported");
+      } else if (code == BinaryConsts::RefCastNull && !ref->type.isNullable()) {
```
kripken, line 6918:
```suggestion
      } else if (code == BinaryConsts::RefCastNull && ref->type.isNonNullable()) {
```

kripken (approved):

PR #5338, opened 2022-12-12
[Wasm GC] Fix GlobalStructInference on unrefined globals

kripken:
If a global's type is not fully refined, then when `--gsi` replaces a reference with
a `global.get`, we end up with a type that might not be good enough. For example,
if the type is `any` then it is not a subtype of `eq` and we can't do `ref.eq` on it, which
this pass requires. We also can't just do `struct.get` on it if it is a too-distant parent
or such.

Found by the fuzzer.

### DIFF
--- a/src/passes/GlobalStructInference.cpp
+++ b/src/passes/GlobalStructInference.cpp
@@ -123,14 +123,12 @@ struct GlobalStructInference : public Pass {
 
       auto type = global->init->type.getHeapType();
 
-      // The global's type must be a subtype of |eq| for us to do a comparison
-      // check on it later. For example a global declared as type |any| but that
-      // contains (ref $A) is not something we can optimize, as ref.eq on a
-      // global.get of that global will not validate. (This should not be a
-      // problem after GlobalSubtyping runs, which will specialize the type of
-      // the global.)
-      if (!global->type.isRef() ||
-          !HeapType::isSubType(global->type.getHeapType(), HeapType::eq)) {
+      // The global's declared type must match the init's type. If not, say if
+      // we had a global declared as type |any| but that contains (ref $A), then
+      // that is not something we can optimize, as ref.eq on a global.get of
+      // that global will not validate. (This should not be a problem after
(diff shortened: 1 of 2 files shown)


src/passes/GlobalStructInference.cpp
```diff
@@ -123,6 +123,16 @@ struct GlobalStructInference : public Pass {
 
       auto type = global->init->type.getHeapType();
 
+      // The global's declared type must match the init's type. If not, say if
+      // we had a global declared as type |any| but that contains (ref $A), then
+      // that is not something we can optimize, as ref.eq on a global.get of
+      // that global will not validate. (This should not be a problem after
+      // GlobalSubtyping runs, which will specialize the type of the global.)
```
tlively, line 130: Do we need to worry about the case where `GlobalSubtyping` might refine the global to be some LUB, but it is still refined enough that we could optimize it here?

kripken: In practice no, since GSI works in a simple way: it connects globals to struct types by seeing `global g = new Type()`, so a 1:1 correspondence. In theory though one day if it's possible to do a `select` in a global init then we might want to support more.

src/passes/GlobalStructInference.cpp
```diff
@@ -197,14 +207,22 @@ struct GlobalStructInference : public Pass {
           return;
         }
 
+        // We must ignore the case of a non-struct heap type, that is, a bottom
+        // type (which is all that is left after we've already ruled out
+        // unreachable).
+        auto heapType = type.getHeapType();
+        if (!heapType.isStruct()) {
+          return;
+        }
```
tlively, line 216: Can we add a test for this case as well?

tlively (approved):

tlively (commented):
Does it make sense to add a test where the global has a bottom type as well? Maybe not because then it wouldn't be initialized with a `struct.new`, I guess.

kripken:
Yes, we only look at `struct.new`s so a global with a bottom type would be treated like a global with an `i32` and skipped over.

Issue #5245, opened 2022-11-12
Consider running tests with -D_LIBCPP_ENABLE_ASSERTIONS=1


Maximus-:
There's a handful of instances across binaryen/wasm-opt where there is "technically C++ Undefined Behavior". These code defects may be causing nondeterminism, or random failures. 

For example, https://github.com/WebAssembly/binaryen/blob/main/src/wasm/wasm-s-parser.cpp#L1154, `str.size()` is checked to be `>= 3`, and then `str[3]` is accessed. There's a few more instances in wasm-s-parser.cpp, for what it's worth. They should all be easy fixes :) 

Building all of binaryen with -D_LIBCPP_ENABLE_ASSERTIONS=1 should point out these issues pretty quickly, or running the existing fuzzer after enabling this build macro. More info on this build macro: https://libcxx.llvm.org/DesignDocs/DebugMode.html (This is a clang-LLVM only thing).

kripken:
Sounds good to me. A PR adding this for one of our CI test runs would be great. The relevant file for this is `.github/workflows/ci.yml` I believe.

PR #5340, opened 2022-12-13
Add -D_LIBCPP_ENABLE_ASSERTIONS=1 to ASAN/UBSAN CI configurations.

Maximus-:
-D_LIBCPP_ENABLE_ASSERTIONS=1 enables container checking for C++ STL types, like std::vector, std::string, and std::optional. Note, it does not do any validation for iterators. This should address #5245.

I suspect there will be many test failures here before this can be merged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -138,7 +138,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       ASAN_OPTIONS: "symbolize=1"
-      COMPILER_FLAGS: "-fsanitize=address"
+      COMPILER_FLAGS: "-fsanitize=address -D_LIBCPP_ENABLE_ASSERTIONS=1"
     steps:
     - uses: actions/setup-python@v1
       with:
@@ -202,7 +202,7 @@ jobs:
     name: ubsan
     runs-on: ubuntu-latest
     env:
-      COMPILER_FLAGS: "-fsanitize=undefined -fno-sanitize-recover=all"
+      COMPILER_FLAGS: "-fsanitize=undefined -fno-sanitize-recover=all -D_LIBCPP_ENABLE_ASSERTIONS=1"
       CC: "clang"
       CXX: "clang++"
     steps:


kripken:
No errors! But do we need to actually use `libc++` for them to have an effect? I think we use the default on CI, which might be libstdc++?

Maximus-:
Ah yeah, few things going on here. Indeed, LLVM Libc++ isn't installed, so it's not using that one. I went ahead and setup an Ubuntu 22.04 Linux VM to test if the LLVM LIbc++ it ships is even new enough, and it seems it's not. 

```
mxms@test2:~$ cat /etc/os-release 
PRETTY_NAME="Ubuntu 22.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.1 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy

mxms@test2:~$ dpkg --list | grep libc++
ii  libc++-14-dev:amd64             1:14.0.0-1ubuntu1                       amd64        LLVM C++ Standard library (development files)
ii  libc++-dev:amd64                1:14.0-55~exp2                          amd64        LLVM C++ Standard library (development files)
ii  libc++1-14:amd64                1:14.0.0-1ubuntu1                       amd64        LLVM C++ Standard library
ii  libc++abi-14-dev:amd64          1:14.0.0-1ubuntu1                       amd64        LLVM low level support for a standard C++ library (development files)
ii  libc++abi-dev:amd64             1:14.0-55~exp2                          amd64        LLVM low level support for a standard C++ library (development files)
ii  libc++abi1-14:amd64             1:14.0.0-1ubuntu1                       amd64        LLVM low level support for a standard C++ library
```

GitHub Actions Ubuntu-latest is 20.04... so uh, not really sure what the best course of action is here. We can leave this change and wait until GitHub Actions and Ubuntu both upgrade, or we can run the tests locally with the asserts enabled and put them somewhere.

Let me know your thoughts,
Max

kripken:
If we think this can catch real bugs then it might be worth building libc++ from source on CI or another of the options you mention. However, I'm not sure how many bugs to expect it to find. Maybe as a first step you can test this locally and see if anything in the test suite fails for you?

Maximus-:
Yeah that's how I got here - I'll post some bugs here as soon as I get a second :)

PR #5345, opened 2022-12-13
[NFC] Add some notes to pass.h on closed world

kripken:


src/pass.h
```diff
@@ -191,6 +191,26 @@ struct PassOptions {
...
+  // but we also want to keep types of things on the boundary unchanged. For
+  // example, we should not change an exported function's signature, as the
+  // outside may need that type to properly call the export.
+  //   * For now we disallow nontrivial types on the boundary, that is, you
+  //     cannot use a custom struct type as a function parameter type. Such a
+  //     type would not be optimizable due to the constraint to not modify the
+  //     type of the import/export, but it is very simple to use a type in a
+  //     single import/export and then all of its subtypes become unoptimizable,
+  //     which can mean all the types in your module if it is a root type such
+  //     as $java.lang.Object. For that reason we error on using such types on
```
tlively, line 211:
```suggestion
  //     single import/export and then all of its supertypes become unoptimizable,
  //     which can mean all the types in your module if it is a root type such
  //     as $java.lang.Object. For that reason we error on using such types on
```
The $java.lang.Object example would actually be fine. It's a type very deep in the hierarchy that might be a problem, since none of its supertypes would be able to change.

kripken: It's both, isn't it? Imagine that we see that field `$hash` is never used anywhere in all of `$java.lang.Object` and everything else. If `$java.lang.Object` cannot be modified then GTO can't remove it, since subtypes can't have fewer fields.

tlively: Whether a private subtype of a public supertype can be optimized depends on the optimization. Refining a field or removing a field not present in the supertype would be fine, for example. But a private supertype of a public subtype can never be optimized in any way because that would change the identity of the public type.

kripken: Right, fair point, the main issue is that way.

src/pass.h
```diff
@@ -206,11 +206,11 @@ struct PassOptions {
...
   //     type would not be optimizable due to the constraint to not modify the
   //     type of the import/export, but it is very simple to use a type in a
-  //     single import/export and then all of its subtypes become unoptimizable,
-  //     which can mean all the types in your module if it is a root type such
-  //     as $java.lang.Object. For that reason we error on using such types on
-  //     the boundary for now. Instead, use basic types like anyref, externref,
-  //     etc.
+  //     single import/export and then all of its supertypes become
+  //     unoptimizable, which can mean all the types in your module if it is a
+  //     root type such as $java.lang.Object. For that reason we error on using
```
tlively, line 211: The java.lang.Object example doesn't fit with the previous sentence because it doesn't have supertypes.

kripken: Right, thanks, rewritten.

tlively (approved):

PR #5346, opened 2022-12-13
Add memory: init, copy, fill support to Multi-Memory Lowering Pass 

ashleynh:
This PR adds support for [memory.init](https://webassembly.github.io/spec/core/exec/instructions.html#xref-syntax-instructions-syntax-instr-memory-mathsf-memory-init-x), [memory.copy](https://webassembly.github.io/spec/core/exec/instructions.html#xref-syntax-instructions-syntax-instr-memory-mathsf-memory-init-x), and [memory.fill](https://webassembly.github.io/spec/core/exec/instructions.html#xref-syntax-instructions-syntax-instr-memory-mathsf-memory-fill) instructions in the multi-memory lowering pass. Also includes optional bounds checks per the wasm spec guidelines.

Note: This PR depends on the refactor from https://github.com/WebAssembly/binaryen/pull/5352 landing first.

### DIFF
--- a/src/passes/MultiMemoryLowering.cpp
+++ b/src/passes/MultiMemoryLowering.cpp
@@ -210,7 +210,8 @@ struct MultiMemoryLowering : public Pass {
           builder.makeLocalGet(destIdx, parent.pointerType),
           builder.makeLocalGet(sizeIdx, parent.pointerType),
           memory);
-        std::vector<Expression*> exprs = {destSet, localSet, sizeSet, boundsCheck};
+        std::vector<Expression*> exprs = {
+          destSet, localSet, sizeSet, boundsCheck};
         if (additionalCheck) {
           exprs.push_back(additionalCheck);
         }
@@ -222,7 +223,9 @@ struct MultiMemoryLowering : public Pass {
       return destValue;
     }
 
-    Expression* getSource(MemoryCopy* curr, Index sizeIdx = Index(-1), Index sourceIdx = Index(-1)) {
+    Expression* getSource(MemoryCopy* curr,
+                          Index sizeIdx = Index(-1),
+                          Index sourceIdx = Index(-1)) {
       Expression* sourceValue =
         addOffsetGlobal(curr->source, curr->sourceMemory);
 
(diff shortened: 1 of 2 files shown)


src/passes/MultiMemoryLowering.cpp
```diff
@@ -115,100 +115,174 @@ struct MultiMemoryLowering : public Pass {
       replaceCurrent(builder.makeCall(funcName, {}, curr->type));
     }
 
+    Expression* makeGtuTrap(Expression* leftOperand,
```
tlively, line 118: It might be good to add "Add" to the name because it adds before doing the comparison.
```suggestion
    Expression* makeAddGtuTrap(Expression* leftOperand,
```

src/passes/MultiMemoryLowering.cpp
```diff
@@ -115,100 +115,174 @@ struct MultiMemoryLowering : public Pass {
...
+                            Expression* rightOperand,
+                            Expression* greaterThanCondition) {
```
tlively, line 120: Maybe `limit` instead of `greaterThanCondition`, since it's not a complete condition (like a gtu expression) by itself.

tlively (commented):
It would be helpful to split this into a non-functional change with just the refactoring followed by another change adding the new functionality.

src/passes/MultiMemoryLowering.cpp
ashleynh: Makes sense, thanks!

src/passes/MultiMemoryLowering.cpp
ashleynh: Thank you! Needed naming help here.

test/lit/passes/multi-memory-lowering.wast
```diff
@@ -628,6 +628,182 @@
    )
   )
   )
+
+  ;; CHECK:      (func $memory.fill
+  ;; CHECK-NEXT:  (memory.fill
+  ;; CHECK-NEXT:   (i32.const 0)
+  ;; CHECK-NEXT:   (i32.const 1)
+  ;; CHECK-NEXT:   (i32.const 2)
```
kripken, line 636: The order of these three instructions (0, 1, 2) should remain the same (imagine they are three calls, that have side effects, so they have to keep running in the same order). But atm the order in the output for bounds checks looks like `2, 0, 1`.

If there isn't a better way, a simple approach for stuff like this is to allocate 3 local vars up front and write the inputs to them immediately, and afterwards use new LocalGets of their indexes.

kripken (approved):

PR #5347, opened 2022-12-14
Do not optimize public types

tlively:
Do not optimize or modify public heap types in any way. Public heap types
include the types of imported or exported functions, tables, globals, etc. This
is important to maintain the public interface of a module and ensure it can
still link interact as intended with the outside world.

Also add validation error if we find any nontrivial public types that are not
the types of imported or exported functions. This error is meant to help the
user ensure that type optimizations are not silently inhibited. In the future,
we may want to add options to silence this error or downgrade it to a warning.

This commit only updates the type updating machinery to avoid updating public
types. It does not update any optimization passes accordingly. Since we avoid
modifying public signature types already, this is not expected to break
anything, but in the future once we have function subtyping or if we make the
error optional, we may have to update some of our optimization passes.


tlively:
* main
  * **PR #5347** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5347" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5347?utm_source=stack-comment).

tlively:
@kripken, can I hand this off to you to update the type optimization passes as necessary using `ModuleUtils::getPublicHeapTypes` and `ModuleUtils::getPrivateHeapTypes` to determine which types to optimize? Perhaps we can collaboratively create a stack of PRs on top of this one and then merge them all at once.

kripken:
Sounds good, I'll look into that tomorrow.

src/ir/module-utils.cpp
```diff
@@ -198,10 +198,92 @@ void setIndices(IndexedHeapTypes& indexedTypes) {
...
+  InsertOrderedSet<HeapType> publicTypes;
+
+  auto notePublic = [&](HeapType type) {
+    // TODO: Gracefully handle public types with arbitrary public children here
+    // and leave it to separate code to handle erroring or warning if
+    // optimizations would be inhibited.
+    if (type.isBasic()) {
+      return;
+    }
+    // All the rec group members are public as well.
```
kripken, line 211: In the common case wouldn't that make all types public, if they are all in one big rec group?

src/ir/module-utils.cpp
```diff
@@ -198,10 +198,92 @@ void setIndices(IndexedHeapTypes& indexedTypes) {
...
+      // did collect them.
+      for (auto child : member.getReferencedHeapTypes()) {
+        if (!child.isBasic()) {
+          std::string typeName = child.toString();
+          if (auto it = wasm.typeNames.find(child);
+              it != wasm.typeNames.end()) {
+            typeName = it->second.name.toString();
+          }
+          Fatal() << "nontrivial public type found during type optimizations: "
+                  << typeName;
```
kripken, line 229: Shouldn't this only error in closed-world?

src/ir/module-utils.cpp
tlively: Yes. Consumers should not put their public function types in the same rec groups as their private, internal types.

src/ir/module-utils.cpp
tlively: In open world we currently do not optimize any types anyway, so for now this code path should never be reached.

kripken:
Actually, can this land before the followup work to use it? It seems like it could be a standalone PR. Then I can land independent PRs to use it in passes later.

Separately though I was thinking we'd want the validator to also check this condition, and it would be good to add tests for that.

src/ir/module-utils.cpp
kripken: I see, makes sense.

src/ir/module-utils.cpp
kripken: I see. What happens in nominal mode, though? Don't we emulate nominal using a single big rec group in the output?

tlively: Yes, in the output, but the internal representation in nominal mode only uses singleton rec groups.

tlively:
I don't think this can land independently because it will break passes like GTO that may try to optimize subtype chains rooted in public types in ways that would produce invalid types when the public supertypes are silently not updated.

src/ir/module-utils.cpp
kripken: Wouldn't this workflow break then: optimize in nominal mode, then write a binary, then read and optimize that binary in normal mode?

kripken:
If we have a validation error then I think GTO would not break? No chain should be rooted in a public type in closed-world, that is an error.

src/ir/module-utils.cpp
tlively: Yes, that would break. But it won't be a problem once we delete nominal mode (as long as producers update to not use single huge rec groups).

tlively:
> If we have a validation error then I think GTO would not break? No chain should be rooted in a public type in closed-world, that is an error.

No, this would not be an error, at least not with how I've understood the errors we've discussed. There could be an exported global with a struct type with only trivial fields. Since it doesn't spread its publicness to other types, that wouldn't be an error in this PR, but it could have arbitrary private subtypes that would get optimized.

kripken:
My understanding of our conversation was that this should error:
```wat
(module
  (struct $A ..)
  (global $A (ref $A) (..))
  (export "A" (global $A))
)
```
`$A` is a public type because it is on the boundary (an export). It is a custom user type. And the idea was that custom types on the boundary would error in closed-world, and instead people should be using `externref / anyref` etc. (or use open-world).

That seems simple and good enough, but I might be missing something?

tlively:
I see, so we aim to allow one level of user-definition for the types of functions on the boundary and zero levels of user-definition for globals. What about for tables and tags? Or maybe we just want to error if there are any public struct or array types. But what about MVP func types with subtypes? I guess once we implement final types, we won't need to worry about that because MVP func types won't be able to have subtypes.

The more I think about it, the less I think these naive restrictions make sense. What if I have this module:

```
(module 
  (type $A (func))
  (type $B (func (param (ref $A))))
  (func (export "a") (type $A) ...)
  (func (export "b") (type $B) ...)
)
```

Under our rules this would error out because `$B` refers to another user-defined type. But that type, `$A`, is already public anyway, so it doesn't make any sense to error out here.

Can we come up with a consistent set of rules?

kripken:
I would error in that last case. I do see that it is somewhat ironic, but for consistency erroring seems simple and I don't know that it's a use case we need to support in closed world.

I see the rule as pretty simple here: **Can the outside get or receive a user-defined type**. For a global, that means whether the global has such a type, since that is how users interact with globals. For a function, that means whether a parameter or return has such a type, since that is how users get and receive values when calling a function.

Ruling those cases out leaves closed-world in the situation where stuff can escape to the outside, but only as an opaque type (and the user promises not to inspect the contents of that opaque type). Escaping objects can be cached on the outside and passed back in etc. but that is all safe in closed-world.

tlively:
I've implemented the changes we talked about earlier and added a test. The PR description is updated as well.

src/tools/wasm-opt.cpp
```diff
@@ -359,7 +363,11 @@ int main(int argc, const char* argv[]) {
     auto runPasses = [&]() {
       options.runPasses(wasm);
       if (options.passOptions.validate) {
-        bool valid = WasmValidator().validate(wasm);
+        WasmValidator::Flags flags = WasmValidator::Globally;
+        if (options.passOptions.closedWorld) {
+          flags |= WasmValidator::ClosedWorld;
```
kripken, line 368: This pattern repeats a bit - I wonder if the `WasmValidator` constructor could perhaps take as a parameter a `PassOptions` and read what it needs from there? But anyhow that could be a followup.

src/wasm/wasm-validator.cpp
```diff
@@ -3500,6 +3501,32 @@ static void validateFeatures(Module& module, ValidationInfo& info) {
   }
 }
 
+static void validateClosedWorldInterface(Module& module, ValidationInfo& info) {
+  // Error if there are any publicly exposed heap types beyond the types of
+  // publicly exposed functions.
```
kripken, line 3506: Do we explain why this is anywhere? Might be worth adding a comment on the reasons we decided to define things that way, either here or in `pass.h`.

test/lit/passes/remove-unused-module-elements-refs.wast
```diff
@@ -1,22 +1,14 @@
 ;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
-;; RUN: foreach %s %t wasm-opt --remove-unused-module-elements --closed-world --nominal -all -S -o - | filecheck %s
-;; RUN: foreach %s %t wasm-opt --remove-unused-module-elements                --nominal -all -S -o - | filecheck %s --check-prefix OPEN_WORLD
 
-;; Test both open world (default) and closed world. In a closed world we can do
-;; more with function refs, as we assume nothing calls them on the outside, so
-;; if no calls exist to the right type, the function is not reached.
+;; RUN: foreach %s %t wasm-opt --remove-unused-module-elements --nominal -all -S -o - | filecheck %s
```
kripken, line 3: Was closed-world removed here because the test doesn't pass validation any more? I think it probably still had useful things though. Maybe worth making a new test file for closed world that removes just the non-validating stuff? Otherwise, where are we testing the closed-world-specific code paths in this pass?

test/lit/passes/remove-unused-module-elements-refs.wast
```diff
@@ -96,11 +64,8 @@
   )
 
   ;; CHECK:      (func $target-B (type $B)
-  ;; CHECK-NEXT:  (unreachable)
+  ;; CHECK-NEXT:  (nop)
```
kripken, line 67: Why did these change?

kripken (commented):
lgtm % comments

src/wasm/wasm-validator.cpp
tlively: Would it make sense to explain this in the `--help` text for `--closed-world`?

test/lit/passes/remove-unused-module-elements-refs.wast
tlively: That's right. Will create a separate test for the closed-world-specific parts.

test/lit/passes/remove-unused-module-elements-refs.wast
tlively: Now that we're no longer passing --closed-world, we assume the reference to `target-B` may be called from the outside, so we no longer empty the function. I'll move this test to a separate file where we can still pass --closed-world.

src/wasm/wasm-validator.cpp
kripken: I think the `--help` text is usually pretty brief. We've been leaving longer details to places like `pass.h` for trapsNeverHappen etc. Though that isn't a great place, really. Maybe the main markdown page is better, which has a section on the IR? Anyhow, for now I'd say let's either add to `pass.h` as before, or to the markdown.

tlively:
@kripken, all the feedback should have been addressed now.

test/lit/passes/remove-unused-module-elements-closed-world.wast
```diff
@@ -0,0 +1,32 @@
...
+  ;; CHECK-NEXT:   (ref.func $unused)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $exported (export "foo")
+    ;; This pass can't prove that this reference won't escape.
+    (drop
+      (ref.func $unused)
+    )
+  )
+)
```
kripken, line 32: Is this single test enough? The old file had several places that checked the ability of closed-world to optimize things like this. I think it had various corner cases there.

test/lit/passes/remove-unused-module-elements-refs.wast
```diff
@@ -96,63 +62,40 @@
...
-  ;; OPEN_WORLD-NEXT:  )
-  ;; OPEN_WORLD-NEXT:  (drop
-  ;; OPEN_WORLD-NEXT:   (ref.func $target-A)
-  ;; OPEN_WORLD-NEXT:  )
-  ;; OPEN_WORLD-NEXT: )
-  (func $foo (export "foo") (param $A (ref null $A))
+  (func $foo (export "foo")
     (call_ref $A
-      (local.get $A)
+      (ref.null $A)
```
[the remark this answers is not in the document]
kripken: (For closed world we can't have the param, but this file is now for open world only, so it should be ok to leave it?)

kripken (commented):
Last questions are just about that one test being split.

askeksa-google:
I tested this change with dart2wasm, with `wasm-opt -all --closed-world -tnh -O3 --gufa -O3`.

At first, I got a ton of errors about publicly exposed types. I thought it was probably because some exported functions had types defined in the same rec group as those GC types. I moved the function types for the exported functions to singleton rec groups, after which I get this error:
```
Fatal: Internal GlobalTypeRewriter build error: Heap type has an undeclared supertype at index 2
```
on this module: [ffi_native_test.wasm.gz](https://github.com/WebAssembly/binaryen/files/10245007/ffi_native_test.wasm.gz)

tlively:
Thanks, @askeksa-google, will investigate.

test/lit/passes/remove-unused-module-elements-refs.wast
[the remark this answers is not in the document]
tlively: Ah, this is a mistake. I started by initially reverting the test to before we updated RemoveUnusedModuleElements for open world, and it looks like I missed fixing this up afterward.

test/lit/passes/remove-unused-module-elements-closed-world.wast
tlively: I think this test is enough because the only place we use `closedWorld` in RemoveUnusedModuleElements is to determine what to do in precisely this situation.

kripken: Skimming it now, I see we check various orderings of how `RefFunc`s appear interleaved with `CallRef`s. We also have things like `Verify that we do not crash on an unreachable call_ref` that I think is worth checking in both modes. So I do think we want to basically convert this file to work in closed world rather than add a single tailored test. If you want I can look into that.

tlively: Ah, right, because all these tests predate support for _open_ world, not closed world. I can fix it up.

tlively:
@askeksa-google that problem is fixed now. It was a real bug with this PR that the test suite didn't catch because it overwhelmingly runs in --nominal mode.

kripken (approved):

kripken:
The fuzzer fails quite quickly on this,
```
[wasm-validator error in module] publicly exposed type disallowed with a closed world: $(struct (field (mut i32))), on 
```
Do we have a plan for that? We can disable fuzzing closed world for now, but that's the main mode our users use...

kripken:
The fuzzer found a failure that starts with this landing:

```wat
(module
 (type $super (array i32))
 (type $sub (array_subtype i32 $super))

 (func $export (export "export") (result (ref $super))
  (array.new $sub
   (i32.const 1)
   (i32.const 1)
  )
 )
 (func $helper (param $x i32) ;; This signature is prunable, causing type edits.
  (unreachable)
 )
)
```

```
$ bin/wasm-opt -all  --signature-pruning b.wat --nominal
[wasm-validator error in function 0] function body type must match, if function returns, on 
(array.new_elem $type$1 0
 (i32.const 1)
 (i32.const -1)
)
Fatal: error after opts
```

Looks like somehow after rewriting types we end up without the array types not being subtypes any more..?

tlively:
That should be worked around by #5367.

PR #5351, opened 2022-12-14
Properly use pass options in nested pass runners

kripken:
This fixes a TODO.

There is a runtime cost to this in higher opt levels, as passing through `-O3`
makes nested optimization work take longer. But it can lead to better results,
as the test diff here shows. As in `-O3` we are ok to take longer to run in
return for that, I think it's the right tradeoff. The only ways to avoid a compile-
time regression here are to manually add places that lower the opt level when
creating a nested pass runner, which we might consider if we get reports of
problems, but those seem like hacks.

In measurements, I see no difference on say `-O1`, but `-O3` is 6% slower to
compile on Poppler for example.


kripken:
Alternatively, if we are worried about compile time here, I could add changes to lower the opt level in nested runners in relevant places.

tlively (approved):

kripken:
Maybe another option here is to reduce the opt level in nested runners. Reducing it to a maximum of 1 avoids any regression, I see. It also wouldn't give any new benefits, but maybe that's the safer route, I pushed that now. Note that there are no test changes now.

With the new commit, what this change does is propagate options with a limited opt/shrink level. That still propagates options like trapsNeverHappen, and still allows an opt level of more than 0 (before we had 0), but just to a max of 1. 1 makes sense as we have predictable linear time there, which is reasonable for a nested runner I think.

tlively:
How are nested runners typically used? Propagating the options unmodified makes more sense to me, but maybe nested runners are typically used in places where automatically capping the optimization level at 1 makes sense?

kripken:
A typical use case is a pass that runs another pass internally, or even the entire set of passes. For example `simplify-globals` will find functions it propagated a constant into a `global.get` and then re-optimize those functions. Doing that at level 3 is potentially quite slow, so 1 is kind of a compromise. I do think that is the general pattern for nested pass runners: they have less urgency to run all the possible passes at full strength.

I can leave a TODO here to explore raising the levels all the way. For now though this just raises from 0 to 1, which has very low risk of compile-time regressions compared to 2 or 3 (where we tolerate nonlinear compile times).

tlively:
Ok, capping a 1 with a TODO to consider further sounds good to me.

kripken:
Turns out that even going 0=>1 uncovers at least one fuzzer bug, fixed in the above commit. Now that we actually optimize in simplify-globals, we need to handle the case of the optimizer there seeing a call with the effects of writing to a global (we had an assert on that never happening, but with function effects that can happen, and so a GlobalSet is not the only thing that can set a global).

tlively (approved):

PR #5356, opened 2022-12-15
In --debug mode, print partial wasm data that was read

### LINKED ISSUE #5344: Parse errors during `wasm-opt --print` should print out wat leading up to the parse error
If `wasm-opt --print` encounters some kind of parse error (an opcode it doesn't recognize), it just bails and gives a single parse error. It would be really really nice if it would at least print out everything it had parsed up to that point anyway. This would help in a couple situations:

1) This would help people writing code generators localize their issue better, where exactly wasm-opt is unable to parse
2) For those of us working on the absolute bleeding edge, we are sometimes emitting opcodes that binaryen does not yet recognize, but it would be really helpful to at least be able to look at the types sections, imports/exports, globals and how they are encoded.

kripken:
If `wasm-opt` or `wasm-dis` are given an invalid binary, after the error
message we can also print out the wasm we did manage to read. That
includes global stuff like imports and also all the functions up until
there. This can help debugging in some situations.

Only do this when `--debug` is passed as it can be very verbose and
in general users might not want it.

This is technically easy to do, it turns out, since we already use a
thrown exception on an error in parsing, and we fill up the wasm as
we go, so it just contains what we've read so far, and we can just
print it.

Fixes #5344


test/lit/binary/debug-bad-binary.test
```diff
@@ -0,0 +1,42 @@
+# Verify --debug mode prints partial results even after an error.
```
tlively, line 1: I think we generally use `;;` comment syntax even in .test files.

tlively (approved):

test/lit/binary/debug-bad-binary.test
kripken: Ah, I copied it from https://github.com/WebAssembly/binaryen/blob/main/test/lit/binary/stacky-eh.test - should I change that too?

tlively: It's not that important. I suspected we might not be consistent here...

kripken: Ah, already done... a small step towards eventual consistency.

PR #5358, opened 2022-12-16
[Wasm GC] Optimize away null arms that would trap

kripken:
E.g.
```wat
(struct.get
  (select
    (ref.null ..)
    (something)
    (condition)
  )
)
```

If traps-never-happen then this can be

```wat
(drop (condition))
(struct.get
  (something)
)
```

That is, we can remove the arm that is null, as it would trap but traps are
assumed to not happen.

This pattern may not be very common now, but a new optimization I want
to add soon will start to emit it, so it will be good to optimize it well.

Also fix a bug this uncovers on `struct.set` on a null type.


src/passes/OptimizeInstructions.cpp
```diff
@@ -1513,10 +1513,74 @@ struct OptimizeInstructions
...
+      //
+      // That is, we will by assumption not read from the null, so remove that
+      // arm.
+      //
+      // TODO We could recurse here.
+      // TODO We could do similar things for casts (rule out an impossible arm).
+      if (auto* iff = ref->dynCast<If>()) {
+        if (iff->ifFalse) {
+          if (iff->ifTrue->type.isNull()) {
+            ref = builder.makeBlock({builder.makeDrop(iff->condition),
```
tlively, line 1549: If we had an `assert` or `assume` pseudoinstruction or intrinsic, we could even preserve and propagate knowledge about the condition's value. Maybe worth mentioning in a TODO?

src/passes/OptimizeInstructions.cpp
```diff
@@ -1513,10 +1513,74 @@ struct OptimizeInstructions
...
+                                     builder.makeDrop(iff->ifTrue),
```
tlively, line 1550: The `ifTrue` block should not be included because we're assuming it would never be executed at runtime. This would be different for a `select`.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1513,10 +1513,74 @@ struct OptimizeInstructions
...
+                                     iff->ifFalse});
+            return false;
+          }
+          if (iff->ifFalse->type.isNull()) {
+            ref = builder.makeSequence(
+              builder.makeDrop(iff->condition),
+              getResultOfFirst(iff->ifTrue, builder.makeDrop(iff->ifFalse)));
```
tlively, line 1557: Same here: no need to include `ifFalse`.

src/passes/OptimizeInstructions.cpp
kripken: Interesting, yeah, added a TODO to think about that in this context.

src/passes/OptimizeInstructions.cpp
kripken: Good catch... I missed it in copy-paste, and the fuzzer somehow missed it too... fixed.

tlively (approved):

PR #5359, opened 2022-12-16
Fixed fuzzing of closed-world after #5347

kripken:
A initial content testcase may only work in open world, so check for that
using the existing mechanism of checking if such testcases work with out
feature flags.


tlively (approved):
LGTM if it works!

kripken:
Several thousand iterations look good, and I manually validated that we still check both open and closed world, with and without initial contents.

PR #5360, opened 2022-12-16
Fuzzing: Work around canonicalization of function types in fuzzer for closed world

kripken:
If the canonical heap type for a signature is private, then if we create a new
function with that type and export it we'd be exporting a private type, and
error in closed world.

The error is slightly more complex, in that that function type itself will
actually be accepted in closed world since it ignores types of exported
functions. But subtypes and supertypes of that type can cause problems.

To avoid that, create fresh types in the fuzzer.

### DIFF
--- a/src/tools/fuzzing/fuzzing.cpp
+++ b/src/tools/fuzzing/fuzzing.cpp
@@ -23,6 +23,22 @@ namespace wasm {
 
 namespace {
 
+// Construct a fresh signature type for a given signature, and return its heap
+// type. This is useful when we do not want to get the default behavior of
+// reusing the "canonical" heap type for a signature, which can cause problems
+// in closed-world mode - in that mode the canonical heap type may be private,
+// so we should use a fresh type whenever we construct a function that will be
+// exported.
+static HeapType getFreshSignatureHeapType(Signature sig) {
+  TypeBuilder builder(1);
+  builder.setHeapType(0, sig);
+  return (*builder.build())[0];
+}
+
+static HeapType getFreshSignatureHeapType(Type params, Type results) {
+  return getFreshSignatureHeapType(Signature(params, results));
+}
+
 // Weighting for the core make* methods. Some nodes are important enough that
 // we should do them quite often.
 
@@ -254,8 +270,11 @@ void TranslateToFuzzReader::setupMemory() {
   }
   contents.push_back(builder.makeLocalGet(0, Type::i32));
   auto* body = builder.makeBlock(contents);
-  auto* hasher = wasm.addFunction(builder.makeFunction(
-    "hashMemory", Signature(Type::none, Type::i32), {Type::i32}, body));
+  auto* hasher = wasm.addFunction(
+    builder.makeFunction("hashMemory",
+                         getFreshSignatureHeapType(Type::none, Type::i32),
+                         {Type::i32},
+                         body));
   wasm.addExport(
     builder.makeExport(hasher->name, hasher->name, ExternalKind::Function));
   // Export memory so JS fuzzing can use it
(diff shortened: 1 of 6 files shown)


src/tools/fuzzing/fuzzing.cpp
```diff
@@ -23,6 +23,22 @@ namespace wasm {
...
+// Construct a fresh signature type for a given signature, and return its heap
+// type. This is useful when we do not want to get the default behavior of
+// reusing the "canonical" heap type for a signature, which can cause problems
+// in closed-world mode - in that mode the canonical heap type may be private,
+// so we should use a fresh type whenever we construct a function that will be
+// exported.
+static HeapType getFreshSignatureHeapType(Signature sig) {
+  TypeBuilder builder(1);
+  builder.setHeapType(0, sig);
+  return (*builder.build())[0];
```
tlively, line 35: This will usually not create a fresh type due to canonicalization, although it may happen to create a new type the first time it is called.

tlively (commented):
Can you explain more about what problem this is trying to solve?

kripken:
This is the initial contents that led to the error:
```wat
(module
  (type $sub (func_subtype $super))

  (type $super (func_subtype func))

  (global $g (ref null $super) (ref.null $sub))
)
```

The fuzzer then ends up creating lots of functions that have type `$sub`. That then errors on `$super` in the closed-world validation logic. `$super` is not the type of an exported function, so it is not ignored there. But `$sub` is a public type, and so all its referred types are public too, which means `$super` is. So we error.

src/tools/fuzzing/fuzzing.cpp
kripken: Ah, the error happened in nominal mode, so I guess I just checked there. It did work in that case.

I guess for the isorecursive case we need to do... something more? But as we fuzz only nominal for now, this at least unblocks us right now.

tlively:
The error is working as intended, at least. Can we update the fuzzer to not create public functions with function types that have supertypes? We can try to extend it more precisely later.

A fix that only works in nominal mode isn't very attractive since the nominal type system will go away as soon as I get around to it.

kripken:
> Can we update the fuzzer to not create public functions with function types that have supertypes?

I don't see how? If it just skips creating such exported functions it might end up creating no functions at all, or very few. That's the case in this particular bug, most of the exports ended up getting that type `$sub`.

> A fix that only works in nominal mode isn't very attractive since the nominal type system will go away as soon as I get around to it.

At that time we can extend `getFreshSignatureHeapType`, I think? We need something like that anyhow for optimization passes eventually - right now we just error in MergeTypes, but could use this there. (The update could create larger / more complicated rec groups until it succeeds to make a new type, or such.)

tlively:
Maybe if it tries to make a function with a function type that has a supertype, just make a function with the topmost supertype instead?

kripken:
Hmm, interesting, that might work around this.

Meanwhile the fuzzer found another similar issue: LegalizeJSInterface can also run into the same problem. Basically any pass that generates exports and uses the default function type for those exports can hit it, not just the fuzzer itself.

We could add a utility that all these passes call, I suppose. Another option might be for closed world to automatically "close" the world when it runs, like in a new pass or during binary writing (that is, to apply the workaround there). But I'm not sure either is good - thoughts?

kripken:
I pushed an attempt to fix this using supers, and with code callable from various places.

tlively:
> Basically any pass that generates exports and uses the default function type for those exports can hit it

Is this in nominal mode or isorecursive mode? Also, I don't see how this can be the same issue the fuzzer found; "default" function types (i.e. the ones produced by the `HeapType(Signature)` constructor can never have super types in either mode afaict. (This will not be true in nominal mode once I implement proper function subtyping, though, so it would be good to delete nominal mode before then.)

kripken:
Chatting offline, the key issue is that in nominal mode a subtype can appear before its super. So the subtype can be the canonical type for a signature. Then when we reuse it in new exports etc. we not only make it public but also its supertype, and the latter ends up not validating in closed world (the subtype validates since it is an exported function type, which we must allow).

But, this is superseded by #5364 as suggested by @tlively as a more general fix.

We may need this PR in the future though, or at least something in its general current shape, which is to add a "make this validate in closed world" method that passes might need, which can involve making new fresh types etc. But for now we can close it.

PR #5361, opened 2022-12-17
Remove unused types during type optimizations

tlively:
The type rewriting utility in type-updating.cpp gathers all the used heap types,
then rewrites them to newly built and possibly modified heap types. The problem
is that for the isorecursive type system, the set of "used" heap types was
overly broad because it also included unused heap types that are in a rec group
with used types. In the context of emitting a binary, it is important to treat
these types as used because failing to emit them would change the identity of
the used types, but in the context of type optimizations it is ok to treat them
as truly unused because we are changing type identities anyway.

Update the type rewriting utility to only include truly used types in the set of
output types. This causes all existing type optimizations to implicitly drop
unused types, but only if they find any other optimizations to do and actually
run the rewriter utitility. Their output will also still include unused types
that were used before their optimizations were applied.

To overcome these limitations and better match the optimizing power of nominal
mode, which never includes unused types in the output, add a new type
optimization pass that removes unused types and does nothing else and run it
near the end of the global optimization pipeline.


tlively:
* main
  * **PR #5361** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5361" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5361?utm_source=stack-comment).

src/passes/RemoveUnusedTypes.cpp
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright 2021 WebAssembly Community Group participants
```
kripken, line 2:
```suggestion
 * Copyright 2022 WebAssembly Community Group participants
```
Almost 2023 :exploding_head:

test/lit/passes/signature-refining_gto.wat
```diff
@@ -1,8 +1,10 @@
-;; RUN: foreach %s %t wasm-opt --nominal --signature-refining --gto --roundtrip -all -S -o - | filecheck %s
+;; RUN: foreach %s %t wasm-opt --nominal --signature-refining --gto                       --roundtrip -all -S -o - | filecheck %s
+;; RUN: foreach %s %t wasm-opt           --signature-refining --gto --remove-unused-types --roundtrip -all -S -o - | filecheck %s --check-prefix ISOREC
```
kripken, line 2: Please add a comment as to what this is testing (contrasted to nominal on the line before).

src/passes/pass.cpp
```diff
@@ -620,6 +623,7 @@ void PassRunner::addDefaultGlobalOptimizationPrePasses() {
     }
     addIfNoDWARFIssues("remove-unused-module-elements");
     if (options.closedWorld) {
+      addIfNoDWARFIssues("remove-unused-types");
```
kripken, line 626: Can we skip adding this pass in nominal mode?

tlively: Sure, but after we delete nominal mode it won't have made a difference.

test/lit/passes/signature-refining_gto.wat
```diff
@@ -1,8 +1,10 @@
...
 
 (module
  ;; The type $A should not be emitted at all (see below).
  ;; CHECK-NOT: (type $A
+ ;; ISOREC-NOT: (type $A
```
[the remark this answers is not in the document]
tlively: The auto-update script currently removes any `XXX-NOT` lines, but those are an important part of this test. Maybe we could update the script to preserve those lines instead.

src/passes/pass.cpp
kripken: Right, fair point.

kripken:
lgtm with comments on the issues mentioned.

kripken (approved):

PR #5362, opened 2022-12-18
Multi-Memory Only Wasm-Split

ashleynh:
This PR removes the "global" and "in-memory" options from the wasm-split tool's instrumenter, leaving secondary memory as the only option for profile-data collection. After the instrumenter creates and uses the secondary memory, the multi-memory lowering pass will be run to lower the memories into a single combined memory.


ashleynh:
Tests are failing because https://github.com/WebAssembly/binaryen/pull/5363 needs to be submitted first.

ashleynh:
One emscripten wasm-split test is failing, `other.test_split_main_module`:

`failed to asynchronously prepare wasm: LinkError: WebAssembly.instantiate(): memory import 14 is smaller than initial 257, got 256`

PR #5363, opened 2022-12-19
Maintain first memory import/export in Multi-Memory Lowering Pass

ashleynh:
This PR maintains the first memory's import/export in the single combined memory after multi-memories are lowered.


tlively (commented):
Looks good, but it would be good to add a test. It may also be worth considering whether we should error out if more than one original memory is exported because in that case the host would have to use the offset globals to access at least one of the memories and we probably don't want to have to support that right now.

Relatedly, do we error out if there are imported memories? If not, we should.

test/lit/wasm-split/multi-memory-lowering-export.wast
```diff
@@ -0,0 +1,9 @@
+;; RUN: wasm-opt %s --enable-multi-memories --multi-memory-lowering --enable-bulk-memory --enable-extended-const --enable-simd --enable-threads -S -o - | filecheck %s
```
tlively, line 1: For the non-error tests, it would be good to put the magic comment at top so they can be auto-updated.

tlively (approved):
LGTM % comments

PR #5364, opened 2022-12-19
[Wasm GC] Do not cache signature types in nominal mode if they have a super

kripken:
This reduces the amount of public types, since if there is a super then using the
type in a public place would make the super also public. It is safer for closed-world
mode to reuse types without supers.


test/lit/parse-nominal-types-no-sig-sharing.wast
```diff
@@ -0,0 +1,39 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
+
+;; RUN: foreach %s %t wasm-opt --nominal -all -S -o - | filecheck %s
+;; RUN: foreach %s %t wasm-opt --nominal -all --roundtrip -S -o - | filecheck %s
```
tlively, line 4: Do these need --closed-world?

tlively (approved):

test/lit/parse-nominal-types-no-sig-sharing.wast
kripken: Ah, with closed world we'd also get a validation check here, good idea. Adding.

kripken:
(this passed over 10K fuzzer iterations, so hopefully this fixes this entire issue)

PR #5365, opened 2022-12-20
[NFC] Add `TypeBuilder::buildRelaxed`

tlively:
Add a new method for building types that implicitly creates a single large rec
group and does not require that the supertypes appear before their subtypes. Use
this method to simplify type-updating.cpp, which does not really care about type
validation. This is also a step toward removing the nominal type system, since
using `buildRelaxed` from the text and binary parsers will emulate the current
nominal behavior even if the internal type representation is isorecursive.

### DIFF
--- a/src/wasm-type.h
+++ b/src/wasm-type.h
@@ -647,11 +647,12 @@ struct TypeBuilder {
   };
 
   // Returns all of the newly constructed heap types. May only be called once
-  // all of the heap types have been initialized with `setHeapType`. In nominal
-  // mode, all of the constructed HeapTypes will be fresh and distinct. In
-  // nominal mode, will also produce a fatal error if the declared subtype
-  // relationships are not valid.
+  // all of the heap types have been initialized with `setHeapType`. In relaxed
+  // mode, the types are allowed to appear in any order and be arbitrarily
+  // mututally recursive because they will be implicitly reordered and put into
+  // a single rec group, so there should not be any existing rec groups.
   BuildResult build();
+  BuildResult buildRelaxed();
 
   // Utility for ergonomically using operator[] instead of explicit setHeapType
   // and getTempHeapType methods.
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -27,6 +27,7 @@
 #include "compiler-support.h"
 #include "support/hash.h"
 #include "support/insert_ordered.h"
+#include "support/topological_sort.h"
 #include "wasm-features.h"
 #include "wasm-type-printing.h"
 #include "wasm-type.h"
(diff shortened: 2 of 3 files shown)


tlively:
* main
  * **PR #5365** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5365" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5365?utm_source=stack-comment).

src/wasm-type.h
```diff
@@ -647,11 +647,12 @@ struct TypeBuilder {
...
 
   // Returns all of the newly constructed heap types. May only be called once
-  // all of the heap types have been initialized with `setHeapType`. In nominal
-  // mode, all of the constructed HeapTypes will be fresh and distinct. In
-  // nominal mode, will also produce a fatal error if the declared subtype
-  // relationships are not valid.
+  // all of the heap types have been initialized with `setHeapType`. In relaxed
+  // mode, the types are allowed to appear in any order and be arbitrarily
+  // mututally recursive because they will be implicitly reordered and put into
+  // a single rec group, so there should not be any existing rec groups.
```
kripken, line 653: Hmm, I am not 100% happy with "relaxed" here but I can't find a better name...

(I considered "nominal", "singleton", "simple" but all have downsides)

src/wasm/wasm-type.cpp
```diff
@@ -2979,6 +2983,87 @@ TypeBuilder::BuildResult TypeBuilder::build() {
...
+  // looping when tryiing to sort in the next step.
+  std::unordered_map<HeapType, size_t> originalTypeIndices;
+  std::vector<HeapType> originalTypes;
+  originalTypes.reserve(size());
+  for (size_t i = 0; i < size(); ++i) {
+    auto type = getTempHeapType(i);
+    originalTypeIndices[type] = i;
+  }
+
+  if (auto err = getSupertypeCycle(originalTypes)) {
```
kripken, line 2997: Nothing seems to be inserted into `originalTypes`? Just a reservation.

tlively:
One more problem to solve before deleting nominal mode is that the public and private types will all be lumped together and cause a closed world validation error. On the other hand, nominal does the same lumping in its output, so arguably it should never pass closed world validation.

One fix would be to find all the public types and put them in their own separate rec group, but that would require type-updating.cpp to be able to update public types while currently it is only able to update private types.

Another "fix" would be to start requiring producers to produce valid isorecursive types and not even try to emulate nominal mode after we delete it.

tlively:
> One fix would be to find all the public types and put them in their own separate rec group, but that would require type-updating.cpp to be able to update public types while currently it is only able to update private types.

Another similar fix would be to have the parser build the public types (limited to just the types of function exports) first followed by the rest of the types. This works as long as our closed world public type validation would pass, but wouldn't generalize beyond that.

tlively:
Another fix would be to build all the types together but then split them out into their minimal rec groups immediately afterward.

kripken:
> the public and private types will all be lumped together

I don't follow. TypeRewriter just rewrites private types, so the public ones stay where they were?

tlively:
The real goal of this PR is to work toward deleting the nominal type system by providing a variant of `TypeBuilder::build` that the parsers can call in emulated-nominal mode to provide something that looks like today's nominal mode behavior on top of isorecursive typing. As it is, deleting nominal typing and using `TypeBuilder::buildRelaxed` in the parsers would not be sufficient because it would put the public and private types together into a single rec group and therefore fail closed world validation.

kripken:
I see, thanks. Yes, in the parsers it doesn't sound like enough.

PR #5373, opened 2022-12-21
[Parser] Parse data segments

tlively:
Parse active and passive data segments, including all their variations and
abbreviations as well as data segments declared inline in memory declarations.
Switch to parsing data strings, memory limits, and memory types during the
ParseDecls phase so that the inline data segments can be completely parsed
during that phase and never revisited. Parsing the inline data segments in a
later phase would not work because they would be incorrectly inserted at the end
of the data segment index space.

Also update the printer to print a memory use on active data segments that are
initialized in a non-default memory.

PR #5372, opened 2022-12-21
[NFC][Parser] Track definition indices

tlively:
For each definition in a module, record that definition's index in the relevant
index space. Previously the index was inferred from its position in a list of
module definitions, but that scheme does not scale to data segments defined
inline inside memory definitions because these data segments occupy a slot in
the data segment index space but do not have their own independent definitions.

PR #5374, opened 2022-12-22
[Parser] Parse array creation and data segment instructions

tlively:


tlively:
* main
  * **PR #5372** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5372" target="_blank">Graphite</a> 
    * **PR #5373** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5373" target="_blank">Graphite</a> 
      * **PR #5374** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5374" target="_blank">Graphite</a>  👈
        * **PR #5375** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5375" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5374?utm_source=stack-comment).

src/wasm/wat-parser.cpp
```diff
@@ -1450,6 +1466,22 @@ struct ParseDefsCtx : TypeParserCtx<ParseDefsCtx> {
...
   }
 
+  Result<uint32_t> getDataFromIdx(uint32_t idx) {
+    if (idx >= wasm.dataSegments.size()) {
+      return in.err("data index out of bounds");
+    }
+    return idx;
+  }
+
+  Result<uint32_t> getDataFromName(Name name) {
```
kripken, line 1476: Could this perhaps be `getDataIndexFromName` and the above be `getValidDataIndex`?

kripken (approved):
lgtm % name suggestions

src/wasm/wat-parser.cpp
tlively: These should be consistent with the other `getXFromIdx` and `getXFromName` functions, so if we rename these we would want to rename them all. I would rather do that in a separate PR if we want to that at all.

I'm guessing the motivation here is that fact that data segments are referred to by index rather than by Name in Binaryen IR, but that's an unnecessary inconsistency that I would like to fix at some point.

kripken: Followup sounds good.

(To me, reading `getData` was not clear enough to understand what this was actually getting, which is the index.)

tlively:
⏳ This pull request is set to merge as part of a Graphite merge job
Stack job ID: YT2CYTWYcz4xraCFUPzB.
See details on [graphite.dev](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5374)

PR #5376, opened 2022-12-28
Move assert in MultiMemoryLowering

ashleynh:
Moving the assert that checks whether the DataSegment* offset type is Const. This assert only needs to happen when the DataSegment belongs to a memory other than the first.


tlively (commented):
Do we care about missing the trap for segments that have a non-constant offset that would have been out of bounds at runtime but is in bounds after multi-memory lowering?

kripken:
Good point, might be worth a TODO for now. But as with other traps I don't think it's urgent personally.

ashleynh:
Added the trap TODO, thanks

PR #5378, opened 2022-12-30
wasm2js: Avoid emitting non-JS code during opt

willcohen:
As noted in #4806, trying to optimize past level 0 can result in passes emitting non-JS code, which is then unable to be converted during final output.

This commit creates a new targetJs option in PassOptions, which can be checked inside each pass where non-JS code might be emitted.

This commit initially adds that logic to OptimizeInstructions, where this issue was first noticed.

### DIFF
--- a/src/pass.h
+++ b/src/pass.h
@@ -217,6 +217,12 @@ struct PassOptions {
   bool closedWorld = false;
   // Whether to try to preserve debug info through, which are special calls.
   bool debugInfo = false;
+  // Whether we are targeting JS. In that case we want to avoid emitting things
+  // in the optimizer that do not translate well to JS, or that could cause us
+  // to need extra lowering work or even a loop (where we optimize to something
+  // that needs lowering, then we lower it, then we can optimize it again to the
+  // original form).
+  bool targetJS = false;
   // Arbitrary string arguments from the commandline, which we forward to
   // passes.
   std::unordered_map<std::string, std::string> arguments;
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -3669,7 +3669,10 @@ private:
     {
       // ~(1 << x) aka (1 << x) ^ -1  ==>  rotl(-2, x)
       Expression* x;
-      if (matches(curr, binary(Xor, binary(Shl, ival(1), any(&x)), ival(-1)))) {
+      // Note that we avoid this in JS mode, as emitting a rotation would
+      // require lowering that rotation for JS in another cycle of work.
+      if (matches(curr, binary(Xor, binary(Shl, ival(1), any(&x)), ival(-1))) &&
+          !getPassOptions().targetJS) {
         curr->op = Abstract::getBinary(type, RotL);
         right->value = Literal::makeFromInt32(-2, type);
         curr->left = right;
--- a/src/wasm2js.h
+++ b/src/wasm2js.h
@@ -173,8 +173,11 @@ public:
     // and store would need to do a check. Given that, we can just ignore
     // implicit traps like those when optimizing. (When not optimizing, it's
     // nice to see codegen that matches wasm more precisely.)
+    // It is also important to prevent the optimizer from adding new things that
+    // require additional lowering, as we could hit a cycle.
     if (options.optimizeLevel > 0) {
       options.ignoreImplicitTraps = true;
+      options.targetJS = true;
     }
   }
 
--- a/test/wasm2js/target_js.wast
+++ b/test/wasm2js/target_js.wast
@@ -1,0 +1,13 @@
+(module
+  (func $func (export "func") (param $x i32) (result i32)
+    ;; Do not turn xor into something that does not express well in JS
+    ;; when optimizing.
+    (i32.xor
+      (i32.shl
+        (i32.const 1)
+        (local.get $x)
+      )
+      (i32.const -1)
+    )
+  )
+)
--- a/test/wasm2js/target_js.2asm.js
+++ b/test/wasm2js/target_js.2asm.js
@@ -1,0 +1,27 @@
+
+function asmFunc(imports) {
+ var Math_imul = Math.imul;
+ var Math_fround = Math.fround;
+ var Math_abs = Math.abs;
+ var Math_clz32 = Math.clz32;
+ var Math_min = Math.min;
+ var Math_max = Math.max;
+ var Math_floor = Math.floor;
+ var Math_ceil = Math.ceil;
+ var Math_trunc = Math.trunc;
+ var Math_sqrt = Math.sqrt;
+ var nan = NaN;
+ var infinity = Infinity;
+ function func(x) {
+  x = x | 0;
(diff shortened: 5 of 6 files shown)


test/wasm2js/target_js.wast
```diff
@@ -0,0 +1,29 @@
...
+ (func $0 (param $0 i32) (result i32)
```
kripken, line 7: Please make the testcase minimal - we don't need all the entries in the table here, I am pretty sure (do we need the elem at all?)

test/wasm2js/target_js.wast
```diff
@@ -0,0 +1,29 @@
...
+  (unreachable)
+ )
+ (func $1 (param $0 i32) (param $1 i32)
+  (i32.store
+   (i32.const 1)
+   (i32.xor
```
kripken, line 13: I think we only need the xor for this test? We could have one function that returns an i32, which is the xor.

test/wasm2js/target_js.wast
```diff
@@ -0,0 +1,29 @@
...
+    (i32.shl
```
kripken, line 14: Please add an explanation in the testcase, that even when optimizing we should not turn the xor into something that does not express well in JS.

src/passes/OptimizeInstructions.cpp
```diff
@@ -3669,7 +3669,8 @@ struct OptimizeInstructions
     {
       // ~(1 << x) aka (1 << x) ^ -1  ==>  rotl(-2, x)
       Expression* x;
-      if (matches(curr, binary(Xor, binary(Shl, ival(1), any(&x)), ival(-1)))) {
+      if (matches(curr, binary(Xor, binary(Shl, ival(1), any(&x)), ival(-1))) &&
```
kripken, line 3672:
```suggestion
      // Note that we avoid this in JS mode, as emitting a rotation would
      // require lowering that rotation for JS in another cycle of work.
      if (matches(curr, binary(Xor, binary(Shl, ival(1), any(&x)), ival(-1))) &&
```

src/pass.h
```diff
@@ -217,6 +217,8 @@ struct PassOptions {
   bool closedWorld = false;
   // Whether to try to preserve debug info through, which are special calls.
   bool debugInfo = false;
+  // Whether to target JS, to avoid optimization loops when using wasm2js
+  bool targetJs = false;
```
kripken, line 221:
```suggestion
  bool targetJS = false;
```

src/pass.h
```diff
@@ -217,6 +217,8 @@ struct PassOptions {
   bool closedWorld = false;
   // Whether to try to preserve debug info through, which are special calls.
   bool debugInfo = false;
+  // Whether to target JS, to avoid optimization loops when using wasm2js
```
kripken, line 220:
```suggestion
  // Whether we are targeting JS. In that case we want to avoid emitting things in
  // the optimizer that do not translate well to JS, or that could cause us to need
  // extra lowering work or even a loop (where we optimize to something that
  // needs lowering, then we lower it, then we can optimize it again to the original
  // form).
```

kripken (commented):
Thanks! Looks good aside from some minor comments.

test/wasm2js/target_js.wast
willcohen: @kripken Many thanks for reviewing this. Shrinking the test case any smaller than the following seems to result, as best as I can tell, with `wasm2js -O1` failing to keep enough code in to hit unreachable in the before case. To maintain the unreachable before/after problem, I was able to shrink the table, simplify the elem, and remove the superfluous first function. It's possible I'm missing additional minimization but am not sure how to rework further.

```
(module
 (type $i32_=>_none (func (param i32)))
 (import "env" "memory" (memory $mimport$0 256 32768))
 (table $0 3 3 funcref)
 (elem (i32.const 1) $0)
 (func $0 (param $0 i32) (param $1 i32)
  (i32.store
   (i32.const 1)
   (i32.xor
    (i32.shl
     (i32.const 1)
     (local.get $0)
    )
    (i32.const -1)
   )
  )
  (call_indirect (type $i32_=>_none)
   (i32.const 1)
   (i32.const 0)
  )
  (unreachable)
 )
)
```

test/wasm2js/target_js.wast
willcohen: As noted above, if I remove the table and simplify the function to xor, is there another way to retain the function at all? Anything further than the revised wast noted in my other reply seems to excise the function entirely at wasm2js -O1.

willcohen:
@kripken pending potentially shrinking the minimal test case further, all other changes have been adopted with the latest force push.

test/wasm2js/target_js.wast
kripken: Ah, yes, you need to export the function. Without that, only the table was keeping it alive...

Here is a minified version you can use:
```wat
(module
  (func $func (export "func") (param $x i32) (result i32)
    ;; Do not turn xor into something that does not express well in JS
    ;; when optimizing.
    (i32.xor
      (i32.shl
        (i32.const 1)
        (local.get $x)
      )
      (i32.const -1)
    )
  )
)
```

willcohen: Many thanks. Test case modified accordingly.

src/passes/OptimizeInstructions.cpp
```diff
@@ -3669,7 +3669,10 @@ struct OptimizeInstructions
     {
       // ~(1 << x) aka (1 << x) ^ -1  ==>  rotl(-2, x)
       Expression* x;
-      if (matches(curr, binary(Xor, binary(Shl, ival(1), any(&x)), ival(-1)))) {
+      // Note that we avoid this in JS mode, as emitting a rotation would
+      // require lowering that rotation for JS in another cycle of work.
+      if (matches(curr, binary(Xor, binary(Shl, ival(1), any(&x)), ival(-1))) &&
+          getPassOptions().targetJS == false) {
```
kripken, line 3675:
```suggestion
          !getPassOptions().targetJS) {
```

src/wasm2js.h
```diff
@@ -173,8 +173,12 @@ class Wasm2JSBuilder {
     // and store would need to do a check. Given that, we can just ignore
     // implicit traps like those when optimizing. (When not optimizing, it's
     // nice to see codegen that matches wasm more precisely.)
+    // Similarly, when targeting JS, it is important to clarify this for each pass,
+    // so no optimizations past level 0 end up adding non-JS things that cannot be
+    // targeted in the final output.
```
kripken, line 178:
```suggestion
    // It is also important to prevent the optimizer from adding new things that
    // require additional lowering, as we could hit a cycle.
```

kripken (approved):
Looks great, just two suggestions for comment modifications.

kripken:
Looks like there is a code style error on CI (over 80 columns).

willcohen:
> Looks like there is a code style error on CI (over 80 columns).

Re-filled paragraph based on line length and force pushed. Apologies.

willcohen:
Thank you!

kripken:
Thanks for the PR! Great to have this fixed.

PR #5380, opened 2023-01-03
[Wasm GC] Do not refine types of exported globals in closed world

kripken:
Doing so can cause us to switch from a private type to a public type.

Also refactor export-utils to make this easy.

Found by the fuzzer.


src/ir/export-utils.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "ir/export-utils.h"
+#include "wasm.h"
+
+namespace wasm::ExportUtils {
+
+namespace {
+
+// Returns a vector of all the things that are exported of a particular kind,
+// given that kind and a way to look them up on the module.
+template<typename T, ExternalKind K, typename G>
```
tlively, line 26: It would be possible to infer `G` from `K` or vice versa, but if this will always be wrapped in a non-template function, then it wouldn't be worth it.

tlively (approved):
Looks good! It took me a minute to page back in why this change shouldn't apply to open-world as well 😅

src/ir/export-utils.cpp
kripken: Yeah, I think maybe it's not worth it here. But adding a `getKindFromType` function does seem possible in principle (maybe we already have one and I'm not fully paged in yet as well? :smile: but I can't seem to see one)

PR #5381, opened 2023-01-03
[Wasm GC] Fix GlobalStructInference reasoning on unoptimizability

### LINKED ISSUE #5379: Optimization ignoring external values when using `--type-ssa`
If I optimize [h107.wasm.gz](https://github.com/WebAssembly/binaryen/files/10336760/h107.wasm.gz) with `wasm-opt -all -g --closed-world --type-ssa -O2`, the resulting module breaks.

It optimizes this function:
```wat
 (func $_BoxedDouble.toInt (type $type$55) (param $0 (ref $type$44)) (result i64)
  // ...
  (return
   (i64.trunc_sat_f64_s
    (struct.get $type$9 1
     (ref.cast $type$9
      (local.get $0)
     )
    )
   )
  )
 )
```
into
```wat
 (func $_BoxedDouble.toInt (type $type$3065) (param $0 (ref $type$2733)) (result i64)
  // ...
  (i64.trunc_sat_f64_s
   (select
    (f64.const 0)
    (f64.const 1)
    (ref.eq
     (local.get $0)
     (global.get $global$10)
    )
   )
  )
 )
```
Notice how it apparently has concluded that either the given object is a particular global whose field value is 1, or the field value is 0, so it replaced the read of the field by a conditional select.

However, the `$_BoxedDouble.toInt` function is called from this function (optimized version):
```wat
 (func $_stringRead (type $type$2695) (param $0 externref) (param $1 f64) (result f64)
  (local $2 (ref $type$2667))
  (f64.convert_i64_s
   (call_indirect $0 (type $type$2697)
    (local.tee $2
     (ref.cast $type$2667
      (ref.as_data
       (extern.internalize
        (local.get $0)
       )
      )
     )
    )
    (call $_BoxedDouble.toInt
     (struct.new $type$2696
      (i32.const 218)
      (local.get $1)
     )
    )
    (i32.sub
     (struct.get $type$2667 0
      (local.get $2)
     )
     (i32.const 229)
    )
   )
  )
 )
```
The `$_stringRead` function is exported from the module, so any value could be passed as `$1`. This value ends up in the object passed to `$_BoxedDouble.toInt`. Apparently Binaryen ignores this fact and assumes that only values from inside the module can end up in the field.

kripken:
We have a data structure there, typeGlobals, which maps types to the list of
globals for that type. Previously we used the convention of not having an
entry in the map to mean that a type is unoptimizable. However, this was not
used consistently, and in fact one place could insert to the map in a dangerous
way: a subtype's global is added to the list of globals of the super, and
`typeGlobals[super].add(sub-global)` would then effectively make an unoptimizable
super into an optimizable one.

To fix that, check for unoptimizability before propagating sub-globals. We do
still use the convention of not keeping data in typeGlobals for unoptimizable
things as it is a minor optimization to avoid wasted work.

Fixes #5379 (cc @askeksa-google )


test/lit/passes/gsi.wast
```diff
@@ -1268,3 +1268,162 @@
     )
   )
 )
+
+;; Two subtypes, each with a global. A get of the parent can be optimized into
+;; a select, as it must read one of the children.
```
tlively, line 1273: This should be gated behind --closed-world, but it doesn't look like it is. With an open world, some other value with the same type may have flowed in.

tlively (approved):
LGTM % comment, probably to be a addressed in a follow-up if it is correct.

test/lit/passes/gsi.wast
kripken: Still paging back here, but wouldn't that be an issue only for a public type in closed world mode?

tlively: No, because in closed world mode the only public types are function types (at least for now). If we had a public struct type in closed world mode, then I think we would still want the closed world assumption that only the current module can produce values of that type and that the environment could still only hold on to them and pass them back in but not inspect or produce them. The exception would be for the public types of imported globals, which must be produced by the environment. So really there are three levels of visibility in a generalized closed world mode:

1. imported public - There is an import with this type, so the environment must be able to create and inspect values with this type. This is only possible for function types under the current closed-world validation rules.
2. public - The type identity must be preserved so that the module signature does not change, but we can still assume that the environment does not produce or inspect values of this type. This visibility is not allowed under the current closed-world validation rules.
3. private - Arbitrary type modifications are allowed.

This pass might need to updated to handle "imported public" struct types correctly in the future. I'm not sure.

kripken: I see. Yeah, more to think about here, long term.

For now this pass only runs in closed world mode, I see, so I think we are ok.

https://github.com/WebAssembly/binaryen/blob/e761198190d5c5e6e0a3735c3dab59d02c970e43/src/passes/pass.cpp#L625-L628

tlively: Ah, then it looks like I missed adding an error to this pass to validate that we're running in closed world mode in #5367. Would you mind adding that error here and updating the test?

kripken: Is there a reason to do it in this PR? I can do it separately.

kripken: Oh, I guess it will conflict immediately in 2 places. I'll do it here...

tlively (approved):
Thanks!

test/lit/passes/gsi.wast
kripken: Unfortunately it's not simple, as several testcases need nontrivial modifications. I'll leave it out of this PR and look into it separately.

tlively: Sounds good 👍

PR #5386, opened 2023-01-04
Allow non-nullable ref.cast of nullable references

### LINKED ISSUE #5382: Fuzz bug on "ref.cast null on non-nullable input not yet supported"
```wat
(module
 (rec
  (type ${i32} (struct (field i32)))
  (type $none_=>_ref?|{i32}| (func (result (ref null ${i32}))))
 )
 (func $0 (type $none_=>_ref?|{i32}|) (result (ref null ${i32}))
  (local $0 (ref null ${i32}))
  (ref.cast null ${i32}
   (local.tee $0
    (ref.as_non_null
     (ref.null none)
    )
   )
  )
 )
)
```
STR:
```
$ bin/wasm-opt a.wat -all --coalesce-locals --roundtrip
[parse exception: ref.cast null on non-nullable input not yet supported (at 0:40)]
Fatal: error in parsing wasm binary
```

cc @tlively

tlively:
This new cast configuration was not expressible with the legacy cast
instructions. Although it is valid in Wasm, do not allow nullable casts of
non-nullable references, since those would unnecessarily lose type information.
Convert such casts to be non-nullable in the parsers and during expression
finalization.

Fixes #5382.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -207,7 +207,6 @@ def init_important_initial_contents():
         os.path.join('lit', 'passes', 'optimize-instructions-call_ref.wast'),
         os.path.join('lit', 'passes', 'inlining_splitting.wast'),
         os.path.join('heap-types.wast'),
-        os.path.join('spec', 'ref_cast.wast'),
     ]
     RECENT_DAYS = 30
 
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -207,6 +207,7 @@ def init_important_initial_contents():
         os.path.join('lit', 'passes', 'optimize-instructions-call_ref.wast'),
         os.path.join('lit', 'passes', 'inlining_splitting.wast'),
         os.path.join('heap-types.wast'),
+        os.path.join('spec', 'ref_cast.wast'),
     ]
     RECENT_DAYS = 30
 
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -6920,13 +6920,10 @@ bool WasmBinaryBuilder::maybeVisitRefCast(Expression*& out, uint32_t code) {
     } else {
       nullability = code == BinaryConsts::RefCast ? NonNullable : Nullable;
     }
-    // Only accept instructions emulating the legacy behavior for now.
-    if (ref->type.isRef()) {
-      if (nullability == NonNullable && ref->type.isNullable()) {
-        throwError("ref.cast on nullable input not yet supported");
-      } else if (nullability == Nullable && ref->type.isNonNullable()) {
-        throwError("ref.cast null on non-nullable input not yet supported");
-      }
+    // Implicitly convert nullable casts of non-null references to non-nullable
+    // casts to avoid losing type information.
+    if (ref->type.isRef() && ref->type.isNonNullable()) {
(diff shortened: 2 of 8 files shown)


tlively:
* main
  * **PR #5386** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5386" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5386?utm_source=stack-comment).

scripts/fuzz_opt.py
```diff
@@ -207,6 +207,7 @@ def init_important_initial_contents():
         os.path.join('lit', 'passes', 'optimize-instructions-call_ref.wast'),
         os.path.join('lit', 'passes', 'inlining_splitting.wast'),
         os.path.join('heap-types.wast'),
+        os.path.join('spec', 'ref_cast.wast'),
```
kripken, line 210: Intentional?

(you can just set RECENT_DAYS to 1 or 2, btw)

src/wasm/wasm-binary.cpp
```diff
@@ -6920,13 +6920,10 @@ bool WasmBinaryBuilder::maybeVisitRefCast(Expression*& out, uint32_t code) {
...
-    if (ref->type.isRef()) {
-      if (nullability == NonNullable && ref->type.isNullable()) {
-        throwError("ref.cast on nullable input not yet supported");
-      } else if (nullability == Nullable && ref->type.isNonNullable()) {
-        throwError("ref.cast null on non-nullable input not yet supported");
-      }
+    // Implicitly convert nullable casts of non-null references to non-nullable
+    // casts to avoid losing type information.
+    if (ref->type.isRef() && ref->type.isNonNullable()) {
+      nullability = NonNullable;
```
kripken, line 6926: Do we need this? It could be left for optimizations.

scripts/fuzz_opt.py
tlively: This was intentional, but I didn't know about RECENT_DAYS. I'll revert and use that instead. Meanwhile I found a place where OptimizeInstructions needs to be updated, so I'll push an update for that.

src/wasm/wasm-binary.cpp
tlively: Yes, because otherwise our validation check for this will fail. We could remove the validation rule as well, but I think it's helpful to simply disallow the suboptimal configuration.

kripken: I see. Makes sense, but it will also need to be in the C API and every location that creates such an instruction. I don't feel strongly though. If you do want to go this route, please document it in the readme among our list of IR differences from wasm.

tlively: Actually, since finalization already implements this automatic transition away from the undesirable configuration, we probably don't need to implement it anywhere else, even in the parsers. Does that sound right?

kripken: Good point, yeah, finalization is the proper single place to do this canonicalization. I'd fuzz it carefully though to make sure we're not missing something.

tlively: Fuzzed 58k iterations with auto-initial-contents over night, so I think this should be ready to go.

kripken (approved):

PR #5387, opened 2023-01-04
[NFC] Rename context methods for getting item indices

tlively:
Clarify in the names that these methods return logical indices in the various
module index spaces.

### DIFF
--- a/src/wasm/wat-parser.cpp
+++ b/src/wasm/wat-parser.cpp
@@ -528,7 +528,7 @@ struct NullTypeParserCtx {
   void appendLocal(LocalsT&, Name, TypeT) {}
 
   Result<Index> getTypeIndex(Name) { return 1; }
-  Result<HeapTypeT> getHeapTypeFromIdx(Index) { return Ok{}; }
+  Result<HeapTypeT> getHeapTypeIdxFromIdx(Index) { return Ok{}; }
 
   DataStringT makeDataString() { return Ok{}; }
   void appendDataString(DataStringT&, std::string_view) {}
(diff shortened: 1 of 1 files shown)


tlively:
* main
  * **PR #5387** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5387" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5387?utm_source=stack-comment).

tlively:
As discussed in https://github.com/WebAssembly/binaryen/pull/5374#discussion_r1060850947. I don't actually like this rename, though, and would prefer not to land this. Perhaps we could rename `getDataFromIdx` and `getDataFromName` to `getDataSegFromIdx` and `getDataSegFromName` to resolve the original concern about clarity instead?

tlively:
Maybe instead of `IdxFromIdx`, which uses "index" in two different ways, we could use `IdxFromInt`. WDYT?

src/wasm/wat-parser.cpp
```diff
@@ -1161,7 +1162,7 @@ struct ParseModuleTypesCtx : TypeParserCtx<ParseModuleTypesCtx>,
     : TypeParserCtx<ParseModuleTypesCtx>(typeIndices), in(in), wasm(wasm),
       types(types), implicitTypes(implicitTypes) {}
 
-  Result<HeapTypeT> getHeapTypeFromIdx(Index idx) {
+  Result<HeapTypeT> getHeapTypeIdxFromInt(Index idx) {
```
kripken, line 1165: This returns a `HeapType`, doesn't it, not an index? At least in some code paths we have `HeapTypeT = HeapType`.

tlively: Yes, but in Binaryen IR the "index" type used to refer to a heap type is `HeapType`. Basically the concrete types here are completely irrelevant; we're talking about indices in the sense of the Wasm abstract semantics here.

kripken: Hmm, I see. I guess in that case I'm really not sure if this is a significant improvement. The reader needs to keep a bunch of stuff in their head both before this PR and after, that is. So I'd be ok without this PR - what do you think?

tlively: Yeah, I don't feel strongly about it. Will close for now.

PR #5393, opened 2023-01-04
[Parser] Parse blocks

tlively:
Parse both the folded and unfolded forms of blocks and structure the code to
make supporting additional block instructions like if-else and try-catch
relatively simple.

Parsing block types is extra fun because they may implicitly define new
signature heap types via a typeuse, but only if their types are not given by a
single result type. To figuring out whether a new type may be introduced in all
the relevant parsing stages, always track at least the arity of parsed results.

The parser parses block labels, but more work will be required to support branch
instructions that use them.

### DIFF
--- a/src/wasm/wat-parser.cpp
+++ b/src/wasm/wat-parser.cpp
@@ -471,7 +471,8 @@ struct NullTypeParserCtx {
   using HeapTypeT = Ok;
   using TypeT = Ok;
   using ParamsT = Ok;
-  using ResultsT = Ok;
+  using ResultsT = size_t;
+  using BlockTypeT = Ok;
   using SignatureT = Ok;
   using StorageT = Ok;
   using FieldT = Ok;
@@ -504,8 +505,12 @@ struct NullTypeParserCtx {
   ParamsT makeParams() { return Ok{}; }
   void appendParam(ParamsT&, Name, TypeT) {}
 
-  ResultsT makeResults() { return Ok{}; }
-  void appendResult(ResultsT&, TypeT) {}
+  // We have to count results because whether or not a block introduces a
+  // typeuse that may implicitly define a type depends on how many results it
+  // has.
+  size_t makeResults() { return 0; }
+  void appendResult(size_t& results, TypeT) { ++results; }
+  size_t getResultsSize(size_t results) { return results; }
 
   SignatureT makeFuncType(ParamsT*, ResultsT*) { return Ok{}; }
 
@@ -534,6 +539,10 @@ struct NullTypeParserCtx {
   void appendDataString(DataStringT&, std::string_view) {}
 
   MemTypeT makeMemType(Type, LimitsT, bool) { return Ok{}; }
+
+  BlockTypeT getBlockTypeFromResult(size_t results) { return Ok{}; }
+
+  Result<> getBlockTypeFromTypeUse(Index, TypeUseT) { return Ok{}; }
 };
 
 template<typename Ctx> struct TypeParserCtx {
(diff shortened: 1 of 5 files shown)


tlively:
* main
  * **PR #5393** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5393" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5393?utm_source=stack-comment).

src/wasm/wat-parser.cpp
```diff
@@ -1310,7 +1360,7 @@ struct ParseDefsCtx : TypeParserCtx<ParseDefsCtx> {
       for (Index i = 0; i < expr->type.size(); ++i) {
         CHECK_ERR(push(pos,
                        builder.makeTupleExtract(
-                         builder.makeLocalGet(*scratchIdx, expr->type[i]), i)));
+                         builder.makeLocalGet(*scratchIdx, expr->type), i)));
```
kripken, line 1363: This was wrong before?

test/lit/wat-kitchen-sink.wast
```diff
@@ -701,6 +701,112 @@
...
+ ;; CHECK-NEXT:  (drop
+ ;; CHECK-NEXT:   (tuple.extract 1
+ ;; CHECK-NEXT:    (local.get $scratch)
+ ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT:  (unreachable)
+ ;; CHECK-NEXT: )
+ (func $block-folded
+  nop
+  (block $l (result i32) (result) (result i32)
```
kripken, line 745: this `(result)` is... valid? :scream:  I guess it has to be unless the spec special cases it...

test/lit/wat-kitchen-sink.wast
```diff
@@ -701,6 +701,112 @@
...
+ ;; CHECK-NEXT:    (nop)
+ ;; CHECK-NEXT:    (unreachable)
+ ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT:  (drop
+ ;; CHECK-NEXT:   (tuple.extract 0
+ ;; CHECK-NEXT:    (local.get $scratch)
+ ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT:  (drop
```
kripken, line 736: We don't support using the results yet, and so that part is not tested and we can only show drops here, is that right? (If it isn't, I missed it in the code...)

kripken (approved):

src/wasm/wat-parser.cpp
kripken: Ah, yes, looks like it certainly was.

test/lit/wat-kitchen-sink.wast
tlively: Yes, and even worse, you can put `(result)` literally anywhere in a wat file because it is an abbreviation that is entirely rewritten away. We only support it in reasonable places, though.

test/lit/wat-kitchen-sink.wast
tlively: No, we support using the results of blocks just fine. The drops here could be any instruction. Once the inner block is finished and its scope is popped, it is pushed into the outer scope just like any other instruction. See the calls to `ctx.appendInstr()` inside `instrs()`.

PR #5397, opened 2023-01-05
Support br_on_cast null

tlively:
As well as br_on_cast_fail null. Unlike the existing br_on_cast* instructions,
these new instructions treat the cast as succeeding when the input is a null.
Update the internal representation of the cast type in `BrOn` expressions to be
a `Type` rather than a `HeapType` so it will include nullability information.

Also update and improve `RemoveUnusedBrs` to handle the new instructions
correctly and optimize in more cases.

### DIFF
--- a/test/lit/passes/remove-unused-brs-gc.wast
+++ b/test/lit/passes/remove-unused-brs-gc.wast
@@ -166,19 +166,23 @@
  )
 
  ;; CHECK:      (func $br_on_cast_no (type $none_=>_ref|$struct|) (result (ref $struct))
- ;; CHECK-NEXT:  (block $block
+ ;; CHECK-NEXT:  (local $struct (ref null $struct))
+ ;; CHECK-NEXT:  (block $block (result (ref $struct))
  ;; CHECK-NEXT:   (drop
- ;; CHECK-NEXT:    (ref.null none)
+ ;; CHECK-NEXT:    (br_on_cast $block $struct
+ ;; CHECK-NEXT:     (local.get $struct)
+ ;; CHECK-NEXT:    )
  ;; CHECK-NEXT:   )
  ;; CHECK-NEXT:   (unreachable)
  ;; CHECK-NEXT:  )
  ;; CHECK-NEXT: )
  (func $br_on_cast_no (result (ref $struct))
+  (local $struct (ref null $struct))
   (block $block (result (ref $struct))
    (drop
     (br_on_cast $block $struct
      ;; As above, but now the type is nullable, so we cannot infer anything.
-     (ref.null $struct)
+     (local.get $struct)
     )
    )
    (unreachable)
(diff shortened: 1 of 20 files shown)


tlively:
* main
  * **PR #5397** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5397" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5397?utm_source=stack-comment).

test/spec/ref_cast.wast
```diff
@@ -131,6 +155,7 @@
 (assert_return (invoke "test-ref-test-any") (i32.const 1))
 (assert_return (invoke "test-ref-cast-struct"))
 (assert_return (invoke "test-br-on-cast-struct") (i32.const 1))
+(assert_return (invoke "test-br-on-cast-null-struct") (i32.const 1))
```
kripken, line 158: Two new exports, but only one new invoke of an export - is the latter missing?

test/passes/Oz_fuzz-exec_all-features.txt
```diff
@@ -159,14 +158,10 @@
      (i32.const 1)
     )
     (drop
-     (br_on_cast_fail $any $extendedstruct
+     (br $any
```
kripken, line 161: Does this fail because extendedstruct is not a subtype of struct? That might be a longstanding bug in the testcase if so. But might not be worth changing.

test/lit/passes/remove-unused-brs-gc.wast
```diff
@@ -141,19 +144,36 @@
...
+ ;; CHECK-NEXT:    (ref.null none)
  ;; CHECK-NEXT:   )
  ;; CHECK-NEXT:   (unreachable)
  ;; CHECK-NEXT:  )
  ;; CHECK-NEXT: )
  (func $br_on_cast_no (result (ref $struct))
-  (local $temp (ref null $struct))
   (block $block (result (ref $struct))
    (drop
     (br_on_cast $block $struct
```
kripken, line 179: comment on line 180 says "we cannot infer anything" but it seems like we optimize the br away entirely, so the comment is stale?

kripken (commented):
lgtm % comments

test/spec/ref_cast.wast
tlively: Oops, yes, good catch.

test/lit/passes/remove-unused-brs-gc.wast
tlively: Ah, that's what the new test above is supposed to cover, so rather than updating this comment, I'll update this test to get it back in a state where we can't optimize.

test/passes/Oz_fuzz-exec_all-features.txt
tlively: Yes, it would appear so. I guess we never updated it when we switched away from structural subtyping. I'll fix it.

kripken (approved):

tlively:
Fuzzed over 10k iterations, so going ahead and landing.

PR #5398, opened 2023-01-06
[Wasm GC] Fix optimizations on ref.cast of null, and optimize to ref.as_non_null

kripken:
We were checking the heap type, but now casts care about the nullability as
well.

If the nullability is the only problem, that is, the heap type will be fine but we
might have a null, we can at least switch a `ref.cast` (non-null) to a
`ref.as_non_null`.

I believe this might be related to recent cast/null changes. Found by the fuzzer.

### DIFF
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -1963,7 +1963,7 @@ struct OptimizeInstructions
     }
 
     // Check whether the cast will definitely succeed.
-    if (HeapType::isSubType(curr->ref->type.getHeapType(), intendedType)) {
+    if (Type::isSubType(curr->ref->type, curr->type)) {
       replaceCurrent(curr->ref);
 
       // We must refinalize here, as we may be returning a more specific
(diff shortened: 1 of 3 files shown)


src/passes/OptimizeInstructions.cpp
```diff
@@ -1963,7 +1963,7 @@ struct OptimizeInstructions
     }
 
     // Check whether the cast will definitely succeed.
-    if (HeapType::isSubType(curr->ref->type.getHeapType(), intendedType)) {
+    if (Type::isSubType(curr->ref->type, curr->type)) {
```
tlively, line 1966: Instead of using `curr->ref->type` here, we could use the more refined of `cur->ref->type` and `fallthrough->type` for some extra precision in case type information was lost somewhere.

We can also check to see whether the cast will definitely fail by subsequently checking `!Type::isSubType(curr->type, curr->ref->type)`. It looks like we try to do that above with a call to `canBeCastTo`, but that only looks at heap types, so it's not as precise as it could be.

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -3120,4 +3120,67 @@
...
+
+  ;; CHECK:      (func $ref-cast-heap-type (type $ref?|$B|_ref|$B|_=>_none) (param $null-b (ref null $B)) (param $b (ref $B))
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (local.get $b)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (ref.as_non_null
+  ;; CHECK-NEXT:    (local.get $null-b)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
```
tlively, line 3132: Do we refinalize when we do this optimization since the result becomes more refined?

tlively (commented):
Looks good, but it looks like there's room to further optimize and/or simplify the code.

test/lit/passes/optimize-instructions-gc.wast
kripken: Yes, there is a refinalize in this PR.

src/passes/OptimizeInstructions.cpp
kripken: Agreed, we can do more here. I suggest we leave those as followups?

tlively (approved):
Further work in follow ups sgtm.

PR #5399, opened 2023-01-06
Consolidate br_on* operations

tlively:
The `br_on{_non}_{data,i31,func}` operations are deprecated and directly
representable in terms of the new `br_on_cast` and `br_on_cast_fail`
instructions, so remove their dedicated IR opcodes in favor of representing them
as casts. `br_on_null` and `br_on_non_null` cannot be consolidated the same way
because their behavior is not directly representable in terms of `br_on_cast`
and `br_on_cast_fail`; when the cast to null bottom type succeeds, the null
check instructions implicitly drop the null value whereas the cast instructions
would propagate it.

Add special logic to the binary writer and printer to continue emitting the
deprecated instructions for now. This will allow us to update the test suite in
a separate future PR with no additional functional changes.

Some tests are updated because the validator no longer allows passing non-func
data to `br_on_func`. Doing so has not made sense since we separated the three
reference type hierarchies.


tlively:
* main
  * **PR #5399** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5399" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5399?utm_source=stack-comment).

kripken (commented):
Why are there test changes? Can this be an NFC change that only alters how we represent things internally?

tlively:
> Why are there test changes? Can this be an NFC change that only alters how we represent things internally?

primarily in cast-to-basic.wast and in gufa-refs.wast. Making this pure NFC isn't possible because we no longer have a way to differentiate between e.g. `br_on_cast  data` and `br_on_data`. Working around the change in validation behavior would make us less correct, so we don't want to do that either.

test/lit/passes/gufa-refs.wast
```diff
@@ -361,29 +361,30 @@
...
     ;; type and the i31 is not. The order in which we process this matters:
     ;;
     ;;  * If the i31 arrives first, we'll filter it out as incompatible with the
     ;;    type of the block. Then the null arrives and the final result is that
     ;;    null, which we can then optimize the block to return.
     ;;  * Or, if the null arrives first, then when the i31 arrives the
-    ;;    combination of nullfunc + i31 is Many (since the types are
-    ;;    incompatible). We then filter that to the block's type, ending up with
-    ;;    a cone of funcref. We cannot optimize in that case, unlike before.
+    ;;    combination of none + i31 is ??????
```
tlively, line 387: I think I updated the test so it's still testing what it is supposed to, but I'm not sure how to update this text here since the types are now compatible. Can you tell me if the test looks good and what this text should say?

test/passes/remove-unused-brs_all-features.wast
```diff
@@ -96,55 +96,4 @@
...
  )
-
- ;; a br_on of the obviously incorrect kind can just flow out the value as the
- ;; break is never taken
- (func $br_on-to-flow
-  ;; brs to func
-  (drop
-   (block $func (result (ref null func))
-    (drop
-     (br_on_func $func
```
kripken, line 107: Should this be replaced with a test using `br_on_cast`? I don't see one in this file.

test/lit/passes/gufa-refs.wast
kripken: If the types are now compatible then I think this is testing something else. Can the `br_on_func` that gets an `i31` be replaced with a `br_on_cast` that receives an incompatible type?

test/passes/remove-unused-brs_all-features.wast
tlively: No, `br_on_cast` is already tested in test/lit/passes/remove-unused-brs-gc.wast, so I intentionally did not replace the test in this file.

test/lit/passes/gufa-refs.wast
tlively: We could do e.g. `(br_on_cast $label i31 (struct.new ...))`, but there's no longer any way to combine types in completely different hierarchies, which is what we had before and is what created the `Many` mentioned in the comment.

kripken: I see. I guess this test has not been entirely testing the same thing then, maybe since the null changes actually.

I think we don't need it in that case. Let's delete it - it tested something that can't be expressed now.

kripken (approved):
lgtm with that

test/lit/passes/gufa-refs.wast
tlively: Sounds good, thanks!

PR #5400, opened 2023-01-06
[Wasm GC] Optimize impossible ref.casts better

kripken:
Also add some comments on related optimization opportunities.

### DIFF
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -1902,6 +1902,18 @@ struct OptimizeInstructions
     return HeapType::isSubType(a, b) || HeapType::isSubType(b, a);
   }
 
+  bool canBeCastTo(Type a, Type b) {
+    // A value can be cast to the other if the heap type can be cast, or if a
+    // null can work.
+    if (a.isNullable() && b.isNullable()) {
+      return true;
+    }
+    if (a.isRef() && b.isRef() && canBeCastTo(a.getHeapType(), b.getHeapType())) {
+      return true;
+    }
+    return false;
+  }
+
   void visitRefCast(RefCast* curr) {
     // Note we must check the ref's type here and not our own, since we only
     // refinalize at the end, which means our type may not have been updated yet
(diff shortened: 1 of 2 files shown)


src/passes/OptimizeInstructions.cpp
```diff
@@ -1902,6 +1902,19 @@ struct OptimizeInstructions
     return HeapType::isSubType(a, b) || HeapType::isSubType(b, a);
   }
 
+  bool canBeCastTo(Type a, Type b) {
```
tlively, line 1905: Instead of having this function that returns `true` when the cast may succeed or will certainly succeed, how about we have just a single method that returns `Success`, `Failure`, or `Unknown` just by checking `a <: b` and `b <: a`? That code would be simpler and would give us more information.

kripken: Makes sense. Would you mind if I leave that for a followup? I already have a PR for after this that would conflict.

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -2091,35 +2091,40 @@
...
   ;; NOMNL-NEXT: )
   (func $consecutive-opts-with-unreachable (param $func funcref)
     (drop
       (ref.cast $struct
         ;; Casting a funcref to i31 will definitely fail, so this will be
-        ;; replaced with an unreachable. But it should be enclosed in a block of
-        ;; the previous type, so that the outside ref.cast null is not confused. This
-        ;; is a regression test for a bug where we replace this node with an
+        ;; replaced with an unreachable. Likewise, the cast must fail. This is
+        ;; a regression test for a bug where we replace this node with an
```
tlively, line 2127: Can we update this test to not cast across type hierarchies? If we don't do it now, I'll have to do it soon when I update `ref.as*` to be implemented with `ref.cast`.

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -3183,4 +3188,94 @@
...
+    )
+    (drop
+      (ref.cast null $struct
+        (local.get $b)
+      )
+    )
+    ;; This last case is the only one that can succeed. We leave it alone, but
+    ;; in theory we could optimize it to { ref == null ? null : trap }.
+    (drop
+      (ref.cast null $struct
```
tlively, line 3276: We should at least optimize this to `ref.cast null none` to make the result type more specific.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1943,26 +1956,21 @@ struct OptimizeInstructions
...
-        replaceCurrent(builder.makeBlock(
-          {builder.makeDrop(curr->ref), builder.makeUnreachable()},
-          curr->type));
-        return;
-      }
-      // Otherwise, we are not sure what it is, and need to wait for runtime
-      // to see if it is a null or not. (We've already handled the case where
-      // we can see the value is definitely a null at compile time, earlier.)
+    // Check whether the cast will definitely fail.
+    if (!canBeCastTo(fallthrough->type, curr->type)) {
```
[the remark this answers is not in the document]
kripken: True. In my followup PR I handle that actually, so let's defer to then?

test/lit/passes/optimize-instructions-gc.wast
kripken: Sounds good. I'll delete the test as if we didn't cast across hierarchies it would overlap with existing tests for impossible casts.

test/lit/passes/optimize-instructions-gc.wast
kripken: Sounds good, I noted myself a TODO.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1987,10 +1995,25 @@ struct OptimizeInstructions
...
     // non-null.
+    //
+    // Note that we could do something similar for a failed cast, that is,
+    // handle the situation where the entire cast might succeed, but the heap
+    // type part will definitely fail. For example, the input might be a
+    // nullable array while the output might be a nullable struct. That is, a
+    // situation where the only way the cast succeeds is if the input is null.
+    // However, optimizing this would mean emitting something like
+    //
+    //   ref == null ? null : trap
```
tlively, line 2013: It would be nice to update this comment to mention the `ref.cast null none` idea, which would be strictly better than this code pattern.

tlively (approved):

src/passes/OptimizeInstructions.cpp
tlively: But no need if this comment is going to be short-lived anyway.

kripken: It's already deleted in the followup :smile:

kripken:
(I could work on a single big PR for all this, but with all the corner cases I'm trying to make the PRs as small as possible.)

tlively:
Yep, small PRs are nice. This is a great use case for stacked PRs, where I can get a better sense for what's coming next.

PR #5401, opened 2023-01-06
Replace `RefIs` with `RefIsNull`

tlively:
The other `ref.is*` instructions are deprecated and expressible in terms of
`ref.test`. Update binary and text parsing to parse those instructions as
`RefTest` expressions. Also update the printing and emitting of `RefTest`
expressions to emit the legacy instructions for now to minimize test changes and
make this a mostly non-functional change. Since `ref.is_null` is the only
`RefIs` instruction left, remove the `RefIsOp` field and rename the expression
class to `RefIsNull`.

The few test changes are due to the fact that `ref.is*` instructions are now
subject to `ref.test` validation, and in particular it is no longer valid to
perform a `ref.is_func` on a value outside of the `func` type hierarchy.

### DIFF
--- a/test/lit/passes/optimize-instructions-gc.wast
+++ b/test/lit/passes/optimize-instructions-gc.wast
@@ -238,17 +238,13 @@
   ;; CHECK-NEXT:   )
   ;; CHECK-NEXT:  )
   ;; CHECK-NEXT:  (drop
-  ;; CHECK-NEXT:   (i32.eqz
-  ;; CHECK-NEXT:    (ref.is_null
-  ;; CHECK-NEXT:     (local.get $func)
-  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:   (ref.is_func
+  ;; CHECK-NEXT:    (local.get $func)
   ;; CHECK-NEXT:   )
   ;; CHECK-NEXT:  )
   ;; CHECK-NEXT:  (drop
-  ;; CHECK-NEXT:   (i32.eqz
-  ;; CHECK-NEXT:    (ref.is_null
-  ;; CHECK-NEXT:     (local.get $i31)
-  ;; CHECK-NEXT:    )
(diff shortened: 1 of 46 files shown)


tlively:
* main
  * **PR #5401** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5401" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5401?utm_source=stack-comment).

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -238,17 +238,13 @@
   ;; CHECK-NEXT:   )
   ;; CHECK-NEXT:  )
   ;; CHECK-NEXT:  (drop
-  ;; CHECK-NEXT:   (i32.eqz
-  ;; CHECK-NEXT:    (ref.is_null
-  ;; CHECK-NEXT:     (local.get $func)
-  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:   (ref.is_func
+  ;; CHECK-NEXT:    (local.get $func)
```
kripken, line 242: Why is this now less optimized?

kripken: A PR before this sounds good, just so we don't have a regression in the middle (could affect bisection in the future, in principle). I could do it if you want as I assume I wrote the original.

tlively: Actually, this has significant overlap with your ongoing changes to `ref.cast` optimization, so it might be better to wait and reuse the code we end up with there.

kripken: Ok, fair enough, let's not block on that then, and leave it.

kripken (approved):

PR #5402, opened 2023-01-06
[Wasm GC] More minor cast optimizations

kripken:
Look for definitely-failing casts along all the fallthrough values. Specifically, if any
cast in the middle proves the final cast will fail, then we know we will trap.

Fully optimize redundant casts, considering both the type and the heap type.

Combine a cast with a `ref.as_non_null`.


src/passes/OptimizeInstructions.cpp
```diff
@@ -2023,97 +2061,61 @@ struct OptimizeInstructions
...
+      // having a heap supertype which is non-nullable, and the parent having
+      // a heap subtype which is nullable, like this:
+      //
+      //   (ref.cast null $B
+      //     (ref.cast $A
+      //
+      // We can optimize that to
+      //
+      //   (ref.cast $B
+      //     (ref.as_non_null $A
```
tlively, line 2097: Would it be worth saving the allocation of the `ref.as_non_null` node and finishing here instead of falling through to the next rule?

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -985,10 +981,8 @@
 
   ;; CHECK:      (func $flip-cast-of-as-non-null (type $anyref_=>_none) (param $x anyref)
   ;; CHECK-NEXT:  (drop
-  ;; CHECK-NEXT:   (ref.as_non_null
-  ;; CHECK-NEXT:    (ref.cast null $struct
-  ;; CHECK-NEXT:     (local.get $x)
-  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:   (ref.cast $struct
```
tlively, line 984: Looks like the comment on new line 1033 needs to be updated.

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -1628,59 +1607,43 @@
...
   ;; CHECK-NEXT: )
   ;; NOMNL:      (func $ref-eq-possible-b (type $eqref_eqref_=>_none) (param $x eqref) (param $y eqref)
   ;; NOMNL-NEXT:  (drop
   ;; NOMNL-NEXT:   (ref.eq
-  ;; NOMNL-NEXT:    (ref.as_non_null
-  ;; NOMNL-NEXT:     (ref.cast null $A
-  ;; NOMNL-NEXT:      (local.get $x)
-  ;; NOMNL-NEXT:     )
+  ;; NOMNL-NEXT:    (ref.cast $A
+  ;; NOMNL-NEXT:     (local.get $x)
```
tlively, line 1633: It looks like the comment on 1652 should be updated as well.

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -2728,34 +2707,65 @@
...
+  ;; NOMNL-NEXT:   )
+  ;; NOMNL-NEXT:  )
+  ;; NOMNL-NEXT:  (drop
+  ;; NOMNL-NEXT:   (block (result (ref $struct))
+  ;; NOMNL-NEXT:    (drop
+  ;; NOMNL-NEXT:     (ref.cast $array
+  ;; NOMNL-NEXT:      (local.get $x)
+  ;; NOMNL-NEXT:     )
   ;; NOMNL-NEXT:    )
+  ;; NOMNL-NEXT:    (unreachable)
```
tlively, line 2721: Would `vacuum` or some other pass get rid of the cast here since it's only effect is to trap, but it's followed by an unreachable anyway? We could do this even without --traps-never-happen.

src/passes/OptimizeInstructions.cpp
kripken: Yeah, I guess it's almost as simple to just optimize it fully there, I'll do that.

test/lit/passes/optimize-instructions-gc.wast
kripken: Hmm, atm no, we don't look forward in the block for this sort of thing.

In TNH mode the cast will just be removed anyhow due to the drop, so you're right this can be optimized, but only in non-TNH. For that reason it seems low priority to do, but in principle we could.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1956,14 +1955,53 @@ struct OptimizeInstructions
...
+          // the local state in this function is in sync which this change; it's
+          // easier to just do another clean pass on this node.)
+          replaceCurrent(curr);
+          return;
+        }
+
+        auto* last = ref;
+        ref = Properties::getImmediateFallthrough(
+          ref, getPassOptions(), *getModule());
+        if (ref == last) {
```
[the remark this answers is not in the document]
kripken: Thinking more on this, I think the current API makes sense. `getFallthrough` is usually called when we want to get the actual value we should care about, which is either the node itself, or a child (or child's child) if a child falls through. That is, we can do `curr = getFallthrough(curr)` and that is a no-op if there is no fallthrough. If instead it returned an optional, we'd need to write `if (auto fallthrough = getFallthrough(curr)) curr = *fallthrough` for that common pattern.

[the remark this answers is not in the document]
tlively: Makes sense to me!

PR #5403, opened 2023-01-06
Fix a bug optimizing out br_on_cast

tlively:
We were considering casts between unrelated types as unconditionally failing,
but in the case where the unrelated types are nullable, the cast could still
succeed if the value is null.

This bug was introduced in #5397.

### DIFF
--- a/src/ir/gc-type-utils.h
+++ b/src/ir/gc-type-utils.h
@@ -62,10 +62,10 @@ inline EvaluationResult evaluateKindCheck(Expression* curr) {
           return flip ? Failure : Success;
         }
         // If the cast type is unrelated to the type we have and it's not
-        // possible for the cast succeed anyway because the value is null, then
-        // the cast will certainly fail. TODO: This is essentially the same as
-        // `canBeCastTo` in OptimizeInstructions. Find a way to deduplicate this
-        // logic.
+        // possible for the cast to succeed anyway because the value is null,
+        // then the cast will certainly fail. TODO: This is essentially the same
+        // as `canBeCastTo` in OptimizeInstructions. Find a way to deduplicate
+        // this logic.
         if (!Type::isSubType(br->castType, br->ref->type) &&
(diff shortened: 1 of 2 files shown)


tlively:
* main
  * **PR #5403** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5403" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5403?utm_source=stack-comment).

src/ir/gc-type-utils.h
```diff
@@ -61,9 +61,13 @@ inline EvaluationResult evaluateKindCheck(Expression* curr) {
         if (Type::isSubType(br->ref->type, br->castType)) {
           return flip ? Failure : Success;
         }
-        // If the cast type is unrelated to the type we have, the cast will
-        // certainly fail.
-        if (!Type::isSubType(br->castType, br->ref->type)) {
+        // If the cast type is unrelated to the type we have and it's not
+        // possible for the cast succeed anyway because the value is null, then
```
kripken, line 65:
```suggestion
        // possible for the cast to succeed anyway because the value is null, then
```

test/lit/passes/remove-unused-brs-gc.wast
```diff
@@ -144,23 +144,114 @@
...
+ ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:   (unreachable)
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT: )
+ (func $br_on_cast_fail_unrelated (result anyref)
+  (local $nullable-struct2 (ref null $struct2))
+  (block $block (result anyref)
+   (drop
+    ;; This cast can be computed at compile time: it will definitely fail, so we
+    ;; can remove it.
```
kripken, line 231:
```suggestion
    ;; can remove it and leave an unconditional br.
```

kripken (approved):

PR #5404, opened 2023-01-06
[NFC] Refactor cast check logic

kripken:
This is just a start - checking that it looks like a good direction.


tlively (commented):
Direction looks great! Good idea adding `SuccessOnlyIfNull` and `SucessOnlyIfNonNull` as well.

kripken:
Should be complete now.

This adds a new utility for evaluating cast checks, and uses it in multiple places.

tlively (approved):

PR #5417, opened 2023-01-10
[Wasm GC] Add missing RefTest optimizations to parallel RefCast

kripken:
We already handled `Success` and `Failure`. Also handle `SuccessOnlyIfNull` and
`SuccessOnlyIfNonNull`.


tlively (approved):
Great!

PR #5418, opened 2023-01-11
[Wasm GC] Handle an unreachable br_on_cast_fail in DCE

kripken:
Without this we hit an assertion on unreachable not being a heap type.


tlively (approved):

kripken:
`emcc` CI error is unrelated and should be fixed by #5419

PR #5419, opened 2023-01-11
[binaryen.js] Avoid use of the global buffer var which emcc has removed

kripken:
Error on CI that this should fix:

https://github.com/WebAssembly/binaryen/actions/runs/3894806501/jobs/6649299612

I am a little unclear as to why the error began on CI here... the emcc tot build has not been updated in a while as the roller is stalled (at least 6 days it seems). But this error began just over the last 24 hours somehow..? Anyhow, this fix is necessary given the change to remove this var.

### DIFF
--- a/src/js/binaryen.js-post.js
+++ b/src/js/binaryen.js-post.js
@@ -2597,7 +2597,7 @@ function wrapModule(module, self = {}) {
         const ptr = _malloc(size);
         Module['_BinaryenCopyMemorySegmentData'](module, id, ptr);
         const res = new Uint8Array(size);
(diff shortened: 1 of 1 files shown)


sbc100 (approved):
Is there not some way to use `set` here without creating a new subarrary?

Either way this seems like improvement so lgtm.

kripken:
I'm not aware of a way. `.set()` takes a typed array, so we need to make one. But it's worth it for large copies, at least - one JS allocation is not that bad, compared to a manual loop to do the copy (vs `.set()` which does a very efficient copy).

PR #5421, opened 2023-01-11
[NFC] Remove GCTypeUtils::evaluateKindCheck

tlively:
Since we refactored all the old kind-checking instructions to be represented as
general cast instructions, `GCTypeUtils::evaluateKindCheck` had become a
vestigial wrapper around `GCTypeUtils::evaluateCastCheck` that was only used in
RemoveUnusedBrs. Remove `evaluateKindCheck` and use `evaluateCastCheck` in
RemoveUnusedBrs without changing any functionality. A future PR may use the
extra information from `evaluateCastCheck` to further optimize branching casts.

### DIFF
--- a/src/ir/gc-type-utils.h
+++ b/src/ir/gc-type-utils.h
@@ -39,6 +39,22 @@ enum EvaluationResult {
   SuccessOnlyIfNonNull,
 };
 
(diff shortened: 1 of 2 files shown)


kripken (approved):
Nice!

PR #5423, opened 2023-01-12
[Wasm GC] Optimize casts of null values better

tlively:
Instead of only looking at the final fallthrough value and seeing whether it is
a `RefNull` expression to determine if we are casting a null value, check the
type of each intermediate fallthrough value to see if it is a null reference.

Also improve `evaluateCastCheck` to return `Failure` instead of
`SuccessOnlyIfNonNull` if the cast value is a reference to bottom type, since
those can never be non-null.

### DIFF
--- a/src/ir/gc-type-utils.h
+++ b/src/ir/gc-type-utils.h
@@ -95,6 +95,10 @@ inline EvaluationResult evaluateCastCheck(Type refType, Type castType) {
   if (refIsHeapSubType) {
     assert(refType.isNullable());
     assert(castType.isNonNullable());
+    if (refHeapType.isBottom()) {
+      // Non-null references to bottom types do not exist!
+      return Failure;
+    }
     return SuccessOnlyIfNonNull;
   }
 
--- a/src/ir/gc-type-utils.h
+++ b/src/ir/gc-type-utils.h
@@ -85,18 +85,15 @@ inline EvaluationResult evaluateCastCheck(Type refType, Type castType) {
     return SuccessOnlyIfNull;
   }
 
-  // The cast will not definitely succeed nor will it definitely fail.
-  //
-  // Perhaps the heap type part of the cast can be reasoned about, at least.
-  // E.g. if the heap type part of the cast is definitely compatible, but the
-  // cast as a whole is not, that would leave only nullability as an issue,
-  // that is, this means that the input ref is nullable but we are casting to
-  // non-null.
+  // If the heap type part of the cast is compatible but the cast as a whole is
+  // not, we must have a nullable input ref that we are casting to a
+  // non-nullable type.
   if (refIsHeapSubType) {
     assert(refType.isNullable());
     assert(castType.isNonNullable());
     if (refHeapType.isBottom()) {
-      // Non-null references to bottom types do not exist!
+      // Non-null references to bottom types do not exist, so there's no value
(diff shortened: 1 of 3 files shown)


tlively:
* main
  * **PR #5423** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5423" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5423?utm_source=stack-comment).

src/ir/gc-type-utils.h
```diff
@@ -95,6 +95,10 @@ inline EvaluationResult evaluateCastCheck(Type refType, Type castType) {
   if (refIsHeapSubType) {
     assert(refType.isNullable());
     assert(castType.isNonNullable());
+    if (refHeapType.isBottom()) {
+      // Non-null references to bottom types do not exist!
+      return Failure;
```
kripken, line 100: The comment on line 88 needs to be updated (or maybe just removed), since we can no longer say at that point that it will not definitely fail.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1949,12 +1925,27 @@ struct OptimizeInstructions
...
 
-        if (result == GCTypeUtils::Failure) {
+        if (result == GCTypeUtils::Success) {
+          // The cast will succeed, but we can't just remove the cast and
+          // replace it with `ref` because the intermediate expressions might
+          // have had side effects. If we know the value is a null, though, we
+          // can at least put the side effects in a block that directly returns
+          // the null value.
+          // TODO: Track side effects of expressions we've seen to determine
+          // whether we can just replace the cast with `ref`.
```
kripken, line 1942: I think we don't need this TODO as vacuum will remove the dropped ref, or parts of it, when possible. So we don't need to try to do better here.

kripken (approved):

src/passes/OptimizeInstructions.cpp
tlively: This TODO is to allow us to optimize things other than nulls where we wouldn't be able to just drop the child expression because we wouldn't have a value to return afterward.

kripken: Ok, I think we can do that without tracking side effects - maybe that's what confused me. We can add a local.tee to stash the value, and return a local.get of it?

tlively: Oh hmm, yes, that sounds like it would work. Are we worried about cases where it makes the code worse because there actually are side effects that vacuum wouldn't be able to get rid of?

kripken: I'm not very worried about that. In trapsNeverHappen mode vacuum should almost always succeed. And the block vanishes anyhow. I suppose if we wanted to be careful we could gate this opt on trapsNeverHappen, but the benefits may lead to bigger savings later anyhow even without it.

GUFA does this sort of thing and it's not had downsides. Which makes me realize that this overlaps with GUFA as it tracks constant values like nulls through expressions and then emits them on the outside. But as this is quite simple it still seems worthwhile.

tlively: Sounds good, thanks. I'll update the TODO and implement this in a follow-up.

PR #5425, opened 2023-01-12
[Wasm GC] Allow TypeMerging to merge cast types with TNH

tlively:
If traps are assumed never to happen, then ref.cast and call_indirect
instructions cannot differentiate between types because they always succeed.
Take advantage of this fact by only having these instructions inhibit type
merges when we are not assuming traps never happen.


tlively:
* main
  * **PR #5424** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5424" target="_blank">Graphite</a> 
    * **PR #5425** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5425" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5425?utm_source=stack-comment).

kripken:
This PR is to main, so it also includes all the parts of the dependent PR that should land before it. Does Graphite not allow doing PRs relative to branches? I do that manually without Graphite and it seems to work reasonably ok.

kripken:
Oh, it did, and the branch target was just flipped before I read it I guess.

But the other changes + conflict are still here?

tlively:
Yeah, when the previous PR gets squashed and merged instead of merged as-is, GitHub doesn't do a good job of updating the diff even though it changes the target branch. You can always filter the diff to just include the new commits, though.

src/passes/TypeMerging.cpp
```diff
@@ -57,13 +57,21 @@ using ReferredTypes = SmallUnorderedSet<HeapType, 5>;
 struct CastFinder : public PostWalker<CastFinder> {
   ReferredTypes referredTypes;
 
+  bool trapsNeverHappen;
```
kripken, line 60: Maybe add a comment as to why it is valid to ignore certain things when TNH, like in the PR description?

test/lit/passes/type-merging-tnh.wast
```diff
@@ -0,0 +1,113 @@
...
+        )
+        (unreachable)
+      )
+    )
+    (unreachable)
+  )
+)
+
+;; call_indirect does not inhibit merging if traps never happen (although we
+;; haven't implemented function type merging anyway TODO).
```
kripken, line 92:
```suggestion
;; call_indirect should not inhibit merging if traps never happen, but we
;; haven't implemented function type merging yet TODO
```

kripken (approved):
Good idea!

src/passes/TypeMerging.cpp
```diff
@@ -57,13 +57,21 @@ using ReferredTypes = SmallUnorderedSet<HeapType, 5>;
...
+  CastFinder(const PassOptions& options)
```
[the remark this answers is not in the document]
kripken: (blank line between the bool and the constructor)

PR #5426, opened 2023-01-12
[Wasm GC] Optimize successful casts better

tlively:
Optimize ref.cast instructions that must succeed by simply replacing them with
their child in the case where the child has a more refined type or by
propagating a further removed fallthrough value with a more refined type using a
tee.

### DIFF
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -1936,7 +1936,10 @@ struct OptimizeInstructions
         auto result = GCTypeUtils::evaluateCastCheck(ref->type, curr->type);
 
         if (result == GCTypeUtils::Success) {
-          // The cast will succeed. If there were no intermediate expressions,
+          // The cast will succeed. This can only happen if the ref is a subtype of
+          // the cast instruction, which means we can replace the cast with the ref.
+          assert(Type::isSubType(ref->type, cast->type));
+          // If there were no intermediate expressions,
           // we can just skip the cast.
           if (ref == curr->ref) {
             replaceCurrent(ref);
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -1929,23 +1929,41 @@ struct OptimizeInstructions
     Builder builder(*getModule());
     auto nullType = curr->type.getHeapType().getBottom();
     {
-      auto* ref = curr->ref;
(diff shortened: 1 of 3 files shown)


tlively:
* main
  * **PR #5426** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5426" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5426?utm_source=stack-comment).

src/passes/OptimizeInstructions.cpp
```diff
@@ -1929,23 +1929,41 @@ struct OptimizeInstructions
...
             return;
           }
+          // We need to use a tee to return the value since we can't materialize
+          // it directly.
+          auto scratch = builder.addVar(getFunction(), ref->type);
+          *refp = builder.makeLocalTee(scratch, ref, ref->type);
+          replaceCurrent(
+            builder.makeSequence(builder.makeDrop(curr->ref),
+                                 builder.makeLocalGet(scratch, ref->type)));
+          refinalize = true;
```
kripken, line 1965: can check if the type changed here.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1929,23 +1929,41 @@ struct OptimizeInstructions
...
 
         if (result == GCTypeUtils::Success) {
-          // The cast will succeed, but we can't just remove the cast and
-          // replace it with `ref` because the intermediate expressions might
-          // have had side effects. We can replace the cast with a drop followed
-          // by a direct return of the value, though.
-          //
-          // TODO: Do this for non-null values as well by storing the value to
-          // return in a tee.
+          // The cast will succeed. If there were no intermediate expressions,
```
kripken, line 1939:
```suggestion
          // The cast will succeed. This can only happen if the ref is a subtype of
          // the cast instruction, which means we can replace the cast with the ref.
          assert(Type::isSubType(ref->type, cast->type));
          // If there were no intermediate expressions,
```
(without this, it's not obvious why these replacements are all valid)

kripken (approved):

src/passes/OptimizeInstructions.cpp
tlively: Sounds good. I factored this check out so we don't have to duplicate it.

PR #5427, opened 2023-01-13
[Wasm GC] SignaturePruning should not prune fields from a type with subtypes

kripken:
For subtyping to not break, we must leave the number of fields equal in both
super and subtype, for each such pair.

Found by the fuzzer.

### DIFF
--- a/src/passes/SignaturePruning.cpp
+++ b/src/passes/SignaturePruning.cpp
@@ -150,7 +150,7 @@ struct SignaturePruning : public Pass {
       }
     }
 
-    // Types with subtypes cannot be pruned (for a function type B to be a
+    // Types with subtypes cannot be pruned: for a function type B to be a
     // subtype of A, we need them to have the same number of parameters, as B's
     // parameters must be a subtype of A's, and tuple types are only subtypes if
     // their sizes match.
--- a/src/passes/SignaturePruning.cpp
+++ b/src/passes/SignaturePruning.cpp
@@ -150,10 +150,11 @@ struct SignaturePruning : public Pass {
       }
     }
 
-    // Types with subtypes cannot be pruned: for a function type B to be a
-    // subtype of A, we need them to have the same number of parameters, as B's
-    // parameters must be a subtype of A's, and tuple types are only subtypes if
-    // their sizes match.
+    // Types with subtypes cannot be pruned as we must preserve contravariance
+    // of parameters. Likewise, we must preserve covariance of results, so in
+    // practice a type with either a supertype or a subtype cannot be modified.
+    // TODO We could handle "cycles" where we remove fields from a group of
+    //      types with subtyping relations at once.
     SubTypes subTypes(*module);
 
     // Find parameters to prune.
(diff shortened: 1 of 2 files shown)


src/passes/SignaturePruning.cpp
```diff
@@ -149,6 +150,12 @@ struct SignaturePruning : public Pass {
       }
     }
 
+    // Types with subtypes cannot be pruned: for a function type B to be a
+    // subtype of A, we need them to have the same number of parameters, as B's
+    // parameters must be a subtype of A's, and tuple types are only subtypes if
+    // their sizes match.
```
tlively, line 156:
```suggestion
    // Types with subtypes cannot be pruned: for a function type B to be a
    // subtype of A, we need them to have the same number of parameters, as B's
    // parameters must be supertypes of A's, and tuple types are only subtypes if
    // their sizes match.
```
Due to contravariance of parameters ;) Although we could also simplify this whole comment to just say the arities have to match.

src/passes/SignaturePruning.cpp
```diff
@@ -160,6 +167,10 @@ struct SignaturePruning : public Pass {
         continue;
       }
 
+      if (!subTypes.getStrictSubTypes(type).empty()) {
+        continue;
+      }
```
tlively, line 172: Maybe update the comment below to cover both the supertype and subtype cases?

tlively (approved):

src/passes/SignaturePruning.cpp
kripken: Does contravariance always mean the number is the same?

src/passes/SignaturePruning.cpp
kripken: Merged the comments.

src/passes/SignaturePruning.cpp
tlively: No, not by itself.

src/passes/SignaturePruning.cpp
```diff
@@ -149,6 +150,13 @@ struct SignaturePruning : public Pass {
       }
     }
 
+    // Types with subtypes cannot be pruned as we must preserve contravariance
+    // of parameters. Likewise, we must preserve covariance of results, so in
+    // practice a type with either a supertype or a subtype cannot be modified.
+    // TODO We could handle "cycles" where we remove fields from a group of
+    //      types with subtyping relations at once.
```
tlively, line 157:
```suggestion
    // A type must have the same number of parameters and results as its
    // supertypes and subtypes, so we only attempt to modify types without
    // supertypes or subtypes.
    // TODO We could handle "cycles" where we remove fields from a group of
    //      types with subtyping relations at once.
```
It shouldn't really be necessary to talk about co- or contravariance at all here.

src/passes/SignaturePruning.cpp
kripken: Then should the comment not only mention covariance and contravariance but also the number?

PR #5430, opened 2023-01-13
Add Valmari-Lehtinen DFA minimization as a standalone utility

tlively:
We used to have this algorithm in wasm-type.cpp, where we used it to implement
equirecursive type canonicalization, but we removed it when we removed
equirecursive typing. Bring the algorithm back as a standalone utility for
future use in optimization passes. In particular, it will be useful in
TypeMerging for identifying the greatest fixed point of mergeable types rather
than the smallest fixed point.

### DIFF
--- a/src/support/CMakeLists.txt
+++ b/src/support/CMakeLists.txt
@@ -5,6 +5,7 @@ set(support_SOURCES
   colors.cpp
   command-line.cpp
   debug.cpp
+  dfa_minimization.cpp
   file.cpp
   istring.cpp
   path.cpp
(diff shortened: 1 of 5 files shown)


tlively:
* main
  * **PR #5430** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5430" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5430?utm_source=stack-comment).

kripken (commented):
Reviewed all but `dfa_minimization.cpp` and lgtm so far - is that file essentially the same as the previously-reviewed versions? :smile:

tlively:
Yes, it's extremely similar to the previous code. I did a couple things like factor out the setup into separate initialization functions, but the data structures and logic are essentially copy-pasted from what we had before.

kripken (approved):

PR #5431, opened 2023-01-13
Support using JSPI to load the secondary wasm split module.

brendandahl:
When using JSPI with wasm-split, any calls to secondary module functions
will now first check a global to see  if the module is loaded. If not
loaded it will call a JSPI'ed function that will handle loading module.

The setup is split into the JSPI pass and wasm-split tool since the JSPI
pass is first run by emscripten and we need to JSPI'ify the load secondary
module function. wasm-split then injects all the checks and calls to the
load function.


brendandahl:
I'm not thrilled with how code has to be in the JSPI pass and wasm-split. If anyone has suggestions on a different approach I'd love to hear it.

kripken:
Perhaps another option is to have another pass to create the module-loading export? So the new pass would be run first and just make that export. Then JSPI would work normally, and module splitting afterwards.

OTOH there isn't much complexity in doing it as in this PR either.

src/ir/module-splitting.cpp
```diff
@@ -318,7 +338,12 @@ std::pair<std::set<Name>, std::set<Name>>
 ModuleSplitter::classifyFunctions(const Module& primary, const Config& config) {
   std::set<Name> primaryFuncs, secondaryFuncs;
   for (auto& func : primary.functions) {
-    if (func->imported() || config.primaryFuncs.count(func->name)) {
+    // In JSPI mode exported functions cannot to be moved to the secondary
```
tlively, line 341:
```suggestion
    // In JSPI mode exported functions cannot be moved to the secondary
```

src/ir/module-splitting.cpp
```diff
@@ -318,7 +338,12 @@ std::pair<std::set<Name>, std::set<Name>>
...
+    // module since that would make them async when they may not have the JSPI
+    // wrapper. Exported JSPI functions can still benefit from splitting though
+    // since the only the JSPI wrapper stub will remain in the primary module.
```
tlively, line 344:
```suggestion
    // since only the JSPI wrapper stub will remain in the primary module.
```

src/ir/module-splitting.cpp
```diff
@@ -409,6 +434,21 @@ void ModuleSplitter::thunkExportedSecondaryFunctions() {
   }
 }
 
+Expression* ModuleSplitter::maybeLoadSecondary(Builder& builder,
+                                               Expression* callIndirect) {
+  if (config.jspi) {
+    // Check if the secondary module is loaded and if it isn't, call the
+    // function to load it.
```
tlively, line 441: Does this depend on the secondary module being loaded synchronously? Or this this import JSPIed such that execution suspends until it asynchronously resolves? This stuff is so mind bending...

src/ir/module-splitting.cpp
```diff
@@ -409,6 +434,21 @@ void ModuleSplitter::thunkExportedSecondaryFunctions() {
   }
 }
 
+Expression* ModuleSplitter::maybeLoadSecondary(Builder& builder,
+                                               Expression* callIndirect) {
+  if (config.jspi) {
```
tlively, line 439:
```suggestion
  if (!config.jspi) {
    return callIndirect;
```

It would be good to de-nest this function a bit.

src/passes/JSPI.cpp
```diff
@@ -49,6 +50,12 @@
 //      Wrap each export in the comma-separated list. Similar to jspi-imports,
 //      wildcards and separate files are supported.
 //
+//   --pass-arg=jspi-split-module
+//
+//      Enables integration with wasm-split. A JSPI'ed function named
+//      `__load_secondary_module` will be injected that is used by wasm-split to
+//      load a secondary module.
```
tlively, line 57: And we can't have the wasm-split tool handle adding this function because wasm-split runs after Emscripten has already run the JSPI transformation on the module?

src/tools/wasm-split/wasm-split.cpp
```diff
@@ -300,22 +308,13 @@ void splitModule(const WasmSplitOptions& options) {
...
 
-  // Run asyncify on the primary module
-  if (options.asyncify) {
-    PassOptions passOptions;
-    passOptions.optimizeLevel = 1;
-    passOptions.arguments.insert({"asyncify-ignore-imports", ""});
-    PassRunner runner(&wasm, passOptions);
-    runner.add("asyncify");
-    runner.run();
-  }
```
kripken, line 317: Looks like this removes the old asyncify support - is that not needed any more?

src/wasm.h
```diff
@@ -2198,6 +2198,7 @@ class Module {
   Global* getGlobal(Name name);
   Tag* getTag(Name name);
 
+  bool isExported(const Function& func) const;
```
kripken, line 2201: We generally try to keep `wasm.h` pretty minimal. I think a better place for this particular helper would be in `src/ir/export-utils.h`.

src/ir/module-splitting.cpp
brendandahl: `load_secondary_module` will be a JSPI function so it will suspend until the secondary module is loaded. Any exported wasm function that leads to this code path will also need a JSPI wrapper. If the user messes this up there will be a runtime error.

src/passes/JSPI.cpp
brendandahl: Correct.

src/tools/wasm-split/wasm-split.cpp
```diff
@@ -248,6 +248,14 @@ void splitModule(const WasmSplitOptions& options) {
     }
   }
 
+  if (options.jspi) {
+    // The load secondary module function must be kept in the main module.
+    if (keepFuncs.find(ModuleSplitting::LOAD_SECONDARY_MODULE) ==
+        keepFuncs.end()) {
+      keepFuncs.insert(ModuleSplitting::LOAD_SECONDARY_MODULE);
+    }
```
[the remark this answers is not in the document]
brendandahl: Oh yeah, this is a set.

test/lit/wasm-split/jspi-secondary-export.wast
```diff
@@ -0,0 +1,85 @@
...
+ ;; SECONDARY:      (elem (i32.const 0) $foo)
+
+ ;; SECONDARY:      (func $foo (param $0 i32) (result i32)
+ ;; SECONDARY-NEXT:  (i32.const 0)
+ ;; SECONDARY-NEXT: )
+ (func $foo (param i32) (result i32)
+  (i32.const 0)
+ )
+)
+;; PRIMARY:      (export "__load_secondary_module" (func $__load_secondary_module))
```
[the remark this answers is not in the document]
brendandahl: Yeah, good idea.

src/tools/wasm-split/wasm-split.cpp
brendandahl: I thought this was never integrated with emscripten. @tlively should I leave this?

kripken (approved):
lgtm % remaining open discussions

src/ir/module-splitting.cpp
tlively: I see, so it's up to the user to manually choose which exports to JSPI-ify? Can we add an option to trap or otherwise report an error if we ever get to one of these loading sites from a non-JSPI-ified entry point?

brendandahl: > I see, so it's up to the user to manually choose which exports to JSPI-ify?

Yeah, we could do the automatic route, but then you may have exports that switch between sync and async.

> Can we add an option to trap or otherwise report an error if we ever get to one of these loading sites from a non-JSPI-ified entry point?

We could, though v8 will already reports this with an "invalid suspender error". Do we want to add our own trap as well?

tlively: Ah, unless we can provide a better error message, I guess there's nothing more for us to do.

tlively (approved):

PR #5432, opened 2023-01-14
[Wasm GC] Use DFA minimization to merge types

tlively:
Analyze type merging candidates as states in a DFA, using a partition refining
DFA minimization algorithm to more aggressively find sets of mergeable types.
The new implementation can merge types that the previous iteration would have
taken an arbitrary number of iterations to merge or would not have been able to
merge at all.

The new implementation is also careful in its treatment of private and public
types and should work as-is under an open world assumption or a relaxed
closed-world assumption that admits more public types. Specifically, the new
code avoids merging public types into their supertypes, but still allows
private subtypes to be merged into their public supertypes.

The new implementation supports merging function types, which the old
implementation did not.

Finally, the new implementation is able to merge identical sibling types even
when the analysis determines that they cannot be merged into their common
supertype. Extending this capability to sibling top-level types without
nontrivial supertypes is left to a follow-on PR.


tlively:
* main
  * **PR #5430** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5430" target="_blank">Graphite</a> 
    * **PR #5432** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5432" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5432?utm_source=stack-comment).

tlively:
The code should be good to go and you can see some small benefits in the existing tests, but I still have to write a bunch of new tests for this next week.

test/lit/passes/type-merging.wast
```diff
@@ -138,25 +138,23 @@
...
   (func $foo
     ;; $A will remain the same.
     (local $a (ref null $A))
     ;; $B can be merged into $A.
     (local $b (ref null $B))
-    ;; $C refines the field, so it cannot be merged. However, separately, in
-    ;; the type definition of $C, its field of type $B should become $A. That
-    ;; is, $B should no longer be used anywhere.
+    ;; $C refines the field, so normally it wouldn't be able to be merged, but
+    ;; since that refinement does not survive merging, we can still merge $C.
```
kripken, line 157: 1. Please modify this test to test what it used to (that a field is properly updated in a type that is not merged), or add a new one.
2. I am not sure I understand why the new result is valid. Is it only because `$C` is not used anywhere? (if so then I guess 1 could be achieved by using it)

tlively: > I am not sure I understand why the new result is valid. Is it only because `$C` is not used anywhere? (if so then I guess 1 could be achieved by using it)

Note that this is the same result you would get if you had run the previous implementation of type merging for another iteration. Using `$C` in a cast would prevent the merge, yes. I'll update this test to use `$C` in a cast and test all the fun new merging capabilities separately.

tlively: Actually, updating `$C` to use a cast would be testing something materially different. It would be testing that casts inhibit merging rather than that refining inhibits merging. I'll update this test another way.

src/passes/TypeMerging.cpp
```diff
@@ -88,188 +91,331 @@ struct CastFinder : public PostWalker<CastFinder> {
...
   void visitCallIndirect(CallIndirect* curr) {
     if (!trapsNeverHappen) {
-      referredTypes.insert(curr->heapType);
+      castTypes.insert(curr->heapType);
     }
   }
 };
 
+// TODO: This is almost directly copied from type-updating.cpp. Deduplicate
+// this.
```
kripken, line 100: Good idea - I think utilities for supertype-first and subtype-first sorting would be useful in a shared place.

src/passes/TypeMerging.cpp
```diff
@@ -88,188 +91,331 @@ struct CastFinder : public PostWalker<CastFinder> {
...
-        }
-        return getTempType(type);
-      }
+bool TypeMerging::mayBeMergeable(Type a, Type b) {
+  if (a.isTuple() && b.isTuple()) {
+    return mayBeMergeable(a.getTuple(), b.getTuple());
+  }
+  return a == b || (a.isRef() && b.isRef() && !a.getHeapType().isBasic() &&
+                    !b.getHeapType().isBasic());
+}
```
kripken, line 406: In most cases similar to this, `foo(Type ..)` ends up calling `foo(HeapType ..)` on references, that is, it looks through the type (and possibly handles nullability, if necessary). Why does this not do that?

src/passes/TypeMerging.cpp
tlively: Do you mind if I do this as a follow-up?

src/passes/TypeMerging.cpp
tlively: Good point, I think that ignoring the nullability here is a bug. Ignoring the non-basic heap type is intentional, though. We don't know which non-basic heap types are going to end up getting merged, so when creating partitions we ignore differences between non-basic heap type children and the let the DFA state partitioning algorithm take care of differentiating between them.

src/passes/TypeMerging.cpp
kripken: Sure, followup sgtm.

src/passes/TypeMerging.cpp
kripken: I see, makes sense. Please add a comment though, at least for me that is not obvious.

tlively: There is a comment in the latest code 👍

test/lit/passes/type-merging.wast
```diff
@@ -189,24 +324,31 @@
...
-  ;; CHECK:      (type $A (struct_subtype (field (mut i32)) (field (mut i32)) (field (mut (ref null $D))) (field (mut i64)) (field (mut (ref null $I))) $H))
-  (type $A (struct_subtype (field (mut i32)) (field (mut i32)) (field (mut (ref null $E))) (field (mut i64)) (field (mut (ref null $I))) $H))
-  (type $A$to-merge (struct_subtype (field (mut i32)) (field (mut i32)) (field (mut (ref null $E))) (field (mut i64)) (field (mut (ref null $I))) $A))
+  (rec
+    ;; CHECK:      (rec
+    ;; CHECK-NEXT:  (type $C (struct (field (mut i32))))
+
+    ;; CHECK:       (type $D (struct_subtype (field (mut i32)) (field (mut i32)) $C))
+
+    ;; CHECK:       (type $E (struct_subtype (field (mut i32)) (field (mut i32)) $D))
```
kripken, line 333: I guess `$E` appears now when it didn't in the previous output because of the change to hybrid typing?

kripken (commented):
I don't see a test for field updating after this. That is, a field was once `$B` and now it should be `$A`. Maybe there is one outside of the diff, or I missed it?

src/passes/TypeMerging.cpp
```diff
@@ -88,188 +91,345 @@ struct CastFinder : public PostWalker<CastFinder> {
...
+bool TypeMerging::mayBeMergeable(Type a, Type b) {
+  if (a == b) {
+    return true;
+  }
+  if (a.isTuple() && b.isTuple()) {
+    return mayBeMergeable(a.getTuple(), b.getTuple());
+  }
+  // The only thing allowed to differ is the non-basic heap type child, since we
+  // don't know before running the DFA partition refinement whether different
+  // heap type children will end up being merged.
```
kripken, line 409:
```suggestion
  // heap type children will end up being merged (we let the DFA algorithm
  // handle that).
```

test/lit/passes/type-merging.wast
tlively: Possibly, or it could be the bug that we're not properly replacing merged supertypes.

test/lit/passes/type-merging.wast
```diff
@@ -277,15 +416,93 @@
...
+    (local $b (ref null $sub-func))
+    ;; $sub-func-refined will not be merged into $func because it refines a result.
+    (local $c (ref null $sub-func-refined))
+  )
+)
+
+;; Check that public types are not merged.
+;; TODO: $C still appears in the output because type-updating.cpp does not
+;; support merging private types into their public supertypes very well. Fix
+;; this.
```
kripken, line 449: Is this TODO done? I don't see `$C` in the output.

kripken (approved):

test/lit/passes/type-merging.wast
tlively: Ah yes, this was a note I added last night to remind me of what was happening, but it's resolved now.

kripken:
@tlively 

Fuzzer found an issue:
```wat
(module
 (type $type$1 (func (param (ref null $type$0)) (result (ref null $type$1))))
 (type $type$0 (func_subtype (param (ref null $type$1)) (result (ref null $type$0)) $type$1))
 (func $0 (type $type$0) (param $0 (ref null $type$1)) (result (ref null $type$0))
  (unreachable)
 )
)
```

```
$ bin/wasm-opt a.wat -all --nominal --type-merging --closed-world
wasm-opt:src/ir/type-updating.cpp:94: void wasm::GlobalTypeRewriter::update(): Assertion `it->second < i' failed.
Aborted
```

tlively:
Fixed in #5439.

PR #5436, opened 2023-01-17
Factor supertype-first type sort into separate header

tlively:
type-updating.cpp implemented a topological sorts on heap types that would visit
supertypes first. Since this same sort will be used by TypeMerging.cpp in #5432,
factor it out into a shared utility in a new wasm-type-ordering.h header.

At the same time, fix a bug in which the sort would visit types not in the input
collection. Concretely, this bug would cause public supertypes of private types
to be visited when only private types should have been visited. This
functionality change will be tested in #5432.

In principle the subtype-first sort used in subtypes.h could also be moved to
this header, but that is not yet duplicated in the code base and is more
efficient because it depends on implementation details of its current context,
so do not move it for now.

### DIFF
--- a/src/wasm-type-ordering.h
+++ b/src/wasm-type-ordering.h
@@ -44,7 +44,7 @@ struct SupertypesFirst : TopologicalSort<HeapType, SupertypesFirst<T>> {
     // Types that are not supertypes of others are the roots.
     for (auto [type, isSuper] : typeSet) {
       if (!isSuper) {
-        this->template push(type);
+        this->push(type);
       }
     }
   }
--- a/src/wasm-type-ordering.h
+++ b/src/wasm-type-ordering.h
@@ -19,30 +19,40 @@
 
 #include <unordered_set>
 
+#include "support/insert_ordered.h"
 #include "support/topological_sort.h"
 #include "wasm-type.h"
 
 namespace wasm::HeapTypeOrdering {
 
-struct SupertypesFirst : TopologicalSort<HeapType, SupertypesFirst> {
-  SupertypesFirst(const std::vector<HeapType>& types) {
-    std::unordered_set<HeapType> supertypes;
+// Given a collection of types, iterate through it such that each type in the
+// collection is visited only after its immediate children in the collection are
+// visited.
+template<typename T>
+struct SupertypesFirst : TopologicalSort<HeapType, SupertypesFirst<T>> {
+  // For each type in the input collection, whether it is a supertype. Used to
+  // track membership in the input collection.
+  InsertOrderedMap<HeapType, bool> typeSet;
+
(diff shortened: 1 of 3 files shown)


tlively:
* main
  * **PR #5430** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5430" target="_blank">Graphite</a> 
    * **PR #5432** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5432" target="_blank">Graphite</a> 
      * **PR #5436** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5436" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5436?utm_source=stack-comment).

kripken (approved):

tlively:
This is no longer NFC and is now a prerequisite for #5432 instead of a follow-on. I've updated the PR title and description to match the new contents. PTAL!

src/wasm-type-ordering.h
```diff
@@ -0,0 +1,62 @@
...
+
+#include "support/insert_ordered.h"
+#include "support/topological_sort.h"
+#include "wasm-type.h"
+
+namespace wasm::HeapTypeOrdering {
+
+// Given a collection of types, iterate through it such that each type in the
+// collection is visited only after its immediate children in the collection are
+// visited.
```
kripken, line 30: What does "children" mean here? Subtypes? But then it sounds like this visits subtypes first, not supertypes?

src/wasm-type-ordering.h
```diff
@@ -0,0 +1,62 @@
...
+struct SupertypesFirst : TopologicalSort<HeapType, SupertypesFirst<T>> {
+  // For each type in the input collection, whether it is a supertype. Used to
+  // track membership in the input collection.
+  InsertOrderedMap<HeapType, bool> typeSet;
+
+  SupertypesFirst(const T& types) {
+    for (auto type : types) {
+      typeSet.insert({type, false});
+      if (auto super = type.getSuperType()) {
+        typeSet[*super] = true;
```
kripken, line 41: Hmm, I don't understand this code, sorry. `super` might not be in `types`, so why can we add it to `typeSet`?

Second, can't the `false` on line 39 end up overwriting a `true` that was written earlier for that particular type?

src/wasm-type-ordering.h
tlively: Yep, sorry, this should say "supertype" rather than "children".

src/wasm-type-ordering.h
tlively: Good catches. Fixes pushed now.

kripken (approved):

PR #5437, opened 2023-01-18
Add a TypeNameGenerator that uses names from a Module

tlively:
If the module does not have a name for a particular type, the new utility falls
back to use a different user-configurable type name generator, just like the
existing IndexedTypeNameGenerator does.

Also change how heap types are printed by this printing machinery (which is
currently only used for debugging) so that their names are printed in addition
to their contents. This makes the printer much more useful for debugging.


tlively:
* main
  * **PR #5437** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5437" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5437?utm_source=stack-comment).

src/wasm-type-printing.h
```diff
@@ -93,6 +94,33 @@ struct IndexedTypeNameGenerator
...
+  ModuleTypeNameGenerator(
+    const Module& wasm,
+    std::enable_if_t<std::is_same_v<T, DefaultTypeNameGenerator>>* = nullptr)
+    : ModuleTypeNameGenerator(wasm, defaultGenerator) {}
+
+  TypeNames getNames(HeapType type) {
+    if (auto it = wasm.typeNames.find(type); it != wasm.typeNames.end()) {
+      return it->second;
+    }
+    return fallback.getNames(type);
```
kripken, line 120: Not new in this PR, but `getNames` should maybe return a reference and not a copy..?

kripken (approved):

src/wasm-type-printing.h
tlively: I don't think that's possible, since in general the generators may be creating new names on the fly.

PR #5438, opened 2023-01-18
[Wasm GC] Merge root types with identical shapes

tlively:
The TypeMerging optimization usually tries to merge types into their supertypes,
but it can also merge sibling types together without merging them into their
common supertype. Implement this kind of sibling merging for root types that
have no nontrivial supertypes. Root types were previously never merged.

The implementation requires hashing and comparing the top-level structure of the
root types so that root types with the same structure can be assigned to the
same partition for DFA minimization.

### DIFF
--- a/src/passes/TypeMerging.cpp
+++ b/src/passes/TypeMerging.cpp
@@ -126,14 +126,35 @@ struct TypeMerging : public Pass {
   std::vector<HeapType> getPublicChildren(HeapType type);
   DFA::State<HeapType> makeDFAState(HeapType type);
   void applyMerges(const TypeUpdates& merges);
+};
 
-  bool mayBeMergeable(HeapType sub, HeapType super);
-  bool mayBeMergeable(const Struct& a, const Struct& b);
-  bool mayBeMergeable(Array a, Array b);
-  bool mayBeMergeable(Signature a, Signature b);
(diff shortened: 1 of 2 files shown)


tlively:
* main
  * **PR #5438** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5438" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5438?utm_source=stack-comment).

kripken (commented):
Code looks right to me, but do we have test coverage for all the code paths in the shape eq/hash logic? (I guess we can't test the hashing that much, but the eq at least) That code is pretty verbose and feels like there is a risk of a typo or other mistake.

That code is vaguely structurally similar to other areas of code, but I can't think of an obvious way to share it.

tlively:
Yeah, this is similar to `RecGroupHasher` in wasm-type.cpp and is even more similar to code we used to have for equirecursive canonicalization, but I wrote this new code fresh. I don't anticipate having to do this kind of comparing/hashing elsewhere, thankfully, so I don't think it's worth trying to factor out.

I'll add more robust testing for this tomorrow.

kripken (approved):

PR #5439, opened 2023-01-19
[Wasm GC] Do not merge supertypes into subtypes

tlively:
In TypeMerging we previously merged all subsequent types in a refined partition
into whichever type happened to be first in the partition, but when that first
type happened to be a subtype of one of the other types in the partition, that
would cause type-updating.cpp to try to update that subtype's supertype to be
itself, causing an assertion failure.

Fix the problem by ensuring that the merge target is not a subtype of any other
types in the partition.


tlively:
* main
  * **PR #5438** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5438" target="_blank">Graphite</a> 
    * **PR #5439** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5439" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5439?utm_source=stack-comment).

kripken (approved):

kripken:
This PR fixes the previous error, but I see an error even with this merged in:
```wat
(module
 (rec
  (type $type$7 (struct ))
  (type $type$5 (struct ))
  (type $type$6 (struct_subtype  $type$5))
  (type $type$4 (func (param f64) (result f64)))
  (type $type$3 (func (param f32) (result f32)))
  (type $type$2 (func (param (ref none) (ref $type$6) (ref $type$7))))
  (type $type$1 (struct ))
  (type $type$0 (func (result structref)))
 )
 (func $0 (type $type$0) (result structref)
  (struct.new_default $type$1)
 )
 (func $1 (type $type$2) (param $0 (ref none)) (param $1 (ref $type$6)) (param $2 (ref $type$7))
  (nop)
 )
 (func $2 (type $type$3) (param $0 f32) (result f32)
  (if (result f32)
   (f32.eq
    (local.get $0)
    (local.get $0)
   )
   (local.get $0)
   (f32.const 0)
  )
 )
 (func $3 (type $type$4) (param $0 f64) (result f64)
  (if (result f64)
   (f64.eq
    (local.get $0)
    (local.get $0)
   )
   (local.get $0)
   (f64.const 0)
  )
 )
)
```

```
$ bin/wasm-opt -all --closed-world a.wat --type-merging
wasm-opt: binaryen/src/ir/type-updating.cpp:94: void wasm::GlobalTypeRewriter::update(): Assertion `it->second < i' failed.
Aborted
```

Same assertion as before it looks like, but maybe a different cause?

tlively:
That new fuzz bug is fixed by #5442.

PR #5441, opened 2023-01-19
Fix segment fault of API BinaryenModuleParse (#5440)

### LINKED ISSUE #5440: BinaryenModuleParse segment fault when input parameter is a read only char pointer.
## Environment
Binaryen Version:
```shell
$ git show
commit 594ff7b9609656edb83187cb4600b23b3f2fde37 (HEAD -> main, origin/main, origin/HEAD)
```

OS version:
```shell
$ uname -a
Linux ubuntu 5.15.0-25-generic #25-Ubuntu SMP Wed Mar 30 15:54:22 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```
Compiler Version
```shell
$ g++ --version
g++ (Ubuntu 11.2.0-19ubuntu1) 11.2.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

## Reproduce steps
1. Build and install binaryen 
```shell
cmake -DCMAKE_BUILD_TYPE=Debug . && make && sudo make install
```
2.  Create following test case into main.c
```C
#include <binaryen-c.h>

int main() {
  const char *wat = "(module"
                    "(type $i32_i32_=>_i32 (func (param i32 i32) (result i32)))"
                    "(func $adder (param $0 i32) (param $1 i32) (result i32)"
                    "(i32.add"
                    "(local.get $0)"
                    " (local.get $1)"
                    ")))";

  BinaryenModuleRef module = BinaryenModuleParse(wat);

  BinaryenModulePrint(module);

  BinaryenModuleDispose(module);

  return 0;
}
```
3. Build with gcc
`
$ gcc -o main.bin main.c -L /usr/local/lib -lbinaryen
`
4. Run the test case, it crash with segment fault
```shell
$ ./main.bin
Segmentation fault (core dumped)
```
## Root cause analyse
In binaryen-c.cpp line 4166,
`
SExpressionParser parser(const_cast<char*>(text));
`
const_cast is dangerous because const char* may be put into a read only region by compiler. Write to the address of this pointer can lead to permission fault.

If run the program in debugger, the fault reason is straight forward:
In wasm-s-parser.cpp inline 321, the
 `
input[0] = 0;
`
write the address of the const char*, which lead to segment fault.
image

Changqing-JING:
fix #5440

### DIFF
--- a/src/binaryen-c.cpp
+++ b/src/binaryen-c.cpp
@@ -5453,7 +5453,7 @@ void BinaryenModuleSetFeatures(BinaryenModuleRef module,
 BinaryenModuleRef BinaryenModuleParse(const char* text) {
   auto* wasm = new Module;
   try {
-    SExpressionParser parser(const_cast<char*>(text));
+    SExpressionParser parser(text);
     Element& root = *parser.root;
     SExpressionWasmBuilder builder(*wasm, *root[0], IRProfile::Normal);
   } catch (ParseException& p) {
--- a/src/wasm-s-parser.h
+++ b/src/wasm-s-parser.h
@@ -91,16 +91,16 @@ public:
 // Generic S-Expression parsing into lists
 //
 class SExpressionParser {
-  char* input;
+  const char* input;
   size_t line;
-  char* lineStart;
+  char const* lineStart;
   SourceLocation* loc = nullptr;
 
   MixedArena allocator;
 
 public:
   // Assumes control of and modifies the input.
-  SExpressionParser(char* input);
+  SExpressionParser(const char* input);
   Element* root;
 
 private:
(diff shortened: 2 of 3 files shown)


kripken (approved):
Thanks!

Measuring speed (this does an extra copy), this seems 2% slower than what we have now. I think that's ok as text parsing speed is not the most important area for us, and there is also a new parser that will replace this code eventually anyhow.

kripken:
I think the error on CI was also present on main. Merging latest main should fix it.

Changqing-JING:
> 
Thank you very much! I have rebased the PR to main head

PR #5442, opened 2023-01-19
[Wasm GC] Fix supertype ordering bug in GlobalTypeRewriter

tlively:
GlobalTypeRewriter made sure to order supertypes before their subtypes so that
type building would succeed, but this ordering previously did not account for
the fact that supertypes might be replaced by the type updating. When a
supertype is replaced, the replacement might not previously have had a
supertype/subtype relationship with the subtype, so it might have happened to be
ordered after the subtype.

Fix the problem by taking supertype replacements into account when ordering
types in the GlobalTypeRewriter. This requires generalizing the
`SupertypesFirst` utility in wasm-type-ordering.h to allow users to modify how
supertypes are found.

### DIFF
--- a/src/ir/type-updating.cpp
+++ b/src/ir/type-updating.cpp
@@ -36,7 +36,22 @@ void GlobalTypeRewriter::update() {
   // come before their subtypes.
   Index i = 0;
   auto privateTypes = ModuleUtils::getPrivateHeapTypes(wasm);
-  for (auto type : HeapTypeOrdering::SupertypesFirst(privateTypes)) {
+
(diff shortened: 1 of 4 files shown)


tlively:
* main
  * **PR #5442** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5442" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5442?utm_source=stack-comment).

kripken (approved):

PR #5450, opened 2023-01-24
Fix a typo in Flat IR validation logic

kripken:
`doVisitFunction` does not exist, and we don't have static checking of that
sadly. Could `overrides` work here..?


tlively (approved):
Oh wow, so we just weren't running this verification?

Unfortunately `override` only works with proper virtual methods, not with the CRTP nonsense we use.

kripken:
Yes, we were missing this part of the validation. It might matter less than the rest though.

PR #5451, opened 2023-01-24
Avoid spurious warnings in pass skipping

kripken:
Nested runners should be ignored, as they run some internal stuff in
certain passes, which would not contain the pass the user asked to
skip with `--skip-pass`.

Diff without whitespace is trivial.


tlively (approved):

tlively:
I could see this being somewhat confusing for passes like inlining that run an entire optimization pipeline in a nested pass runner, but I don't know that it would be worth the extra complexity to "fix" that.

kripken:
Good point, this would miss the nested full pipeline run of `wasm-opt --inlining-optimizing`. But yeah, not worth it to fix I think.

PR #5452, opened 2023-01-25
Merge more sibling types in TypeMerging

tlively:
TypeMerging was previously able to merge identical root types and identical
siblings that were children of a distinct type, but only when the siblings had
the same top-level structure as their parent. Improve the optimization by also
merging identical children when they have a different top-level structure from
their parent.

This solution is not fully general because it does not merge shape-refining
children of separate types that would become identical siblings only after their
parent types are merged, but that full generality would require either
modifying Valmari-Lehtinen DFA minimization or performing the DFA minimization
in a loop until reaching a fixpoint.

### DIFF
--- a/src/passes/TypeMerging.cpp
+++ b/src/passes/TypeMerging.cpp
@@ -187,8 +187,20 @@ void TypeMerging::run(Module* module_) {
   // Map each type to its partition in the list.
   std::unordered_map<HeapType, Partitions::iterator> typePartitions;
 
-  // Map the top-level structures of root types to their partitions in the list.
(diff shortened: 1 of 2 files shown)


tlively:
* main
  * **PR #5452** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5452" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5452?utm_source=stack-comment).

kripken (approved):

PR #5459, opened 2023-01-26
[Strings] Add experimental StringNew variants

kripken:
`string.from_code_point` makes a string from an int code point.

`string.new_utf8*_try` makes a utf8 string and returns null on a UTF8 encoding
error rather than trap.


test/lit/passes/O1_skip.wast
```diff
@@ -1,6 +1,3 @@
-;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
-;; NOTE: This test was ported using port_passes_tests_to_lit.py and could be cleaned up.
```
tlively, line 2: Is this a mistake?

kripken: No, it's intentional - this test now has a manual `CHECK-NOT` so it can't be auto-updated (the update removes that line!). I missed this when landing and it's annoying in local work. I can open a separate PR, but it is trivial?

src/wasm/wasm-binary.cpp
```diff
@@ -7184,7 +7184,12 @@ bool WasmBinaryBuilder::maybeVisitStringNew(Expression*& out, uint32_t code) {
...
   Expression* start = nullptr;
   Expression* end = nullptr;
-  if (code == BinaryConsts::StringNewWTF8) {
+  bool try_ = false;
+  if (code == BinaryConsts::StringNewWTF8 ||
+      code == BinaryConsts::StringNewUTF8Try) {
+    if (code == BinaryConsts::StringNewUTF8Try) {
+      try_ = true;
+    }
     if (getInt8() != 0) {
```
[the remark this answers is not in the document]
kripken: Hmm, I think you're right, I'll add a TODO here. When we implement multimemory support for these we should fix that.

test/lit/passes/O1_skip.wast
kripken: Moved out.

PR #5461, opened 2023-01-26
[Wasm GC] Add CastRefining pass

kripken:
If a compiler emits "abstract" types, that are never created, then at least in
trapsNeverHappen mode we can refine their types to a proper subtype if one
exists. That is, if we have `ListType` and it has a single subtype `I32ListType`,
and we never create `ListType`, then casts to it are the same as casts to the
specific child `I32ListType` (since otherwise they'd trap, but in TNH we can
assume that's not the case).

This can help remove types "in the middle" that we have difficulty removing
otherwise. MergeTypes won't help if an abstract type adds a field to its super,
for example. 

In J2Wasm this happens, I assume, because the Java class hierarchy is
mirrored into wasm types, and in some cases abstract types there end up
with a single child. This affects 3% of all casts there, and the refining leads
to some later optimizations like removing other casts etc. Not a huge effect
but it seems useful.

While computing this we also find out which types are never created and
never have any subtype that is created. Casts to them must trap, and we
optimize that as well (even in non-TNH).


tlively:
> This can help remove types "in the middle" that we have difficulty removing
otherwise. MergeTypes won't help if an abstract type adds a field to its super,
for example.

I think we should be able to improve MergeTypes (or add a really simple additional pass) to get rid of all intermediate supertypes that are not used in the code. Maybe we should just do that in this pass since it's already doing the required analysis? I can take a look at that after this lands if you'd like.

kripken:
That would be a great followup!

Doing it in this pass might make sense since we've found the unused ones, yeah, and hopefully made some more unused by the refining. OTOH it means adding TypeRewriter logic here, I guess, which MergeTypes already has. However you think is cleaner sgtm.

test/lit/passes/O1_skip.wast
```diff
@@ -7,7 +7,6 @@
 ;; run it directly: the skip instruction overrides everything else.
 
 ;; There should also be no warning in the output.
-;; CHECK-NOT: warning:
```
tlively, line 10: This looks like an auto-update mistake.

src/passes/CastRefining.cpp
```diff
@@ -0,0 +1,284 @@
...
+    void visitRefCast(RefCast* curr) {
+      auto castType = curr->getCastType();
+      if (auto heapType = getHeapTypeIfRelevant(castType)) {
+        if (parent.createdTypesOrSubTypes.count(*heapType) == 0) {
+          // Nothing is created of this type or any subtype, so the cast can
+          // only pass through a null, at most.
+          Builder builder(*getModule());
+          Expression* rep = nullptr;
+          if (castType.isNullable()) {
+            // TODO: Without TNH we can still optimize here, to do a null check
```
tlively, line 233: This could be as simple as this:
```
curr->castType() = Type(heapType->getBottom(), Nullable);
```
Maybe worth doing in the same PR?

test/lit/passes/cast-refining.wast
```diff
@@ -0,0 +1,1039 @@
...
+
+  ;; YESTNH:      (type $C2 (struct_subtype  $B))
+  ;; NO_TNH:      (type $C2 (struct_subtype  $B))
+  (type $C2 (struct_subtype $B))
+
+  ;; YESTNH:      (func $ref.cast (type $anyref_=>_none) (param $x anyref)
+  ;; YESTNH-NEXT:  (drop
+  ;; YESTNH-NEXT:   (block
+  ;; YESTNH-NEXT:    (drop
+  ;; YESTNH-NEXT:     (ref.cast $A
```
tlively, line 556: I wonder if it would be worth updating these casts to be casts to bottom type in addition to sequencing an `unreachable` after them. That would remove the type use and further empower e.g. MergeTypes without having to run an Vacuum in between.

test/lit/passes/cast-refining.wast
```diff
@@ -0,0 +1,1039 @@
...
+  ;; NO_TNH-NEXT:   (block $block0 (result (ref any))
+  ;; NO_TNH-NEXT:    (br_on_non_null $block0
+  ;; NO_TNH-NEXT:     (local.get $x)
+  ;; NO_TNH-NEXT:    )
+  ;; NO_TNH-NEXT:    (unreachable)
+  ;; NO_TNH-NEXT:   )
+  ;; NO_TNH-NEXT:  )
+  ;; NO_TNH-NEXT: )
+  (func $br_on (param $x anyref)
+    ;; We don't optimize here yet. TODO
```
tlively, line 839: The $A is optimized to $B, so it would be good to clarify what optimization is missing.

tlively (commented):
Very nice!

What does the pass do with function types? Does it optimize out all their casts because it doesn't think functions are ever created? Same question for arrays.

tlively:
It also might be good to restrict the analysis to private types and assume that public types are never abstract. This should be testable today by importing functions with public types and never taking a reference to them so that the rest of the analysis doesn't consider them to be instantiated without this new logic.

src/passes/CastRefining.cpp
```diff
@@ -0,0 +1,284 @@
...
+    template<typename T> void refineCast(T* curr) {
+      auto& castType = curr->getCastType();
+      if (auto heapType = getHeapTypeIfRelevant(castType)) {
+        auto iter = parent.refinableTypes.find(*heapType);
+        if (iter != parent.refinableTypes.end()) {
+          // We can refine this cast.
+          castType = Type(iter->second, castType.getNullability());
+        }
+      }
+    }
```
tlively, line 222: Thinking more about optimizing abstract supertypes out of the type hierarchy: Since this pass has the exact analysis we would need to do that, we could just let the global type rewriter machinery replace the cast types rather than doing it manually. 

Specifically, we would want the type rewriter to replace abstract types with their single topmost instantiated subtype or with bottom if they have no instantiated subtypes.

kripken:
> Doing it in this pass might make sense since we've found the unused ones, yeah, and hopefully made some more unused by the refining. OTOH it means adding TypeRewriter logic here, I guess, which MergeTypes already has. However you think is cleaner sgtm.

Thinking more on that, TypeMerging is not run by default because it can limit later optimizations. But your followup idea here would not have such downsides IIUC. That would be a reason to do it in this pass I think.

kripken:
> What does the pass do with function types? Does it optimize out all their casts because it doesn't think functions are ever created? Same question for arrays.

Good catch, this was just totally broken for arrays and funcs... disappointing the fuzzer didn't catch that. I guess we have too few tests with array or func casts, and the fuzzer doesn't synthesize such casts either. Fixed.

src/passes/CastRefining.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+    SubTypes subTypes(*module);
+
+    // Compute createdTypesOrSubTypes by starting with the created types and
+    // then propagating subtypes.
+    createdTypesOrSubTypes = createdTypes;
+    for (auto type : subTypes.getDepthSort()) {
+      // If any of our subtypes are created, so are we.
+      for (auto subType : subTypes.getStrictSubTypes(type)) {
```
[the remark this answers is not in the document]
kripken: All those name changes sgtm.

src/passes/CastRefining.cpp
```diff
@@ -0,0 +1,284 @@
...
+        if (createdTypesOrSubTypes.count(subType)) {
+          createdTypesOrSubTypes.insert(type);
+          break;
+        }
+      }
+    }
+
+    if (trapsNeverHappen) {
```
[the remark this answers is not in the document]
kripken: Good idea, done.

src/passes/CastRefining.cpp
kripken: Good point, we can just do that here. And we can do it in TNH mode too, and leave the other part for OptimizeInstructions (I'll verify we already do it, or add it if not).

test/lit/passes/cast-refining.wast
kripken: Hmm, I'm not sure it's worth it. CastTypes will definitely be followed up with Vacuum in the normal pipeline (it's in pre global opts, so it runs before all the func opts).

test/lit/passes/cast-refining.wast
kripken: Fixed. I changed the test to focus on the optimization we don't do, and clarify that.

src/passes/CastRefining.cpp
kripken: TypeRewriter could do it, I guess, yeah. It would be more complicated, though, I think? We need the same analysis as here, plus TypeRewriter boilerplate.

kripken:
> It also might be good to restrict the analysis to private types and assume that public types are never abstract.

Can you say why? I don't see a reason to prevent public types from being abstract. That can happen as the result of opts.

src/passes/CastRefining.cpp
tlively: We could actually make this much simpler. If this pass only did the analysis and the type rewriting, then all the optimizations this pass currently implements could be handled by a subsequent `OptimizeInstructions` pass.

If I did this as a follow up, I would end up deleting all the optimization code in this pass and replacing it with type rewriting code, keeping only the analysis. Maybe that's ok if we want to verify that none of the optimization tests would regress after that change.

tlively:
> > It also might be good to restrict the analysis to private types and assume that public types are never abstract.
> 
> Can you say why? I don't see a reason to prevent public types from being abstract. That can happen as the result of opts.

Consider this perfectly valid closed world module that would be misoptimized if we allowed its public type to be considered abstract:

```
(module
  (type $public (func))
  (type $private (func_subtype $public))

  (import  "env" "g" (global $public (ref $public))

  (func $test (export "test") (type $public)
    (if
      (ref.test $public (global.get $public))
      (nop) ;; ok
      (unreachable) ;; error!
    )
  )
)
```

If `public` were considered abstract, the cast would be optimized to `ref.test $private (global.get $public)`, which might fail while the original cast always succeeded.

Under closed world, we could fix this by differentiating between "imported public" types that we know are created by the environment because that is necessary for instantiation and "linkage public" types that we assume the environment never creates, but rather uses only for linking imports and exports.

But looking forward to supporting this optimization with an open world, we would need to assume that all public types can be instantiated, so in that case we would simply need to disallow public types from being considered abstract.

src/passes/CastRefining.cpp
kripken: I think I see what you're saying, but I'm not sure it will remain that way in the future. There are in principle situations where you can refine a cast without also refining all other casts throughout the program in an identical way, which is what rewriting at the type level would require. Does that make sense or am I missing your point?

kripken:
We might disagree on what closed world means, given that example...

I believe the point of closed world is this, which is true for all types:

https://github.com/WebAssembly/binaryen/blob/d19c31b371aab47dedd166c2d056f1c583c6a9a0/src/pass.h#L187-L191

That is tempered by this:

https://github.com/WebAssembly/binaryen/blob/d19c31b371aab47dedd166c2d056f1c583c6a9a0/src/pass.h#L199-L203

But that practical constraint does **not** mean that in a closed world public types can be inspected and manipulated by the outside. Closed world doesn't mean "closed except for public types"; it means "closed, but for technical reasons sometimes public types are needed - like function types - so leave those alone, **but they still aren't inspected/manipulated by the outside**.

If you disagree we should discuss, and depending on that we may need to update those docs.

src/passes/CastRefining.cpp
tlively: > There are in principle situations where you can refine a cast without also refining all other casts throughout the program in an identical way

I think doing that sort of optimization would require a dataflow analysis (possibly interprocedural) to determine what specific types can flow into a specific cast. If a type has multiple topmost instantiated children but only a single topmost instantiated child flows into a particular cast, that cast can be optimized even if other casts to the same type couldn't be optimized.

If we want to implement that, we could either factor out the analysis in this pass to share with a new pass that does the data flow analysis as well, or we could add the data flow analysis and specific cast optimization to this pass as well. Either way, the optimizations currently done by this pass would be more powerful (because of the type optimizations) and also simpler if expressed as a type replacement followed by OptimizeInstructions.

tlively:
IMO it would be overly restrictive in the long term if we disallowed closed world modules from being able to import GC data from their environment, even if we continue assuming that the environment will not otherwise create, inspect, or manipulate that data. In the long term, I expect that we will want to be able to annotate individual types as `public` or `private` and have all the optimizations do the right thing.

If you don't think the example above should be a valid closed world module, then we would have to update our implementation of the closed world validation to reject it. But I think it would be much more forward looking to accept that as a valid closed world module and make the optimization work correctly for it. Let's discuss more tomorrow.

kripken:
After discussion I think we agreed for now to focus on the simple closed-world case, in which we assume no public types, and only keep what is forced to be public unchanged. That means the module above should not validate. However, in the longer term we'll want to optimize open world as much as possible, including with "public/private" annotations, and we may want to write future-facing code when relevant so it fits in that model (but we don't have a clear model for that yet).

src/passes/CastRefining.cpp
kripken: Talking offline, I can see how the refining part of this pass makes sense to do as a type rewriting. It would also remove unneeded types we've proven are safe to remove.

kripken: Ok, rewritten as a rewrite. This does look nicer I think!

To do this I did an NFC refactor to move the TypeMapper helper class from TypeMerging to a header.

src/passes/CastRefining.cpp
```diff
@@ -0,0 +1,292 @@
...
+    // To optimize we rewrite types. That is, if we want to optimize all casts
+    // of $A to instead cast to the refined type $B, we can do that by simply
+    // replacing all appearances of $A with $B. That is possible here since we
+    // only optimize when we know $A is never created, and we are removing all
+    // casts to it, which means no other references to it are needed - so we can
+    // just rewrite all references to $A to point to $B. Doing such a rewrite
+    // will also remove the unneeded type from the type section, which is nice
+    // for code size.
+    //
+    // Note that this could in theory limit optimizability, but it does not. It
```
tlively, line 200: Maybe rather than saying it limits optimizability "in theory" but not actually, which is somewhat confusing, it would be better to say that it "may allow a later TypeMerging to merge more types, which could inhibit more optimizations after that pass."

kripken:
With the removal of unneeded types, this PR now makes the J2Wasm binary almost 0.5% smaller. Not bad when this doesn't even have the followups to OptimizeInstruction to optimize using the changes here.

src/passes/CastRefining.cpp
```diff
@@ -0,0 +1,292 @@
...
+        // Go up the chain of supertypes, skipping things we are mapping away,
+        // as those things will not appear in the output. This skips B in the
+        // example above.
+        while (1) {
+          if (mapping.count(*super)) {
+            super = super->getSuperType();
+          } else {
+            return super;
+          }
+        }
```
tlively, line 277: This should check that the new `super` exists before accessing it (and it would be good to add a test that exercises this). It can also be a two-liner:
```suggestion
        for (; super && mapping.count(*super); super = super->getSuperType()) {}
        return super;
```

src/passes/CastRefining.cpp
```diff
@@ -0,0 +1,292 @@
...
+      }
+    };
+
+    CastRefiningTypeMapper(*module, mapping).map();
+
+    // Refinalize, as RefCasts may have new types now.
```
tlively, line 283: Lots of things may have new types now!

test/lit/passes/cast-refining.wast
```diff
@@ -0,0 +1,1124 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
+
+;; RUN: foreach %s %t wasm-opt --cast-refining --traps-never-happen -all --closed-world --nominal -S -o - | filecheck %s --check-prefix=YESTNH
+;; RUN: foreach %s %t wasm-opt --cast-refining                      -all --closed-world --nominal -S -o - | filecheck %s --check-prefix=NO_TNH
```
tlively, line 4: How about running these tests with `--hybrid` instead of `--nominal` in the hopes that we can eventually remove nominal typing? You'll just want to add a `--remove-unused-types` at the end so the unused types actually get removed from the test output.

test/lit/passes/cast-refining.wast
```diff
@@ -0,0 +1,1124 @@
...
+
+;; Run in both TNH and non-TNH mode.
+
+;; $A :> $B :> $C :> $D :> $E
+;;
+;; $A and $D have no struct.news, so we can optimize casts of them to their
```
tlively, line 10: Not just casts!

test/lit/passes/cast-refining.wast
```diff
@@ -0,0 +1,1124 @@
...
+  ;; NO_TNH-NEXT:   )
+  ;; NO_TNH-NEXT:  )
+  ;; NO_TNH-NEXT:  (drop
+  ;; NO_TNH-NEXT:   (ref.cast $E
+  ;; NO_TNH-NEXT:    (local.get $x)
+  ;; NO_TNH-NEXT:   )
+  ;; NO_TNH-NEXT:  )
+  ;; NO_TNH-NEXT: )
+  (func $ref.cast (param $x anyref)
+    ;; List out all possible casts for comprehensiveness. For other instructions
```
tlively, line 120: You could also use e.g. locals to show that the types have been updated. That would make for shorter tests.

src/passes/CastRefining.cpp
```diff
@@ -0,0 +1,292 @@
...
+    for (auto type : subTypes.types) {
+      if (!type.isStruct()) {
+        // TODO: support arrays and funcs
+        continue;
+      }
+
+      // Add a mapping of types that are never created (and none of their
+      // subtypes) to the bottom type. This is valid because all locations of
+      // that type, like a local variable, will only contain null at runtime.
+      // Likewise, if we have a ref.test of such a type, we can only be looking
```
tlively, line 229: It would be good to test that nothing goes wrong if we end up with a local (or anything else) typed as a non-nullable reference to bottom.

test/lit/passes/cast-refining.wast
```diff
@@ -0,0 +1,1124 @@
...
+    ;; This returns null.
+    (drop
+      (ref.cast null $C2
+        (local.get $x)
+      )
+    )
+  )
+)
+
+;; Function subtyping, which is a TODO - for now we do nothing.
```
tlively, line 923: It would be good to test with public function types as well. Even with the most restrictive closed world validation, we'll need to make sure nothing goes wrong when there is a public abstract supertype of a private subtype.

tlively:
> With the removal of unneeded types, this PR now makes the J2Wasm binary almost 0.5% smaller. Not bad when this doesn't even have the followups to OptimizeInstruction to optimize using the changes here.

Did you run OptimizeInstructions as-is after this pass? IIRC, most of the relevant optimizations are already in place. Am I misremembering?

kripken:
> IIRC, most of the relevant optimizations are already in place. 

`ref.cast` etc. of a bottom type are not optimized now. Some of the other stuff is present though. Good point though, I'm not sure where most of the benefit is, so it might already be present.

src/passes/CastRefining.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+  void visitStructNew(StructNew* curr) {
+    auto type = curr->type;
+    if (type != Type::unreachable) {
+      types.insert(type.getHeapType());
+    }
+  }
+};
+
+struct CastRefining : public Pass {
```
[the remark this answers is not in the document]
kripken: I agree `CastRefining` isn't the best name. But I think the future improvements would be to this pass, to add data flow analysis to it - it would reuse all the info we already have in this pass. So if we want a better name it should cover the existing pass + that extension I think.

src/passes/CastRefining.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+    // Use what we found about abstract types and never-created types to
+    // optimize.
+    optimize(module, subTypes);
+  }
+
+  void computeAbstractTypes(const SubTypes& subTypes) {
+    // Abstract types are those with no news, i.e., the complement of
+    // |createdTypes|. As mentioned above, we can only optimize this case if
+    // traps never happen.
```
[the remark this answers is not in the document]
kripken: I think we can, I'll add a TODO. In general though as all our main wasm GC users can use trapsNeverHappen atm I think that's low priority.

src/passes/CastRefining.cpp
kripken: The possible issue is even without TypeMerging, though. TypeMerging is risky (can limit opts), and this paragraph is trying to say that **this** optimization may appear to have the same risk - but in fact it does not.

src/passes/CastRefining.cpp
kripken: Good catch. This was actually already tested, but I guess nothing bad happened to happen from dereferencing the optional (whatever we found there wasn't in the mapping, so it worked out... Fixed.

test/lit/passes/cast-refining.wast
kripken: Do we want to remove nominal? I think it's useful in the long term too, and it makes tests easier to read and reason about, at least for me.

kripken:
Feedback is all either addressed or responded to.

src/passes/CastRefining.cpp
tlively: Maybe something like "Even though this pass removes types, it does not on its own inhibit further optimizations" would be clearer, then.

test/lit/passes/cast-refining.wast
tlively: Ideally I would like to make `--nominal` into a parser flag at most and transition away from using it ourselves so our test files can be fully valid Wasm. The only thing stopping me from doing that now is that our closed-world validation means that we would have to be more sophisticated than just sticking all the types in the same rec group.

src/passes/CastRefining.cpp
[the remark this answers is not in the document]
tlively: I think if you squint at it `AbstractTypeRefining` could cover that use case. It refines all the globally abstract types everywhere, and in the future it would also refine specific uses of types that are only "locally abstract."

src/passes/CastRefining.cpp
```diff
@@ -61,7 +61,9 @@ struct NewFinder : public PostWalker<NewFinder> {
 };
 
 struct CastRefining : public Pass {
-  // Only changes cast types, but not locals.
+  // Changes types, either by refining them without changing nullability, or
+  // changing them to be null types, so we never add new non-nullable locals
+  // here.
```
tlively, line 66:
```suggestion
  // Changes types, either by refining them without changing nullability, or
  // changing them to be bottom types, so we never add new non-nullable locals
  // here.
```

Although "changing them to be bottom types" is already included in "refining them." The pass never changes nullability, even when refining to a bottom type.

test/lit/passes/cast-refining.wast
```diff
@@ -757,6 +790,29 @@
...
+  ;; YESTNH-NEXT:  (local $A (ref none))
+  ;; YESTNH-NEXT:  (local $B (ref none))
+  ;; YESTNH-NEXT:  (local $C1 (ref none))
+  ;; YESTNH-NEXT:  (local $C2 nullref)
+  ;; YESTNH-NEXT:  (nop)
+  ;; YESTNH-NEXT: )
+  ;; NO_TNH:      (func $locals (type $none_=>_none)
+  ;; NO_TNH-NEXT:  (local $A (ref none))
+  ;; NO_TNH-NEXT:  (local $B (ref none))
+  ;; NO_TNH-NEXT:  (local $C1 (ref none))
```
tlively, line 804: Is there a pass responsible for optimizing out anything that claims to have an impossible type?

test/lit/passes/cast-refining.wast
```diff
@@ -1011,6 +1067,106 @@
...
+  )
+
+  ;; YESTNH:      (func $C (type $A)
+  ;; YESTNH-NEXT:  (nop)
+  ;; YESTNH-NEXT: )
+  ;; NO_TNH:      (func $C (type $A)
+  ;; NO_TNH-NEXT:  (nop)
+  ;; NO_TNH-NEXT: )
+  (func $C (export "C") (type $A)
+  )
```
tlively, line 1113: Is this intended to have `(type $A)`? It looks like it's meant to have `(type $C)`, but that would break closed-world validation by making `$B` public, so maybe this should be removed.

test/lit/passes/cast-refining.wast
```diff
@@ -1011,6 +1067,106 @@
...
+  ;; NO_TNH-NEXT:  )
+  ;; NO_TNH-NEXT:  (drop
+  ;; NO_TNH-NEXT:   (ref.cast $C
+  ;; NO_TNH-NEXT:    (local.get $x)
+  ;; NO_TNH-NEXT:   )
+  ;; NO_TNH-NEXT:  )
+  ;; NO_TNH-NEXT: )
+  (func $casts (param $x funcref)
+    ;; $A and $C have functions of their types, so in theory we could optimize
+    ;; $B here.
```
tlively, line 1151: I don't think it's the existence of functions, but rather of `ref.func`s, that make a function type non-abstract. Throw a `ref.func $B` in here somewhere that expects to receive a `(ref $A)` and watch the sparks fly in the future when this pass tries to rewrite abstract public type `$A` to `$B`.

tlively (approved):
LGTM % ongoing pass name discussion and fixing (and possibly improving) the test with public types.

src/passes/CastRefining.cpp
kripken: Sounds good, rewritten with that base.

src/passes/CastRefining.cpp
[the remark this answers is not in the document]
kripken: Makes sense, I guess either can work. Renamed as you suggested.

test/lit/passes/cast-refining.wast
kripken: I see. Sounds worth discussing.

Needing to add `--remove-unused-types` here though seems like a downside to me. These tests are a good example of where nominal typing makes things simpler.

test/lit/passes/cast-refining.wast
kripken: In practice GUFA will do so, since no content will arrive there. We could add a small peephole optimization as well for compiler efficiency I suppose.

test/lit/passes/cast-refining.wast
kripken: Right, good point, this should be type `$C` but not exported.

test/lit/passes/cast-refining.wast
kripken: Good point to also test references. Both are an issue, though, aren't they? With or without references we can't consider a type abstract if it is the type of an actual thing that exists.

kripken:
All feedback should be addressed.

test/lit/passes/cast-refining.wast
tlively: Oh yeah, I suppose that's right. We can't necessarily refine a function type if there is a function that implements it.

tlively (approved):
LGTM!

PR #5467, opened 2023-01-31
OptimizeInstructions: Handle signed overflow properly in canOverflow()

### LINKED ISSUE #5464: OptimizeInstructions fuzz error on i64 added constants
```wat
(module  
 (func "export"
  (if
   (i64.gt_s
    (i64.const 127)            
    (i64.add
     (i64.const 18014398509481984)
     (i64.or
      (i64.const 0)
      (i64.const 9223372036854775807)
     )
    )
   )
   (unreachable)
  )
 )
)
```

```
$ bin/wasm-opt a.wat -all --fuzz-exec --optimize-instructions
[fuzz-exec] calling export
[trap unreachable]
[fuzz-exec] calling export
[fuzz-exec] comparing export
[fuzz-exec] optimization passes changed results
```

kripken:
We only checked for an unsigned overflow, which was wrong.

Fixes #5464

### DIFF
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -3761,7 +3761,11 @@ private:
 
   // Returns true if the given binary operation can overflow. If we can't be
   // sure either way, we return true, assuming the worst.
-  bool canOverflow(Binary* binary) {
+  //
+  // We can check for an unsigned overflow (more than the max number of bits) or
+  // a signed one (where even reaching the sign bit is an overflow, as that
+  // would turn us from positive to negative).
+  bool canOverflow(Binary* binary, bool signed_) {
     using namespace Abstract;
 
     // If we know nothing about a limit on the amount of bits on either side,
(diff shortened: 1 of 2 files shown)


src/passes/OptimizeInstructions.cpp
```diff
@@ -3761,7 +3761,11 @@ struct OptimizeInstructions
 
   // Returns true if the given binary operation can overflow. If we can't be
   // sure either way, we return true, assuming the worst.
-  bool canOverflow(Binary* binary) {
+  //
+  // We can check for an unsigned overflow (more than the max number of bits) or
+  // a signed one (where even reaching the sign bit is an overflow, as that
+  // would turn us from positive to negative).
+  bool canOverflow(Binary* binary, bool signed_) {
```
tlively, line 3768: Can we do `isSignedOp` inside this function instead of having it take `signed_` as a parameter?

tlively (approved):
Huh, how'd we miss this?

PR #5470, opened 2023-01-31
RemoveUnusedModuleElements: Support function subtyping

kripken:
When we see a `call_ref` or `call_indirect` of a heap type, we might be calling a
subtype of that type.

### DIFF
--- a/src/passes/RemoveUnusedModuleElements.cpp
+++ b/src/passes/RemoveUnusedModuleElements.cpp
@@ -292,7 +292,7 @@ struct Analyzer {
     return worked;
   }
 
-  std::unique_ptr<SubTypes> subTypes;
+  std::optional<SubTypes> subTypes;
 
   void useCallRefType(HeapType type) {
     if (!subTypes) {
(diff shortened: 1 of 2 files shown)


src/passes/RemoveUnusedModuleElements.cpp
```diff
@@ -291,24 +292,32 @@ struct Analyzer {
     return worked;
   }
 
+  std::unique_ptr<SubTypes> subTypes;
```
tlively, line 295: Is there any reason to prefer the indirection?
```suggestion
  std::optional<SubTypes> subTypes;
```

tlively (approved):

src/passes/RemoveUnusedModuleElements.cpp
kripken: I used `move` for the actual construction. Is that right or is there some better C++ way?

tlively: I don't think `move` should be necessary here since what you're assigning is already an rvalue reference (I think). Are you getting an error without it?

kripken: It seems to build both ways.

PR #5473, opened 2023-02-01
Fix issues with ref.cast_nop

kripken:
It did not have proper annotation for the safety field, and also
it could not handle basic heap types.

If we don't want to fix the latter in this way, we would need to
have optimization passes check `safety`, since the reason I noticed
this is that CastRefining will turn casts to cast to null, which then
errors for the `_nop` variant.

Is this even still needed, I wonder?

### DIFF
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -6935,8 +6935,7 @@ void WasmBinaryBuilder::visitRefAsCast(RefCast* curr, uint32_t code) {
 bool WasmBinaryBuilder::maybeVisitRefCast(Expression*& out, uint32_t code) {
   if (code == BinaryConsts::RefCastStatic || code == BinaryConsts::RefCast ||
       code == BinaryConsts::RefCastNull || code == BinaryConsts::RefCastNop) {
-    bool legacy =
-      code == BinaryConsts::RefCastStatic;
+    bool legacy = code == BinaryConsts::RefCastStatic;
     auto heapType = legacy ? getIndexedHeapType() : getHeapType();
     auto* ref = popNonVoidExpression();
     Nullability nullability;
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -6936,7 +6936,7 @@ bool WasmBinaryBuilder::maybeVisitRefCast(Expression*& out, uint32_t code) {
   if (code == BinaryConsts::RefCastStatic || code == BinaryConsts::RefCast ||
       code == BinaryConsts::RefCastNull || code == BinaryConsts::RefCastNop) {
     bool legacy =
-      code == BinaryConsts::RefCastStatic || code == BinaryConsts::RefCastNop;
+      code == BinaryConsts::RefCastStatic;
     auto heapType = legacy ? getIndexedHeapType() : getHeapType();
     auto* ref = popNonVoidExpression();
     Nullability nullability;
--- a/src/wasm-delegations-fields.def
+++ b/src/wasm-delegations-fields.def
@@ -620,6 +620,7 @@ switch (DELEGATE_ID) {
   }
   case Expression::Id::RefCastId: {
     DELEGATE_START(RefCast);
+    DELEGATE_FIELD_INT(RefCast, safety);
     DELEGATE_FIELD_CHILD(RefCast, ref);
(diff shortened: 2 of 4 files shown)


src/wasm/wasm-binary.cpp
```diff
@@ -6935,8 +6935,7 @@ void WasmBinaryBuilder::visitRefAsCast(RefCast* curr, uint32_t code) {
 bool WasmBinaryBuilder::maybeVisitRefCast(Expression*& out, uint32_t code) {
   if (code == BinaryConsts::RefCastStatic || code == BinaryConsts::RefCast ||
       code == BinaryConsts::RefCastNull || code == BinaryConsts::RefCastNop) {
-    bool legacy =
-      code == BinaryConsts::RefCastStatic || code == BinaryConsts::RefCastNop;
+    bool legacy = code == BinaryConsts::RefCastStatic;
```
tlively, line 6938: Did the V8 implementation get updated to take an arbitrary heap type instead of a type index, then?

tlively (approved):
LGTM. I hadn't realized that V8 updated their implementation to match the new cast structure. No one has needed this recently, but I wouldn't be surprised if it is useful again in the future for estimating the benefits we could get from e.g. eliminating receiver casts somehow.

src/wasm/wasm-binary.cpp
kripken: I'm not sure. @jakobkummerow , does V8 use the same heap type encoding in `ref.cast` and `ref.cast_nop`?

jakobkummerow: We haven't changed `ref.cast_nop`, so it still uses the same immediate as the old `ref.cast_static` (`0xfb45`, i.e. an "index immediate", not a "heap type immediate").

If we want to keep `ref.cast_nop` for future experimentation, then it probably makes sense to update it to be consistent with the new `ref.cast [null]` (`0xfb41`/`0xfb49`) -- the question is which of these? Or would we need two variants of `ref.cast_nop`? Or should we use the opportunity to drop it?

tlively: Dropping it seems fine since it would take some amount of work and coordination to update it. We can do that work once someone asks for it.

kripken: Probably makes sense to do the least amount of work for now since we might want this for future experimentation but are not sure. For now in Binaryen I think that's what's in this PR, which makes it behave like `ref.cast` without null (`0xfb41`).

tlively: I would rather fix the bug where the `safety` field was skipped without making our version of `ref.cast_nop` incompatible with V8's version. The binary format should not change and `ref.cast_nop` should not work with basic heap types.

kripken: If `ref.cast_nop` doesn't work with basic heap types then a bunch of optimizations will need to check if a cast is a nop or not, in order to avoid applying a basic heap type there (which is how this problem was detected). I'd rather not spread that complexity in the optimizer.

tlively: Makes sense. Can we delete `ref.cast_nop`, then? Having an incompatible implementation seems worse than removing it, and I agree about not wanting to spread complexity.

kripken: I was only thinking about leaving it because you said we might want it later? Leaving it, but just applying this PR to avoid problems with having casts to null, seems simplest to me, and less work overall if we end up needing it later.

tlively: I guess we can do that if you prefer having it in a broken state than deleting it altogether. Code LGTM in that case, but if it were up to me I would rather delete it than keep it in a broken state. It wouldn't be much work to bring back in the future if we need it.

jakobkummerow: By "broken state" you mean incompatibility with V8, right? We can update V8's understanding of `ref.cast_nop` to match the immediate of `0xfb41`, that's easy. And I don't think we need to worry about breaking changes to the binary format for this, because it's an unsafe experimental instruction that nobody should use in production modules anyway.

tlively (approved):
LGTM

jakobkummerow:
Updated in V8: crrev.com/5472313c96a96a7ef2c2f74b6a9b631bba1a107a

PR #5474, opened 2023-02-01
[Wasm GC] Fix struct.set / ref.as_non_null ordering issue

kripken:
If traps can happen, then we can't always remove a trap on null
on the ref input to `struct.set`, since it has two children,
```wat
(struct.set
  (ref.as_non_null X)
  (call $foo))
```
Removing the `ref.as` would not prevent a trap, as the `struct.set`
will trap, but it does move the trap to after the call which is bad.

### DIFF
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -1506,7 +1506,8 @@ struct OptimizeInstructions
               if (child == input) {
                 seenInput = true;
               } else if (seenInput) {
-                if (EffectAnalyzer(options, *getModule(), child).hasSideEffects()) {
+                if (EffectAnalyzer(options, *getModule(), child)
+                      .hasSideEffects()) {
                   return;
                 }
               }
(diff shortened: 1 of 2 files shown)


src/passes/OptimizeInstructions.cpp
```diff
@@ -1473,10 +1473,48 @@ struct OptimizeInstructions
...
+            // If there is such a child, it is a problem as mentioned above (it
+            // is fine for such a child to appear *before* |input|, as then we
+            // wouldn't be reordering effects).
+            bool seenInput = false;
+            for (auto* child : ChildIterator(parent)) {
+              if (child == input) {
+                seenInput = true;
+              } else if (seenInput) {
+                if (EffectAnalyzer(options, *getModule(), child)
+                      .hasSideEffects()) {
```
tlively, line 1510: Would it be sufficient to check for control flow side effects here? Updating a local certainly won't matter, for instance. I guess updating a global might matter, though?

But I guess you'll say that side effects are rare enough that it doesn't really matter and we should do the simplest thing here 👍

tlively (approved):

src/passes/OptimizeInstructions.cpp
kripken: Good point, we could do a little better here. We can also ignore traps, since traps are ok to reorder. I added a TODO as it seems low priority, and would probably require some tinkering in `effects.h`.

Issue #5343, opened 2022-12-13
Support try-catch in Asyncify / lower wasm exceptions to MVP


kripken:
See discussion in https://github.com/WebAssembly/binaryen/pull/5143 - lowering wasm exceptions to wasm MVP might be the best approach for this.

caiiiycuk:
Sorry, still misunderstand this. What is old js exceptions? asmjs exceptions?  

wasm+asyncify didn't support any exceptions even simplest one? like

```
try{
//...
} catch {
//...
}
```

I just trying to understand, is dosbox-x exceptions handling not working because of bug, or because try/catch is not implemented (this task).

kripken:
There are two modes, see https://emscripten.org/docs/porting/exceptions.html

Basically the old mode is `-fexceptions`. It works with Asyncify (but there might be bugs). The new mode is `-fwasm-exceptions`. That is much faster and smaller, but it doesn't work with Asyncify yet.

caiiiycuk:
Hi @tlively. I have motivation to implement this. I hope I don't bother you with some questions. From bird's-eye the  lower wasm exceptions to MVP seems doable. 

My first idea is to replace the try blocks with just a block with br statement. When a br occurs, I store the exception data somewhere in memory and then process the blocks associated with the catches until exception is consumed.

Something like this:
```
**FROM**
try 
  throw ...
catch
  ...

**TO**
block
  // store exception data
  br 1

// catch realted to thrown exception
block
  // load exception data
```

I excuse that it very raw, but I don't want to implement something not working :)

Anyway,
1. when we up to function level, **after each** function call we need to check if we processing some exception right now. This means that I need to put some logic after each call/call_indirect instruction. Do you think is it will bloat the binary very fast?
2. try/delegate instruction, I don't have an idea how to handle it, it breaks the linear execution flow, and makes this simple idea not working. Maybe there is option to not generate them, but again it will bloat binary.
3. Can we just get implementation from wasm2c? Even more what happen if I do wasm2c -> c2wasm loop, can it work as workaorund?

tlively:
This sounds like a good direction. I do think that having to check for exceptions after every call would bloat the binary, but it may not be avoidable. Maybe you could do an analysis to find calls that cannot possibly throw exceptions to skip instrumenting them.

For try-delegate, I expect you will have to put in more branch instructions to transfer control to the correct handler block.

[...]

PR #5475, opened 2023-02-02
Add partial support for -fwasm-exceptions in Asyncify (#5343)

caiiiycuk:
Hi. This PR does not do lowering wasm exceptions to MVP as suggested in #5343, so I need to explain situation:

I am trying to port a dosbox/dosbox-x core as is to WebAssembly. The dosbox core is syncrhnous, it runs emulation in forever while loop, events are processed inside this loop. To make this loop asynchrous I use Asyncify, however the unmodified core uses exceptions in certain situations.

Asyncify does not support exceptions, this PR should change the situation in some cases including dosbox/x case.
The main idea of emulation loop is follow:
```c++
void emulation() {
  do {
    em_sleep(); // <-- **START_UNWIND**
    try {
      int ret = cpudecode();
      // rest logic
    } catch (PageFault&) {
      // load failed page 
    } catch (Some&) {
      // do some 
    }
  } while (ret);
}

int cpudecode() {
  // some logic that also can call **emulation** recursively
}
```

The catch blocks is used to recover core, or to jump out from some of deep cpudecode calls. **There is no situation when emulation loop is called when catch block somewhere in stack:**

```c++
void some() {
  while () {
    try {
      ...
      some(); // can call start_unwind
    } catch () {
      // never calls to start_unwind
    }
  }
}
```

This  also means that catch block can't change the state, and so **only try block can**. But on other hand try block it self should works fine with current asyncify code. When we are doing rewind we are only visiting try blocks and never catch blocks, so it work out of the box.

For such type of synchronous code we can add exceptions support, we need only verify that unwind didn't start from a catch block (**when catch is not in stack**). We **can validate** this requirment and throw exception if it breaks or ignore the unwind call. 

This is very small overhead and simple solution to bring exceptions support in Asyncify, more over for some codebases it should be simple to rewrite the synchronous loop to not call unwind from a catch blocks.

This PR is tested on dosbox-x, it works fine. 
[emscripten-discuss initial context](https://groups.google.com/g/emscripten-discuss/c/jjLr6S-a87w/m/LQn7Elh6AgAJ?utm_medium=email&utm_source=footer)

---

There is one thing in impelementation that probably is not correct:
```c++
            auto decIf = builder->makeIf(
              br->condition,
              builder->makeSequence(builder->makeDec(amount), br),
              br->value);
            br->condition = nullptr;
            replaceCurrent(decIf);

```
I have a global variable that I need to decrease on some amount just before `br_if` instruction,
but only when condition is **true**. I tried lot of other variants but no luck:

1. Adding new instruction into condition body:

```c++
            Index newLocal = builder->addVar(func, br->condition->type);
            auto setLocal = builder->makeLocalSet(newLocal, br->condition);
            auto getLocal =
              builder->makeLocalGet(newLocal, br->condition->type);
            
            br->condition = builder->makeBlock(
              {setLocal,
               builder->makeIf(getLocal, builder->makeDec(amount)),
               getLocal},
              br->condition->type);
```

Result:
> Assertion failed: (!resultUsed || curr->type != Type::none), function optimize, file Vacuum.cpp, line 95.
```c++
>    // resultUsed only makes sense when the type is concrete
>    assert(!resultUsed || curr->type != Type::none);  <-- **ASSERT**
>    // If we actually need the result, then we must not change anything.
>    // TODO: maybe there is something clever though?
>    if (resultUsed) {
>      return curr;
>    }
```

2. Just wrap the br_if with new block:

```c++
            replaceCurrent(builder->makeSequence(
              builder->makeIf(br->condition, builder->makeDec(amount)), br));
```

Result:
> Assertion failed: (values.size() == 1), function getSingleValue, file wasm-interpreter.h, line 67.
```c++
>    if (curr->condition) {
>      Flow conditionFlow = visit(curr->condition);
>      if (conditionFlow.breaking()) {
>        return conditionFlow;
>      }
>      condition = conditionFlow.getSingleValue().getInteger() != 0; // <-- **ASSERT**
```

3. Just wrap the br_if but save condition result into local var like in 1. Results is sames as in 2.

4. Completely replace br_if with if and br:

```c++
            auto decIf = builder->makeIf(
              br->condition,
              builder->makeSequence(builder->makeDec(amount), br));
            br->condition = nullptr;
            replaceCurrent(decIf);
```

Results:
> Works fine on small wasm files
> On big wasm file have this error:
>
> [wasm-validator error in function CDROM_Interface_Image::LoadCueSheet\28char*\29] unexpected true: if without else must not return a value in body, on 
```wast
>(if
> (i32.and
>  (i32.load8_u offset=16
>   (i32.add
>    (local.get $8)
>    (i32.load
>     (i32.sub
>      (i32.load
>       (local.get $8)
>      )
>      (i32.const 12)
>     )
>    )
>   )
>  )
>  (i32.const 5)
> )
> (block (result i32)
>  (global.set $__asyncify_catch_counter
>   (i32.sub
>    (global.get $__asyncify_catch_counter)
>    (i32.const 2)
>   )
>  )
>  (br $label$1
>   (i32.const 0)
>  )
> )
>)
```
OR
> [wasm-validator error in function CDROM_Interface_Image::LoadCueSheet\28char*\29] unexpected false: can only drop a valid value, on 
```wast
>(drop
> (if
>  (i32.and
>   (i32.load8_u offset=16
>    (i32.add
>     (local.get $8)
>     (i32.load
>      (i32.sub
>       (i32.load
>        (local.get $8)
>       )
>       (i32.const 12)
>      )
>     )
>    )
>   )
>   (i32.const 5)
>  )
>  (block (result i32)
>   (global.set $__asyncify_catch_counter
>    (i32.sub
>     (global.get $__asyncify_catch_counter)
>     (i32.const 2)
>    )
>   )
>   (br $label$1
>    (i32.const 0)
>   )
>  )
> )
>)
```

5. I thinking to not use replaceCurrent, find the parent block of br_if and insert builder->makeDec just before br, but seems even more cumbersome cause parent block can be `Block`, `Loop`, `If`, `Drop`, `Try`, etc. I don't have straightforward algorithm in my mind.

Thanks!


src/passes/Asyncify.cpp
```diff
@@ -1214,6 +1244,222 @@ struct AsyncifyFlow : public Pass {
...
+    };
+    NestedLevelWalker nestedLevelWalker;
+    nestedLevelWalker.expressionNestedLevel = &expressionNestedLevel;
+    nestedLevelWalker.walk(func->body);
+
+    // with this walker we will handle those counters:
+    // - entering into catch (= pop)  +1
+    // - return                       -N (nested catch count up to root)
+    // - break                        -N (nested catch count up to label)
+    // - exiting from catch           -1
```
kripken, line 1328: I think a simpler approach might be to use a `try-finally` pattern? That would also handle nested try-catch, for example. Wasm doesn't have finally, but you can do `catch_all + rethrow`:
```wat
(try
  (do
    COUNTER++;
  )
  (catch_all
    COUNTER--;
    (rethrow 0)
  )
)
```

kripken (commented):
Interesting idea! I think this makes sense to do.

src/passes/Asyncify.cpp
caiiiycuk: Hi, I alreay use try/catch_all pattern, for all catch blocks, but return/break can't be handled by try/catch_all, so they are also needed. But, I realized that I need modify only root catch block, because we not interested to count nested blocks. I will try to add lit tests, and refactor this...

kripken: Oh, right, you also need return/break code too... I forgot `catch_all` is not really `finally`...

caiiiycuk:
Refactored, now I change only top-level catch blocks. Also added lit test, for me seems correct. Dosbox-x works fine. 
Only here could be a problem:
```c++
// br_if case with condition and value

            Index newLocal = builder->addVar(func, br->value->type);
            auto setLocal = builder->makeLocalSet(newLocal, br->value);
            auto getLocal = builder->makeLocalGet(newLocal, br->value->type);
            auto condition = br->condition;
            br->condition = nullptr;
            br->value = getLocal;
            auto decIf =
              builder->makeIf(condition,
                              builder->makeSequence(builder->makeDec(), br),
                              getLocal);
            replaceCurrent(builder->makeSequence(setLocal, decIf));
```

Not sure how to deal if sideeffect will be in `setLocal`, currenlty it executes before condition....
@kripken please take a look

src/passes/Asyncify.cpp
```diff
@@ -1214,6 +1244,234 @@ struct AsyncifyFlow : public Pass {
...
+    };
+    NestedLevelWalker nestedLevelWalker;
+    nestedLevelWalker.expressionCatchCount = &expressionCatchCount;
+    nestedLevelWalker.walk(func->body);
+
+    // with this walker we will handle those changes of counter:
+    // - entering into catch (= pop)  +1
+    // - return                       -1
+    // - break                        -1
+    // - exiting from catch           -1
```
kripken, line 1333: Maybe I don't understand what the counter is. In the previous class we counted the nesting level of catch bodies. Why do we want to count say a break as -1? (and why -1 and not some other value?)

caiiiycuk: We need to understand are we inside a catch block when we call 'unwind'. No matter are we inside only top-level catch or some nested catch. Obviously entering or exiting a nested catch didn’t change the fact that we were already in the catch block. Because of that I change counter only when entering/exiting inside top-level catches. 

The comment is not clear, I think it should be:
```
    // - entering into catch (= pop)                +1
    // - return                                     -1
    // - break (target outside of top-level catch)  -1
    // - break (target inside of top-level catch)    0 (ignored)
    // - exiting from catch                         -1
```

src/passes/Asyncify.cpp
```diff
@@ -1134,6 +1154,16 @@ struct AsyncifyFlow : public Pass {
       } else if (doesCall(curr)) {
         results.push_back(makeCallSupport(curr));
         continue;
+      } else if (auto* iTry = curr->dynCast<Try>()) {
+        if (item.phase == Work::Scan) {
```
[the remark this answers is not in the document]
caiiiycuk: correct, will add

src/passes/Asyncify.cpp
```diff
@@ -238,6 +249,11 @@
 //      an unwind/rewind in an invalid place (this can be helpful for manual
 //      tweaking of the only-list / remove-list, see later).
 //
+//   --pass-arg=asyncify-ignore-unwind-from-catch
+//
+//      This enables extra check before unwind, if it called from within catch
+//      block then it silently ignored (-fwasm-exceptions support)
```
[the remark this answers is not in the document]
caiiiycuk: it silently ignore, will try to write better comment)

src/passes/Asyncify.cpp
```diff
@@ -163,6 +170,10 @@
 //    calls, so that you know when to start an asynchronous operation and
 //    when to propagate results back.
 //
+//  * asyncify_get_catch_counter(): call this to get the current value of the
+//    internal "__asyncify_catch_counter" variable (only when assertions
+//    or ignore mode are enabled).
```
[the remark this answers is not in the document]
caiiiycuk: Just when `unreachable execution` happens you can call this function to see how many functions in stack uses catches before unwind. Like:
```
0 - unreachable happened because of some other assertion (probably not-in asyncify list)
1 - only one function involved
2 - two functions
3 - etc.
```

Not very handy because you can't understand the exact function names. But check for zero is useful, because there is no other way to understand if problem related to catches or non-asyncify function.

Better solution is to write in to console the reason of unreachable just before it happens, but idk is it possible somehow.

src/passes/Asyncify.cpp
caiiiycuk: Even more:

```
    // - entering top-level catch (= pop)           +1
    // - entering nested catch (= pop)               0 (ignored)
    //
    // - return inside top-level/nested catch       -1
    // - return outside top-level/nested catch       0 (ignored)
    //
    // - break target outside of top-level catch    -1
    // - break target inside of top-level catch      0 (ignored)
    // - break outside top-level/nested catch        0 (ignored)
    //
    // - exiting from top-level catch               -1
    // - exiting from nested catch                   0 (ignored)
```

kripken: Thanks, I see. I think what's confusing me is that more than one class is used to achieve this? I think there's a simpler way. See `ExpressionStackWalker` - you can use that traversal to get the stack of all the parents. Then you can simply walk up the stack to see if you are inside a catch body or not.

src/passes/Asyncify.cpp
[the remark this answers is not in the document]
kripken: Wouldn't the stack trace in the VM show the function names? I think I understand the extra info you are trying to add here, but if a user runs in a browser debugger I think they can get enough to debug with? The stack trace + they can see, visually, if they are inside a catch or not.

allsey87:
@kripken is there anything preventing this PR from being merged?

caiiiycuk:
I'm sorry, I want to finish this and add the changes suggested by @kripken, I'll try to find some time to finish it (not sure I can do it anytime soon). Btw I've used it many times and it works great in the case of js-dos.

allsey87:
Excellent work @caiiiycuk! I look forward to seeing this merged!

allsey87:
@caiiiycuk I am currently testing your branch and have updated the binaries `emsdk/upstream/bin` with those from the build. However I am getting some errors when I try to build my test program:

> em++: warning: ASYNCIFY=1 is not compatible with -fwasm-exceptions. Parts of the program that mix ASYNCIFY and exceptions will not compile. [-Wemcc]
error: DISABLE_EXCEPTION_THROWING was set (likely due to -fno-exceptions), which means no C++ exception throwing support code is linked in, but exception catching code appears. Either do not set DISABLE_EXCEPTION_THROWING (if you do want exception throwing) or compile all source files with -fno-except (so that no exceptions support code is required); also make sure DISABLE_EXCEPTION_CATCHING is set to the right value - if you want exceptions, it should be off, and vice versa.

I think the warning is fine. But I am confused as to why `DISABLE_EXCEPTION_THROWING` is being set. I have added `-fwasm-exceptions` to both the linker flags and compiler flags (both to the C++ and C flags (I guess the latter is not necessary)). Do have any suggestions on how to fix this? I am using Emscripten 3.1.32.

caiiiycuk:
I just it like this
```
            -fwasm-exceptions
            "-sASYNCIFY=1"
            "-sASYNCIFY_IMPORTS=['syncSleep']"
            "-sASYNCIFY_WHITELIST=@${TARGETS_DIR}/dosbox-x-asyncify.txt"
```

Did you check, maybe you set flag `-fno-exceptions` somewhere. 
Also you need to set `-fwasm-exceptions` both on compile and linking stage

src/passes/Asyncify.cpp
caiiiycuk: Done!

caiiiycuk:
@kripken I apologize for the delay. I have made all the changes you requested. Please take a look and let me know if there's anything else to do.

caiiiycuk:
@kripken friendly ping

allsey87:
I have been using this for several weeks now and I haven't found any issues yet! Excellent work @caiiiycuk!

caiiiycuk:
@allsey87 thanks, great to know

kripken:
@caiiiycuk Sorry, I've been meaning to get to this but have been busy. Thanks for updating this!

There has recently been a suggestion to make some changes in how wasm exceptions works in the spec, see https://github.com/WebAssembly/exception-handling/issues/280 - I have no idea what will happen there, but it might mean that if we land this we'd need to then refactor it significantly. On the positive side, they do mention it is a goal to make things like Asyncify easier, so perhaps this will become simpler actually. Either way, though, I think we should wait on this PR to see what the spec changes are first, to avoid wasted work for all of us.

caiiiycuk:
@kripken No worries, all good. At least it works for my case. :)

fs-eire:
Is there any chance that we can merge this? JSPI still need to take some time...

kripken:
The new wasm EH proposal I mentioned [above](https://github.com/WebAssembly/binaryen/pull/5475#issuecomment-1652418845) is in the process of being implemented now. The main component is this:

https://github.com/WebAssembly/binaryen/blob/3049fb83c3f9c17e78157c43b5592055732db8ff/src/wasm.h#L1473-L1474

If we want to land this PR then targeting that would make sense, as the old EH will be removed. However, the new wasm EH is not done yet (e.g. parts of the optimizer don't work on it) so it's too early for that, I'm afraid.

hoodmane:
@kripken @sbc100 Is there any chance we can get this merged? We are trying to move Rust over to using exception handling so that it will be JSPI compatible. See the discussion here:
https://github.com/rust-lang/rust/pull/147224

The fact that only wasm-eh supports JSPI and only JS eh supports ASYNCIFY makes the migration pathway very hard. Runtime support for new wasm EH is also still very limited and it will be years before we can use it everywhere. So we need this as an intermediate step.

kripken:
@hoodmane Is the idea to move rust to wasm EH while still using Asyncify, then move it to JSPI later (still using wasm EH)?

> Runtime support for new wasm EH is also still very limited and it will be years before we can use it everywhere

Is there really that big a time difference between legacy wasm EH and new wasm EH? The [features page](https://webassembly.org/features/) suggests major browsers have supported the latter for over 6 months now, if I read that right.

(But, separately: Yes, if this is generally useful in the meantime, which it sounds like it might be given comments here, I am open to this getting reviewed and landed.)

hoodmane:
> Is the idea to move rust to wasm EH while still using Asyncify, then move it to JSPI later (still using wasm EH)?

Exactly. The full story:

Rust is only willing to support one stable Emscripten EH ABI.

I want to use JSPI for Pyodide so I want Rust to support legacy wasm eh. Asyncify has unacceptable performance for my use case. Pyodide is set up so that it works either way but stack switching will fail at runtime if JSPI support is not there. On the other hand, I'm not willing to use the new error handling yet for Pyodide since it isn't supported enough.

Other people want to use ASYNCIFY because they need stack switching in all cases. They want an exception handling ABI which is:
1. not new EH (not enough runtime support yet)
2. support for ASYNCIFY
3. supported by stable Rust

Me changing the Rust stable EH ABI from JS EH to legacy wasm EH is thus creating a problem for them.

kripken:
Is it perhaps an option for Rust to move to (either) wasm EH, and people that are using Asyncify (which is not compatible with wasm EH) can use an older Rust?

I believe Asyncify is not a heavily-used feature, so yes using older Rust may be an inconvenience, but hopefully not an insurmountable one, and for not many people?

hoodmane:
I am perfectly happy with that approach, I just want Rust to switch to wasm-eh by any means. @curiousdannii is using ASYNCIFY with Rust and is less happy about it. I'm not sure what the standards are for whether or not we block a target change, maybe @workingjubilee can say more about how the Rust project decides this sort of question.

hoodmane:
It's also worth noting that it will continue to be possible to use JS eh with nightly Rust but @curiousdannii doesn't want to do that either. Currently Pyodide exclusively uses nightly Rust so we can use wasm eh, so it would be a question of who is privileged. I would argue that Pyodide is a substantially larger project and we should get preference.

But since this PR already exists, if all it would take from the Emscripten side is to merge it it would remove the whole knot. When JSPI gets universal enough support, we can remove it.

hoodmane:
According to @purplesyringa, z3-sys is the most popular project that would be broken by this change:
https://docs.rs/z3-sys/latest/z3_sys/

purplesyringa:
@hoodmane That might have been misleading, sorry -- z3-sys does not use `-sASYNCIFY` and would not be broken by this deficiency in emscripten, but rather by the fact that the EH ABI is changed at all because it links C++ code compiled with `-fexceptions`. That's a separate problem.

curiousdannii:
I'll switch to Rust nightly if I need to. Better than being forever stuck on an old Rust version until JSPI is usable (probably several years still).

kripken:
@hoodmane 

> But since this PR already exists, if all it would take from the Emscripten side is to merge it it would remove the whole knot. When JSPI gets universal enough support, we can remove it.

Fair enough, yeah, it sounds like there are a lot of moving pieces here, and the simplest thing for everyone might be to land this.

@caiiiycuk Would you have time to finish this for landing? If so I can review this in detail, and we'll go from there.

caiiiycuk:
@kripken yep, I’m already trying to keep it up to date — usually I just rebase it, and it works out of the box. Last time I used it with a combination of em 4.0.6, and it worked just fine.

@hoodmane 
However, I’m a bit out of context. Is it okay that it only partially supports exceptions? I mean, the codebase shouldn’t change the Asyncify state from within catch blocks — is that acceptable for the named projects?

hoodmane:
We'll have to find out I guess.

hoodmane:
I think it should be enough because they are Rust projects and they don't use `catch_unwind()` at all so they have no catch blocks at all.

curiousdannii:
I'm linking a Rust static lib with C++ code which does use exceptions... But I have no idea how to even find out if this PR would be enough (other than eventually trying it).

caiiiycuk:
@kripken, I rebased it, please review.

workingjubilee:
> I'm not sure what the standards are for whether or not we block a target change, maybe @workingjubilee can say more about how the Rust project decides this sort of question.

We prefer to make decisions by consensus but in a case where there simply is no way of completely satisfying both parties, we are forced to pick, no? We could pick differently than one prefers, but as mentioned, preference mostly goes out the window if the emsdk simply supports the combinatorics necessary.

>  However, I’m a bit out of context. Is it okay that it only partially supports exceptions? I mean, the codebase shouldn’t change the Asyncify state from within catch blocks — is that acceptable for the named projects?

I'm not entirely sure, but this sounds like it should be fine. It is fairly rare for Rust code to tamper that much with how unwinding happens, and the result of throwing Rust panics into C++ or C++ exceptions into Rust can be quite explosive. I mean, on the Rust side the program _can_ terminate correctly, but we often wind up picking "terminate" if any other path would be UB, and *throwing* a foreign exception into C++ can be very that. So it seems to me to be very unlikely that @curiousdannii has been successfully linking with code that extensively micromanages the unwinding state instead of just doing normal throw-catch-maybe-throw-again patterns.

curiousdannii:
Yeah I haven't been doing anything tricky.

The one catch I see is that indirect calls typically need to be instrumented for Asyncify, and Rust seems to use them a lot. It's very plausible that the Rust exceptions code would use indirect calls. I don't know whether that use would be covered by this PR or not. 

You can opt specific functions out of Asyncify instrumentation (asyncify_remove), but it's hard to know when it's safe to. My asyncify_advise log has over 1500 lines.

workingjubilee:
Well. Our formatting code typically goes through a vtable for size reasons, which I believe winds up with an indirect call, and while if memory serves the unwinding itself doesn't involve formatting, the panic messaging sure does.

workingjubilee:
It is a tad too late in the day for me to explain in detail what the difference is between "starting a panic", "running the panic handler" (which usually is the thing that does the message printing, if I am not somehow misremembering everything...), and "unwinding (due to a panic)" but these are technically all their own little thing and it's not entirely clear to me what ASYNCIFY needs and doesn't need, here.

This is why "inscrutable binary editing with unclear parameters but it works, probably" is usually something to which we say, "Have fun with that!" Except that it is apparently core to some usage of the emscripten target, and, well, if we don't support what the emsdk can do, then what exactly *does* define the emscripten target for us? That is a philosophical question which may be useful to address one day but also may be nice to kick down the road a bit.

caiiiycuk:
Yes, sometimes it’s really hard to understand all the technical details, especially when you don’t fully know the supporting technology.

If we simplify things a lot, code like this is considered normal:

```cpp
while (true) {
    try {
       // ...
       asyncify_sleep(10);
    } catch (...) {
       // anything
    }
}
```

while code like this is not supported:

```cpp
while (true) {
    try {
       // ...
       asyncify_sleep(10);
    } catch (...) {
       // anything
       asyncify_sleep(10); // <--- not supported
    }
}
```

I believe this approach will succeed because code is typically written like in block 1 — a `catch` block is used to handle an exception and then return the execution flow to a valid, normal state.

At least in the case of incorrect code organization, we’ll see a clear, understandable error indicating that asyncify_sleep was called in the wrong place.

Overall, if we look only at Emscripten, I don’t see any negative side effects from merging this PR. I’ve been maintaining it for about two years without major issues.

src/passes/Asyncify.cpp
```diff
@@ -96,6 +96,13 @@
 // Overall, this should allow good performance with small overhead that is
 // mostly noticed at rewind time.
 //
+// Exceptions handling (-fwasm-exceptions) is partially supported. Asyncify
+// can't start unwind operation when a catch block is in the stack trace.
+// If assertions mode is enabled then pass will check if unwind called from
+// within catch block or not, and if so throw an unreachable exception.
+// If "ignore unwind from catch" mode is enable then Asyncify will skip
+// any unwind call from within catch block.
```
kripken:
```suggestion
// Exceptions handling (-fwasm-exceptions) is partially supported, everything
// except for handling unwinding from within a catch block. If assertions mode
// is enabled then this pass will check for that problem, and if so, throw an
// unreachable exception. (If "ignore unwind from catch" mode is enabled then
// Asyncify will silently skip any unwind call from within catch blocks, see
// below.)
```

src/passes/Asyncify.cpp
```diff
@@ -239,6 +246,12 @@
 //      an unwind/rewind in an invalid place (this can be helpful for manual
 //      tweaking of the only-list / remove-list, see later).
 //
+//   --pass-arg=asyncify-ignore-unwind-from-catch
+//
+//      This enables additional check to be performed before unwinding. In
+//      cases where the unwind operation is triggered from the catch block,
+//      it will be silently ignored (-fwasm-exceptions support)
```
kripken:
```suggestion
//      When an unwind operation is triggered from inside a wasm-exceptions
//      catch block, which is not supported, silently ignore it rather than
//      fail during rewinding later.
```

src/passes/Asyncify.cpp
```diff
@@ -1218,6 +1248,202 @@ struct AsyncifyFlow : public Pass {
   }
 };
 
+// Add catch block counters to verify that unwind is not called from catch
```
kripken:
```suggestion
// Add catch block counters to verify that unwinding is not done from catch
```

src/passes/Asyncify.cpp
```diff
@@ -1218,6 +1248,202 @@ struct AsyncifyFlow : public Pass {
...
+// block.
```
kripken: Could we depend on the IR structure to find if we are inside a catch? See `ExpressionStackWalker` which maintains a stack of parents. `findBreakTarget()` there shows an example of scanning the parents.

caiiiycuk: Cool, I will try to adopt this.

curiousdannii:
> At least in the case of incorrect code organization, we’ll see a clear, understandable error indicating that asyncify_sleep was called in the wrong place.

At compile time or run time?

caiiiycuk:
> > At least in the case of incorrect code organization, we’ll see a clear, understandable error indicating that asyncify_sleep was called in the wrong place.
> 
> At compile time or run time?

At runtime. However, you can also avoid this by using a special flag that ignores `asyncify_sleep` in incorrect situations — though the app might get stuck there in forever loop.

curiousdannii:
Specifically asyncify_sleep or all async functions?

The other big question is if you think async in catch is a fundamental incompatibility, or instead a big difficulty that might be supportable with a lot more work?

fs-eire:
Is it possible to use any compile time analysis tool or runtime check to tell whether any rewind/unwind happening inside a catch block?

caiiiycuk:
> Specifically asyncify_sleep or all async functions?
> 
> The other big question is if you think async in catch is a fundamental incompatibility, or instead a big difficulty that might be supportable with a lot more work?

Any async function — they’re all implemented the same way. But I don’t know if it’s even possible to make Asyncify work inside a `catch` block. There’s also an attempt here: #5143 — maybe if that idea is developed further, something could come out of it. Still, I think Asyncify is rarely used in practice, and Asyncify inside a `catch` block is an even rarer case. It seems easier to just rewrite the code to avoid the need for infinite loops inside `catch` blocks altogether.



> Is it possible to use any compile time analysis tool or runtime check to tell whether any rewind/unwind happening inside a catch block?

Yes, to some extent — it’s similar to automatically generating a list of functions that modify `asyncify_state`. But if there’s a virtual function call inside (through a pointer), then you really can’t be sure anymore.

caiiiycuk:
> Is it possible to use any compile time analysis tool or runtime check to tell whether any rewind/unwind happening inside a catch block?

Potentially, it’s possible to implement a check with two outcomes:

1. 100% no `catch` block calls an async function (**ALL GOOD**)
2. Some functions that modify the `asyncify` state might be called from `catch` + list of this functions.

src/passes/Asyncify.cpp
caiiiycuk: @kripken 

I have been thinking about this task, and so far I do not understand how to make it more efficient. We do not actually need a parent stack, because we do not care which nested catch block we are currently in. Essentially, I only want to know whether we are inside a catch or not. In the current code, we have a counter for this, which is effectively just a boolean flag. If I am not mistaken, it is always either 0 or 1, which means we could add additional assert that checks it falls within that range.

The core problem is that at compile time we cannot determine whether a function that changes the Asyncify state called inside a catch or not. More precisely, we could determine that if there were no virtual function calls. I have not yet found a way to work around this at compile time, which is why a counter is used.

I still do not quite understand how to apply a parent stack here. I could maintain such a stack at runtime, but that seems worse than just using an integer counter.

**Or** is the idea to keep the counter, but update it using the information obtained from _ExpressionStackWalker_?

caiiiycuk: Oh, obviously it can be greater than 1, but at least it cannot be less than 0.

kripken: > The core problem is that at compile time we cannot determine whether a function that changes the Asyncify state called inside a catch or not.

We may not know if there is a catch further up, at runtime, but we will reach it when we unwind, I think? So the compile-time knowledge seems enough. Here is what I mean:
```js
function foo() {
  try {
    ..
  } catch (e) {
    bar();
    // after call to bar
  }
}

function bar() {
  importThatPauses();
}
```
Say we reach that call to bar, then it calls the import that pauses. Inside bar, we don't know that we are inside a catch from earlier. But after bar unwinds normally, we reach "`after call to bar`" and we can have a static assert there about not unwinding.

That is, each catch block can, statically, get added checks to not unwind, after each possibly-unwinding operation there?

kripken: (and iirc we already have a way to add checks for "`should not be unwinding here`", so I hope this is not hard to add)

caiiiycuk: Ahh, sounds correct. But then we cannot support `--pass-arg=asyncify-ignore-unwind-from-catch` if I understood correctly. But we can add those static checks in case if this flag is not passed.

kripken: I see now, thanks. So that option makes it necessary to know immediately if we are nested under a catch...

src/passes/Asyncify.cpp
```diff
@@ -1218,6 +1248,202 @@ struct AsyncifyFlow : public Pass {
   }
 };
 
+// Add catch block counters to verify that unwinding is not done from catch
+// block.
```
kripken:
```suggestion
// Add catch block counters to verify that unwinding is not done from inside
// a catch block. Note that we need to know if we are nested under a catch
// immediately - we can't depend on unwinding noticing it when we reach
// that parent catch - because of the option asyncify-ignore-unwind-from-catch.
```

src/passes/Asyncify.cpp
```diff
@@ -1218,6 +1248,202 @@ struct AsyncifyFlow : public Pass {
...
+    //
+    // - return inside top-level/nested catch       -1
+    // - return outside top-level/nested catch       0 (ignored)
+    //
+    // - break target outside of top-level catch    -1
+    // - break target inside of top-level catch      0 (ignored)
+    // - break outside top-level/nested catch        0 (ignored)
+    //
+    // - exiting from top-level catch               -1
+    // - exiting from nested catch                   0 (ignored)
```
kripken: Unfortunately, detecting an exit from a catch is not as simple as this. You would also need to handle `br_on` and `return_call`, off the top of my head.

We could in theory say this feature doesn't support GC or tail calls, but imagine this is linked with a module using those features, and then inlining happens, so it seems worrying.

How important is `asyncify-ignore-unwind-from-catch`? Without that, we could use the more static approach I mentioned before.

kripken: (Oh, and also `try_table` - people could mix exnref EH and non-exnref EH.)

caiiiycuk: Personally I don't use asyncify-ignore-unwind-from-catch, I provided it for completeness, and because it is easy. So, I think can remove it, but can't guess if it can be used by someone else. Also I think it is easy to fix in code.

caiiiycuk: I mean if needed devs can fix code to support something like ignore this asyncify_sleep

kripken: Ok, I don't think anyone else mentioned a need for that option, so maybe it is best to remove it. Can maybe wait a bit to see if anyone on this PR responds.

workingjubilee:
> while code like this is not supported:
> 
> ```c++
> while (true) {
>     try {
>        // ...
>        asyncify_sleep(10);
>     } catch (...) {
>        // anything
>        asyncify_sleep(10); // <--- not supported
>     }
> }
> ```
> 
> I believe this approach will succeed because code is typically written like in block 1 — a `catch` block is used to handle an exception and then return the execution flow to a valid, normal state.
> 

Hmm, that may be slightly easier to run into than I thought! Though I still wouldn't expect it to happen via the Rust side because of how Rust code uses exceptions only, well, exceptionally.

@curiousdannii It *may* be possible to mitigate the indirection problem if necessary... or at least help you with the `ASYNCIFY_REMOVE` list... so please feel free to ask us about this if you wind up having troubles after updating to an emcc with this branch in it. Though not in this thread, obviously, I mean a new issue or asking on Zulip or such.

curiousdannii:
Even with JS exceptions I need to work out what I can asyncify_remove. Here's what I'm working with. Wish me luck 😂
[glulxe-log.txt](https://github.com/user-attachments/files/23406702/glulxe-log.txt)

Edit: great news: I've discovered that at least one of the apps I build in my project can go back to using ASYNCIFY_IGNORE_INDIRECT. I'd disabled it when I first switched to Rust. Possibly changing to panic=abort made the difference.

caiiiycuk:
@kripken 

Please take a look at the last two commits. I attempted to implement the following algorithm:

1. We iterate over all function calls.
2. We determine whether the call is made inside a `catch` block.
3. If it is, we insert a check after the call to ensure that `ASYNCIFY_STATE` is still `NORMAL`.

More details on [point 2](https://github.com/WebAssembly/binaryen/pull/5475/files#diff-8d9a63ebb26e7ffd05e820c49277e5571ea7764bbcfe547d624f979a0cd42489R1396-R1409), since I am not entirely sure the current implementation covers all cases—mainly because I may not be aware of every detail.

I use `ExpressionStackWalker` to locate the nearest parent `Try` block. Once it is found, I need to determine whether the call occurs inside the `try` body or inside one of the `catch` blocks. In practice, it is enough to check whether the current instruction belongs to the `try.body`. To do this, I simply take the first child of the `Try` block from the stack, and if that child is `try->body`, then the call is considered safe; otherwise, I insert the corresponding check.

I tested this code with js-dos, and so far I haven’t encountered any issues. However, I am still not completely confident in it overall.

src/passes/Asyncify.cpp
```diff
@@ -239,6 +246,12 @@
 //      an unwind/rewind in an invalid place (this can be helpful for manual
 //      tweaking of the only-list / remove-list, see later).
 //
+//   --pass-arg=asyncify-ignore-unwind-from-catch
+//
+//      When an unwind operation is triggered from inside a wasm-exceptions
+//      catch block, which is not supported, silently ignore it rather than
+//      fail during rewinding later.
```
kripken:
```suggestion
//      fail during rewinding later. (This is unsafe in general.)
```

src/passes/Asyncify.cpp
```diff
@@ -317,13 +330,16 @@
...
+#include "ir/branch-utils.h"
 #include "ir/effects.h"
+#include "ir/eh-utils.h"
 #include "ir/find_all.h"
 #include "ir/linear-execution.h"
 #include "ir/literal-utils.h"
 #include "ir/memory-utils.h"
 #include "ir/module-utils.h"
 #include "ir/names.h"
+#include "ir/parents.h"
```
kripken:
```suggestion
```

src/passes/Asyncify.cpp
```diff
@@ -317,13 +330,16 @@
 
 #include "asmjs/shared-constants.h"
 #include "cfg/liveness-traversal.h"
+#include "ir/branch-utils.h"
 #include "ir/effects.h"
+#include "ir/eh-utils.h"
```
kripken:
```suggestion
```

src/passes/Asyncify.cpp
```diff
@@ -317,13 +330,16 @@
 
 #include "asmjs/shared-constants.h"
 #include "cfg/liveness-traversal.h"
+#include "ir/branch-utils.h"
```
kripken:
```suggestion
```

src/passes/Asyncify.cpp
```diff
@@ -1316,6 +1344,80 @@ struct AsyncifyAssertInNonInstrumented : public Pass {
...
+  AsyncifyAssertUnwindCorrectness(ModuleAnalyzer* analyzer, Module* module) {
+    this->analyzer = analyzer;
+    this->module = module;
+  }
+
+  std::unique_ptr<Pass> create() override {
+    return std::make_unique<AsyncifyAssertUnwindCorrectness>(analyzer, module);
+  }
+
+  bool isFunctionParallel() override { return true; }
```
kripken: As a convention we prefer this at the top of the class, before all other internals.

src/passes/Asyncify.cpp
```diff
@@ -1316,6 +1344,80 @@ struct AsyncifyAssertInNonInstrumented : public Pass {
...
+
+  void runOnFunction(Module*, Function* function) override {
+    auto builder = std::make_unique<Builder>(*module);
+
+    struct UnwindWalker : WalkerPass<ExpressionStackWalker<UnwindWalker>> {
+      Function* function;
+      Builder* builder;
+
+      void replaceCallWithCheck(Call* call) {
```
kripken:
```suggestion
      // Adds a check for Call that is inside a Catch block (we do not handle unwinding there).
      void checkCallInsideCatch(Call* call) {
```

src/passes/Asyncify.cpp
```diff
@@ -1316,6 +1344,80 @@ struct AsyncifyAssertInNonInstrumented : public Pass {
...
+              call,
+              check,
+            },
+            call->type));
+        }
+      }
+
+      void visitCall(Call* curr) {
+        assert(!expressionStack.empty());
+        std::vector<int> parents;
```
kripken:
```suggestion
```

src/passes/Asyncify.cpp
```diff
@@ -1316,6 +1344,80 @@ struct AsyncifyAssertInNonInstrumented : public Pass {
...
+        Index i = expressionStack.size() - 1;
```
kripken:
```suggestion
        // Go up the stack and see if we are in a Catch.
        Index i = expressionStack.size() - 1;
```

src/passes/Asyncify.cpp
```diff
@@ -1316,6 +1344,80 @@ struct AsyncifyAssertInNonInstrumented : public Pass {
...
+            assert(i + 1 < expressionStack.size());
+            if (expressionStack[i + 1] != aTry->body) {
+              replaceCallWithCheck(curr);
+            }
+            break;
+          }
+          i--;
+        }
+      };
+    };
```
kripken: We should also handle CallIndirect and CallRef here.

src/passes/Asyncify.cpp
```diff
@@ -1316,6 +1344,80 @@ struct AsyncifyAssertInNonInstrumented : public Pass {
...
+            {
+              call,
+              check,
+            },
+            call->type));
+        }
+      }
+
+      void visitCall(Call* curr) {
+        assert(!expressionStack.empty());
```
kripken: A `return_call` (`curr->isReturn`) can be ignored here: It returns first, leaving the Catch, before calling.

src/passes/Asyncify.cpp
```diff
@@ -1316,6 +1344,80 @@ struct AsyncifyAssertInNonInstrumented : public Pass {
...
+  }
+
+  std::unique_ptr<Pass> create() override {
+    return std::make_unique<AsyncifyAssertUnwindCorrectness>(analyzer, module);
+  }
+
+  bool isFunctionParallel() override { return true; }
+
+  void runOnFunction(Module*, Function* function) override {
+    auto builder = std::make_unique<Builder>(*module);
```
kripken: Builders are very cheap to construct - please create one in `replaceCallWithCheck`, as needed, which would be simpler.

src/passes/Asyncify.cpp
```diff
@@ -1814,14 +1920,14 @@ struct Asyncify : public Pass {
     module->addGlobal(std::move(asyncifyData));
   }
 
-  void addFunctions(Module* module) {
+  void addFunctions(Module* module, bool asserts) {
     Builder builder(*module);
     auto makeFunction = [&](Name name, bool setData, State state) {
+      auto* body = builder.makeBlock();
```
kripken: Why move this?

src/passes/Asyncify.cpp
```diff
@@ -1814,14 +1920,14 @@ struct Asyncify : public Pass {
     module->addGlobal(std::move(asyncifyData));
   }
 
-  void addFunctions(Module* module) {
+  void addFunctions(Module* module, bool asserts) {
```
kripken: New param seems unused?

src/passes/Asyncify.cpp
```diff
@@ -1759,8 +1861,12 @@ struct Asyncify : public Pass {
       // Add asserts in non-instrumented code. Note we do not use an
       // instrumented pass runner here as we do want to run on all functions.
       PassRunner runner(module);
-      runner.add(std::make_unique<AsyncifyAssertInNonInstrumented>(
-        &analyzer, pointerType, asyncifyMemory));
+      if (asserts) {
```
kripken: This is already inside an `if (asserts)`, line 1860

caiiiycuk:
@kripken PR updated, please check. If everything is ok, then I will squash into one commit.

kripken:
(It might be nice to also test `call_indirect` and `call_ref`, but I don't want to hold up landing on that, given the intended use cases. Based on feedback we can improve later.)

kripken:
Oh, before I land, please provide the commit message. (Either update comment 0, or squash and put it in there, or here in a comment.)

caiiiycuk:
> Oh, before I land, please provide the commit message. (Either update comment 0, or squash and put it in there, or here in a comment.)

What about current one? 
"Add partial support for -fwasm-exceptions in Asyncify"

kripken:
Ok, sounds good. I'll elaborate a little in the text.

hoodmane:
So to get this out we need a release of Binaryen and then a release of Emscripten that bumps the Binaryen version. What's the Binaryen release policy? Can we get a release for this?

hoodmane:
(Also, thanks @caiiiycuk for finishing this two and a half years after the PR was opened!)

sbc100:
> So to get this out we need a release of Binaryen and then a release of Emscripten that bumps the Binaryen version. What's the Binaryen release policy? Can we get a release for this?

Not really answering you question, but you can always get the latest emscripten+binaryen with `./emsdk install tot` (normally lags commits by about 2 hours)

caiiiycuk:
> (Also, thanks @caiiiycuk for finishing this two and a half years after the PR was opened!)

Thank you. Now I can use normal emscripten to build js-dos :)

kripken:
PR up to prepare for a release: https://github.com/WebAssembly/binaryen/pull/8063

caiiiycuk:
> PR up to prepare for a release: #8063

Ahh, I forgot to ask. Is there any way to have a message when reaching "unreachable"? Currently it's not clear why we have an error in runtime, because error message is no sens (unreachable executed).

kripken:
@caiiiycuk The best way is to call an import with a special name. E.g. the SafeHeap pass will call `env.segfault` when it segfaults (the JS import does a throw). But that requires changes on the JS side.

Otherwise the stack trace can be useful. But wasm sadly doesn't have a way to report different error types, or even do error logging...

kripken:
@hoodmane Release 125 is complete.

fs-eire:
> @hoodmane Release 125 is complete.

Do we also need emscripten to upgrade the version to use emsdk with this change?

hoodmane:
Tip of tree emscripten now has it: https://github.com/emscripten-core/emscripten/pull/25834
If you want to use a released version of emscripten, you'll have to wait for 4.0.21 which will likely be about two weeks.

fs-eire:
I try to use ASYNCIFY together with `-fwasm-exceptions` but I ran into the following error:

```
[100%] Linking CXX executable ort-wasm-simd-threaded.jsep.mjs
em++: warning: ASYNCIFY=1 is not compatible with -fwasm-exceptions. Parts of the program that mix ASYNCIFY and exceptions will not compile. [-Wemcc]
unexpected expr type
UNREACHABLE executed at /b/s/w/ir/cache/builder/emscripten-releases/binaryen/src/passes/Flatten.cpp:231!
unexpected expr typeem++: error: '/mnt/vss/_work/onnxruntime/onnxruntime/cmake/external/emsdk/upstream/bin/wasm-opt --safe-heap --asyncify --pass-arg=asyncify-asserts --pass-arg=asyncify-propagate-addlist '--pass-arg=asyncify-imports@env.invoke_*,env.__asyncjs__*,*.fd_sync,*.emscripten_promise_await,*.emscripten_idb_load,*.emscripten_idb_store,*.emscripten_idb_delete,*.emscripten_idb_exists,*.emscripten_idb_clear,*.emscripten_sleep,*.emscripten_wget_data,*.emscripten_scan_registers,*._load_secondary_module,*.emscripten_fiber_swap,*._emval_await' ort-wasm-simd-threaded.jsep.wasm -o ort-wasm-simd-threaded.jsep.wasm -g --mvp-features --enable-threads --enable-bulk-memory --enable-bulk-memory-opt --enable-call-indirect-overlong --enable-exception-handling --enable-multivalue --enable-mutable-globals --enable-nontrapping-float-to-int --enable-reference-types --enable-sign-ext --enable-simd' failed (received SIGABRT (-6))
gmake[3]: *** [CMakeFiles/onnxruntime_webassembly.dir/build.make:191: ort-wasm-simd-threaded.jsep.mjs] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:935: CMakeFiles/onnxruntime_webassembly.dir/all] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:942: CMakeFiles/onnxruntime_webassembly.dir/rule] Error 2
gmake: *** [Makefile:466: onnxruntime_webassembly] Error 2
```

@kripken could you please help to take a look? I am trying an ASYNCIFY build with `-fwasm-exception -sWASM_LEGACY_EXCEPTIONS=0` and I assume it should work. (emsdk version: 4.0.23)

caiiiycuk:
> I try to use ASYNCIFY together with `-fwasm-exceptions` but I ran into the following error:
> 
> ```
> [100%] Linking CXX executable ort-wasm-simd-threaded.jsep.mjs
> em++: warning: ASYNCIFY=1 is not compatible with -fwasm-exceptions. Parts of the program that mix ASYNCIFY and exceptions will not compile. [-Wemcc]
> unexpected expr type
> UNREACHABLE executed at /b/s/w/ir/cache/builder/emscripten-releases/binaryen/src/passes/Flatten.cpp:231!
> unexpected expr typeem++: error: '/mnt/vss/_work/onnxruntime/onnxruntime/cmake/external/emsdk/upstream/bin/wasm-opt --safe-heap --asyncify --pass-arg=asyncify-asserts --pass-arg=asyncify-propagate-addlist '--pass-arg=asyncify-imports@env.invoke_*,env.__asyncjs__*,*.fd_sync,*.emscripten_promise_await,*.emscripten_idb_load,*.emscripten_idb_store,*.emscripten_idb_delete,*.emscripten_idb_exists,*.emscripten_idb_clear,*.emscripten_sleep,*.emscripten_wget_data,*.emscripten_scan_registers,*._load_secondary_module,*.emscripten_fiber_swap,*._emval_await' ort-wasm-simd-threaded.jsep.wasm -o ort-wasm-simd-threaded.jsep.wasm -g --mvp-features --enable-threads --enable-bulk-memory --enable-bulk-memory-opt --enable-call-indirect-overlong --enable-exception-handling --enable-multivalue --enable-mutable-globals --enable-nontrapping-float-to-int --enable-reference-types --enable-sign-ext --enable-simd' failed (received SIGABRT (-6))
> gmake[3]: *** [CMakeFiles/onnxruntime_webassembly.dir/build.make:191: ort-wasm-simd-threaded.jsep.mjs] Error 1
> gmake[2]: *** [CMakeFiles/Makefile2:935: CMakeFiles/onnxruntime_webassembly.dir/all] Error 2
> gmake[1]: *** [CMakeFiles/Makefile2:942: CMakeFiles/onnxruntime_webassembly.dir/rule] Error 2
> gmake: *** [Makefile:466: onnxruntime_webassembly] Error 2
> ```
> 
> @kripken could you please help to take a look? I am trying an ASYNCIFY build with `-fwasm-exception -sWASM_LEGACY_EXCEPTIONS=0` and I assume it should work. (emsdk version: 4.0.23)

Works for me with
```
-fwasm-exceptions -sASYNCIFY=1 "-sASYNCIFY_IMPORTS=[...]" -sASYNCIFY_WHITELIST=@
 ```
 
NOTE:  You also need to set `-fwasm-exceptions` for compiler.
 
 Reproduced when add `-sWASM_LEGACY_EXCEPTIONS=0`, I think you can omit it.

**Comment by fs-eire** [comment]
@caiiiycuk  thanks for your advice. I removed `-sWASM_LEGACY_EXCEPTIONS=0` and only keep `-fwasm-exception`. The build passes with the following warning:
```
em++: warning: ASYNCIFY=1 is not compatible with -fwasm-exceptions. Parts of the program that mix ASYNCIFY and exceptions will not compile. [-Wemcc]
```

I am totally confused about the flags. My understanding is that there are 2 types of EH, the legacy EH and native EH. When using legacy, WASM_LEGACY_EXCEPTIONS=1 (default value) and use `-fexception`. When using native, WASM_LEGACY_EXCEPTIONS=0 and use `-fwasm-exception`. I don't understand what it is using for `-fwasm-exception` with WASM_LEGACY_EXCEPTIONS=1.

This problem seems only happening to Asyncify. For JSPI I think `-fwasm-exception -sWASM_LEGACY_EXCEPTIONS=0` works well for my build.

fs-eire:
UPDATE: I use `wasm-objdump` to check what is included in the .wasm file. It turns out that using `-fwasm-exception -sWASM_LEGACY_EXCEPTIONS=0` generates wasm containing `try_table` and using `-fwasm-exception` generates wasm containing `try`.

So it looks that I may have some misunderstanding. The actual is: using `-fwasm-exception` does not necessarily mean to use native EH, and native EH is never supported with asyncify.

PR #5479, opened 2023-02-02
Optimize ref.as_non_null removal effect computation

kripken:
Followup to #5474 - turns out it's simple. And probably worth doing as tees are
common.


tlively (approved):
Nice!

PR #5480, opened 2023-02-03
[Strings] Add experimental string.hash instruction

kripken:
See https://github.com/WebAssembly/stringref/issues/60


tlively (approved):
Shockingly simple! Good idea to lump it in with the measuring instructions.

PR #5482, opened 2023-02-03
[C API] Add Relaxed SIMD operations

dcodeIO:
Exposes the constants

**Unary**

* BinaryenRelaxedTruncSVecF32x4ToVecI32x4
* BinaryenRelaxedTruncSVecF32x4ToVecI32x4
* BinaryenRelaxedTruncZeroSVecF64x2ToVecI32x4
* BinaryenRelaxedTruncZeroUVecF64x2ToVecI32x4

**Binary**

* BinaryenRelaxedSwizzleVecI8x16
* BinaryenRelaxedMinVecF32x4
* BinaryenRelaxedMaxVecF32x4
* BinaryenRelaxedMinVecF64x2
* BinaryenRelaxedMaxVecF64x2
* BinaryenRelaxedQ15MulrSVecI16x8
* BinaryenDotI8x16I7x16SToVecI16x8

**SIMDTernary**

* BinaryenRelaxedFmaVecF32x4
* BinaryenRelaxedFmsVecF32x4
* BinaryenRelaxedFmaVecF64x2
* BinaryenRelaxedFmsVecF64x2
* BinaryenLaneselectI8x16
* BinaryenLaneselectI16x8
* BinaryenLaneselectI32x4
* BinaryenLaneselectI64x2
* BinaryenDotI8x16I7x16AddSToVecI32x4

so the respective instructions can be produced and inspected with the C API.


tlively (approved):
LGTM, although it's possible that some of these names may change in the future to align with the final relaxed SIMD spec. For example there is some change of "fma" becoming "mul_add." Would name changes like that be ok for your use case if they were called out in the changelog?

dcodeIO:
Sure, name changes shouldn't be a problem.

Perhaps a question: On the AS side, I'm currently reassigning to the naming scheme from [relaxed-simd/Overview](https://github.com/WebAssembly/relaxed-simd/blob/main/proposals/relaxed-simd/Overview.md#binary-format). Do you know if this scheme is somewhat final? Just wondering because it states `madd` (vs `mul_add`).

tlively:
Yes, that's the most up-to-date list and its `madd` is more authoritative than my comment about `mul_add` :) However the CG voted last week to give that instruction fused semantics, so its name might go back to `fma` 🤷‍♂️ We expect the proposal to finally go to phase 4 at the next meeting, so we should have final answers by then.

PR #5484, opened 2023-02-07
[Wasm GC] Optimize casts to bottom types

kripken:
A cast to a non-nullable null (an impossible type) must trap.

In traps-never-happen mode, a cast that either returns a null or traps will
definitely return a null.

Followup to #5461 which emits casts to bottom types.


src/ir/gc-type-utils.h
```diff
@@ -70,6 +70,19 @@ inline EvaluationResult evaluateCastCheck(Type refType, Type castType) {
...
+    if (refType.isNullable() && castType.isNullable()) {
+      return SuccessOnlyIfNull;
+    } else {
+      return Failure;
+    }
+  }
+
   auto refIsHeapSubType = HeapType::isSubType(refHeapType, castHeapType);
   auto castIsHeapSubType = HeapType::isSubType(castHeapType, refHeapType);
   bool heapTypesCompatible = refIsHeapSubType || castIsHeapSubType;
```
tlively, line 88: This smaller diff would accomplish the same thing.
```suggestion
  // The dynamic heap type of the reference might be a subtype of the cast type only if the types are related and the cast type is instantiable.
  bool heapTypesCompatible = (refIsHeapSubType || castIsHeapSubType) && !castHeapType.isBottom();
```

test/lit/passes/optimize-instructions-gc-tnh.wast
```diff
@@ -607,6 +607,99 @@
...
+      )
+    )
+    (drop
+      (ref.cast none
+        (local.get $nullable-ref)
+      )
+    )
+    ;; Nullable casts might succeed, but can only return a null. In TNH mode we
+    ;; we dismiss the possibility of a trap and just return a null. Or, if the
+    ;; input is non-nullable, we will trap in either mode.
```
tlively, line 690: It would be clearer to comment on the two casts below separately. The first is more similar to the two casts above than to the last cast.

tlively (approved):

src/ir/gc-type-utils.h
kripken: What does "instantiable" mean? A null is a real value, so I guess you mean "not created by struct.new" or such?

But even so I'm not sure how to read the English here. Why can't the cast type be a bottom type for the types to be compatible? A null input value would be compatible with a null cast type, wouldn't it?

kripken: I do see how the logic is repeated, though... but I'm not sure how to explain a merge of the code...

kripken: I pushed a commit that unifies the code, but now the comment is twice as long :thinking:

tlively: It’s not the reference type but rather the bottom heap type that is not instantiable, i.e. not able to be instantiated.

“compatible” here means it’s possible for the dynamic heap type of the referenced object to be a subtype of the cast target heap. If the referenced heap type is bottom, the referenced object cannot exist so the heap types cannot be compatible.

kripken: I guess you're special-casing null in the last paragraph? In my mind a null value **is** compatible with a null cast target. Why is that wrong?

That is, no instantiated object exists (as you said) but a null is still a real value, and it has a heap type, and that heap type is compatible with the cast type's heap type. So the heap types are compatible.

tlively: FWIW I'm fine with the way you've written it as well, where special casing bottom is done separately from the computation of "compatibility."

src/ir/gc-type-utils.h
```diff
@@ -70,18 +70,25 @@ inline EvaluationResult evaluateCastCheck(Type refType, Type castType) {
...
-    // If at least one is not null, then since the heap types are not compatible
-    // we must fail.
+  if (!heapTypesCompatible || castHeapType.isBottom()) {
+    // The cases of the heap types not being compatible, or the cast type being
+    // a bottom type, have the same outcomes:
+    //  * If the heap types are not compatible, and if one is not null, then
+    //    they cannot be equal since a null is the only overlap.
+    //  * If we are casting to a null type, then both the reference must be
+    //    nullable (so a null can appear) and the cast type as well (or else the
+    //    cast type is an impossible type, a non-nullable null).
```
tlively, line 85:
```suggestion
    // If the heap types are incompatible or if it is impossible to have a non-null reference to the target heap type, then the only way the cast can succeed is if it allows nulls and the input is null.
```

tlively (commented):
I think we can simplify the comment while still making the situation clear.

PR #5487, opened 2023-02-10
[Wasm GC] Fix array.new order of operand execution

kripken:


test/lit/exec/array.wast
```diff
@@ -0,0 +1,34 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --output=fuzz-exec and should not be edited.
+
+;; RUN: wasm-opt %s -all --fuzz-exec -q -o /dev/null 2>&1 | filecheck %s
+
+(module
+ (type $array (array (mut i8)))
+
+ ;; CHECK:      [fuzz-exec] calling func
+ ;; CHECK-NEXT: [fuzz-exec] note result: func => 10
+ (func "func" (result i32)
```
tlively, line 10: I'm confused about how complicated this test is to read. Would it work to have just a single `array.new` with expressions containing different `return` instructions as its two arguments?

kripken: Ah, good point. This testcase actually shows a crash in the interpreter (we end up with the wrong type somewhere),
```
wasm-interpreter.h:207: wasm::Flow wasm::ExpressionRunner<SubType>::visit(wasm::Expression*) [with SubType = wasm::ModuleRunner]: Assertion `Type::isSubType(type, curr->type)' failed.
```
(we have a non-nullable type but a null is present). But the underlying cause is what is fixed in this PR, and you're right that that can be tested directly. Changed to that.

tlively (approved):

PR #5491, opened 2023-02-14
[Strings] Initial string execution support

kripken:
* Store string data as GC data. Inefficient (one Const per char), but ok for now.
* Implement `string.new_wtf16` and `string.const`, enough for basic testing.
* Create strings in `makeConstantExpression`, which enables ctor-eval support.
* Print strings in fuzz-exec which makes testing easier.


src/wasm/literal.cpp
```diff
@@ -71,8 +71,9 @@ Literal::Literal(const uint8_t init[16]) : type(Type::v128) {
 
 Literal::Literal(std::shared_ptr<GCData> gcData, HeapType type)
   : gcData(gcData), type(type, NonNullable) {
-  // The type must be a proper type for GC data.
-  assert((isData() && gcData) || (type.isBottom() && !gcData));
+  // The type must be a proper type for GC data: either a struct, array, or
+  // string; or a null.
+  assert(((isData() || isString()) && gcData) || (type.isBottom() && !gcData));
```
tlively, line 76: Hmm, we might want to consider including strings in `isData` since there's no obvious reason to exclude them. That might break some assumptions, though, so I wouldn't want to do it in this PR. Another option would be to remove `isData` entirely now that there is no `data` type.

src/wasm/wasm-type.cpp
```diff
@@ -897,7 +897,8 @@ bool Type::isFunction() const {
 
 bool Type::isData() const {
   if (isBasic()) {
-    return false;
+    // The only basic type that is considered data is a string.
+    return isString();
```
tlively, line 901: Oh, if we already consider strings to be data, then there's some code above that we can simplify.

src/wasm/wasm-type.cpp
```diff
@@ -1297,6 +1300,8 @@ bool HeapType::isArray() const {
   }
 }
 
+bool HeapType::isString() const { return *this == HeapType::string; }
```
tlively, line 1303: This is the only predicate we have that is expressible as an equality comparison. Should we just use an equality comparison everywhere instead? We don't have `isNone()`, for example.

test/ctor-eval/strings.wast
```diff
@@ -0,0 +1,16 @@
+(module
```
tlively, line 1: I don't suppose you would want to start a `test/lit/ctor-eval` directory and make this a lit test, would you?

src/wasm/wasm-type.cpp
kripken: Which code do you mean?

src/wasm/wasm-type.cpp
kripken: I think it's clearer with `isString()`, but yeah, maybe it's excessive. I lean towards `isString()` for now as the proposal is still experimental, and it's easier to change (e.g. maybe we'll want to consider string iterators as strings as well..? I'm not sure).

test/ctor-eval/strings.wast
kripken: I think we discussed this last time as well, and there was some technical issue iirc? Each of these tests is spread over 3 files now, that might be it.

Anyhow, if you're not happy with adding more here, I can make the PR do a little more in that area, enough for Precompute to be used for testing instead. I pushed that now.

src/wasm/wasm-type.cpp
tlively: The code I commented on in literal.cpp.

tlively (approved):
Nice, the precompute lit test looks good. LGTM % the question of whether we can simplify the code in literal.cpp.

src/wasm/wasm-type.cpp
tlively: Yeah, fair enough that we might end up with multiple string types.

src/wasm/literal.cpp
kripken: I agree we probably shouldn't do anything in this PR, as you said. Makes sense to consider both of the options you suggested later.

tlively: But it looks like we already are including strings in `Type::isData()` down in wasm-type.cpp, so the current state seems inconsistent. Am I missing something?

kripken: Ah, now I see what you mean, sorry for missing it before... I didn't put your comments together in my head.

Yeah, good point, this is inconsistent. Maybe the simplest thing is to go all-in with strings as data. That has a simple logic: they are backed by heap-allocated data, like structs and arrays.

kripken: Done.

src/literal.h
```diff
@@ -92,6 +92,8 @@ class Literal {
   bool isConcrete() const { return type.isConcrete(); }
   bool isNone() const { return type == Type::none; }
   bool isFunction() const { return type.isFunction(); }
+  // Whether this is GC data, that is, something stored on the heap (aside for a
+  // null). This includes structs, arrays, and also strings.
```
tlively, line 96:
```suggestion
  // Whether this is GC data, that is, something stored on the heap (aside from a
  // null or i31). This includes structs, arrays, and also strings.
```

tlively (approved):

kripken:
One place had to be fixed, as `isData()` there meant 'struct or array' but *not* string. Specifically string is not a subtype of eq, but the others are.

tlively:
Yeah, I really think we should get rid of `isData()` soon...

PR #5493, opened 2023-02-15
Never flip a call from unreachable to reachable during inlining

kripken:
See example in the new comment. In general, we never want to go from
unreachable to reachable (refinalize doesn't even try), as it misses out on
DCE opportunities. Also it may have validation issues, which is how the
fuzzer found this.

Remove code in the same place that was redundant after the refinalize
was added in #5492. That simplifies some existing testcases slightly, by
removing an unneeded `br` we added, and now we add a new `unreachable`
at the end in other cases that need it.

### DIFF
--- a/src/passes/Inlining.cpp
+++ b/src/passes/Inlining.cpp
@@ -404,14 +404,32 @@ static Expression* doInlining(Module* module,
   updater.walk(contents);
   block->list.push_back(contents);
   block->type = retType;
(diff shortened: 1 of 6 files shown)


tlively (approved):

kripken:
The fuzzer keeps finding stuff I've missed, sorry... hopefully now it is good.

tlively (approved):

PR #5500, opened 2023-02-16
[Strings] Add support for strings in getLiteral and Literal()

kripken:
This is enough for DAE and other opts to run on string consts.


src/wasm/literal.cpp
```diff
@@ -76,6 +76,16 @@ Literal::Literal(std::shared_ptr<GCData> gcData, HeapType type)
   assert((isData() && gcData) || (type.isBottom() && !gcData));
 }
 
+Literal::Literal(std::string string)
+  : gcData(nullptr), type(Type(HeapType::string, NonNullable)) {
+  // TODO: we could in theory internalize strings
```
tlively, line 81: Should we use `IString` instead of `std::string` so we get that right away?

kripken: I'd rather leave optimization for later, since I'm not sure what will be the more common use case in the code. Also to really internalize we'd need to replace the use of GCData in Literal with an IString, which would not be simple.

Issue #5503, opened 2023-02-17
Incorrect debug location for call instruction with sourcemap


JesseCodeBones:
Currently I found the call instruction debug location is always incorrect even source map format is correct.
The call instruction always reuse the debug location of its latest parameter instruction.
this is the screen shot:
![image](https://user-images.githubusercontent.com/56120624/219599331-bb2f52f5-a4a6-4bab-a83d-8918299cb7d4.png)

I found the root cause is when binaryen write source map it take the sequence:
call -> param1 -> param2 (asume we have two parameter).
but when we read from the wasm and source map, it take the sequence:
param1 -> param2 -> call.
with the logic of functionality `WasmBinaryBuilder::readNextDebugLocation()`
* it firstly get the correct positon info of call instruction, but it drops the information because the pos of param1 is pointed to same offset location of wasm binaryen file.
* when it comes to call instruction, it will reuse the debug location if its latest parameter.

***below is my decode for debug locations of call of wasm
```
.....call location......
SAIT:
9
0
4
-9
.....param1 location......
AAAI:
0 <---the offset of wasm is 0, so it will drop the call location information
0
0
4
.....param2 location......
EAAE:
2
0
0
2
...........
```

JesseCodeBones:
Proposal:  
For sourcemap generation, it should loop the parameters firstly, after that write the debug location of call instruction, is it correct?

kripken:
Thanks for the report and investigation!

I'm not sure I understand the problem yet. Maybe I am not familiar enough with the sourcemap code in `wasm-binary.cpp`. Can you perhaps create a small standalone testcase that shows the issue? (like maybe a simple wat file that when roundtripped the source map locations change in a wrong way) Or, if that's simpler, can you write out the code you mean in the last comment?

JesseCodeBones:
> Thanks for the report and investigation!
> 
> I'm not sure I understand the problem yet. Maybe I am not familiar enough with the sourcemap code in `wasm-binary.cpp`. Can you perhaps create a small standalone testcase that shows the issue? (like maybe a simple wat file that when roundtripped the source map locations change in a wrong way) Or, if that's simpler, can you write out the code you mean in the last comment?

No problem, I will let you knoe once I finished it😁

JesseCodeBones:
> Thanks for the report and investigation!
> 
[...]

PR #5504, opened 2023-02-20
fix: sourcemap generation for call instruction

JesseCodeBones:
fix: [5503](https://github.com/WebAssembly/binaryen/issues/5503)

### DIFF
--- a/src/wasm-binary.h
+++ b/src/wasm-binary.h
@@ -1430,7 +1430,7 @@ class WasmBinaryBuilder {
   MixedArena& allocator;
   const std::vector<char>& input;
   std::istream* sourceMap;
-  std::pair<uint32_t, Function::DebugLocation> nextDebugLocation;
+  std::tuple<uint32_t, uint32_t, Function::DebugLocation> nextDebugLocation; // available pos, previous pos, next debug location
   bool debugInfo = true;
   bool DWARF = false;
   bool skipFunctionBodies = false;
--- a/src/wasm-binary.h
+++ b/src/wasm-binary.h
@@ -1430,7 +1430,12 @@ class WasmBinaryBuilder {
   MixedArena& allocator;
   const std::vector<char>& input;
   std::istream* sourceMap;
-  std::tuple<uint32_t, uint32_t, Function::DebugLocation> nextDebugLocation; // available pos, previous pos, next debug location
+  struct NextDebugLocation {
+    uint32_t availablePos;
+    uint32_t previousPos;
+    Function::DebugLocation next;
+  };
+  NextDebugLocation nextDebugLocation;
   bool debugInfo = true;
   bool DWARF = false;
   bool skipFunctionBodies = false;
--- a/src/wasm-stack.h
+++ b/src/wasm-stack.h
@@ -232,7 +232,6 @@ void BinaryenIRWriter<SubType>::visitPossibleBlockContents(Expression* curr) {
 
 template<typename SubType>
 void BinaryenIRWriter<SubType>::visit(Expression* curr) {
-  emitDebugLocation(curr);
   // We emit unreachable instructions that create unreachability, but not
   // unreachable instructions that just inherit unreachability from their
   // children, since the latter won't be reached. This (together with logic in
@@ -257,6 +256,7 @@ void BinaryenIRWriter<SubType>::visit(Expression* curr) {
   if (Properties::isControlFlowStructure(curr)) {
     Visitor<BinaryenIRWriter>::visit(curr);
   } else {
+    emitDebugLocation(curr);
     emit(curr);
   }
 }
--- a/src/wasm/wasm-binary.cpp
+++ b/src/wasm/wasm-binary.cpp
@@ -1572,8 +1572,8 @@ void WasmBinaryWriter::writeField(const Field& field) {
 WasmBinaryBuilder::WasmBinaryBuilder(Module& wasm,
                                      FeatureSet features,
                                      const std::vector<char>& input)
-  : wasm(wasm), allocator(wasm.allocator), input(input), sourceMap(nullptr),
-    nextDebugLocation(0, 0, {0, 0, 0}), debugLocation() {
+  : wasm(wasm), allocator(wasm.allocator), input(input),
+    sourceMap(nullptr), nextDebugLocation{0, 0, {0, 0, 0}}, debugLocation() {
   wasm.features = features;
 }
 
@@ -2741,7 +2741,7 @@ void WasmBinaryBuilder::readSourceMapHeader() {
 
   mustReadChar('\"');
   if (maybeReadChar('\"')) { // empty mappings
-    std::get<0>(nextDebugLocation) = 0;
+    nextDebugLocation.availablePos = 0;
     return;
   }
   // read first debug location
@@ -2750,7 +2750,8 @@ void WasmBinaryBuilder::readSourceMapHeader() {
   uint32_t lineNumber =
     readBase64VLQ(*sourceMap) + 1; // adjust zero-based line number
   uint32_t columnNumber = readBase64VLQ(*sourceMap);
-  nextDebugLocation = {position, position, {fileIndex, lineNumber, columnNumber}};
+  nextDebugLocation = {
+    position, position, {fileIndex, lineNumber, columnNumber}};
 }
 
 void WasmBinaryBuilder::readNextDebugLocation() {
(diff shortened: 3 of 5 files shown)


src/wasm-stack.h
```diff
@@ -232,7 +232,9 @@ void BinaryenIRWriter<SubType>::visitPossibleBlockContents(Expression* curr) {
 
 template<typename SubType>
 void BinaryenIRWriter<SubType>::visit(Expression* curr) {
-  emitDebugLocation(curr);
+  if (!curr->is<Call>()) {
+    emitDebugLocation(curr);
+  }
```
kripken, line 237: Please add a comment explaining why `Call` is special here. That's not obvious to me, sorry if I didn't understand something you already explained in the issue.

In particular, why is `Call` different from, say, `Binary`? And what about `CallIndirect` and `CallRef`?

kripken (commented):
I took at look at the [repo](https://github.com/JesseCodeBones/binaryen_sourcemap_fix) you provided (thanks!), but the testcase is large - looks like it uses all of Binaryen? Instead can you make a small testcase that we can also add to the test suite here?

I made a quick attempt at a lit test now:

```lit
;; RUN: wasm-opt %s -o %t.wasm -osm %t.map -q
;; RUN: wasm-opt %t.wasm -ism %t.map -q -o - -S | filecheck %s

(module
  (func $foo (param $x i32) (param $y i32)
    ;;@ src.cpp:10:1
    (if
      ;;@ src.cpp:20:1
      (i32.add
        ;;@ src.cpp:30:1
        (local.get $x)
        ;;@ src.cpp:40:1
        (local.get $y)
      )
      ;;@ src.cpp:50:1
      (return)
    )
    ;;@ src.cpp:60:1
    (call $foo
      ;;@ src.cpp:70:1
      (local.get $x)
      ;;@ src.cpp:80:1
      (local.get $y)
    )
  )
)

;; CHECK: (module
;; CHECK-NEXT:  (func $0 (param $0 i32) (param $1 i32)
;; CHECK-NEXT: 
;; CHECK-NEXT:   ;;@ src.cpp:40:1
;; CHECK-NEXT:   (if
;; CHECK-NEXT:    (i32.add
;; CHECK-NEXT:     ;;@ src.cpp:30:1
;; CHECK-NEXT:     (local.get $0)
;; CHECK-NEXT:     ;;@ src.cpp:40:1
;; CHECK-NEXT:     (local.get $1)
;; CHECK-NEXT:    )
;; CHECK-NEXT:    ;;@ src.cpp:50:1
;; CHECK-NEXT:    (return)
;; CHECK-NEXT:   )
;; CHECK-NEXT:   ;;@ src.cpp:80:1
;; CHECK-NEXT:   (call $0
;; CHECK-NEXT:    ;;@ src.cpp:70:1
;; CHECK-NEXT:    (local.get $0)
;; CHECK-NEXT:    ;;@ src.cpp:80:1
;; CHECK-NEXT:    (local.get $1)
;; CHECK-NEXT:   )
```

Something does indeed look wrong with the call. But also with the add... so I'm not sure what's going on.

JesseCodeBones:
> 

yeah, it seems affact all the misordered instructions, like `if` `call` and etc.
the misorder means, it will be first looped by IR tree walker (source map generation), but latest with wasm binary parser (sourcemap reader).
for example:
### tree IR:
```
(call add
  i32.const 1
  i32.const 2
)
```

### binaryen parser:

```
i32.const 1
i32.const 2
call add
```
it is better to combine the debug information generation with binary generation, ie `emitDebugLocation` and `emit`,I will have a test, what do you think?

JesseCodeBones:
@kripken 
Hi Alon,
for the fixed test, I found the position for the fixed file is also incorrect, I changed it.
here is the detail:
### source map decode
```
moBAEA:
643
0
2
0
...........
4BAKA:
28
0
5
0
...........
QAJA:
8
0
-4
0
...........
OADA:
7
0
-1
0
...........
OAAA:
7
0
0
0
...........
uCAKA:
39
0
5
0
...........
```
### wasm decode
```
...
 0002d5: 21 01                      |         local.set 1
 0002d7: 0b                         |       end
 0002d8: 0c 01                      |       br 1
 0002da: 0b                         |     end
 0002db: 0b                         |   end
 0002dc: 20 04                      |   local.get 4
 0002de: 0f                         |   return
 0002df: 00                         |   unreachable
 0002e0: 00                         |   unreachable
 0002e1: 0b                         | end
 0002e2: 00                         | unreachable
 0002e3: 0b                         | end
```

according to source map position, the latest postion of wasm should be:  
643+28+8+7+7+39 = 0x2DC  
so it should be attached to position:  
` 0002dc: 20 04                      |   local.get 4`

src/wasm-binary.h
```diff
@@ -1430,7 +1430,7 @@ class WasmBinaryBuilder {
   MixedArena& allocator;
   const std::vector<char>& input;
   std::istream* sourceMap;
-  std::pair<uint32_t, Function::DebugLocation> nextDebugLocation;
+  std::tuple<uint32_t, uint32_t, Function::DebugLocation> nextDebugLocation; // available pos, previous pos, next debug location
```
kripken, line 1433: I think it would be good to use a struct here. Something like
```cpp
struct NextDebugLocation {
  uint32_t available;
  uint32_t previous;
  uint32_t next;
};
```

kripken (commented):
Nice! This looks much better. I also see the results on the `lit` testcase from my comment look better with this patch applied (still not perfect, but that can be left for later).

I think it would be good to get that lit test into the test suite. Here is an updated version with your patch:
```lit
;; RUN: wasm-opt %s -o %t.wasm -osm %t.map -g -q
;; RUN: wasm-opt %t.wasm -ism %t.map -q -o - -S | filecheck %s

(module
  (func $foo (param $x i32) (param $y i32)
    ;;@ src.cpp:10:1
    (if
      ;;@ src.cpp:20:1
      (i32.add
        ;;@ src.cpp:30:1
        (local.get $x)
        ;;@ src.cpp:40:1
        (local.get $y)
      )
      ;;@ src.cpp:50:1
      (return)
    )
    ;;@ src.cpp:60:1
    (call $foo
      ;;@ src.cpp:70:1
      (local.get $x)
      ;;@ src.cpp:80:1
      (local.get $y)
    )
  )
)

;; CHECK:  (func $foo (param $x i32) (param $y i32)
;; CHECK-NEXT:   ;;@ src.cpp:20:1
;; CHECK-NEXT:   (if
;; CHECK-NEXT:    (i32.add
;; CHECK-NEXT:     ;;@ src.cpp:30:1
;; CHECK-NEXT:     (local.get $x)
;; CHECK-NEXT:     ;;@ src.cpp:40:1
;; CHECK-NEXT:     (local.get $y)
;; CHECK-NEXT:    )
;; CHECK-NEXT:    ;;@ src.cpp:50:1
;; CHECK-NEXT:    (return)
;; CHECK-NEXT:   )
;; CHECK-NEXT:   ;;@ src.cpp:60:1
;; CHECK-NEXT:   (call $foo
;; CHECK-NEXT:    ;;@ src.cpp:70:1
;; CHECK-NEXT:    (local.get $x)
;; CHECK-NEXT:    ;;@ src.cpp:80:1
;; CHECK-NEXT:    (local.get $y)
;; CHECK-NEXT:   )
```
Please add it in `tests/lit/source-map.wast`.

kripken:
I verified that this patch passes the emscripten source map tests as well.

src/wasm-binary.h
JesseCodeBones: Hi Alon,
I updated the fix according to your comments, and for the next step (add `lit` test to main test suite), I donot know how to begin the work, can you give some advice? thank you.

kripken: For the test, just create a new file at `tests/lit/source-map.wast`, paste the test contents from [here](https://github.com/WebAssembly/binaryen/pull/5504#pullrequestreview-1309749525) into that file, and add it using `git add`, then commit it. It should just work.

JesseCodeBones: The lit test just passed as well:
```
/binaryen$ python build/bin/binaryen-lit -vv test/lit/source-map.wast 
-- Testing: 1 tests, 1 workers --
PASS: Binaryen lit tests :: source-map.wast (1 of 1)

Testing Time: 0.06s
  Passed: 1
```

JesseCodeBones: Hi Alon,
the test has been added, please kindly review, thank you.

kripken:
Thanks! I'm still reading the code, but to fix the CI error please apply the style diff from here:

https://github.com/WebAssembly/binaryen/actions/runs/4249525050/jobs/7400946814

kripken (approved):
lgtm aside from the code style that needs to be fixed.

JesseCodeBones:
> lgtm aside from the code style that needs to be fixed.

fixed it now:)

kripken (approved):
Great, thanks!

PR #5507, opened 2023-02-21
[Strings] Add hashing and equality support for strings

kripken:
This is enough to test RSE.

### DIFF
--- a/src/wasm/literal.cpp
+++ b/src/wasm/literal.cpp
@@ -432,6 +432,15 @@ bool Literal::operator==(const Literal& other) const {
       assert(func.is() && other.func.is());
       return func == other.func;
     }
+    if (type.isString()) {
+      if (!gcData && !other.gcData) {
+        return true;
+      }
+      if (!gcData || !other.gcData) {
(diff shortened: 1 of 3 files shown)


src/wasm/literal.cpp
```diff
@@ -432,6 +432,15 @@ bool Literal::operator==(const Literal& other) const {
...
     }
+    if (type.isString()) {
+      if (!gcData && !other.gcData) {
+        return true;
+      }
+      if (!gcData || !other.gcData) {
+        return false;
+      }
+      return gcData->values == other.gcData->values;
+    }
```
tlively, line 443: Null `gcData` is only allowed when the type is a reference to bottom, so we don't need to handle it here.

```suggestion
    if (type.isString()) {
      return gcData->values == other.gcData->values;
    }
```

tlively (approved):
LGTM % comment.

src/wasm/literal.cpp
kripken: Good catch!

PR #5509, opened 2023-02-22
feat(asyncify): allow importing asyncify.get_state from the module

lbguilherme:
It is currently not possible to read the asyncify state from inside the module. It should be possible to "import" asyncify.get_state for that.


kripken (commented):
Makes sense to add this. Please just add a test for it. Looks like the best place might be `test/lit/passes/asyncify_enable-multivalue.wast`.

PR #5512, opened 2023-02-22
[wasm-ctor-eval] Add v128 load/store support

kripken:


scripts/update_lit_checks.py
```diff
@@ -40,6 +40,11 @@
 ALL_ITEMS = '|'.join(['type', 'import', 'global', 'memory', 'data', 'table',
                       'elem', 'tag', 'export', 'start', 'func'])
 ITEM_NAME = r'\$?[^\s()]*|"[^\s()]*"'
+# FIXME: This does not handle nested string contents. For example,
+#  (data (i32.const 10) "hello(")
+# will look unterminated, due to the '(' inside the string. As a result, the
+# code below will consider more elements after the |data| to be part of it,
+# until it sees enough closing ')' symbols.
```
tlively, line 47: I wonder if it would be worth escaping parentheses in strings to work around this.

Something I've idly thought about is whether it would be possible to rewrite this script in C++ using the new wat parser rather than leaning so heavily on regexes. That would avoid problems like this.

tlively (approved):

PR #5513, opened 2023-02-22
[Fuzzer] Simplify the hang limit mechanism

kripken:
Previously the idea was that we started with `HANG_LIMIT = 10` or so, and we'd decrement
it by one in each potentially-recursive call and loop entry. When we reached 0 we'd start
to unwind the stack. Then, after we unwound it all the way, we'd reset `HANG_LIMIT` before
calling the next export.

That approach adds complexity that each "execution wrapper", like for JS or for `--fuzz-exec`,
had to manually reset `HANG_LIMIT`. That was done by calling an export. Calls to those
exports had to appear in various places, which is sort of a hack.

The new approach here does the following when the hang limit reaches zero: It resets
`HANG_LIMIT`, and it traps. The trap unwinds the call stack all the way out. When the next
export is called, it will have a fresh hang limit since we reset it before the trap.

This does have downsides. Before, we did not always trap when we hit the hang limit but
rather we'd emit something unreachable, like a return. The idea was that we'd leave the
current function scope at least, so we don't hang forever. That let us still execute a small
amount of code "on the way out" as we unwind the stack. I'm not sure it's worth the
complexity for that.

The advantages of this PR are to simplify the code, and also it makes more fuzzing
approaches easy to implement. I'd like to add a `wasm-ctor-eval` fuzzer, and having to add
hacks to call the hang limit init export in it would be tricky. With this PR, the execution
model is simple in the fuzzer: The exports are called one by one, in order, and that's it -
no extra magic execution needs to be done.

Also bump the hang limit from 10 to 100, just to give some more chance for code to run.


aheejin:
I like the simplicity of this idea! But this way if a function decrements `HANG_LIMIT_GLOBAL` by, say, 90, and then we run the next function, `HANG_LIMIT_GLOBAL` wouldn't have been reset to 100, and that function only has 10 tries before trapping. Also this seems to make each function execution's result to be dependent of the execution order. Is that fine?

tlively (approved):
Nice simplification!

kripken:
Good idea! We already have a wrapping mechanism, the invoker, so adding a random reset there some of the time should improve things. I added that now.

I also made the smaller of the two `-ttf` tests print the output. We removed it because it was verbose, but it's useful to see changes like this. In particular you can see a reset of the limit here:

https://github.com/WebAssembly/binaryen/pull/5513/files#diff-8a7f00ac3f0114900a98b0ae0f4475ce22f9c2a413202ceae7c2f57699b87eabR428-R430

kripken:
Oh, no, this doesn't work. Adding resets of the hang counter in userspace can lead to a hang. The fuzzer found this:
```js
function invoke() {
  limit = 100;
}
function other() {
  while (1) {
    if (!limit) trap
    limit--;
    invoke(); // resets the limit!
  }
}
```
Later functions can have random calls to earlier functions, so this can happen...

tlively:
It would work if the module could never call the exported wrappers directly, right?

kripken:
Yes, but that would require more work. I think it's not worth it as the simple approach is fairly good too.

kripken:
I'm not worried about debugging - the hang limit global is just another global at that point when you look at the wasm. I've never had a case where I wanted to manually adjust the limit when debugging a testcase at least.

aheejin (approved):

PR #5515, opened 2023-02-23
[Wasm GC] Parse (sub $super _) for array, func, and struct

wingo:
The pretty-printer will still serialize these using the old func_subtype, array_subtype, and struct_subtype syntax, though.

### DIFF
--- a/src/wasm/wasm-s-parser.cpp
+++ b/src/wasm/wasm-s-parser.cpp
@@ -52,8 +52,7 @@ namespace wasm {
 static Name STRUCT("struct"), FIELD("field"), ARRAY("array"),
   FUNC_SUBTYPE("func_subtype"), STRUCT_SUBTYPE("struct_subtype"),
   ARRAY_SUBTYPE("array_subtype"), EXTENDS("extends"), REC("rec"), I8("i8"),
-  I16("i16"), DECLARE("declare"), ITEM("item"), OFFSET("offset"),
-  SUB("sub");
+  I16("i16"), DECLARE("declare"), ITEM("item"), OFFSET("offset"), SUB("sub");
 
 static Address getAddress(const Element* s) {
   return std::stoll(s->toString());
@@ -905,12 +904,12 @@ void SExpressionWasmBuilder::preParseHeapTypes(Element& module) {
         throw ParseException("invalid 'sub' form", kind.line, kind.col);
       }
       super = def[1];
-      Element &subtype = *def[2];
+      Element& subtype = *def[2];
       if (!subtype.isList() || subtype.size() < 1) {
-        throw ParseException("invalid subtype definition", subtype.line,
-                             subtype.col);
+        throw ParseException(
+          "invalid subtype definition", subtype.line, subtype.col);
       }
-      Element &subtypeKind = *subtype[0];
+      Element& subtypeKind = *subtype[0];
       if (subtypeKind == FUNC) {
         builder[index] = parseSignatureDef(subtype, 0);
       } else if (subtypeKind == STRUCT) {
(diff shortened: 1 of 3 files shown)


test/lit/wat-kitchen-sink.wast
```diff
@@ -121,6 +121,9 @@
  ;; CHECK:      (type $s8 (struct (field i32) (field i64) (field $z f32) (field f64) (field (mut i32))))
  (type $s8 (struct i32 (field) i64 (field $z f32) (field f64 (mut i32))))
 
+ ;; CHECK:      (type $s9 (struct_subtype (field i32) $s2))
+ (type $s9) (sub $s2 (struct i64))
```
kripken, line 125:
```suggestion
 (type $s9 (sub $s2 (struct i64)))
```
I think?

src/wasm/wasm-s-parser.cpp
```diff
@@ -898,30 +899,62 @@ void SExpressionWasmBuilder::preParseHeapTypes(Element& module) {
...
-      } else if (kind == FUNC_SUBTYPE && super->str() == FUNC) {
-        // OK; no supertype
-        super = nullptr;
-      } else if ((kind == STRUCT_SUBTYPE || kind == ARRAY_SUBTYPE) &&
-                 super->str() == DATA) {
-        // OK; no supertype
-        super = nullptr;
+    if (kind == SUB) {
+      if (def.size() != 3) {
+        throw ParseException("invalid `sub' form", kind.line, kind.col);
```
tlively, line 905: I don't think we have precedent for using backticks in messages.
```suggestion
        throw ParseException("invalid 'sub' form", kind.line, kind.col);
```

test/lit/wat-kitchen-sink.wast
tlively: I would not expect wat-kitchen-sink.wast to change at all, since it tests the new parser but you modified the old parser. The new parser should already parse the standards-track format.

wingo:
Thanks for the feedback, and apologies for the month delay!  I changed the error message to have straight ticks and reverted the test change.

tlively:
Thanks! I do think it would be good to have a test for this, just not in that particular file. How about porting test/lit/isorecursive-good.wast to the new syntax?

wingo:
ran clang-format, updated isorecursive-good.test though i have not been able to run it yet :/

tlively:
Is there a problem trying to run the test that we can help with? `./check.py lit` or `bin/binaryen-lit test/lit/isorecursive-good.wast` should work after you've installed filecheck and lit via `pip3 install -r requirements-dev.txt`.

wingo:
> Is there a problem trying to run the test that we can help with? `./check.py lit` or `bin/binaryen-lit test/lit/isorecursive-good.wast` should work after you've installed filecheck and lit via `pip3 install -r requirements-dev.txt`.

Thank you for this note, not sure how i missed this incantation in the `README.md`.  The test passes for me though I see that apparently CI is not working.  (Could you flip the bit for me that makes CI run automatically?  Otherwise feedback/refresh cycles are longer.)

wingo:
Followup patch fixes bug subtyping `$func` which would be seen as subtyping `func`; same with `$data`.

src/wasm/wasm-s-parser.cpp
```diff
@@ -898,30 +898,62 @@ void SExpressionWasmBuilder::preParseHeapTypes(Element& module) {
...
-        super = nullptr;
-      } else if ((kind == STRUCT_SUBTYPE || kind == ARRAY_SUBTYPE) &&
-                 super->str() == DATA) {
-        // OK; no supertype
-        super = nullptr;
+    if (kind == SUB) {
+      if (def.size() != 3) {
+        throw ParseException("invalid 'sub' form", kind.line, kind.col);
+      }
+      super = def[1];
```
tlively, line 906: The standard format also allows `sub` definitions with zero specified supertypes, but we don't need to support that in this patch (or even in this parser, probably).

tlively (approved):

wingo:
thanks for your patience @tlively :)

PR #5517, opened 2023-02-23
Fix validation of DataDrop

### LINKED ISSUE #5511: Segfault in MemoryPacking
```wat
(module
 (memory $0 (shared 16 17))
 (func $1
  (data.drop 1)
  (unreachable)
 )
)
```
```
$ bin/wasm-opt a.wat -all --memory-packing
Segmentation fault
```
Looks like it crashes in `MemoryPacking::optimizeSegmentOps`.

cc @tlively

kripken:
`shouldBeFalse` was misused in that place.

Fixes #5511

### DIFF
--- a/src/wasm/wasm-validator.cpp
+++ b/src/wasm/wasm-validator.cpp
@@ -138,8 +138,9 @@ struct ValidationInfo {
                      Function* func = nullptr) {
     if (result) {
       fail("unexpected true: " + std::string(text), curr, func);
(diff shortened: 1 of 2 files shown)


kripken:
I reverted a change to alter the meaning of the result. The result is used elsewhere it turns out, so just keep it as is and fix DataDrop).

tlively (approved):
Thank you!

PR #5518, opened 2023-02-23
[wasm-ctor-eval] Fix evalling of multiple ctors with GC

kripken:
The logic worked for a single ctor but with more than one we ended up
not copying the recently-added globals.

An existing test changes slightly because now the module is in a valid
state during all phases of computation, and when we pick names for
new globals we are affected by the number of existing globals (previously
the number of globals was too low in the interim, but we were lucky
enough to not hit a problem).


kripken:
Oops, this is wrong, sorry.

PR #5519, opened 2023-02-23
JSPI - Replace function table references with JSPI'ed wrapper.

brendandahl:
This makes it possible to get the JSPI'ed version of the function from the function table.
Needed for async bindings with embind.


src/passes/JSPI.cpp
```diff
@@ -140,6 +140,24 @@ struct JSPI : public Pass {
...
+          auto iter = wrappedExports.find(get->func);
+          if (iter == wrappedExports.end()) {
+            continue;
+          }
+          auto* replacementRef = builder.makeRefFunc(
+            iter->second, module->getFunction(iter->second)->type);
+          segment->data[i] = replacementRef;
+        }
+      }
+    }
```
kripken, line 159: Should this be done for all RefFuncs, also not in the table (like, say, in a GC struct or array)? If so then it would be more general to write a `Walker` and to use `visitRefFunc` there. That would be able to find all RefFuncs anywhere, including in the table.

brendandahl: I need to learn more about the GC proposal. Are those exposed outside of wasm? I think we only want to replace references that are exposed.

kripken: I guess in practice now only the table and normal exported functions are used for that. But in theory a GC array could contain references and one of them could be returned from an export.

It's probably about the same amount of code, so there's not a big benefit either way. sgtm to go with this for now.

kripken (approved):

PR #5520, opened 2023-02-24
remote unnecessary echo PATH in CI

Changqing-JING:
The runner ubuntu-latest is Ubuntu 22, which no longer use llvm-8, so this PATH can be remoted.

This llvm-8 here is misleading. Some developers spend a lot of time to compile llvm-8 from source, because llvm-8 is not available in new Ubuntu apt. But actually CI is not using llvm-8.


kripken (approved):
Thanks!

PR #5522, opened 2023-02-24
[wasm-ctor-eval] Properly handle multiple ctors with GC

kripken:
Before, a single ctor with GC worked, but any subsequent ones simply dropped
the globals from the previous ones, because we were missing an `addGlobal` in
an important place.

Also, we can get confused about which global names are in use in the module, so fix
that as well by storing them directly (we keep removing and re-adding globals, so
we can't use the normal module mechanism to find which names are in use).


kripken:
(Changes to the existing test are just that now we compute new global names slightly differently.)

tlively (approved):

PR #5524, opened 2023-02-24
Emit source map information for control flow structures

kripken:
With this, the sourcemap testcase outputs the exact same thing as the input.

Followup to #5504

cc @JesseCodeBones

### DIFF
--- a/test/lit/source-map.wast
+++ b/test/lit/source-map.wast
@@ -26,8 +26,9 @@
 )
 
 ;; CHECK:  (func $foo (param $x i32) (param $y i32)
-;; CHECK-NEXT:   ;;@ src.cpp:20:1
+;; CHECK-NEXT:   ;;@ src.cpp:10:1
 ;; CHECK-NEXT:   (if
+;; CHECK-NEXT:    ;;@ src.cpp:20:1
 ;; CHECK-NEXT:    (i32.add
 ;; CHECK-NEXT:     ;;@ src.cpp:30:1
 ;; CHECK-NEXT:     (local.get $x)
--- a/src/wasm-stack.h
+++ b/src/wasm-stack.h
@@ -251,12 +251,12 @@ void BinaryenIRWriter<SubType>::visit(Expression* curr) {
     // `curr` is not reachable, so don't emit it.
     return;
   }
+  emitDebugLocation(curr);
   // Control flow requires special handling, but most instructions can be
   // emitted directly after their children.
   if (Properties::isControlFlowStructure(curr)) {
     Visitor<BinaryenIRWriter>::visit(curr);
   } else {
-    emitDebugLocation(curr);
     emit(curr);
   }
 }


JesseCodeBones:
Hi Alon,
I had a test with follow source code (assemblyscript).
```typescript
export function main():void{
    add(
        1,2);
}
```
here is source map generated by binaryen:
```
...........<- call instruction
OAIT:
7
0
4
-9
...........<- first parameter
AACI:
0   <-----the offset is 0, that means the param1 and call instruction used the same position. which could be incorrect
0
1
4
...........
EAAE:
2
0
0
2
```
and here is the outputs of binary module printer (the call instruction debug info is missed):
```
(func $assembly/index/main
  (drop
   (call $assembly/index/add
    ;;@ assembly/index.ts:7:8
    (i32.const 1)
    ;;@ assembly/index.ts:7:10
    (i32.const 2)
   )
  )
```
I did not going to detail of the root cause, just report what I found, I will update later if I have some new found, Thank you.

JesseCodeBones:
Hi Alon,
I found if there is duplicate debug location for different instructions, the binaryen will dismiss some of them, here is the lit test, I am still finding the root cause and solution

```
;; RUN: wasm-opt %s -o %t.wasm -osm %t.map -g -q
;; RUN: wasm-opt %t.wasm -ism %t.map -q -o - -S | filecheck %s
(module
  (func $foo (param $a i32) (param $b i32) (result i32)
    (return
        (i32.add 
            (local.get $a)
            (local.get $b)
        )
    )
  )
  (func $bar (param $a i32)
    ;;@ src.cpp:10:1
    (local.set $a
     (i32.const 1)
    )
    ;;@ src.cpp:10:1
    (drop
        ;;@ src.cpp:10:1
        (call $foo
            ;;@ src.cpp:20:1
            (local.get $a)
            ;;@ src.cpp:30:1
            (i32.const 42)
        )
    )
    (return)
  )
)

;; CHECK:  (module
;; CHECK-NEXT:   (type $i32_i32_=>_i32 (func (param i32 i32) (result i32)))
;; CHECK-NEXT:    (type $i32_=>_none (func (param i32)))
;; CHECK-NEXT:    (func $foo (param $a i32) (param $b i32) (result i32)
;; CHECK-NEXT:     (return
;; CHECK-NEXT:      (i32.add
;; CHECK-NEXT:       (local.get $a)
;; CHECK-NEXT:       (local.get $b)
;; CHECK-NEXT:      )
;; CHECK-NEXT:     )
;; CHECK-NEXT:    )
;; CHECK-NEXT:    (func $bar (param $a i32)
;; CHECK-NEXT:     ;;@ src.cpp:10:1
;; CHECK-NEXT:     (local.set $a
;; CHECK-NEXT:      (i32.const 1)
;; CHECK-NEXT:     )
;; CHECK-NEXT:     (drop
;; CHECK-NEXT:      (call $foo
;; CHECK-NEXT:       ;;@ src.cpp:20:1
;; CHECK-NEXT:       (local.get $a)
;; CHECK-NEXT:       ;;@ src.cpp:30:1
;; CHECK-NEXT:       (i32.const 42)
;; CHECK-NEXT:      )
;; CHECK-NEXT:     )
;; CHECK-NEXT:     (return)
;; CHECK-NEXT:    )
;; CHECK-NEXT:   )

```

Regards

kripken:
cc @aheejin - not sure if you're familiar with this part of the debugging logic (it's source maps and not DWARF) but maybe you're interested in it too.

JesseCodeBones:
> @JesseCodeBones That does look like a real bug. Source maps support referring to the same source line from multiple places. And I agree it can happen in real-world code because sometimes a single source line ends up turned into multiple wasm lines, like you said. If you can find a fix that would be great!
> 
> I think that is separate from this PR, though? I see the same results on that testcase with this PR or not. Does this PR look ok to you to land?

yeah, I can handle it after this pr, it may take sometime and I have some personal issue to be handled, I suppose it is not an urgen or blocking issue

kripken:
Sounds good, thanks, and yes, this it not urgent of course.

test/lit/source-map.wast
```diff
@@ -26,8 +26,9 @@
 )
 
 ;; CHECK:  (func $foo (param $x i32) (param $y i32)
-;; CHECK-NEXT:   ;;@ src.cpp:20:1
```
aheejin, line 29: I'm not very familiar with this part of the code. A few questions:
- Why didn't we print the debug location for control flow structure before?
- It looks this `src.cpp:20:1` is the debug info for not `if` but `i32.add`, given that we didn't print debug info for control flow structures. Then why was it printed before `if` and not before `i32.add`?

kripken: Unfortunately I don't think anyone is very familiar with it :smile: It was written a long time ago (and not by me).

I can't see a reason for not printing it for control flow structures before. That might have just been an omission.

I'm not sure about the second question. I know there are some heuristics for how to print and emit stuff, that involve looking at recent instructions or such, so probably related to that?

Overall, my approach here considering this unfamiliar code is to get more tests on it, and improve on them. This PR seems to do that as after it we appear to perfectly roundtrip the input.

aheejin (approved):

PR #5533, opened 2023-03-01
Fuzzer: Only use RefAs in a function context

kripken:
Line 1980/1981 below is where we use a RefAs. We need to avoid getting
there if we are not in a function.


tlively (approved):
This is because `ref.as_non_null` is not a constant instruction?

PR #5539, opened 2023-03-02
Fuzzer: Pick from existing heap types in the module

kripken:
The next step after this will be to use new generated heap types there.


kripken:
(Some nondeterminism has been sorted out.)

tlively (approved):
LGTM! Does this already run into issues with uninhabitable types?

kripken:
Yes, we do have some noninhabitable types in the test suite. We could either skip those tests or filter such types out. I pulled in the code to do that now, as it seems best to just do it here (we might forget to unskip later, and I have the code already as it was planned for the next PR).

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -275,9 +275,71 @@ void TranslateToFuzzReader::setupMemory() {
...
+    // original type to be noninhatible).
+    auto maybeAdd = [&](Type type) {
+      // Non-refs are always ok. Nullable refs are ok, since we can create an
+      // instance with a null there.
+      if (!type.isRef() || type.isNullable()) {
+        return;
+      }
+
+      auto heapType = type.getHeapType();
+      if (heapType.isStruct() || heapType.isArray()) {
```
tlively, line 310: Don't forget function types!

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -275,9 +275,71 @@ void TranslateToFuzzReader::setupMemory() {
 void TranslateToFuzzReader::setupHeapTypes() {
   // Start with any existing heap types in the module, which may exist in any
   // initial content we began with.
-  interestingHeapTypes = ModuleUtils::collectHeapTypes(wasm);
+  auto possibleHeapTypes = ModuleUtils::collectHeapTypes(wasm);
 
   // TODO: use heap type fuzzer to add new types in addition to the previous
+
+  // Filter away uninhabitable heap types, that is, heap types that we cannot
+  // construct, like a type with a non-nullable reference to itself. For
```
tlively, line 283: I don't see where cycles are actually detected here.

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -275,9 +275,71 @@ void TranslateToFuzzReader::setupMemory() {
...
+  // simplicity, don't look for such type loops of aribtrary size but just limit
+  // the search to a reasonable amount (to avoid possible slowness).
+  const size_t MAX_SEARCH = 100;
+  std::unordered_set<HeapType> uninhabitable;
```
tlively, line 287: It doesn't look like this is used.

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -275,9 +275,71 @@ void TranslateToFuzzReader::setupMemory() {
...
+    }
+
+    std::vector<HeapType> seen;
+    seen.push_back(t);
+    size_t next = 0;
+    bool fail = false;
+
+    // Add a child type (a field in a struct, etc.) to the list of seen types,
+    // if it is something we need to look at (that is, if it can cause the
+    // original type to be noninhatible).
```
tlively, line 301:
```suggestion
    // original type to be uninhabitable).
```

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -275,9 +275,71 @@ void TranslateToFuzzReader::setupMemory() {
...
+      }
+
+      auto heapType = type.getHeapType();
+      if (heapType.isStruct() || heapType.isArray()) {
+        seen.push_back(heapType);
+        if (seen.size() >= MAX_SEARCH) {
+          fail = true;
+        }
+      } else if (heapType.isBottom()) {
+        // This is a non-nullable bottom type, which is noninhabitable.
```
tlively, line 316:
```suggestion
        // This is a non-nullable bottom type, which is uninhabitable.
```

tlively:
It would be good to factor out the code for identifying uninhabitable types to be shared between the type fuzzer and the main fuzzer.

src/tools/fuzzing/fuzzing.cpp
kripken: Do we hit a problem with them? The only issue is failing to create a GC object, but functions are always possible to create, of any type.

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -275,9 +275,71 @@ void TranslateToFuzzReader::setupMemory() {
...
+  // construct, like a type with a non-nullable reference to itself. For
+  // simplicity, don't look for such type loops of aribtrary size but just limit
+  // the search to a reasonable amount (to avoid possible slowness).
+  const size_t MAX_SEARCH = 100;
+  std::unordered_set<HeapType> uninhabitable;
+  for (auto t : possibleHeapTypes) {
+    if (t.isBasic() || t.isBottom()) {
+      // These types are handled directly in the random code generators.
+      continue;
+    }
```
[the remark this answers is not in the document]
tlively: Maybe make it an assertion for now so we can't forget to update it once it becomes relevant?

src/tools/fuzzing/fuzzing.cpp
tlively: Oh, good point. I guess I could update `makeInhabitable` to also assume function references are always inhabitable.

src/tools/fuzzing/fuzzing.cpp
tlively: Oh, I see, it's depending on hitting the recursion limit to detect cycles. That seems wasteful!

kripken: It's very simple though :smile: look at how little code it is. With a reasonable recursion limit it should be cheap.

src/tools/fuzzing/fuzzing.cpp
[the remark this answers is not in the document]
kripken: Done.

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -270,6 +272,77 @@ void TranslateToFuzzReader::setupMemory() {
...
+  // simplicity, don't look for such type loops of aribtrary size but just limit
+  // the search to a reasonable amount (to avoid possible slowness). That is, we
+  // start at a given type, expand out its children, and continue to expand
+  // recursively. If there is a cycle then this would continue forever, so it
+  // must hit any fixed limit.
+  const size_t MAX_SEARCH = 100;
+  for (auto t : possibleHeapTypes) {
+    // These types are handled directly in the random code generators, and we
+    // assume they are not emitted by the code before us in this function.
+    assert(!t.isBasic() && !t.isBottom());
```
kripken, line 293: This assertion you suggested actually gets hit. Specifically, I run `HeapTypeGenerator::create` and then `HeapTypeGenerator::makeInhabitable` and I get the heap type `none` in the vector.

tlively: Oh yeah, the assertion is only meant to stick around until the next PR that uses `HeapTypeGenerator`.

kripken: Oh, ok. I'll leave it here then.

kripken:
> It would be good to factor out the code for identifying uninhabitable types to be shared between the type fuzzer and the main fuzzer.

I considered that but I'm not sure I see an obvious way. Do you?

tlively:
The `findUninhabitable` function in wasm-fuzz-types.cpp is almost what we need, except that it's actually recursive and returns an optional heaptype when really it would be fine to simplify it to return a bool.

kripken:
Ok, I can take a look at that once your PR lands.

tlively:
Thanks! I just landed it.

kripken:
Ok, this now includes a refactoring of your code to let it be used from the main fuzzer.

src/tools/fuzzing/heap-types.h
```diff
@@ -44,6 +44,24 @@ struct HeapTypeGenerator {
...
+  static std::optional<HeapType>
+  findUninhabitable(HeapType type,
+                    std::unordered_set<HeapType>& visited,
+                    std::unordered_set<HeapType>& visiting);
+
+  static std::optional<HeapType>
+  findUninhabitable(HeapType parent,
+                    Type type,
+                    std::unordered_set<HeapType>& visited,
+                    std::unordered_set<HeapType>& visiting);
```
tlively, line 64: Do these need to be public or can we make them implementation details in the .cpp file?

src/tools/wasm-fuzz-types.cpp
```diff
@@ -561,7 +501,7 @@ void Fuzzer::checkInhabitable() {
   std::unordered_set<HeapType> visited, visiting;
   bool haveUninhabitable = false;
   for (auto type : types) {
-    if (findUninhabitable(type, visited, visiting)) {
+    if (HeapTypeGenerator::findUninhabitable(type, visited, visiting)) {
```
tlively, line 504: To make `findUninhabitable` an implementation detail, we can use `haveUninhabitable = HeapTypeGenerator::getUninhabitable(types).size() == types.size())`. That will let us make `findUninhabitable` return bool and simplify the implementation as well.

tlively (commented):
Alternatively, I'd be happy if we landed this and I can go in and clean things up afterward.

src/tools/wasm-fuzz-types.cpp
kripken: Hmm, there is code lower down that looks at individual types too, though? That's why atm `findUninhabitable` is not just an implementation detail. But that later code could be removed though if we think those checks + fatal errors are not needed, like this perhaps?
```diff
diff --git a/src/tools/wasm-fuzz-types.cpp b/src/tools/wasm-fuzz-types.cpp
index d10510222..1f0aa1b4a 100644
--- a/src/tools/wasm-fuzz-types.cpp
+++ b/src/tools/wasm-fuzz-types.cpp
@@ -498,27 +498,9 @@ void Fuzzer::checkInhabitable() {
   }
 
   // Check whether any of the original types are uninhabitable.
-  std::unordered_set<HeapType> visited, visiting;
-  bool haveUninhabitable = false;
-  for (auto type : types) {
-    if (HeapTypeGenerator::findUninhabitable(type, visited, visiting)) {
-      haveUninhabitable = true;
-      break;
-    }
-  }
-  visited.clear();
-  visiting.clear();
-
-  if (haveUninhabitable) {
-    // Verify that the transformed types are inhabitable.
-    for (auto type : inhabitable) {
-      if (auto uninhabitable =
-            HeapTypeGenerator::findUninhabitable(type, visited, visiting)) {
-        IndexedTypeNameGenerator print(inhabitable);
-        Fatal() << "Found uninhabitable type: " << print(*uninhabitable);
-      }
-    }
-  } else if (getTypeSystem() == TypeSystem::Isorecursive) {
+  haveUninhabitable = HeapTypeGenerator::getInhabitable(types).size() != types.size();
+
+  if (!haveUninhabitable && getTypeSystem() == TypeSystem::Isorecursive) {
     // Verify the produced inhabitable types are the same as the original types.
     if (types.size() != inhabitable.size()) {
       Fatal() << "Number of inhabitable types does not match number of "
```
I'd also be ok leaving this for after this lands as you said.

tlively: The way I would fix this would be to make that "Found uninhabitable type" error actually print all of the uninhabitable types rather than the arbitrary uninhabitable type we happen to find first. We could do `set(types) \ `set(inhabitable)` types to find the uninhabitable types.

kripken: But the set `inhabitable` would be generated by filtering out the uninhabitable ones. So we'd be using the exact same process twice to filter to inhabitable types? Given that I think we can just delete it here. But I feel like I might be missing something.

tlively: no, I'd use a normal membership test on a `std::unordered_set` of the uninhabitable types to find inhabitable types. It's ok to do some extra slow work in this failure case to print a more useful message.

kripken: Ok, I tried to write it that way, PTAL.

I suspect I have done something wrong, as this now fails your fuzzer,
```
$ ./bin/wasm-fuzz-types
Iteration 1
Running with seed 5435209648995731794
Fatal: makeInhabitable incorrectly changed type (type $0 (array i16))
```
Maybe I don't quite understand how to use your APIs properly? Could be something obvious in the last commit that I'm doing wrong that you can see.

kripken: (It seems line new line 528 might be wrong, as it should only look at uninhabitable types..? But that line is unchanged in this PR, so I'm confused.)

kripken: (But to look at inhabitable types specifically we'd need that helper function `findUninhabitable` again... which we've said we want to avoid here.)

tlively: Oh, I just realized I wrote `set(inhabitable)` above when I meant `set(uninhabitable)`. Sorry for the confusion! Taking a look at the latest code now...

kripken:
This has been rebased after your changes @tlively

tlively (approved):
LGTM!

kripken:
Thanks!

Meanwhile this has been fuzzed for many hundreds of thousands of iterations at this point, so hopefully it is stable. Landing.

PR #5541, opened 2023-03-03
Add a fuzzer utility for ensuring types are inhabitable

tlively:
Some valid GC types, such as non-nullable references to bottom heap types and
types that contain non-nullable references to themselves, are uninhabitable,
meaning it is not possible to construct values of those types. This can cause
problems for the fuzzer, which generally needs to be able to construct values of
arbitrary types.

To simplify things for the fuzzer, introduce a utility for transforming type
graphs such that all their types are inhabitable. The utility performs a DFS to
find cycles of non-nullable references and breaks those cycles by introducing
nullability.

The new utility is itself fuzzed in the type fuzzer.

### DIFF
--- a/src/tools/fuzzing/heap-types.cpp
+++ b/src/tools/fuzzing/heap-types.cpp
@@ -678,6 +678,18 @@ struct Inhabitator {
   // Uniquely identify fields as an index into a type.
   using FieldPos = std::pair<HeapType, Index>;
 
+  // When we make a reference nullable, we typically need to make the same
+  // reference in other types nullable to maintain valid subtyping. Which types
+  // we need to update depends on the variance of the reference, which is
+  // determined by how it is used in its enclosing heap type.
+  //
+  // An invariant field of a heaptype must have the same type in subtypes of
+  // that heaptype. A covariant field of a heaptype must be typed with a subtype
+  // of its original type in subtypes of the heaptype. A contravariant field of
+  // a heap type must be typed with a supertype of its original type in subtypes
+  // of the heaptype.
+  enum Variance { Invariant, Covariant, Contravariant };
+
   // The input types.
   const std::vector<HeapType>& types;
 
@@ -688,8 +700,6 @@ struct Inhabitator {
 
   Inhabitator(const std::vector<HeapType>& types)
     : types(types), subtypes(types) {}
-
-  enum Variance { Invariant, Covariant, Contravariant };
 
   Variance getVariance(FieldPos field);
   void markNullable(FieldPos field);
--- a/src/tools/fuzzing/heap-types.cpp
+++ b/src/tools/fuzzing/heap-types.cpp
@@ -16,6 +16,8 @@
 
 #include <variant>
 
+#include "ir/subtypes.h"
+#include "support/insert_ordered.h"
 #include "tools/fuzzing/heap-types.h"
 #include "tools/fuzzing/parameters.h"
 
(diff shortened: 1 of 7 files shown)


tlively:
* main
  * **PR #5541** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5541" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5541?utm_source=stack-comment).

src/tools/fuzzing/heap-types.cpp
```diff
@@ -654,4 +656,334 @@ HeapTypeGenerator::create(Random& rand, FeatureSet features, size_t n) {
...
+
+  // The fields we will make nullable.
+  std::unordered_set<FieldPos> nullables;
+
+  SubTypes subtypes;
+
+  Inhabitator(const std::vector<HeapType>& types)
+    : types(types), subtypes(types) {}
+
+  enum Variance { Invariant, Covariant, Contravariant };
```
kripken, line 692: Please add a comment for what these are.

src/tools/fuzzing/heap-types.cpp
```diff
@@ -654,4 +656,334 @@ HeapTypeGenerator::create(Random& rand, FeatureSet features, size_t n) {
...
+void Inhabitator::markBottomRefsNullable() {
+  for (auto type : types) {
+    auto children = type.getTypeChildren();
+    for (size_t i = 0; i < children.size(); ++i) {
+      auto child = children[i];
+      if (child.isRef() && child.getHeapType().isBottom() &&
+          child.isNonNullable()) {
+        markNullable({type, i});
+      }
+    }
```
kripken, line 780: I think we need to handle `HeapType::ext` as well. A `(ref ext)` can't be created by the fuzzer, unlike `(ref any)`.

src/tools/fuzzing/heap-types.cpp
```diff
@@ -654,4 +656,334 @@ HeapTypeGenerator::create(Random& rand, FeatureSet features, size_t n) {
...
+
+  // Copy types. Update references to point to the corresponding new type and
+  // make them nullable where necessary.
+  auto updateType = [&](FieldPos pos, Type& type) {
+    if (!type.isRef()) {
+      return;
+    }
+    auto heapType = type.getHeapType();
+    auto nullability = type.getNullability();
+    if (auto it = typeIndices.find(heapType); it != typeIndices.end()) {
```
kripken, line 877: Is this line so we refer to the temp types in the builder?

src/tools/fuzzing/heap-types.cpp
```diff
@@ -654,4 +656,334 @@ HeapTypeGenerator::create(Random& rand, FeatureSet features, size_t n) {
...
+  deduplicated = inhabitator.build();
+
+  // Re-duplicate and re-insert basic types as necessary.
+  std::vector<HeapType> result;
+  for (size_t i = 0; i < types.size(); ++i) {
+    if (deduplicatedIndices[i] == (size_t)-1) {
+      assert(types[i].isBasic());
+      result.push_back(types[i]);
+    } else {
+      result.push_back(deduplicated[deduplicatedIndices[i]]);
```
kripken, line 983: Don't the duplicates get pushed at the end, but they weren't necessarily there before? (If so, Changing positions inside a rec group can change whether groups get merged, so can't that be an issue?)

src/tools/fuzzing/heap-types.cpp
tlively: Are you looking for a comment explaining what variance is or just why we need to care about it here?

src/tools/fuzzing/heap-types.cpp
tlively: Right, we need the newly built types to refer to other newly built types rather than the old types where possible. Right now the only case where this doesn't apply is for references to basic heap types.

src/tools/fuzzing/heap-types.cpp
tlively: No, we're building up `result` one item at a time, so the item that gets pushed here will be at index `i` in the results. `deduplicatedIndices[i]` tells us where in the (potentially smaller) list of built types we can find the right type, whether or not that type was a duplicate.

src/tools/fuzzing/heap-types.cpp
kripken: A quick summary of what these type-theoretic concepts are would be nice. I don't think we assume that knowledge in general.

test/lit/fuzz-types/nominal.test
```diff
@@ -20,3 +20,27 @@
...
+;; CHECK-NEXT: Inhabitable types:
+;; CHECK-NEXT:
+;; CHECK-NEXT: Built 20 types:
+;; CHECK-NEXT: (type $0 (type $0 (struct (field (mut i16) i31ref f32 f32 f64))))
+;; CHECK-NEXT: (type $1 (type $1 (func (param f64 v128))))
+;; CHECK-NEXT: (type $2 (type $2 (struct (field (mut (ref null $19)) f64 arrayref))))
+;; CHECK-NEXT: (type $3 (type $3 (struct_subtype (field (mut i16) (ref i31) f32 f32 f64) $0)))
+;; CHECK-NEXT: (type $4 (type $4 (struct)))
+;; CHECK-NEXT: (type $5 none)
+;; CHECK-NEXT: (type $6 (type $6 (array (mut eqref))))
```
kripken, line 33: The printing here looks wrong: `(type $6 (type $6` etc.

test/lit/fuzz-types/nominal.test
```diff
@@ -20,3 +20,27 @@
...
+;; CHECK-NEXT: (type $10 (type $10 (struct_subtype (field (mut i16) (ref i31) f32 f32 f64) $0)))
+;; CHECK-NEXT: (type $11 (type $11 (func (result f64))))
+;; CHECK-NEXT: (type $12 (type $12 (func_subtype (param f64 v128) $1)))
+;; CHECK-NEXT: (type $13 (type $13 (func_subtype (param f64 v128) $12)))
+;; CHECK-NEXT: (type $14 (type $14 (func_subtype (result f64) $11)))
+;; CHECK-NEXT: (type $15 (type $15 (func_subtype (result f64) $14)))
+;; CHECK-NEXT: (type $16 (type $16 (func (param (ref struct)) (result structref))))
+;; CHECK-NEXT: (type $17 (type $17 (array (mut (ref $2)))))
+;; CHECK-NEXT: (type $18 (type $18 (array (ref $10))))
+;; CHECK-NEXT: (type $19 (type $19 (struct_subtype (field (mut i16) (ref i31) f32 f32 f64 (mut f32)) $0)))
```
kripken, line 46: Other than that, these results look identical to before - was any work done to make them inhabitable?

test/lit/fuzz-types/nominal.test
tlively: Ah yes. That's actually not related to this PR, so I'll fix that separately.

test/lit/fuzz-types/isorecursive.test
```diff
@@ -30,3 +30,37 @@
...
+;; CHECK-NEXT: (rec
+;; CHECK-NEXT:  (type $10 (type $10 (func_subtype (param (ref eq) i32 i64 f64 f64 anyref v128) (result i64) $9)))
+;; CHECK-NEXT:  (type $11 (type $11 (array_subtype (mut funcref) $6)))
+;; CHECK-NEXT:  (type $12 (type $12 (array nullref)))
+;; CHECK-NEXT:  (type $13 none)
+;; CHECK-NEXT:  (type $14 (type $14 (array (ref $6))))
+;; CHECK-NEXT:  (type $15 (type $15 (array i32)))
+;; CHECK-NEXT: )
+;; CHECK-NEXT: (rec
+;; CHECK-NEXT:  (type $16 (type $16 (array_subtype nullref $12)))
```
tlively, line 62: This `nullref` is new.

test/lit/fuzz-types/nominal.test
tlively: Unfortunately these seeds don't require much fixing up. I commented separately to point out the one change I did notice.

The fuzzing was extremely effective in catching bugs with this implementation, though. Most bugs were surfaced with just a handful of iterations. Now the fuzzer runs for hundreds of thousands of iterations without finding new problems.

src/tools/fuzzing/heap-types.cpp
tlively: I think we do assume it in our implementation of subtyping, but I'll add a brief explanation here.

test/lit/fuzz-types/isorecursive.test
kripken: Ah, I just compared the nominal one manually, and saw no diff... but yes, this looks fixed up. sgtm

kripken (approved):

tlively:
Actually using this utility in the full fuzzer will require some extra work because there may be e.g. `struct.get`s that were previously non-nullable and used in a context that required them to be non-nullable, but now they are nullable. Would it be helpful if I wrote a pass that did the necessary fixups as a follow-up?

kripken:
Hmm, I don't think that's necessary. This will be fixing up types before we create anything with them. So there won't be `struct.get`s yet.

tlively:
Sounds good unless we want to start using initial contents that already contain uninhabitable types.

kripken:
https://github.com/WebAssembly/binaryen/pull/5539 filters out noninhabitable types, which will handle that.

tlively:
Oh, if our plan is to filter types anyway, then the complexity of this type transformation probably isn't warranted 😅

kripken:
I think both are helpful. Crude filtering can remove lots of types in some cases. But we also need filtering so handle existing testcases.

PR #5542, opened 2023-03-03
Note function signature param/result features for validation

kripken:
As with #5535, this was not noticed because it can only happen on very
small modules where the param/result type appears nowhere else but
in a function signature, but my wip fuzzer branch now finds it.

### DIFF
--- a/test/lit/validation/ref-func-simd.wast
+++ b/test/lit/validation/ref-func-simd.wast
@@ -1,0 +1,16 @@
+;; A function type that uses a SIMD type requires the SIMD feature.
+;; RUN: not wasm-opt --enable-reference-types %s 2>&1 | filecheck %s
+
+;; CHECK: all used types should be allowed
+
+(module
+  (type $simd-user (func (param v128)))
+
+  (global $g (ref null $simd-user) (ref.null func))
+)
+
+;; But it passes with the feature enabled.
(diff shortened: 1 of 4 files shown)


test/lit/validation/ref-func-simd.wast
```diff
@@ -0,0 +1,16 @@
...
+  (type $simd-user (func (param v128)))
+
+  (global $g (ref null $simd-user) (ref.null func))
+)
+
+;; But it passes with the feature enabled.
+;; RUN: wasm-opt --enable-reference-types --enable-simd %s
+
+
+
```
tlively, line 16: It would be good to remove the unnecessary whitespace.

tlively (approved):

tlively:
We have a `HeapTypeGraphWalker` utility in wasm-type.cpp that might be useful.

kripken:
Thanks, but I think we just need a recursion guard? I pushed that now. Diff without whitespace is much smaller, this just wraps the inside in a function.

tlively (commented):
It seems fishy that we recurse into function types but not structs or arrays.

kripken:
Good point, that was wrong too. I pushed a fix now that `HeapTypeGraphWalker`, it does seem like we should not duplicate walking logic here.

I also reused `getTypeChildren` from your PR #5541 so this should land after that.

PR #5543, opened 2023-03-03
Fix type printing in the type fuzzer

tlively:
In #5437 we updated type printing so that printing a heap type would print its
name in addition to its contents. We had already been separately printing type
names in the type fuzzer, so after that change we were printing each type name
twice. Remove the redundant printing in the fuzzer to fix the error.


tlively:
* main
  * **PR #5543** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5543" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5543?utm_source=stack-comment).

kripken (approved):

PR #5555, opened 2023-03-08
Integrate the heap type fuzzer into the main fuzzer

kripken:
With this we generate random GC types that may be used in creating
instructions later.

We don't create many instructions yet, which will be the next step after
this.

Also add some trivial assertions in some places, that have helped
debugging in the past.


kripken:
This uncovered some UB in call order in the heap fuzzer. Last two commits should fix it. This is noticeable on our CI because gcc and clang have opposite behavior in that UB.

tlively (approved):
Makes sense, thanks for the fix!

kripken:
I pushed a commit to stop fuzzing TypeMerging for now due to https://github.com/WebAssembly/binaryen/issues/5556. With that things look stable, and this can land, WDYT?

tlively (approved):
Sounds good! I should be able to get to the TypeMerging bug today, too.

PR #5558, opened 2023-03-09
Emit the fuzzer hashMemory function after modifications

kripken:
Previously we emitted it early, and would then modify it in random ways
like other initial content. But this function is called frequently during
execution, so if we were unlucky and modded that function to trap then
basically all other functions would trap as well.

After fixing this, some places assert on not having any functions or types
to pick a random one from, so fix those places too.


kripken:
(actual hashing code is just moved around)

PR #5562, opened 2023-03-10
Fuzzer: CompareVMs: Do not compare when hitting a host limitation

kripken:
For example, we might hit an allocation limit in the wasm, but the
optimized wasm might optimize that allocation out. So we need to
ignore comparisons in such cases, as we cannot expect the output
to be identical. We already do similar things for FuzzExec and
#5560 adds it for TrapsNeverHappen; this adds it to CompareVMs.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -839,7 +839,10 @@ class CompareVMs(TestCaseHandler):
                 # NaNs can differ from wasm VMs
                 return not NANS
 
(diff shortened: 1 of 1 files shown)


kripken:
Friendly ping on this.

Also it might help with https://github.com/WebAssembly/binaryen/pull/5572#issuecomment-1468311039 (although that sounds like the flipped situation perhaps? if so, I may have a fix for it later on, I'm still thinking about that)

tlively (approved):

PR #5564, opened 2023-03-10
Fuzzer: Avoid emitting massive nested structs

kripken:
The nesting limit of around 20 was enough to cause exponential blowup. A 20K
input file lead to a 2GB wasm in one case I saw (!) which takes many seconds to
fuzz.

Instead, reduce the limit, and also check if `random` tells us that the random
input is done; when that's done we should stop, which limits us to `O(input size)`.


src/tools/fuzzing/fuzzing.cpp
```diff
@@ -2178,15 +2178,18 @@ Expression* TranslateToFuzzReader::makeConstCompoundRef(Type type) {
...
   // Note that we might have cycles of types where some are non-nullable. We
   // will only stop here when we exceed the nesting and reach a nullable one.
   // (This assumes there is a nullable one, that is, that the types are
   // inhabitable.)
-  const auto LIMIT = 2 * NESTING_LIMIT;
+  const auto LIMIT = NESTING_LIMIT + 1;
   AutoNester nester(*this);
-  if (type.isNullable() && (oneIn(10) || nesting >= LIMIT)) {
+  if (type.isNullable() &&
+      (random.finished() || nesting >= LIMIT || oneIn(LIMIT - nesting + 1))) {
```
tlively, line 2192: Where is `nesting` updated? And what is an `AutoNester`?

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -2178,15 +2178,18 @@ Expression* TranslateToFuzzReader::makeConstCompoundRef(Type type) {
...
+  // this limit is really a last resort we want to never reach. Also, increase
+  // the chance to emit a null as |nesting| rises, to avoid deep recursive
+  // structures.
   //
   // Note that we might have cycles of types where some are non-nullable. We
   // will only stop here when we exceed the nesting and reach a nullable one.
   // (This assumes there is a nullable one, that is, that the types are
   // inhabitable.)
-  const auto LIMIT = 2 * NESTING_LIMIT;
+  const auto LIMIT = NESTING_LIMIT + 1;
```
tlively, line 2189: Why the `+ 1`? It looks like just reusing `NESTING_LIMIT` would work as well.

src/tools/fuzzing/fuzzing.cpp
kripken: AutoNester increments `nesting` and auto-decrements it when we exit the function scope (using a destructor).

src/tools/fuzzing/fuzzing.cpp
kripken: As the comment above says, this tries to be higher than that limit so that the normal nesting-limiting mechanisms can kick in first, hopefully. They are less "severe", in that here we'll emit a null which is likely to cause a trap if it is used.

tlively (approved):

kripken:
I noticed in the fuzzer that we also need to handle the non-nullable case, since we have such things in our test suite. This emits a cast to non-null of a null in that case, which is bad but less bad than infinite recursion. At least it is rare.

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -2178,18 +2178,30 @@ Expression* TranslateToFuzzReader::makeConstCompoundRef(Type type) {
...
+std::cout << "maek " << type << " : " << type.getHeapType() << " : " << nesting << '\n';
```
tlively, line 2190: stale!

tlively (commented):
Isn't the nesting limit supposed to avoid infinite recursion? We also filter out infinite non-nullable cycles, so why is this latest code necessary?

src/tools/fuzzing/fuzzing.cpp
kripken: heh, thanks!

kripken:
> Isn't the nesting limit supposed to avoid infinite recursion?

It is, but it is enforced manually. We do it in `make(type)` which is the central place, but the problem is that `makeConstantRef` can recurse with itself, so we need to manually check for the limit here.

> We also filter out infinite non-nullable cycles, so why is this latest code necessary?

Only because of existing content. We don't create those, and we don't even add them to `interestingHeapTypes`, but they do exist in the module, and we have things like `mutate` that replace a thing with a type with another thing of the same type. (We could in principle modify those places instead, but it would be more work I think, and less useful.)

kripken:
I also added a fix for globals. We can't try to emit an uninhabitable one, or else the new code in this PR could end up (if we recurse too much) emitting a RefAsNonNull in a global. To avoid that, only pick inhabitable types there.

kripken:
Well, let's not overthink it... in the very rare case we end up with a RefAs in a global, we can flip it to be an MVP type. I  added that now.

kripken:
This seems to work well for now. We can improve later if we have more ideas.

PR #5567, opened 2023-03-10
Fix ambiguous operators under C++20

tlively:
When resolving `operator!=`, C++20 also considers `operator==` implementations
when the types on `operator!=` do not match exactly. This caused the modified
code to have no most-specific overload to choose, resulting in an error. This is
actually a bug in the language that is being fixed, but there exist compilers
without the fix applied.

Work around the problem by updating the types in the declaration of `operator==`
and `operator!=` to be more exact.

This is a copy of #5029 with formatting fixes.

### DIFF
--- a/src/support/parent_index_iterator.h
+++ b/src/support/parent_index_iterator.h
@@ -49,10 +49,12 @@ template<typename Parent, typename Iterator> struct ParentIndexIterator {
 
   Iterator& self() { return *static_cast<Iterator*>(this); }
 
-  bool operator==(const Iterator& other) const {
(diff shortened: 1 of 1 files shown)


tlively:
* main
  * **PR #5567** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5567" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5567?utm_source=stack-comment).

kripken (approved):

PR #5569, opened 2023-03-10
Fuzzer: Limit array sizes

kripken:
Even with a 1% chance of a huge array, there is a second problem aside from
hitting an allocation failure, which is DoS - building such a huge array of
Literals takes noticeable time in the fuzzer. Instead, just limit array max sizes,
which is consistent with what we do for struct sizes etc.

### DIFF
--- a/src/tools/fuzzing/parameters.h
+++ b/src/tools/fuzzing/parameters.h
@@ -38,7 +38,7 @@ constexpr int MAX_TUPLE_SIZE = 6;
 // The maximum number of struct fields.
 static const int MAX_STRUCT_SIZE = 6;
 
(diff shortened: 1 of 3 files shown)


src/tools/fuzzing/parameters.h
```diff
@@ -38,6 +38,9 @@ constexpr int MAX_TUPLE_SIZE = 6;
 // The maximum number of struct fields.
 static const int MAX_STRUCT_SIZE = 6;
 
+// The maximum number of arrays.
```
tlively, line 41:
```suggestion
// The maximum number of elements in an array.
```

tlively (approved):

PR #5571, opened 2023-03-13
Fix fuzzer emitting invalid constant expressions

tlively:
The fuzzer had code to avoid emitting `global.get` of locally defined (i.e.
non-imported) globals in global initializers and data segment offsets, but that
code only handled top-level `global.get` because it predated the extended-const
proposal. Unfortunately this bug went undetected until #5557, which fixed the
validator to make it reject invalid uses of `global.get` in constant
expressions.

Fix the bug so the validator no longer produces invalid modules.

### DIFF
--- a/src/tools/fuzzing/fuzzing.cpp
+++ b/src/tools/fuzzing/fuzzing.cpp
@@ -313,10 +313,12 @@ void TranslateToFuzzReader::setupGlobals() {
       global->module = global->base = Name();
       global->init = makeConst(global->type);
     } else {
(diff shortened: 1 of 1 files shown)


tlively:
* main
  * **PR #5571** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5571" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5571?utm_source=stack-comment).

kripken (approved):

PR #5573, opened 2023-03-14
Fuzzer: Pick interesting subtypes in getSubType(HeapType)

kripken:


kripken:
> but I guess that doesn't cover types from the initial contents

Yes, exactly, this is meant to be fully general. We may add more types in the future in other ways.

> I'm not sure calculating subtypes lazily buys us anything, either; it can end up being more expensive than eagerly computing them after just a single supertype is queried since Type::isSubType is linear in depth of the potential subtype.

? Without memoizing the list of subtypes per heap type, every time we want to pick a subtype we'd do `O(#interesting types)` work. But with this, we do `O(1)` after the first time.

(Another option I considered is to do constant work by picking up to K random indexes in the interesting types, and seeing if they happen to be subtypes, but that seems like it wouldn't scale.)

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -3331,7 +3327,25 @@ HeapType TranslateToFuzzReader::getSubType(HeapType type) {
...
+  auto iter = interestingHeapSubTypes.find(type);
+  if (iter == interestingHeapSubTypes.end()) {
+    std::vector<HeapType> subTypes;
+    for (auto possible : interestingHeapTypes) {
+      // We avoid storing the type itself among its subtypes since there is
+      // already a good chance to return the type itself from this function.
+      if (possible != type && HeapType::isSubType(possible, type)) {
+        subTypes.push_back(possible);
+      }
+    }
```
tlively, line 3341: This loop is `O(n^2)` where `n` is the number of heap types because `HeapType::isSubType` is linear in the number of heap types (in the worst case).

In contrast, walking the type graph once and collecting all the subtypes for all types naively without using `HeapType::isSubType` would be `O(n^2)`, but would collect the subtypes for _all_ types, not just a single type.

(It's also possible to collect all the subtypes for all types in linear time using a fancy flattened representation of the subtype tree, but the complexity is certainly not worth it.)

kripken: Now I see, thanks! I had assumed not many would be called, and didn't think it through that even a single iteration is quadratic.

Rewritten to be efficient.

tlively (approved):

PR #5574, opened 2023-03-14
[Wasm GC] Remove RefIsFunc and RefIsI31

kripken:
These are deprecated, and we can implement them using `ref.test`.

V8 has already removed the deprecated ones. After this PR, this module
successfully validates there:
```wat
(module
 (func $test (result i32)
  (ref.is_func
   (ref.null func)
  )
 )
)
```


tlively (approved):
LGTM. It would be good to either clarify in the commit title that this is only removing them from the binary input and output, or to go all the way and remove them from the text input and output as well.

PR #5575, opened 2023-03-14
Fuzzer: Generate both immutable and mutable globals

kripken:


src/tools/fuzzing/fuzzing.cpp
```diff
@@ -1471,7 +1472,7 @@ Expression* TranslateToFuzzReader::makeGlobalSet(Type type) {
     return makeTrivial(Type::none);
   }
   auto name = pick(it->second);
-  if (isValidGlobal(name)) {
+  if (isValidGlobal(name) && wasm.getGlobal(name)->mutable_) {
```
tlively, line 1475: Maybe we should track `mutableGlobalsByType` as well to make it more likely that we successfully create a `GlobalSet` here.

kripken: Good idea, done.

tlively (approved):

PR #5576, opened 2023-03-15
Support interpretation of extern.externalize and extern.internalize

tlively:
To allow the external and internal reference values to be differentiated yet
round-trippable, set the `Literal` type to externref on external references, but
keep the gcData the same for both. The only exception is for i31 references, for
which the externalized version gets a `gcData` that contains a copy of the
original i31 literal.

### DIFF
--- a/src/literal.h
+++ b/src/literal.h
@@ -52,10 +52,6 @@ class Literal {
     // also use this to store String data, as it is similarly stored on the
     // heap.
     std::shared_ptr<GCData> gcData;
-    // TODO: Literals of type `anyref` can only be `null` currently but we
-    // will need to represent external values eventually, to
-    // 1) run the spec tests and fuzzer with reference types enabled and
-    // 2) avoid bailing out when seeing a reference typed value in precompute
   };
 
 public:
@@ -664,6 +660,9 @@ public:
   Literal relaxedFmsF32x4(const Literal& left, const Literal& right) const;
   Literal relaxedFmaF64x2(const Literal& left, const Literal& right) const;
   Literal relaxedFmsF64x2(const Literal& left, const Literal& right) const;
+
+  Literal externalize() const;
+  Literal internalize() const;
 
 private:
   Literal addSatSI8(const Literal& other) const;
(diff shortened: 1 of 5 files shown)


tlively:
* main
  * **PR #5576** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5576" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5576?utm_source=stack-comment).

src/literal.h
```diff
@@ -665,6 +661,9 @@ class Literal {
   Literal relaxedFmaF64x2(const Literal& left, const Literal& right) const;
   Literal relaxedFmsF64x2(const Literal& left, const Literal& right) const;
 
+  Literal externalize() const;
```
kripken, line 664: Please document in this header how the encoding is done for externalized/internalized data.

src/wasm/literal.cpp
```diff
@@ -2590,4 +2592,42 @@ Literal Literal::relaxedFmsF64x2(const Literal& left,
...
+      }
+      case HeapType::string:
+      case HeapType::stringview_wtf8:
+      case HeapType::stringview_wtf16:
+      case HeapType::stringview_iter:
+        WASM_UNREACHABLE("TODO: string literals");
+      default:
+        break;
+    }
+    WASM_UNREACHABLE("unexpected type");
```
kripken, line 2616:
```suggestion
        WASM_UNREACHABLE("unexpected type");
    }
```

src/wasm/literal.cpp
```diff
@@ -2590,4 +2592,42 @@ Literal Literal::relaxedFmsF64x2(const Literal& left,
...
+  }
+  return Literal(gcData, HeapType::ext);
+}
+
+Literal Literal::internalize() const {
+  assert(Type::isSubType(type, Type(HeapType::ext, Nullable)) &&
+         "can only internalize external references");
+  if (isNull()) {
+    return Literal(std::shared_ptr<GCData>{}, HeapType::none);
```
kripken, line 2625: Can function references not be externalized and internalized? (They'd need a different bottom type.)

kripken (approved):
Nice encoding!

src/wasm/literal.cpp
tlively: That is my understanding, yes. There's no way to convert from the `func` hierarchy to any other hierarchy.

PR #5577, opened 2023-03-15
[Wasm GC] Fuzz ref.test

kripken:


src/tools/fuzzing/fuzzing.cpp
```diff
@@ -3090,6 +3091,43 @@ Expression* TranslateToFuzzReader::makeRefEq(Type type) {
...
+      // Cast is a subtype of ref.
+      refType = getReferenceType();
+      castType = getSubType(refType);
+      break;
+    case 2:
+      // Ref is a subtype of cast.
+      castType = getReferenceType();
+      refType = getSubType(castType);
+      break;
+      break;
```
tlively, line 3123:
```suggestion
```

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -3090,6 +3091,43 @@ Expression* TranslateToFuzzReader::makeRefEq(Type type) {
...
+  Type refType, castType;
+  switch (upTo(3)) {
+    case 0:
+      // Totally random.
+      refType = getReferenceType();
+      castType = getReferenceType();
+      // They must share a bottom type in order to validate.
+      if (refType.getHeapType().getBottom() ==
+          castType.getHeapType().getBottom()) {
+        break;
```
tlively, line 3108: Since there are only three hierarchies, it seems reasonable to have `getInternalReferenceType`, `getExternalReferenceType`, and `getFunctionReferenceType` and avoid potential mismatches entirely. WDYT?

kripken: Interesting. That could work, but it might need changes later. That is, the current code would work if we added more hierarchies in the future since it is fully general.

It does have the downside of having a lower chance of unrelated pairs, but I don't think that's a big deal. It's not hugely lower, and it's not the common case anyhow so less need to fuzz.

tlively: For the first point, adding a new hierarchy with all the new types and instructions would require a bunch of work in the fuzzer anyway, so I'm not too worried about having to add some extra code here to support it as well.

I'm not sure I understand the second point about the chance being "not hugely lower." My suggestion is that we avoid mismatches by construction, so the chances go to zero, which seems significant.

kripken: It's true new hierarchies would require work. But the current code would mean no work would be required at this location. Maybe not a big difference, but it's nice. And the code is simpler.

The second point was that it's fine if we have, say 10% to get unrelated types, and 45%/45% for one being a subtype of the other or vice versa. That's good enough. The "failure" case here is just to fall back to one being a subtype of the other, which is still useful to fuzz.

tlively: Ok fair enough. Another question is how much do we care about wasting entropy? It seems better not to have to throw away random decisions we've made.

tlively (approved):

src/tools/fuzzing/fuzzing.cpp
kripken: In general it's best not to waste entropy, yeah, but also it's not a big deal - like the Fed, we can always print more. (This comment is not financial advice; I have the fiscal understanding of a first-grader.)

Overall when there's a tradeoff with code clarity and simplicity I think I'd prefer those over wasting some small amount of entropy, which is the case here. Wasted entropy just makes us very slightly slower here.

PR #5579, opened 2023-03-15
Use SmallVector in TypeUpdating::updateParamTypes()

kripken:
Noticed when profiling for https://github.com/WebAssembly/binaryen/issues/5561#issuecomment-1470783261

### DIFF
--- a/src/wasm-builder.h
+++ b/src/wasm-builder.h
@@ -19,6 +19,7 @@
 
 #include "ir/manipulation.h"
 #include "parsing.h"
+#include "support/small_vector.h"
 #include "wasm.h"
 
 namespace wasm {
@@ -227,6 +228,13 @@ public:
     ret->name = name;
     ret->list.set(items);
     ret->finalize(type);
+    return ret;
+  }
+  template<size_t N>
+  Block* makeBlock(const SmallVector<Expression*, N>& items) {
+    auto* ret = wasm.allocator.alloc<Block>();
+    ret->list.set(items);
+    ret->finalize();
     return ret;
   }
   If* makeIf(Expression* condition,
--- a/src/ir/type-updating.cpp
+++ b/src/ir/type-updating.cpp
@@ -19,6 +19,7 @@
 #include "ir/local-structural-dominance.h"
 #include "ir/module-utils.h"
 #include "ir/utils.h"
+#include "support/small_vector.h"
 #include "support/topological_sort.h"
 #include "wasm-type-ordering.h"
 #include "wasm-type.h"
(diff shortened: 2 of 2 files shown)


src/wasm-builder.h
```diff
@@ -229,6 +230,13 @@ class Builder {
     ret->finalize(type);
     return ret;
   }
+  template<size_t N>
+  Block* makeBlock(const SmallVector<Expression*, N>& items) {
```
tlively, line 234: Would it be worth deduplicating all the different versions of `makeBlock` into a few templatized versions that can take any iterable for the items?

tlively (approved):

tlively:
How much does this improve performance, out of curiosity?

TerrorJack:
I tried the artifacts on this branch to run the same test in the same environment in #5561. The run time is now down to 18s.

kripken:
@tlively I didn't see a speed difference in my own local tests (perhaps because I can't reproduce the malloc contention issue on my machine). But I verified that on common codebases there are fewer params than this, and it will definitely save some work in those cases, so it looks right in general.

@TerrorJack Interesting, to what is that compared to? I see various numbers in that issue but all are much larger than 18s so I'm not sure which is relevant.

src/wasm-builder.h
tlively: #5583

kripken:
I'm still confused @TerrorJack , sorry, since the toplevel post contains this:

```
Result sans mimalloc:

$ time bench ./test.sh
benchmarking ./test.sh
time                 221.9 s    (218.3 s .. 224.6 s)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 219.3 s    (218.1 s .. 220.3 s)
std dev              1.411 s    (1.155 s .. 1.566 s)
variance introduced by outliers: 19% (moderately inflated)

real    58m35.860s
user    129m50.133s
sys     2395m41.639s

Result with mimalloc:

$ time bench ./test.sh 
benchmarking ./test.sh
time                 14.06 s    (13.38 s .. 14.86 s)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 13.94 s    (13.82 s .. 14.03 s)
std dev              123.8 ms   (76.09 ms .. 151.4 ms)
variance introduced by outliers: 19% (moderately inflated)


real    3m43.584s
user    45m38.783s
sys     0m40.349s
```

To which of those numbers should the 18 seconds be compared? The 219 seems too high (this PR can't help that much) but the 14 is too low (since you say it decreased, not increased). Apologies, I'm probably not reading something right...

PR #5582, opened 2023-03-16
[Wasm GC] wasm-ctor-eval: Handle externalized data

kripken:


src/tools/wasm-ctor-eval.cpp
```diff
@@ -572,13 +572,28 @@ struct CtorEvalExternalInterface : EvallingModuleRunner::ExternalInterface {
...
+
+    // An externalized i31 has both original.isData and value.isData as being
+    // false, so we need to check both. That is, we can hand off for normal
+    // handling anything that is not (externalized or not) GC data. GC data is
+    // the only thing that requires special handling here by adding globals as
+    // needed.
+    if (!original.isData() && !value.isData()) {
+      // This can be handled normally. We can pass in the original value here as
+      // makeConstantExpression() knows how to handle externalization itself.
+      return builder.makeConstantExpression(original);
```
tlively, line 596: Eventually we will have to support externref values that actually are external, not just just externalized internal references. (The upstream spec tests depend on this, so we will need to implement it if we ever want to be able to run those.) This code doesn't look forward compatible with that because it seems unlikely that `makeConstantExpression` will be able to do anything with a truly external reference. Should we check more explicitly for i31 here?

kripken: Makes sense. I added some assertions and now we check that external things are either data or i31.

tlively (approved):
LGTM!

PR #5583, opened 2023-03-16
[NFC] Templatize `makeBlock` so it can be used with any iterable

tlively:
Replace the different overloads we previously had for different kinds of
containers with generic templates. We still need dedicated overloads for
`std::initializer_list` because it is never inferred in a template context,
though. Also, since `std::initializer_list` does not allow subscripting, update
the arena vector implementation to use iterators instead now that initializer
lists can be passed down to that layer without being reified as vectors.

### DIFF
--- a/src/wasm-builder.h
+++ b/src/wasm-builder.h
@@ -167,7 +167,6 @@ public:
   }
 
   // IR nodes
-
   Nop* makeNop() { return wasm.allocator.alloc<Nop>(); }
   Block* makeBlock(Expression* first = nullptr) {
     auto* ret = wasm.allocator.alloc<Block>();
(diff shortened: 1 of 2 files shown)


tlively:
* main
  * **PR #5583** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5583" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5583?utm_source=stack-comment).

src/wasm-builder.h
```diff
@@ -183,52 +182,47 @@ class Builder {
     ret->finalize();
     return ret;
   }
-  Block* makeBlock(const std::vector<Expression*>& items) {
+
+  template<typename T>
+  using bool_if_not_expr_t =
+    std::enable_if_t<std::negation_v<std::is_convertible<T, Expression*>>,
```
kripken, line 188: Btw, any idea why I couldn't get this to work with `std::is_base_of` instead of `is_convertible`?

kripken (approved):
Thanks!

src/wasm-builder.h
tlively: I think it has to do with having to remove the pointer, presumably using `std::remove_pointer`, although TBH I couldn't get that working either, so I'm definitely missing something else there, too.

PR #5592, opened 2023-03-20
[Wasm GC] Fix detection of externalize/internalize as constant

kripken:
Both `isValidInConstantExpression` and `isSingleConstantExpression` must look
recursively at the internals of a RefAs that externalizes and internalizes, or else
we might do something like externalize a `local.get`, which is not constant.

`getLiteral` must handle externalize/internalize as well, and return a properly-
modified literal.

Without these fixes the testcase hits different internal assertions, and we either
fail to recognize something is constant or not, or think that it is but fail to
produce a literal for it.

### DIFF
--- a/src/ir/properties.cpp
+++ b/src/ir/properties.cpp
@@ -36,6 +36,8 @@ bool isGenerative(Expression* curr, FeatureSet features) {
   return scanner.generative;
 }
 
+// Checks an expression in a shallow manner (i.e., does not check children) as
+// to whether it is valid in a wasm constant expression.
 static bool isValidInConstantExpression(Module& wasm, Expression* expr) {
   if (isSingleConstantExpression(expr) || expr->is<StructNew>() ||
       expr->is<ArrayNew>() || expr->is<ArrayNewFixed>() || expr->is<I31New>() ||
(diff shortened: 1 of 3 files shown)


src/ir/properties.cpp
```diff
@@ -43,6 +43,12 @@ static bool isValidInConstantExpression(Module& wasm, Expression* expr) {
     return true;
   }
 
+  if (auto* refAs = expr->dynCast<RefAs>()) {
+    if (refAs->op == ExternExternalize || refAs->op == ExternInternalize) {
+      return isValidInConstantExpression(wasm, refAs->value);
+    }
+  }
```
tlively, line 50: We shouldn't need to recurse here because recursion should be handled by `isValidConstantExpression`.

tlively (approved):
LGTM % comment.

Also, as a follow-up it might make sense to reimplement `isSingleConstantExpression` in terms of `isValidConstantExpression` plus a check for `global.get` of imported globals so we don't have two separate-but-similar implementations of these similar concepts.

kripken:
Agreed a followup makes sense here. There is a TODO to merge those functions and we should probably do it.

PR #5593, opened 2023-03-20
Basic implementation of wasm-quickbind.

eyebrowsoffire:


tlively:
@eyebrowsoffire and I chatted elsewhere about potentially using and improving wasm-split, since it is more general than this proposed tool.

PR #5597, opened 2023-03-22
Use more than a single wasm page in Asyncify fuzzing

kripken:
I saw a testcase fail on the internal assertion of the buffer being too small.
Enlarge it to use as much of the memory we have anyhow to reduce that
risk (we can use 15 pages instead of 1, without changing anything else).

### DIFF
--- a/scripts/fuzz_shell.js
+++ b/scripts/fuzz_shell.js
@@ -45,7 +45,9 @@ var Asyncify = {
   sleeps: 0,
   maxDepth: 0,
   DATA_ADDR: 4,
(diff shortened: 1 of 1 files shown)


tlively (approved):
LGTM. Is it ever useful to fuzz the case where the buffer is too small?

kripken:
If it's too small we trap and the user just needs to increase the size, there's nothing else to do there, really. I suppose in theory we could fuzz to make sure we trap instead of doing something else, but we have regular tests for that.

PR #5603, opened 2023-03-29
Do not treat `atomic.fence` as using a memory

tlively:
Update RemoveUnusedModuleElements so that it no longer keeps the memory alive
due to an `atomic.fence` instruction and update validation to allow modules to
use `atomic.fence` without a memory.

PR #5602, opened 2023-03-29
[NFC] Port remove-unused-module-elements_all-features.wast to lit

tlively:
Port the test automatically using scripts/port_passes_tests_to_lit.py.

PR #5601, opened 2023-03-29
[NFC] Simplify initialization in RemoveUnusedModuleElements.cpp

tlively:
Use copy-list-initialization to shorten the code and reduce visual redundancy.

PR #5604, opened 2023-03-29
Support multiple memories in RemoveUnusedModuleElements

tlively:
Add support for memory and data segment module elements and treat them uniformly
with other module elements rather than as special cases. There is a cyclic
dependency between memories (or tables) and their active segments because
exported or accessed memories (or tables) keep their active segments alive, but
active segments for imported memories (or tables) keep their memories (or
tables) alive as well.

### DIFF
--- a/src/passes/RemoveUnusedModuleElements.cpp
+++ b/src/passes/RemoveUnusedModuleElements.cpp
@@ -681,7 +681,9 @@ struct RemoveUnusedModuleElements : public Pass {
       }
     }
 
-    // Active segments that write to imported tables and memories are roots.
+    // Active segments that write to imported tables and memories are roots
+    // because those writes are externally observable even if the module does
+    // not otherwise use the tables or memories.
     ModuleUtils::iterActiveDataSegments(*module, [&](DataSegment* segment) {
       if (module->getMemory(segment->memory)->imported() &&
           !segment->data.empty()) {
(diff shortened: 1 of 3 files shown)


tlively:
* main
  * **PR #5601** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5601" target="_blank">Graphite</a> 
    * **PR #5602** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5602" target="_blank">Graphite</a> 
      * **PR #5603** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5603" target="_blank">Graphite</a> 
        * **PR #5604** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5604" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5604?utm_source=stack-comment).

src/passes/RemoveUnusedModuleElements.cpp
```diff
@@ -643,20 +681,26 @@ struct RemoveUnusedModuleElements : public Pass {
...
+        roots.emplace_back(ModuleElementKind::Memory, curr->value);
       }
     }
-    // Check for special imports, which are roots.
-    bool importsMemory = false;
-    if (!module->memories.empty() && module->memories[0]->imported()) {
-      importsMemory = true;
-    }
+
+    // Active segments that write to imported tables and memories are roots.
```
kripken, line 689: Why are these roots? The import may not be used, in which case neither are they.

src/passes/RemoveUnusedModuleElements.cpp
```diff
@@ -702,6 +746,24 @@ struct RemoveUnusedModuleElements : public Pass {
...
+    });
+    module->removeMemories([&](Memory* curr) {
+      return !needed({ModuleElementKind::Memory, curr->name});
+    });
+    module->removeTables([&](Table* curr) {
+      return !needed({ModuleElementKind::Table, curr->name});
+    });
+    module->removeDataSegments([&](DataSegment* curr) {
+      return !needed({ModuleElementKind::DataSegment, curr->name});
+    });
```
kripken, line 766: Looks like the first and last 3 here are duplicates? The first three may have been intended to be deleted?

src/passes/RemoveUnusedModuleElements.cpp
tlively: Even if the module never uses the imports, the fact that the segments are written to the imported table or memory during initialization is externally visible, so we have to make sure it still happens after the optimization.

kripken: Oh, good point...

kripken (approved):
lgtm minus that one 3/3 fix.

tlively:
The fuzzer found a bug where removing segments produces invalid modules because the segment indices in various instructions aren't updated. I've wanted to replace the indexes with names to match all other module elements for a long time, and now I finally have a reason to do that 🎉

tlively:
This depends on #5618 to resolve that fuzzer bug.

tlively:
Rebased and fuzzing again.

PR #5608, opened 2023-03-31
[Wasm GC] Fix nondeterminism in SignaturePruning

kripken:
The order of iteration on `sigFuncs` there can matter, as each time we
prune we end up changing things that can affect other prunings (specifically,
`ParamUtils::removeParameter` can decide that it can't remove a parameter
based on the effects on arguments, and current limitations in how we handle
that; and pruning can affect effects).

### DIFF
--- a/src/passes/SignaturePruning.cpp
+++ b/src/passes/SignaturePruning.cpp
@@ -35,6 +35,7 @@
 #include "ir/type-updating.h"
 #include "param-utils.h"
 #include "pass.h"
+#include "support/insert_ordered.h"
 #include "support/sorted_vector.h"
 #include "wasm-type.h"
 #include "wasm.h"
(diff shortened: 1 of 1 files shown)


tlively (approved):
LGTM, although the nondeterminism here is (at least partially) a symptom of unrealized optimization potential.

If we made sure to replace constant parameters in callers before callees, we would be able to propagate those constants more frequently where they are passed unmodified to the callees. A "naive" approach would be to do a topological sort and break cycles arbitrarily, but the choice of how to break cycles could still be a source of non-determinism and missed optimizations. A complete solution would do a coinductive largest fixed point analysis on the recursive definition of constant parameters instead.

Maybe worth a TODO? 😆

kripken:
Might be worth improving, yeah. Reading the code more carefully, I found this:

https://github.com/WebAssembly/binaryen/blob/dba0a08963d0452a773b9abbdd789503945c5f43/src/passes/param-utils.cpp#L73-L76

So the issue is unreachability, really, and can occur if DCE has not been run recently enough. I'm not sure that's worth a TODO? We usually just leave stuff like that.

kripken:
Landing to try to fix j2wasm's nondeterminism issue. Can continue to discuss TODOs and followups in parallel.

kripken:
Ah, good point, that's entirely separate. I opened #5611

PR #5609, opened 2023-03-31
Switch fuzzer to the hybrid type system

kripken:
This is the default, and also used by J2Wasm.

Fuzzing for a day with this I found no issues.


tlively (approved):
🎉

PR #5611, opened 2023-03-31
Add a TODO in SignaturePruning

kripken:


src/passes/SignaturePruning.cpp
```diff
@@ -163,6 +163,13 @@ struct SignaturePruning : public Pass {
     SubTypes subTypes(*module);
 
     // Find parameters to prune.
+    //
+    // TODO: The order matters here, and more than one cycle can find more work
+    //       in some cases, as finding a parameter is a constant and removing it
+    //       can lead to another call (that receives that parameter's value) to
+    //       now have constant parameters as well, and so it becomes
+    //       optimizable. We could do a topological sort here (but if this is
```
tlively, line 171: topological sort or greatest fixed point analysis! I bet there are a nonzero number of cases where the same constant is propagated through a recursive call sequence.

tlively (approved):

PR #5613, opened 2023-03-31
[NFC] Remove our bespoke `make_unique` implementation

tlively:
This code predates our adoption of C++14 and can now be removed in favor of
`std::make_unique`, which should be more efficient.

PR #5615, opened 2023-04-03
Fix a call to make_unique without std

kripken:
Followup to #5613

Not sure how this builds for us, but it breaks on the J2Wasm roll on `make_unique` not being declared.


tlively (approved):
Weird!

PR #5616, opened 2023-04-03
[Wasm GC] Avoid refining in TypeUpdating unreachability propagation

kripken:
That code should only propagate unreachability, and not refine. If it refines when
we call `finalize()` then other code around it might end up invalid (as it could be
partially refined).

### DIFF
--- a/src/ir/type-updating.h
+++ b/src/ir/type-updating.h
@@ -248,14 +248,22 @@ struct TypeUpdater
           return; // did not turn
         }
       } else if (auto* iff = curr->dynCast<If>()) {
-        // may not be unreachable if just one side is
+        // We only want to change a concrete type to unreachable here, so undo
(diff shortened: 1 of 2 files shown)


tlively (commented):
It seems like it would be much simpler and harder to break in the future if we just ran refinalize here. Is that not a good option?

kripken:
Refinalizing here would mean we need to refinalize throughout `TypeUpdating`, which would be a larger refactoring, unfortunately, and this code is somewhat tricky to get right (since unreachability is tricky). I do think it's unlikely we'll have more changes here for this reason, at least, as the new code applies a generic way of only updating unreachability (undoing non-unreachability changes), that is, the new code is not specific to the internals of If or Try.

tlively (approved):

PR #5617, opened 2023-04-03
[Wasm GC] Fuzz RefCast

kripken:
Depends on #5616 which was uncovered by this fuzzing.


src/tools/fuzzing/fuzzing.cpp
```diff
@@ -3418,23 +3462,54 @@ Type TranslateToFuzzReader::getSubType(Type type) {
...
 }
 
+Nullability TranslateToFuzzReader::getSuperType(Nullability nullability) {
+  if (nullability == Nullable) {
+    return Nullable;
+  }
+  return getNullability();
+}
+
+HeapType TranslateToFuzzReader::getSuperType(HeapType type) {
```
tlively, line 3487: This could also include basic supertypes like `eq` and `any`, etc.

tlively (approved):

src/tools/fuzzing/fuzzing.cpp
kripken: I think those already get enough representation from `getReferenceType()` and friends. Though I agree it's not ideal to not do it here, as it seems natural to add them. But the bigger picture I think is that it could be nice to refactor so that `HeapType::getSuperType()` returns basic types too, which would make this just work as expected. That's a longer-term refactoring though as code depends on it not doing that atm...

tlively: `ref.cast` of a basic heap type is sufficiently different and interesting compared to `ref.cast` of user defined types that it seems worth fuzzing, even basic heap types are well-represented elsewhere already.

PR #5618, opened 2023-04-04
Use Names instead of indices to identify segments

tlively:
All top-level Module elements are identified and referred to by Name, but for
historical reasons element and data segments were referred to by index instead.
Fix this inconsistency by using Names to refer to segments from expressions that
use them. Also parse and print segment names like we do for other elements.

The C API is partially converted to use names instead of indices, but there are
still many functions that refer to data segments by index. Finishing the
conversion can be done in the future once it becomes necessary.


tlively:
* main
  * **PR #5618** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5618" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5618?utm_source=stack-comment).

src/passes/MemoryPacking.cpp
```diff
@@ -596,25 +595,11 @@ void MemoryPacking::createSplitSegments(
...
-        } else if (auto* curr = referrer->dynCast<DataDrop>()) {
-          curr->segment = segmentIndex;
-        } else if (auto* curr = referrer->dynCast<ArrayNewSeg>()) {
-          curr->segment = segmentIndex;
-        } else {
-          WASM_UNREACHABLE("Unexpected segment operation");
-        }
-        return referrer;
-      };
-    }
```
kripken, line 617: Nice this logic gets simpler with Names!

src/wasm/wasm-stack.cpp
```diff
@@ -687,26 +687,27 @@ void BinaryInstWriter::visitSIMDLoadStoreLane(SIMDLoadStoreLane* curr) {
...
+  o << U32LEB(parent.getDataSegmentIndex(curr->segment));
 }
 
 void BinaryInstWriter::visitMemoryCopy(MemoryCopy* curr) {
   o << int8_t(BinaryConsts::MiscPrefix);
   o << U32LEB(BinaryConsts::MemoryCopy);
-  o << int8_t(parent.getMemoryIndex(curr->destMemory))
-    << int8_t(parent.getMemoryIndex(curr->sourceMemory));
+  o << U32LEB(parent.getMemoryIndex(curr->destMemory));
+  o << U32LEB(parent.getMemoryIndex(curr->sourceMemory));
```
kripken, line 704: drive-bys? lgtm

kripken (approved):
Nice!

Heh, the changes to the `limit-segments` test are so big they can't be viewed on github even by clicking to expand...

lgtm with fuzzing.

src/wasm/wasm-stack.cpp
tlively: Yep 👍

tlively:
Ran 7.5k iterations on the fuzzer, so landing this before we end up with merge conflicts.

PR #5621, opened 2023-04-04
[Wasm GC] Fix CoalesceLocals ineffective tee removal with type changes

kripken:
When we remove a tee there, we must not change the type at all.

Must land after #5619


tlively (approved):
These blocks will eventually get optimized out, right?

kripken:
Yes, Vacuum removes such blocks.

PR #5622, opened 2023-04-04
[Wasm GC] Fuzz struct.new and array.new

kripken:
Repurpose `makeBasicRef, makeCompoundRef` to generate not just "constant"
refs but any reference, and use those to create StructNew/ArrayNew.

The key changes are to add `makeCompoundRef` to `make()`, and to make
the function call `make()` for children, where possible, instead of just
`makeTrivial()`.

### DIFF
--- a/src/tools/fuzzing/fuzzing.cpp
+++ b/src/tools/fuzzing/fuzzing.cpp
@@ -2279,7 +2279,11 @@ Expression* TranslateToFuzzReader::makeCompoundRef(Type type) {
       // here. This nesting makes it much less likely for a recursive data
       // structure to end up as a massive tree of struct.news, since the nesting
       // limitation code at the top of this function will kick in.
-      nester.add(values.size());
+      if (!values.empty()) {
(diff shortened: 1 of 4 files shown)


src/tools/fuzzing/fuzzing.cpp
```diff
@@ -3176,13 +3196,6 @@ Expression* TranslateToFuzzReader::makeRefCast(Type type) {
   return builder.makeRefCast(make(refType), type, RefCast::Safe);
 }
 
-Expression* TranslateToFuzzReader::makeI31New(Type type) {
```
tlively, line 3179: Was this never called?

tlively (approved):

src/tools/fuzzing/fuzzing.cpp
kripken: It was called by being taken by address in `make()`. But it is no longer needed, since now we add `make*Ref()` there, and `i31` is just one of the cases handled by those methods.

Issue #5595, opened 2023-03-22
setMemory fails for large data after update to v112


mnater:
Hi

I'm using binaryen.js to prepopulate memory with a larger amount of data. Since the update to v112 this doesn't work anymore.
Reduced testcase:
````javascript
import binaryen from "binaryen";

const dataLength = 100_000;
const mockData = new Uint8Array(dataLength);
for (let i = mockData.length - 1; i >= 0; i--) {
  mockData[i] = i & 255;
}

// Create a module and prepopulate memory
const myModule = new binaryen.Module();

myModule.setMemory(2, -1, "mem", [{
  "data": mockData,
  "offset": myModule.i32.const(0),
  "passive": false
}]);


const binary = myModule.emitBinary();
console.log(binary.length); // emits 100034 on v111
````
In v112 I get `Bus error: 10`
(Node v19.6.0 on macOS 12.6.3 )

The maximum `dataLength` where the code above still works in v112 is `67_312`.

I also tried to split data in several chunks but I hit the same issues:

````javascript
import binaryen from "binaryen";

const dataLength = 100_000;
const mockData = new Uint8Array(dataLength);
for (let i = mockData.length - 1; i >= 0; i--) {
  mockData[i] = i & 255;
}

// Create a module and prepopulate memory
const myModule = new binaryen.Module();

const memSegments = [];
const segmentSize = 50_000
const segmentCount = Math.ceil(dataLength / segmentSize);
for (let i = 0; i < segmentCount; i += 1) {
    memSegments.push({
        "data": mockData.subarray(i * segmentSize, ((i + 1) * segmentSize)),
        "offset": myModule.i32.const(i * segmentSize),
        "passive": false
    });
}

console.log(memSegments);

myModule.setMemory(
    2,
    -1,
    "mem",
    memSegments
);

const binary = myModule.emitBinary();
console.log(binary.length);
````

I think this is due the change from `allocate()` to `stackAlloc()` in https://github.com/WebAssembly/binaryen/blame/main/src/js/binaryen.js-post.js#L2572

How can I use `setMemory` with a larger amount of data?

kripken:
Hmm, those should both allocate on the stack, but it's possible that an emscripten change to the default stack size affected this. To test that, building binaryen.js with `-sSTACK_SIZE=8MB` (added to CMakeLists along the other such flags) should fix it, if that's right.

mnater:
[...]

PR #5625, opened 2023-04-04
fix: use _malloc/_free to allocate memory in setMemory

mnater:
see issue #5595


CMakeLists.txt
```diff
@@ -27,6 +27,12 @@ if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE "Release")
 endif()
 
+# Optionally change stack size (e.g.: emcmake cmake . -Dbinaryen_STACK_SIZE=8388608)
+if(NOT binaryen_STACK_SIZE)
```
kripken, line 31: Looks like the other options are in the form `BYN_` so please make this `BYN_STACK_SIZE`. Or, as this is only for binaryen.js/wasm, perhaps `BYN_WASM_STACK_SIZE`.

kripken (commented):
malloc/free looks good!

Please remove the other changes, though - I don't think we need to make the stack size customizable now. (And it has a downside as it bakes 64K as the default, rather than deferring to emscripten to set the default.)

mnater:
I don't know a clean way to remove the older commits. 🤷‍♂️
I'll close this PR and send a new one.

mnater:
#5646

PR #5626, opened 2023-04-04
Avoid imported memories in the fuzzer

kripken:
We do similar things for globals already.

This fixes a crash that started to happen after recent landings added
tests with imported memories, but the problem is long-standing.

### DIFF
--- a/src/tools/fuzzing/fuzzing.cpp
+++ b/src/tools/fuzzing/fuzzing.cpp
@@ -415,8 +415,7 @@ void TranslateToFuzzReader::finalizeMemory() {
       memory->initial,
       Address((maxOffset + Memory::kPageSize - 1) / Memory::kPageSize));
   }
-  memory->initial =
-    std::max(memory->initial, USABLE_MEMORY);
+  memory->initial = std::max(memory->initial, USABLE_MEMORY);
   // Avoid an unlimited memory size, which would make fuzzing very difficult
   // as different VMs will run out of system memory in different ways.
   if (memory->max == Memory::kUnlimitedSize) {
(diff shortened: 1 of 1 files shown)


kripken:
Actually the issue was multi-memory: we already did this for the first memory, and just needed to loop.

Also clean up that code while I'm around, to avoid repeating `wasm.memories[0]` all the time.

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -196,6 +196,7 @@ void TranslateToFuzzReader::build() {
 void TranslateToFuzzReader::setupMemory() {
   // Add memory itself
   MemoryUtils::ensureExists(&wasm);
+  auto& memory = wasm.memories[0];
```
tlively, line 199: Would it make sense to stop special-casing memory 0 entirely and treat memories uniformly?

kripken: Perhaps, but I'm not sure we want full uniformity. We need some special conditions on the memories we use, like being of a reasonable known size (so we don't constantly have traps). It might actually be good that we just modify one memory atm to that shape and not all of them, but it does mean our fuzzing of the others is limited in other ways.

For imports, though, we just need to modify them all, or else we immediately trap on any use.

tlively: Even if we want to occasionally be more permissive, there's no reason memory 0 should be special. I'm concerned that treating it specially in the fuzzer will mask bugs where we inappropriately treat it as special elsewhere.

kripken: Agreed, yeah, it has downsides. Worth considering separately from this PR I think (which does not make us more specific to memory 0).

tlively (approved):
Makes sense to address uniformity separately. LGTM!

PR #5631, opened 2023-04-05
Fuzzer: Improve mutate()

kripken:
1. Pick a subtype when mutating.
2. Don't use a fixed 10% chance to mutate, but pick a mutation rate in each function.

### DIFF
--- a/src/tools/fuzzing/fuzzing.cpp
+++ b/src/tools/fuzzing/fuzzing.cpp
@@ -802,12 +802,14 @@ void TranslateToFuzzReader::mutate(Function* func) {
   // average. To achieve that, we raise r/100, which is in the range [0, 1], to
   // the 9th power, giving us a number also in the range [0, 1] with a mean of
   //   \integral_0^1 t^9 dx = 0.1 * t^10 |_0^1 = 0.1
+  // As a result, we get a value in the range of 0-100%. We'll adjust 0% to 1%
+  // so we avoid doing nothing at all, and note that 100% is ok since we can't
+  // replace everything anyhow (see below).
   double t = r;
   t = t / 100;
   t = t * t * t * t * t * t * t * t * t;
   Index percentChance = t * 100;
   if (percentChance == 0) {
-    // Give at least a 1% chance, or else we'll be doing nothing here.
     percentChance = 1;
   }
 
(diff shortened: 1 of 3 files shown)


src/tools/fuzzing/fuzzing.cpp
```diff
@@ -786,30 +786,54 @@ void TranslateToFuzzReader::recombine(Function* func) {
...
+  // Prefer lower numbers: We want something like a 10% chance to mutate on
+  // average. To achieve that, we raise r/100, which is in the range [0, 1], to
+  // the 9th power, giving us a number also in the range [0, 1] with a mean of
+  //   \integral_0^1 t^9 dx = 0.1 * t^10 |_0^1 = 0.1
+  // As a result, we get a value in the range of 0-100%. We'll adjust 0% to 1%
+  // so we avoid doing nothing at all, and note that 100% is ok since we can't
+  // replace everything anyhow (see below).
+  double t = r;
+  t = t / 100;
+  t = t * t * t * t * t * t * t * t * t;
```
tlively, line 810: Why not `pow(t, 9)`?

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -823,13 +847,15 @@ void TranslateToFuzzReader::mutate(Function* func) {
...
         // not, changing an offset, etc.
         // Perform a general replacement. (This is not always valid due to
-        // nesting of labels, but we'll fix that up later.)
-        // TODO: pick a subtype of the current type
-        replaceCurrent(parent.make(curr->type));
+        // nesting of labels, but we'll fix that up later.) We both pick a
+        // possible subtype here, and then pick something of that type, so that
+        // we should give a possibility for any valid replacement here to
+        // appear.
+        replaceCurrent(parent.make(parent.getSubType(curr->type)));
```
tlively, line 854: The first thing `make` does is get a subtype, so we don't need to do it here.

tlively (approved):
LGTM % comments.

src/tools/fuzzing/fuzzing.cpp
kripken: Very good point, I missed that... removed this part.

src/tools/fuzzing/fuzzing.cpp
kripken: Also good point, done.

tlively (approved):

PR #5636, opened 2023-04-06
Allow vector additions to be optimized

### LINKED ISSUE #5633: SIMD add not being optimized away
The `i64x2.add` in the following WAT isn't being optimized at all (`build/bin/wasm-opt --enable-simd -O3 -S -o - test.wat`):

```wat
(module
    (export "v128-add" (func $v128-add))
    (export "i64-add" (func $i64-add))
    (func $v128-add (param v128) (result v128)
        (i64x2.add (local.get 0) (v128.const i64x2 0 0))
    )
    (func $i64-add (param i64) (result i64)
        (i64.add (local.get 0) (i64.const 0))
    )
)
```

It outputs:

```wat
(module
 (type $v128_=>_v128 (func (param v128) (result v128)))
 (type $i64_=>_i64 (func (param i64) (result i64)))
 (export "v128-add" (func $v128-add))
 (export "i64-add" (func $i64-add))
 (func $v128-add (; has Stack IR ;) (param $0 v128) (result v128)
  (i64x2.add
   (local.get $0)
   (v128.const i32x4 0x00000000 0x00000000 0x00000000 0x00000000)
  )
 )
 (func $i64-add (; has Stack IR ;) (param $0 i64) (result i64)
  (local.get $0)
 )
)
```

I'd have expected that add to be optimized away into `(local.get $0)`. How do I get that to work?

CountBleck:
This change allows optimizeAddedConstants to work on vector additions and subtractions. There were no other changes. I don't know how what I did worked, or why it worked without additional changes, but it worked.

Fixes #5633.

### DIFF
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -669,7 +669,11 @@ struct OptimizeInstructions
       // note that both left and right may be consts, but then we let
       // precompute compute the constant result
     } else if (curr->op == AddInt32 || curr->op == AddInt64 ||
-               curr->op == SubInt32 || curr->op == SubInt64) {
+               curr->op == SubInt32 || curr->op == SubInt64 ||
+               curr->op == AddVecI8x16 || curr->op == AddVecI16x8 ||
+               curr->op == SubVecI8x16 || curr->op == SubVecI16x8 ||
+               curr->op == AddVecI32x4 || curr->op == AddVecI64x2 ||
+               curr->op == SubVecI32x4 || curr->op == SubVecI64x2) {
       if (auto* ret = optimizeAddedConstants(curr)) {
         return replaceCurrent(ret);
       }
@@ -2769,7 +2773,7 @@ private:
   // and combine them note that we ignore division/shift-right, as rounding
   // makes this nonlinear, so not a valid opt
   Expression* optimizeAddedConstants(Binary* binary) {
-    assert(binary->type.isInteger());
+    assert(binary->type.isInteger() || binary->type.isVector());
 
     uint64_t constant = 0;
     std::vector<Const*> constants;


src/passes/OptimizeInstructions.cpp
```diff
@@ -2769,7 +2773,7 @@ struct OptimizeInstructions
   // and combine them note that we ignore division/shift-right, as rounding
   // makes this nonlinear, so not a valid opt
   Expression* optimizeAddedConstants(Binary* binary) {
-    assert(binary->type.isInteger());
+    assert(binary->type.isInteger() || binary->type.isVector());
 
     uint64_t constant = 0;
```
kripken, line 2778: I think this needs to be generalized. This is the constant value that is added up, and 64 bits is enough for i32 and i64 but not v128. Instead, this can use a `Literal` which abstracts over all constants.

And for testing, see for example `test/lit/passes/optimize-instructions-strings.wast` - please create a parallel one for `simd` instead of `strings`.

CountBleck: How do I generalize that? I can make `constant` and `mul` `Literal`s and update everything accordingly, but that doesn't seem to work properly. It seems that v128 `Literal`s can't be added or multiplied with other `Literal`s, which makes sense.

kripken: I think we just need to add vector Literals with other vector Literals? That is, we'd be gathering Literals all of the same type in all cases (all i32, or all i64, or all v128), and we need to make Literals of the right type for shifts etc.

CountBleck: How do I add two v128 Literals together if the type of add (`i32x4`, `f64x2`, etc.) can vary? I'm unfamiliar with Binaryen in general, so I don't know any tricks that I could use.

kripken: Ah, yes, you'd need to make the adds use the right operation. So right now where it does `Abstract::getBinary(binary->type, Abstract::Add)`, which abstracts over `i32` and `i64`, I guess we'd need to look for SIMD operations, and handle each of them there (using the [right method](https://github.com/WebAssembly/binaryen/blob/18898347e7c4ceb86988bdb8594160883f7886f0/src/literal.h#L531-L533). But maybe there's an even better way somehow, using the interpreter.

@tlively I'm guessing there is some clever abstract interpretation solution to this..?

tlively: Yes ;) It would be similar to equipping precompute-propagate with the ability to do (some) algebra, though, so not a tiny project, even with an abstract interpretation framework in place.

PR #5637, opened 2023-04-06
Implement array.fill, array.init_data, and array.init_elem

tlively:
These complement array.copy, which we already supported, as an initial complete
set of bulk array operations. Replace the WIP spec tests with the upstream spec
tests, lightly edited for compatibility with Binaryen.


tlively:
* main
  * **PR #5637** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5637" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5637?utm_source=stack-comment).

src/ir/possible-contents.cpp
```diff
@@ -988,7 +988,8 @@ struct InfoCollector
     auto* set = builder.makeArraySet(curr->destRef, curr->destIndex, get);
     visitArraySet(set);
   }
-
+  void visitArrayFill(ArrayFill* curr) { WASM_UNREACHABLE("unimplemented"); }
+  void visitArrayInit(ArrayInit* curr) { WASM_UNREACHABLE("uimplemented"); }
```
kripken, line 992: The fuzzer will crash on these (any testcase with the instructions + `--gufa`).

I think their implementation can be similar to `ArraySet`.

src/wasm-interpreter.h
```diff
@@ -3576,6 +3625,77 @@ class ModuleRunnerBase : public ExpressionRunner<SubType> {
...
+    switch (curr->op) {
+      case InitData: {
+        auto* seg = wasm.getDataSegment(curr->segment);
+        auto elem = curr->ref->type.getHeapType().getArray().element;
+        size_t elemSize = elem.getByteSize();
+        uint64_t readSize = (uint64_t)sizeVal * elemSize;
+        if (offsetVal + readSize > seg->data.size()) {
+          trap("out of bounds segment access in array.init_data");
+        }
+        if (offsetVal + sizeVal > 0 && droppedSegments.count(curr->segment)) {
```
kripken, line 3670: Should `offsetVal + sizeVal` be just `sizeVal`? If the size is 0, it seems like the offset should not matter.

src/wasm-interpreter.h
```diff
@@ -3576,6 +3625,77 @@ class ModuleRunnerBase : public ExpressionRunner<SubType> {
...
+        if ((uint64_t)offsetVal + sizeVal > seg->data.size()) {
+          trap("out of bounds segment access in array.init");
+        }
+        // TODO: Check whether the segment has been dropped once we support
+        // dropping element segments.
+        for (size_t i = 0; i < sizeVal; i++) {
+          // TODO: This is not correct because it does not preserve the identity
+          // of references in the table! ArrayNewSeg suffers the same problem.
+          // Fixing it will require changing how we represent segments, at least
+          // in the interpreter.
```
kripken, line 3690: To make sure I follow is the issue something like a `struct.new` in a table, that would be re-evaluated each time? While `ref.func` is fine?

test/lit/wat-kitchen-sink.wast
```diff
@@ -1627,6 +1629,22 @@
...
+ ;; CHECK-NEXT:   (local.get $3)
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT: )
+ (func $array-fill (param (ref $a2) i32 f32 i32)
+  local.get 0
+  local.get 1
+  local.get 2
+  local.get 3
+  array.fill $a2
+ )
```
kripken, line 1646: just `fill` for now and not `init*`?

kripken (approved):
lgtm with comments addressed + fuzzing

src/wasm-interpreter.h
tlively: No, dropped segments semantically have their length truncated to zero, so if offset and size are both zero, there's no trap, but if either is nonzero there will be a trap. This could perhaps be more clearly expressed, but this code is copied from the implementation of `MemoryInit`.

src/wasm-interpreter.h
tlively: Yes, exactly.

test/lit/wat-kitchen-sink.wast
tlively: We don't parse element segment references yet, so I left `ArrayInit` as future work.

src/wasm-interpreter.h
kripken: I see. Maybe my confusion is about the spec then. It seems odd to trap on trying to load 0 items from offset 17 but not to trap on trying to load 0 items from offset 0 (since both do no actual reads).

src/ir/possible-contents.cpp
tlively: What should I do to model reading from the data or element segment in `ArrayInit`?

kripken: Looks like `TableGet` just models the value as a root (an unknown value that could be anything). I think we should do the same here. But that value would need to be written into the array's storage instead of just declaring this expression to be a root.

If it doesn't look obvious how to do that I can look into it. I'm not sure if it will be trivial or not.

tlively: Thanks. I'm not sure how to hook up the root value with the synthetic write, so I'll leave that to you and look forward to seeing what's involved.

src/wasm-interpreter.h
tlively: It's about keeping the bounds checks uniform in the spec. For the spec, the only rule is "trap if offset + length is out of bounds." Since there are no special cases for zero sized accesses, you get the  somewhat counterintuitive behavior of trapping on out-of-bounds but not in-bounds or at-bounds zero-sized accesses. Combine that with the bounds changing for dropped segments, and you get some pretty weird behavior for anything that doesn't model things exactly like the spec.

src/ir/possible-contents.cpp
tlively: My fuzzing did find this after 7591 iterations, so at least it's rare.

kripken: Ah, I was hoping to fix this before it lands, as that frequency still means it'll be found many times a day. But no worries, I'll look into this right now.

tlively: Oops, sorry about that.

kripken: No problem, I think I found the solution already, PR is up (#5638)

kripken:
Fuzzer found a bug:
```wat
(module
 (type $[mut:i32] (array (mut i32)))
 (type $ref|[mut:i32]|_i32_i32_i32_=>_none (func (param (ref $[mut:i32]) i32 i32 i32)))
 (global $global$0 (mut i32) (i32.const -69))
 (global $global$1 (mut i32) (i32.const 100))
 (memory $0 (shared 16 17))
 (data $0 "")
 (data $1 (i32.const 0) "")
 (data $2 (i32.const 14) "")
 (data $3 (i32.const 30) "")
 (func $0 (param $0 (ref $[mut:i32])) (param $1 i32) (param $2 i32) (param $3 i32)
  (array.init_data $[mut:i32] $0
   (local.get $0)
   (i32.const 0)
   (i32.const 0)
   (i32.const 1)
  )
 )
)
```
```
$ bin/wasm-opt w.wasm -all --memory-packing
[wasm-validator error in function 0] unexpected false: array.init_data segment must exist, on 
(array.init_data $[mut:i32] $0
 (local.get $0)
 (i32.const 0)
 (i32.const 0)
 (i32.const 1)
)
Fatal: error after opts
```

PR #5638, opened 2023-04-06
[Wasm GC] Fix GUFA on ArrayInit 

kripken:
Looks like it is fairly simple to do, but it requires some hacks...


tlively (approved):

tlively:
Thanks! I wouldn't have thought to use a fake local.

PR #5639, opened 2023-04-06
[GUFA] Refine global types during flow

kripken:
Previously `(ref.as_non_null (global.get ..))` would return the global with no changes,
and if the global was nullable then the type didn't match the output, which hit an
assertion (where GUFA checks that the contents match the declared type in the wasm).

To fix this, refine global types, that is, the type we track on GlobalInfo may be more
refined than the global itself. In the above example, if the global is nullable then
the GlobalInfo would point to that global but have a non-nullable type.

In fact the code was already prepared for this, and few changes were needed.

### DIFF
--- a/src/ir/possible-contents.cpp
+++ b/src/ir/possible-contents.cpp
@@ -202,13 +202,19 @@ void PossibleContents::intersectWithFullCone(const PossibleContents& other) {
     // or immutable global is not removed by intersection, if the type is in the
     // cone we are intersecting with.
     if (isSubType) {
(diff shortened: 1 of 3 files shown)


tlively (approved):
Is it worth adding a .wast lit test demonstrating the result of this extra refinement?

kripken:
Hmm, I tried using `ref.eq`, since global identity could show this, but the problem is GUFA only optimizes the not-equal case of `ref.eq` atm. So I'm not sure how to do it. But the unit test checks the issue directly, at least.

PR #5640, opened 2023-04-06
[Wasm GC] Handle packed fields in GUFA and CFP

kripken:
The same bug was present in both: We ignored packing, so writing a larger
value than fits in the field would lead to us propagating that original value.


kripken:
(oops, I missed there is already `Field::getByteSize()`. refactored to use that)

tlively (approved):

PR #5643, opened 2023-04-07
[NFC] Use the new getField() in the heap type fuzzer

kripken:


src/tools/fuzzing/heap-types.cpp
```diff
@@ -712,21 +713,13 @@ struct Inhabitator {
...
-  }
-  if (type.isArray()) {
-    if (type.getArray().element.mutable_ == Mutable) {
-      return Invariant;
-    } else {
-      return Covariant;
-    }
+  auto field = GCTypeUtils::getField(type, idx);
+  assert(field);
+  if (field->.mutable_ == Mutable) {
```
tlively, line 718: Whoa, I've never seen `->.` before. I'm surprised that compiles. Does it work if you just do `field->mutable_`?

tlively (approved):

src/tools/fuzzing/heap-types.cpp
kripken: It didn't compile... somehow I forgot to check that this compiles before pushing it :smile:

kripken: (should be ok now)

tlively (approved):

PR #5646, opened 2023-04-07
use malloc/free instead of stackAlloc in setMemory (cleaned up)

### LINKED ISSUE #5595: setMemory fails for large data after update to v112

mnater:
cleaned PR. Original: #5625

fixes #5595

### DIFF
--- a/src/js/binaryen.js-post.js
+++ b/src/js/binaryen.js-post.js
@@ -2569,13 +2569,13 @@ function wrapModule(module, self = {}) {
       const segmentOffset = new Array(segmentsLen);
       for (let i = 0; i < segmentsLen; i++) {
         const { data, offset, passive } = segments[i];
-        segmentData[i] = stackAlloc(data.length);
+        segmentData[i] = _malloc(data.length);
         HEAP8.set(data, segmentData[i]);
         segmentDataLen[i] = data.length;
         segmentPassive[i] = passive;
         segmentOffset[i] = offset;
       }
-      return Module['_BinaryenSetMemory'](
+      const ret = Module['_BinaryenSetMemory'](
         module, initial, maximum, strToStack(exportName),
         i32sToStack(segmentData),
         i8sToStack(segmentPassive),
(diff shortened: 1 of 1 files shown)


kripken (approved):
Thanks!

PR #5651, opened 2023-04-10
[Wasm GC] Fuzz struct.get and array.get

kripken:
Fuzzing with this locally has led to many of the recent fixes.


kripken:
It's tricky to show this output since it's random. Perhaps we could write a script that searches the space of random inputs until it finds a small case that shows it... or we could refactor the code to allow it to be generated manually too.

PR #5652, opened 2023-04-10
[GUFA] Fix packed field filtering

kripken:
Technically we need to filter both before and after combining, that is,
if a location's contents will be filtered by `F()` then if the new contents
are `x` and old contents `y` then we need to end up with
`F(F(x) U F(y))`. That is, filtering before is necessary to ensure the union
of content does not end up unnecessarily large, and the filtering
after is necessary to ensure the final result is properly filtered to fit.
(If our representation were perfect then this would not be needed, but
it is not, as the union of two exact types can end up as a very large
cone, for example.)

For efficiency we have been filtering afterwards. But that is not enough
for packed fields, it turns out, where we must filter before. If we don't,
then if inputs `0` and `0x100` arrive to an `i8` field then combining them
we get "unknown integer" (which is then filtered by `0xff`, but it's too
late). By filtering before, the actual values are both `0` and we end up
with that as the only possible value.

it turns out that filtering before is enough for such fields, so do only
that.

### DIFF
--- a/src/ir/possible-contents.cpp
+++ b/src/ir/possible-contents.cpp
@@ -1695,6 +1695,33 @@ bool Flower::updateContents(LocationIndex locationIndex,
   std::cout << '\n';
 #endif
 
+  auto location = getLocation(locationIndex);
+
+  // Handle special cases: Some locations can only contain certain contents, so
+  // filter accordingly. In principle we need to filter both before and after
+  // combining with existing content; filtering afterwards is obviously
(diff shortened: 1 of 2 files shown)


tlively:
I think I need to sit down and read this code more closely sometime because my mental model for how I expect GUFA to work does not involve any filtering whatsoever :(

tlively (approved):

kripken:
(Happy to meet up and discuss anything that isn't clear!)

PR #5655, opened 2023-04-11
[Wasm GC] Fuzz struct.set and array.set

kripken:
Example output of the `array.set` length check:
```wat
          (if
           (i32.lt_u
            (local.tee $8
             (select
              (i32.const -65536)
              (ref.is_null
               (array.new_fixed $[i8])
              )
              (local.get $4)
             )
            )
            (array.len
             (local.tee $14
              (array.new $[mut:ref?|{i32_mut:ref|i31|_mut:i16_mut:f32}|]
               (ref.null none)
               (i32.const 88)
              )
             )
            )
           )
           (array.set $[mut:ref?|{i32_mut:ref|i31|_mut:i16_mut:f32}|]
            (local.get $14)
            (local.get $8)
            (ref.null none)
           )
          )
```
Locals `$14, $8` store the ref and the index, which are each used twice.

### DIFF
--- a/src/tools/fuzzing.h
+++ b/src/tools/fuzzing.h
@@ -334,7 +334,9 @@ private:
   Expression* makeRefTest(Type type);
   Expression* makeRefCast(Type type);
   Expression* makeStructGet(Type type);
+  Expression* makeStructSet(Type type);
   Expression* makeArrayGet(Type type);
+  Expression* makeArraySet(Type type);
   Expression* makeI31Get(Type type);
   Expression* makeMemoryInit();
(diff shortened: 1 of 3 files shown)


tlively (approved):
LGTM, but would it make sense to factor out some of the bounds checking shared between array.get and array.set?

kripken:
A few lines of code could be shared, yeah. The problem is that the helper function would need multiple out params (or in params) so it feels kind of clunky just to save a few lines. If i run into the same thing with more instructions though then I might try again.

PR #5657, opened 2023-04-12
Fuzzer: When nested under makeTrivial(), avoid normal make()

kripken:
Without this, in certain complex operations we could end up calling a nested
`make()` operation that included nontrivial things, which could cause problems.
The specific problem I encountered was in `fixAfterChanges()` we tried to fix up
a duplicate label, but calling `makeTrivial()` emitted something very large that
happened to include a new block with a new label nested under a `struct.get`,
and that block's label conflicted with a label we'd already processed.

### DIFF
--- a/src/tools/fuzzing/fuzzing.cpp
+++ b/src/tools/fuzzing/fuzzing.cpp
@@ -1048,6 +1048,10 @@ void TranslateToFuzzReader::addInvocations(Function* func) {
 
 Expression* TranslateToFuzzReader::make(Type type) {
   auto subtype = getSubType(type);
+  if (trivialNesting) {
+    // We are nested under a makeTrivial call, so only emit something trivial.
+    return makeTrivial(type);
+  }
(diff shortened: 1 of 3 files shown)


src/tools/fuzzing/fuzzing.cpp
```diff
@@ -1212,6 +1216,14 @@ Expression* TranslateToFuzzReader::_makeunreachable() {
 }
 
 Expression* TranslateToFuzzReader::makeTrivial(Type type) {
+  struct TrivialNester {
```
tlively, line 1219: This is an unfortunate amount of boilerplate just to run `trivialNesting--` to run at scope exit, but I couldn't find a better way to do it outside of boost.

tlively (approved):

src/tools/fuzzing/fuzzing.cpp
kripken: Yeah... other languages can do better, e.g. `defer` in Go and Zig.

PR #5659, opened 2023-04-12
[NFC] Refactor fuzzer array check logic

kripken:
Turns out we'll need array checks for other array operations
like `array.copy`, so it's worth the refactor.

### DIFF
--- a/src/tools/fuzzing/fuzzing.cpp
+++ b/src/tools/fuzzing/fuzzing.cpp
@@ -3287,7 +3287,10 @@ Expression* TranslateToFuzzReader::makeStructSet(Type type) {
   return builder.makeStructSet(fieldIndex, ref, value);
 }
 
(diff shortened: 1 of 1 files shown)


src/tools/fuzzing/fuzzing.cpp
```diff
@@ -3287,6 +3287,30 @@ Expression* TranslateToFuzzReader::makeStructSet(Type type) {
   return builder.makeStructSet(fieldIndex, ref, value);
 }
 
+static auto makeArrayBoundsCheck(Expression* ref,
```
tlively, line 3290: TIL you can use auto like this and even return and use unnameable types. Neat!

tlively (approved):

src/tools/fuzzing/fuzzing.cpp
kripken: Even nicer would be to return an anonymous struct like some languages have... but yes, this is not bad...

PR #5661, opened 2023-04-13
[Wasm GC] Improve GC operation coverage by using locals more

kripken:
When we emit e.g. a `struct.get`'s reference, this PR makes us prefer a non-nullable
value, and even to reuse an existing local if possible. By doing that we reduce
the risk of a trap, and also by using locals we end up testing operations on the
same data, like this:
```js
x = new A();
x.a = ..
foo(x.a)
```
In contrast, without this PR each of those `x.` uses might be `new A()`.

### DIFF
--- a/src/tools/fuzzing/fuzzing.cpp
+++ b/src/tools/fuzzing/fuzzing.cpp
@@ -2371,6 +2371,40 @@ Expression* TranslateToFuzzReader::makeCompoundRef(Type type) {
   }
 }
 
+Expression* TranslateToFuzzReader::makeTrappingRefUse(HeapType type) {
+  auto percent = upTo(100);
+  // Only give a low probability to emit a nullable reference.
+  if (percent < 5) {
+    return make(Type(type, Nullable));
+  }
+  // Otherwise, usually emit a non-nullable one.
+  auto nonNull = Type(type, NonNullable);
+  if (percent < 70 || !funcContext) {
+    return make(nonNull);
+  }
+  // With significant probability, try to use an existing value. it is better to
+  // have patterns like this:
+  //
+  //  (local.set $ref (struct.new $..
+  //  (struct.get (local.get $ref))
+  //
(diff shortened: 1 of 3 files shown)


src/tools/fuzzing/fuzzing.cpp
```diff
@@ -2323,6 +2323,13 @@ Expression* TranslateToFuzzReader::makeCompoundRef(Type type) {
   // least avoids infinite recursion here, and we emit a valid (but not that
   // useful) wasm.
   if (type.isNonNullable() && (random.finished() || nesting >= LIMIT)) {
+    // If we have a function context then we can at least emit a local.get,
```
tlively, line 2326: It would be good to move some of this info into the comment above, which currently makes it sound like we will definitely trap at runtime.

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -2323,6 +2323,13 @@ Expression* TranslateToFuzzReader::makeCompoundRef(Type type) {
...
+    // perhaps, which is less bad. Note that we need to check typeLocals
+    // manually here to avoid infinite recursion (as makeLocalGet will fall back
+    // to us, if there is no local).
+    if (funcContext && !funcContext->typeLocals[type].empty()) {
```
tlively, line 2330: Does this look for locals holding subtypes as well? If not, it could.

tlively (approved):

src/tools/fuzzing/fuzzing.cpp
kripken: Good point, it doesn't. I added a TODO.

src/tools/fuzzing/fuzzing.cpp
kripken: Good idea, I updated the comment above.

tlively (approved):

PR #5663, opened 2023-04-13
[Wasm GC] Fuzz array.copy and array.fill

kripken:


kripken:
This found a difference between V8 and Binaryen:
```
(module
 (type $type$0 (func))
 (rec
  (type $type$6 (func (result i64)))
  (type $type$7 (struct (field f32) (field (ref null $type$7)) (field i64) (field (mut i32)) (field i16)))
  (type $type$5 (array (ref null $type$9)))
  (type $type$8 (struct (field (mut i31ref)) (field i32) (field i64) (field nullref)))
  (type $type$4 (array_subtype nullref $type$5))
  (type $type$9 (struct (field (mut i32)) (field eqref) (field externref) (field (mut funcref)) (field i64)))
 )
 (rec
  (type $type$2 (array_subtype nullref $type$4))
  (type $type$3 (func (result f32)))
  (type $type$1 (array (mut i16)))
 )
 (func $0 (type $type$0)
  (local $0 (ref null $type$1))
  (array.fill $type$1
   (ref.as_non_null
    (local.get $0)
   )
   (i32.const 0)
   (i32.const -65537)
   (i32.const 0)
  )
 )
)
```
V8 complains with
```
a.wasm.js:38: CompileError: WebAssembly.Module(): Compiling function #0 failed: array.init[2] expected type i16, found i32.const of type i32 @+97
```
That wasm looks valid to me, and it is odd it mentions `array.init` and not `array.fill`. Is this a V8 bug perhaps @jakobkummerow ?

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -3404,6 +3426,64 @@ Expression* TranslateToFuzzReader::makeArraySet(Type type) {
...
+  }
+  auto arrayType = pick(mutableArrays);
+  auto element = arrayType.getArray().element;
+  auto* index = make(Type::i32);
+  auto* ref = makeTrappingRefUse(arrayType);
+  if (oneIn(2)) {
+    // ArrayFill
+    auto* value = make(element.type);
+    auto* length = make(Type::i32);
+    // Only rarely emit a plain get which might trap. See related logic in
```
tlively, line 3442:
```suggestion
    // Only rarely emit a plain array.fill that might trap. See related logic in
```

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -3404,6 +3426,64 @@ Expression* TranslateToFuzzReader::makeArraySet(Type type) {
...
+    // ::makePointer().
+    if (allowOOB && oneIn(10)) {
+      // TODO: fuzz signed and unsigned, and also below
```
tlively, line 3445:
```suggestion
```

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -3404,6 +3426,64 @@ Expression* TranslateToFuzzReader::makeArraySet(Type type) {
...
+  } else {
+    // ArrayCopy. Here we must pick a source array whose element type is a
+    // subtype of the destination.
+    auto srcArrayType = pick(mutableArrays);
+    auto srcElement = srcArrayType.getArray().element;
+    if (!Type::isSubType(srcElement.type, element.type) ||
+        element.packedType != srcElement.packedType) {
+      // TODO: A matrix of which arrays are subtypes of others. For now, if we
+      // didn't get what we want randomly, just copy from the same type to
+      // itself.
```
tlively, line 3462: Alternatively we could filter `mutableArrays` before picking.

src/tools/fuzzing/fuzzing.cpp
```diff
@@ -3404,6 +3426,64 @@ Expression* TranslateToFuzzReader::makeArraySet(Type type) {
...
+      srcArrayType = arrayType;
+      srcElement = element;
+    }
+    auto* srcIndex = make(Type::i32);
+    auto* srcRef = makeTrappingRefUse(srcArrayType);
+    auto* length = make(Type::i32);
+    if (allowOOB && oneIn(10)) {
+      // TODO: fuzz signed and unsigned, and also below
```
tlively, line 3470:
```suggestion
```

tlively (approved):
Looks good! Mostly just comments on comments.

jakobkummerow:
Thanks for reporting this. V8-side fix in flight: https://chromium-review.googlesource.com/c/v8/v8/+/4421302

PR #5665, opened 2023-04-13
[Wasm GC] Ignore unreachable sets for purposes of non-nullable locals

### LINKED ISSUE #5599: Invalid non-nullable local binary emitting related to unreachability
```wat
(module
 (func $test (param $ref (ref struct)) 
  (local $local (ref struct))
  (drop
   (block (result i32)
    (local.set $local
     (unreachable)       ;; this causes validation of the later local.get to fail
     ;; (local.get $ref) ;; this would be ok
    )
    (i32.const 0)
   )
  )
  (drop
   (local.get $local)
  )
 )
)
```
`bin/wasm-as -all --nominal wat -o wasm` emits a binary that V8 complains about, `uninitialized non-defaultable local`. Binaryen does accept the binary as valid, oddly. Also oddly, replacing the unreachable with something reachable fixes it in V8.

I think Binaryen thinks this validates since the unnamed block is not emitted in the binary format, so the `local.set` dominates the `local.get`. But perhaps unreachability makes us behave differently somehow?

kripken:
As pointed out by @tlively in #5599, we don't emit unreachable code in the binary
format, which means such sets can't help gets validate.

Fixes #5599


tlively (approved):

kripken:
It turns out this is tricky. If we land this PR, then any pass that does a refinalize will need to fix up non-nullable locals, beceause refinalization can propagate unreachability - and with this PR, unreachability affects non-nullable local validation.

In practice , that means adding non-nullable fixups to 24 more passes (!) (all those that don't do it now, but do perform a refinalize). That's so many passes that probably we would want to just constantly run non-nullable fixups after every pass and not bother with `requiresNonNullableLocalFixups()` as a way to opt out, and we'd just suffer the slower compile times. But that seems a high cost for this particular corner case...

tlively:
Filed https://github.com/WebAssembly/function-references/issues/98 to discuss a spec change.

kripken:
> Another workaround would be to treat these local.sets specially in the binary writer and emit them even though we otherwise wouldn't.

Unfortunately, I found this wouldn't be simple, as atm we stop at the first block etc. child that is unreachable. So to do this, we'd need to continue onward to check for sets, which would add complexity and overhead.

tlively:
What about the solution of continuing to skip emitting things even beyond the end of blocks, as long as those blocks are unnamed?

kripken:
That doesn't seem simple. It would affect multiple functions, as e.g. `visitPossibleBlockContents` just sees a single block. So we'd need to pass that state around somehow. And we need to preserve the other properties that this code handles, like this:

https://github.com/WebAssembly/binaryen/blob/5679e3b1ea0d4155a6b5c3b93fa6aef34023365f/src/wasm-stack.h#L237-L241

It might be worth trying though, I'm just saying it's not obvious to me that it will work or work easily.

(OTOH, just running non-nullable local fixups, or just running dce before binary emitting, would definitely work and be very simple - but slow us down.)

tlively:
This would all be much easier if we could separate iteration from other logic using generators built on C++20 coroutines :'(

kripken:
I had the idea to notice in ReFinalize when we turn a set unreachable. That is rare, so doing more work in that case would be fine. And it's a change in one place (ReFinalize) and not in multiple passes. That works ok (with some fallout in `LocalRefining`, see code), but it doesn't address the more general problem, a set that is after unreachable code but not unreachable itself,
```wat
(module
 (func $2
  (local $3 (ref array))
  (drop
   (block (result nullref)
    (unreachable)
    (local.set $3
     (ref.as_non_null
      (ref.null none)
     )
    )
    (ref.null none)
   )
  )
  (drop
   (local.get $3)
  )
 )
)
```
That won't validate in v8 but it does work in Binaryen. Noticing that in ReFinalize is not simple and would, I think, add overhead (we'd need to track when we turn any child of anything unreachable, and see if there's a set after it).

kripken:
As of #5677 we have a fuzzer workaround here, and nothing else seems urgent. Closing, and we'll see how the spec and other discussions go around this topic.

PR #5666, opened 2023-04-14
Port a few more tests off of --nominal

tlively:


test/lit/passes/gufa-refs.wast
```diff
@@ -1537,15 +1536,20 @@
 
 ;; Arrays get/set
 (module
-  (type $nothing (array_subtype (mut (ref null any)) data))
+  (rec
+    ;; CHECK:      (rec
+    ;; CHECK-NEXT:  (type $nothing (array (mut anyref)))
```
tlively, line 1541: `$nothing` and `$something-child` appear in the output where they did not previously because this test file does not run `--remove-unused-types`. Normally I would fix that by adding `--remove-unused-types` to the run command, but I can't do that yet without also using `--closed-world`, which would require more changes to this file.

PR #5672, opened 2023-04-15
Remove the nominal type system

tlively:
And since the only type system left is the standard isorecursive type system,
remove `TypeSystem` and its associated APIs entirely. Delete a few tests that
only made sense under the isorecursive type system.


tlively:
* main
  * **PR #5672** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5672" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5672?utm_source=stack-comment).

tlively:
Building new type definitions that refer to previously built types is still allowed. In the extreme, we could even build one rec group at a time, as long as we build them in order. The newly built types and groups may or may not be the same as previously built types and groups.

PR #5676, opened 2023-04-19
Disable sign extension in SignExtLowering.cpp

tlively:
The sign extension lowering pass would previously lower away the sign extension
instructions, but it wouldn't disable the sign extension feature, so follow-on
passes such as optimize-instructions could reintroduce sign extension
instructions.

Fix the pass to disable the sign extension feature to prevent sign extension
instructions from being reintroduced later.

### DIFF
--- a/src/passes/SignExtLowering.cpp
+++ b/src/passes/SignExtLowering.cpp
@@ -68,6 +68,14 @@ struct SignExtLowering : public WalkerPass<PostWalker<SignExtLowering>> {
       }
     }
   }
+
+  void run(Module* module) override {
+    if (!module->features.has(FeatureSet::SignExt)) {
+      return;
+    }
+    super::run(module);
+    module->features.disable(FeatureSet::SignExt);
+  }
 };
 
 Pass* createSignExtLoweringPass() { return new SignExtLowering(); }
--- a/src/passes/pass.cpp
+++ b/src/passes/pass.cpp
@@ -408,7 +408,8 @@ void PassRegistry::registerPasses() {
                "apply more specific subtypes to signature types where possible",
                createSignatureRefiningPass);
   registerPass("signext-lowering",
-               "lower sign-ext operations to wasm mvp",
+               "lower sign-ext operations to wasm mvp and disable the sign "
+               "extension feature",
                createSignExtLoweringPass);
   registerPass("simplify-globals",
                "miscellaneous globals-related optimizations",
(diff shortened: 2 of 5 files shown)


tlively:
* main
  * **PR #5676** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5676" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5676?utm_source=stack-comment).

src/passes/SignExtLowering.cpp
```diff
@@ -68,6 +68,14 @@ struct SignExtLowering : public WalkerPass<PostWalker<SignExtLowering>> {
       }
     }
   }
+
+  void run(Module* module) override {
+    if (!module->features.has(FeatureSet::SignExt)) {
```
sbc100, line 73: Is it possible the module could contain sign ext instructions but this feature not be enabled?  (I guess in that case it would not validate?)

sbc100 (approved):

PrintessEditor:
Wow! Thank you very much for this fast fix! :)

src/passes/SignExtLowering.cpp
sbc100: I also wonder why the test we have in emscripten for this was passing.. investigating.

sbc100: The emscripten test merely checks that `--signext-lowering` is part of the wasm-opt command line .. I didn't both to actually decompile the module to verify binaryen was actually doing the right thing.

kripken:
If we do want to bundle the operations, it might also be good to document that in the pass description perhaps.

kripken:
I can imagine testing/fuzzing situations where we might want to lower certain instructions but maybe generate more of them later (could be more efficient than otherwise). In particular, the fuzzer assumes features stay the same throughout the pipeline, though I don't know if there is any danger to that assumption changing. And I agree that the obvious use case is to lower and disable the feature.

Looks like in existing lowering passes we do that in MultiMemory, for the reason of avoiding the feature appearing in the features section later, but we don't do it for Memory64. We should probably make a general decision and be consistent about it. I don't have strong feelings so we could separately adjust Memory64 as well to disable the feature.

tlively:
I agree we should be consistent. I can update the Memory64 lowering pass as well.

tlively:
@tlively queued this pull request to merge with [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5676).

tlively:
@tlively queued this pull request to merge with [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5676).

PR #5677, opened 2023-04-19
Fuzzer: Run --dce when GC is enabled

kripken:
DCE at the end avoids issues with non-nullable local operations in unreachable
code, which is still being discussed. This PR avoids fuzzer errors for now, but we
should revert it when we have a proper fix.

See

* https://github.com/WebAssembly/binaryen/issues/5599
* https://github.com/WebAssembly/binaryen/pull/5665
* https://github.com/WebAssembly/function-references/issues/98


tlively (approved):
This seems like a good fix for now. If we get the spec change in, we can revert this without doing anything else.

PR #5678, opened 2023-04-19
Remove the ability to construct basic types in a TypeBuilder

tlively:
This capability was originally introduced to support calculating LUBs in the
equirecursive type system, but has not been needed for anything except tests
since the equirecursive type system was removed. Since building basic heap types
is no longer useful and was a source of significant complexity, remove the APIs
that allowed it and the tests that used those APIs.

Also remove test/example/type-builder.cpp, since a significant portion of it
tested the removed APIs and the rest is already better tested in
test/gtest/type-builder.cpp.


tlively:
* main
  * **PR #5678** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5678" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5678?utm_source=stack-comment).

test/example/type-builder.cpp
```diff
@@ -1,233 +0,0 @@
-#include <cassert>
-#include <iostream>
-
-#include "wasm-type-printing.h"
-#include "wasm-type.h"
-
-using namespace wasm;
-
-// Check that the builder works when there are duplicate definitions
-void test_canonicalization() {
```
kripken, line 10: Is this test no longer needed? Not all tests in this file test constructing basic types AFAICT.

tlively: Everything else here is already well tested in test/gtest/type-builder.cpp.

kripken (approved):

PR #5681, opened 2023-04-20
[Wasm GC] Fix a trapsNeverHappen corner case with if/select of a trapping arm

kripken:
The logic says that if an if/select has an arm that returns a null type, and the if/select
goes into a cast, then we can ignore that arm in tnh mode (as it would trap, and we
are ignoring the possibility of a trap). But it is not enough to return a null type - the
null must actually flow out, rather than say a return be executed before.

One existing test needed adjustment, as it used calls for "thing with effects". But a
call can transfer control flow when EH is enabled, and this pass has `-all`. Rather
than mess with the features, I switched the effects to be locals.

### DIFF
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -1630,8 +1630,7 @@ struct OptimizeInstructions
       // function and the null does not actually flow out. We can only optimize
       // here if a null definitely flows out (as only that would cause a trap).
       auto flowsOutNull = [&](Expression* child) {
-        return child->type.isNull() &&
-               !effects(child).transfersControlFlow();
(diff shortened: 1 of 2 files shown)


tlively (approved):
That the calls inhibit this optimization when EH is enabled makes me think that it would be a good idea to infer `nothrow` attributes for functions.

kripken:
> That the calls inhibit this optimization when EH is enabled makes me think that it would be a good idea to infer nothrow attributes for functions.

True. But that could happen automatically by using [function effects](https://github.com/WebAssembly/binaryen/blob/113db86e5e83e8eacd676871b32a651fc9ce064c/src/pass.h#L232-L238).

tlively:
Aha, I forgot we were already doing this.

PR #5682, opened 2023-04-20
[Wasm GC] ReFinalize when needed in SimplifyGlobals

kripken:


tlively (commented):
Would it make sense to relax this validation requirement instead? Do we have any other situations where refining the type of a value can cause a validation failure? That seems bad in general. OTOH, do we have other situations where we refine values without refinalizing? If not, we should make this change anyway.

kripken:
In general refinalizing is the right thing to do, as it lets later optimizations do more, so this is worth landing I think.

But yes, that ref.cast situation is the only case where we actually error. I'd be happier if we removed it (it is odd) but I think your reasoning when you added it was that it would prevent missed optimizations?

tlively:
Ok, let's land this and also remove that validation check. If we were consistent about checking that everything had the most precise possible type in the validator, that might be ok, but doing it just for this one instruction is strange.

tlively (approved):

PR #5683, opened 2023-04-20
[NFC] Simplify rec group initialization in wasm-type.cpp

tlively:
Now that we no longer support constructing basic heap types in TypeBuilder, we
can fully initialize rec groups when they are created, rather than having to
initialize them later during the build step after any basic types have been
canonicalized. Alongside that change, also simplify the process of initializing
a type builder slot to avoid completely overwriting the HeapTypeInfo in the slot
and avoid the hacky workarounds that required.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -120,7 +120,6 @@ struct HeapTypeInfo {
   constexpr bool isArray() const { return kind == ArrayKind; }
   constexpr bool isData() const { return isStruct() || isArray(); }
 
-  HeapTypeInfo& operator=(const HeapTypeInfo& other);
   bool operator==(const HeapTypeInfo& other) const;
   bool operator!=(const HeapTypeInfo& other) const { return !(*this == other); }
 };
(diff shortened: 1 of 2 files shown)


tlively:
* main
  * **PR #5683** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5683" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5683?utm_source=stack-comment).

src/wasm/wasm-type.cpp
```diff
@@ -2247,14 +2248,21 @@ struct TypeBuilder::Impl {
...
-      hti.recGroup = info->recGroup;
-      *info = std::move(hti);
-      info->isTemp = true;
+      info->kind = hti.kind;
+      switch (info->kind) {
+        case HeapTypeInfo::SignatureKind:
+          info->signature = hti.signature;
+          break;
+        case HeapTypeInfo::StructKind:
+          info->struct_ = std::move(hti.struct_);
```
kripken, line 2260: Why is this moved but nothing else?

kripken (approved):

PR #5686, opened 2023-04-24
[Wasm GC] Ignore GC cycle leaks in LSan

kripken:
Leaks happen since we use `std::shared_ptr` which does not handle cycles.
But since Binaryen isn't used in long-running code it's probably find to just
let them leak, and ignore them in LSan.


kripken:
Interesting, thanks, let me try it that way. It does seem cleaner to do in the source, and also I'm running into problems on CI that are annoying with the old way (why would the `example` tests start to fail..?)

tlively (approved):

PR #5690, opened 2023-04-26
[NFC] Track the kinds of items that names refer to in wasm-delegations-fields

kripken:
This does two things that could be split out, but I'm uploading this PR to get
feedback on the direction.

1. This makes delegations-fields track Kinds. That is, rather than say a field is just a Name, we can say it is a name of kind Function. This allows users to track references to functions, tables, memories, etc., in a simple and generic way, avoiding duplicated code which we have atm. (In particular this will help wasm-merge in the future.) This also uses that functionality in two small places to show the benefits (see `memory-utils.cpp`).
2. This refactors `ArrayNewSeg/ArrayInit` to allow that refactoring. Before this PR they have one field with a segment name. After this PR the fields are separate, so we can mark each as what it refers to. This also removes the `op` field which said what the field referred to, which is no longer needed.


src/wasm.h
```diff
@@ -1615,8 +1604,11 @@ class ArrayNewSeg : public SpecificExpression<Expression::ArrayNewSegId> {
 public:
   ArrayNewSeg(MixedArena& allocator) {}
 
-  ArrayNewSegOp op;
-  Name segment;
+  // One of the two should be defined, which specifies if this is an
+  // array.new_data or array.new_elem.
```
tlively, line 1608: I do think it would make more sense to split this into separate `ArrayNewData` and `ArrayNewSeg` rather than having two different names. Almost all the visitors do different (but similar) things for the two operations, so I think this would increase complexity less than having these mutually exclusive fields.

tlively (commented):
This looks like a useful change to wasm-delegations-fields.def 👍

src/wasm-delegations-fields.def
```diff
@@ -665,8 +693,8 @@ switch (DELEGATE_ID) {
   }
   case Expression::Id::ArrayNewSegId: {
     DELEGATE_START(ArrayNewSeg);
-    DELEGATE_FIELD_INT(ArrayNewSeg, op);
-    DELEGATE_FIELD_NAME(ArrayNewSeg, segment);
+    DELEGATE_FIELD_NAME_KIND(ArrayNewSeg, dataSegment, ModuleItemKind::DataSegment);
+    DELEGATE_FIELD_NAME_KIND(ArrayNewSeg, elemSegment, ModuleItemKind::ElementSegment);
```
tlively, line 697: What if we kept `op` and `segment`, but then used the ternary operator on `op` to determine what `ModuleItemKind` to pass in for `segment`? I'd really like to avoid having two different mutually exclusive names in the class.

kripken: We can't quite do that here. Note that we don't have a `curr` object at all, as this is all quite abstract, and in fact we may be operating on more than one object at a time (as equality comparisons do). That is, we don't have a way to query a specific instance at this level.

Can you explain why you don't like having mutually exclusive names? If it's size then we do remove the `op` field, which balances things out. And as mentioned above we do have precedent for this. It seems ok to me?

tlively: Too bad :( It's much nicer when we can use the C++ type system to make it impossible to represent meaningless values rather than having to remember and check a bunch of extra constraints. IMO the cleanest solution here is to split the classes and factor out shared code where possible. I would rather have ~100 extra lines of code than have these mutually exclusive name fields 😅

kripken: Fair enough, it is something the reader needs to keep in mind. I'll work on a patch.

kripken:
Renamed and rebased onto the earlier PR.

src/ir/memory-utils.cpp
```diff
@@ -37,10 +38,33 @@ bool flatten(Module& wasm) {
...
+#define DELEGATE_FIELD_OPTIONAL_CHILD(id, field)
+#define DELEGATE_FIELD_INT(id, field)
+#define DELEGATE_FIELD_INT_ARRAY(id, field)
+#define DELEGATE_FIELD_LITERAL(id, field)
+#define DELEGATE_FIELD_NAME(id, field)
+#define DELEGATE_FIELD_NAME_VECTOR(id, field)
+#define DELEGATE_FIELD_SCOPE_NAME_DEF(id, field)
+#define DELEGATE_FIELD_SCOPE_NAME_USE(id, field)
+#define DELEGATE_FIELD_SCOPE_NAME_USE_VECTOR(id, field)
+#define DELEGATE_FIELD_ADDRESS(id, field)
```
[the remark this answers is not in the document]
kripken: Interesting idea. Worth thinking about, but I'm not sure as in general you need to be careful about omitting any of them, so it's actually good to be forced to write them out, in a way.

In general the way we've avoided the boilerplate from being an issue is writing higher-level C++ abstractions over this type of code. We could do more of that as an alternative.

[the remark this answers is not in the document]
tlively: Yeah, I would definitely want this to be an opt-in mechanism so that by default you are forced to provide all the definitions. For some cases, the code only cares about a single type of thing, and even if we add new types of things in the future, the code still will not care about them.

PR #5691, opened 2023-04-28
Split the maybeSplit into canSplit/doSplit

gkdn:
This patch doesn't inline the methods to help with the review. I will follow up with another patch to inline them.

### DIFF
--- a/src/passes/Inlining.cpp
+++ b/src/passes/Inlining.cpp
@@ -540,10 +540,12 @@ struct FunctionSplitter {
   // the number of iterations, etc.). Therefore this function will only find out
   // if we *can* split, but not actually do any splitting.
   bool canSplit(Function* func) {
+    // TODO: we should not get here if function marked uninlineable.
     if (!canHandleParams(func)) {
       return false;
     }
 
+    // TODO: inline this after review
     return maybeSplit(func);
   }
 
@@ -554,10 +556,8 @@ struct FunctionSplitter {
   // This is called when we are definitely inlining the function, and so it will
   // perform the splitting (if that has not already been done before).
   Function* getInlineableSplitFunction(Function* func) {
-    Function* inlineable = nullptr;
-    [[maybe_unused]] auto success = maybeSplit(func, &inlineable);
-    assert(success && inlineable);
-    return inlineable;
+    // TODO: inline this after review
+    return doSplit(func);
   }
 
   // Clean up. When we are done we no longer need the inlineable functions on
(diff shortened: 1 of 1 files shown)


gkdn:
@kripken

gkdn:
@kripken I added another commit that inlines the functions. The first commit will be easier to review since less code is moved around.

src/passes/Inlining.cpp
```diff
@@ -584,10 +584,12 @@ struct FunctionSplitter {
 private:
   // Information about splitting a function.
   struct Split {
-    // Whether we can split the function. If this is false, the other two will
-    // remain nullptr forever; if this is true then we will populate them the
+    enum class Kind { None, PatternA, PatternB };
```
kripken: Please add a comment that the patterns are described later down.

src/passes/Inlining.cpp
```diff
@@ -607,24 +609,11 @@ struct FunctionSplitter {
   // Check if we can split a function. Returns whether we can. If the out param
   // is provided, also actually does the split, and returns the inlineable split
   // function in that out param.
-  bool maybeSplit(Function* func, Function** inlineableOut = nullptr) {
+  bool maybeSplit(Function* func) {
```
kripken: It looks like this function no longer does splitting, so "maybe split" might be confusing. How about `canSplit`? Looks like it returns true iff we can. Or perhaps `computeSplit` as it computes the split kind and stashes that.

src/passes/Inlining.cpp
```diff
@@ -773,20 +737,63 @@ struct FunctionSplitter {
...
+    return true;
+  }
+
+  Function* doSplit(Function* func) {
+    auto iter = splits.find(func->name);
+    assert (iter != splits.end());
+
+    auto& split = iter->second;
+    if (split.inlineable) {
+      // We all performed the split.
```
kripken: `all` => `already`

src/passes/Inlining.cpp
```diff
@@ -540,91 +540,15 @@ struct FunctionSplitter {
   // the number of iterations, etc.). Therefore this function will only find out
   // if we *can* split, but not actually do any splitting.
   bool canSplit(Function* func) {
+    // TODO: we should not get here if function marked uninlineable.
```
kripken, line 543: What did you have in mind here?

(if you mean the existing `uninlineable` field then that is only used for `canHandleParams`, which is checked below us)

src/passes/Inlining.cpp
kripken: Oh, sorry, it looks like the patterns appear before this actually.

src/passes/Inlining.cpp
kripken: Looks like this was addressed in a later commit.

src/passes/Inlining.cpp
kripken: Ah, this seems to have been removed in later commits.

src/passes/Inlining.cpp
kripken: (lgtm % this question)

gkdn: I was thinking to have worthInlining to do the checking for that (had the follow up in here: cl/527177213)

kripken: I see. That sounds ok, but can you remove the comment from this PR then? (If it only makes sense in the context of the later PR, then it makes more sense there, and removing it here will avoid confusion in reading the code in the future e.g. in bisection).

kripken (approved):

PR #5692, opened 2023-04-28
[NFC] Refactor each of ArrayNewSeg and ArrayInit into subclasses for Data/Elem

kripken:
```
ArrayNewSeg => ArrayNewSegData, ArrayNewSegElem
ArrayInit => ArrayInitData, ArrayInitElem
```

Basically we remove the opcode and use the class type to differentiate them.

This adds some code but it makes the representation simpler and more compact in
memory, and it will help with #5690


scripts/gen-s-parser.py
```diff
@@ -591,8 +591,8 @@
     ("struct.set",           "makeStructSet(s)"),
     ("array.new",            "makeArrayNew(s, false)"),
     ("array.new_default",    "makeArrayNew(s, true)"),
-    ("array.new_data",       "makeArrayNewSeg(s, NewData)"),
-    ("array.new_elem",       "makeArrayNewSeg(s, NewElem)"),
+    ("array.new_data",       "makeArrayNewSegData(s)"),
+    ("array.new_elem",       "makeArrayNewSegElem(s)"),
```
tlively, line 595: We can probably remove `Seg` from the class names since it doesn't appear in the instruction names.

src/wasm-interpreter.h
```diff
@@ -3606,43 +3620,54 @@ class ModuleRunnerBase : public ExpressionRunner<SubType> {
...
+    auto sizeFlow = self()->visit(curr->size);
+    if (sizeFlow.breaking()) {
+      return sizeFlow;
+    }
+
+    uint64_t offset = offsetFlow.getSingleValue().getUnsigned();
+    uint64_t size = sizeFlow.getSingleValue().getUnsigned();
+
+    auto heapType = curr->type.getHeapType();
+    [[maybe_unused]] auto elemType = heapType.getArray().element.type;
```
tlively, line 3653: It looks like this is actually unused. Should we remove it?

tlively (approved):
I didn't read _too_ closely  on the assumption that this is largely mechanical. Is there anywhere I should pay special attention?

LGTM % comments otherwise

kripken:
Yes, almost entirely mechanical, but I did add some template usage to reduce the code size increase in a few places.

kripken:
Good ideas, done.

PR #5693, opened 2023-04-28
[NFC] Inlining: Avoid double-checking canHandleParams.

gkdn:


src/passes/Inlining.cpp
```diff
@@ -85,10 +85,7 @@ struct FunctionInfo {
   }
 
   // See pass.h for how defaults for these options were chosen.
-  bool worthInlining(PassOptions& options) {
-    if (uninlineable) {
-      return false;
-    }
+  bool worthDirectInlining(PassOptions& options) {
```
kripken, line 88: This change is correct, but I think it is a little surprising to read. This location here is in `FunctionInfo`, and the function `worthInlining` looks at the info in that class, compares to the options, and returns whether it is worthwhile. So it is odd for one of the fields, `uninlineable`, to not be checked here like all the others.

That is, this PR seems to move the check for that internal field to an external location outside of the class, which seems less good to me as it might make future refactorings harder.

`canHandleParams` is not a slow function, so I am not worried about it being called more than once. But maybe I am missing something?

gkdn: I generally agree with point that it is not the best place.
Where I was trying to get with this direction is to replace this boolean state with an "inlining mode". With that when we do the analysis, we can decide what kind of inlining we want to do and remember that.  (i.e. won't need to call worthInlining check again to understand what kind of inlining we want to do when we are actually performing)
So:
The next PR will first introduce this InliningMode which will avoid us doing full worthInlining checks again.
Then the PR after that one will change partial inlining analysis to sometimes indicate a direct inlining would be better.

If the general direction makes sense but this PR by itself not a good intermediate step, I can also merge it with the next one.

kripken: I think I see. Sgtm to merge with the next PR then, as this PR by itself does not look like it makes complete sense by itself.

PR #5695, opened 2023-05-01
[NFC] Start tracking InliningMode and memoize it to avoid re-analysis.

gkdn:
This also creates the backbone for canSplit to return different inlining modes.

### DIFF
--- a/src/passes/Inlining.cpp
+++ b/src/passes/Inlining.cpp
@@ -678,7 +678,8 @@ struct FunctionSplitter {
   // perform the splitting (if that has not already been done before).
   Function* getInlineableSplitFunction(Function* func,
                                        InliningMode inliningMode) {
-    assert(inliningMode == InliningMode::SplitPatternA || inliningMode == InliningMode::SplitPatternB);
+    assert(inliningMode == InliningMode::SplitPatternA ||
+           inliningMode == InliningMode::SplitPatternB);
     auto& split = splits[func->name];
 
     if (!split.inlineable) {
--- a/src/passes/Inlining.cpp
+++ b/src/passes/Inlining.cpp
@@ -50,6 +50,14 @@ namespace wasm {
 
 namespace {
 
+enum class InliningMode {
+  Unknown,
+  Uninlineable,
+  Direct,
+  SplitPatternA,
+  SplitPatternB
+};
+
 // Useful into on a function, helping us decide if we can inline it
 struct FunctionInfo {
   std::atomic<Index> refs;
@@ -58,7 +66,7 @@ struct FunctionInfo {
   bool hasLoops;
   bool hasTryDelegate;
   bool usedGlobally; // in a table or export
-  bool uninlineable;
+  InliningMode inliningMode;
 
   FunctionInfo() { clear(); }
 
@@ -69,7 +77,7 @@ struct FunctionInfo {
     hasLoops = false;
     hasTryDelegate = false;
     usedGlobally = false;
-    uninlineable = false;
+    inliningMode = InliningMode::Unknown;
   }
 
   // Provide an explicit = operator as the |refs| field lacks one by default.
@@ -80,15 +88,12 @@ struct FunctionInfo {
     hasLoops = other.hasLoops;
     hasTryDelegate = other.hasTryDelegate;
     usedGlobally = other.usedGlobally;
-    uninlineable = other.uninlineable;
+    inliningMode = other.inliningMode;
     return *this;
   }
 
   // See pass.h for how defaults for these options were chosen.
-  bool worthInlining(PassOptions& options) {
-    if (uninlineable) {
-      return false;
-    }
+  bool worthDirectInlining(PassOptions& options) {
     // Until we have proper support for try-delegate, ignore such functions.
     // FIXME https://github.com/WebAssembly/binaryen/issues/3634
     if (hasTryDelegate) {
@@ -181,7 +186,7 @@ struct FunctionInfoScanner
     auto& info = (*infos)[curr->name];
 
     if (!canHandleParams(curr)) {
-      info.uninlineable = true;
+      info.inliningMode = InliningMode::Uninlineable;
     }
 
     info.size = Measurer::measure(curr->body);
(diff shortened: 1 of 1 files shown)


src/passes/Inlining.cpp
```diff
@@ -50,6 +50,14 @@ namespace wasm {
 
 namespace {
 
+enum class InliningMode {
+  Unknown,
+  Uninlineable,
+  Direct,
+  SplitPatternA,
+  SplitPatternB
```
kripken, line 58:
```suggestion
  // We do not know yet if this function can be inlined, as that has
  // not been computed yet.
  Unknown,
  // This function cannot be inlinined in any way.
  Uninlineable,
  // This function can be inlined normally.
  Direct,
  // This function cannot be inlined normally, but we can use split inlining,
  // using pattern "A" or "B" (see below).
  SplitPatternA,
  SplitPatternB
```

src/passes/Inlining.cpp
```diff
@@ -200,7 +205,7 @@ struct InliningAction {
 };
 
 struct InliningState {
-  std::unordered_set<Name> worthInlining;
+  std::unordered_map<Name, InliningMode> inlinableFunctions;
```
kripken, line 208:
```suggestion
  // Maps functions worth inlining to the mode with which we can inline them.
  std::unordered_map<Name, InliningMode> inlinableFunctions;
```

src/passes/Inlining.cpp
```diff
@@ -539,43 +544,28 @@ struct FunctionSplitter {
...
-    if (iter != splits.end()) {
-      return iter->second.splitKind != Split::Kind::None;
-    }
-
-    // The default value of split.splittable is false, so if we fail we just
-    // need to return false from this function. If, on the other hand, we can
-    // split, then we will update this split info accordingly.
-    auto& split = splits[func->name];
-
+  InliningMode getSplitDrivenInliningMode(Function* func) {
```
kripken, line 547: The comment above needs updating, in particular "returns true" is no longer accurate as the return type is changed.

Also, how about removing "Driven" from the name? I think it's clearer without it.

src/passes/Inlining.cpp
```diff
@@ -683,15 +669,13 @@ struct FunctionSplitter {
   //
   // This is called when we are definitely inlining the function, and so it will
   // perform the splitting (if that has not already been done before).
-  Function* getInlineableSplitFunction(Function* func) {
-    auto iter = splits.find(func->name);
-    assert(iter != splits.end());
+  Function* getInlineableSplitFunction(Function* func,
+                                       InliningMode inliningMode) {
```
kripken, line 673:
```suggestion
                                       InliningMode inliningMode) {
    assert(inliningMode == InliningMode::SplitPatternA || inliningMode == InliningMode::SplitPatternB);
```

src/passes/Inlining.cpp
```diff
@@ -1079,20 +1058,29 @@ struct Inlining : public Pass {
     });
   }
 
-  bool worthInlining(Name name) {
+  InliningMode getInliningMode(Name name) {
+    auto& info = infos[name];
+
+    if (info.inliningMode != InliningMode::Unknown) {
```
kripken, line 1064: Where/how is the default value set to Unknown? I'd guess we need line 69 to change from
```cpp
InliningMode inliningMode;
```
to
```cpp
InliningMode inliningMode = InliningMode::Unknown;
```
But perhaps there is some C++ default value spec thing I'm forgetting?

kripken (commented):
"Direct" doesn't seem clear enough to me to mean "non-split inlining" aka "normal inlining." I don't have a much better idea, but perhaps "Normal"?

src/passes/Inlining.cpp
gkdn: I put the funky name because getSplitIniningMode may return a non-split inlining mode in the  next PR (i.e. it may return Direct). 
If you don't think that would be confusing, I can do the rename.

src/passes/Inlining.cpp
gkdn: The FunctionInfo constructor calls `clear()` where are all the initializations are. So I had the the initialization there next to other field initializations.

src/passes/Inlining.cpp
kripken: I'm not sure without seeing the next PR, but perhaps let's rename it here and then rename it again in the next PR? Hopefully we can find a better name at that time - it sounds like the purpose of the function might be changing.

src/passes/Inlining.cpp
kripken: Right, sorry, I forgot we have that pattern here... iirc it's because the atomic one can't be initialized inline so for consistency we moved them all to stay together.

src/passes/Inlining.cpp
gkdn: Done.

kripken:
> "Direct" doesn't seem clear enough to me to mean "non-split inlining" aka "normal inlining." I don't have a much better idea, but perhaps "Normal"?

I still think this comment from before is worth discussing. What do you think about "Direct => Normal"? It would help code like this I think:
```diff
+  // This function can be inlined normally.
+  Direct,
```
Changing that to "inlined directly" isn't that clear to me, while changing the option to "Normal" seems more understandable.

gkdn:
> > "Direct" doesn't seem clear enough to me to mean "non-split inlining" aka "normal inlining." I don't have a much better idea, but perhaps "Normal"?
> 
> I still think this comment from before is worth discussing. What do you think about "Direct => Normal"? It would help code like this I think:
> 
> ```diff
> +  // This function can be inlined normally.
> +  Direct,
> ```
> 
> Changing that to "inlined directly" isn't that clear to me, while changing the option to "Normal" seems more understandable.

We can call it `Normal`. Or maybe `Full` to contrast with "partial"?

kripken:
I like `Full`, good idea. Then the comment could be
```
// This function can be inlined fully, that is, normally: the entire function can be
// inlined. This is in contrast to partial inlining, see below.
Full,

gkdn:
> I like `Full`, good idea. Then the comment could be
> 
> ```
> // This function can be inlined fully, that is, normally: the entire function can be
> // inlined. This is in contrast to partial inlining, see below.
> Full,
> ```

Ok renamed to Full.

kripken (approved):

PR #5696, opened 2023-05-01
[NFC] Scan first in UniqueNameMapper

kripken:
We can scan far faster than the name mapping process itself, so do that
first, as in the common case nothing needs to be fixed up.

This together with a PR to be opened will change the profile of
`--inlining-optimizing` from being 50% UniqueNameMapper (:scream: ) to
around 10%.

cc @gkdn

### DIFF
--- a/src/wasm/parsing.cpp
+++ b/src/wasm/parsing.cpp
@@ -108,7 +108,8 @@ void UniqueNameMapper::clear() {
 namespace {
 
 struct DuplicateNameScanner
-  : public PostWalker<DuplicateNameScanner, UnifiedExpressionVisitor<DuplicateNameScanner>> {
+  : public PostWalker<DuplicateNameScanner,
+                      UnifiedExpressionVisitor<DuplicateNameScanner>> {
 
   // Whether things are ok. If not, we need to fix things up.
   bool ok = true;
--- a/src/wasm/parsing.cpp
+++ b/src/wasm/parsing.cpp
@@ -16,6 +16,7 @@
 
 #include "parsing.h"
 #include "ir/branch-utils.h"
+#include "support/small_set.h"
 
 namespace wasm {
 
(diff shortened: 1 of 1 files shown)


src/wasm/parsing.cpp
```diff
@@ -105,6 +106,39 @@ void UniqueNameMapper::clear() {
...
+
+    // Whether things are ok. If not, we need to fix things up.
+    bool ok = true;
+
+    // It is rare to have many nested names in general, so track the seen names
+    // as we go in an efficient way.
+    SmallUnorderedSet<Name, 10> seen;
+
+    void visitExpression(Expression* curr) {
+      BranchUtils::operateOnScopeNameDefs(curr, [&](Name& name) {
```
tlively, line 122: Does this traverse the entire expression tree rooted at `curr` or does it only visit `curr`? It would be unfortunate if a performance optimization accidentally introduced quadratic behavior!

tlively (commented):
It seems strange to me that scanning to look for uniqueness problems is so much faster than scanning to fix those problems. This is also does more work overall when there is a problem that needs fixing. Is there room to optimize the existing implementation of `UniqueNameMapper` instead of working around its poor performance? I noticed that it uses `std::map`. Can we use `std::unordered_map` instead, for instance?

tlively:
This still seems like a lot of code and complexity to implement a fast path optimization. For another example of low hanging fruit in `UniqueNameMapper`, the `uniqueToSource` method is completely unused in the code base. I would strongly prefer if we optimized the core `UniqueNameMapper` functionality first, then considered adding this fast path if it still showed strong speedups. Adding `SmallSet` (or even starting to use absl data structures) sounds like a good idea to me and would also help in future performance optimizations.

kripken:
I refactored it to move the class out. Perhaps now it is clearer how small of a change this is? It just adds a few lines to `uniquify()`, and the helper class is trivial, the only functional lines there are
```cpp
    BranchUtils::operateOnScopeNameDefs(curr, [&](Name& name) {
      if (!name.is()) {
        return;
      }
      if (seen.count(name)) {
        ok = false;
      } else {
        seen.insert(name);
      }
    });
```

I.e., if the scope def is seen, error; if not, note it. That's all. I think it's an extremely simple helper class, in fully idiomatic Binaryen form, isn't it? (Btw, we might actually reuse it in the validator - that also has to check for name duplicates. It might add some overhead to do that, though, as right now it's interleaved in the single walk there.)

I am pushing back here because I think it would be a waste of time to optimize the name mapping logic, because it simply does much more work than the scanner. By a large margin. I don't think it's possible that we'll optimize it to the point where we don't want to spend 20 or so lines on a simple helper class to get an early exit, as this PR does. And an early exit is just the right thing to do for something with this much runtime overhead: 50% (!) of inlining time, and inlining is one of our heavier optimizations.

What I would consider a viable option is to investigate an algorithm for the name mapper that has overhead similar to the scanner. A rewrite of the mapper like that could conceivably result in better code for similar speed. I'm not sure I know how to do that, though - it seems like both maps and stacks are required (stacks, because duplicate names can nest, that's valid wat).

Another viable option might be to to investigate a way to avoid even adding a block during inlining - we may be able to do so in the common case.

But both of those viable options may end up more complex than this PR's early exit.

kripken:
> What I would consider a viable option is to investigate an algorithm for the name mapper that has overhead similar to the scanner. 

Actually maybe that is possible. A set like the scanner uses can be used to check if we need to do any name mapping at all, and all other code paths could be behind a check if we saw a such a need. Those should be well-predicted branches (assuming we only rarely find a conflict), so maybe it could be fast... I'll experiment tomorrow.

kripken:
Unfortunately, it looks like I was right before, and the idea in the last comment doesn't work out. See the commit before last which is then reverted, it implements the best way I can think of to do name mapping here with small sets/vectors, and avoiding work when there are no changes. But this is still around 1.5-2x slower than the quick early scan.

What I think is happening is that even with the effort to skip work when there are no fixups, the scanner just does a simple postorder walk, while the commit before last also has to visit control flow structures as it goes in both pre and post-order (so that it can see where a duplicate name scope begins and ends, if it is a duplicate). That's not a lot of work, but compared to a fast postorder walk that does almost no actual work it seems it's still significant.

tlively (approved):
Great, I like this a lot better factored out into a separate util like this. Thanks for all your experimentation here!

PR #5697, opened 2023-05-01
Generate unique block names when inlining

kripken:
Each time we inline we put the contents in a block. Before we used the same name
each time we inlined the same method, and as a result had many conflicts if a
function was inlined many times.

This is not 100% NFC as it does change block names, which is observable in the IR
(as can be seen in the test updates).

This plus #5696 leads to the 5x speedup of UniqueNameManner as mentioned in
that PR. This PR makes us emit more unique names, and that PR takes advantage of
names being unique to exit early after scanning.

cc @gkdn


tlively (approved):

kripken:
I'll wait to land this as we are still debating in #5696 (if that doesn't land, this matters much less), and also it has a conflict related to @gkdn 's PRs which are still in flight (might be simpler to wait for them all to land first).

PR #5698, opened 2023-05-02
Fallback to direct inline if the outline will be inlined.

gkdn:
This workarounds the extra work around the edge case where;
 - Function is too big to full-inline
 - It is a candidate for partial inline
 - Outlined version becomes eligible for full-inline.

In such a case, binaryen would introduce a temporary state with partial inlined functions and later on inline them. J2CL hit this scenario for String literal which resulted in significant regressions in compilation time.

This patch updates partial inlining analysis to identify the edge case and direct to full-inling when that happens.

It probably worths adding new tests but unfortunately I'm not familiar enough with Binaryen to add such tests.

### DIFF
--- a/src/passes/Inlining.cpp
+++ b/src/passes/Inlining.cpp
@@ -588,8 +588,7 @@ struct FunctionSplitter {
       // return), and we would not even attempt to do splitting.
       assert(body->is<Block>());
 
-      auto outlinedFunctionSize =
-        info.size - Measurer::measure(iff);
+      auto outlinedFunctionSize = info.size - Measurer::measure(iff);
       // If outlined function will be worth normal inline, skip the intermediate
       // state and inline fully.
       if (outlinedFunctionWorthInlining(info, outlinedFunctionSize)) {
@@ -676,8 +675,8 @@ struct FunctionSplitter {
     }
     // Success, this matches the pattern.
 
-    // If the outlined function will be worth inlining normally, skip the intermediate
-    // state and inline fully.
+    // If the outlined function will be worth inlining normally, skip the
+    // intermediate state and inline fully.
     if (numIfs == 1) {
       auto outlinedFunctionSize = Measurer::measure(iff->ifTrue);
       if (outlinedFunctionWorthInlining(info, outlinedFunctionSize)) {
@@ -754,8 +753,7 @@ private:
     // loop or calls even though this section may not have.
     // This is a conservative estimate, that is, it will return true only when
     // it should, but might return false when a more precise analysis would
-    return true. And it is a practical
-    // estimation to avoid extra future work.
+    // return true. And it is a practical estimation to avoid extra future work.
     info = origin;
     info.size = sizeEstimate;
     return info.worthFullInlining(options);
--- a/src/passes/Inlining.cpp
+++ b/src/passes/Inlining.cpp
@@ -588,8 +588,8 @@ struct FunctionSplitter {
       // return), and we would not even attempt to do splitting.
       assert(body->is<Block>());
 
-      Index outlinedFunctionSize =
-        info.size - Measurer::measure(iff->condition) - /* if itself */ 1;
+      auto outlinedFunctionSize =
+        info.size - Measurer::measure(iff);
       // If outlined function will be worth normal inline, skip the intermediate
       // state and inline fully.
       if (outlinedFunctionWorthInlining(info, outlinedFunctionSize)) {
@@ -676,10 +676,10 @@ struct FunctionSplitter {
     }
     // Success, this matches the pattern.
 
-    // If outlined function will be worth normal inline, skip the intermediate
+    // If the outlined function will be worth inlining normally, skip the intermediate
     // state and inline fully.
     if (numIfs == 1) {
-      Index outlinedFunctionSize = Measurer::measure(iff->ifTrue);
+      auto outlinedFunctionSize = Measurer::measure(iff->ifTrue);
       if (outlinedFunctionWorthInlining(info, outlinedFunctionSize)) {
         return InliningMode::Full;
       }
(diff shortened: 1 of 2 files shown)


src/passes/Inlining.cpp
```diff
@@ -588,6 +588,14 @@ struct FunctionSplitter {
       // return), and we would not even attempt to do splitting.
       assert(body->is<Block>());
 
+      Index outlinedFunctionSize =
+        info.size - Measurer::measure(iff->condition) - /* if itself */ 1;
```
kripken, line 592:
```suggestion
        info.size - Measurer::measure(iff);
```
This should be identical, and is simpler. Was the current code an optimization to save work? If so I think it's too small to be worth optimizing. (In particular, a risk is that the measure metric could be improved in the future, so this math might end up wrong.)

src/passes/Inlining.cpp
```diff
@@ -666,8 +674,17 @@ struct FunctionSplitter {
         assert(iff->ifTrue->type == Type::unreachable);
       }
     }
-
     // Success, this matches the pattern.
+
+    // If outlined function will be worth normal inline, skip the intermediate
+    // state and inline fully.
+    if (numIfs == 1) {
+      Index outlinedFunctionSize = Measurer::measure(iff->ifTrue);
```
kripken, line 682: I don't follow this. We outline the body of the if, not the condition?

src/passes/Inlining.cpp
```diff
@@ -730,6 +747,17 @@ struct FunctionSplitter {
   // staying constant.
   std::unordered_map<Name, Split> splits;
 
+  bool outlinedFunctionWorthInlining(FunctionInfo& origin, Index sizeEstimate) {
+    FunctionInfo info;
+    // This is not accurate, for example the origin function may have
+    // loop or calls even though this section may not have.
+    // However since this is more conservative and it is a practical
```
kripken, line 754:
```suggestion
    // This is a conservative estimate, that is, it will return true only when
    // it should, but might return false when a more precise analysis would
    return true. And it is a practical
```

src/passes/Inlining.cpp
gkdn: This wasn't intended as an optimization; was trying to avoid the "false" body since I assume this pattern could "technically" work with an "else" as well. 
However since absence of "else" is enforced, I can simplify it.

src/passes/Inlining.cpp
gkdn: My understanding is with single if condition PatternB looks like:

```
  if (<condition>) {
    outlined(...);
  }
```

Did I get it wrong?

kripken: Ah, sorry, I read that wrong. You are correct. I read it as measuring `iff->condition` but your code is right as it has `iff->ifTrue`.

src/passes/Inlining.cpp
```diff
@@ -588,6 +588,14 @@ struct FunctionSplitter {
       // return), and we would not even attempt to do splitting.
       assert(body->is<Block>());
 
+      Index outlinedFunctionSize =
```
kripken, line 591:
```suggestion
      auto outlinedFunctionSize =
```

src/passes/Inlining.cpp
kripken, line 682:
```suggestion
      auto outlinedFunctionSize = Measurer::measure(iff->ifTrue);
```

src/passes/Inlining.cpp
```diff
@@ -552,7 +552,7 @@ struct FunctionSplitter {
   // into account (like limitations on which functions can be inlined into in
   // each iteration, the number of iterations, etc.). Therefore this function
   // may only find out if we *can* split, but not actually do any splitting.
-  InliningMode getSplitInliningMode(Function* func) {
+  InliningMode getSplitDrivenInliningMode(Function* func, FunctionInfo& info) {
```
kripken, line 555: Please add/update the comment above here, as now the function can return `Full` as well.

kripken (commented):
Code lgtm with a comment update, see below.

Aside from that it looks like tests need to be updated here, see the CI failure on `passes/inlining_splitting.wast`. To do that, you can run
```
./scripts/update_lit_checks.py test/lit/passes/inlining_splitting.wast
```
It's possible that a test that should not be inlining normally is now doing so, so that test might need adjustment.

Also, please test the new functionality in that file. You can basically add new modules at the end, and that update script will add the checks for them. Let me know if something in how that works isn't clear.

kripken:
While writing tests for this the behavior surprised me and I realized that I'd gotten something wrong. We have partial inlining on by default (in `-O3`), it's partial inlining *ifs* that we don't. That means that this change is theoretically noticeable by people doing normal `-O3` builds in Emscripten, which is the recommended link opt for release builds.

However, I think this is not going to be noticeable because all it does is avoid wasted work, correct? Even before this PR we could partially inline and then inline the outlined code later; with this PR we just do that as a single step. So even before this PR we would, in effect, normally inline some functions larger than the inlining limit (if they were slightly larger + had the shape that partial inlining looks for).

If that's correct then please clarify that in the code.

kripken:
(tests in https://github.com/WebAssembly/binaryen/pull/5700)

gkdn:
> While writing tests for this the behavior surprised me and I realized that I'd gotten something wrong. We have partial inlining on by default (in `-O3`), it's partial inlining _ifs_ that we don't. That means that this change is theoretically noticeable by people doing normal `-O3` builds in Emscripten, which is the recommended link opt for release builds.

To double check: this PR didn't start "new" partial inlinings, right?

> 
> However, I think this is not going to be noticeable because all it does is avoid wasted work, correct? Even before this PR we could partially inline and then inline the outlined code later; with this PR we just do that as a single step. So even before this PR we would, in effect, normally inline some functions larger than the inlining limit (if they were slightly larger + had the shape that partial inlining looks for).
> 
> If that's correct then please clarify that in the code comments.

Yes the intention is to only avoid wasted work, not change optimization decision. The only change I am expecting is potentially due to difference that might occur from inlining order change.

I earlier had comment like this where we were returning 'Full':
```
    // If the outlined function will be worth inlining normally, skip the
    // intermediate state and inline fully now.
```

I now add following to the function doc:
```
  // Note that to avoid wasteful work, this function may return "Full' inlining
  // mode instead of a split inining. That is; if it detects that a partial
  // inlining will trigger a follow up full inline of the splitted function
  // it will instead return "InliningMode::Full" directly.
```

Let me know if that helps or if we need more explanation.

PR #5701, opened 2023-05-04
Fix DeadArgumentElimination return value opts on nesting+recursion

kripken:
The testcase here has a recursive call that is also nested in itself, something
like
```wat
(return
  (call $me
    (return
      ..
```
This found a bug in our return value removal logic. When we remove a return
value we both modify call sites (to add drops) and modify returns
(to remove their values). One of those uses pointers into the IR which the other
invalidated, so the order of the two matters. This PR just reorders that code to
fix the bug.


kripken:
Yes, the fuzzer. It doesn't take AI for machines to beat humans sometimes...

PR #5702, opened 2023-05-04
[NFC] Add some comments for split inlining crossing the normal limit

kripken:
PTAL @gkdn

### DIFF
--- a/src/passes/Inlining.cpp
+++ b/src/passes/Inlining.cpp
@@ -556,7 +556,7 @@ struct FunctionSplitter {
   // Note that to avoid wasteful work, this function may return "Full" inlining
   // mode instead of a split inlining. That is, if it detects that a partial
   // inlining will trigger a follow up full inline of the splitted function
-  // it will instead return "InliningMode::Full" directly. In more detail,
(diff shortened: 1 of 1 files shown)


src/passes/Inlining.cpp
```diff
@@ -553,10 +553,53 @@ struct FunctionSplitter {
   // each iteration, the number of iterations, etc.). Therefore this function
   // may only find out if we *can* split, but not actually do any splitting.
   //
-  // Note that to avoid wasteful work, this function may return "Full' inlining
-  // mode instead of a split inining. That is; if it detects that a partial
+  // Note that to avoid wasteful work, this function may return "Full" inlining
+  // mode instead of a split inlining. That is, if it detects that a partial
   // inlining will trigger a follow up full inline of the splitted function
-  // it will instead return "InliningMode::Full" directly.
+  // it will instead return "InliningMode::Full" directly. In more detail,
```
gkdn, line 559: I think adding then could be helpful while reading:

"splitted function, **then** it will instead"

gkdn (approved):
LGTM thanks.

PR #5704, opened 2023-05-04
[Wasm GC] Automatically make RefCast heap types more precise

### LINKED ISSUE #5703: TypeRefining fuzz bug
```wat
(module
 (rec
  (type $A (struct (field (ref struct))))
  (type $B (struct))
 )

 (func $0 (result (ref $A))
  (struct.new $A
   (ref.cast struct
    (struct.get $A 0
     (struct.new $A
      (struct.new_default $B)
     )
    )
   )
  )
 )
)
```

STR:
```wat
$ bin/wasm-opt a.wat -all --type-refining --closed-world
[wasm-validator error in function 0] struct.new operand 0 must have proper type, on 
(struct.new $A
 (ref.cast struct
  (struct.get $A 0
   (struct.new $A
    (struct.new_default $B)
   )
  )
 )
)
Fatal: error after opts
```

kripken:
We already did this for nullablilty, and so for the same reasons we should do it
for heap types as well. Also, I realized that doing so  would solve #5703, which
is the new test added for TypeRefining here.

The fuzz bug solved here is that our analysis of struct gets/sets will skip
copy operations - a read from a field that is written into it. And we skip
fallthrough values while doing so, since it doesn't matter if the read goes
through an if arm or a cast. An if would automatically get a more precise
type during refinalize, so this PR does the same for a cast basically.

Fixes #5703

### DIFF
--- a/test/lit/cast-to-basic.wast
+++ b/test/lit/cast-to-basic.wast
@@ -16,19 +16,17 @@
     )
   )
 
-  ;; CHECK:      (func $cast (type $none_=>_none)
+  ;; CHECK:      (func $cast (type $structref_=>_none) (param $x structref)
   ;; CHECK-NEXT:  (drop
-  ;; CHECK-NEXT:   (ref.cast null none
-  ;; CHECK-NEXT:    (ref.null none)
+  ;; CHECK-NEXT:   (ref.cast null struct
+  ;; CHECK-NEXT:    (local.get $x)
   ;; CHECK-NEXT:   )
   ;; CHECK-NEXT:  )
   ;; CHECK-NEXT: )
-  (func $cast
-    ;; Note that this will not round-trip precisely because Binaryen IR will
-    ;; apply the more refined type to the cast automatically (in finalize).
+  (func $cast (param $x (ref null struct))
     (drop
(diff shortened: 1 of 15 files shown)


test/lit/cast-to-basic.wast
```diff
@@ -18,12 +18,14 @@
...
   ;; CHECK:      (func $cast (type $none_=>_none)
   ;; CHECK-NEXT:  (drop
-  ;; CHECK-NEXT:   (ref.cast null struct
+  ;; CHECK-NEXT:   (ref.cast null none
   ;; CHECK-NEXT:    (ref.null none)
   ;; CHECK-NEXT:   )
   ;; CHECK-NEXT:  )
   ;; CHECK-NEXT: )
   (func $cast
+    ;; Note that this will not round-trip precisely because Binaryen IR will
```
tlively, line 27: Since this test is explicitly testing round-tripping, it might be good to update this test case to use a `(ref null struct)` parameter instead.

tlively (approved):
Makes sense, and this is consistent with what we do with e.g. StructGet, where we recalculate the output type based on the input ref type.

PR #5707, opened 2023-05-09
Remove TypeUpdater in Vacuum and always ReFinalize

kripken:
`TypeUpdater::remove` must be called after removing a thing from the
tree. If not, then we can get confused by something like this:
```wat
(block $b
  (br $b)
)
```
If we first call `TypeUpdater::remove` then we see that the block's only
`br` is going away, so it becomes unreachable. But when we then remove
the `br` then the block should have type `none`. Removing the `br` first
from the IR, and then calling `TypeUpdater::remove`, is the safe way to do it.

However, changing that order in Vacuum is not trivial. After looking into
this, I realized that it is just simpler to remove TypeUpdater entirely.
Instead, we can ReFinalize at the end unconditionally. This has the downside
that we do not propagate type updates "as we go", but that should be very
rare.

Another downside is that `TypeUpdater` tracks the number of `br`s, which
can help remove code like in the one test that regresses here (see comment
there). But I'm not sure that removal was valid - Vacuum should not really be
doing it, and it looks like its related to this bug actually. Instead, we have a
dedicated pass for removing unused `br`s - RemoveUnusedBrs - so leave
things for it.

This PR's benefit, aside from now handling the fuzz testcase, is that it makes
the code simpler and faster. I see a 10-25% speedup on the Vacuum pass on
various binaries I tested on. (Vacuum is one of our faster passes anyhow,
though, so the runtime of `-O1` is not much improved.)

Another minor benefit might be that running ReFinalize more often can
propagate type info more quickly, thanks to #5704 etc. But that is probably
very minor.


tlively:
> This has the downside
> that we do not propagate type updates "as we go", but that should be very
> rare.

Wouldn't it be very common that `unreachable` gets propagated up the tree?

tlively (approved):
Looks great! Always nice to delete so much code. How would you feel about rebasing the test changes on top of #5708?

kripken:
Unreachability can be propagated up, true. But vacuum doesn't actually benefit much from unreachability, unlike DCE. Vacuum focuses more on reachable code, and actually perhaps another simplification could be to make it ignore unreachable code entirely - could remove some corner cases, but I'm not sure.

Test changes lgtm, I can wait for those to land first.

Issue #2767, opened 2020-04-15
wasm-link?


kripken:
In the past we had `wasm-merge` and then removed it for lack of use cases and the belief that `lld` would generally be the linker for everything. However, the need for something like `wasm-merge` has [come up](https://github.com/WebAssembly/binaryen/issues/2276#issuecomment-517367353) since then more than once, and I think I have an idea for a new design and goal for it.

## Goal 

Do the same things conceptually that a wasm VM would do at the "**link**" stage, but as a toolchain tool at compile time. That is, instead of shipping multiple wasm files and linking them at runtime on the client, you can link them at compile time. Hence I suggest the name `wasm-link`.

edit: To clarify, what I mean by "what a wasm VM would do" is "what would happen if a wasm VM were given a set of modules and names, and instantiated the first, then instantiated the second with the first imported with its name, then instantiated the third with the first two imported, and so forth." That is, the most simple and naive linking possible of wasm modules, without any special things done in the middle by JS loader code or anything like that.

Concretely, that means:

* This runs on final wasm modules, not on wasm object files.
* Each input file also has the module name it is imported as in other modules in this link - same as the VM would have.
* In general, all we do is to concatenate, basically, keeping the modules unchanged in their behavior but linking them into one big module. But we also do what the VM would do at wasm link time, like match up imports in one module to exports in the other:
```wat
(module ;; imported with module name "one"
  (export "foo" (func $internal_foo))
  (func $internal_foo ..)
)

+

(module ;; imported with some other module name
  (import "one" "foo" (func $foo))
  (func $call-to-foo
    (call $foo) 
  )
)

== linked to =>

(module
  (func $internal_foo ..)
  (func $call-to-foo
    (call $internal_foo) 
  )
)
```
* In the future interface types will give us a lot more to do here, like fuse lifting/lowering code etc.
* Note that if more than one input module has a memory or a table, multi-memory/multi-table are necessary, we need to handle duplicate internal names, etc.

## Use cases

* The specific use case I have myself atm is I want to explore replacing JS glue code in Emscripten with wasm code using reftypes. For now I'm experimenting with handwritten wat files, and I need a way to link those in. 
* A JS bundler does similar things to JS, and I remember plans to do the same for wasm files there. I'm not sure if that's been done in any of them or not (lack of multiple memories likely prevents it so far). I think it would be good to have a simple standalone tool in binaryen for this, that could perhaps be used by bundlers if they want.
[...]

Issue #2276, opened 2019-08-01
Compiling and linking wat files


ljulliar:
If I understand correctly wasm-as turns a wat source file into wasm file. But is it a wasm object file that can be linked with other wasm object files (with wasm-ld for instance) ?

I saw that wasm-merge disappeared from binaryen some time ago. THat was the tool to "link" several wasm files together wasn't it ? Any alternative tools to link wasm files together ?

Thank you!

tlively:
Binaryen currently doesn't know how to reason fully about wasm object files. Can I ask where your wasm modules are coming from and why you want to merge them? That will help us figure out what the best tool for your job is.

ljulliar:
I am actually developing quite a sizeable application directly using the WASM text format (WAT). I handle everything in a single file right now but it's already quite bulky and I feel a need to split this into several modules and use the proper import / export in each of them and link them together at the end.

My problam is that:
1) wat2wasm (from wabt) generate incomplete wasm object files right now (symbil table is missing) and wasm-ld isn't happy about it
2) llvm-as doesn't seem to accept the WAT format as an input
3) and binary ditched wasm-merge

So right now I haven't yet found a working compilation chain that write WAT files and turn then into a single WASM binary file.

Hope this helps

kripken:
Have you considered using something like [AssemblyScript](https://github.com/AssemblyScript/assemblyscript)? It lets you write code that translates 1 to 1 to wasm, like wast, and I believe it supports separate files etc.

We could restore/rewrite wasm-merge in Binaryen, though, that's come up more than once - there are various use cases it turns out.

tlively:
Closing this because we have a new implementation of `wasm-merge` available. Binaryen still doesn't know how to reason about object files, but that's WAI for now.

Issue #2174, opened 2019-06-17
wasm-merge - why it has been deleted


geloizi:
Curios why it was decided to remove `wasm-merge` from the project? I s there any alternative "recommended way" of merging wasm modules?

tlively:
We deleted it because we weren't using it and weren't aware of any other uses of it. We now have support for WebAssembly object files in LLVM's lld linker, which provides a much more sophisticated way to merge wasm modules, albeit only those that are valid object files. The use case that is no longer supported is merging separate fully-linked modules together, but it seems that generally it would be better not to do that.

Heath123:
Not all compilers use LLVM or object files though (e.g. AssemblyScript)

Heath123:
This could still be useful for modules that don't use linear memory, or ones that has been specifically compiled to only use a certain segment of memory that won't overlap with anything else

Heath123:
If I need to do this, what should I do? Use an older version of Binaryen?

kripken:
@Heath123 At the moment yes, an older version is the best option.

I do think we should restore `wasm-merge` as various use cases have been accumulating, including yours. Hopefully someone will have time for this soon.

PR #5709, opened 2023-05-09
Reintroduce wasm-merge 

kripken:
We used to have a `wasm-merge` tool but removed it for a lack of use cases. Recently
use cases have been showing up in the wasm GC space and elsewhere, as people are
using more diverse toolchains together, for example a project might build some C++
code alongside some wasm GC code. Merging those wasm files together can allow
for nice optimizations like inlining and better DCE etc., so it makes sense to have a
tool for merging.

Removal: https://github.com/WebAssembly/binaryen/pull/1969

Requests:
* https://github.com/WebAssembly/binaryen/issues/2174
* https://github.com/WebAssembly/binaryen/issues/2276
* https://github.com/WebAssembly/binaryen/issues/2767

This PR is a compete rewrite of wasm-merge, not a restoration of the original
codebase. The original code was quite messy (my fault), and also, since then
we've added multi-memory and multi-table which makes things a lot simpler.

The linking semantics are as described in the "wasm-link" issue #2767 : all we do
is merge normal wasm files together and connect imports and export. That is, we
have a graph of modules and their names, and each import to a module name can
be resolved to that module. Basically, like a JS bundler would do for JS, or, in other
words, we do the same operations as JS code would do to glue wasm modules
together at runtime, but at compile time. See the README update in this PR for a
concrete example.

There are no plans to do more than that simple bundling, so this should not
really overlap with `wasm-ld`'s use cases.

This should be fairly fast as it works in linear time on the total input code. However,
it won't be as fast as `wasm-ld`, of course, as it does build Binaryen IR for each
module. An advantage to working on Binaryen IR is we could optimize right after
merging, but in this first version no optimizations are done yet.

Perhaps this new tool should be given a new name, to differentiate it from the
old `wasm-merge`? Other name options might be `wasm-link` or `wasm-bundle`
perhaps.


sbc100:
I'm curious how it handles the module name part of the imports.   i.e. what happens if I have two imports in one of my wasm module: `module1.foo` and `module2.foo` .. then the second module exports `foo`.  Does it resolve both of the imports?  Are the module names simply ignored?

kripken:
> I'm curious how it handles the module name part of the imports. i.e. what happens if I have two imports in one of my wasm module: module1.foo and module2.foo .. then the second module exports foo. Does it resolve both of the imports? Are the module names simply ignored?

Each module is given a name for purposes of linking. That is, wasm imports have a module name and a base name, but module exports do not, but in this tool we actually give each module its module name.

Concretely, if one module imports `module2.foo` then we look for a module that was given name `module2`, and an export `foo` from it.

sbc100:
> > I'm curious how it handles the module name part of the imports. i.e. what happens if I have two imports in one of my wasm module: module1.foo and module2.foo .. then the second module exports foo. Does it resolve both of the imports? Are the module names simply ignored?
> 
> Each module is given a name for purposes of linking. That is, wasm imports have a module name and a base name, but module exports do not, but in this tool we actually give each module its module name.
> 
> Concretely, if one module imports `module2.foo` then we look for a module that was given name `module2`, and an export `foo` from it.
> 
> edit: and that is done on the commandline, `wasm-merge a.wasm module1 b.wasm module2` will merge `a.wasm` and `b.wasm`, and it gives `a.wasm` the module name `module1`, and it gives `b.wasm` `module2`.

Sounds good.  Perhaps we could use the module name from the name section to allow the user so skip the specification of the module name on the command line if they like.

How about `a.wasm:module1` for the long hand and just `a.wasm` for the shorthand which uses the name from the name section?

kripken:
> Perhaps we could use the module name from the name section to allow the user so skip the specification of the module name on the command line if they like.

Oh, interesting, I'd forgotten there was a name section subsection for that...

> How about a.wasm:module1 for the long hand and just a.wasm for the shorthand which uses the name from the name section?

I'm a little worried about things like `:` being used as separators since then we can run into issues with names or paths that contain it. E.g. on windows, `C:\wasms\a.wasm:module1` would not work. That's why I used the normal positional argument separation (`wasm-merge a.wasm module1`). But I'm not that happy with that either...

I do agree we should use the name section for the name as a default. Some ideas:
```python
# --name=[index of module, name of module]
wasm-merge a.wasm b.wasm --name=0,module1

# --name is needed once for each module but can be empty for ones we use the name section for
wasm-merge a.wasm b.wasm --name= --name=module1
```
I'm not that happy with those, though...

scripts/fuzz_opt.py
```diff
@@ -1241,6 +1246,53 @@ def handle(self, wasm):
...
+        make_random_input(second_size, second_input)
+        second_wasm = abspath('second.wasm')
+        run([in_bin('wasm-opt'), second_input, '-ttf', '-o', second_wasm] + FUZZ_OPTS + FEATURE_OPTS)
+
+        # sometimes also optimize the second module
+        if random.random() < 0.5:
+            opts = get_random_opts()
+            run([in_bin('wasm-opt'), second_wasm, '-o', second_wasm, '-all'] + FEATURE_OPTS + opts)
+
+        # merge the wasm files. note that we must pass -all, as even if the two
```
tlively, line 1271: In principle you could just append `--enable-multi-memories --enable-reference-types` rather than `-all`, but idk if there would be much benefit to the extra precision.

scripts/fuzz_opt.py
```diff
@@ -1241,6 +1246,53 @@ def handle(self, wasm):
...
+        # verify that merging in the second module did not alter the output.
+        output = run_bynterp(wasm, ['--fuzz-exec-before', '-all'])
+        output = fix_output(output)
+        merged_output = run_bynterp(merged, ['--fuzz-exec-before', '-all'])
+        merged_output = fix_output(merged_output)
+        # a complication is that the second module's exports are appended, so we
+        # have extra output. to handle that, just prune the tail, so that we
+        # only compare the original exports from the first module.
+        # TODO: compare the second module's exports to themselves as well, but
+        #       they may have been renamed due to overlaps...
```
tlively, line 1291: Silently renaming exports seems bad! Should this be an error by default instead?

src/ir/module-utils.h
```diff
@@ -159,6 +156,15 @@ inline void copyModule(const Module& in, Module& out) {
   for (auto& curr : in.dataSegments) {
     copyDataSegment(curr.get(), out);
   }
+}
+
+inline void copyModule(const Module& in, Module& out) {
+  // we use names throughout, not raw pointers, so simple copying is fine
+  // for everything *but* expressions
+  for (auto& curr : in.exports) {
+    out.addExport(new Export(*curr));
```
tlively, line 165: Can we use a `std::unique_ptr` here instead of a raw pointer?

src/ir/module-utils.h
```diff
@@ -132,12 +132,9 @@ inline DataSegment* copyDataSegment(const DataSegment* segment, Module& out) {
...
 
-inline void copyModule(const Module& in, Module& out) {
-  // we use names throughout, not raw pointers, so simple copying is fine
-  // for everything *but* expressions
-  for (auto& curr : in.exports) {
-    out.addExport(new Export(*curr));
-  }
+// Copies named toplevel module items (things of kind ModuleItemKind). See
+// copyModule() for something that also copies exports, the start function, etc.
+inline void copyModuleItems(const Module& in, Module& out) {
```
tlively, line 137: Is this ever independently useful? If not, I suggest that we roll it into `copyModule`. Also, these methods seem expensive enough that it would make more sense to put them in a .cpp file rather than in a header.

tlively (commented):
Initial comments. Next will have to read wasm-merge.cpp and the tests.

scripts/fuzz_opt.py
kripken: Yeah, I don't think it would help to be precise - just more code to read.

scripts/fuzz_opt.py
kripken: I don't think this should error, since there are use cases like e.g. two modules export "main" and so after renaming we end up with "main" and "main_1". The user of `wasm-merge` knows that the first is the first module's and so forth, so they can call them at the appropriate times. (The renamings are deterministic, which makes that possible.)

src/ir/module-utils.h
kripken: Good point, done.

src/ir/module-utils.h
kripken: Yes, it's used in both `copyModule` and in `wasm-merge` - maybe I don't understand your question? Are you suggesting to duplicate that code?

I agree we should move this out of headers. I can do that afterwards.

tlively: Great, I didn't get far enough to see that it's used in two places. Moving the code out of the header as a follow-up sgtm.

scripts/fuzz_opt.py
tlively: In general I wouldn't expect a user to have a detailed enough global understanding of their system to anticipate name conflicts. In particular for the example of `main`, I expect that merges with two exported `main`s are more likely to be accidental than intentional. I agree that it should be possible to let wasm-merge handle resolve conflicts automatically, but I think that should be opt-in and that erroring out would be a safer default.

Also, WDYT about resolving conflicts or even handling all exports by renaming each _export_ from an original _module_ to _module.export_ in the combined module? That would make conflicts impossible by construction.

kripken: We might want multiple modes here, yeah. I'm not sure what the default should be, though. Renaming all exports to `module.export` would mean we don't handle well the simple use case of a main module being merged with side modules, and the side modules have no final exports to the outside - in that case we just want the main module's exports to be kept as they are, and the side module's exports to be "dropped."

So maybe we want some modes like this:

* Rename conflicting export names.
* Error on conflicting export names.
* Keep exports from one module, drop all the rest. (Or, perhaps a set of modules could keep their exports, with either renaming or erroring.)

src/tools/wasm-merge.cpp
```diff
@@ -0,0 +1,527 @@
...
+    kindModuleExportMaps[ex->kind][exportInfo.moduleName][exportInfo.baseName] =
+      ex->value;
+  }
+
+  // Find all the imports and see which have corresponding exports, which means
+  // there is an internal item we can refer to. We build up a map of the names
+  // that we should update.
+  KindNameMaps kindNameMaps;
+  ModuleUtils::iterImportable(merged, [&](ExternalKind kind, Importable* curr) {
+    if (curr->imported()) {
```
tlively, line 367: We don't have an iterator utility that does this check for us?

src/tools/wasm-merge.cpp
```diff
@@ -0,0 +1,527 @@
...
+      auto internalName = kindModuleExportMaps[kind][curr->module][curr->base];
+      if (internalName.is()) {
+        // We found something to fuse! Add it to the maps for renaming.
+        kindNameMaps[ModuleItemKind(kind)][curr->name] = internalName;
```
tlively, line 371: Does the conversion from `ExternalKind` to `ModuleItemKind` depend on the enum values being the same? If so, that seems brittle and we might want to make that more explicit with a conversion function or something.

src/tools/wasm-merge.cpp
```diff
@@ -0,0 +1,527 @@
...
+    assert(exportModuleMap.count(ex.get()));
+    ExportInfo& exportInfo = exportModuleMap[ex.get()];
+    kindModuleExportMaps[ex->kind][exportInfo.moduleName][exportInfo.baseName] =
+      ex->value;
+  }
+
+  // Find all the imports and see which have corresponding exports, which means
+  // there is an internal item we can refer to. We build up a map of the names
+  // that we should update.
+  KindNameMaps kindNameMaps;
```
tlively, line 365: Can we call this `importNamesToUpdate` or something that describes its use rather than its type?

src/tools/wasm-merge.cpp
```diff
@@ -0,0 +1,527 @@
...
+    if (options.debug) {
+      std::cerr << "reading input '" << inputFile << "' as '" << inputFileName
+                << "'...\n";
+    }
+
+    // For the first input, we'll just read it in directly. For later inputs,
+    // we read them and then merge.
+    std::unique_ptr<Module> laterInput;
+    Module* currModule;
+    if (first) {
```
tlively, line 472: It might be simpler to just use `i == 0` instead of a separate `first` variable.

tlively (commented):
Comments through wasm-merge.cpp. Still have to read the tests.

src/tools/wasm-merge.cpp
```diff
@@ -0,0 +1,527 @@
...
+// even if we fuse - as it is simple to leave that for later optimizations
+// (removing unwanted exports can be done using wasm-metadce, see
+// https://github.com/WebAssembly/binaryen/wiki/Pruning-unneeded-code-in-wasm-files-with-wasm-metadce#example-pruning-exports
+// ).
+//
+// Note that we allow "forward references" - a reference from an earlier module
+// to a later one. If one instantiates the wasm modules in sequence then that is
+// impossible to do, and to work around it e.g. emscripten dynamic linking
+// support will add a thunk, but ES module support should allow it for wasm, and
+// so we support it here for full generality.
```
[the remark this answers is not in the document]
kripken: See some rationale here:

https://exploringjs.com/es6/ch_modules.html#sec_design-goals-es6-modules

(cyclic imports was a design goal in fact).

src/tools/wasm-merge.cpp
kripken: We could add one. We have ones for imported/defined functions, etc. But as this may be the only place that uses it, it might be excessive.

src/tools/wasm-merge.cpp
kripken: Yes, this is an annoyance since C++ has no good way to extend an enum (there's a comment in `wasm.h` on that).

Perhaps an explicit conversion would be a good idea, I agree.

src/tools/wasm-merge.cpp
kripken: Good point, the use matters more. I changed `NameMap` etc. to `NameUpdates` etc. since these are updates to names.

src/tools/wasm-merge.cpp
```diff
@@ -0,0 +1,527 @@
...
+//      (call $f) ;; call $f directly since "foo.bar" resolved as $f
+//    )
+//    (func $f
+//      ..
+//    )
+//  )
+//
+// We call that process "fusing" of imports to exports. Note that we don't
+// bother to optimize here - we don't remove either the export or the import,
+// even if we fuse - as it is simple to leave that for later optimizations
```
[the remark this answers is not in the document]
kripken: Makes sense. The simplest way to do that is to run the optimizer pass for it, I added that now.

kripken:
Thanks for the comments, all feedback so far should be addressed now.

src/tools/wasm-merge.cpp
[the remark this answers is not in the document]
tlively: It looks like this doesn't extend to cycles between Wasm modules, though: https://github.com/WebAssembly/esm-integration/tree/main/proposals/esm-integration#snapshotting-imports

src/tools/wasm-merge.cpp
tlively: I agree that this single use case probably doesn't warrant a new utility.

src/tools/wasm-merge.cpp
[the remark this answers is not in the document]
kripken: Ah, interesting, I didn't know that. I'll update the comment then.

src/tools/wasm-merge.cpp
tlively: If there's at least a comment about keeping them in sync in wasm.h, then it's probably fine.

src/tools/wasm-merge.cpp
tlively: Looks good, thanks!

src/tools/wasm-merge.cpp
kripken: Yes, the comment says they must overlap:

https://github.com/WebAssembly/binaryen/blob/7d5d24f400019ff023b65ccdb3c1d8d07ebb00a5/src/wasm.h#L2071-L2074

test/lit/merge/memory_data.wat
tlively, line 1: It would be good to check that memory imports and exports are fused correctly, too.

test/lit/merge/renamings.wat
```diff
@@ -0,0 +1,360 @@
...
+
+  ;; CHECK:      (elem $other func $foo_3 $other)
+
+  ;; CHECK:      (elem $bar_2 func $other $foo_3)
+
+  ;; CHECK:      (tag $foo (param i32))
+  (tag $foo (param i32))
+
+  ;; CHECK:      (tag $bar (param i64))
+  (tag $bar (param i64))
```
tlively, line 68: Where did the tags from the other module go? It looks like they are correctly referenced, but I don't actually see their declarations. Is this a validator bug, too, or am I blind?

test/lit/merge/renamings.wat
```diff
@@ -0,0 +1,360 @@
...
+
+  ;; This memory has a conflict in second.wat, and so second.wat's $foo
+  ;; will be renamed.
+  (memory $foo 10 20)
+
+  (memory $bar 30 40)
```
tlively, line 74: If you reorder the module elements to match the output order, the check lines will match up better.

test/lit/merge/start.flip.wat.second
```diff
@@ -0,0 +1,8 @@
+(module
+  (func $start
+    ;; Not a start function, but the name overlaps so it will get deduplicated.
```
tlively, line 3: Looks like this is getting optimized out of the merged module.

test/lit/merge/start.wat.second
```diff
@@ -0,0 +1,15 @@
+(module
+  (start $start)
+
+  (func $start
+    (drop
+      (i32.const 1)
+    )
+  )
+
+  (func $user
```
tlively, line 10: This is being optimized out, too.

test/lit/merge/start3.wat.second
```diff
@@ -0,0 +1,16 @@
+(module
+  (start $start)
+
+  (func $start
+    (drop
+      (i32.const 1)
+    )
+  )
+
+  (func $user
```
tlively, line 10: ditto.

test/lit/merge/table_elem.wat
```diff
@@ -0,0 +1,112 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
+
+;; RUN: wasm-merge %s first %s.second second -all -S -o - | filecheck %s
+
+;; Test we rename tables and element segments properly at the module scope.
+;; Table $foo has a name collision, and both of the element segments' names.
```
tlively, line 6: Unfinished sentence?

test/lit/merge/table_elem.wat
tlively, line 1: This test seems redundant with renamings.wat. Can it be removed?

test/lit/merge/cycle.wat.second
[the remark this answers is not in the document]
kripken: About the suffix, yes :smile: The other way lit gets angry about having a file with a suffix it understands that doesn't have any lit stuff in it.

About multiple files, I'm not sure how we could put them in one. The lit runner would need to split them somehow..?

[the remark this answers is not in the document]
kripken: Note that the general use case is probably not going to be a text file with a module name, or even a binary with the name section name for a module. Imagine a module coming from emscripten that is linked with one from Dart.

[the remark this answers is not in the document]
sbc100: llvm has a tool called `split-file` test files as part of the running the test.    I wonder if that is part of the open source lit ?

test/lit/merge/memory_data.wat
kripken: Makes sense, done.

test/lit/merge/cycle.wat.second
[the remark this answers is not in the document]
tlively: We also have our own `foreach` script that does something similar. Adding a `split` script, whether taken from elsewhere or written from scratch, sounds reasonable. That shouldn't block this PR, though.

test/lit/merge/renamings.wat
kripken: I'm not sure why they end up sorted to an odd position by the auto-updater, but they are down on lines 102,104.

test/lit/merge/renamings.wat
kripken: It's annoying though because sometimes after I do it it gets reordered again anyhow by the updater...

Rather than us working on this manually each time it would probably be good to improve the updater script?

tlively:
For the CLI UI for naming modules, I propose that the optional flag `--name=foo` means the next given input file will be assigned name "foo". Otherwise,  if the next given input file has a module name (in the text or in the name section), then that name should be used. Otherwise, the base filename (without extension) of the input file should be used as the module name.

test/lit/merge/start.flip.wat.second
kripken: Good point, I thought this didn't matter when I added global dce but it's best to keep the alive to see the deduplication etc.

test/lit/merge/table_elem.wat
kripken: I think we need both. The other one has no elem that refers to a table. That is, one test has passive segments, the other active. I guess they could be merged (no pun intended) but I think it might be clearer as it is.

kripken:
> For the CLI UI for naming modules, I propose that the optional flag --name=foo means the next given input file will be assigned name "foo". Otherwise, if the next given input file has a module name (in the text or in the name section), then that name should be used. Otherwise, the base filename (without extension) of the input file should be used as the module name.

The one maybe odd part there is the mixture of positional and non-positional items. Making one depend on the other is not something I'm familiar with in commandline tools. Or is there precedent?

test/lit/merge/cycle.wat.second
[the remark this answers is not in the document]
kripken: I am a little wary of adding more complexity to the lit runner, just because we have existing issues with the auto updating (stuff like update lines not being next to the module item they should be, as we see in this PR, or being too close to the module after rather than the one before, that they came from). The more complexity we add means more stuff for the auto-updater to handle. But maybe if we focused on it more we could get it to a better place, so I'm not opposed to this.

test/lit/merge/renamings.wat
tlively: The updater is just using the tool output and not reordering anything itself, so the question is why `Print.cpp` isn't giving you a stable order.

kripken: Ah, it could be Print. But I believe Print is deterministic, so I suspect it's something in the interaction between the two tools. See the last two commits: the one before last is a manual reordering, and the last is the effects of the auto-updater, which didn't change it as much as I worried, but still causes enough reordering that I'm not sure it's better overall.

kripken:
Another option for the commandline UI might be to forgo positional arguments entirely, as they are potentially confusing when mixed with others. We could have this:
```
wasm-merge --input=a.wasm --name=foo --input=b.wasm --name=bar --output=merged.wasm
```
Basically we could have `--input/-i` for inputs like we have `--output/-o` for the output.

tlively:
One similar use of `--flags` to affect positional arguments is the `--whole-archive` and `--no-whole-archive` linker flags.

test/lit/merge/cycle.wat.second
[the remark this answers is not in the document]
tlively: Keeping the tests as they are now seems fine for this PR. If I get time I might experiment with a split script, but I agree we shouldn't require any more complexity in the existing update script.

tlively:
Code and tests LGTM. I believe the only remaining questions are 1) the CLI UI for naming modules, and 2) what options to provide and what to do by default for conflicting export names.

test/lit/merge/table_elem.wat
tlively: Sounds good.

kripken:
> One similar use of --flags to affect positional arguments is the --whole-archive and --no-whole-archive linker flags.

Very good point... I guess there is precedent for such stuff.

I don't really have a strong feeling of what's best for the CLI UI. Perhaps we can bikeshed this on the Monday meeting if there's time.

For handling of export conflicts, do you agree we should have 3 modes like I was getting at before?

1. Rename to avoid conflicts. (Helps if the user knows and expects conflicts, e.g. a `main` in each module.)
2. Error on conflicts. (Avoids surprises.)
3. Avoid conflicts by keeping the exports of exactly one module. (Best for the main module + side modules use case.)

If so perhaps 2 should be the default, as the least surprising, as you've been suggesting?

tlively:
> For handling of export conflicts, do you agree we should have 3 modes like I was getting at before?

Yes! Sounds good.

> If so perhaps 2 should be the default, as the least surprising, as you've been suggesting?

That sounds great to me.

kripken:
Ok, 3 modes + new default are now implemented.

src/tools/wasm-merge.cpp
```diff
@@ -110,6 +110,27 @@ namespace {
...
+  // to make them unique. For example, this is useful if you merge several
+  // modules that each have a "main" export, and it's fine if those are renamed
+  // to "main", "main_1", "main_2" etc. - you'll decide when to call each of
+  // those and do so at the right time.
+  RenameExportConflicts,
+  // Keep the exports from the first module and remove all others. For example,
+  // this is useful when the first module is the main program and the others are
+  // libraries of code that it uses, but that do not have any exports intended
+  // to be used by anyone other than the main program.
+  KeepOnlyFirstModuleExports,
```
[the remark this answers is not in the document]
tlively: Sorry I didn't catch this earlier when you were describing the 3 modes you had in mind!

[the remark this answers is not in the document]
kripken: Interesting. That would allow more things to work. But it would also add more effort for the user in the common case this is meant to address, which is a main module + side modules, since we never want the side module's exports to remain. But maybe that's just more work we should expect the user to do, hmm... they also likely would want to prune some main module exports.

test/lit/merge/export_options.wat
```diff
@@ -0,0 +1,59 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
+
+;; If asked to, we rename the conflicts. The second "func" export will become
+;; "func_1".
+;; RUN: wasm-merge %s first %s.second second --rename-export-conflicts        -S -o - | filecheck %s --check-prefix RENAME
```
tlively, line 5:
```suggestion
;; RUN: wasm-merge %s first %s.second second --rename-export-conflicts -S -o - | filecheck %s --check-prefix RENAME
```

tlively (approved):
LGTM 🎉 🎉 🎉

PR #5711, opened 2023-05-09
Add a "mayNotReturn" effect

kripken:
This changes loops from having the effect "may trap (timeout)" to having
"may not return." The only noticeable difference is in TrapsNeverHappen mode,
which ignores the former but not the latter. So after this PR, in TNH mode we do
not optimize away an infinite loop that seems to have no other side effects. We
may also use this for other things in the future, like continuations/stack switching.

There are upsides and downsides to allowing the optimizer to remove infinite
loops (the C and C++ communities have had interesting discussions on that topic
over the years...) but it seems safer to not optimize them out for now, to let the
most code work properly. If a need comes up to optimize such code, we can look
at various options then (like a flag to ignore infinite loops).

As suggested by @tlively in #5228, which this may unblock.

### DIFF
--- a/src/ir/effects.h
+++ b/src/ir/effects.h
@@ -122,10 +122,6 @@ public:
   // *do* mark a potentially infinite number of allocations as trapping, as all
   // VMs would trap eventually, and the same for potentially infinite recursion,
   // etc.
-  //   * We assume that VMs will timeout eventually, so any loop that we cannot
-  //     prove terminates is considered to trap. (Some VMs might not have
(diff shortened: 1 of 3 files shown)


tlively (approved):
Looks good, thanks! I guess in principle we could add this effect to function calls (or at least import calls) as well, but we don't need to because those are already considered to have all effects?

kripken:
Yes, calls are assumed to have all effects. To avoid setting all effects inside `visitCall` etc. we check for the `call` effect, which is basically "anything".

Issue #5607, opened 2023-03-30
Parse error for `string.measure_utf8`, `string.measure_wtf8`


wingo:
Consider test.wat:
```wat
(module
 (func $f (export "f") (param (ref string)) (result i32)
   (string.measure_utf8 (local.get 0))))
```

Fails to assemble:

```
$ ~/src/binaryen/bin/wasm-as -d --all-features -o /tmp/test.wasm /tmp/test.wat
Loading '/tmp/test.wat'...
s-parsing...
w-parsing...
[parse exception: string.measure_utf8 (at 3:3)]Fatal: error in parsing input
```

Similarly for `string.measure_wtf8`, though the debug diagnostic is different:
```
$ cat /tmp/test.wat
(module
 (func $f (export "f") (param (ref string)) (result i32)
   (string.measure_wtf8 (local.get 0))))
$ ~/src/binaryen/bin/wasm-as -d --all-features -o /tmp/test.wasm /tmp/test.wat
Loading '/tmp/test.wat'...
s-parsing...
w-parsing...
[parse exception: expected string (at 3:24)]Fatal: error in parsing input
```

wingo:
One more, fwiw:
```wat
(module
 (type $bytes (array (mut i8)))
 (func $f (export "f") (param (ref string) (ref $bytes)) (result i32)
   (string.encode_lossy_utf8_array (local.get 0) (local.get 1) (i32.const 0))))
```

kripken:
It looks like we expect `string.measure_wtf8 utf8`. Did this change perhaps? Or maybe I read the spec incorrectly...

Should be a simple fix. But is the text format still in flux? And should we focus on the old or the new parser @tlively ?

tlively:
Probably still the old parser, since the new parser is still missing a few important things.

wingo:
@kripken Since https://github.com/WebAssembly/stringref/pull/46 the "policy" is part of the instruction -- there are no more "utf8" etc particles.  Format not in flux.

kripken:
Got it, thanks @tlively @wingo 

Should be easy to fix this, but I think we'll need to continue to support the old notation for a while until our users migrate (j2wasm). Actually an option might be to only do this in the new parser, and then when they switch they can do it atomically with the notation change.

tlively:
I had envisioned a graceful update to the new parser, which implies that either the old parser would have to support the new syntax at some point or the new parser would have  to support the old syntax temporarily.

[...]

PR #5714, opened 2023-05-10
[Strings] Adopt new instruction binary encoding

vouillon:
See https://github.com/WebAssembly/stringref/pull/46.
This format is already adopted by V8: https://chromium-review.googlesource.com/c/v8/v8/+/3892695.
The text format is left unchanged (see https://github.com/WebAssembly/binaryen/issues/5607 for a discussion on the subject).

I have also added support for `string.encode_lossy_utf8` and `string.encode_lossy_utf8 array` (by allowing the `replace` policy for Binaryen's `string.encode_wtf8` instruction).


kripken:
Thanks for the PR!

You say V8 already implemented this change - is it backwards compatible? AFAIK our state before this PR works in V8, and the linked V8 CL is quite old, so I'm surprised things have been working so far for people using Binaryen+V8...

vouillon:
No, this changes is not backwards compatible. I investigated this because `string.new_wtf8_array replace` was not working.
Maybe they are using `string.new_wtf8 wtf8` which happens to be interpreted by V8 as instruction `string.new_utf8` followed by a `nop` (opcode 0x01)?

kripken:
Interesting... maybe it happens to work out that way.

kripken (approved):
Looks right to me. I'm not 100% sure we understand why the same binaries seem to work before and after this change (that is, with v8 and binaryen using incompatible versions) but it seems like this should only help. And if we see breakage after it lands we can back it out and investigate further.

PR #5716, opened 2023-05-11
TrapsNeverHappen Fuzzing: Handle timeouts in optimized code

kripken:
As the optimizer uses TNH mode more (#5228), we can run into situations where
TNH on a trapping module can do things like hang forever:

```c
while (1) {
  if (condition) trap();
}
```

The if can be optimized out if we assume traps never happen, and then we
hang.

This can only happen if the program traps, so we could just not optimize
modules that trap with TNH in the fuzzer. But then a single trap would skip
an entire module, so we have existing logic to ignore output from the point
of the trap. That doesn't help with an infinite loop, though, so this PR adds
a test to see if the optimized output hangs, and if so, skip the wasm.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -1132,6 +1132,12 @@ class TrapsNeverHappen(TestCaseHandler):
     def handle_pair(self, input, before_wasm, after_wasm, opts):
         before = run_bynterp(before_wasm, ['--fuzz-exec-before'])
 
+        if before == IGNORE:
+            # There is no point to continue since we can't compare this output
+            # to anything, and there is a risk since if we did so we might run
+            # into an infinite loop (see below).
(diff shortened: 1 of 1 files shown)


kripken:
Unfortunately these infinite loops appear to not be rare... I saw another in the fuzzer now. Perhaps the timeout should be reduced.

tlively (approved):
LGTM.

Another solution might be to insert more hang guards on loops after optimizing with TNH.

kripken:
Meanwhile I think #5717 is the best solution here. It's much faster I see (very noticeably so), and more precise. Sorry for the noise. Closing this one.

Adding hang checks is an interesting idea too. At a glance we'd need to refactor that logic out of fuzzer.cpp into a pass, so I'd prefer #5717

PR #5717, opened 2023-05-11
Prune trapping code during TrapsNeverHappen fuzzing

kripken:
Alternative to #5716 - this prunes the trapping export and all others after it
from the wasm, which makes us operate only on non-trapping code, which
is safe for TNH.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -1221,7 +1221,6 @@ class TrapsNeverHappen(TestCaseHandler):
 
             # re-execute the now safe wasm
             before = run_bynterp(before_wasm, ['--fuzz-exec-before'])
-            print(before)
             assert TRAP_PREFIX not in before, 'we should have fixed this problem'
 
         after_wasm_tnh = after_wasm + '.tnh.wasm'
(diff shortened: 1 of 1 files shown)


scripts/fuzz_opt.py
```diff
@@ -1164,23 +1200,30 @@ def handle_pair(self, input, before_wasm, after_wasm, opts):
...
-                # allocations, if that is what the program normally does (and
-                # the normal run only avoided that by trapping).
-                assert IGNORE in after
-                return
-            after_index = after.index(call_line)
-            after = after[:after_index]
+            trapping_export = get_export_from_call_line(call_line)
+
+            # now that we know the trapping export, we can leave only the safe
+            # ones that are before it
```
tlively, line 1206: There's no good way to remove all the trapping exports but leave all the non-trapping exports in place?

kripken: Unfortunately no (before or after this PR) because of global effects. If we call exports `a,b,c` and only `b` traps then it is possible that calling `a,c` will now trap in `c` when it didn't before. In other words, we have a single linear stream of execution of the exports in order, and once that traps we have to stop considering code from that point onward - running any of the later code is risky.

tlively: Makes sense. I guess we could run in a loop, always just removing the first trapping function until there are no more traps, but that sounds expensive.

tlively (approved):

PR #5718, opened 2023-05-12
Update LogExecution pass to emit log ordinals in the wasm function ordinals order

juj:
Update LogExecution pass to emit log ordinals in the wasm function ordinals order, so that they can be correlated with functions afterwards.

This is something I found I needed to do in order to get a PGO based analysis to line up with wasm function indices.

### DIFF
--- a/test/lit/passes/log-execution.wast
+++ b/test/lit/passes/log-execution.wast
@@ -83,11 +83,11 @@
   ;; CHECK-NEXT:   )
   ;; CHECK-NEXT:   (block
   ;; CHECK-NEXT:    (call $log_execution
-  ;; CHECK-NEXT:     (i32.const 10)
+  ;; CHECK-NEXT:     (i32.const 11)
   ;; CHECK-NEXT:    )
   ;; CHECK-NEXT:    (loop
   ;; CHECK-NEXT:     (call $log_execution
-  ;; CHECK-NEXT:      (i32.const 11)
+  ;; CHECK-NEXT:      (i32.const 10)
   ;; CHECK-NEXT:     )
   ;; CHECK-NEXT:     (block
   ;; CHECK-NEXT:      (drop
--- a/test/lit/passes/log-execution_arg.wast
+++ b/test/lit/passes/log-execution_arg.wast
@@ -13,7 +13,7 @@
 
   ;; CHECK:      (func $nopp
   ;; CHECK-NEXT:  (call $log_execution
-  ;; CHECK-NEXT:   (i32.const 0)
+  ;; CHECK-NEXT:   (i32.const 2)
   ;; CHECK-NEXT:  )
   ;; CHECK-NEXT:  (nop)
   ;; CHECK-NEXT: )


src/passes/LogExecution.cpp
```diff
@@ -39,23 +40,47 @@ namespace wasm {
 Name LOGGER("log_execution");
 
 struct LogExecution : public WalkerPass<PostWalker<LogExecution>> {
-  void visitLoop(Loop* curr) { curr->body = makeLogCall(curr->body); }
+  std::map<Function*, Index> functionOrdinals;
```
tlively, line 43: It doesn't look like this needs to be iterated over.
```suggestion
  std::unordered_map<Function*, Index> functionOrdinals;
```

src/passes/LogExecution.cpp
```diff
@@ -39,23 +40,47 @@ namespace wasm {
...
 
-  void visitReturn(Return* curr) { replaceCurrent(makeLogCall(curr)); }
+  Index nextFreeIndex = 0;
+
+  // Tries to convert a string to a function index. Returns (Index)-1 on
+  // failure.
+  Index stringToIndex(const char* s) {
+    for (const char* q = s; *q; ++q)
+      if (!isdigit(*q))
+        return (Index)-1;
```
tlively, line 52: Binaryen style is to always use braces.
```suggestion
    for (const char* q = s; *q; ++q) {
      if (!isdigit(*q)) {
        return (Index)-1;
      }
    }
```

src/passes/LogExecution.cpp
```diff
@@ -81,14 +106,40 @@ struct LogExecution : public WalkerPass<PostWalker<LogExecution>> {
...
+    for (auto& func : curr->functions) {
+       if (func->imported())
+         ++idx;
+    }
+
+    for (auto& func : curr->functions) {
+      if (func->imported())
+        continue;
+
+      Index currentFunctionIndex = (Index)stringToIndex(func->name.toString().c_str());
```
tlively, line 122:
```suggestion
      Index currentFunctionIndex = stringToIndex(func->name.toString().c_str());
```

Also, why try to convert function names to indices at all? It would be more robust to avoid trying to interpret the names so that all function names can be treated uniformly.

juj: Hey @tlively, it's been a while, though I am now coming back to this functionality, and look to get this landed.

Updated the PR here: the idea here is to ensure that Wasm functions that do not have a debug symbol name string, would be guaranteed to get the log ID that matches its function ordinal. It gives consistency, and allows users to map the log IDs then to e.g. with external symbol maps.

Revised the PR to make sure this is the case always, even if some producer might e.g. give debug symbol names to only some of the functions, and not all.

juj:
Updated this PR to latest, looking for a review. Thanks!

juj:
@kripken Maybe you might be available to review?

src/passes/LogExecution.cpp
```diff
@@ -39,6 +41,8 @@ namespace wasm {
 Name LOGGER("log_execution");
 
 struct LogExecution : public WalkerPass<PostWalker<LogExecution>> {
+  std::unordered_map<Function*, Index> functionOrdinals;
```
kripken, line 44: What is a "function ordinal"? I don't see "ordinal" anywhere existing in `src/`.

If it is a new concept please add a comment explaining it.

If this is a function index, please rename it to that. Skimming the code below that might be the case?

juj: They are the function indices. Good idea, I'll rename.

I'll post this in a bit, I find there is some kind of bug in my use, and not sure if it could be related to this code path; so I'll investigate that and then update this PR.

PR #5722, opened 2023-05-15
Optimize ArrayNew zero construction

kripken:
All array elements have the same type, so we can construct a single zero
and just copy it.

This makes ArrayNew of large arrays 2x faster. I also experimented with
putting `Literal::makeZero` in a header, in hopes of inlining leading to
licm helping here, but that did not help at all unfortunately, at least not
in gcc.


tlively (approved):

tlively:
This could have `[NFC]` in the commit title

PR #5723, opened 2023-05-16
EffectAnalyzer: Do not clear break targets before walk()/visit()

kripken:
I honestly can't think of why we did that clearing... seems odd. Maybe I'm
forgetting something? Overall, in general we depend on repeated calls to
walk/visit accumulating effects; if we want to clear stuff then we create a
new EffectAnalyzer. So I think this PR is a correct change.

Removing that fixes the attached testcase. Also added a unit test. The fuzzer
seems happy as well after several thousand iterations.

### DIFF
--- a/test/example/cpp-unit.cpp
+++ b/test/example/cpp-unit.cpp
@@ -637,6 +637,19 @@ void test_effects() {
     assert_equal(effects.readsMutableStruct, false);
     assert_equal(effects.writesStruct, false);
   }
+
+  {
+    EffectAnalyzer effects(options, module);
+
+    // If we break, then we transfer control flow.
+    effects.breakTargets.insert("block");
+    assert_equal(effects.transfersControlFlow(), true);
+
+    // Repeated walks accumulate effects, that is, old effects are not
+    // removed.
+    effects.walk(&nop);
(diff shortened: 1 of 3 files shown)


test/example/cpp-unit.cpp
```diff
@@ -637,6 +637,19 @@ void test_effects() {
     assert_equal(effects.readsMutableStruct, false);
     assert_equal(effects.writesStruct, false);
   }
+
+  {
+    EffectAnalyzer effects(options, module);
```
tlively, line 642: Do we have an effect analyzer gtest suite we could put this in instead?

tlively (approved):
Makes sense to me.

test/example/cpp-unit.cpp
kripken: I looked for that first, but we don't.

Separately it would be nice to migrate these tests, I agree.

kripken:
164 K fuzzer iterations overnight look ok...

aheejin:
Sorry for the late reply. I traced when that was first added and it was in this big PR (#807), so no clue either. 🤷🏻

PR #5725, opened 2023-05-16
[Docs] Add docs for wasm-merge export handling flags

kripken:


README.md
```diff
@@ -709,6 +714,33 @@ merged.exports.main();
...
+By default `wasm-merge` errors if there are overlapping export names. That is,
+`wasm-merge` will automatically handle overlapping function names and so forth,
+because those are not externally visible (the code still behaves the same), but
+if we renamed exports then the outside would need to be modified to expect the
+new export names, and so we error instead on such name conflicts.
+
+If you do want exports to be renamed, run `wasm-merge` with
+`--rename-export-conflicts`. Later exports will have a suffix appended to them
+to ensure they do not overlap with previous exports. The suffixes are
+deterministic, so once you see what they are you can call them from the outside.
```
tlively, line 728: The fact that the renamed names depend on an internal implementation detail of how we generate fresh names is a little unfortunate. It would be nice to have a formally documented renaming scheme instead. (But that doesn't affect this PR.)

tlively (approved):

README.md
kripken: Good point, yeah, if this ends up popular then it might be worth making a true documented ABI for how we do the naming scheme.

PR #5727, opened 2023-05-16
Emit function types on function imports

kripken:
I'm not sure if this was an oversight, or there is a reason not to do it,
but without this certain problems can be hard to debug.


tlively (approved):
I'm pretty sure omitting the type is just wrong, so strongly LGTM.

PR #5731, opened 2023-05-17
Consider local sets with immediate casts in the OptimizeCasts pass.

zm2he:
For instance, the below
```
(local.set $a
  (ref.cast .. 
    (local.get $ref)
  )
)
(local.get $a)
```

Can be converted to
```
(local.set $a
  (local.tee $temp
    (ref.cast ..
       (local.get $ref)
    )
  )
)
(local.get $temp)
```
Currently, this change only applies the optimization if the immediate child of the local.set is a cast and cannot look through other operations (e.g. a cast which has been wrapped in a block or a cast which has been wrapped in a tee). This is added as a TODO.

### DIFF
--- a/src/passes/OptimizeCasts.cpp
+++ b/src/passes/OptimizeCasts.cpp
@@ -75,10 +75,8 @@
 // TODO: Move casts earlier in a basic block as well, at least in traps-never-
 //       happen mode where we can assume they never fail.
 // TODO: Look past individual basic blocks?
-// TODO: Look at LocalSet as well and not just Get. That would add some overlap
-//       with the other passes mentioned above, but once we do things like
-//       moving casts earlier as in the other TODO, we'd be doing uniquely
-//       useful things with LocalSet here.
+// TODO: When looking at Local Sets, check fallthroughs/descendants for casts
+//       instead of just the immediate child
 //
 
 #include "ir/linear-execution.h"
@@ -104,8 +102,13 @@ struct BestCastFinder : public LinearExecutionWalker<BestCastFinder> {
   // This is tracked in each basic block, and cleared between them.
   std::unordered_map<Index, Expression*> mostCastedGets;
 
-  // For each most-downcasted local.get, a vector of other local.gets that could
-  // be replaced with gets of the downcasted value.
+  // Map local indices to the current downcasting of local.set to those indices.
+  //
+  // Also tracked in each basic block and cleared between them.
+  std::unordered_map<Index, Expression*> curCastedSets;
+
+  // For each most-downcasted local.get or local.set, a vector of other
+  // local.gets that could be replaced with gets of the downcasted value.
   //
   // This is tracked until the end of the entire function, and contains the
   // information we need to optimize later. That is, entries here are things we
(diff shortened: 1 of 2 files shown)


kripken:
Hmm, I'm confused by the example. It looks like in both cases the `local.get` at the end reads out a value that is a `ref.cast` of `local.get $ref`. So I don't see how the new `local.tee` improves matters. Sorry if I'm missing something obvious.

zm2he:
In this example, there is probably not much practical benefit, but my main approach is suppose $B is a subclass of $A, and we have this local variable $a of type $A. If $a is first assigned to something cast to $B, and it is later assigned to something of type $A. Then if we tee the first set into some local variable of type $B, we can use this new local variable of type $B, even though $a needs to remain of type $A because of the later set.

If you have any other suggestions for utilizing the local sets, let me know.

src/passes/OptimizeCasts.cpp
```diff
@@ -104,8 +102,13 @@ struct BestCastFinder : public LinearExecutionWalker<BestCastFinder> {
   // This is tracked in each basic block, and cleared between them.
   std::unordered_map<Index, Expression*> mostCastedGets;
 
-  // For each most-downcasted local.get, a vector of other local.gets that could
-  // be replaced with gets of the downcasted value.
+  // Map local indices to the current downcasting of local.set to those indices.
+  //
+  // Also tracked in each basic block and cleared between them.
+  std::unordered_map<Index, Expression*> curCastedSets;
```
tlively, line 108: It would be helpful to mention why there is only one current casted set for each local as opposed to having a  "most" casted set of many like we have for gets. As the reader, I initially expected to see a `mostCastedSets` map exactly mirroring `mostCastedGets`.

Also, we usually use `curr` rather than `cur` as an abbreviation of "current" in Binaryen.

kripken:
I see, thanks! Ok, IIUC this is meant to help with this:

```wat
(module
  (type $A (struct))
  (type $B (struct_subtype $A))
  (import "a" "b" (func $use (param (ref any))))
  (func $test (param $param-a (ref $A)) (param $param-ref (ref any))
    (local $a (ref $A))
    (local.set $a
      (local.get $param-a)
    )
    (call $use
      (local.get $a) ;; this use already has optimal type
    )
    (local.set $a
      (ref.cast $B
        (local.get $param-ref)
      )
    )
    (call $use
      (local.get $a) ;; this use could be of type $B due to the cast
    )
  )
)
```

However, we actually already have a pass for that, SSA. Here is the output of running `--ssa` on that:

```wat
 (func $test (type $ref|$A|_ref|any|_=>_none) (param $param-a (ref $A)) (param $param-ref (ref any))
  (local $a (ref $A))
  (local $3 (ref $A))
  (local $4 (ref $A))
  (local.set $3
   (local.get $param-a)
  )
  (call $use
   (local.get $3)
  )
  (local.set $4
   (ref.cast $B
    (local.get $param-ref)
   )
  )
  (call $use
   (local.get $4)
  )
 )
```
`SSA` created those two new locals, `$3` and `$4`, one for each `local.set`. This is kind of the opposite of `coalesce-locals`: coalescing merges indexes while ssa "untangles" them into multiple locals, which can then potentially each be optimized differently.

So this is a useful optimization, but it does already exist unless I'm missing something. Perhaps the TODO in OptimizeCasts wasn't clear enough, sorry about that. Let's chat more offline when we get a chance.

src/passes/OptimizeCasts.cpp
```diff
@@ -114,22 +117,50 @@ struct BestCastFinder : public LinearExecutionWalker<BestCastFinder> {
...
+    auto setIter = curCastedSets.find(curr->index);
+
+    if (getIter != mostCastedGets.end()) {
+      auto* bestCast = getIter->second;
+      if (setIter != curCastedSets.end()) {
+        // Always use a cast in local.set if it is equal or better than
+        // a local.get since we know it is always before any gets that
+        // retrieve the set value from the index
+        if (bestCast->type == setIter->second->type ||
+            Type::isSubType(setIter->second->type, bestCast->type)) {
```
tlively, line 148: Subtyping includes equality, so there's no need to separate check equality in the code. I do think it's helpful for the reader that you call out "equal or better" in the comment, though.

```suggestion
        if (Type::isSubType(setIter->second->type, bestCast->type)) {
```

tlively:
Oh wow, I hadn't considered that the SSA pass would already handle this. I guess I'm also confused about what the TODO is about!

test/lit/passes/optimize-casts.wast
```diff
@@ -274,9 +284,21 @@
     (drop
       (local.get $x)
     )
+    (local.set $x
+      ;; Cannot look through for sets at the moment
```
tlively, line 288: Generally we try to make comments full sentences, including `.`, although we're not perfect about it.
```suggestion
      ;; Cannot look through for sets at the moment.
```

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +442,165 @@
...
+  ;; CHECK-NEXT:   (local.get $1)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (ref.cast $B
+  ;; CHECK-NEXT:    (local.get $1)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (ref.cast $B
+  ;; CHECK-NEXT:    (local.get $1)
```
tlively, line 535: Why did this cast to `$A` in the source change to a cast to `$B` in the output?

tlively (commented):
Great work! I finished reviewing like I normally would, even though there is that outstanding question about whether we need this optimization after all. Feel free to hold off on addressing the comments until we resolve that question.

kripken:
Apologies, that TODO was not very clear. In a meeting we realized it overlaps with SimplifyLocals. Consider:
```wat
(local.set $x
  (ref.cast $Y
    (local.get $z)))
```
This pass will see that `$z`, the input to the cast, is cast to `$Y` and use that. One can also look at the output from the cast and use the fact that it is in local `$x`. But SimplifyLocals does that.

Higher up the text mentions the overlap between the passes and that once the pass does cross-basic block etc. things then the overlap with SimplifyLocals would be much smaller. But for now, that TODO would be reimplementing SimplifyLocal's EquivalentOptimizer segment, so it's likely not worth atm.

Sorry for not writing a clearer comment in that third TODO!

First and second TODOs there did seem clear to @zm2he and myself though.

zm2he:
Closed for overlap with other passes at the moment. Similar pull request addressing the TODO might be opened later when other optimizations to `OptimizeCasts` make this useful.

PR #5732, opened 2023-05-18
Improve TODO docs in OptimizeCasts

kripken:
cc @tlively @zm2he

### DIFF
--- a/src/passes/OptimizeCasts.cpp
+++ b/src/passes/OptimizeCasts.cpp
@@ -72,13 +72,18 @@
 // sense in new pass anyhow, and things should be simpler overall to keep such
 // casts all in one pass, here.
 //
-// TODO: Move casts earlier in a basic block as well, at least in traps-never-
-//       happen mode where we can assume they never fail.
-// TODO: Look past individual basic blocks?
-// TODO: Look at LocalSet as well and not just Get. That would add some overlap
-//       with the other passes mentioned above, but once we do things like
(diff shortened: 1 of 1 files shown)


src/passes/OptimizeCasts.cpp
```diff
@@ -72,13 +72,18 @@
...
+// TODO: 1. Move casts earlier in a basic block as well, at least in
+//          traps-never-happen mode where we can assume they never fail, and
+//          perhaps in other situations too.
+// TODO: 2. Look past individual basic blocks? This may be worth considering
+//          given the pattern of a cast appearing in an if condition that is
+//          then used in an if arm, for example, where simple dominance shows
+//          the cast can be reused.
+// TODO: 3. Look at LocalSet as well and not just Get. That would add some
+//          overlap with the other passes mentioned above, but once we do the
+//          first two TODOs above then we'd be adding some novel things here.
```
tlively, line 84: Can you explain more about what the novel parts would be and what corresponding improvements could be made to the other passes? This will help me understand more and will also inform our work this summer 🎉

kripken: Sure, I added more details now.

Note that the 3rd TODO is a little speculative and needs more thought, as I added more on now. The first two definitely look worthwhile to me, but I'm not sure of the 3rd - let's reconsider that one later.

tlively (approved):
LGTM, thanks!

PR #5734, opened 2023-05-18
Update br_on_cast binary and text format

tlively:
The final versions of the br_on_cast and br_on_cast_fail instructions have two
reference type annotations: one for the input type and one for the cast target
type. In the binary format, this is represented as a flags byte followed by two
encoded heap types. Since these instructions have been in flux for a while, do
not attempt to maintain backward compatibility with older versions of the
instructions. Instead, upgrade all of the tests at once to use the new versions
of the instructions. Drop some binary tests of deprecated instruction encodings
that would be more effort to update than they're worth.


tlively:
* main
  * **PR #5734** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5734" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5734?utm_source=stack-comment).

test/lit/passes/gufa-refs.wast
```diff
@@ -5387,7 +5387,7 @@
     (drop
       (block $B (result (ref $B))
         (drop
-          (br_on_cast $B $B
+          (br_on_cast $B anyref (ref $B)
```
kripken, line 5390: why not `anyref => (ref $A)` ?

tlively: For a lot of these tests, it was simpler to just blindly type "anyref" as the input type rather than put thought into what more specific input type would make sense. The input type is thrown away during parsing, so it doesn't affect the IR. If you'd like, I can take another pass and replace "anyref" with more precise types.

Another complication is that the cast type is required to be a subtype of the input type (although so far I haven't implemented this yet in the PR), so I can't always use the precise type of ref operand, either.

kripken: Sounds good, I was just checking I wasn't missing something.

kripken (approved):
lgtm with fuzzing

test/lit/passes/gufa-refs.wast
tlively: I started implementing this subtyping constraint, but it turns out to be unfortunately complicated, so I filed https://github.com/WebAssembly/gc/issues/381 to try to get the constraint dropped instead.

tlively:
I think in principle there is a chance that this PR will produce a binary rejected by V8 due to the restriction https://github.com/WebAssembly/gc/issues/381 is aiming to lift, but OTOH, the fuzzer has been through 12k iterations with the CompareVMs probability set to 1, so it seems that we're not hitting this issue in practice.

Should I merge this as-is or should we hold off and see how that issue is resolved and how V8 is updated?

kripken:
I think it's ok to land, since in the worst case we can revert it. Unless there is large followup work on the same code we expect to make a reversion difficult? (I can't think of any.)

tlively:
No, I expect a revert would be straightfoward. The fuzzer has now completed over 20k iterations. I'll land this.

kripken:
> I think it's ok to land, since in the worst case we can revert it.

It turns out that dropping the old opcodes here causes problems for Kotlin, and could delay them upgrading their Binaryen version. After some discussion, it might make sense for Binaryen to match what V8 does, which is to keep the old opcodes for now and probably until the final encoding of the binary format. For now I'll revert this as discussed above, and @tlively when you return we'll talk about what to do going forward here.

PR #5735, opened 2023-05-19
Fuzzer: Do DCE for reference types, not just GC

kripken:
Even with just reference types enabled we can hit the problem that
DCE is working around here. For example `(ref func)`.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -188,12 +188,12 @@ def randomize_fuzz_settings():
     else:
         LEGALIZE = False
 
-    # if GC is enabled then run --dce at the very end, to ensure that our
+    # if reftypes are enabled then run --dce at the very end, to ensure that our
     # binaries validate in other VMs, due to how non-nullable local validation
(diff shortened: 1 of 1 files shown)


tlively (approved):

tlively:
Wait, but reference types alone doesn't enable non-nullable reference types. I believe those come with typed function references, which we combine into the GC feature.

kripken:
Oh, good point...

Looking closer at the failure, we end up with `(ref func)` on some `block`s and `select`s. I'm guessing what happens is that we have `(ref.func $foo)` + a refinalize which propagates that non-nullable type.

We could in theory set the type of `ref.func` to be nullable when GC is not enabled..? But I seem to remember us deciding to always use non-nullable types for `ref.func` internally for simplicity. Special-casing this behavior in refinalize also seems annoying. So I think we are doing the best thing when we allow some non-nullable types to appear even with just reference types.

tlively:
Yes, I see the problem. Will make a PR.

tlively:
See #5737 for a fix.

PR #5737, opened 2023-05-19
Fix emitting of function reference types without GC

tlively:
We previously had logic to emit GC types used in the IR as their corresponding
top types when GC was not enabled (so e.g. nullfuncref would be emitted as
funcref), but the logic was not robust enough and non-null function references
were not properly emitted as funcref.

Refactor the relevant code to be more robust and future-proof, and add a test
demonstrating that the lowering works as intended.


tlively:
* main
  * **PR #5737** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5737" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5737?utm_source=stack-comment).

src/wasm/wasm-binary.cpp
```diff
@@ -1393,6 +1393,18 @@ void WasmBinaryWriter::writeInlineBuffer(const char* data, size_t size) {
 
 void WasmBinaryWriter::writeType(Type type) {
   if (type.isRef()) {
+    // The only reference types allowed without GC are funcref and externref. We
+    // internally use more refined versions of those types, but we cannot emit
+    // thoes more refined types.
```
kripken, line 1398:
```suggestion
    // those more refined types.
```

test/lit/downgrade-reftypes.wast
```diff
@@ -0,0 +1,61 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py and should not be edited.
+
+;; Write to a binary, lowering away refined GC types.
+;; RUN: wasm-as %s -all --disable-gc -g -o %s.wasm
+
+;; Read it back and verify that the types were lowered away.
+;; RUN: wasm-dis %s.wasm -all -o - | filecheck %s
```
kripken, line 7: You can just do `wasm-opt --roundtrip` for this kind of thing.

kripken (approved):
lgtm % (comments + fuzzing)

test/lit/downgrade-reftypes.wast
kripken: Though maybe as it is is more explicit? I don't feel strongly.

tlively: In particular I want to write without GC enabled and then read with GC enabled. I don't think --roundtrip would let me do that.

kripken: Ah, good point...

PR #5744, opened 2023-05-26
Move casts which are immediate children of local.gets to earlier local.gets

zm2he:
In the `OptimizeCasts` pass, it is useful to move more refined casts as early as possible without causing side-effects. This will allow such casts to potentially trap earlier, and will allow the `OptimizeCasts` pass to use more refined casts earlier. This change allows a more refined cast to be duplicated at an earlier `local.get` expression. The later instance of the cast will then be eliminated in a later optimization pass.

For example, if we have the following instructions:
```
(drop
  (local.get $x)
)
(drop
  (ref.cast $A
    (local.get $x)
)
(drop
  (ref.cast $B
    (local.get $x)
  )
)
```
Where `$B` is a sublcass of `$A`, we can convert this to:
```
(drop
  (ref.cast $B
    (local.get $x)
  )
)
(drop
  (ref.cast $A
    (local.get $x)
)
(drop
  (ref.cast $B
    (local.get $x)
  )
)
```
Currently this change only moves casts whose immediate value (i.e. child) is a `local.get`. Furthermore, the change current allows only one cast to be moved up to an earlier `local.get` location. `RefCast`s are given precedence. 

This is implemented as a separate walker traversal before the existing `OptimizeCast` pass. `EffectAnalyzers` are used to analyze the expressions between a cast an a `local.get` we want to move to for side effects if we move the cast up.

### DIFF
--- a/src/passes/OptimizeCasts.cpp
+++ b/src/passes/OptimizeCasts.cpp
@@ -89,9 +89,9 @@
 //    (ref.cast .. (local.get $ref))
 //  )
 //
-// TODO: 1. Move casts earlier in a basic block as well, at least in
-//          traps-never-happen mode where we can assume they never fail, and
-//          perhaps in other situations too.
+// TODO: 1. Check if casts whose fallthroughs, but not immediate falthroughs 
+//          are local.gets can be moved earlier, and also see if we can move
+//          multiple casts to the same local.get.
 // TODO: 2. Look past individual basic blocks? This may be worth considering
 //          given the pattern of a cast appearing in an if condition that is
 //          then used in an if arm, for example, where simple dominance shows
--- a/src/passes/OptimizeCasts.cpp
+++ b/src/passes/OptimizeCasts.cpp
@@ -90,7 +90,7 @@
 //  )
 //
 // Note that right now, we only consider RefAs with op RefAsNonNull as a cast.
-// RefAS with ExternInternalize and ExternExternalize are not considered casts
+// RefAs with ExternInternalize and ExternExternalize are not considered casts
 // when obtaining fallthroughs, and so are ignored.
 //
 // TODO: 1. Look past individual basic blocks? This may be worth considering
@@ -129,6 +129,10 @@ struct RefCastInfo {
   // Indicate whether the cast is already at the target. If it is, then
   // the information is just for determining if a potential cast is less
   // refined than the existing cast, and will be removed later.
+  //
+  // Note that we track a cast already on the get since we only want to move
+  // better casts to it: if the best cast is already on the get, there is no
+  // work to do.
   bool atTarget;
 
   RefCastInfo() {}
@@ -151,8 +155,6 @@ struct EarlyCastFinder
   std::vector<LocalGet*> earliestRefAsReachable;
 
   // Used to analyze expressions to see if casts can be moved past them.
-  RefCast dummyRefCast;
-  RefAs dummyRefAs;
   EffectAnalyzer testRefCast;
   EffectAnalyzer testRefAs;
 
@@ -165,7 +167,7 @@ struct EarlyCastFinder
   std::unordered_map<LocalGet*, RefCastInfo> refCastToApply;
 
   // Maps local.gets to a RefAs to move to it. This is the final result.
-  // As right now only RefAsNonNull is the only non Extern cast, we only
+  // As of right now only RefAsNonNull is the only non-extern cast, so we only
   // have one type of RefAs cast to move. The boolean indicates if the
   // ref.as_non_null is already at the local.get, or whether we actually
   // need to move it.
@@ -175,13 +177,11 @@ struct EarlyCastFinder
     : options(options), numLocals(func->getNumLocals()),
       earliestRefCastReachable(func->getNumLocals(), nullptr),
       earliestRefAsReachable(func->getNumLocals(), nullptr),
-      dummyRefCast(module->allocator), dummyRefAs(module->allocator),
       testRefCast(options, *module), testRefAs(options, *module) {
 
-    // Only RefAsNonNull produces implicit traps, so we use this when analyzing
-    // RefAs. ExternInternalize and ExternExternalize ops can actually be moved
-    // further since they are infallible, but we don't have special cases for
-    // them currently.
+    // TODO: generalize this when we handle more than RefAsNonNull
+    RefCast dummyRefCast(module->allocator);
+    RefAs dummyRefAs(module->allocator);
     dummyRefAs.op = RefAsNonNull;
 
     testRefCast.visit(&dummyRefCast);
@@ -215,6 +215,7 @@ struct EarlyCastFinder
 
   void visitLocalSet(LocalSet* curr) {
     visitExpression(curr);
+
     earliestRefCastReachable[curr->index] = nullptr;
     earliestRefAsReachable[curr->index] = nullptr;
   }
@@ -469,12 +470,12 @@ struct OptimizeCasts : public WalkerPass<PostWalker<OptimizeCasts>> {
 
     bool castsToMove = earlyCastFinder.hasCastsToMove();
     if (castsToMove) {
-      // Duplicate casts to earlier locations if possible
+      // Duplicate casts to earlier locations if possible.
       EarlyCastApplier earlyCastApplier(earlyCastFinder);
       earlyCastApplier.walkFunctionInModule(func, getModule());
 
       // Adding more casts causes types to be refined, that should be
-      // propagated. Especially those of nested casts.
+      // propagated.
       ReFinalize().walkFunctionInModule(func, getModule());
     }
 
--- a/src/passes/OptimizeCasts.cpp
+++ b/src/passes/OptimizeCasts.cpp
@@ -89,7 +89,7 @@
 //    (ref.cast .. (local.get $ref))
 //  )
 //
-// TODO: 1. Check if casts whose fallthroughs, but not immediate falthroughs 
+// TODO: 1. Check if casts whose fallthroughs, but not immediate fallthroughs
 //          are local.gets can be moved earlier, and also see if we can move
 //          multiple casts to the same local.get.
 // TODO: 2. Look past individual basic blocks? This may be worth considering
--- a/src/passes/OptimizeCasts.cpp
+++ b/src/passes/OptimizeCasts.cpp
@@ -121,24 +121,17 @@ namespace wasm {
 
 namespace {
 
-// Contains information about a RefCast we want to move to some target.
+// Contains information about a RefCast we want to move to a target LocalGet.
 struct RefCastInfo {
-  Type type;
-  RefCast::Safety safety;
+  LocalGet* target = nullptr;
+  RefCast* bestCast = nullptr;
+};
 
-  // Indicate whether the cast is already at the target. If it is, then
-  // the information is just for determining if a potential cast is less
-  // refined than the existing cast, and will be removed later.
-  //
-  // Note that we track a cast already on the get since we only want to move
-  // better casts to it: if the best cast is already on the get, there is no
-  // work to do.
-  bool atTarget;
-
-  RefCastInfo() {}
-
-  RefCastInfo(Type type, RefCast::Safety safety, bool atTarget)
-    : type(type), safety(safety), atTarget(atTarget) {}
+// Contains information about a RefAs we want to move to a target LocalGet.
+// Currently only RefAsNonNull will be moved.
+struct RefAsInfo {
+  LocalGet* target = nullptr;
+  RefAs* bestCast = nullptr;
 };
 
 // Find a cast to move earlier to another local.get. More refined subtypes are
@@ -150,34 +143,38 @@ struct EarlyCastFinder
   size_t numLocals;
 
   // Tracks the current earliest local.get that we can move a cast to without
-  // side-effects.
-  std::vector<LocalGet*> earliestRefCastReachable;
-  std::vector<LocalGet*> earliestRefAsReachable;
+  // side-effects, and the most refined cast that we can move to it (could be
+  // already at the earliest local.get).
+  //
+  // Note that we track a cast already on the get since we only want to move
+  // better casts to it: if the best cast is already on the get, there is no
+  // work to do.
+  std::vector<RefCastInfo> currRefCastMove;
+  std::vector<RefAsInfo> currRefAsMove;
 
   // Used to analyze expressions to see if casts can be moved past them.
   EffectAnalyzer testRefCast;
   EffectAnalyzer testRefAs;
 
-  // Maps local.gets to a RefCast to move to it. This is the final result.
-  // We currently only support moving one RefCast to a local.get because
-  // if a local.get is cast to multiple types which are not in a direct
-  // subtype relationship (i.e. one is not a subtype of the other), then
-  // a trap is inevitable, and we assume this would already be optimized
-  // away beforehand.
-  std::unordered_map<LocalGet*, RefCastInfo> refCastToApply;
+  // Maps LocalGets to the most refined RefCast to move to it, to be used by the
+  // EarlyCastApplier. If the most refined RefCast is already at the desired
+  // LocalGet, it does not appear here. In the normal case, only one RefCast
+  // needs to be moved to a LocalGet. If a LocalGet is cast to multiple types
+  // which are not subtypes of each other the a trap is inevitable, and we
+  // assume this would already be optimized away beforehand, so we don't care
+  // about this special case.
+  std::unordered_map<LocalGet*, RefCast*> refCastToApply;
 
-  // Maps local.gets to a RefAs to move to it. This is the final result.
-  // As of right now only RefAsNonNull is the only non-extern cast, so we only
-  // have one type of RefAs cast to move. The boolean indicates if the
-  // ref.as_non_null is already at the local.get, or whether we actually
-  // need to move it.
-  std::unordered_map<LocalGet*, bool> refAsToApply;
+  // Maps LocalGets to a RefAs to move to it, to be used by the
+  // EarlyCastApplier. As of right now RefAsNonNull is the only non-extern cast,
+  // so we only have one type of RefAs cast to move.
+  std::unordered_map<LocalGet*, RefAs*> refAsToApply;
 
   EarlyCastFinder(PassOptions options, Module* module, Function* func)
     : options(options), numLocals(func->getNumLocals()),
-      earliestRefCastReachable(func->getNumLocals(), nullptr),
-      earliestRefAsReachable(func->getNumLocals(), nullptr),
-      testRefCast(options, *module), testRefAs(options, *module) {
+      currRefCastMove(func->getNumLocals()),
+      currRefAsMove(func->getNumLocals()), testRefCast(options, *module),
+      testRefAs(options, *module) {
 
     // TODO: generalize this when we handle more than RefAsNonNull
     RefCast dummyRefCast(module->allocator);
@@ -188,10 +185,52 @@ struct EarlyCastFinder
     testRefAs.visit(&dummyRefAs);
   }
 
+  void flushRefCastResult(size_t index, Module& module) {
+    if (currRefCastMove[index].target) {
+      if (currRefCastMove[index].bestCast) {
+        // If the fallthrough is equal to the target, this means the cast is
+        // already at the target local.get and doesn't need to be duplicated
+        // again.
+        auto* fallthrough = Properties::getFallthrough(
+          currRefCastMove[index].bestCast, options, module);
+        if (fallthrough != currRefCastMove[index].target) {
+          refCastToApply[currRefCastMove[index].target] =
+            currRefCastMove[index].bestCast;
+        }
+        currRefCastMove[index].bestCast = nullptr;
+      }
+      currRefCastMove[index].target = nullptr;
+    }
+  }
+
+  void flushRefAsResult(size_t index, Module& module) {
+    if (currRefAsMove[index].target) {
+      if (currRefAsMove[index].bestCast) {
+        // As in flushRefCastResult, we need to check if the cast is already at
+        // the target and thus does not need to be moved.
+        auto* fallthrough = Properties::getFallthrough(
+          currRefAsMove[index].bestCast, options, *getModule());
+        if (fallthrough != currRefAsMove[index].target) {
+          refAsToApply[currRefAsMove[index].target] =
+            currRefAsMove[index].bestCast;
+        }
+        currRefAsMove[index].bestCast = nullptr;
+      }
+      currRefAsMove[index].target = nullptr;
+    }
+  }
+
   static void doNoteNonLinear(EarlyCastFinder* self, Expression**) {
     for (size_t i = 0; i < self->numLocals; i++) {
-      self->earliestRefCastReachable[i] = nullptr;
-      self->earliestRefAsReachable[i] = nullptr;
+      self->flushRefCastResult(i, *self->getModule());
+      self->flushRefAsResult(i, *self->getModule());
+    }
+  }
+
+  void visitFunction(Function* curr) {
+    for (size_t i = 0; i < numLocals; i++) {
+      flushRefCastResult(i, *getModule());
+      flushRefAsResult(i, *getModule());
     }
   }
 
@@ -202,13 +241,13 @@ struct EarlyCastFinder
 
     if (testRefCast.invalidates(currAnalyzer)) {
       for (size_t i = 0; i < numLocals; i++) {
-        earliestRefCastReachable[i] = nullptr;
+        flushRefCastResult(i, *getModule());
       }
     }
 
     if (testRefAs.invalidates(currAnalyzer)) {
       for (size_t i = 0; i < numLocals; i++) {
-        earliestRefAsReachable[i] = nullptr;
+        flushRefAsResult(i, *getModule());
       }
     }
   }
@@ -216,39 +255,35 @@ struct EarlyCastFinder
   void visitLocalSet(LocalSet* curr) {
     visitExpression(curr);
 
-    earliestRefCastReachable[curr->index] = nullptr;
-    earliestRefAsReachable[curr->index] = nullptr;
+    flushRefCastResult(curr->index, *getModule());
+    flushRefAsResult(curr->index, *getModule());
   }
 
   void visitLocalGet(LocalGet* curr) {
     visitExpression(curr);
 
-    if (!earliestRefCastReachable[curr->index]) {
-      earliestRefCastReachable[curr->index] = curr;
+    if (!currRefCastMove[curr->index].target) {
+      currRefCastMove[curr->index].target = curr;
     }
 
-    if (!earliestRefAsReachable[curr->index]) {
-      earliestRefAsReachable[curr->index] = curr;
+    if (!currRefAsMove[curr->index].target) {
+      currRefAsMove[curr->index].target = curr;
     }
   }
 
   void visitRefAs(RefAs* curr) {
     visitExpression(curr);
 
-    // Right now, since all other ops are not casts, we only care about
-    // RefAsNonNull If more cast ops are added, the program must be changed to
-    // accomodate more than one type of op.
+    // TODO: support more than RefAsNonNull
     if (curr->op != RefAsNonNull) {
       return;
     }
 
     auto* fallthrough = Properties::getFallthrough(curr, options, *getModule());
-
     if (auto* get = fallthrough->dynCast<LocalGet>()) {
-      auto castIter = refAsToApply.find(earliestRefAsReachable[get->index]);
-      if (castIter == refAsToApply.end()) {
-        refAsToApply.emplace(earliestRefAsReachable[get->index],
-                             get == earliestRefAsReachable[get->index]);
+      RefAsInfo& bestMove = currRefAsMove[get->index];
+      if (bestMove.target && !bestMove.bestCast) {
+        bestMove.bestCast = curr;
       }
     }
   }
@@ -285,29 +320,28 @@ struct EarlyCastFinder
     // This would produce a trap and should already be optimized away
     // beforehand.
     //
-    // Note that if the best cast is already at the target location, we will
-    // keep it for comparison purposes but remove it once the traversal is done.
+    // If the best cast is already at the target location, we will still track
+    // it in currRefCastMove to see if we can obtain a better cast. However, we
+    // won't flush it.
 
     auto* fallthrough = Properties::getFallthrough(curr, options, *getModule());
     if (auto* get = fallthrough->dynCast<LocalGet>()) {
-      auto castIter = refCastToApply.find(earliestRefCastReachable[get->index]);
-
-      if (castIter == refCastToApply.end()) {
-        RefCastInfo newCast(curr->type,
-                            curr->safety,
-                            get == earliestRefCastReachable[get->index]);
-        refCastToApply.emplace(earliestRefCastReachable[get->index], newCast);
-      } else if (curr->type != castIter->second.type &&
-                 Type::isSubType(curr->type, castIter->second.type)) {
-        castIter->second =
-          RefCastInfo(curr->type,
-                      curr->safety,
-                      get == earliestRefCastReachable[get->index]);
-      } else if (curr->type == castIter->second.type &&
-                 curr->safety == RefCast::Safety::Safe &&
-                 castIter->second.safety == RefCast::Safety::Unsafe) {
-        castIter->second.safety = curr->safety;
-        castIter->second.atTarget = get == earliestRefCastReachable[get->index];
+      RefCastInfo& bestMove = currRefCastMove[get->index];
+      if (bestMove.target) {
+        if (!bestMove.bestCast) {
+          // If there isn't any other cast to move, the current cast is the best
+          bestMove.bestCast = curr;
+        } else if (bestMove.bestCast->type != curr->type &&
+                   Type::isSubType(curr->type, bestMove.bestCast->type)) {
+          // If the current cast is more refined than the best cast to move,
+          // change the best cast to move
+          bestMove.bestCast = curr;
+        }
+        // We don't care about the safety of the cast at present. If there are
+        // two casts with the same type one being safe and one being unsafe, the
+        // first cast that we visit will be chosen to be moved. Perhaps in the
+        // future we can consider prioritizing unsafe casts over safe ones since
+        // users may be more interested in that.
       }
     }
   }
@@ -328,25 +362,23 @@ struct EarlyCastApplier : public PostWalker<EarlyCastApplier> {
   EarlyCastApplier(EarlyCastFinder& finder) : finder(finder) {}
 
   // RefCast casts are applied before RefAs casts. If there are multiple
-  // casts to apply to a location, they are nested within one another.
+  // casts to apply to a location, they are nested within one another. Only
+  // at most one RefCast and at most one RefAs can be applied.
   void visitLocalGet(LocalGet* curr) {
     Expression* currPtr = curr;
 
     auto refCastIter = finder.refCastToApply.find(curr);
     if (refCastIter != finder.refCastToApply.end()) {
-      if (!refCastIter->second.atTarget) {
-        currPtr = replaceCurrent(Builder(*getModule())
-                                   .makeRefCast(currPtr,
-                                                refCastIter->second.type,
-                                                refCastIter->second.safety));
-      }
+      currPtr = replaceCurrent(Builder(*getModule())
+                                 .makeRefCast(currPtr,
+                                              refCastIter->second->type,
+                                              refCastIter->second->safety));
     }
 
     auto refAsIter = finder.refAsToApply.find(curr);
     if (refAsIter != finder.refAsToApply.end()) {
-      if (!refAsIter->second) {
-        replaceCurrent(Builder(*getModule()).makeRefAs(RefAsNonNull, currPtr));
-      }
+      replaceCurrent(
+        Builder(*getModule()).makeRefAs(refAsIter->second->op, currPtr));
     }
   }
 };
@@ -464,12 +496,11 @@ struct OptimizeCasts : public WalkerPass<PostWalker<OptimizeCasts>> {
       return;
     }
 
-    // Look for casts which can be moved earlier
+    // Look for casts which can be moved earlier.
     EarlyCastFinder earlyCastFinder(getPassOptions(), getModule(), func);
     earlyCastFinder.walkFunctionInModule(func, getModule());
 
-    bool castsToMove = earlyCastFinder.hasCastsToMove();
-    if (castsToMove) {
+    if (earlyCastFinder.hasCastsToMove()) {
       // Duplicate casts to earlier locations if possible.
       EarlyCastApplier earlyCastApplier(earlyCastFinder);
       earlyCastApplier.walkFunctionInModule(func, getModule());
@@ -484,8 +515,7 @@ struct OptimizeCasts : public WalkerPass<PostWalker<OptimizeCasts>> {
     finder.options = getPassOptions();
     finder.walkFunctionInModule(func, getModule());
 
-    bool castsToOptimize = !finder.lessCastedGets.empty();
-    if (castsToOptimize) {
+    if (!finder.lessCastedGets.empty()) {
       // Apply the requests: use the best casts.
       FindingApplier applier(finder);
       applier.walkFunctionInModule(func, getModule());
--- a/src/passes/OptimizeCasts.cpp
+++ b/src/passes/OptimizeCasts.cpp
@@ -176,7 +176,7 @@ struct EarlyCastFinder
       currRefAsMove(func->getNumLocals()), testRefCast(options, *module),
       testRefAs(options, *module) {
 
-    // TODO: generalize this when we handle more than RefAsNonNull
+    // TODO: generalize this when we handle more than RefAsNonNull.
     RefCast dummyRefCast(module->allocator);
     RefAs dummyRefAs(module->allocator);
     dummyRefAs.op = RefAsNonNull;
@@ -273,8 +273,8 @@ struct EarlyCastFinder
     }
 
     // As we only move RefAsNonNull RefAs casts right now, we should
-    // ignore a LocalGet if the type is already non-nullable, and
-    // adding an extra ref.as_non_null has no effect
+    // ignore a LocalGet if the type is already non-nullable, as
+    // adding an extra ref.as_non_null has no effect.
     if (!currRefAsMove[curr->index].target && curr->type.isNullable()) {
       currRefAsMove[curr->index].target = curr;
     }
@@ -283,7 +283,7 @@ struct EarlyCastFinder
   void visitRefAs(RefAs* curr) {
     visitExpression(curr);
 
-    // TODO: support more than RefAsNonNull
+    // TODO: support more than RefAsNonNull.
     if (curr->op != RefAsNonNull) {
       return;
     }
--- a/src/passes/OptimizeCasts.cpp
+++ b/src/passes/OptimizeCasts.cpp
@@ -336,7 +336,7 @@ struct EarlyCastFinder
     auto* fallthrough = Properties::getFallthrough(curr, options, *getModule());
     if (auto* get = fallthrough->dynCast<LocalGet>()) {
       auto& bestMove = currRefCastMove[get->index];
-      // Do not move a cast if it's type is not related to the target
+      // Do not move a cast if its type is not related to the target
       // local.get's type (i.e. not in a subtyping relationship). Otherwise
       // a type error will occur. Also, if the target local.get's type is
       // already more refined than this current cast, there is no point in
--- a/src/passes/OptimizeCasts.cpp
+++ b/src/passes/OptimizeCasts.cpp
@@ -272,7 +272,10 @@ struct EarlyCastFinder
       currRefCastMove[curr->index].target = curr;
     }
 
-    if (!currRefAsMove[curr->index].target) {
+    // As we only move RefAsNonNull RefAs casts right now, we should
+    // ignore a LocalGet if the type is already non-nullable, and
+    // adding an extra ref.as_non_null has no effect
+    if (!currRefAsMove[curr->index].target && curr->type.isNullable()) {
       currRefAsMove[curr->index].target = curr;
     }
   }
--- a/src/passes/OptimizeCasts.cpp
+++ b/src/passes/OptimizeCasts.cpp
@@ -336,7 +336,13 @@ struct EarlyCastFinder
     auto* fallthrough = Properties::getFallthrough(curr, options, *getModule());
     if (auto* get = fallthrough->dynCast<LocalGet>()) {
       auto& bestMove = currRefCastMove[get->index];
-      if (bestMove.target) {
+      // Do not move a cast if it's type is not related to the target
+      // local.get's type (i.e. not in a subtyping relationship). Otherwise
+      // a type error will occur. Also, if the target local.get's type is
+      // already more refined than this current cast, there is no point in
+      // moving it.
+      if (bestMove.target && bestMove.target->type != curr->type &&
+          Type::isSubType(curr->type, bestMove.target->type)) {
         if (!bestMove.bestCast) {
           // If there isn't any other cast to move, the current cast is the
           // best.
--- a/src/passes/OptimizeCasts.cpp
+++ b/src/passes/OptimizeCasts.cpp
@@ -89,14 +89,11 @@
 //    (ref.cast .. (local.get $ref))
 //  )
 //
-// TODO: 1. Check if casts whose fallthroughs, but not immediate fallthroughs
-//          are local.gets can be moved earlier, and also see if we can move
-//          multiple casts to the same local.get.
-// TODO: 2. Look past individual basic blocks? This may be worth considering
+// TODO: 1. Look past individual basic blocks? This may be worth considering
 //          given the pattern of a cast appearing in an if condition that is
 //          then used in an if arm, for example, where simple dominance shows
 //          the cast can be reused.
-// TODO: 3. Look at LocalSet as well and not just Get. That would add some
+// TODO: 2. Look at LocalSet as well and not just Get. That would add some
 //          overlap with the other passes mentioned above (SimplifyLocals and
 //          RedundantSetElimination also track sets and can switch a get to use
 //          a better set's index when that refines the type). But once we do the
@@ -134,12 +131,6 @@ struct RefCastInfo {
 
   RefCastInfo(Type type, RefCast::Safety safety, bool atTarget)
     : type(type), safety(safety), atTarget(atTarget) {}
-
-  void modify(Type newType, RefCast::Safety newSafety, bool newAtTarget) {
-    type = newType;
-    safety = newSafety;
-    atTarget = newAtTarget;
-  }
 };
 
 // Contains information about a RefAs we want to move to some target.
@@ -152,7 +143,7 @@ struct RefAsInfo {
   RefAsInfo(RefAsOp op, bool atTarget) : op(op), atTarget(atTarget) {}
 };
 
-// Find casts to move earlier to another local.gets. More refined subtypes are
+// Find casts to move earlier to another local.get. More refined subtypes are
 // chosen over less refined ones.
 struct EarlyCastFinder
   : public LinearExecutionWalker<EarlyCastFinder,
@@ -160,7 +151,7 @@ struct EarlyCastFinder
   PassOptions options;
   size_t numLocals;
 
-  // For each index, tracks a list of most refined casts that we can move to the
+  // For each local index, tracks a list of most refined casts that we can move to the
   // earliest reachable local.get.
   std::vector<SmallVector<RefCastInfo, 5>> movedRefCast;
   std::vector<SmallVector<RefAsInfo, 3>> movedRefAs;
@@ -193,7 +184,7 @@ struct EarlyCastFinder
 
     // Only RefAsNonNull produces implicit traps, so we use this when analyzing
     // RefAs. ExternInternalize and ExternExternalize ops can actually be moved
-    // further since they are infalible, but we don't have special cases for
+    // further since they are infallible, but we don't have special cases for
     // them currently.
     dummyRefAs.op = RefAsNonNull;
 
@@ -300,6 +291,39 @@ struct EarlyCastFinder
 
   void visitRefCast(RefCast* curr) {
     visitExpression(curr);
+
+    // Using fallthroughs here is fine due to the following cases.
+    // Suppose we have types $A->$B->$C (where $C is the most refined)
+    // and $D, which is an unrelated type.
+    // Case 1:
+    // (ref.cast $A (ref.cast $C (local.get $x)))
+    //
+    // ref.cast $C is initially chosen for $x. Then we consider ref.cast $A.
+    // Since $A is less refined than $C, we ignore it.
+    //
+    // Case 2:
+    // (ref.cast $C (ref.cast $A (local.get $x)))
+    //
+    // ref.cast $A is initially chosen for $x. Then we consider ref.cast $C,
+    // which is more refined than ref.cast $A, so we replace it with ref.cast $C.
+    //
+    // Case 3:
+    // (ref.cast $B (ref.cast $B (local.get $x)))
+    //
+    // We initially choose to move the inner ref.cast $B. When we consider the
+    // outer ref.cast $B, we can see that it has the same type as tge existing
+    // ref.cast $B, so we ignore it.
+    //
+    // Case 4:
+    // (ref.cast $D (ref.cast $C (local.get $x)))
+    //
+    // We first choose to move ref.cast $C. Since $D is neither a subtype nor
+    // a supertype of $C (if $B->$D, this also applies), then we add ref.cast $D
+    // as an additional cast to apply. ref.cast $C is nested in ref.cast $D.
+    //
+    // Note that if the best cast is already at the target location, we will
+    // keep it for comparison purposes but remove it once the traversal is done.
+
     auto* fallthrough = Properties::getFallthrough(curr, options, *getModule());
     if (auto* get = fallthrough->dynCast<LocalGet>()) {
 
@@ -313,9 +337,7 @@ struct EarlyCastFinder
         if (curr->type != currInfo.type) {
           if (Type::isSubType(curr->type, currInfo.type)) {
             isNewType = false;
-            currInfo.modify(curr->type,
-                            curr->safety,
-                            get == earliestRefCastReachable[get->index]);
+            currInfo = RefCastInfo(curr->type, curr->safety, get == earliestRefCastReachable[get->index]);
             break;
           } else if (Type::isSubType(currInfo.type, curr->type)) {
             isNewType = false;
@@ -511,10 +533,7 @@ struct OptimizeCasts : public WalkerPass<PostWalker<OptimizeCasts>> {
       EarlyCastApplier earlyCastApplier(earlyCastFinder);
       earlyCastApplier.walkFunctionInModule(func, getModule());
 
-      // LocalGet type changes must be propagated here as well. This
-      // is because moving casts earlier may produce nested casts
-      // which means that the actual type of the moved cast might
-      // not be accurate.
+      // Adding more casts causes types to be refined, that should be propagated. Especially those of nested casts.
       ReFinalize().walkFunctionInModule(func, getModule());
     }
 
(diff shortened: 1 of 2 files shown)


src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +120,256 @@ namespace wasm {
...
+  // Maps local.gets to the "best" RefCast to move to it.
+  std::unordered_map<Expression*, RefCast*> refCastToApply;
+
+  static void doNoteNonLinear(EarlyCastFinder* self, Expression**) {
+    self->castMoveSearchers.clear();
+  }
+
+  void visitExpression(Expression* curr) {
+    for (auto& indexSearcher : castMoveSearchers) {
+      indexSearcher.second.visitExpression(curr);
```
kripken, line 280:
```suggestion
    for (auto& [_, searcher] : castMoveSearchers) {
      searcher.visitExpression(curr);
```
I think this style is clearer, and we use it a lot. Note how it gets rid of the `.second` which can be a little unclear when reading code.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +120,256 @@ namespace wasm {
...
+    }
+  }
+
+  void visitLocalSet(LocalSet* curr) {
+    for (auto& indexSearcher : castMoveSearchers) {
+      indexSearcher.second.visitExpression(curr);
+    }
```
kripken, line 287: You can call `visitExpression()` to avoid duplicating this code, perhaps with a comment to the effect of "do the default behaviors we perform on any expression"

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +120,256 @@ namespace wasm {
...
+
+  void visitLocalGet(LocalGet* curr) {
+    for (auto& indexSearcher : castMoveSearchers) {
+      indexSearcher.second.visitExpression(curr);
+    }
+    auto iter = castMoveSearchers.find(curr->index);
+    if (iter == castMoveSearchers.end()) {
+      RefCast testRefCast(getModule()->allocator);
+      EffectAnalyzer efa(options, *getModule());
+      efa.visit(&testRefCast);
```
kripken, line 302: Reading this code, it's not obvious to me why we compute the effects of a RefCast when we are processing a LocalGet, so I think that is worth a comment at least.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +120,256 @@ namespace wasm {
...
+      IndexEarlyCastSearcher searcher(options, *getModule());
+      searcher.visitLocalGet(curr, *getModule());
+      castMoveSearchers.emplace(curr->index, searcher);
```
kripken, line 305: IIUC the idea here is to create a Searcher for each local get index we see? I am a little worried about efficiency here, as LocalGets are among the most common instructions, and most of the time this optimization would not find work to do. And the Searcher class above looks like it has a significant size, so creating it is not very cheap.

This might be ok as it is. But perhaps there is a more efficient way? I am thinking, vaguely, of something like this:

* When we see a LocalGet, we just note the existence of the get for that index, perhaps adding to a `std::vector<SmallVector<Info>>` in which `vector[index]` is the vector of info for each get (which is typically small, so a SmallVector is more efficient). The info might be the get + the type it is cast to if it has a parent that is a cast - I hope we don't need any more.
* When we see a trap effect, we clear that vector, since we can't move past a trap. That might hard-code the fact that casts have only that effect, but I think it's ok if the efficiency gain is significant, and if we have an assertion elsewhere that verifies that is the only effect.
* When we see a cast of a get, we look in that vector and see if we can improve things, and if so we make a note for later.

I haven't thought that through entirely, but what do you think of the direction?

kripken (commented):
Very nice! I read around half the code, and have some thoughts on the overall design, but this is looking very promising.

(The UBSan error may be real. If so I'd guess it's perhaps related to `sampleRefAs` but I don't have a more specific idea. Hopefully in a debug build + ubsan you can get a stack trace locally and it's easy to figure out. But based on the other comments perhaps we can avoid using `sampleRefAs` so I'd ignore it for now.)

src/passes/OptimizeCasts.cpp
zm2he: It seems like this was something from an earlier iteration that I forgot to delete. We don't need to compute the effects of a RefCast here. I will remove it in the next commit.

src/passes/OptimizeCasts.cpp
zm2he: Done in new change.

src/passes/OptimizeCasts.cpp
```diff
@@ -215,7 +483,21 @@ struct OptimizeCasts : public WalkerPass<PostWalker<OptimizeCasts>> {
...
+    if (!earlyCastFinder.refAsToApply.empty() ||
+        !earlyCastFinder.refCastToApply.empty()) {
+      EarlyCastApplier earlyCastApplier(earlyCastFinder);
+      earlyCastApplier.walkFunctionInModule(func, getModule());
+    }
+
+    // Find the best casts that we want to use.
     BestCastFinder finder;
     finder.options = getPassOptions();
     finder.walkFunctionInModule(func, getModule());
```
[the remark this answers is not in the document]
zm2he: Done in new change.

src/passes/OptimizeCasts.cpp
```diff
@@ -215,7 +483,21 @@ struct OptimizeCasts : public WalkerPass<PostWalker<OptimizeCasts>> {
...
 
-    // First, find the best casts that we want to use.
+    // Look for casts which can be moved earlier
+    EarlyCastFinder earlyCastFinder;
+    earlyCastFinder.options = getPassOptions();
+    earlyCastFinder.walkFunctionInModule(func, getModule());
+    earlyCastFinder.collectCastMoves();
+
+    // If there are casts which can be moved earlier, duplicate them to those
+    // locations
```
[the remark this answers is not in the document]
zm2he: Comment removed.

src/passes/OptimizeCasts.cpp
zm2he: Not in new change.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +120,273 @@ namespace wasm {
...
+          movedRefAs[get->index].emplace_back(
+            curr->op, get == earliestRefAsReachable[get->index]);
+        }
+      }
+    }
+  }
+
+  void visitRefCast(RefCast* curr) {
+    visitExpression(curr);
+    auto* fallthrough = Properties::getFallthrough(curr, options, *getModule());
```
kripken, line 303: After offline discussion I think fallthrough here is ok, but please add a detailed comment summarizing what we talked about (i.e., why it is ok - why the various possible situations end up not as problems).

src/passes/OptimizeCasts.cpp
```diff
@@ -72,9 +72,26 @@
...
+//  )
+//
+// could be converted to:
+//
+//  (some.operation
+//    (ref.cast (local.get $ref))
+//    (ref.cast .. (local.get $ref))
+//  )
+//
+// TODO: 1. Check if casts whose fallthroughs, but not immediate fallthroughs
```
kripken, line 92: The current code seems to do this TODO?

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +120,273 @@ namespace wasm {
...
+  RefCastInfo() {}
+
+  RefCastInfo(Type type, RefCast::Safety safety, bool atTarget)
+    : type(type), safety(safety), atTarget(atTarget) {}
+
+  void modify(Type newType, RefCast::Safety newSafety, bool newAtTarget) {
+    type = newType;
+    safety = newSafety;
+    atTarget = newAtTarget;
+  }
```
kripken, line 142: I don't think this is needed, since you can replace `foo.modify(x, y, z)` with `foo = RefCastInfo(x, y, z)`.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +120,273 @@ namespace wasm {
...
+struct RefAsInfo {
+  RefAsOp op;
+  bool atTarget;
+
+  RefAsInfo() {}
+
+  RefAsInfo(RefAsOp op, bool atTarget) : op(op), atTarget(atTarget) {}
+};
+
+// Find casts to move earlier to another local.gets. More refined subtypes are
```
kripken, line 155:
```suggestion
// Find casts to move earlier to other local.gets. More refined subtypes are
```
or
```suggestion
// Find casts to move earlier to another local.get. More refined subtypes are
```

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +120,273 @@ namespace wasm {
...
+// chosen over less refined ones.
+struct EarlyCastFinder
+  : public LinearExecutionWalker<EarlyCastFinder,
+                                 UnifiedExpressionVisitor<EarlyCastFinder>> {
+  PassOptions options;
+  size_t numLocals;
+
+  // For each index, tracks a list of most refined casts that we can move to the
```
kripken, line 163:
```suggestion
  // For each local index, tracks a list of most refined casts that we can move to the
```

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +120,273 @@ namespace wasm {
...
+  // earliest reachable local.get.
+  std::vector<SmallVector<RefCastInfo, 5>> movedRefCast;
+  std::vector<SmallVector<RefAsInfo, 3>> movedRefAs;
```
kripken, line 166: (why 5 and 3 and not the same value in both?)

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +120,273 @@ namespace wasm {
...
+    : options(options), numLocals(func->getNumLocals()),
+      movedRefCast(func->getNumLocals()), movedRefAs(func->getNumLocals()),
+      earliestRefCastReachable(func->getNumLocals(), nullptr),
+      earliestRefAsReachable(func->getNumLocals(), nullptr),
+      dummyRefCast(module->allocator), dummyRefAs(module->allocator),
+      testRefCast(options, *module), testRefAs(options, *module) {
+
+    // Only RefAsNonNull produces implicit traps, so we use this when analyzing
+    // RefAs. ExternInternalize and ExternExternalize ops can actually be moved
+    // further since they are infalible, but we don't have special cases for
```
kripken, line 196:
```suggestion
    // further since they are infallible, but we don't have special cases for
```

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +120,273 @@ namespace wasm {
...
+        for (size_t i = 0; i < movedRefAs[get->index].size(); i++) {
+          if (movedRefAs[get->index][i].op == curr->op) {
+            alreadyAdded = true;
+            break;
+          }
+        }
+
+        if (!alreadyAdded) {
+          movedRefAs[get->index].emplace_back(
+            curr->op, get == earliestRefAsReachable[get->index]);
```
kripken, line 295: Why doesn't this place add to the final output place, `refAsToApply`? I'm having a hard time understanding why it is queued here and processed later. Sorry if I'm missing something obvious.

(I guess queuing can help by letting us decide later what is more optimal. But also updating `refAsToApply` to something better if we find it seems equally good.)

src/passes/OptimizeCasts.cpp
zm2he: No particular reason. I picked 3 for `movedRefAs` because there are only 3 members of `RefAsOp`. I just picked 5 for `RefCast`, but 3 is probably also reasonable, since I assume there shouldn't be too many different types. If there is a way to specify this without hard-coding the values, let me know.

src/passes/OptimizeCasts.cpp
zm2he: You're right, it could be added immediately. I just followed the pattern in `visitRefCast`, where I queue it because there may be more refined casts later.  But for  a `RefAs`, since the operators are orthogonal to each other and neither is "better", this doesn't matter.

src/passes/OptimizeCasts.cpp
zm2he: Fixed.

src/passes/OptimizeCasts.cpp
zm2he: Fixed.

src/passes/OptimizeCasts.cpp
zm2he: Fixed.

src/passes/OptimizeCasts.cpp
zm2he: Removed.

src/passes/OptimizeCasts.cpp
zm2he: Removed.

src/passes/OptimizeCasts.cpp
zm2he: Added.

src/passes/OptimizeCasts.cpp
```diff
@@ -215,22 +500,38 @@ struct OptimizeCasts : public WalkerPass<PostWalker<OptimizeCasts>> {
...
+    bool castsToMove = earlyCastFinder.hasCastsToMove();
+    if (castsToMove) {
+      // Duplicate casts to earlier locations if possible
+      EarlyCastApplier earlyCastApplier(earlyCastFinder);
+      earlyCastApplier.walkFunctionInModule(func, getModule());
+
+      // LocalGet type changes must be propagated here as well. This
+      // is because moving casts earlier may produce nested casts
+      // which means that the actual type of the moved cast might
+      // not be accurate.
```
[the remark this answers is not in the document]
zm2he: Reworded.

src/passes/OptimizeCasts.cpp
kripken: Can't we do the same for visitRefCast? Again, if we see a more refined cast later can't be update the final output place as we go?

I mean, if we see
1. A LocalGet
2. A cast to $B
3. A cast to $C

Then at step 3 we can overwrite what step 2 wrote.

kripken (commented):
Nice work!

I finished reading and (I hope) understanding all the code this time.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,235 @@ namespace wasm {
...
+    //
+    // Note that if the best cast is already at the target location, we will
+    // keep it for comparison purposes but remove it once the traversal is done.
+
+    auto* fallthrough = Properties::getFallthrough(curr, options, *getModule());
+    if (auto* get = fallthrough->dynCast<LocalGet>()) {
+      auto castIter = refCastToApply.find(earliestRefCastReachable[get->index]);
+
+      if (castIter == refCastToApply.end()) {
+        RefCastInfo newCast(curr->type,
```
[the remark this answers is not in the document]
zm2he: @kripken For 3., should I prefer the experimental cast then? Or just ignore it?

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,235 @@ namespace wasm {
...
+struct EarlyCastFinder
+  : public LinearExecutionWalker<EarlyCastFinder,
+                                 UnifiedExpressionVisitor<EarlyCastFinder>> {
+  PassOptions options;
+  size_t numLocals;
+
+  // Tracks the current earliest local.get that we can move a cast to without
+  // side-effects.
+  std::vector<LocalGet*> earliestRefCastReachable;
+  std::vector<LocalGet*> earliestRefAsReachable;
```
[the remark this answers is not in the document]
zm2he: @kripken yes, `atTarget` would not be necessary, but wouldn't it save some extra `getFallthrough` calls?

src/passes/OptimizeCasts.cpp
```diff
@@ -215,22 +463,35 @@ struct OptimizeCasts : public WalkerPass<PostWalker<OptimizeCasts>> {
...
+    earlyCastFinder.walkFunctionInModule(func, getModule());
+
+    bool castsToMove = earlyCastFinder.hasCastsToMove();
+    if (castsToMove) {
+      // Duplicate casts to earlier locations if possible
+      EarlyCastApplier earlyCastApplier(earlyCastFinder);
+      earlyCastApplier.walkFunctionInModule(func, getModule());
+
+      // Adding more casts causes types to be refined, that should be
+      // propagated. Especially those of nested casts.
```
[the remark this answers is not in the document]
zm2he: I meant that nested casts would cause at least one of the casts types to become potentially inaccurate. But since now that I think of it, it happens to any expression, I should just remove this as you suggest.

src/passes/OptimizeCasts.cpp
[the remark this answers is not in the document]
kripken: I guess either way is ok. But it's worth a comment at least mentioning that the ordering between them is arbitrary, just so people reading the code don't wonder "this must be obvious because there's no comment, but I don't know why this is how it is"

src/passes/OptimizeCasts.cpp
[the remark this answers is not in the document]
kripken: Yes, it's a tradeoff. I'm actually not sure how the tradeoff works in terms of runtime performance, since I can see benefits both ways (more `getFallthrough` calls, as you said, vs. smaller data structures). Given that, I think smaller and clearer code seems best.

src/passes/OptimizeCasts.cpp
```diff
@@ -72,14 +72,32 @@
...
+//
+// could be converted to:
+//
+//  (some.operation
+//    (ref.cast (local.get $ref))
+//    (ref.cast .. (local.get $ref))
+//  )
+//
+// Note that right now, we only consider RefAs with op RefAsNonNull as a cast.
+// RefAS with ExternInternalize and ExternExternalize are not considered casts
```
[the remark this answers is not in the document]
zm2he: Done.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,235 @@ namespace wasm {
...
+  std::vector<LocalGet*> earliestRefCastReachable;
+  std::vector<LocalGet*> earliestRefAsReachable;
+
+  // Used to analyze expressions to see if casts can be moved past them.
+  RefCast dummyRefCast;
+  RefAs dummyRefAs;
+  EffectAnalyzer testRefCast;
+  EffectAnalyzer testRefAs;
+
+  // Maps local.gets to a RefCast to move to it. This is the final result.
```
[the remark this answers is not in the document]
zm2he: Done.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,235 @@ namespace wasm {
...
+  // We currently only support moving one RefCast to a local.get because
+  // if a local.get is cast to multiple types which are not in a direct
+  // subtype relationship (i.e. one is not a subtype of the other), then
+  // a trap is inevitable, and we assume this would already be optimized
+  // away beforehand.
+  std::unordered_map<LocalGet*, RefCastInfo> refCastToApply;
+
+  // Maps local.gets to a RefAs to move to it. This is the final result.
+  // As right now only RefAsNonNull is the only non Extern cast, we only
```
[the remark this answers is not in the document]
zm2he: Done.

src/passes/OptimizeCasts.cpp
```diff
@@ -215,22 +463,35 @@ struct OptimizeCasts : public WalkerPass<PostWalker<OptimizeCasts>> {
...
     }
 
-    // First, find the best casts that we want to use.
+    // Look for casts which can be moved earlier
+    EarlyCastFinder earlyCastFinder(getPassOptions(), getModule(), func);
+    earlyCastFinder.walkFunctionInModule(func, getModule());
+
+    bool castsToMove = earlyCastFinder.hasCastsToMove();
+    if (castsToMove) {
+      // Duplicate casts to earlier locations if possible
```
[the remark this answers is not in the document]
zm2he: Done.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,235 @@ namespace wasm {
...
 namespace {
 
+// Contains information about a RefCast we want to move to some target.
+struct RefCastInfo {
+  Type type;
+  RefCast::Safety safety;
+
+  // Indicate whether the cast is already at the target. If it is, then
+  // the information is just for determining if a potential cast is less
+  // refined than the existing cast, and will be removed later.
```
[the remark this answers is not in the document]
zm2he: Done.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,235 @@ namespace wasm {
...
+      earliestRefAsReachable[curr->index] = curr;
+    }
+  }
+
+  void visitRefAs(RefAs* curr) {
+    visitExpression(curr);
+
+    // Right now, since all other ops are not casts, we only care about
+    // RefAsNonNull If more cast ops are added, the program must be changed to
+    // accomodate more than one type of op.
```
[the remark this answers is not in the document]
zm2he: Done.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,235 @@ namespace wasm {
...
+
+    if (testRefAs.invalidates(currAnalyzer)) {
+      for (size_t i = 0; i < numLocals; i++) {
+        earliestRefAsReachable[i] = nullptr;
+      }
+    }
+  }
+
+  void visitLocalSet(LocalSet* curr) {
+    visitExpression(curr);
```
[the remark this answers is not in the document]
zm2he: Done.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,235 @@ namespace wasm {
...
+  size_t numLocals;
+
+  // Tracks the current earliest local.get that we can move a cast to without
+  // side-effects.
+  std::vector<LocalGet*> earliestRefCastReachable;
+  std::vector<LocalGet*> earliestRefAsReachable;
+
+  // Used to analyze expressions to see if casts can be moved past them.
+  RefCast dummyRefCast;
+  RefAs dummyRefAs;
```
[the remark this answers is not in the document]
zm2he: Done.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,235 @@ namespace wasm {
...
+    : options(options), numLocals(func->getNumLocals()),
+      earliestRefCastReachable(func->getNumLocals(), nullptr),
+      earliestRefAsReachable(func->getNumLocals(), nullptr),
+      dummyRefCast(module->allocator), dummyRefAs(module->allocator),
+      testRefCast(options, *module), testRefAs(options, *module) {
+
+    // Only RefAsNonNull produces implicit traps, so we use this when analyzing
+    // RefAs. ExternInternalize and ExternExternalize ops can actually be moved
+    // further since they are infallible, but we don't have special cases for
+    // them currently.
```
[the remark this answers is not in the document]
zm2he: Done.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,235 @@ namespace wasm {
...
+  // a trap is inevitable, and we assume this would already be optimized
+  // away beforehand.
+  std::unordered_map<LocalGet*, RefCastInfo> refCastToApply;
+
+  // Maps local.gets to a RefAs to move to it. This is the final result.
+  // As right now only RefAsNonNull is the only non Extern cast, we only
+  // have one type of RefAs cast to move. The boolean indicates if the
+  // ref.as_non_null is already at the local.get, or whether we actually
+  // need to move it.
+  std::unordered_map<LocalGet*, bool> refAsToApply;
```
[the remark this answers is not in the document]
zm2he: Bool replaced.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,235 @@ namespace wasm {
...
+  RefAs dummyRefAs;
+  EffectAnalyzer testRefCast;
+  EffectAnalyzer testRefAs;
+
+  // Maps local.gets to a RefCast to move to it. This is the final result.
+  // We currently only support moving one RefCast to a local.get because
+  // if a local.get is cast to multiple types which are not in a direct
+  // subtype relationship (i.e. one is not a subtype of the other), then
+  // a trap is inevitable, and we assume this would already be optimized
+  // away beforehand.
```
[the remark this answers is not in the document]
zm2he: Done.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,235 @@ namespace wasm {
...
+    // ref.cast $B, so we ignore it.
+    //
+    // Case 4:
+    // (ref.cast $D (ref.cast $C (local.get $x)))
+    //
+    // This would produce a trap and should already be optimized away
+    // beforehand.
+    //
+    // Note that if the best cast is already at the target location, we will
+    // keep it for comparison purposes but remove it once the traversal is done.
```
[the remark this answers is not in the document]
zm2he: Clarified.

src/passes/OptimizeCasts.cpp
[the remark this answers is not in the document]
zm2he: Removing special treatment of safety for now.

src/passes/OptimizeCasts.cpp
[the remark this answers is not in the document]
zm2he: Changes made.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,235 @@ namespace wasm {
...
+      return;
+    }
+
+    auto* fallthrough = Properties::getFallthrough(curr, options, *getModule());
+
+    if (auto* get = fallthrough->dynCast<LocalGet>()) {
+      auto castIter = refAsToApply.find(earliestRefAsReachable[get->index]);
+      if (castIter == refAsToApply.end()) {
+        refAsToApply.emplace(earliestRefAsReachable[get->index],
+                             get == earliestRefAsReachable[get->index]);
```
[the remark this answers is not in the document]
zm2he: Done.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,268 @@ namespace wasm {
...
+
+// Find a cast to move earlier to another local.get. More refined subtypes are
+// chosen over less refined ones.
+struct EarlyCastFinder
+  : public LinearExecutionWalker<EarlyCastFinder,
+                                 UnifiedExpressionVisitor<EarlyCastFinder>> {
+  PassOptions options;
+  size_t numLocals;
+
+  // Tracks the current earliest local.get that we can move a cast to without
```
kripken, line 145: Please start with `For each local index,` - otherwise it's not obvious what these vectors are indexed on, I think.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,268 @@ namespace wasm {
...
+
+  // Used to analyze expressions to see if casts can be moved past them.
+  EffectAnalyzer testRefCast;
+  EffectAnalyzer testRefAs;
+
+  // Maps LocalGets to the most refined RefCast to move to it, to be used by the
+  // EarlyCastApplier. If the most refined RefCast is already at the desired
+  // LocalGet, it does not appear here. In the normal case, only one RefCast
+  // needs to be moved to a LocalGet. If a LocalGet is cast to multiple types
+  // which are not subtypes of each other the a trap is inevitable, and we
```
kripken, line 163:
```suggestion
  // which are not subtypes of each other then a trap is inevitable, and we
```

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,268 @@ namespace wasm {
...
+  std::vector<RefAsInfo> currRefAsMove;
+
+  // Used to analyze expressions to see if casts can be moved past them.
+  EffectAnalyzer testRefCast;
+  EffectAnalyzer testRefAs;
+
+  // Maps LocalGets to the most refined RefCast to move to it, to be used by the
+  // EarlyCastApplier. If the most refined RefCast is already at the desired
+  // LocalGet, it does not appear here. In the normal case, only one RefCast
+  // needs to be moved to a LocalGet. If a LocalGet is cast to multiple types
```
kripken, line 162:
```suggestion
  // needs to be moved to a LocalGet; if a LocalGet is cast to multiple types
```
These two sentences are really tied together, and this helps understand them I think.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,268 @@ namespace wasm {
...
+    // TODO: generalize this when we handle more than RefAsNonNull
+    RefCast dummyRefCast(module->allocator);
+    RefAs dummyRefAs(module->allocator);
+    dummyRefAs.op = RefAsNonNull;
+
+    testRefCast.visit(&dummyRefCast);
+    testRefAs.visit(&dummyRefAs);
+  }
+
+  void flushRefCastResult(size_t index, Module& module) {
```
kripken, line 188: Please add a comment about what this does, like
```
We track information as we go, looking for the best cast to move backwards, and when we hit a barrier - a position we can't optimize past - then we flush/finalize the work we've done so far, since nothing better can appear later.
```

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,268 @@ namespace wasm {
...
+    if (currRefCastMove[index].target) {
+      if (currRefCastMove[index].bestCast) {
+        // If the fallthrough is equal to the target, this means the cast is
+        // already at the target local.get and doesn't need to be duplicated
+        // again.
+        auto* fallthrough = Properties::getFallthrough(
+          currRefCastMove[index].bestCast, options, module);
+        if (fallthrough != currRefCastMove[index].target) {
+          refCastToApply[currRefCastMove[index].target] =
+            currRefCastMove[index].bestCast;
```
kripken, line 198: This code can be clearer with capturing the long repeated expressions `currRefCastMove[index].target` and `currRefCastMove[index].bestCast` to locals. Using `&` on the local allows it to be modified later. I mean, this section can start with
```cpp
auto& target = currRefCastMove[index].target;
..
```

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,268 @@ namespace wasm {
...
+
+  static void doNoteNonLinear(EarlyCastFinder* self, Expression**) {
+    for (size_t i = 0; i < self->numLocals; i++) {
+      self->flushRefCastResult(i, *self->getModule());
+      self->flushRefAsResult(i, *self->getModule());
+    }
+  }
+
+  void visitFunction(Function* curr) {
+    for (size_t i = 0; i < numLocals; i++) {
```
kripken, line 231: This can actually just call `doNoteNonLinear` to avoid repeating the code. Or perhaps both can call a helper with a nice name, like `flushAll()`.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,268 @@ namespace wasm {
...
+    visitExpression(curr);
+
+    // TODO: support more than RefAsNonNull
+    if (curr->op != RefAsNonNull) {
+      return;
+    }
+
+    auto* fallthrough = Properties::getFallthrough(curr, options, *getModule());
+    if (auto* get = fallthrough->dynCast<LocalGet>()) {
+      RefAsInfo& bestMove = currRefAsMove[get->index];
```
kripken, line 284:
```suggestion
      auto& bestMove = currRefAsMove[get->index];
```

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,268 @@ namespace wasm {
...
+    // If the best cast is already at the target location, we will still track
+    // it in currRefCastMove to see if we can obtain a better cast. However, we
+    // won't flush it.
+
+    auto* fallthrough = Properties::getFallthrough(curr, options, *getModule());
+    if (auto* get = fallthrough->dynCast<LocalGet>()) {
+      RefCastInfo& bestMove = currRefCastMove[get->index];
+      if (bestMove.target) {
+        if (!bestMove.bestCast) {
+          // If there isn't any other cast to move, the current cast is the best
```
kripken, line 332:
```suggestion
          // If there isn't any other cast to move, the current cast is the best.
```

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,268 @@ namespace wasm {
...
+          bestMove.bestCast = curr;
+        } else if (bestMove.bestCast->type != curr->type &&
+                   Type::isSubType(curr->type, bestMove.bestCast->type)) {
+          // If the current cast is more refined than the best cast to move,
+          // change the best cast to move
```
kripken, line 337:
```suggestion
          // change the best cast to move.
```

src/passes/OptimizeCasts.cpp
zm2he: Done.

src/passes/OptimizeCasts.cpp
zm2he: Done.

src/passes/OptimizeCasts.cpp
zm2he: Done.

src/passes/OptimizeCasts.cpp
zm2he: Done.

src/passes/OptimizeCasts.cpp
zm2he: Done.

src/passes/OptimizeCasts.cpp
zm2he: Done.

src/passes/OptimizeCasts.cpp
zm2he: Done.

src/passes/OptimizeCasts.cpp
zm2he: Done.

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +413,261 @@
...
+    ;; Casts cannot be moved past local sets.
+    (local.set $x
+      (call $get)
+    )
+    (drop
+      (local.get $x)
+    )
+    (drop
+      ;; This will be moved above as the first RefAs.
+      (ref.as_non_null
```
kripken, line 495: This cast does nothing since `$x` is already non-nullable, so as mentioned in a previous comment, we probably shouldn't optimize here. Making the param nullable seems like a good idea.

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +413,261 @@
...
+        (local.get $x)
+      )
+    )
+    (local.set $x
+      (call $get)
+    )
+    (drop
```
kripken, line 502: Is this trio from here testing the moving of both  RefAs and a RefCast? A comment would be good. Also, with this is a separate function, the function name could hint at what is tested.

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +413,261 @@
...
+  ;; CHECK-NEXT:    (local.get $2)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (ref.cast $A
+  ;; CHECK-NEXT:    (local.get $2)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $testMoveCast (param $x (ref struct))
```
kripken, line 472: We generally don't use `test` at the start of test functions, and we use `move-cast` (and not `moveCast`).

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +413,261 @@
...
+      (call $get)
+    )
+    (drop
+      (local.get $x)
+    )
+    (drop
+      (ref.as_non_null
+        (local.get $x)
+      )
+    )
```
kripken, line 509: Testing the reverse order of the RefAs/RefCast would be good.

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +413,261 @@
...
+    (drop
+      (ref.as_non_null
+        (local.get $x)
+      )
+    )
+  )
+
```
kripken, line 670: Another thing to test is the corner case of a tee between the local.get and the cast:
```wat
(ref.cast $A
  (local.tee $x    ;; and also $y
    (local.get $x)
  )
)
```

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +413,261 @@
     )
   )
 
+  ;; CHECK:      (func $testMoveCast (type $ref|struct|_=>_none) (param $x (ref struct))
+  ;; CHECK-NEXT:  (local $1 (ref $B))
+  ;; CHECK-NEXT:  (local $2 (ref $A))
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (ref.cast $B
+  ;; CHECK-NEXT:    (local.tee $1
+  ;; CHECK-NEXT:     (ref.cast $B
```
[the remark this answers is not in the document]
zm2he: After the `EarlyCastApplier` pass, this was
```
(ref.cast $A
  (ref.cast $B
    (local.get $x)
  )
)
```
However, `ReFinalize()` converts the cast to `$A` into a cast to `$B`.

test/lit/passes/optimize-casts.wast
zm2he: The function `$testRefAsAndRefCast$` is already doing much of this, so I will be removing this.

test/lit/passes/optimize-casts.wast
```diff
@@ -190,15 +207,24 @@
         (local.get $x)
       )
     )
+    (global.set $a
```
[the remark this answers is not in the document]
zm2he: Done.

test/lit/passes/optimize-casts.wast
zm2he: Fixed.

test/lit/passes/optimize-casts.wast
zm2he: Added in `$move-ref.as-and-ref.cast`.

test/lit/passes/optimize-casts.wast
zm2he: Fixed.

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +413,261 @@
...
+    )
+    (drop
+      (local.get $x)
+    )
+    (drop
+      (ref.cast $B
+        (local.get $x)
+      )
+    )
+    ;; Casts cannot be moved past local sets.
```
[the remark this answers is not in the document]
zm2he: Separated into `$move-as`.

test/lit/passes/optimize-casts.wast
```diff
@@ -190,15 +207,24 @@
         (local.get $x)
       )
     )
+    (global.set $a
+      (i32.const 30)
```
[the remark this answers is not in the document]
zm2he: Done.

test/lit/passes/optimize-casts.wast
```diff
@@ -54,7 +60,9 @@
 
   ;; CHECK:      (func $ref.as-no (type $ref|$A|_=>_none) (param $x (ref $A))
   ;; CHECK-NEXT:  (drop
-  ;; CHECK-NEXT:   (local.get $x)
+  ;; CHECK-NEXT:   (ref.as_non_null
+  ;; CHECK-NEXT:    (local.get $x)
+  ;; CHECK-NEXT:   )
```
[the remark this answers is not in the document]
zm2he: Done.

test/lit/passes/optimize-casts.wast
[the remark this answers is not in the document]
kripken: I see, thanks! Makes sense.

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,278 @@ namespace wasm {
...
+  void visitLocalGet(LocalGet* curr) {
+    visitExpression(curr);
+
+    if (!currRefCastMove[curr->index].target) {
+      currRefCastMove[curr->index].target = curr;
+    }
+
+    // As we only move RefAsNonNull RefAs casts right now, we should
+    // ignore a LocalGet if the type is already non-nullable, and
+    // adding an extra ref.as_non_null has no effect
```
kripken, line 277:
```suggestion
    // adding an extra ref.as_non_null has no effect.
```

src/passes/OptimizeCasts.cpp
```diff
@@ -102,6 +121,278 @@ namespace wasm {
...
+
+  void visitLocalGet(LocalGet* curr) {
+    visitExpression(curr);
+
+    if (!currRefCastMove[curr->index].target) {
+      currRefCastMove[curr->index].target = curr;
+    }
+
+    // As we only move RefAsNonNull RefAs casts right now, we should
+    // ignore a LocalGet if the type is already non-nullable, and
```
kripken, line 276:
```suggestion
    // ignore a LocalGet if the type is already non-nullable, as
```

test/lit/passes/optimize-casts.wast
zm2he: Added.

test/lit/passes/optimize-casts.wast
```diff
@@ -30,8 +35,8 @@
   ;; CHECK-NEXT:  )
   ;; CHECK-NEXT: )
   (func $ref.as (param $x (ref null $A))
-    ;; After the first ref.as, we can use the cast value in later gets, which is
-    ;; more refined.
+    ;; We duplicate the ref.as to the first local.get, since it is more refined
+    ;; than the local.get alone. We then use the refined index throughout.
     (drop
       (local.get $x)
     )
```
kripken, line 42: The comment on line 51 now applies to line 44 as well: on both of them we don't need the RefAs any more, but we leave it there for other opts.

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +412,282 @@
...
+  ;; CHECK-NEXT:   (local.get $1)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (ref.cast $B
+  ;; CHECK-NEXT:    (local.get $1)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $move-cast (param $x (ref struct))
+    (drop
```
kripken, line 436: An explanation of what to expect would be good - it really helps speed up reading a testcase when you come back to it later and don't remember all the context.

Here, for example, I'd write
```
The most refined cast of $x is to $B, which we can move up to
the top and reuse from there.
```
And to address my question from earlier, which a future reader might also have, can add "(the cast to $A will also turn into a cast to $B due to ReFinalize)".

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +412,282 @@
...
+  ;; CHECK-NEXT:     (local.get $x)
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $move-cast-side-effects (param $x (ref struct)) (param $y (ref struct))
+    (drop
+      (local.get $x)
+    )
+    ;; Cannot move past global set due to trap possibility.
```
kripken, line 529:
```suggestion
    ;; Cannot move past global set due to trap possibility, so the cast to $A will
    ;; move up to here but not further up.
```

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +412,282 @@
...
+    )
+    (drop
+      (local.get $y)
+    )
+    (drop
+      (ref.cast $A
+        (local.get $x)
+      )
+    )
+    ;; Casts to $x cannot be moved past local.set.
```
kripken, line 544:
```suggestion
    ;; Casts to $x cannot be moved past local.set $x, but the cast of $y will be.
```

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +412,282 @@
...
+    (local.set $x
+      (local.get $y)
+    )
+    (drop
+      (ref.cast $B
+        (local.get $y)
+      )
+    )
+    (global.set $a
```
kripken, line 553: What is tested from this line onwards?

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +412,282 @@
...
+      (i32.const 20)
+    )
+    (drop
+      (ref.cast $B
+        (ref.as_non_null
+          (local.get $x)
+        )
+      )
+    )
+  )
```
kripken, line 563: This test is great for showing that we track `RefCast` separately in each local index. Please add such a test for `RefAs` as well.

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +412,282 @@
...
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (ref.cast $A
+  ;; CHECK-NEXT:    (local.get $3)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $move-ref.as-and-ref.cast (param $x (ref null struct)) (param $y (ref null struct))
+    (drop
```
kripken, line 625: Why do we test multiple local indexes here? I'm not sure what this particular test is focused on. Comments might help.

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +412,282 @@
...
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (ref.cast $A
+  ;; CHECK-NEXT:    (local.tee $x
+  ;; CHECK-NEXT:     (local.get $x)
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $move-over-tee (param $x (ref null struct)) (param $y (ref null struct))
+    (drop
```
kripken, line 679: Comment please. And should the name be "*no* move over tee"?

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +412,282 @@
...
+      (local.get $x)
+    )
+    (drop
+      ;; This will be moved above as the first RefAs.
+      (ref.as_non_null
+        (local.get $x)
+      )
+    )
+  )
```
kripken, line 476: These first two testcases are very good for the basics. I think we should add the other possible orderings in them, that is, 5 more for RefCast and one more for RefAs. We don't optimize in all of them, but that's important to test as well.

(Unless we already check all the variations elsewhere?)

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +412,282 @@
...
+    (drop
+      (ref.cast $A
+        (local.tee $x
+          (local.get $x)
+        )
+      )
+    )
+  )
+
   ;; CHECK:      (func $get (type $none_=>_ref|struct|) (result (ref struct))
```
kripken, line 691: Do we have a test of the `noteNonLinear` code path? Not sure I saw one. A way to test that would be to add control flow, e.g., moving part of a pattern we optimize into an `if` arm.

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +412,282 @@
...
+      (local.get $x)
+    )
+    (drop
+      ;; This will be moved above as the first RefAs.
+      (ref.as_non_null
+        (local.get $x)
+      )
+    )
+  )
+
```
kripken, line 477: Please also test where the param is nullable for RefAs, to verify we do *not* optimize there. (In general, knowing where we do not optimize is almost as important as where we do, both to avoid wasted work and to avoid the risk of misoptimizations.)

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +412,282 @@
...
+      (ref.cast $A
+        (local.tee $x
+          (local.get $x)
+        )
+      )
+    )
+  )
+
   ;; CHECK:      (func $get (type $none_=>_ref|struct|) (result (ref struct))
   ;; CHECK-NEXT:  (unreachable)
```
kripken, line 692: It would be good to have a test of `get, cast to A, cast to A`, that is, of identical subsequent casts (in particular, to see that we are deterministic there).

kripken (commented):
I ran this code on a large real-world Java program to see if anything went wrong. No problems, and it found 50 places to optimize out of 50,000 casts. So this isn't the most common situation, but it definitely happens frequently enough to be worth optimizing, which is great.

kripken:
The fuzzer found what looks like a bug:
```wat
(module
 (type $type$1 (struct (field (mut (ref null $type$1))) (field f32) (field (ref null $type$1)) (field (mut externref))))
 (type $type$0 (func (param f64 f64)))
 (func $0 (type $type$0) (param $0 f64) (param $1 f64)
  (local $2 (ref null $type$1))
  (struct.set $type$1 0
   (ref.as_non_null
    (local.get $2)
   )
   (select (result (ref null $type$1))
    (ref.null none)
    (struct.new_default $type$1)
    (ref.eq
     (ref.as_non_null
      (local.get $2)
     )
     (ref.as_i31
      (ref.as_non_null
       (local.get $2)
      )
     )
    )
   )
  )
 )
)
```

```
$ bin/wasm-opt a.wat -all --optimize-casts
[wasm-validator error in function 0] unexpected false: struct.set ref must be a struct, on 
(ref.as_non_null
 (ref.as_i31
  (local.get $2)
 )
)
Fatal: error after opts

**Comment by kripken** [comment]
Another fuzzer testcase, which I think is the same issue:
```wat
(module
 (type $type$3 (struct ))
 (type $type$1 (struct (field i16) (field f64) (field structref) (field (mut i16))))
 (type $type$0 (func (result (ref null $type$3))))
 (type $type$2 (array (mut i32)))
 (func $0 (type $type$0) (result (ref null $type$3))
  (local $0 (ref null $type$1))
  (local $1 (ref $type$1))
  (local.set $1
   (ref.as_non_null
    (local.get $0)
   )
  )
  (drop
   (ref.cast $type$2
    (ref.as_non_null
     (local.get $0)
    )
   )
  )
  (struct.new_default $type$3)
 )
)
```

zm2he:
The main error encountered by the fuzzer seems to be that there is a situation
```
(local.set $a
  (local.get $b)
)
```
Where both $a and $b have compatible types. However, we insert a cast with a type incompatible with neither $a nor $b 
```
(local.set $a
  (ref.cast $C
    (local.get $b)
  )
)
```
there is an error because $a is now getting a value with a wrong type.

To solve this, I check the type of the local.get to ensure that I only move casts to it where are more refined sub-types of its type. This also has the additional advantage of reducing casts which aren't necessary because the local index is already more refined.

I don't need to deal with $a and $b being of incompatible types because that is already detected before this pass.

test/lit/passes/optimize-casts.wast
zm2he: Added.

test/lit/passes/optimize-casts.wast
zm2he: done.

test/lit/passes/optimize-casts.wast
zm2he: Done.

test/lit/passes/optimize-casts.wast
zm2he: Done.

test/lit/passes/optimize-casts.wast
zm2he: Done.

test/lit/passes/optimize-casts.wast
zm2he: Done.

test/lit/passes/optimize-casts.wast
zm2he: Done.

test/lit/passes/optimize-casts.wast
zm2he: Changed to using only one index. I just added the second index because I thought maybe I shouldn't be using the global sets so much when trying to test multiple cases in a function.

test/lit/passes/optimize-casts.wast
zm2he: Done. Made cases for moving over the tee of the same local index as the `local.get`, and moving over the tee of a different local index.

test/lit/passes/optimize-casts.wast
zm2he: I changed the parameters to nullable when testing RefAs.

test/lit/passes/optimize-casts.wast
zm2he: I was testing that when the best cast is already at the target, that I do nothing instead of duplicating the cast again. Moved it to a different function.

test/lit/passes/optimize-casts.wast
zm2he: Made `$move-ref-as-for-separate-index`.

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +420,789 @@
...
+  ;; CHECK-NEXT:    (local.get $1)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (local.get $1)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $move-cast-5 (param $x (ref struct))
+    (drop
+      ;; This is already the most refined cast, so nothing will be moved.
```
kripken, line 600:
```suggestion
      ;; This first location is already the most refined cast, so nothing will be moved up.
      ;; (But we will save the cast to a local and reuse it below.)
```

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +420,789 @@
...
+  ;; CHECK-NEXT:   (ref.cast $D
+  ;; CHECK-NEXT:    (ref.as_non_null
+  ;; CHECK-NEXT:     (local.get $2)
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $avoid-erroneous-cast-move
+    ;; This tests whether we are able to avoid moving a cast earlier if
+    ;; we know doing so will violate typing rules.
```
kripken, line 696:
```suggestion
    ;; This tests shows we avoid moving a cast earlier if
    ;; doing so would violate typing rules.
```

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +420,789 @@
...
+    (local $a (ref null $A))
+    (local $b (ref $A))
+    (local.set $b
+      (ref.as_non_null
```
kripken, line 700: Why is `ref.as_non_null` needed in this test? I think you can show the problem with only `ref.cast`.

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +420,789 @@
...
+        ;; We could move the ref.cast $D here. However, as $a is already known
+        ;; to have type ref null $A, not type $D, it would fail. Moving the cast
```
kripken, line 702:
```suggestion
        ;; to have type ref null $A, not type $D, it would fail, since those types are incompatible. Moving the cast
```

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +420,789 @@
...
+        ;; will also cause the local.set $b to fail, since $b is of type
+        ;; ref null $A, not $D.
+
+        ;; By checking the type of $a, we can avoid moving the ref.cast $D here.
+        ;; If the type of $a is wrong to begin with, the error will be caught
+        ;; before the OptimizeCast pass is run.
```
kripken, line 708:
```suggestion
```
This comment duplicates the above, I think. And no need to mention the last sentence - that's just the normal situation that we can assume modules are valid (in passes, since we validate before).

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +420,789 @@
...
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (local.get $1)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $move-as-2 (param $x (ref null struct))
+    (drop
+      ;; This is already the most refined cast, so nothing is done.
```
kripken, line 764: Something is done, however:
```suggestion
      ;; This is already the most refined cast, so the cast is not copied (but we do save it to a local and use it it below).
```

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +420,789 @@
...
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $move-over-tee (param $x (ref null struct)) (param $y (ref null struct))
+    (local $a (ref null struct))
+    (drop
+      (local.get $x)
+    )
+    (drop
+      ;; We can move this ref.cast because the local.tee sets another local index.
```
kripken, line 1096: What if it's an identical index? Please test that as well.

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +420,789 @@
...
+    (drop
+      ;; This is the first ref.cast $A which appears, so it will be moved
+      ;; up to the first local.get $x.
+      (ref.cast $A
+        (local.get $x)
+      )
+    )
+    (drop
+      ;; Since a ref.cast $A has already been moved, this will not move.
+      (ref.cast $A
```
kripken, line 1138: Both comments here seem confusing, since it's not observable which cast was moved. But the point of the test is to see that the result makes sense and this corner case doesn't cause a problem of any kind.

kripken (commented):
Code lgtm, still some minor test comments though.

test/lit/passes/optimize-casts.wast
zm2he: If it is not visible which one is selected, should I mention which one the implementation selects?

test/lit/passes/optimize-casts.wast
zm2he: Done.

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +420,789 @@
...
+      (local.get $x)
+    )
+    (drop
+      ;; Since we know $x is of type $B, this cast to a less refined type $A
+      ;; will not be moved higher.
+      (ref.cast $A
+        (local.get $x)
+      )
+    )
+  )
```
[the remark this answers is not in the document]
zm2he: Added as `$no-move-ref.as-to-non-nullable-local`.

test/lit/passes/optimize-casts.wast
zm2he: Done.

test/lit/passes/optimize-casts.wast
zm2he: Done.

test/lit/passes/optimize-casts.wast
zm2he: Done.

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +420,789 @@
...
+    ;; Since these casts cannot be moved further, nothing should happen. The ref.cast $B
+    ;; should not be re-inserted to this local.get for instance.
+    (drop
+      (ref.cast $B
+        (ref.as_non_null
+          (local.get $x)
+        )
+      )
+    )
+  )
```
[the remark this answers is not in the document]
zm2he: Done.

test/lit/passes/optimize-casts.wast
kripken: Perfect, thanks! Sorry for missing that.

test/lit/passes/optimize-casts.wast
```diff
@@ -387,6 +420,808 @@
...
+  ;; CHECK-NEXT:   (ref.cast $B
+  ;; CHECK-NEXT:    (ref.as_non_null
+  ;; CHECK-NEXT:     (local.get $x)
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $unoptimizable-nested-casts (param $x (ref null struct))
+    ;; No optimizations should be made here for this nested cast.
+    ;; This test is here to ensure this.
```
kripken, line 1053: Good idea to test this as well.

kripken (approved):
Looks great! Thanks for your patience with my reviews.

PR #5746, opened 2023-05-30
StackIR: Remove nops

### LINKED ISSUE #5745: `wasm-opt` adds `nop` instruction to empty exported function
Running `wasm-opt -Oz` or `wasm-opt -O4` on a module exporting an empty function adds an unnecessary `nop` instruction.

This:
```wat
(module
  (type (func))
  (func (type 0))
  (export "example" (func 0)))
```
becomes this:
```wat
(module
  (type (func))
  (func (type 0)
    nop)
  (export "example" (func 0)))
```

It's not a huge deal, but it seems strange to me that optimization (especially `-Oz`!) would grow the output even by one byte.

kripken:
No `nop` instruction is necessary in wasm, so in StackIR we can simply
remove them all.

(Unless I'm forgetting something..?)

Fixes #5745

### DIFF
--- a/test/lit/passes/roundtrip-gc.wast
+++ b/test/lit/passes/roundtrip-gc.wast
@@ -12,7 +12,7 @@
  ;; CHECK-NEXT:    (local.set $0
  ;; CHECK-NEXT:     (struct.new_default $\7bi32\7d)
  ;; CHECK-NEXT:    )
- ;; CHECK-NEXT:    (nop)
+ ;; CHECK-NEXT:    (call $other)
  ;; CHECK-NEXT:    (local.get $0)
  ;; CHECK-NEXT:   )
  ;; CHECK-NEXT:   (i32.const 1)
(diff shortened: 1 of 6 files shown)


test/lit/passes/roundtrip-gc.wast
```diff
@@ -21,20 +21,24 @@
...
   (call $help
    (struct.new_default ${i32})
-   ;; Stack IR optimizations can remove this block, leaving a nop in an odd
+   ;; Stack IR optimizations can remove this block, leaving a call in an odd
    ;; "stacky" location. On load, we will use a local to work around that. It
    ;; is fine for the local to be non-nullable since the get is later in that
    ;; same block.
    (block $block (result i32)
-    (nop)
+    (call $other)
```
kripken, line 29: This change avoids the test becoming trivial (the new opt removes a nop, after which a lot more optimizations kick in that we don't want here).

aheejin (approved):

PR #5753, opened 2023-06-05
Partial WIP code to fuzz exceptions+multivalue

kripken:
cc @aheejin

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -405,7 +405,10 @@ def pick_initial_contents():
         #   )
         #  )
         # )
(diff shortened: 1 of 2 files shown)


kripken:
This might be useful already, but to really get coverage we'd need to add the TODOs mentioned here.

kripken:
This doesn't take long to find an issue, on seed `14874664545984875933`,
```
wasm-opt: wasm-binary.cpp:1462: void wasm::WasmBinaryWriter::writeType(wasm::Type): Assertion `!type.isTuple() && "Unexpected tuple type"' failed.
```
I haven't investigated it yet.

PR #5756, opened 2023-06-07
[Strings] Fix non-nullable string emitting in the binary format

kripken:
Related to #5737 which did something similar for other types.

Found by the fuzzer.


tlively (commented):
I would have thought that stringref required GC. Is that not the case?

tlively (approved):
I guess you could plausibly implement stringref with ref counting, so LGTM.

kripken:
Yeah, though I doubt any VMs will do so in practice.

PR #5757, opened 2023-06-07
TypeRefining: Fix a bug with chains of StructGets

kripken:
If we have

```wat
(struct.get $A
  (struct.get $B
```

then if both types end up refined we may have a problem. If the inner one is
refined to emit `nullref` then the outer one no longer knows what type it is,
since it depends on the type of the `ref` child for that in our IR.

To avoid that, note all the types first before optimizing.

### DIFF
--- a/src/passes/TypeRefining.cpp
+++ b/src/passes/TypeRefining.cpp
@@ -250,7 +250,29 @@ struct TypeRefining : public Pass {
       }
 
       void visitStructGet(StructGet* curr) {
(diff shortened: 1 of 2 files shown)


tlively (approved):
Another fix would be to just replaceCurrent with an unreachable sequence, which would avoid having to break the usual visitor pattern. I'm ok with either way, though.

tlively (approved):

PR #5758, opened 2023-06-07
Strings: Add initial validation checks

kripken:
This is far from comprehensive, but it checks strings being enabled for
all the instructions. Without this, the fuzzer can get confused because it
checks if code validates and then proceeds under that assumption, so
any missing validation checks can cause problems (specifically, if we have
a `string.const` without strings enabled then we error during writing of
the string, since we don't do the initial pass to find all strings to
deduplicate them).


kripken:
I can also make a helper function to avoid the duplication. But I don't seem to see us do that elsewhere.

tlively (approved):
Alternatively, what if the validator checked that the type of each expression is allowed under the given feature set?

tlively:
I guess that wouldn't work when the operands are unreachable, though, so checking explicitly is more correct.

PR #5762, opened 2023-06-09
Update br_on_cast binary and text format

tlively:
Update br_on_cast binary and text format

The final versions of the br_on_cast and br_on_cast_fail instructions have two
reference type annotations: one for the input type and one for the cast target
type. In the binary format, this is represented as a flags byte followed by two
encoded heap types. Since these instructions have been in flux for a while, do
not attempt to maintain backward compatibility with older versions of the
instructions. Instead, upgrade all of the tests at once to use the new versions
of the instructions. Drop some binary tests of deprecated instruction encodings
that would be more effort to update than they're worth.

Restore binary parsing of previous br_on_cast_{fail} instructions


tlively:
* main
  * **PR #5762** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5762" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5762?utm_source=stack-comment).

tlively:
This is an updated version of #5734, which was previously landed then reverted. See the second commit here for the changes relative to that PR. The difference is that this new PR still parses the old binary format to avoid breaking toolchains that are still emitting it.

kripken (approved):
Please fuzz this.

@bashor is there a simple way we can verify that this change will not cause issues for Kotlin? If there's an example binary you can provide that we can see works, maybe.

tlively:
FWIW I manually tested that a test .wasm file emitted without this change was properly parsed with this change applied. A real, nontrivial module would be better, but at least it's something.

tlively:
Fuzzer seems happy.

bashor:
Here is a Compose app [compose-circels.zip](https://github.com/WebAssembly/binaryen/files/11728581/compose-circels.zip), it contains `br_on_cast_fail` instructions. Now we don't generate `br_on_cast`.

kripken:
Thanks @bashor , I verified that that demo works when processed by Binaryen after this PR landed, and I see no errors in either Binaryen or Chrome (unstable). So everything looks good.

PR #5764, opened 2023-06-12
EffectAnalyzer: Assume we execute the two things whose effects we compare

kripken:
`EffectAnalyzer::canReorder/invalidate` now assume that the things from whom we
generated the effects both execute. If they both execute then we can do more work
in TrapsNeverHappen mode, since we can then reorder this for example:
```wat
(global.set ..)
(i32.load ..)
```
The load may trap, but in TNH mode we assume it won't. So we can reorder those
two. However, if they did not both execute then we could be in this situation:
```wat
(global.set ..)
(br_if ..)
(i32.load)
```
Reordering the load and the set here would be invalid, because we could make the
load execute when it didn't execute before, and it could now start to actually
trap at runtime.

This new assumption seems obvious, since we don't compare the effects of
things unless they are adjacent and with no control flow between them - otherwise,
why compare them? To be sure, I manually reviewed every single use of
`EffectAnalyzer::canReorder/invalidate` in the entire codebase.

I've also been fuzzing this for several days (hundreds of thousands of iterations),
and have not seen any problem.

This was motivated by seeing that #5744 should be able to do more work in TNH
mode, but it wasn't. New tests show the benefits there in OptimizeCasts as well
as in SimplifyLocals.


src/ir/effects.h
```diff
@@ -271,10 +292,18 @@ class EffectAnalyzer {
...
-        (other.trap && writesGlobalState())) {
-      return true;
+    // state is modified. However, in trapsNeverHappen mode we assume traps do
+    // not occur in practice, which lets us ignore this, at least in the case if
+    // the trap will definitely happen, since we can remove traps but not move
+    // them around, so we must check if control flow can transfer. Note that we
+    // use the property of the things being compared both executing, see the
+    // comment on this function.
+    if (!trapsNeverHappen || transfersControlFlow() ||
+        other.transfersControlFlow()) {
```
tlively, line 302: It seems that what we're assuming is not "both analyzed code fragments are executed," but rather "both analyzed code fragments are executed unless the first transfers control." It's probably worth clarifying the opening comment and the comment on `canReorder`.

kripken: Good point, that wasn't very clear. Updated.

src/ir/effects.h
```diff
@@ -217,8 +217,32 @@ class EffectAnalyzer {
...
+  //
+  //   A
+  //   (local.set 0 (i32.const 0))
+  //   B
+  //
+  // But here it is not:
+  //
+  //   A
+  //   (br_if 0 (local.get 0)) ;; this may transfer control flow away
+  //   B
```
aheejin, line 238: In this case, if neither A nor B returns true for `transferscontrolFlow()`, because this `br_if` in between, we cannot reorder A and B, right? I understand we only compare adjacent expressions but then is this example in the comment appropriate?

aheejin (approved):
> This new assumption seems obvious, since we don't compare the effects of things unless they are adjacent and with no control flow between them - otherwise, why compare them? To be sure, I manually reviewed every single use of `EffectAnalyzer::canReorder/invalidate` in the entire codebase.

I wish we can do some `assert` on this but given that we don't have a link to siblings it wouldn't be trivial I guess...

kripken:
> I wish we can do some assert on this but given that we don't have a link to siblings it wouldn't be trivial I guess...

Yeah.

In fact sometimes there aren't specific Expression nodes we compare, as the effects might be accumulated from several of them. But we do always do that for consecutive nodes.

src/ir/effects.h
kripken: Correct, we also cannot reorder them in this case.

Are you saying a better example is something that we *can* reorder? I'm not sure that exists, now that I think of it. That is, for us to be able to reorder two things we do need them to be consecutive, in general. Unless they are trivial things like say a `local.get` and a `nop`. But maybe I misunderstood your question?

tlively (approved):

src/ir/effects.h
aheejin: This function assumes the two expressions are adjacent, and invalidates the expression when either of them contains a control flow, but this example in the comment seems to be where two expressions (A and B) are not adjacent. That was my question.

kripken: I see now, thanks. This is an example of something that the function should not even be called with. I'll rephrase to make the clearer.

PR #5765, opened 2023-06-12
[NFC] Simplify the HeapTypeStore

tlively:
Now that we support only isorecursive typing, which canonicalizes heap types by
the rec group rather than individually, we no longer need a canonicalizing store
for heap types. Simplify the `Store` implementation, which was previously
generic over both `HeapType`s and `Type`s, to instead work only for `Type`s.
Replace `globalHeapTypeStore` with a simple vector to keep canonicalized
`HeapTypeInfo`s alive. Also simplify global canonicalization to not replace heap
type uses, since that is already done separately as part of isorecursive rec
group canonicalization.

This simplification does require adding new information to
`CanonicalizationState`, but further work will simplify that away as well.

### DIFF
--- a/src/wasm/wasm-type.cpp
+++ b/src/wasm/wasm-type.cpp
@@ -119,9 +119,6 @@ struct HeapTypeInfo {
   constexpr bool isStruct() const { return kind == StructKind; }
   constexpr bool isArray() const { return kind == ArrayKind; }
   constexpr bool isData() const { return isStruct() || isArray(); }
-
-  bool operator==(const HeapTypeInfo& other) const;
-  bool operator!=(const HeapTypeInfo& other) const { return !(*this == other); }
 };
 
 // Helper for coinductively checking whether a pair of Types or HeapTypes are in
@@ -588,10 +585,6 @@ HeapTypeInfo::~HeapTypeInfo() {
       return;
   }
   WASM_UNREACHABLE("unexpected kind");
-}
-
-bool HeapTypeInfo::operator==(const HeapTypeInfo& other) const {
-  return this == &other;
 }
 
 struct TypeStore {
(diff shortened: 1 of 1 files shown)


tlively:
* main
  * **PR #5765** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5765" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5765?utm_source=stack-comment).

src/wasm/wasm-type.cpp
```diff
@@ -734,7 +699,7 @@ struct RecGroupStore {
     auto group = asHeapType(info).getRecGroup();
     auto canonical = insert(group);
     if (group == canonical) {
-      globalHeapTypeStore.insert(std::move(info));
+      globalHeapTypeStore.emplace_back(std::move(info));
```
kripken, line 702: Why is this safe to do without taking the mutex?

src/wasm/wasm-type.cpp
```diff
@@ -2501,30 +2462,25 @@ void globallyCanonicalize(CanonicalizationState& state) {
...
   // Keep a lock on the global HeapType store as long as it can reach temporary
   // types to ensure that no other threads observe the temporary types, for
   // example if another thread concurrently constructs a new HeapType with the
   // same shape as one being canonicalized here. This cannot happen with Types
   // because they are hashed in the global store by pointer identity, which has
   // not yet escaped the builder, rather than shape.
-  std::lock_guard<std::recursive_mutex> lock(globalHeapTypeStore.mutex);
-  std::unordered_map<HeapType, HeapType> canonicalHeapTypes;
+  std::lock_guard<std::recursive_mutex> lock(globalHeapTypeStoreMutex);
+  // std::unordered_map<HeapType, HeapType> canonicalHeapTypes;
```
kripken, line 2476: is this still needed?

tlively: No, removed.

kripken (approved):
lgtm with fuzzing

PR #5767, opened 2023-06-13
Rename WasmBinaryBuilder to WasmBinaryReader (NFC)

aheejin:
We have `WasmBinaryBuilder` that read binary into Binaryen IR and `WasmBinaryWriter` that writes Binaryen IR to binary. To me `WasmBinaryBuilder` sounds similar to `WasmBinaryWriter`, which builds binary. How about renaming it to `WasmBinaryReader`?


kripken (approved):
Good idea!

tlively (approved):
LGTM as well

PR #5770, opened 2023-06-14
[EH] Add pass to remove EH instructions

aheejin:
This pass strips all EH stuff, including EH instructions and tags, from the input module and disables the EH feature from the features section.

1. This removes `catch` and `catch_all` blocks from the code. So
```wast
(try
  (do
    (some code)
  )
  (catch
    ...
  )
)
```
becomes just `(some code)`. Note that all `rethrow`s will be removed with `catch`es.

2. This converts 'throw (...)` into `unreachable`. 

3. This removes all tags from the module, which are unused anyway after 1 and 2.

4. This removes exception handling feature from the features section.

You can use the pass with
```console
$ wasm-opt --enable-exception-handling --strip-eh INPUT -o OUTPUT
```
This is not an optimization pass, so it is not run unless you specify the pass.

This is in effect similar to Clang's `-fignore-exceptions`, in which you can throw but it will result in a crash and we compile away all landing pads. This can be used for people who don't (or can't) use `-fignore-exceptions` in their build settings or who want to compile away `catch` blocks later.

Closes emscripten-core/emscripten#19585.


kripken (commented):
What do you think about also replacing `Throw`? Using `getDroppedChildren(,.., unreachable)` that would turn `(throw $e (local.get))` into `(unreachable)`. That could save a bit on code size.

Actually, another benefit to doing that is that then the wasm would contain no more EH instructions (since `Rethrow` also can't exist at that point). That would mean we can remove the exceptions feature from being set and also the wasm can run in older VMs.

juj:
Absolutely awesome!

@sbc100 @kripken I wonder if it would be sensible to enable this pass when `-sDISABLE_EXCEPTION_CATCHING=1` is set at link time? That would match the old behavior iiuc? Or was there a flag to append custom Binaryen passes at Emscripten link time?

kripken:
@juj you can append passes using `-sBINARYEN_EXTRA_PASSES=pass1,pass2`.

(I think running this pass by default would add overhead, so that might be better to avoid. Though I do see your point that it is consistent with the fastcomp behavior, so I'm not totally opposed.)

sbc100:
Should we call this pass `--eh-lowering`  or some thing like that match the existing `--signext-lowering`.  Or maybe just `--remove-eh`?

kripken:
"remove" seems more appropriate to me here since it's destructive. That is, the existing "lowering" passes are like polyfills in that they keep behavior the same.

aheejin:
@kripken 
> What do you think about also replacing `Throw`? Using `getDroppedChildren(,.., unreachable)` that would turn `(throw $e (local.get))` into `(unreachable)`. That could save a bit on code size.
> 
> Actually, another benefit to doing that is that then the wasm would contain no more EH instructions (since `Rethrow` also can't exist at that point). That would mean we can remove the exceptions feature from being set and also the wasm can run in older VMs.

Given that practically `throw` will be in only in a library function for C++ (`__cxa_throw`) I don't think the code size will be meaningfully saved. I think Not having to turn on EH is a plus. But that way we also remove the info for the exception at the same time (thrown value and possibly stack trace in case it's enabled) that can be useful. Would that be fine?

@juj 
> @sbc100 @kripken I wonder if it would be sensible to enable this pass when `-sDISABLE_EXCEPTION_CATCHING=1` is set at link time? That would match the old behavior iiuc? Or was there a flag to append custom Binaryen passes at Emscripten link time?

`DISABLE_EXCEPTION_CATCHING` only applies to Emscripten EH (the old EH), so making it control Wasm EH seems weird. 

@sbc100 @kripken 
> "remove" seems more appropriate to me here since it's destructive. That is, the existing "lowering" passes are like polyfills in that they keep behavior the same.

I chose `--remove-***` for the same reason that it doesn't preserve behaviors.

sbc100:
> I chose `--remove-***` for the same reason that it doesn't preserve behaviors.

Bikeshedding:  How about `--strip-` to match long standing linker flags `--strip-debug` and `--strip-all`?

kripken:
@aheejin 

> But that way we also remove the info for the exception at the same time (thrown value and possibly stack trace in case it's enabled) that can be useful. Would that be fine?

Ah, good point. This would lose a little debuggability, as the value would be lost. (I think the stack trace would be ok, though, as traps can have stack traces.)

The debug benefit of the value seems minor to me. Smaller than the benefit of removing the feature entirely. How about removing the value for now, and if this comes up later we could add it back as an option? (the option could either leave throw, or perhaps set the throen value in a new mutable exported global or such)

aheejin:
> > I chose `--remove-***` for the same reason that it doesn't preserve behaviors.
> 
> Bikeshedding: How about `--strip-` to match long standing linker flags `--strip-debug` and `--strip-all`?

Sounds fine to me.

aheejin:
> @aheejin
> 
> > But that way we also remove the info for the exception at the same time (thrown value and possibly stack trace in case it's enabled) that can be useful. Would that be fine?
> 
> Ah, good point. This would lose a little debuggability, as the value would be lost. (I think the stack trace would be ok, though, as traps can have stack traces.)
> 
> The debug benefit of the value seems minor to me. Smaller than the benefit of removing the feature entirely. How about removing the value for now, and if this comes up later we could add it back as an option? (the option could either leave throw, or perhaps set the throen value in a new mutable exported global or such)

Yeah, then will convert the `throw` to `unreachable` for now. @juj Would that be OK with your usage?

aheejin:
- Renamed the pass (`RemoveCatches` -> `StripEH` and option (`--remove-catches` -> `--strip-eh`)
- Converted `throw` to `unreachable`

src/passes/StripEH.cpp
```diff
@@ -0,0 +1,51 @@
...
+    auto& wasm = *getModule();
+    Builder builder(wasm);
+    replaceCurrent(getDroppedChildrenAndAppend(curr,
+                                               wasm,
+                                               getPassOptions(),
+                                               builder.makeUnreachable(),
+                                               DropMode::IgnoreParentEffects));
+  }
+
+  void visitTry(Try* curr) { replaceCurrent(curr->body); }
```
kripken, line 46: Does the body always have the same type as the Try? If the Try has type`i32` and the body has type `unreachable` (possible with an If, not sure offhand about Try) then this will error as a refinalize is needed.

kripken (commented):
Please also remove the feature, like this pass does for multi-memory:

https://github.com/WebAssembly/binaryen/blob/00211a90878cf0f37fbfb124332c4abc9863032b/src/passes/MultiMemoryLowering.cpp#L380

Unfortunately that can only be done in a non-function-parallel pass. So like that other pass, this will need to be non-parallel, and run a small parallel pass inside it I think.

src/passes/StripEH.cpp
aheejin: Can you give an example of this that fails validation after this without refinalization? I can't really think of one. Now that the `try` is gone, we have its parent operation that expects an `i32` but rather gets `unreachable`, but that's fine, no? For example, 
```
(i32.add
  (try (result i32)
    (throw ...)
    ...
  )
  (i32.const 0)
)
```
Will become
```wast
(i32.add
  (unreachable)
  (i32.const 0)
)
```
And 
```wast
(block (result i32)
  (try (result i32)
    (throw ...)
    ...
  )
)
```
will become
```wast
(block (result i32)
  (unreachable)
)
```
and these seem to be OK.

In which case we need refinalization in general?

kripken: In your first piece of code the type of the `i32.add` itself is `i32`, while in the second it should be `unreachable`. That is, after one of the add's children becomes unreachable, we need to update the parent.

With `BINARYEN_PRINT_FULL=1` those types are printed, and with `BINARYEN_PASS_DEBUG=1` the types are checked, so I think in that mode you'll get an error on that testcase.

In general, replacing something with an identical type is always ok with no additional work. Replacing something with another type - either unreachable, or a subtype - requires type updates in the parent (and perhaps further), which we need refinalize for.

aheejin: I see, fixed it and added a test for it.

src/passes/StripEH.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include <ir/utils.h>
+
+namespace wasm {
+
+namespace {
+
+struct StripEHImpl : public WalkerPass<PostWalker<StripEHImpl>> {
+  bool refinalize = false;
+
+  bool isFunctionParallel() override { return true; }
```
kripken, line 33:
```suggestion
  bool isFunctionParallel() override { return true; }

  bool refinalize = false;
```
(as a convention we always put `isFunctionParallel` first)

src/passes/StripEH.cpp
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  void visitTry(Try* curr) {
+    replaceCurrent(curr->body);
+    refinalize = true;
+  }
+
+  void visitFunction(Function* curr) {
+    if (refinalize) {
+      std::cout << curr->name << std::endl;
```
kripken, line 56:
```suggestion
```

kripken (approved):

PR #5771, opened 2023-06-14
Liveness Analysis Proof of Concept

zm2he:
This introduces a Liveness analysis as an initial proof of concept for the future creation of a monotone flow-sensitive static analysis framework. Tests are included in `test/gtest/cfg.cpp` under the names `CFGTest.LinearLiveness` and `CFGTest.Nonlinearliveness`.

The Liveness Analysis consists of:

- `BlockState` which contains before and after states for a CFG node. It also applies the transfer using each expression in the given CFG node.
- `MonotoneCFGAnalyzer` which builds a graph of `BlockState`s and runs a backward worklist algorithm over the graph of `BlockState`s.
- `BitsetPowersetLattice` which implements a powerset lattice using a bitset.

An example lattice for sign analysis is also included mostly for illustrative purposes.

There are several limitations. The Liveness-analysis specific transfer function and the `BitsetPowersetLattice` are hardcoded into the analysis instead of being templated or interchangeable. Future changes will extract these implementation specific details.

### DIFF
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -1,0 +1,27 @@
+#ifndef wasm_analysis_lattice_h
+#define wasm_analysis_lattice_h
+
+#include <bitset>
+#include "wasm.h"
+
+namespace wasm::analysis {
+
+enum LatticeComparison {
+    NO_RELATION,
+    EQUAL,
+    LESS,
+    GREATER
+};
+
+template<size_t N>
+struct BitsetPowersetLattice {
+    std::bitset<N> value;
+    static bool isTop(const BitsetPowersetLattice<N> & element);
+    static bool isBottom(const BitsetPowersetLattice<N> & element);
+    static LatticeComparison compare(const BitsetPowersetLattice<N> & left, const BitsetPowersetLattice<N> & right);
+    static BitsetPowersetLattice<N> getLeastUpperBound(const BitsetPowersetLattice<N> & left, const BitsetPowersetLattice<N> & right);
+};
+
+}
+
+#endif // wasm_analysis_lattice_h
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -33,10 +33,6 @@ constexpr bool is_lattice =
 template<size_t N> struct BitsetPowersetLattice {
   std::bitset<N> value;
 
-  // Sets the bitvector to be identical to that of a source lattice
-  // element.
-  void copy(BitsetPowersetLattice<N>& src);
-
   static BitsetPowersetLattice<N> getBottom();
   static bool isTop(const BitsetPowersetLattice<N>& element);
   static bool isBottom(const BitsetPowersetLattice<N>& element);
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -16,7 +16,7 @@ constexpr bool has_compare = std::is_invocable_r<LatticeComparison,
                                                  const T&>::value;
 template<typename T>
 constexpr bool has_getLeastUpperBound = std::
-  is_invocable_r<T, decltype(T::getLeastUpperBound), const T&, const T&>::value;
+  is_invocable_r<bool, decltype(&T::getLeastUpperBound), T, const T&>::value;
 template<typename T>
 constexpr bool has_isTop =
   std::is_invocable_r<bool, decltype(T::isTop), const T&>::value;
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -9,41 +9,54 @@ namespace wasm::analysis {
 
 enum LatticeComparison { NO_RELATION, EQUAL, LESS, GREATER };
 
+template<typename T>
+constexpr bool has_compare = std::is_invocable_r<LatticeComparison,
+                                                 decltype(T::compare),
+                                                 const T&,
+                                                 const T&>::value;
+template<typename T>
+constexpr bool has_getLeastUpperBound = std::
+  is_invocable_r<T, decltype(T::getLeastUpperBound), const T&, const T&>::value;
+template<typename T>
+constexpr bool has_isTop =
+  std::is_invocable_r<bool, decltype(T::isTop), const T&>::value;
+template<typename T>
+constexpr bool has_isBottom =
+  std::is_invocable_r<bool, decltype(T::isBottom), const T&>::value;
+
+template<typename T>
+constexpr bool is_lattice =
+  has_compare<T>&& has_getLeastUpperBound<T>&& has_isTop<T>&& has_isBottom<T>;
+
 // Represents a powerset lattice element (i.e. a set) as a bitvector. A true
 // means that an element is present in the set.
 template<size_t N> struct BitsetPowersetLattice {
   std::bitset<N> value;
 
-  // Copies sets the bitvector to be identical to that of a source lattice
+  // Sets the bitvector to be identical to that of a source lattice
   // element.
-  inline void copy(BitsetPowersetLattice<N>& src);
+  void copy(BitsetPowersetLattice<N>& src);
 
-  // Returns a bottom lattice element to use.
-  static inline BitsetPowersetLattice<N> getBottom();
-
-  // Indicates whether a lattice element is the top element.
-  static inline bool isTop(const BitsetPowersetLattice<N>& element);
-
-  // Indicates whether a lattice element is the bottom element.
-  static inline bool isBottom(const BitsetPowersetLattice<N>& element);
+  static BitsetPowersetLattice<N> getBottom();
+  static bool isTop(const BitsetPowersetLattice<N>& element);
+  static bool isBottom(const BitsetPowersetLattice<N>& element);
 
   // Compares two lattice elements and returns a result indicating the
   // left element's relation to the right element.
-  static inline LatticeComparison
-  compare(const BitsetPowersetLattice<N>& left,
-          const BitsetPowersetLattice<N>& right);
+  static LatticeComparison compare(const BitsetPowersetLattice<N>& left,
+                                   const BitsetPowersetLattice<N>& right);
 
   // Returns a lattice element which is the least upper bound of the left and
   // right elements.
-  static inline BitsetPowersetLattice<N>
+  static BitsetPowersetLattice<N>
   getLeastUpperBound(const BitsetPowersetLattice<N>& left,
                      const BitsetPowersetLattice<N>& right);
 
   // Same as above function, but does this in place.
-  inline void getLeastUpperBound(const BitsetPowersetLattice<N>& right);
+  void getLeastUpperBound(const BitsetPowersetLattice<N>& right);
 
   // Prints out the bits in the bitvector for a lattice element.
-  inline void print(std::ostream& os);
+  void print(std::ostream& os);
 };
 
 } // namespace wasm::analysis
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -9,26 +9,41 @@ namespace wasm::analysis {
 
 enum LatticeComparison { NO_RELATION, EQUAL, LESS, GREATER };
 
-// represents a powerset lattice element (i.e. set) as a bitvector. A true
+// Represents a powerset lattice element (i.e. a set) as a bitvector. A true
 // means that an element is present in the set.
 template<size_t N> struct BitsetPowersetLattice {
   std::bitset<N> value;
-  // Returns a bottom element to use.
-  static BitsetPowersetLattice<N> getBottom();
 
-  // indicates whether a lattice is the top element.
-  static bool isTop(const BitsetPowersetLattice<N>& element);
+  // Copies sets the bitvector to be identical to that of a source lattice
+  // element.
+  inline void copy(BitsetPowersetLattice<N>& src);
 
-  // indicates whether a lattice is the bottom element.
-  static bool isBottom(const BitsetPowersetLattice<N>& element);
-  static LatticeComparison compare(const BitsetPowersetLattice<N>& left,
-                                   const BitsetPowersetLattice<N>& right);
-  static BitsetPowersetLattice<N>
+  // Returns a bottom lattice element to use.
+  static inline BitsetPowersetLattice<N> getBottom();
+
+  // Indicates whether a lattice element is the top element.
+  static inline bool isTop(const BitsetPowersetLattice<N>& element);
+
+  // Indicates whether a lattice element is the bottom element.
+  static inline bool isBottom(const BitsetPowersetLattice<N>& element);
+
+  // Compares two lattice elements and returns a result indicating the
+  // left element's relation to the right element.
+  static inline LatticeComparison
+  compare(const BitsetPowersetLattice<N>& left,
+          const BitsetPowersetLattice<N>& right);
+
+  // Returns a lattice element which is the least upper bound of the left and
+  // right elements.
+  static inline BitsetPowersetLattice<N>
   getLeastUpperBound(const BitsetPowersetLattice<N>& left,
                      const BitsetPowersetLattice<N>& right);
 
-  // prints out the bits in the bitvector.
-  void print(std::ostream& os);
+  // Same as above function, but does this in place.
+  inline void getLeastUpperBound(const BitsetPowersetLattice<N>& right);
+
+  // Prints out the bits in the bitvector for a lattice element.
+  inline void print(std::ostream& os);
 };
 
 } // namespace wasm::analysis
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -2,6 +2,7 @@
 #define wasm_analysis_lattice_h
 
 #include <bitset>
+#include <iostream>
 #include "wasm.h"
 
 namespace wasm::analysis {
@@ -16,10 +17,12 @@ enum LatticeComparison {
 template<size_t N>
 struct BitsetPowersetLattice {
     std::bitset<N> value;
+    static BitsetPowersetLattice<N> getBottom();
     static bool isTop(const BitsetPowersetLattice<N> & element);
     static bool isBottom(const BitsetPowersetLattice<N> & element);
     static LatticeComparison compare(const BitsetPowersetLattice<N> & left, const BitsetPowersetLattice<N> & right);
     static BitsetPowersetLattice<N> getLeastUpperBound(const BitsetPowersetLattice<N> & left, const BitsetPowersetLattice<N> & right);
+    void print(std::ostream& os);
 };
 
 }
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -27,4 +27,6 @@ struct BitsetPowersetLattice {
 
 } // namespace wasm::analysis
 
+#include "powerset-lattice-impl.h"
+
 #endif // wasm_analysis_lattice_h
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -1,28 +1,25 @@
 #ifndef wasm_analysis_lattice_h
 #define wasm_analysis_lattice_h
 
+#include "wasm.h"
 #include <bitset>
 #include <iostream>
-#include "wasm.h"
 
 namespace wasm::analysis {
 
-enum LatticeComparison {
-    NO_RELATION,
-    EQUAL,
-    LESS,
-    GREATER
-};
+enum LatticeComparison { NO_RELATION, EQUAL, LESS, GREATER };
 
-template<size_t N>
-struct BitsetPowersetLattice {
-    std::bitset<N> value;
-    static BitsetPowersetLattice<N> getBottom();
-    static bool isTop(const BitsetPowersetLattice<N> & element);
-    static bool isBottom(const BitsetPowersetLattice<N> & element);
-    static LatticeComparison compare(const BitsetPowersetLattice<N> & left, const BitsetPowersetLattice<N> & right);
-    static BitsetPowersetLattice<N> getLeastUpperBound(const BitsetPowersetLattice<N> & left, const BitsetPowersetLattice<N> & right);
-    void print(std::ostream& os);
+template<size_t N> struct BitsetPowersetLattice {
+  std::bitset<N> value;
+  static BitsetPowersetLattice<N> getBottom();
+  static bool isTop(const BitsetPowersetLattice<N>& element);
+  static bool isBottom(const BitsetPowersetLattice<N>& element);
+  static LatticeComparison compare(const BitsetPowersetLattice<N>& left,
+                                   const BitsetPowersetLattice<N>& right);
+  static BitsetPowersetLattice<N>
+  getLeastUpperBound(const BitsetPowersetLattice<N>& left,
+                     const BitsetPowersetLattice<N>& right);
+  void print(std::ostream& os);
 };
 
 } // namespace wasm::analysis
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -9,16 +9,25 @@ namespace wasm::analysis {
 
 enum LatticeComparison { NO_RELATION, EQUAL, LESS, GREATER };
 
+// represents a powerset lattice element (i.e. set) as a bitvector. A true
+// means that an element is present in the set.
 template<size_t N> struct BitsetPowersetLattice {
   std::bitset<N> value;
+  // Returns a bottom element to use.
   static BitsetPowersetLattice<N> getBottom();
+
+  // indicates whether a lattice is the top element.
   static bool isTop(const BitsetPowersetLattice<N>& element);
+
+  // indicates whether a lattice is the bottom element.
   static bool isBottom(const BitsetPowersetLattice<N>& element);
   static LatticeComparison compare(const BitsetPowersetLattice<N>& left,
                                    const BitsetPowersetLattice<N>& right);
   static BitsetPowersetLattice<N>
   getLeastUpperBound(const BitsetPowersetLattice<N>& left,
                      const BitsetPowersetLattice<N>& right);
+
+  // prints out the bits in the bitvector.
   void print(std::ostream& os);
 };
 
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -16,7 +16,7 @@ constexpr bool has_compare = std::is_invocable_r<LatticeComparison,
                                                  const T&>::value;
 template<typename T>
 constexpr bool has_getLeastUpperBound = std::
-  is_invocable_r<bool, decltype(&T::getLeastUpperBound), T, const T&>::value;
+  is_invocable_r<void, decltype(&T::getLeastUpperBound), T, const T&>::value;
 template<typename T>
 constexpr bool has_isTop =
   std::is_invocable_r<bool, decltype(T::isTop), const T&>::value;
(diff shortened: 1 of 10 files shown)


kripken (commented):
Generally looks good!

Will we not have a generic Lattice header from which all lattices inherit? It could use CRTP or such.

src/analysis/sign-lattice.cpp
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        return lub;
+    }
+};
+
+static_assert(std::is_invocable_r<LatticeComparison, decltype(SignLattice::compare), const SignLattice &, const SignLattice &>::value);
+static_assert(std::is_invocable_r<SignLattice, decltype(SignLattice::getLeastUpperBound), const SignLattice &, const SignLattice &>::value);
+static_assert(std::is_invocable_r<bool, decltype(SignLattice::isTop), const SignLattice &>::value);
+static_assert(std::is_invocable_r<bool, decltype(SignLattice::isBottom), const SignLattice &>::value);
```
tlively, line 58: It would be good to package these lattice properties up as a templated constexpr value so we can do something like `static_assert(is_lattice<SignLattice>)`.

src/analysis/sign-lattice.cpp
```diff
@@ -0,0 +1,60 @@
...
+namespace wasm::analysis {
+
+struct SignLattice {
+public:
+    enum Sign {
+        BOTTOM,
+        NEGATIVE,
+        ZERO,
+        POSITIVE,
+        TOP
```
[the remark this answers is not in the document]
tlively: Even if we use "any" and "none" for the elements of this particular lattice, we would still want to use `getTop` and `getBottom()` for the lattice methods because top and bottom do not always correspond to elements that could be referred to as "any" or "none". Reading this analysis code is going to require some familiarity with the relevant formalisms, so I think we should lean into using terms from the literature and have ample documentation for readers and users to bridge the gap.

src/analysis/lattice.h
```diff
@@ -0,0 +1,27 @@
...
+namespace wasm::analysis {
+
+enum LatticeComparison {
+    NO_RELATION,
+    EQUAL,
+    LESS,
+    GREATER
+};
+
+template<size_t N>
```
tlively, line 16: In general the size of the set may not be known at compile time (for example if it is a set of local variables, we don't know how many there will be until we know what function we are analyzing), so actually I think using `std::vector<bool>` would be better than using a `bitset` after all.

(Alternatively, if this lattice is just to prove out the infrastructure rather than for near-term use in a real analysis, it would be fine to keep `N` a compile time constant)

src/analysis/lattice.h
```diff
@@ -0,0 +1,27 @@
...
+struct BitsetPowersetLattice {
+    std::bitset<N> value;
+    static bool isTop(const BitsetPowersetLattice<N> & element);
+    static bool isBottom(const BitsetPowersetLattice<N> & element);
+    static LatticeComparison compare(const BitsetPowersetLattice<N> & left, const BitsetPowersetLattice<N> & right);
+    static BitsetPowersetLattice<N> getLeastUpperBound(const BitsetPowersetLattice<N> & left, const BitsetPowersetLattice<N> & right);
+};
```
tlively, line 23: It would be good for the `BitsetPowersetLattice` to be templated on `T`, the element type of the set, and to have a constructor that takes a `std::vector<T>`. In addition to supplying the set of elements, the vector ensures each element has a unique index for use in the bit vector.

tlively (commented):
Looks good! Left some high-level comments for now. Happy to discuss any of them further.

src/analysis/lattice.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include "wasm.h"
+#include <bitset>
+#include <iostream>
+
+namespace wasm::analysis {
+
+enum LatticeComparison { NO_RELATION, EQUAL, LESS, GREATER };
+
+template<size_t N> struct BitsetPowersetLattice {
```
kripken, line 12: Comments on each class would be helpful, I think. Just reading `BitsetPowerset` doesn't immediately suggest what that is, to me at least.

src/analysis/monotone-analyzer.h
```diff
@@ -0,0 +1,59 @@
...
+#include "cfg.h"
+#include "lattice.h"
+#include "wasm-traversal.h"
+
+namespace wasm::analysis {
+
+template<size_t N> struct MonotoneCFGAnalyzer;
+
+template<size_t N>
+struct BlockState : public UnifiedExpressionVisitor<BlockState<N>> {
```
tlively, line 17: `BlockState` currently bundles multiple concepts for a single block together:
  - The dependent blocks
  - The transfer function
  - The states associated with a block

This is a sort of object-oriented approach where all the state for a single object (i.e block) is put together and represented by a single class, but the downside is that putting all the state together makes it difficult to expose and configure different pieces separately. For example, the dependent blocks and transfer function will be specific to each analysis, so they should be provided by the framework user, but the user shouldn't also need to manage the intermediate lattice states for each block, since that can be the same for every analysis. Having the analysis framework separately manage the dependency graph, transfer function, and intermediate states will make exposing the right interface to the user much simpler.

It's fine if you want to land this with the current design and iterate afterward, though.

src/analysis/monotone-analyzer-impl.h
```diff
@@ -0,0 +1,126 @@
...
+#include <iostream>
+#include <unordered_map>
+
+#include "monotone-analyzer.h"
+
+namespace wasm::analysis {
+template<size_t N>
+BlockState<N>::BlockState(const BasicBlock* underlyingBlock)
+  : index(underlyingBlock->getIndex()), cfgBlock(underlyingBlock),
+    states(underlyingBlock->size() + 1, BitsetPowersetLattice<N>::getBottom()),
```
tlively, line 13: I would expect that we would only store the single entry state for each block, rather than a separate state for each instruction in the block. After the analysis completes, the follow-up code can collect a separate state for each instruction if it needs to, but that shouldn't generally be necessary. For example, consider a live variable analysis, where storing intermediate states for instructions that don't involve locals should never be useful.

src/analysis/monotone-analyzer-impl.h
```diff
@@ -0,0 +1,126 @@
...
+}
+
+template<size_t N> void BlockState<N>::visitLocalGet(LocalGet* curr) {
+  states[currIndex].value[curr->index] = true;
+}
+
+template<size_t N> void BlockState<N>::transfer(std::queue<Index>& worklist) {
+  if (states.size() > 1) {
+    auto cfgIter = cfgBlock->rbegin();
+    currIndex = states.size() - 2;
```
tlively, line 43: Can we do this without using both indices and iterators? It's generally simpler to use just one or the other.

src/analysis/monotone-analyzer-impl.h
```diff
@@ -0,0 +1,126 @@
...
+
+template<size_t N> void BlockState<N>::visitLocalSet(LocalSet* curr) {
+  states[currIndex].value[curr->index] = true;
+}
+
+template<size_t N> void BlockState<N>::visitLocalGet(LocalGet* curr) {
+  states[currIndex].value[curr->index] = true;
+}
+
+template<size_t N> void BlockState<N>::transfer(std::queue<Index>& worklist) {
```
tlively, line 40: It would be nice to have the entire worklist algorithm as a single method of `MonotoneCFGAnalyzer` so the `transfer` function for each block can just take update from an initial state to an end state for a block without worrying about the worklist or any other details of how the analysis is run.

src/analysis/monotone-analyzer-impl.h
```diff
@@ -0,0 +1,126 @@
...
+  for (auto it = cfg->begin(); it != cfg->end(); it++) {
+    result.stateBlocks.emplace_back(&(*it));
+    basicBlockToState[&(*it)] = index++;
+  }
+
+  for (index = 0; index < result.stateBlocks.size(); ++index) {
+    BlockState<N>& currBlock = result.stateBlocks.at(index);
+    BasicBlock::Predecessors preds = currBlock.cfgBlock->preds();
+    BasicBlock::Successors succs = currBlock.cfgBlock->succs();
+    for (auto it = preds.begin(); it != preds.end(); ++it) {
```
tlively, line 90: This can just be:

```suggestion
    for (auto* pred : preds) {
```

src/analysis/powerset-lattice-impl.h
```diff
@@ -0,0 +1,61 @@
...
+    return EQUAL;
+  }
+
+  return NO_RELATION;
+}
+
+template<size_t N>
+BitsetPowersetLattice<N> BitsetPowersetLattice<N>::getLeastUpperBound(
+  const BitsetPowersetLattice<N>& left, const BitsetPowersetLattice<N>& right) {
+  BitsetPowersetLattice<N> result;
```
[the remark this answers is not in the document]
zm2he: Will do.

src/analysis/monotone-analyzer-impl.h
zm2he: Done.

src/analysis/lattice.h
zm2he: Added some comments.

test/gtest/cfg.cpp
```diff
@@ -82,3 +86,86 @@ TEST(CFGTest, Print) {
...
+    )
+  )wasm";
+
+  Module wasm;
+  SExpressionParser parser(moduleText);
+  SExpressionWasmBuilder builder(wasm, *(*parser.root)[0], IRProfile::Normal);
+
+  CFG cfg = CFG::fromFunction(wasm.getFunction("bar"));
+
+  cfg.print(std::cout);
```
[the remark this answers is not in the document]
zm2he: Done.

test/gtest/cfg.cpp
```diff
@@ -82,3 +86,86 @@ TEST(CFGTest, Print) {
...
+
+  const size_t sz = 3;
+  MonotoneCFGAnalyzer<sz> analyzer = MonotoneCFGAnalyzer<sz>::fromCFG(&cfg);
+
+  analyzer.evaluate();
+
+  analyzer.print(std::cout);
+}
+
+TEST(CFGTest, Nonlinearliveness) {
```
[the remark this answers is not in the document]
zm2he: As of right now, the liveness test is hardcoded into the analyzer class, so there aren't any separate implementation files yet.

src/analysis/lattice.h
```diff
@@ -0,0 +1,29 @@
...
+  getLeastUpperBound(const BitsetPowersetLattice<N>& left,
+                     const BitsetPowersetLattice<N>& right);
+  void print(std::ostream& os);
+};
+
+} // namespace wasm::analysis
+
+#include "powerset-lattice-impl.h"
+
+#endif // wasm_analysis_lattice_h
```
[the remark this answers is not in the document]
zm2he: Done.

src/analysis/monotone-analyzer-impl.h
zm2he: I implemented this, but now that we don't store intermediate results, is there any situation where this now ignores the process of applying the LUB between two intermediate states? For example, suppose we have two intermediate states, A, and B. We have just finished calculating A and will now set B = LUB(A, B). If B is a superset of A, then LUB(A, B) = B, but this wouldn't happen if we don't store intermediate states, because we just set B to A.

src/analysis/monotone-analyzer.h
```diff
@@ -0,0 +1,75 @@
...
+
+// A node which contains all the lattice states for a given CFG node.
+template<size_t N>
+struct BlockState : public UnifiedExpressionVisitor<BlockState<N>> {
+  BlockState(const BasicBlock* underlyingBlock);
+
+  void addPredecessor(BlockState* pred);
+  void addSuccessor(BlockState* succ);
+
+  BitsetPowersetLattice<N>& getFirstState();
```
kripken, line 24: Why is this hardcoded on a particular lattice?

src/analysis/monotone-analyzer-impl.h
```diff
@@ -0,0 +1,150 @@
...
+}
+
+template<size_t N> BitsetPowersetLattice<N>& BlockState<N>::getLastState() {
+  return endState;
+}
+
+// In our current limited implementation, we just update a new live variable
+// when it it is used in a get or set.
+template<size_t N> void BlockState<N>::visitLocalSet(LocalSet* curr) {
+  currState.value[curr->index] = true;
```
tlively, line 36: Shouldn't this set the variable to `false`, since its previous value (the "after" state in the backward analysis) is not live?

src/analysis/monotone-analyzer.h
zm2he: I intended this mainly as a proof of concept, so I hardcoded a lot of liveness analysis specific things. I plan to properly template and abstract this in a subsequent PR. However, if you believe that it is more appropriate to template it here, I can also implement that.

src/analysis/powerset-lattice-impl.h
```diff
@@ -0,0 +1,81 @@
...
+#define wasm_analysis_lattice_impl_h
+
+#include "lattice.h"
+
+namespace wasm::analysis {
+
+template<size_t N>
+inline void BitsetPowersetLattice<N>::copy(BitsetPowersetLattice<N>& src) {
+  value.reset();
+  value |= src.value;
```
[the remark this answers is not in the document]
zm2he: I just thought it might do this in-place instead of re-instantiating a bitset. But you're right, I probably should have done the simple thing, as there probably isn't any real difference between them.

src/analysis/monotone-analyzer.h
```diff
@@ -0,0 +1,75 @@
...
+#include "lattice.h"
+#include "wasm-traversal.h"
+
+namespace wasm::analysis {
+
+template<size_t N> struct MonotoneCFGAnalyzer;
+
+// A node which contains all the lattice states for a given CFG node.
+template<size_t N>
+struct BlockState : public UnifiedExpressionVisitor<BlockState<N>> {
```
kripken, line 18: `UnifiedExpressionVisitor` is only needed if you implement `visitExpression` which I don't see here.

src/analysis/powerset-lattice-impl.h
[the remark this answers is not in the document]
kripken: If `value = src.value` works, then do we even need this function? the default `=` operation on `BitsetPowersetLattice<N>` might be fine.

src/analysis/monotone-analyzer.h
zm2he: Changed to visitor.

src/analysis/sign-lattice.cpp
```diff
@@ -0,0 +1,70 @@
...
+  std::is_invocable_r<bool, decltype(T::isTop), const T&>::value;
+template<typename T>
+inline constexpr bool has_isBottom =
+  std::is_invocable_r<bool, decltype(T::isBottom), const T&>::value;
+
+template<typename T>
+inline constexpr bool is_lattice =
+  has_compare<T>&& has_getLeastUpperBound<T>&& has_isTop<T>&& has_isBottom<T>;
+
+static_assert(is_lattice<SignLattice>);
```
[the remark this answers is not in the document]
zm2he: Done.

test/gtest/cfg.cpp
```diff
@@ -82,3 +86,144 @@ TEST(CFGTest, Print) {
...
+        )
+      )
+    )
+  )wasm";
+
+  auto analyzerText = R"analyzer(CFG Analyzer
+State Block: 0
+State at beginning: 11
+State at end: 11
+Intermediate States (reverse order):
```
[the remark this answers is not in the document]
zm2he: Done.

src/analysis/monotone-analyzer-impl.h
zm2he: Fixed.

src/analysis/sign-lattice.cpp
zm2he: Done.

src/analysis/lattice.h
```diff
@@ -0,0 +1,53 @@
...
+// Represents a powerset lattice element (i.e. a set) as a bitvector. A true
+// means that an element is present in the set.
+template<size_t N> struct BitsetPowersetLattice {
+  std::bitset<N> value;
+
+  // Copies sets the bitvector to be identical to that of a source lattice
+  // element.
+  inline void copy(BitsetPowersetLattice<N>& src);
+
+  // Returns a bottom lattice element to use.
```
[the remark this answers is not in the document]
zm2he: Removed it.

src/analysis/powerset-lattice-impl.h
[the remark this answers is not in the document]
zm2he: Changed to the `=` operation.

src/analysis/monotone-analyzer-impl.h
zm2he: In the iterative worklist algorithm, the join doesn't take the LUB of the current state with the inputs, so my original questions was incorrect. 

Each join only takes the LUB of all states it depends upon. We can take the LUB between a basic block's input state and some input multiple times to update it efficiently (only changed values are used to update the LUB) because the transfer function isn't actually applied on the input state (it is only applied to get the next state). 

This means we don't have to worry about the transfer function being applied on some expression A affecting JOIN(A) and trying to maintain that state.

src/analysis/monotone-analyzer-impl.h
```diff
@@ -8,78 +8,83 @@
...
 
-template<size_t N> void BlockState<N>::visitLocalGet(LocalGet* curr) {
+template<size_t N> inline void BlockState<N>::visitLocalGet(LocalGet* curr) {
   currState.value[curr->index] = true;
 }
 
-template<size_t N> void BlockState<N>::transfer() {
+template<size_t N> inline void BlockState<N>::transfer() {
   if (cfgBlock->size() == 0) {
     beginningState.getLeastUpperBound(endState);
```
tlively, line 47: It looks like this is propagating information from the end of an empty block to the beginning of the empty block, which doesn't look correct. The loop below will already run 0 times for an empty block, so I think we can just remove this whole `if`.

src/analysis/monotone-analyzer-impl.h
```diff
@@ -8,78 +8,83 @@
...
-  currState.copy(endState);
+  currState = endState;
 
   while (cfgIter != cfgBlock->rend()) {
     // run transfer function.
     BlockState<N>::visit(*cfgIter);
     ++cfgIter;
   }
-  beginningState.copy(currState);
+  beginningState = currState;
```
tlively, line 60: I'm surprised that we're setting `beginningState` to the result of applying the transfer function. Shouldn't we leave this block's `beginningState` alone and propagate the results to the successor blocks instead?

tlively (commented):
Nice! This is definitely getting there 👍

src/analysis/monotone-analyzer-impl.h
tlively: Ah, I see that the naming here assumes a backward analysis. LGTM for now, although we'll want to fix that later.

test/gtest/cfg.cpp
```diff
@@ -82,3 +86,165 @@ TEST(CFGTest, Print) {
...
+  CFG cfg = CFG::fromFunction(wasm.getFunction("bar"));
+  const size_t sz = 2;
+  MonotoneCFGAnalyzer<sz> analyzer = MonotoneCFGAnalyzer<sz>::fromCFG(&cfg);
+  analyzer.evaluate();
+
+  bool colors = Colors::isEnabled();
+  Colors::setEnabled(false);
+  std::stringstream ss;
+  analyzer.print(ss);
+  Colors::setEnabled(colors);
```
kripken, line 247: Lines `234-236` and `243-247` look repeated in the tests. I assume we'll be adding more in time so the repetition will grow? Perhaps there could be a helper function that does the general outline of things and receives as input the wat text and expected output?

kripken (approved):
Aside from @tlively 's comments overall this looks good to me, and other stuff can be followups.

test/gtest/cfg.cpp
tlively: We should probably use a [test fixture](https://github.com/google/googletest/blob/main/docs/primer.md#test-fixtures-using-the-same-data-configuration-for-multiple-tests-same-data-multiple-tests) for this. I'd be happy deferring this to a standalone follow-up.

src/analysis/lattice.h
```diff
@@ -0,0 +1,62 @@
...
+enum LatticeComparison { NO_RELATION, EQUAL, LESS, GREATER };
+
+template<typename T>
+constexpr bool has_compare = std::is_invocable_r<LatticeComparison,
+                                                 decltype(T::compare),
+                                                 const T&,
+                                                 const T&>::value;
+template<typename T>
+constexpr bool has_getLeastUpperBound = std::
+  is_invocable_r<T, decltype(T::getLeastUpperBound), const T&, const T&>::value;
```
tlively, line 19: for checking member functions, see this working example:

```
#include <type_traits>

struct foo {
  int bar(int, int);
};

static_assert(std::is_invocable_r_v<int, decltype(&foo::bar), foo, int, int>);
```

Note that the method receiver type `foo` is passed to `std::is_invocable` as the first argument, even though that's not really how you call it.

tlively (approved):
LGTM! Let's get this merged 👍

zm2he:
Rebased to resolve merge conflicts from PR #5781.

src/analysis/lattice.h
```diff
@@ -0,0 +1,62 @@
...
+enum LatticeComparison { NO_RELATION, EQUAL, LESS, GREATER };
+
+template<typename T>
+constexpr bool has_compare = std::is_invocable_r<LatticeComparison,
+                                                 decltype(T::compare),
+                                                 const T&,
+                                                 const T&>::value;
+template<typename T>
+constexpr bool has_getLeastUpperBound = std::
+  is_invocable_r<bool, decltype(&T::getLeastUpperBound), T, const T&>::value;
```
tlively, line 19: Since this method returns `void`, I wouldn't expect this assertion to supply a return type.

```suggestion
constexpr bool has_getLeastUpperBound = std::
  is_invocable<decltype(&T::getLeastUpperBound), T, const T&>::value;
```

zm2he: This is a bit of a mix-up. I planned to convert the function to return a bool in later changes because this would let me know whether the lattice element was changed by the LUB, which is convenient when determining which nodes to queue in the worklist. However, I will change it back here.

tlively: Makes sense 👍

zm2he: I've removed the static `getLeastUpperBound` function because the static assert constexpr doesn't seem to like the fact that its name is overloaded with the `getLeastUpperBound` member function and throws compile errors.

src/analysis/lattice.h
zm2he: Assertions are now run in the monotone analyzer now.

test/gtest/cfg.cpp
zm2he: Introduced by #5781. Rebased on it.

src/analysis/monotone-analyzer-impl.h
zm2he: I've removed this special case, since as you said, it can be propagated from the `beginningState` to the `currentState` and then to the `endState` if the loop is not run.

src/analysis/lattice.h
```diff
@@ -0,0 +1,53 @@
...
+          const BitsetPowersetLattice<N>& right);
+
+  // Returns a lattice element which is the least upper bound of the left and
+  // right elements.
+  static inline BitsetPowersetLattice<N>
+  getLeastUpperBound(const BitsetPowersetLattice<N>& left,
+                     const BitsetPowersetLattice<N>& right);
+
+  // Same as above function, but does this in place.
+  inline void getLeastUpperBound(const BitsetPowersetLattice<N>& right);
```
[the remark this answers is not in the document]
zm2he: I've decided to only use the in-place member function now, and have deleted the static function. This over loading is also causes problems with the static assert constexpr.

src/analysis/sign-lattice.cpp
```diff
@@ -0,0 +1,70 @@
...
+    if (left.value == right.value) {
+      return EQUAL;
+    } else if (left.value == BOTTOM || right.value == TOP) {
+      return LESS;
+    } else if (left.value == TOP || right.value == BOTTOM) {
+      return GREATER;
+    } else {
+      return NO_RELATION;
+    }
+  }
```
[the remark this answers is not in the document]
zm2he: Might consider it for this case, but as we mentioned in our discussion, as a general rule, calculating the LUB is going to be very slow for certain lattices, so we want to avoid using it.

src/analysis/lattice.h
```diff
@@ -0,0 +1,53 @@
...
+namespace wasm::analysis {
+
+enum LatticeComparison { NO_RELATION, EQUAL, LESS, GREATER };
+
+// Represents a powerset lattice element (i.e. a set) as a bitvector. A true
+// means that an element is present in the set.
+template<size_t N> struct BitsetPowersetLattice {
+  std::bitset<N> value;
+
+  // Copies sets the bitvector to be identical to that of a source lattice
```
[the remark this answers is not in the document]
zm2he: Done.

src/analysis/lattice.h
zm2he: This is true. The `bitset` powerset lattice is only left here just to prove out the infrastructure. I'm leaving changing it to a dynamically sized powerset lattice in a future PR.

src/analysis/lattice.h
zm2he: I'm leaving this to a follow-up PR.

src/analysis/monotone-analyzer-impl.h
zm2he: I need the iterator to iterate through the CFG block's expressions, while the index is needed to iterate/index the vector of states. The index is a member variable because it lets me access the current state in the visitor without having to change the `Visitor` API.

This is rendered obsolete after we switched to only tracking the beginning and end states for a CFG block instead of tracking state for every expression.

src/analysis/monotone-analyzer-impl.h
zm2he: I'm leaving this for a follow-up PR. However, the step for propagating states between CFG blocks is now in the `MonotoneCFGAnalyzer`.

src/analysis/powerset-lattice-impl.h
```diff
@@ -0,0 +1,61 @@
+#ifndef wasm_analysis_lattice_impl_h
+#define wasm_analysis_lattice_impl_h
+
+#include "lattice.h"
+
+namespace wasm::analysis {
+
+template<size_t N>
+BitsetPowersetLattice<N> BitsetPowersetLattice<N>::getBottom() {
```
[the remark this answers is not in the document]
zm2he: Done.

src/analysis/monotone-analyzer.h
zm2he: This will be left to follow-up PRs where I will try to separate responsibilities between different classes and extract common elements.

src/analysis/lattice.h
```diff
@@ -0,0 +1,53 @@
...
+  // element.
+  inline void copy(BitsetPowersetLattice<N>& src);
```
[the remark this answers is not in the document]
zm2he: Removed the `inline`s in the header files, but not the `impl` header files. The reason mentioned in our discussion was that `inline` is implicitly applied to all member methods inside the scope of a header file's class definition. We need inlines in the `impl` header files because these function definitions are not within the class definition scope.

PR #5772, opened 2023-06-14
[Outlining] StringifyWalker

ashleynh:
This PR is part of the effort to bring Outlining to Binaryen.

StringifyWalker is a new Walker with UnifiedExpressionVisitor.  This walker performs a shallow visit of control-flow (try, if, block, loop) expressions and their simple expression siblings before then visiting the children of each control-flow expression in postorder. As a result, this walker un-nests nested control flow structures, so the expression visit order does not correspond to a normal postorder traversal of the function.


src/passes/Outlining.cpp
```diff
@@ -0,0 +1,211 @@
...
+#include "src/ir/utils.h"
+#include "stringify-walker.h"
+#include "wasm-builder.h"
+#include "wasm-stack.h"
+#include <wasm.h>
+
+namespace wasm {
+
+void StringifyWalker::walkModule(Module* module) {
+  this->pushTask(StringifyWalker::handler, nullptr);
```
tlively, line 32: It would be good to add comments about what the `handler` is for and generally about what this walker is doing.

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,211 @@
...
+    this->emitFunctionBegin(this);
+    this->walk(func->body);
+  });
+}
+
+void StringifyWalker::scanChildren(StringifyWalker* stringify,
+                                   Expression** currp) {
+  Expression* curr = *currp;
+  switch (curr->_id) {
+    case Expression::Id::BlockId: {
```
tlively, line 43: This looks the same as what `PostWalker<...>::scan` would do for a block. Maybe we can just call out to that scan method here for everything except `If`.

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,211 @@
...
+    }
+    case Expression::Id::IfId: {
+      auto* iff = curr->dynCast<If>();
+      stringify->pushTask(StringifyWalker::scan, &iff->ifFalse);
+      // std::cout << "Pushing an task to call StingifyWalker::scan on ifFalse "
+      // << std::endl;
+      stringify->pushTask(StringifyWalker::scan, &iff->ifTrue);
+      // std::cout << "Pushing an task to call StingifyWalker::scan on ifTrue "
+      // << std::endl;
+      stringify->pushTask(StringifyWalker::scan, &iff->condition);
```
tlively, line 64: Since the condition is not actually inside the `if` structure in the binary, it should be pushed during the normal `scan` rather than here in `scanChildren`.

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+void StringifyWalker::walkModule(Module* module) {
+  this->pushTask(StringifyWalker::handler, nullptr);
+  ModuleUtils::iterDefinedFunctions(*module, [&](Function* func) {
+    this->emitFunctionBegin(this);
+    this->walk(func->body);
+  });
+}
+
+void StringifyWalker::scanChildren(StringifyWalker* stringify,
```
tlively, line 39: Maybe call this `deferredScan` or something like that to differentiate it more clearly from normal `scan`, which also scans children.

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,211 @@
...
+    case Expression::Id::LoopId:
+    case Expression::Id::TryId: {
+      self->visitControlFlow(self, currp);
+      // std::cout << "Adding " << name << " to queue" << std::endl;
+      self->queue.push(currp);
+      break;
+    }
+    default: {
+      // std::cout << "Calling PostWalker::scan" << std::endl;
+      PostWalker::scan(self, currp);
```
tlively, line 129: It would be good to add an assertion that `Properties::isControlFlowStructure` here to make sure we update this if we ever add new control flow structures.

src/passes/stringify-walker.h
```diff
@@ -0,0 +1,36 @@
+#ifndef passes_stringify_walker_module_h
+#define passes_stringify_walker_module_h
+
+#include <queue>
+
+namespace wasm {
+
+struct StringifyWalker
```
tlively, line 8: It will be useful to make this a CRTP utility, i.e. to make this a template that takes a `SubType` parameter like other generic walkers. This will let us define multiple concrete subclasses with different implementations of `visitExpression`, `visitControlFlow`, etc. for use in different parts of outlining and in testing.

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,211 @@
...
+  [[maybe_unused]] auto name = getExpressionName(curr);
+  // std::cout << "StringifyWalker::scan() on: " << name << std::endl;
+  // curr->dump();
+
+  switch (curr->_id) {
+    case Expression::Id::BlockId:
+    case Expression::Id::IfId:
+    case Expression::Id::LoopId:
+    case Expression::Id::TryId: {
+      self->visitControlFlow(self, currp);
```
tlively, line 122: I wonder if this could just be a `visitExpression(curr)` call. Then we would leave it up to the client code implementing `visitExpression` to figure out what to do differently with control flow structures, if anything.

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,211 @@
...
+        [[maybe_unused]] auto name = getExpressionName(child);
+        // std::cout << "Pushing a task to call StringifyWalker::scan with block
+        // list item: " << name  << " " << child << std::endl;
+        stringify->pushTask(StringifyWalker::scan, &child);
+      }
+      break;
+    }
+    case Expression::Id::IfId: {
+      auto* iff = curr->dynCast<If>();
+      stringify->pushTask(StringifyWalker::scan, &iff->ifFalse);
```
tlively, line 58: We need to be able to separate the `ifFalse` and `ifTrue` branches with a unique character. In general it looks like we're missing some sort of `visitBlockEnd` method/task that should be called at the end of every control flow block.

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,266 @@
...
+     */
+    self->pushTask(StringifyWalker::handler, nullptr);
+    self->pushTask(StringifyWalker::addUniqueSymbol, &func->body);
+    self->walk(func->body);
+    self->addUniqueSymbol(self, &func->body);
+  });
+}
+
+/*
+ * This handler is responsibe for ensuring the children expressions of control
```
kripken, line 84:
```suggestion
 * This handler is responsible for ensuring the children expressions of control
```

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,266 @@
...
+     *    children
+     * 3. then we call walk, which will visit the function body as a single unit
+     * 4. finally we call addUniqueSymbol directly to ensure the string encoding
+     *    for each function is terminated with a unique symbol, acting as a
+     *    separator between each function in the program string
+     */
+    self->pushTask(StringifyWalker::handler, nullptr);
+    self->pushTask(StringifyWalker::addUniqueSymbol, &func->body);
+    self->walk(func->body);
+    self->addUniqueSymbol(self, &func->body);
```
kripken, line 79: IIUC, this pushes two tasks, so the queue looks like this:
```
[handler, addUniqueSymbol]
```
Then we call walk, which first pushes the body,
```
[handler, addUniqueSymbol, body]
```
and then it pops the body and proceeds to do the walk. When it finishes with the body we are left with
```
[handler, addUniqueSymbol]
```
and then we pop `addUniqueSymbol`, and then pop `handler`.

Is that right?

If so, can't we call `addUniqueSymbol` and `handler` directly after the walk? I feel I'm missing something.

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,266 @@
...
+     * The ordering of the below lines of code are important. On each function
+     * iteration, we:
+     * 1. push a task for calling the handler function, to ensure that each
+     *    function has an opportunity to dequeue from StringifyWalker's
+     *    internally managed queue. This queue exists to provide a way for
+     *    control flow to defer scanning their children.
+     * 2. push a task for adding a unique symbol, so that after the function
+     *    body is visited as a single expression, there is a a separator between
+     *    the symbol for the entire function and the symbol for the function's
+     *    children
```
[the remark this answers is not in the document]
ashleynh: I mean children of the function block here. We evaluate the function block as a unit or single expression, add a unique symbol, and then visit the function's children. Perhaps I can reword.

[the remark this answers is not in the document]
tlively: I think this is confusing because I interpret "function's children" as the children of the root expression, but if the `root` is e.g. an `i32.add`, then its children are visited _before_ the unique symbol this comment is referring to is emitted.

Maybe it would be clearer to  "control flow contents" instead of "control flow children." Then "control flow contents" would mean expressions inside a control flow structure and "children" would refer to all other expressions.

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,266 @@
...
+/*
+ * This walker performs a shallow visit of control-flow (try, if, block, loop)
+ * expressions and their simple expression siblings before then visiting the
+ * children of each control-flow expression in postorder. As a result, this
+ * walker un-nests nested control flow structures, so the expression visit order
+ * does not correspond to a normal postorder traversal of the function.
+ *
+ * For example, the below (contrived) wat:
+ * 1: (block
+ * 2:   (i32.add
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,266 @@
...
+ * 3:     (i32.const 20)
+ * 4:     (i32.const 10))
+ * 5:   (if
+ * 6:     (i32.const 0)
+ * 7:     (then (return (i32.const 1)))
+ * 8:     (else (return (i32.const 0)))
+ * 9:   )
+ *
+ * Would have it's expressions visited in the following order (based on line
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,266 @@
...
+}
+
+void HashStringifyWalker::walkModule(Module* module) {
+  StringifyWalker::walkModule(this, module);
+}
+
+void HashStringifyWalker::addUniqueSymbol(HashStringifyWalker* self,
+                                          Expression** currp) {
+  // string.push_back(monotonic);
+  // monotonic++;
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+  void run(Module* module) override {
+    std::stringstream ss;
+    TestStringifyWalker stringify = TestStringifyWalker(ss);
+    stringify.walkModule(module);
+    std::cout << ss.str();
+  }
+};
+
+Pass* createOutliningPass() { return new Outlining(); }
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/stringify-walker.h
```diff
@@ -0,0 +1,57 @@
...
+  static void addUniqueSymbol(SubType* self, Expression** currp);
+  static void doVisitExpression(SubType* self, Expression** currp);
+  void visitExpression(Expression* curr);
+
+private:
+  static void handler(SubType* self, Expression**);
+  static void deferredScan(SubType* self, Expression** currp);
+};
+
+struct HashStringifyWalker : public StringifyWalker<HashStringifyWalker> {
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/stringify-walker.h
```diff
@@ -0,0 +1,57 @@
...
+
+  std::ostream& os;
+
+  TestStringifyWalker(std::ostream& os);
+
+  void walkModule(Module* module);
+  static void addUniqueSymbol(TestStringifyWalker* self, Expression** currp);
+  void visitExpression(Expression* curr);
+  void print(std::ostream& os);
+};
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/Outlining.cpp
ashleynh: done

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,266 @@
...
+  StringifyWalker<SubType>::deferredScan(self, currp);
+}
+
+template<typename SubType>
+void StringifyWalker<SubType>::deferredScan(SubType* stringify,
+                                            Expression** currp) {
+  Expression* curr = *currp;
+  switch (curr->_id) {
+    case Expression::Id::BlockId: {
+      auto* block = curr->dynCast<Block>();
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,266 @@
...
+      if (block->list.size() > 0) {
+        stringify->pushTask(StringifyWalker::addUniqueSymbol, currp);
+      }
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,266 @@
...
+      auto blockIterator = block->list.end();
+      while (blockIterator != block->list.begin()) {
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,266 @@
...
+        blockIterator--;
+        auto& child = block->list[blockIterator.index];
+        stringify->pushTask(StringifyWalker::scan, &child);
+      }
+      break;
+    }
+    case Expression::Id::IfId: {
+      auto* iff = curr->dynCast<If>();
+      stringify->pushTask(StringifyWalker::addUniqueSymbol, &iff->ifFalse);
+      stringify->pushTask(StringifyWalker::scan, &iff->ifFalse);
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,266 @@
...
+      stringify->pushTask(StringifyWalker::addUniqueSymbol, currp);
+      stringify->pushTask(StringifyWalker::scan, &loop->body);
+      break;
+    }
+    default: {
+      assert(Properties::isControlFlowStructure(curr));
+      auto name = getExpressionName(*currp);
+      std::cout
+        << "deferredScan reached an unimplemented control flow expression: "
+        << name << std::endl;
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,266 @@
...
+    }
+  }
+}
+
+template<typename SubType>
+void StringifyWalker<SubType>::scan(SubType* self, Expression** currp) {
+  Expression* curr = *currp;
+  if (curr->_id == Expression::Id::BlockId || curr->_id == Expression::LoopId ||
+      curr->_id == Expression::TryId || curr->_id == Expression::IfId) {
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,266 @@
...
+    self->pushTask(StringifyWalker::doVisitExpression, currp);
+    self->queue.push(currp);
+  }
+
+  if (curr->_id == Expression::IfId) {
+    auto* iff = curr->dynCast<If>();
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/Outlining.cpp
```diff
@@ -0,0 +1,266 @@
...
+ * 1, 3, 4, 2, 6, 5, 7, 8
+ *
+ * Of note:
+ *   - The add (line 2) binary operator's left and right children are visited
+ * first as they need to be on the stack before the add operation is executed
+ *   - The if-condition (i32.const 0) on line 6 is visited before the if
+ * expression
+ */
+template<typename SubType>
+void StringifyWalker<SubType>::walkModule(SubType* self, Module* module) {
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/Outlining.cpp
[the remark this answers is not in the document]
ashleynh: done

src/passes/Outlining.cpp
ashleynh: completely refactored! take a look, I hope you like it :)

src/wasm-traversal.h
```diff
@@ -290,7 +290,7 @@ struct Walker : public VisitorType {
   };
 
   void pushTask(TaskFunc func, Expression** currp) {
-    assert(*currp);
+    // assert(*currp);
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/stringify-walker.h
```diff
@@ -0,0 +1,68 @@
...
+#include "ir/module-utils.h"
+#include "wasm-traversal.h"
+#include <queue>
+
+namespace wasm {
+
+/*
+ * This walker performs a shallow visit of control-flow (try, if, block, loop)
+ * expressions and their simple expression siblings before then visiting the
+ * children of each control-flow expression in postorder. As a result, this
```
tlively, line 13: Perhaps a clearer version of this first sentence might be "This walker does a normal postorder traversal except that it defers traversing the contents of control flow structures it encounters."

src/passes/stringify-walker.h
```diff
@@ -0,0 +1,68 @@
...
+ *     expression
+ */
+
+template<typename SubType>
+struct StringifyWalker
+  : public PostWalker<SubType, UnifiedExpressionVisitor<SubType>> {
+
+  using Super = PostWalker<SubType, UnifiedExpressionVisitor<SubType>>;
+
+  Module* wasm;
```
tlively, line 46: Walkers already store their modules, so instead of adding this member variable, we can just use `getModule()` where necessary.

test/gtest/stringify.cpp
```diff
@@ -0,0 +1,96 @@
...
+)stringify";
+
+  struct TestStringifyWalker : public StringifyWalker<TestStringifyWalker> {
+    std::ostream& os;
+
+    TestStringifyWalker(std::ostream& os) : os(os){};
+
+    void walkModule(Module* module) { StringifyWalker::walkModule(module); }
+
+    void addUniqueSymbol() { this->os << "adding unique symbol\n"; }
```
tlively, line 75:
```suggestion
    void addUniqueSymbol() { os << "adding unique symbol\n"; }
```

test/gtest/stringify.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+    void visitExpression(Expression* curr) {
+      if (Properties::isControlFlowStructure(curr)) {
+        this->os << "in visitExpression with CF "
+                 << ShallowExpression{curr, this->wasm} << std::endl;
+      } else {
+        this->os << "in visitExpression for "
+                 << ShallowExpression{curr, this->wasm} << std::endl;
```
tlively, line 83: nit: it may be useful to reduce verbosity and just print the ShallowExpression without "in visitExpression with CF" or "in visitExpression for".

tlively (commented):
A bunch more comments, but I promise they're all super minor. After this we will be good to go 🎉

tlively:
It would also be good to mention in the commit description that this will be useful for outlining and perhaps prepend an `[outlining]` tag to the commit title.

test/gtest/stringify.cpp
```diff
@@ -0,0 +1,96 @@
...
+adding unique symbol
+)stringify";
+
+  struct TestStringifyWalker : public StringifyWalker<TestStringifyWalker> {
+    std::ostream& os;
+
+    TestStringifyWalker(std::ostream& os) : os(os){};
+
+    void walkModule(Module* module) { StringifyWalker::walkModule(module); }
+
```
tlively, line 74:
```suggestion
```

test/gtest/stringify.cpp
```diff
@@ -0,0 +1,96 @@
...
+    void addUniqueSymbol() { this->os << "adding unique symbol\n"; }
+
+    void visitExpression(Expression* curr) {
+      if (Properties::isControlFlowStructure(curr)) {
+        this->os << "in visitExpression with CF "
+                 << ShallowExpression{curr, this->wasm} << std::endl;
```
tlively, line 80: I'm pretty sure this will solve the memory errors you're seeing on CI. `this->wasm` is never initialized, so it is reading junk.
```suggestion
                 << ShallowExpression{curr, getModule()} << std::endl;
```
Same below.

ashleynh: it did! done

src/passes/stringify-walker.h
ashleynh: done

src/passes/stringify-walker-impl.h
```diff
@@ -0,0 +1,143 @@
...
+     * 2. push a task for adding a unique symbol, so that after the function
+     *    body is visited as a single expression, there is a a separator between
+     *    the string for the top-level function body and subsequent strings for
+     *    each control flow structure in the function.
+     * 3. then we call walk, which will visit the function body as a single unit
+     * 4. finally we call addUniqueSymbol directly to ensure the string encoding
+     *    for each function is terminated with a unique symbol, acting as a
+     *    separator between each function in the program string
+     */
+    this->walkFunction(func);
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/stringify-walker-impl.h
```diff
@@ -0,0 +1,143 @@
...
+  });
+}
+
+template<typename SubType>
+inline void StringifyWalker<SubType>::doWalkFunction(Function* func) {
+  // call our walk and emit the unique symbol for the function body
+  this->walk(func->body);
+  this->addUniqueSymbol();
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/stringify-walker-impl.h
```diff
@@ -0,0 +1,143 @@
...
+}
+
+template<typename SubType>
+inline void StringifyWalker<SubType>::walk(Expression* curr) {
+  Super::walk(curr);
+  do {
+    this->addUniqueSymbol();
+    this->dequeueControlFlow();
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/stringify-walker-impl.h
```diff
@@ -0,0 +1,143 @@
...
+  } while (controlFlowQueue.size() > 0);
```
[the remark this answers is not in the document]
ashleynh: done

test/gtest/stringify.cpp
ashleynh: done

src/passes/stringify-walker-impl.h
```diff
@@ -0,0 +1,143 @@
...
+
+template<typename SubType>
+void StringifyWalker<SubType>::deferredScan(Expression** currp) {
+  auto self = static_cast<SubType*>(this);
+  Expression* curr = *currp;
+  switch (curr->_id) {
+    case Expression::Id::BlockId: {
+      auto* block = curr->cast<Block>();
+      // TODO: The below code could be simplified if ArenaVector supported
+      // reverse iterators
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/stringify-walker-impl.h
```diff
@@ -0,0 +1,143 @@
...
+      WASM_UNREACHABLE("unexpected expression");
+    }
+  }
+}
+
+template<typename SubType>
+void StringifyWalker<SubType>::doVisitExpression(SubType* self,
+                                                 Expression** currp) {
+  Expression* curr = *currp;
+  self->visitExpression(curr);
```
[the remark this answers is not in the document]
ashleynh: done

test/gtest/stringify.cpp
ashleynh: done

test/gtest/stringify.cpp
ashleynh: done

src/passes/stringify-walker-impl.h
```diff
@@ -0,0 +1,143 @@
...
+ */
+template<typename SubType> void StringifyWalker<SubType>::dequeueControlFlow() {
+  auto& queue = this->controlFlowQueue;
+  if (queue.empty()) {
+    return;
+  }
+
+  Expression** currp = queue.front();
+  queue.pop();
+  this->deferredScan(currp);
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/stringify-walker-impl.h
```diff
@@ -0,0 +1,143 @@
...
+      }
+      break;
+    }
+    case Expression::Id::TryId: {
+      auto* tryy = curr->cast<Try>();
+      Super::walk(tryy->body);
+      this->addUniqueSymbol();
+      for (auto& child : tryy->catchBodies) {
+        Super::walk(child);
+      }
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/stringify-walker-impl.h
```diff
@@ -0,0 +1,120 @@
...
+}
+
+template<typename SubType>
+inline void StringifyWalker<SubType>::scan(SubType* self, Expression** currp) {
+  Expression* curr = *currp;
+  if (Properties::isControlFlowStructure(curr)) {
+    self->controlFlowQueue.push(currp);
+    self->pushTask(doVisitExpression, currp);
+    if (auto* iff = curr->dynCast<If>()) {
+      Super::scan(self, &iff->condition);
```
kripken, line 37: This logic is somewhat similar to the special handling of If here:

https://github.com/WebAssembly/binaryen/blob/84316e8fc3448932c4d62d1e749047aaacf02ef2/src/ir/iteration.h#L143-L155

I wonder if we can write this code using ValueIterator..? Not sure.

src/passes/stringify-walker-impl.h
```diff
@@ -0,0 +1,120 @@
...
+    }
+  } else {
+    Super::scan(self, currp);
+    return;
```
kripken, line 41:
```suggestion
```

src/passes/stringify-walker-impl.h
```diff
@@ -0,0 +1,120 @@
...
+        this->addUniqueSymbol();
+        Super::walk(child);
+      }
+      break;
+    }
+    case Expression::Id::LoopId: {
+      auto* loop = curr->cast<Loop>();
+      Super::walk(loop->body);
+      break;
+    }
```
kripken, line 96: Maybe we can add a ControlChildIterator to parallel ValueChildIterator, that iterates on control flow children, and use that here? It would avoid us needing to update this code each time we add or modify a control flow structure (i.e. it would keep that logic in `iterator.h`.

I mean, in the code here, we could have:
```cpp
for (auto* child : ControlChildIterator(curr)) {
  this->addUniqueSymbol();
  Super::walk(child);
}
```
However that won't work for block, I guess... so we need some special casing here anyhow. Maybe it's not worth it then.

Another option might be `BlockChildIterator` or such, as what we look for here is a child that is a new block scope..?

test/gtest/stringify.cpp
```diff
@@ -0,0 +1,127 @@
...
+          )
+        )
+      )
+    )
+   )
+  )wasm";
+
+  auto stringifyText = R"stringify(in visitExpression for block
+adding unique symbol
+in visitExpression for block $block_a
```
kripken, line 62: Wait, shouldn't we visit the contents of `block_a` before continuing to b? Maybe I'm not understanding the example from the headers. Looking at that example again, it has nested control flow but not sibling control flow - maybe expanding the example to include that could avoid confusion.

src/passes/stringify-walker.h
tlively: I prefer my wording because it better matches the implementation, which uses a queue to defer visiting the children.

src/passes/stringify-walker-impl.h
ashleynh: done

src/passes/stringify-walker-impl.h
```diff
@@ -0,0 +1,143 @@
...
+#define wasm_passes_stringify_walker_impl_h
+
+namespace wasm {
+
+template<typename SubType>
+inline void StringifyWalker<SubType>::doWalkModule(Module* module) {
+  ModuleUtils::iterDefinedFunctions(*module, [&](Function* func) {
+    /*
+     * The ordering of the below lines of code are important. On each function
+     * iteration, we:
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/stringify-walker.h
ashleynh: done

src/passes/stringify-walker-impl.h
```diff
@@ -0,0 +1,143 @@
...
+    Super::scan(self, currp);
+    return;
+  }
+}
+
+/*
+ * This dequeueControlFlow is responsible for ensuring the children expressions
+ * of control flow expressions are visited after the control flow expression has
+ * already been visited. In order to perform this responsibility, the
+ * dequeueControlFlow function needs to always be the very last task in the
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/stringify-walker-impl.h
```diff
@@ -0,0 +1,143 @@
...
+     *    for each function is terminated with a unique symbol, acting as a
+     *    separator between each function in the program string
+     */
+    this->walkFunction(func);
+  });
+}
+
+template<typename SubType>
+inline void StringifyWalker<SubType>::doWalkFunction(Function* func) {
+  // call our walk and emit the unique symbol for the function body
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/stringify-walker-impl.h
```diff
@@ -0,0 +1,143 @@
...
+  } while (controlFlowQueue.size() > 0);
+}
+
+template<typename SubType>
+inline void StringifyWalker<SubType>::scan(SubType* self, Expression** currp) {
+  Expression* curr = *currp;
+  if (Properties::isControlFlowStructure(curr)) {
+    self->controlFlowQueue.push(currp);
+    self->pushTask(doVisitExpression, currp);
+    if (auto* iff = curr->dynCast<If>()) {
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/stringify-walker.h
tlively: I do still think it would be useful to say that this effectively un-nests the control flow, since I think that is a good higher-level intuitive explanation.

test/gtest/stringify.cpp
ashleynh: I updated the comments and added a sibling control flow to the contrived wat example. Hopefully this makes the explanation a bit more clear. Please let me know if not!

This is the correct traversal. 
We want to visit the function body,
then the children of the function body as a single unit: block_a, block_b, block_c, block_d
and then visit the contents of each block. 

If we run into further control flow, we defer visiting the contents of the control flow until after all siblings are visited.

src/passes/stringify-walker-impl.h
ashleynh: I think this is worth adding but would incur quite a bit of boilerplate. I've added a TODO and would like to get to it in a follow-on PR.

src/passes/stringify-walker-impl.h
```diff
@@ -0,0 +1,143 @@
...
+template<typename SubType>
+inline void StringifyWalker<SubType>::doWalkFunction(Function* func) {
+  // call our walk and emit the unique symbol for the function body
+  this->walk(func->body);
+  this->addUniqueSymbol();
+}
+
+template<typename SubType>
+inline void StringifyWalker<SubType>::walk(Expression* curr) {
+  Super::walk(curr);
```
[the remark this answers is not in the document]
tlively: Ok I think this is well covered by the comment with the example wast now.

src/passes/stringify-walker.h
ashleynh: done

src/passes/stringify-walker-impl.h
tlively: Looks like this is still here.

src/passes/stringify-walker-impl.h
kripken: Sounds good. It's not urgent, just an idea (which might not work well, I'm not sure).

test/gtest/stringify.cpp
kripken: I think I see now, thanks! So this doesn't keep a control flow structure close to its children. I was assuming that in my head for some reason but I guess it's not the goal here.

kripken (approved):

src/passes/stringify-walker-impl.h
ashleynh: done! thanks for the tip!

tlively (approved):
🎉

PR #5775, opened 2023-06-16
[NFC] Simplify `Tuple` by making it an alias of `TypeList`

tlively:
Rather than wrap a `TypeList`, make `Tuple` an alias of `TypeList`. This means
removing `Tuple::toString`, but that had no callers and was of limited use for
debugging anyway. In return, the use of tuples becomes much less verbose.

In the future, it may make sense to remove one of `Tuple` and `TypeList`.

### DIFF
--- a/src/wasm-type.h
+++ b/src/wasm-type.h
@@ -51,11 +51,13 @@ void destroyAllTypesForTestingPurposesOnly();
 class Type;
 class HeapType;
 class RecGroup;
-struct Tuple;
 struct Signature;
 struct Field;
 struct Struct;
 struct Array;
+
+using TypeList = std::vector<Type>;
+using Tuple = TypeList;
 
 enum Nullability { NonNullable, Nullable };
 enum Mutability { Immutable, Mutable };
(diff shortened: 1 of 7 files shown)


tlively:
* main
  * **PR #5775** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5775" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5775?utm_source=stack-comment).

src/wasm-type.h
```diff
@@ -453,31 +455,6 @@ class RecGroup {
...
-
-  bool operator==(const Tuple& other) const { return types == other.types; }
-  bool operator!=(const Tuple& other) const { return !(*this == other); }
-  std::string toString() const;
-
-private:
-  void validate() {
-#ifndef NDEBUG
-    for (auto type : types) {
-      assert(type.isSingle());
```
kripken, line 475: Do we have this logic already in the validator? If not it might be good to add.

kripken (approved):
lgtm % question

src/wasm-type.h
tlively: Oh right, I definitely need to replace this with something...

PR #5778, opened 2023-06-20
CtorEval Fuzzer: Generalize export regex

kripken:
Just look for export names as "" with some non-" in the middle.

That should handle everything but escaping of " in an export name (which I don't know
if we can handle in a regex?).

Missing from the old regex: spaces, parens, and probably more. Spaces and parens
are used in the test suite, which is how this was noticed by the fuzzer.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -1263,7 +1263,7 @@ class CtorEval(TestCaseHandler):
 
         # get the list of exports, so we can tell ctor-eval what to eval.
         wat = run([in_bin('wasm-dis'), wasm] + FEATURE_OPTS)
-        p = re.compile(r'^ [(]export "([^"]+)" [(]func')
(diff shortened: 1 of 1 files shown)


tlively (approved):
If you require the close parent to be preceded by nothing or anything except for a `\`, then you should be able to avoid accidentally matching it on an escaped quote.

kripken:
Do you know how to do that in a regex?

kripken:
Doing some tinkering, it seems like `r'"((?:[^\\"]|\\.)*?)"'` might work..

tlively:
I was thinking of something along the lines of `".*[^\\]?"`

kripken:
Thanks, that's even simpler. I keep forgetting that regex `.*` is not "literal" (it still complies with the constraints afterward, rather than consuming everything til the end).

tlively (approved):

PR #5780, opened 2023-06-23
PostEmscripten: Preserve __em_js__ exports (and data) in side modules

sbc100:


tlively (approved):

sbc100:
Added some testing

test/lit/passes/post-emscripten-side-module.wast
```diff
@@ -0,0 +1,49 @@
...
+ (data $data1 (i32.const 1000) "hello world")
+ ;; CHECK:      (data $data2 (i32.const 2000) "hello DELETE ME world")
+ (data $data2 (i32.const 2000) "hello DELETE ME world")
+ ;; CHECK:      (data $data3 (i32.const 3000) "")
+ (data $data3 (i32.const 3000) "some deps")
+ ;; CHECK:      (export "__start_em_asm" (global $em_asm_start))
+ (export "__start_em_asm" (global $em_asm_start))
+ ;; CHECK:      (export "__stop_em_asm" (global $em_asm_stop))
+ (export "__stop_em_asm" (global $em_asm_stop))
+ (export "__start_em_js" (global $em_js_start))
```
kripken, line 39: Why is this one not preserved in the output?

sbc100: For EM_ASM we use the global start/stop symbols and split the resulting string on NULL.

For EM_JS each string is exported via its own global export e.g. `__em_js__foo` so we don't need the start/stop symbols.

PR #5784, opened 2023-06-26
Dynamic Sized Bitvector Powerset Lattice for Static Analysis

zm2he:
This change introduces a finite-height powerset lattice `FinitePowersetLattice` where each element's height is determined when constructed in runtime. This is implemented using a vector of `bool`'s. The previous finite-height powerset lattice implementation had used a `bitset`, which forced the size of each lattice element to be determined statically at compile time, which is a significant drawback, as most powerset lattice sizes are determined by the input program.

This change also modifies `BlockState` and `MonotoneCFGAnalyzer` to be templated on a generic lattice type instead of using a hardcoded lattice. It also fixes an iterator bug in `MonotoneCFGAnalyzer::fromCFG` which assigned a temporary iterator object as predecessor and successor pointers to `BlockState`'s instead of pointers to actual `BlockState` objects.

### DIFF
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -1,9 +1,10 @@
 #ifndef wasm_analysis_lattice_h
 #define wasm_analysis_lattice_h
 
-#include "wasm.h"
-#include <bitset>
 #include <iostream>
+#include <vector>
+
+#include "wasm.h"
 
 namespace wasm::analysis {
 
@@ -30,21 +31,34 @@ constexpr bool is_lattice =
 
 // Represents a powerset lattice element (i.e. a set) as a bitvector. A true
 // means that an element is present in the set.
-template<size_t N> struct BitsetPowersetLattice {
-  std::bitset<N> value;
+class FinitePowersetLattice {
+  std::vector<bool> bitvector;
+  size_t trues;
 
-  static BitsetPowersetLattice<N> getBottom();
-  static bool isTop(const BitsetPowersetLattice<N>& element);
-  static bool isBottom(const BitsetPowersetLattice<N>& element);
+  FinitePowersetLattice(size_t size);
+
+public:
+  void set(size_t index, bool value);
+  bool get(size_t index);
+  size_t size() { return bitvector.size(); }
+  size_t countElements() { return trues; }
+
+  using iterator = std::vector<bool>::const_iterator;
+  iterator begin() { return bitvector.cbegin(); }
+  iterator end() { return bitvector.cend(); }
+
+  static FinitePowersetLattice getBottom(size_t size);
+  bool isTop(const FinitePowersetLattice& element);
+  bool isBottom(const FinitePowersetLattice& element);
 
   // Compares two lattice elements and returns a result indicating the
   // left element's relation to the right element.
-  static LatticeComparison compare(const BitsetPowersetLattice<N>& left,
-                                   const BitsetPowersetLattice<N>& right);
+  static LatticeComparison compare(const FinitePowersetLattice& left,
+                                   const FinitePowersetLattice& right);
 
   // Calculates the LUB of this current (left) lattice element with some right
   // element. It then updates this current lattice element to the LUB in place.
-  void getLeastUpperBound(const BitsetPowersetLattice<N>& right);
+  bool getLeastUpperBound(const FinitePowersetLattice& right);
 
   // Prints out the bits in the bitvector for a lattice element.
   void print(std::ostream& os);
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -20,10 +20,10 @@ constexpr bool has_getLeastUpperBound = std::
   is_invocable_r<void, decltype(&T::getLeastUpperBound), T, const T&>::value;
 template<typename T>
 constexpr bool has_isTop =
-  std::is_invocable_r<bool, decltype(T::isTop), const T&>::value;
+  std::is_invocable_r<bool, decltype(&T::isTop), T>::value;
 template<typename T>
 constexpr bool has_isBottom =
-  std::is_invocable_r<bool, decltype(T::isBottom), const T&>::value;
+  std::is_invocable_r<bool, decltype(&T::isBottom), T>::value;
 
 template<typename T>
 constexpr bool is_lattice =
@@ -33,8 +33,13 @@ constexpr bool is_lattice =
 // means that an element is present in the set.
 class FinitePowersetLattice {
   std::vector<bool> bitvector;
+
+  // Counts the number of elements in the set (i. e. number of trues in the
+  // bitvector).
   size_t trues;
 
+  // Size of the top lattice element must be specified, as it determines the set
+  // of the bitvector.
   FinitePowersetLattice(size_t size);
 
 public:
@@ -47,9 +52,11 @@ public:
   iterator begin() { return bitvector.cbegin(); }
   iterator end() { return bitvector.cend(); }
 
+  // Returns an instance of the bottom lattice element. The size of the the top
+  // lattice element must be specified.
   static FinitePowersetLattice getBottom(size_t size);
-  bool isTop(const FinitePowersetLattice& element);
-  bool isBottom(const FinitePowersetLattice& element);
+  bool isTop();
+  bool isBottom();
 
   // Compares two lattice elements and returns a result indicating the
   // left element's relation to the right element.
@@ -58,6 +65,7 @@ public:
 
   // Calculates the LUB of this current (left) lattice element with some right
   // element. It then updates this current lattice element to the LUB in place.
+  // Returns true if the current lattice element was changed, false otherwise.
   bool getLeastUpperBound(const FinitePowersetLattice& right);
 
   // Prints out the bits in the bitvector for a lattice element.
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -82,10 +82,10 @@ public:
     bool isTop() { return count() == bitvector.size(); }
     bool isBottom() { return count() == 0; }
 
-    // Calculates the LUB of this element with some right element and sets
+    // Calculates the LUB of this element with some other element and sets
     // this element to the LUB in place. Returns true if this element before
     // this method call was different than the LUB.
-    bool makeLeastUpperBound(const Element& right);
+    bool makeLeastUpperBound(const Element& other);
 
     // Prints out the bits in the bitvector for a lattice element.
     void print(std::ostream& os);
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -32,44 +32,49 @@ constexpr bool is_lattice =
 // Represents a powerset lattice element (i.e. a set) as a bitvector. A true
 // means that an element is present in the set.
 class FinitePowersetLattice {
-  std::vector<bool> bitvector;
-
-  // Counts the number of elements in the set (i. e. number of trues in the
-  // bitvector).
-  size_t trues;
-
-  // Size of the top lattice element must be specified, as it determines the set
-  // of the bitvector.
-  FinitePowersetLattice(size_t size);
+  // The size of the set that the powerset lattice was created from. This is
+  // equivalent to the size of the Top lattice element.
+  size_t setSize;
 
 public:
-  void set(size_t index, bool value);
-  bool get(size_t index);
-  size_t size() { return bitvector.size(); }
-  size_t countElements() { return trues; }
+  FinitePowersetLattice(size_t setSize) : setSize(setSize) {}
 
-  using iterator = std::vector<bool>::const_iterator;
-  iterator begin() { return bitvector.cbegin(); }
-  iterator end() { return bitvector.cend(); }
+  class Element {
+  private:
+    std::vector<bool> bitvector;
+    size_t elementSize;
 
-  // Returns an instance of the bottom lattice element. The size of the the top
-  // lattice element must be specified.
-  static FinitePowersetLattice getBottom(size_t size);
-  bool isTop();
-  bool isBottom();
+    Element(size_t latticeSetSize)
+      : bitvector(latticeSetSize), elementSize(0) {}
+
+  public:
+    using iterator = std::vector<bool>::const_iterator;
+    iterator begin() { return bitvector.cbegin(); }
+    iterator end() { return bitvector.cend(); }
+    bool get(size_t index) { return bitvector[index]; }
+    void set(size_t index, bool value);
+    size_t size() { return elementSize; }
+    bool isTop() { return elementSize == bitvector.size(); }
+    bool isBottom() { return elementSize == 0; }
+
+    // Calculates the LUB of this element with some right element and sets
+    // this element to the LUB in place. Returns true if this element before
+    // this method call was different than the LUB.
+    bool getLeastUpperBound(const Element& right);
+
+    // Prints out the bits in the bitvector for a lattice element.
+    void print(std::ostream& os);
+
+    friend FinitePowersetLattice;
+  };
 
   // Compares two lattice elements and returns a result indicating the
   // left element's relation to the right element.
-  static LatticeComparison compare(const FinitePowersetLattice& left,
-                                   const FinitePowersetLattice& right);
+  static LatticeComparison compare(const Element& left, const Element& right);
 
-  // Calculates the LUB of this current (left) lattice element with some right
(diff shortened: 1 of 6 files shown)


src/analysis/powerset-lattice-impl.h
```diff
@@ -5,61 +5,115 @@
...
 namespace wasm::analysis {
 
-template<size_t N>
-inline BitsetPowersetLattice<N> BitsetPowersetLattice<N>::getBottom() {
+inline FinitePowersetLattice::FinitePowersetLattice(size_t size) : trues(0) {
+  bitvector.reserve(size);
+  for (size_t i = 0; i < size; ++i) {
+    bitvector.push_back(false);
+  }
+}
```
tlively, line 13: This can be made a one-liner, so it can probably just go in the .h file rather than the separate impl file.

```suggestion
inline FinitePowersetLattice::FinitePowersetLattice(size_t size) : bitvector(size), trues(0) {}
```

src/analysis/powerset-lattice-impl.h
```diff
@@ -5,61 +5,115 @@
...
-template<size_t N>
-inline bool
-BitsetPowersetLattice<N>::isBottom(const BitsetPowersetLattice<N>& element) {
+inline bool FinitePowersetLattice::isBottom() {
   // Bottom lattice element is the empty set.
-  return element.value.none();
+  return trues == 0;
+}
+
+inline void FinitePowersetLattice::set(size_t index, bool value) {
```
tlively, line 31: In a follow-up where we make `FinitePowersetLattice` generic over the set element type `T`, it will be good to make this a private implementation detail and present a more set-like interface.

src/analysis/powerset-lattice-impl.h
```diff
@@ -5,61 +5,115 @@
...
+// both sides.
+inline bool
+FinitePowersetLattice::getLeastUpperBound(const FinitePowersetLattice& right) {
+  // Both must be from powerset lattice of the same set.
+  assert(right.bitvector.size() == bitvector.size());
+
+  bool modified = false;
+  for (size_t i = 0; i < bitvector.size(); ++i) {
+    if (!bitvector.at(i) && right.bitvector.at(i)) {
+      bitvector.at(i) = true;
```
tlively, line 104: It's more idiomatic to use `bitvector[i]` than `bitvector.at(i)`. If you want the error handling of `at()`, in general you can make that more explicit by adding an assertion `assert(i < bitvector.size())`. In this case, it's probably simpler to assert `bitvector.size() == right.bitvector.size()`.

src/analysis/powerset-lattice-impl.h
```diff
@@ -5,61 +5,115 @@
...
+  }
+
+  return modified;
 }
 
-template<size_t N>
-inline void BitsetPowersetLattice<N>::print(std::ostream& os) {
-  os << value;
+inline void FinitePowersetLattice::print(std::ostream& os) {
+  for (auto it = bitvector.rbegin(); it != bitvector.rend(); ++it) {
```
tlively, line 114: Why print this in reverse?

zm2he: The bitset vector printed out the values in reverse, so I did it here for backward compatibility with the tests. But yes, I should probably change the test instead.

src/analysis/lattice.h
```diff
@@ -19,32 +20,53 @@ constexpr bool has_getLeastUpperBound = std::
...
 
 template<typename T>
 constexpr bool is_lattice =
   has_compare<T>&& has_getLeastUpperBound<T>&& has_isTop<T>&& has_isBottom<T>;
 
 // Represents a powerset lattice element (i.e. a set) as a bitvector. A true
 // means that an element is present in the set.
-template<size_t N> struct BitsetPowersetLattice {
-  std::bitset<N> value;
+class FinitePowersetLattice {
```
kripken, line 34: Btw, it slightly confused me that this is "lattice" but really (as the comment explains) it is a "lattice *element*". But adding "element" to the name would probably not be worth it...

src/analysis/lattice.h
```diff
@@ -19,32 +20,53 @@ constexpr bool has_getLeastUpperBound = std::
...
+  std::vector<bool> bitvector;
+
+  // Counts the number of elements in the set (i. e. number of trues in the
+  // bitvector).
+  size_t trues;
+
+  // Size of the top lattice element must be specified, as it determines the set
+  // of the bitvector.
+  FinitePowersetLattice(size_t size);
```
kripken, line 43: I don't understand the comment. Is this function only used to create the top element, or something like that?

zm2he: It's a constructor. It actually creates the bottom element. But when I refer to the top element, I mean that the bitvector has to be the size of the top element (i.e. contain all possible set elements).

kripken: Ah, ok, then I think the comment could maybe be reworded a little. It isn't the size of the top element that matters here. It is the size of the lattice (which happens to be the size of the top element).

That is, the comment could say, "when creating a lattice element we must pass in the size of the lattice".

tlively: I would say it _is_ the size of the top element that matters here, if only because the top element happens to be the set that we are taking the powerset of. The "size of the lattice" is the number of elements in the powerset, which is not what we are referring to here.

kripken: Ah, fair enough, the lattice is defined as the powerset, correct.

src/analysis/powerset-lattice-impl.h
zm2he: Changed.

src/analysis/lattice.h
zm2he: Added lattice element class.

src/analysis/monotone-analyzer-impl.h
```diff
@@ -7,42 +7,37 @@
...
 
 namespace wasm::analysis {
-template<size_t N>
-inline BlockState<N>::BlockState(const BasicBlock* underlyingBlock)
+inline BlockState::BlockState(const BasicBlock* underlyingBlock, size_t size)
   : index(underlyingBlock->getIndex()), cfgBlock(underlyingBlock),
-    beginningState(BitsetPowersetLattice<N>::getBottom()),
-    endState(BitsetPowersetLattice<N>::getBottom()),
-    currState(BitsetPowersetLattice<N>::getBottom()) {}
+    beginningState(FinitePowersetLattice::getBottom(size)),
```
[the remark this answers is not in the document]
zm2he: Added lattice element class.

src/analysis/lattice.h
```diff
@@ -19,32 +20,53 @@ constexpr bool has_getLeastUpperBound = std::
...
 // Represents a powerset lattice element (i.e. a set) as a bitvector. A true
 // means that an element is present in the set.
-template<size_t N> struct BitsetPowersetLattice {
-  std::bitset<N> value;
+class FinitePowersetLattice {
+  std::vector<bool> bitvector;
+
+  // Counts the number of elements in the set (i. e. number of trues in the
+  // bitvector).
+  size_t trues;
```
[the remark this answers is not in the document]
zm2he: Added lattice element class, with count function for number of members in an element. `setSize` refers to count of top.

src/analysis/lattice.h
```diff
@@ -19,32 +20,53 @@ constexpr bool has_getLeastUpperBound = std::
...
+
+  // Size of the top lattice element must be specified, as it determines the set
+  // of the bitvector.
+  FinitePowersetLattice(size_t size);
+
+public:
+  void set(size_t index, bool value);
+  bool get(size_t index);
+  size_t size() { return bitvector.size(); }
+  size_t countElements() { return trues; }
```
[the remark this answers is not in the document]
zm2he: Done.

src/analysis/lattice.h
```diff
@@ -19,37 +20,89 @@ constexpr bool has_getLeastUpperBound = std::
...
+  // equivalent to the size of the Top lattice element.
+  size_t setSize;
+
+public:
+  FinitePowersetLattice(size_t setSize) : setSize(setSize) {}
+
+  // This represents an element of a powerset lattice. The element is itself a
+  // set which has set members. The bitvector tracks which possible members of
+  // the element are actually present.
+  struct Element {
```
kripken, line 48: This could now be
```suggestion
  struct Element : public std::vector<bool> {
```
That would avoid needing to declare `count` etc as we can use them from the parent.

kripken: Ah, good points. sgtm.

src/analysis/monotone-analyzer-impl.h
```diff
@@ -71,76 +60,72 @@ template<size_t N> inline void BlockState<N>::print(std::ostream& os) {
...
   for (auto it = stateBlocks.rbegin(); it != stateBlocks.rend(); ++it) {
     worklist.push(it->index);
   }
 
   while (!worklist.empty()) {
-    BlockState<N>& currBlockState = stateBlocks[worklist.front()];
+    BlockState<Lattice>& currBlockState = stateBlocks[worklist.front()];
     worklist.pop();
+
+    // Run transfer function on the block.
```
tlively, line 114: This comment doesn't seem too helpful since the method is already called `transfer`. It may be useful to give a higher-level explanation of what this means, though.

src/analysis/lattice.h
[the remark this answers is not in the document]
tlively: current naming lgtm.

src/analysis/lattice.h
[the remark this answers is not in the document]
tlively: current naming lgtm.

src/analysis/powerset-lattice-impl.h
tlively: current solution lgtm.

src/analysis/lattice.h
```diff
@@ -1,55 +1,109 @@
...
+    // we had {true, false, true}, the count would be 2. O(N) operation which
+    // iterates through the bitvector.
+    size_t count();
 
-  static BitsetPowersetLattice<N> getBottom();
-  static bool isTop(const BitsetPowersetLattice<N>& element);
-  static bool isBottom(const BitsetPowersetLattice<N>& element);
+    using iterator = std::vector<bool>::const_iterator;
+    iterator begin() { return bitvector.cbegin(); }
+    iterator end() { return bitvector.cend(); }
```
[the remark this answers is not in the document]
zm2he: Not at the present moment. I will remove it.

src/analysis/lattice.h
```diff
@@ -1,55 +1,109 @@
...
 #include <iostream>
+#include <vector>
+
+#include "wasm.h"
 
 namespace wasm::analysis {
 
 enum LatticeComparison { NO_RELATION, EQUAL, LESS, GREATER };
 
 template<typename T>
```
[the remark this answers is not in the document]
zm2he: Switched to `Lattice`.

src/analysis/lattice.h
```diff
@@ -1,55 +1,109 @@
...
+    bool get(size_t index) { return bitvector[index]; }
+    void set(size_t index, bool value) { bitvector[index] = value; }
+
+    bool isTop() { return count() == bitvector.size(); }
+    bool isBottom() { return count() == 0; }
+
+    // Calculates the LUB of this element with some right element and sets
+    // this element to the LUB in place. Returns true if this element before
+    // this method call was different than the LUB.
+    bool getLeastUpperBound(const Element& right);
```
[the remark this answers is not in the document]
zm2he: I made the change here, as it seems within the scope of the PR with respect to updating/changing the lattice.

src/analysis/lattice.h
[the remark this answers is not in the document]
zm2he: This isn't needed anywhere at present. Removed.

src/analysis/lattice.h
```diff
@@ -1,55 +1,109 @@
...
+
+    // Prints out the bits in the bitvector for a lattice element.
+    void print(std::ostream& os);
+
+    Element(Element&& source) = default;
+    Element(Element& source) = default;
+
```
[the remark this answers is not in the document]
zm2he: Done.

src/analysis/monotone-analyzer-impl.h
zm2he: Updated.

src/analysis/monotone-analyzer-impl.h
```diff
@@ -7,58 +7,47 @@
...
-// when it it is used in a get or set.
-template<size_t N> inline void BlockState<N>::visitLocalSet(LocalSet* curr) {
-  currState.value[curr->index] = false;
+// All states are set to the bottom lattice element using the lattice in this
+// constructor.
+template<typename Lattice>
+inline BlockState<Lattice>::BlockState(const BasicBlock* underlyingBlock,
+                                       Lattice& lattice)
+  : index(underlyingBlock->getIndex()), cfgBlock(underlyingBlock),
+    beginningState(std::move(lattice.getBottom())),
```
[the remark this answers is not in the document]
zm2he: Removed the `std::move` since redundant.

src/analysis/powerset-lattice-impl.h
tlively: I don't think this applies the same way to the new code structure.

tlively (approved):
LGTM!

src/analysis/lattice.h
```diff
@@ -1,55 +1,108 @@
...
+    bool get(size_t index) { return bitvector[index]; }
+    void set(size_t index, bool value) { bitvector[index] = value; }
+
+    bool isTop() { return count() == bitvector.size(); }
+    bool isBottom() { return count() == 0; }
+
+    // Calculates the LUB of this element with some right element and sets
+    // this element to the LUB in place. Returns true if this element before
+    // this method call was different than the LUB.
+    bool makeLeastUpperBound(const Element& right);
```
kripken, line 88:
```suggestion
    bool makeLeastUpperBound(const Element& other);
```
We have the convention that when there are two items with names it is `left, right` as you did below. But when there is just another one in addition to `*this` then we use `other`.

src/analysis/monotone-analyzer-impl.h
```diff
@@ -71,76 +59,75 @@ template<size_t N> inline void BlockState<N>::print(std::ostream& os) {
...
   }
 
   while (!worklist.empty()) {
-    BlockState<N>& currBlockState = stateBlocks[worklist.front()];
+    BlockState<Lattice>& currBlockState = stateBlocks[worklist.front()];
     worklist.pop();
+
+    // For each expression, applies the transfer function, using the expression,
+    // on the state of the expression it depends upon (here the next expression)
+    // to arrive at the expression's state. THe beginning and end states of the
```
kripken, line 115:
```suggestion
    // to arrive at the expression's state. The beginning and end states of the
```

src/analysis/powerset-lattice-impl.h
```diff
@@ -5,61 +5,81 @@
...
+  const FinitePowersetLattice::Element& right) {
+  // Both must be from powerset lattice of the same set.
+  assert(right.bitvector.size() == bitvector.size());
+
+  bool modified = false;
+  for (size_t i = 0; i < bitvector.size(); ++i) {
+    // Bit is flipped on left only if left is false and right is true when left
+    // and right are OR'ed together.
+    modified |= (!bitvector[i] && right.bitvector[i]);
+    bitvector[i] = bitvector[i] || right.bitvector[i];
```
kripken, line 72:
```suggestion
    bitvector[i] = bitvector[i] | right.bitvector[i];
```
It's almost certainly optimized by the compiler, but still, we just want a logical OR here and not control flow branching.

zm2he: @kripken The CI does not like the use of bitwise operands with booleans unless cast to an int:
`error: use of bitwise '|' with boolean operands [clang-diagnostic-bitwise-instead-of-logical]`.

tlively: Probably not worth working around. I would keep using `||`.

src/analysis/monotone-analyzer-impl.h
zm2he: Fixed.

src/analysis/lattice.h
zm2he: Fixed.

src/analysis/powerset-lattice-impl.h
kripken: Interesting. More stuff about C++ I did not know... `||` sgtm then

PR #5787, opened 2023-06-27
Fix opt/shrink levels when running the optimizer multiple times, Part 2

kripken:
This is a followup to #5333 . That fixed the selection of which passes to run, but
forgot to also fix the global state of the current optimize/shrink levels. This PR
fixes that. As a result, running `-O3 -Oz` will now work as expected: the first `-O3`
will run the right passes (as #5333 fixed) and while running them, the global
optimize/shrinkLevels will be `-O3` (and not `-Oz`), which this PR fixes.

A specific result of this is that `-O3 -Oz` used to inline less, since the invocation
of inlining during `-O3` thought we were optimizing for size. The new test verifies
that we do fully inline in the first `-O3` now.


src/tools/optimization-options.h
```diff
@@ -328,32 +328,53 @@ struct OptimizationOptions : public ToolOptions {
...
+    // Flush anything in the current pass runner, and then reset it to a fresh
+    // state so it is ready for new things.
+    auto flushAndReset = [&]() {
+      if (passRunner) {
+        passRunner->run();
+        passRunner.reset();
       }
-      if (pass.shrinkLevel) {
-        passRunner.options.shrinkLevel = *pass.shrinkLevel;
+      passRunner = std::make_unique<PassRunner>(&wasm, passOptions);
```
tlively, line 340: Should this be in an `else`? Otherwise it seems strange to `reset()` the old pass runner just to overwrite it.

tlively (approved):

kripken:
This also now removes the skip-pass warning if the pass was not found. With some work it could still be supported, but I'm not sure it's worth it.

The issue is that `--skip-pass=foo` used to just check if `foo` was among the passes that were skipped. That no longer works since we now create multiple pass runners internally. We could merge the skipped passes etc., of course, but it's not simple - doing it in `optimization-options` would help most use cases, but not others that don't use that logic. As this is a minor warning, I suggest we just drop it.

src/tools/optimization-options.h
kripken: It's intentional to overwrite - it provides a fresh pass runner, with the right settings etc.

I can remove the `.reset()` though, since `passRunner = ` will effectively do a reset as it overwrites. I did that now.

tlively (approved):

PR #5791, opened 2023-06-28
[wasm-metadce] Consider ref.func as introducing a dependency + fix rooting of segment offsets

vouillon:
`wasm-metadce` could wrongly consider some functions as unused since it did not take `ref.func` into consideration.

### DIFF
--- a/src/tools/wasm-metadce.cpp
+++ b/src/tools/wasm-metadce.cpp
@@ -209,7 +209,9 @@ struct MetaDCEGraph {
           // it's an import.
           dceName = parent->importIdToDCENode[parent->getGlobalImportId(name)];
         }
-        if (!parentDceName.isNull()) {
+        if (parentDceName.isNull()) {
+          parent->roots.insert(dceName);
+        } else {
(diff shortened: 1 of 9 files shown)


kripken:
Thanks! This looks mostly correct to me, but reading the code around it I think I see a problem. `InitScanner` does nothing when `parentDceName` is null, so checking that does not actually do anything. And it means we have not been rooting things properly. I think we should fix that first, so that this PR can root properly.

I can look into that fix. Unless you prefer to do it?

vouillon:
Indeed, this looked weird to me. But I did not understand the code well enough to know whether there really was a bug there. After thinking a bit more about that, I think I can write a proper fix.

vouillon:
So, there was a typo in the initial code, which was incorrectly fixed in #4246 by not rooting the global.
I have pushed a fix and a test.

kripken (approved):
Thanks! Looks great.

I think we can add some helper functions and things here to simplify the code, I'll look into that in a followup.

PR #5792, opened 2023-06-28
Limit printing of Literals in a general way

kripken:
Previously we limited printing in a single Literals. But we can have
infinitely recursive GC literals, or just huge graphs even without
infinite recursion where no single Literals is that big (but we still
get exponential blowup).

This should fix an issue I've seen recently where my fuzzing machine
hangs unresponsively. I narrowed that down to a case where we
print such a massive Literals. What appears to happen is that we print
and python captures the stdout in a buffer that grows until the OS
starts to swap and then things just lock up.


aheejin (commented):
- I didn't know we were already limiting literals printed... Then can we read those wat/wast files back?
- So before we only limited the length of the printed literal every time we print `Literals`, which can be recursive, so it didn't do anything depth-wise, which can be infinite or very long. This doesn't limit the length per `Literals` but limit the whole set of top-level `Literals` that can contain many more literals themselves in deep nests. Is my understanding correct?
- Can we have some tests? Or this is a behavior not easy to write in tests?

src/wasm/literal.cpp
```diff
@@ -627,26 +627,56 @@ std::ostream& operator<<(std::ostream& o, Literal literal) {
...
-      if (i == LITERALS_PRINT_LIMIT) {
-        o << "[..]";
-        break;
-      }
+  }
+
+  if (literals.size() == 0) {
+    return o << "()";
+  }
+
```
tlively, line 670: This special case doesn't look necessary, since the loop below will do the same thing for zero iterations.

tlively (approved):
LGTM % comment

tlively:
> * I didn't know we were already limiting literals printed... Then can we read those wat/wast files back?

Since these are literals and not expressions, they do not appear anywhere in wat/wast files.

kripken:
@aheejin 

> So before we only limited the length of the printed literal every time we print Literals, which can be recursive, so it didn't do anything depth-wise, which can be infinite or very long. This doesn't limit the length per Literals but limit the whole set of top-level Literals that can contain many more literals themselves in deep nests. Is my understanding correct?

Yes, exactly.

src/wasm/literal.cpp
kripken: Good point, I refactored that but didn't use the result of that refactoring :clown_face:

kripken:
@aheejin I looked into testing, and it's possible, I added a test now. We mostly print literals in debug contexts, but things like an uncaught wasm exception do end up printed.

While doing that I noticed that we need this logic in `Literal` and not just `Literals`, which is now fixed.

aheejin (approved):

PR #5793, opened 2023-06-29
[NFC] Add a helper to get function DCE names in wasm-metadce

kripken:
PTAL @vouillon


vouillon:
This looks good to me.

kripken:
Thanks @vouillon !

PR #5794, opened 2023-06-29
Make the Transfer Function a Generic, Interchangeable Type for the Monotone Analyzer

zm2he:
This change makes the transfer function an object separate from the monotone analyzer. The transfer function class type is a generic template of the monotone analyzer, and an instance of the transfer function is instantiated and then passed into the monotone analyzer via its constructor.

This change introduces `LivenessTransferFunction` as a transfer function for liveness analysis as an example.

### DIFF
--- a/src/analysis/liveness-transfer-function.h
+++ b/src/analysis/liveness-transfer-function.h
@@ -36,7 +36,7 @@ public:
     for (auto cfgIter = cfgBlock->rbegin(); cfgIter != cfgBlock->rend();
          ++cfgIter) {
       // Run transfer function.
-      LivenessTransferFunction::visit(*cfgIter);
+      visit(*cfgIter);
     }
     currState = nullptr;
   }
@@ -47,38 +47,35 @@ public:
   // run before earlier CFG blocks. This improves performance by
   // reducing the number of state propagations needed, since we are
   // naturally following the backward flow at the beginning.
-  void enqueueWorklist(
-    const std::vector<BlockState<FinitePowersetLattice>>& stateBlocks,
-    std::queue<Index>& worklist) {
-    for (auto it = stateBlocks.rbegin(); it != stateBlocks.rend(); ++it) {
-      worklist.push((*it).getCFGBlock()->getIndex());
+  void enqueueWorklist(CFG* cfg, std::queue<Index>& worklist) {
+    for (auto it = cfg->rbegin(); it != cfg->rend(); ++it) {
+      worklist.push((*it).getIndex());
     }
   }
 
-  // Predecessors depend on use for information.
-  using iterator =
-    std::vector<BlockState<FinitePowersetLattice>*>::const_iterator;
-  iterator depsBegin(BlockState<FinitePowersetLattice>& currBlock) {
-    return currBlock.predecessorsBegin();
-  }
-  iterator depsEnd(BlockState<FinitePowersetLattice>& currBlock) {
-    return currBlock.predecessorsEnd();
+  // Predecessors depend on current basic block for information.
+  _indirect_ptr_vec<BasicBlock> getDependents(const BasicBlock* currBlock) {
+    return currBlock->preds();
   }
 
   // Prints the intermediate states of each BlockState currBlock by applying
   // the transfer function on each expression of the CFG block. This data is
-  // not stored in the BlockState itself.
-  void print(std::ostream& os, BlockState<FinitePowersetLattice>& currBlock) {
+  // not stored in the BlockState itself. Requires the cfgBlock, and a temp
+  // copy of the input state to be passed in, where the temp copy is modified
+  // in place to produce the intermediate states.
+  void print(std::ostream& os,
+             const BasicBlock* cfgBlock,
+             FinitePowersetLattice::Element& inputState) {
     os << "Intermediate States (reverse order): " << std::endl;
-    currState = &currBlock.getInputState();
+    currState = &inputState;
     currState->print(os);
     os << std::endl;
-    auto cfgIter = currBlock.getCFGBlock()->rbegin();
+    auto cfgIter = cfgBlock->rbegin();
 
     // Since we don't store the intermediate states in the BlockState, we need
     // to re-run the transfer function on all the CFG node expressions to
     // reconstruct the intermediate states here.
-    while (cfgIter != currBlock.getCFGBlock()->rend()) {
+    while (cfgIter != cfgBlock->rend()) {
       os << ShallowExpression{*cfgIter} << std::endl;
       LivenessTransferFunction::visit(*cfgIter);
       currState->print(os);
--- a/src/analysis/liveness-transfer-function.h
+++ b/src/analysis/liveness-transfer-function.h
@@ -22,8 +22,8 @@ public:
   }
 
   // Executes the transfer function on all the expressions of the corresponding
-  // CFG node, starting with the node's input state. Returns the final output
-  // state of the node.
+  // CFG node, starting with the node's input state, and changes the input state
+  // to the final output state of the node in place.
   void transfer(const BasicBlock* cfgBlock,
                 FinitePowersetLattice::Element& inputState) {
     // If the block is empty, we propagate the state by inputState =
--- a/src/analysis/liveness-transfer-function.h
+++ b/src/analysis/liveness-transfer-function.h
@@ -10,8 +10,7 @@ class LivenessTransferFunction : public Visitor<LivenessTransferFunction> {
   FinitePowersetLattice::Element* currState;
 
 public:
-  LivenessTransferFunction(FinitePowersetLattice& lattice)
-    : currState(nullptr) {}
+  LivenessTransferFunction() : currState(nullptr) {}
 
   // Transfer function implementation. Modifies the state for a particular
   // expression type. In our current limited implementation, we just update
@@ -34,7 +33,7 @@ public:
   transfer(const BasicBlock* cfgBlock,
            FinitePowersetLattice::Element& inputState) {
     // If the block is empty, we propagate the state by inputState =
-    // outputSTate.
+    // outputState.
 
     FinitePowersetLattice::Element outputState = inputState;
     currState = &outputState;
--- a/src/analysis/liveness-transfer-function.h
+++ b/src/analysis/liveness-transfer-function.h
@@ -7,35 +7,46 @@
 namespace wasm::analysis {
 
 class LivenessTransferFunction : public Visitor<LivenessTransferFunction> {
-  FinitePowersetLattice::Element currState;
+  FinitePowersetLattice::Element* currState;
 
 public:
   LivenessTransferFunction(FinitePowersetLattice& lattice)
-    : currState(lattice.getBottom()) {}
+    : currState(nullptr) {}
 
   // Transfer function implementation. Modifies the state for a particular
   // expression type. In our current limited implementation, we just update
   // state on gets and sets of local indices.
-  void visitLocalSet(LocalSet* curr) { currState.set(curr->index, false); }
-  void visitLocalGet(LocalGet* curr) { currState.set(curr->index, true); }
+  void visitLocalSet(LocalSet* curr) {
+    if (currState) {
+      currState->set(curr->index, false);
+    }
+  }
+  void visitLocalGet(LocalGet* curr) {
+    if (currState) {
+      currState->set(curr->index, true);
+    }
+  }
 
   // Executes the transfer function on all the expressions of the corresponding
-  // CFG node.
-  void transfer(BlockState<FinitePowersetLattice>& currBlock) {
-    // If the block is empty, we propagate the state by endState = currState,
-    // then currState = beginningState.
+  // CFG node, starting with the node's input state. Returns the final output
+  // state of the node.
+  FinitePowersetLattice::Element
+  transfer(const BasicBlock* cfgBlock,
+           FinitePowersetLattice::Element& inputState) {
+    // If the block is empty, we propagate the state by inputState =
+    // outputSTate.
+
+    FinitePowersetLattice::Element outputState = inputState;
+    currState = &outputState;
 
     // Compute transfer function for all expressions in the CFG block.
-    auto cfgIter = currBlock.getCFGBlock()->rbegin();
-    currState = currBlock.getLastState();
-
-    while (cfgIter != currBlock.getCFGBlock()->rend()) {
+    for (auto cfgIter = cfgBlock->rbegin(); cfgIter != cfgBlock->rend();
+         ++cfgIter) {
       // Run transfer function.
       LivenessTransferFunction::visit(*cfgIter);
-      ++cfgIter;
     }
-
-    currBlock.getFirstState() = std::move(currState);
+    currState = nullptr;
+    return outputState;
   }
 
   // Enqueues the worklist before the worklist algorithm is run. For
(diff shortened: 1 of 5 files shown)


tlively (commented):
Initial high-level design comments.

src/analysis/liveness-transfer-function.h
```diff
@@ -0,0 +1,101 @@
+#ifndef wasm_analysis_liveness_transfer_function_h
+#define wasm_analysis_liveness_transfer_function_h
+
+#include "lattice.h"
+#include "monotone-analyzer.h"
+
+namespace wasm::analysis {
+
+class LivenessTransferFunction : public Visitor<LivenessTransferFunction> {
+  FinitePowersetLattice::Element currState;
```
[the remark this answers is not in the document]
zm2he: `currState` is always reset each time the transfer function is applied. It is a member variable to allow the visitor functions to access it.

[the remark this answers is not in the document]
tlively: Since it's logically owned by the core analysis code rather than the transfer function, perhaps it would make sense to make `currState` a pointer to the element in the analysis code rather than a separate value that has to be copied into and out of. WDYT?

src/analysis/liveness-transfer-function.h
```diff
@@ -0,0 +1,101 @@
...
+  iterator depsEnd(BlockState<FinitePowersetLattice>& currBlock) {
+    return currBlock.predecessorsEnd();
+  }
+
+  // We start at the last state and end at the first state in a
+  // backward analysis.
+  FinitePowersetLattice::Element&
+  getInputState(BlockState<FinitePowersetLattice>* currBlock) {
+    return currBlock->getLastState();
+  }
```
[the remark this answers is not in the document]
tlively: IIUC, the worklist algorithm only has to save the input state for each node, and whether the analysis is forward or backward changes what that state means in the results, but the core work list algorithm doesn't need to know anything about that. The only thing that I would expect to change between a forward and backward analysis is the dependencies between the blocks.

src/analysis/liveness-transfer-function.h
```diff
@@ -0,0 +1,101 @@
...
+    currBlock.getFirstState() = std::move(currState);
+  }
+
+  // Enqueues the worklist before the worklist algorithm is run. For
+  // liveness analysis, we are doing a backward analysis, so we want
+  // to enqueue the worklist backward so that later CFG blocks are
+  // run before earlier CFG blocks. This improves performance by
+  // reducing the number of state propagations needed, since we are
+  // naturally following the backward flow at the beginning.
+  void enqueueWorklist(
```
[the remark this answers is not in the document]
tlively: We discussed this further offline. Having the core framework determine a good initial ordering can be done with a topological sort of the dependency graph, perhaps using Tarjan’s SCC algorithm. However, this is probably best left to a separate PR.

src/analysis/liveness-transfer-function.h
```diff
@@ -7,35 +7,45 @@
...
-    : currState(lattice.getBottom()) {}
+  LivenessTransferFunction() : currState(nullptr) {}
 
   // Transfer function implementation. Modifies the state for a particular
   // expression type. In our current limited implementation, we just update
   // state on gets and sets of local indices.
-  void visitLocalSet(LocalSet* curr) { currState.set(curr->index, false); }
-  void visitLocalGet(LocalGet* curr) { currState.set(curr->index, true); }
+  void visitLocalSet(LocalSet* curr) {
+    if (currState) {
```
tlively, line 19: This can probably be an assertion instead of a check, since there should always be a `currState` when we're visiting expressions.

src/analysis/liveness-transfer-function.h
```diff
@@ -0,0 +1,101 @@
+#ifndef wasm_analysis_liveness_transfer_function_h
+#define wasm_analysis_liveness_transfer_function_h
+
+#include "lattice.h"
+#include "monotone-analyzer.h"
+
+namespace wasm::analysis {
+
+class LivenessTransferFunction : public Visitor<LivenessTransferFunction> {
+  FinitePowersetLattice::Element* currState;
```
kripken, line 10:
```suggestion
  FinitePowersetLattice::Element* currState = nullptr;
```
This is a little more idiomatic than doing `currState(nullptr)` in modern C++, I think. It also scales better with multiple constructors (which this doesn't have yet, but might some day).

src/analysis/liveness-transfer-function.h
```diff
@@ -0,0 +1,101 @@
...
+
+public:
+  LivenessTransferFunction() : currState(nullptr) {}
+
+  // Transfer function implementation. Modifies the state for a particular
+  // expression type. In our current limited implementation, we just update
+  // state on gets and sets of local indices.
```
kripken, line 17: Is there anything more in a non-limited future version? I think that's all there is to liveness. So I would shorten this comment. I'd also add a brief reminder of how liveness works. So perhaps the whole comment could be:
```
Transfer function implementation. A local becomes live before a
get, and dead before a set.
```
(that also reminds the reader of the reverse ordering that is used below)

src/analysis/liveness-transfer-function.h
```diff
@@ -0,0 +1,101 @@
...
+  }
+  void visitLocalGet(LocalGet* curr) {
+    if (currState) {
+      currState->set(curr->index, true);
+    }
+  }
+
+  // Executes the transfer function on all the expressions of the corresponding
+  // CFG node, starting with the node's input state. Returns the final output
+  // state of the node.
```
kripken, line 31: This description is very generic, and indeed it looks like there is nothing specific to liveness in the function. Could this not be handled by the parent? That is, the parent could have a flag whether the analysis is forward or backwards (or we could have a ForwardAnalysis and BackwardAnalysis). Then all `LivenessTransferFunction` would contain is just the LocalGet/LocalSet logic, which is all that is specific to liveness.

Likewise, in this idea, `enqueueWorklist` can also be in the parent, which again I think makes sense since it just depends on the order of the analysis.

My motivation here is that we'll have many backwards analyses, and it would be good that we don't duplicate the backwards-ordering code in them all.

src/analysis/liveness-transfer-function.h
zm2he: I was not sure if more logic was needed to account for copies as was done in the `LivenessWalker` . Also whether we wanted to evaluate liveness effects for other expressions. However, for the basic liveness functionality, you're right, that's all there is. I have changed the comment according to your suggestion.

src/analysis/liveness-transfer-function.h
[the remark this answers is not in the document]
zm2he: Will be left to a separate PR.

src/analysis/liveness-transfer-function.h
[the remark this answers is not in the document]
zm2he: Good idea, done.

src/analysis/liveness-transfer-function.h
zm2he: I agree that there is a lot of generic forward/backward functionality which could be re-used. My reasons for doing this are:

- It abstracts away the visitor implementation, in case in future transfer functions, we want to use a different implementation, keep state between expressions, or select expressions types differently (e.g. through casts).
- Right now, the state is set as a pointer member variable of the transfer function so that it can be accessed by the visitor functions without needing it to be passed as a parameter and thus changing the visitor API. It seems cleaner to set it once, mutate it a bunch of times, and then be done instead of resetting the pointer every time we visit an expression.
- The current implementation means nothing has to be parameterized in the parent, and the user is free to implement any arbitrary custom analysis flow, not just forward or backward. Admittedly, there are drawbacks to this, as you pointed out.
- The CFG basic block level seemed a good delineation point between the transfer function and the parent.

I'm not entirely decided about this method either.

src/analysis/liveness-transfer-function.h
```diff
@@ -7,45 +7,38 @@
...
-    currState = &outputState;
+    currState = &inputState;
 
-    // Compute transfer function for all expressions in the CFG block.
+    // This is a backward analysis, so we start from the end of the CFG
+    // and evaluate every expression until the beginning.
     for (auto cfgIter = cfgBlock->rbegin(); cfgIter != cfgBlock->rend();
          ++cfgIter) {
       // Run transfer function.
       LivenessTransferFunction::visit(*cfgIter);
```
tlively, line 39: Can this just be `visit(*cfgIter)`?

src/analysis/monotone-analyzer.h
```diff
@@ -11,48 +11,123 @@
...
   std::vector<BlockState<Lattice>*> successors;
 
-  friend MonotoneCFGAnalyzer<Lattice>;
+public:
+  // All states are set to the bottom lattice element in this constructor.
+  BlockState(const BasicBlock* underlyingBlock, Lattice& lattice);
+
+  // Accessors.
+  typename Lattice::Element& getInputState() { return inputState; }
+  const BasicBlock* getCFGBlock() const { return cfgBlock; }
```
tlively, line 33: These are only used internally by the monotone analyzer code itself, right? If so, I don't see any benefit to using accessor functions rather than just accessing these members directly.

tlively (approved):
LGTM % comments.

Besides the future work on `enqueueWorklist`, I think the interface between the core monotone analysis code and the user-provided transfer function object looks good, although as Alon alluded to, there might be further opportunities to factor out code that will be shared between multiple transfer functions. I think addressing those opportunities in later PRs would be fine.

src/analysis/liveness-transfer-function.h
tlively: It does seem like a good idea to have a `VisitorTransferFunction` CRTP utility that would handle everything here except for the `visit` methods themselves. I imagine that `VisitorTransferFunction` could have a template parameter saying whether it is a forward or backward analysis (or neither) and that it could provide a default implementation of `depsBegin` and `depsEnd` accordingly, which the particular subclass would still be able to replace if necessary.

src/analysis/monotone-analyzer.h
zm2he: We discussed offline that the `BlockState` contents should be made public, but not be exposed to any class other than the `MonotoneCFGAnalyzer`. One reason is that friending `MonotoneCFGAnalyzer` is cumbersome because it has a `TransferFunction` template that `BlockState` should not be aware of.

Since we are doing this, it is also not necessary to store the predecessors and successors in `BlockState` since we can just get the corresponding index from the underling CFG `BasicBlock` and look up the index in `MonotoneCFGAnalyzer`. I have therefore taken out the predecessors and successors from `BlockState` also.

src/analysis/liveness-transfer-function.h
```diff
@@ -47,38 +47,35 @@ class LivenessTransferFunction : public Visitor<LivenessTransferFunction> {
...
-  // Predecessors depend on use for information.
-  using iterator =
-    std::vector<BlockState<FinitePowersetLattice>*>::const_iterator;
-  iterator depsBegin(BlockState<FinitePowersetLattice>& currBlock) {
-    return currBlock.predecessorsBegin();
-  }
-  iterator depsEnd(BlockState<FinitePowersetLattice>& currBlock) {
-    return currBlock.predecessorsEnd();
+  // Predecessors depend on current basic block for information.
+  _indirect_ptr_vec<BasicBlock> getDependents(const BasicBlock* currBlock) {
```
tlively, line 57: `_indirect_ptr_vec` is supposed to be a private implementation detail, so can we use something like `using iterator BasicBlock::Predecessors::iterator` like we had before?

src/analysis/monotone-analyzer.h
```diff
@@ -143,6 +111,7 @@ struct MonotoneCFGAnalyzer {
   Lattice lattice;
   TransferFunction transferFunction;
   std::vector<BlockState<Lattice>> stateBlocks;
+  CFG* cfg = nullptr;
```
tlively, line 114: Can we make the analyzer own the CFG?

src/analysis/liveness-transfer-function.h
```diff
@@ -47,38 +47,35 @@ class LivenessTransferFunction : public Visitor<LivenessTransferFunction> {
...
   // reducing the number of state propagations needed, since we are
   // naturally following the backward flow at the beginning.
-  void enqueueWorklist(
-    const std::vector<BlockState<FinitePowersetLattice>>& stateBlocks,
-    std::queue<Index>& worklist) {
-    for (auto it = stateBlocks.rbegin(); it != stateBlocks.rend(); ++it) {
-      worklist.push((*it).getCFGBlock()->getIndex());
+  void enqueueWorklist(CFG* cfg, std::queue<Index>& worklist) {
+    for (auto it = cfg->rbegin(); it != cfg->rend(); ++it) {
+      worklist.push((*it).getIndex());
```
tlively, line 52: It will be more forward compatible with interprocedural analyses to enqueue `BasicBlock*` rather than indices.

tlively (commented):
Nice, this looks simpler overall now

src/analysis/liveness-transfer-function.h
zm2he: @tlively `getDependents` must be able to return both predecessor, successor, or some arbitrary iterator though.

tlively: Do we have to require it to have some particular type, or can we just let each `TransferFunction` return whatever iterator types it wants, though?

src/analysis/monotone-analyzer.h
zm2he: The CFG has been changed to a reference, but as we discussed offline, the analyzer should not own the CFG because we may want to do multiple analyses on the same CFG or possibly have the transfer function pre-process the CFG also.

src/analysis/liveness-transfer-function.h
zm2he: As discussed offline, we will assert that the method exists with the correct parameter types, but for now we won't be checking the return type, since we don't know right now how to assert on all iterables without a C++20 concept.

src/analysis/liveness-transfer-function.h
zm2he: Saved for later PR.

src/analysis/monotone-analyzer.h
```diff
@@ -71,47 +51,64 @@ constexpr bool is_TransferFunction = has_transfer<TransferFunction, Lattice>&&
...
 
 // Returns an iterable to the CFG BasicBlocks which depend on currBlock for
-// information (e.g. predecessors in a backward analysis). Used to select
-// which blocks to propagate to after applying the transfer function to
-// a block.
+// information (e.g. predecessors in a backward analysis). Used to select which
+// blocks to propagate to after applying the transfer function to a block. At
+// present, we do not know how to use static assertions to check that the return
+// type is any iterable, not just a specific iterable class, so we only assert
+// that the method exists with the following parameters.
```
tlively, line 83:
```suggestion
// blocks to propagate to after applying the transfer function to a block. At
// present, we allow this function to return any iteratable, so we only assert
// that the method exists with the following parameters.
```

src/analysis/monotone-analyzer.h
```diff
@@ -71,47 +51,64 @@ constexpr bool is_TransferFunction = has_transfer<TransferFunction, Lattice>&&
...
 template<typename Lattice, typename TransferFunction>
-struct MonotoneCFGAnalyzer {
+class MonotoneCFGAnalyzer {
   static_assert(is_lattice<Lattice>);
   static_assert(is_TransferFunction<TransferFunction, Lattice>);
 
-  MonotoneCFGAnalyzer(Lattice lattice, TransferFunction transferFunction)
-    : lattice(lattice), transferFunction(transferFunction) {}
+  Lattice lattice;
+  TransferFunction transferFunction;
```
tlively, line 96: Might as well make these references, too, to avoid copying.

tlively (approved):
Please put the comments for each method above the code for checking that the method exists rather than below the code.

 Apart from that and a couple final comments, lgtm!

PR #5799, opened 2023-07-03
[DeadArgumentElimination] Optimize function body after parameter refinement

vouillon:
It can be useful to optimize a function body after its parameters are refined, not just when some parameters are removed.


test/lit/passes/dae-refine-params-and-optimize.wast
```diff
@@ -0,0 +1,31 @@
...
+ ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT: )
+ (func $len (param (ref eq)) (result i32)
+  (drop
+   (block $not_array (result (ref eq))
+     (return
+       (array.len
+         (br_on_cast_fail $not_array (ref eq) (ref array) (local.get 0))))))
+  (i32.const -1))
```
kripken, line 23: Please indent this in the "unrolled" way that we use in our tests,
```wat
 (func $len (param (ref eq)) (result i32)
  (drop
   (block $not_array (result (ref eq))
    (return
     (array.len
      (br_on_cast_fail $not_array (ref eq) (ref array)
       (local.get 0)
      )
     )
    )
   )
  )
  (i32.const -1)
 )
```

test/lit/passes/dae-refine-params-and-optimize.wast
```diff
@@ -0,0 +1,31 @@
...
+ ;; CHECK-NEXT:  (drop
+ ;; CHECK-NEXT:   (return
+ ;; CHECK-NEXT:    (array.len
+ ;; CHECK-NEXT:     (local.get $0)
+ ;; CHECK-NEXT:    )
+ ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT: )
+ (func $len (param (ref eq)) (result i32)
+  (drop
```
kripken, line 18: Maybe worth a comment here about what is expected and what is not. Maybe:

"The `br` is optimized away thanks to the refined type, but the output still has some unoptimized code (`--dce --remove-unused-brs` would get rid of the `drop-return` pattern here, but that is unrelated to this test)."

kripken (commented):
Thanks! Code looks good.

test/lit/passes/dae-refine-params-and-optimize.wast
vouillon: I was mostly interested in that compiling with `-O3` does not produce never-branching `br_on_cast_fail` (since they are crashing V8).

But I'm wondering about this remaining `drop`.  What happens is that the `remove-unused-brs` pass is scheduled after the `dce` pass in `PassRunner::addDefaultFunctionOptimizationPasses`, even though it can expose some dead code.

kripken: Yeah, this does show that the current ordering of passes is not optimal. But I'm not sure we can improve it without increasing compile times (by adding another run of `--dce`). This might be worth thinking about, but in general my feeling is that GC code will want multiple iterations of optimization anyhow, like `-O3 -O3`, which would handle this in most cases.  But at some point perhaps looking into a more efficient pass ordering could make sense.

kripken (approved):

test/lit/passes/dae-refine-params-and-optimize.wast
```diff
@@ -0,0 +1,42 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
+
+;; RUN: foreach %s %t wasm-opt -all --dae-optimizing -S -o - | filecheck %s
+
+;; The br_on_cast_fail is optimized away thanks to the refined type. The output still has some unoptimized code (an additional --dce pass would get rid of the drop-return pattern here), but this is not directly relevant to this test."
```
kripken, line 5:
```suggestion
;; The br_on_cast_fail is optimized away thanks to the refined type. The output still has some unoptimized code (an additional --dce pass would get rid of the drop-return pattern here), but that is not directly relevant to this test.
```

PR #5800, opened 2023-07-04
Generalize Finite Powerset Lattice to Any Given Type

zm2he:
This change enables the `FinitePowersetLattice` to be directly used to represent powerset lattices constructed from a set of arbitrary type members. The `FinitePowersetLattice` is made to map members to bitvector indices. During construction, an ordered vector of members of some templated type `Type` is passed in, and this ordering is used to determine the mapping of `Type` members to bitvector indices. Instead of having users directly manipulate the bitvector for each lattice element, methods `add` and `remove` are introduced to add or remove a lattice element member of `Type`.

### DIFF
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -150,6 +150,7 @@ public:
   static LatticeComparison compare(const Element& left, const Element& right) {
     return FiniteIntPowersetLattice::compare(left, right);
   }
+
   Element getBottom() { return intLattice.getBottom(); }
 };
 
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -40,34 +40,17 @@ constexpr bool is_lattice = has_getBottom<Lattice>&& has_compare<Lattice>&&
   has_makeLeastUpperBound<typename Lattice::Element>&& has_isTop<
     typename Lattice::Element>&& has_isBottom<typename Lattice::Element>;
 
-// Represents a powerset lattice which is constructed from a finite set which
-// can be represented by a bitvector. Set elements are represented by
-// FinitePowersetLattice::Element, which represents members present in each
-// element by bits in the bitvector. Type is the type of a member in a lattice
-// element.
-template<typename Type> class FinitePowersetLattice {
-
+// Represents a powerset lattice constructed from a finite set of consecutive
+// integers from 0 to n which can be represented by a bitvector. Set elements
+// are represented by FinitePowersetLattice::Element, which represents members
+// present in each element by bits in the bitvector.
+class FiniteIntPowersetLattice {
   // The size of the set that the powerset lattice was created from. This is
   // equivalent to the size of the Top lattice element.
   size_t setSize;
 
-  // Maps a bitvector index to some element member of Type.
-  // Used to produce initial ordering of element members.
-  std::vector<Type> memberOrdering;
-
-  // Maps an element member of Type to a bitvector index.
-  std::unordered_map<Type, size_t> memberMapping;
-
 public:
-  // Takes in an ordered list of all elements belonging to the set to create
-  // the powerset lattice from (i.e. the powerset lattice top element). This
-  // is used for mapping these to bitvector indices.
-  FinitePowersetLattice(std::vector<Type> setMembers)
-    : setSize(setMembers.size()), memberOrdering(setMembers) {
-    for (size_t i = 0; i < setSize; ++i) {
-      memberMapping[memberOrdering[i]] = i;
-    }
-  }
+  FiniteIntPowersetLattice(size_t setSize) : setSize(setSize) {}
 
   // This represents an element of a powerset lattice. The element is itself a
   // set which has set members. The bitvector tracks which possible members of
@@ -107,27 +90,8 @@ public:
     // Prints out the bits in the bitvector for a lattice element.
     void print(std::ostream& os);
 
-    friend FinitePowersetLattice;
+    friend FiniteIntPowersetLattice;
   };
-
-  Type indexToMember(size_t index) { return memberOrdering[index]; }
-
-  Type memberToIndex(Type member) { return memberMapping[member]; }
-
-  // Adds member to element.
-  void add(Element* element, Type member) {
-    element->set(memberMapping[member], true);
-  }
-
-  // Removes member from element.
-  void remove(Element* element, Type member) {
-    element->set(memberMapping[member], false);
-  }
-
-  // Checks if member is included in the element set.
-  bool exists(Element* element, Type member) {
-    return element->get(memberMapping[member]);
-  }
 
   // Compares two lattice elements and returns a result indicating the
   // left element's relation to the right element.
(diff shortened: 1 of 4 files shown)


src/analysis/lattice.h
```diff
@@ -42,15 +43,31 @@ constexpr bool is_lattice = has_getBottom<Lattice>&& has_compare<Lattice>&&
...
+  std::vector<Type> memberOrdering;
+
+  // Maps an element member of Type to a bitvector index.
+  std::unordered_map<Type, size_t> memberMapping;
+
 public:
-  FinitePowersetLattice(size_t setSize) : setSize(setSize) {}
+  // Takes in an ordered list of all elements belonging to the set to create
+  // the powerset lattice from (i.e. the powerset lattice top element). This
+  // is used for mapping these to bitvector indices.
```
tlively, line 64: edits for clarity
```suggestion
  // Takes in an ordered list of all elements of the set to create
  // the powerset lattice from (i.e. the powerset lattice top element). This
  // is used for mapping the elements to bitvector indices.
```

src/analysis/lattice.h
```diff
@@ -42,15 +43,31 @@ constexpr bool is_lattice = has_getBottom<Lattice>&& has_compare<Lattice>&&
...
+  FinitePowersetLattice(std::vector<Type> setMembers)
+    : setSize(setMembers.size()), memberOrdering(setMembers) {
```
tlively, line 66: It probably makes sense to move the vector into the lattice to avoid copying it. I see that you are already doing that in the calling code, but we need this part as well to avoid all copies.

```suggestion
  FinitePowersetLattice(std::vector<Type>&& setMembers)
    : setSize(setMembers.size()), memberOrdering(std::move(setMembers)) {
```

src/analysis/lattice.h
```diff
@@ -42,15 +43,31 @@ constexpr bool is_lattice = has_getBottom<Lattice>&& has_compare<Lattice>&&
...
 // FinitePowersetLattice::Element, which represents members present in each
-// element by bits in the bitvector.
-class FinitePowersetLattice {
+// element by bits in the bitvector. Type is the type of a member in a lattice
+// element.
+template<typename Type> class FinitePowersetLattice {
 
   // The size of the set that the powerset lattice was created from. This is
   // equivalent to the size of the Top lattice element.
   size_t setSize;
```
tlively, line 52: We probably don't need to continue storing `setSize` separately if we can now just use `memberOrdering.size()`.

tlively (commented):
It occurred to me that another way to do this would be to keep the old implementation as something like `FiniteIntPowersetLattice` and build a new `FinitePowersetLattice` in terms of that one. That would allow us to skip the extra map lookups in cases where the elements really are just the integers 0 to _n - 1_.

OTOH, things are simpler if we just have a single, more general powerset lattice, so this approach lgtm as well.

src/analysis/lattice.h
```diff
@@ -42,15 +43,31 @@ constexpr bool is_lattice = has_getBottom<Lattice>&& has_compare<Lattice>&&
 // Represents a powerset lattice which is constructed from a finite set which
 // can be represented by a bitvector. Set elements are represented by
 // FinitePowersetLattice::Element, which represents members present in each
-// element by bits in the bitvector.
-class FinitePowersetLattice {
+// element by bits in the bitvector. Type is the type of a member in a lattice
+// element.
+template<typename Type> class FinitePowersetLattice {
```
kripken, line 48: `Type` is a very important class in general in our codebase, so overloading the term here can be a little confusing I think. Perhaps just `T`? That is what we use for similar templating in `SmallVector` and `SmallSet` for example.

test/gtest/cfg.cpp
```diff
@@ -147,11 +147,19 @@ End
...
-  LivenessTransferFunction transferFunction;
+  size_t numLocals = wasm.getFunction("bar")->getNumLocals();
 
-  MonotoneCFGAnalyzer<FinitePowersetLattice, LivenessTransferFunction> analyzer(
-    lattice, transferFunction);
+  std::vector<Index> localIndices(numLocals);
+
+  for (Index i = 0; i < numLocals; ++i) {
+    localIndices[i] = i;
+  }
```
kripken, line 156: Maybe we could add a constructor to handle this common case of the trivial `f(x)=x` mapping?

src/analysis/lattice.h
zm2he: `setSize` is still needed for `FiniteIntPowersetLattice` (the plain bitvector class). However, it it doesn't need to be additionally instantiated for `FinitePowersetLattice`.

zm2he:
> It occurred to me that another way to do this would be to keep the old implementation as something like `FiniteIntPowersetLattice` and build a new `FinitePowersetLattice` in terms of that one. That would allow us to skip the extra map lookups in cases where the elements really are just the integers 0 to _n - 1_.
> 
> OTOH, things are simpler if we just have a single, more general powerset lattice, so this approach lgtm as well.

Have made `FinitePowersetLattice` a composition of `FiniteIntPowersetLattice`.

test/gtest/cfg.cpp
zm2he: If we use `FiniteIntPowersetLattice` now, we should no longer need handle this trivial case. Of course, I can still additionally implement this for `FinitePowersetLattice`.

src/analysis/lattice.h
```diff
@@ -101,7 +101,57 @@ class FinitePowersetLattice {
...
+  // Checks if member is included in the element set.
+  bool exists(Element* element, T member) {
+    return element->get(memberIndices[member]);
+  }
+
+  // We use implementations from FiniteIntPowersetLattice here.
+  static LatticeComparison compare(const Element& left, const Element& right) {
+    return FiniteIntPowersetLattice::compare(left, right);
+  }
+  Element getBottom() { return intLattice.getBottom(); }
```
tlively, line 153: Just for even spacing of methods

```suggestion

  Element getBottom() { return intLattice.getBottom(); }
```

tlively (commented):
Code LGTM. It would be good to add some unit tests of `FinitePowersetLattice`, since the only existing users use the `FiniteIntPowersetLattice` instead.

zm2he:
> Code LGTM. It would be good to add some unit tests of `FinitePowersetLattice`, since the only existing users use the `FiniteIntPowersetLattice` instead.

I think when the reaching definitions analysis is implemented, that will be a good opportunity for adding unit tests using it.

src/analysis/lattice.h
```diff
@@ -39,18 +40,17 @@ constexpr bool is_lattice = has_getBottom<Lattice>&& has_compare<Lattice>&&
...
 
-// Represents a powerset lattice which is constructed from a finite set which
-// can be represented by a bitvector. Set elements are represented by
-// FinitePowersetLattice::Element, which represents members present in each
-// element by bits in the bitvector.
-class FinitePowersetLattice {
-
+// Represents a powerset lattice constructed from a finite set of consecutive
+// integers from 0 to n which can be represented by a bitvector. Set elements
+// are represented by FinitePowersetLattice::Element, which represents members
```
kripken, line 45:
```suggestion
// are represented by FiniteIntPowersetLattice::Element, which represents members
```

kripken (approved):

tlively:
> > Code LGTM. It would be good to add some unit tests of `FinitePowersetLattice`, since the only existing users use the `FiniteIntPowersetLattice` instead.
> 
> I think when the reaching definitions analysis is implemented, that will be a good opportunity for adding unit tests using it.

After talking offline, we decided it would be good to add a few unit tests here just to make sure that new code has tests at the time it is landed.

zm2he:
> > > Code LGTM. It would be good to add some unit tests of `FinitePowersetLattice`, since the only existing users use the `FiniteIntPowersetLattice` instead.
> > 
> > 
> > I think when the reaching definitions analysis is implemented, that will be a good opportunity for adding unit tests using it.
> 
> After talking offline, we decided it would be good to add a few unit tests here just to make sure that new code has tests at the time it is landed.

Unit test added.

test/gtest/cfg.cpp
```diff
@@ -255,3 +255,42 @@ End
...
+  FinitePowersetLattice<std::string>::Element element1 = lattice.getBottom();
+
+  EXPECT_EQ(element1.isBottom(), true);
+
+  lattice.add(&element1, "c");
+  lattice.add(&element1, "d");
+  lattice.add(&element1, "a");
+
+  EXPECT_EQ(element1.isBottom(), false);
+  EXPECT_EQ(element1.isTop(), false);
```
tlively, line 273:
```suggestion
  EXPECT_FALSE(element1.isBottom());
  EXPECT_FALSE(element1.isTop());
```

test/gtest/cfg.cpp
```diff
@@ -255,3 +255,42 @@ End
 
   EXPECT_EQ(ss.str(), analyzerText);
 }
+
+TEST_F(CFGTest, FinitePowersetLatticeFunctioning) {
+
+  std::vector<std::string> initialSet = {"a", "b", "c", "d", "e", "f"};
+  FinitePowersetLattice<std::string> lattice(std::move(initialSet));
+
+  FinitePowersetLattice<std::string>::Element element1 = lattice.getBottom();
```
tlively, line 264: No need to write out such a verbose type.
```suggestion
  auto element1 = lattice.getBottom();
```

test/gtest/cfg.cpp
```diff
@@ -255,3 +255,42 @@ End
...
+  EXPECT_EQ(element1.isBottom(), true);
+
+  lattice.add(&element1, "c");
+  lattice.add(&element1, "d");
+  lattice.add(&element1, "a");
+
+  EXPECT_EQ(element1.isBottom(), false);
+  EXPECT_EQ(element1.isTop(), false);
+
+  FinitePowersetLattice<std::string>::Element element2 = element1;
```
tlively, line 275:
```suggestion
  auto element2 = element1;
```

test/gtest/cfg.cpp
```diff
@@ -255,3 +255,42 @@ End
...
+  element1.print(ss);
+  ss << std::endl;
+  element2.print(ss);
+  ss << std::endl;
+  element2.makeLeastUpperBound(element1);
+  element2.print(ss);
+
+  auto resultText = R"result(101100
+100101
+101101)result";
```
tlively, line 293: Since the indentation is a little weird on these multi-line string literals, it would be better to test these one at a time.

tlively (approved):
LGTM % comments about how to improve the test.

PR #5802, opened 2023-07-06
[NFC] Fix check lines in nominal-good.wast

tlively:
Delete old, unused "NOMINAL" check lines and replace the sole remaining check
prefix, "HYBRID", with the standard "CHECK".

PR #5801, opened 2023-07-06
Print supertype declarations using the standard format

tlively:
Use the standard "(sub $super ...)" format instead of the non-standard
"XXX_supertype ... $super" format. In a follow-on PR implementing final types,
this will allow us to print and parse the standard text format for final types
right away with a smaller diff.

PR #5803, opened 2023-07-06
Initial support for `final` types

tlively:
Implement support in the type system for final types, which are not allowed to
have any subtypes. Final types are syntactically different from similar
non-final types, so type canonicalization is made aware of finality. Similarly,
TypeMerging and TypeSSA are updated to work correctly in the presence of final
types as well.

Implement binary and text parsing and emitting of final types. Use the standard
text format to represent final types and interpret the non-standard
"struct_subtype" and friends as non-final. This allows a graceful upgrade path
for users currently using the non-standard text format, where they can update
their code to use final types correctly at the point when they update to use the
standard format. Once users have migrated to using the fully expanded standard
text format, we can update update Binaryen's parsers to interpret the MVP
shorthands as final types to match the spec without breaking those users.

To make it safe for V8 to independently start interpreting types declared
without `sub` as final, also reserve that shorthand encoding only for types that
have no strict subtypes.


tlively:
* main
  * **PR #5801** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5801" target="_blank">Graphite</a> 
    * **PR #5802** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5802" target="_blank">Graphite</a> 
      * **PR #5803** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5803" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5803?utm_source=stack-comment).

test/lit/isorecursive-good.wast
```diff
@@ -7,66 +7,82 @@
   (rec
     ;; CHECK:      (rec
     ;; CHECK-NEXT:  (type $super-struct (struct (field i32)))
-    (type $super-struct (struct i32))
+    (type $super-struct (sub (struct i32)))
```
kripken, line 10: Why did this change?

tlively: This is to test the new parsing logic that allows using `sub` without specifying any supertype.

kripken (approved):

PR #5805, opened 2023-07-07
GUFA: Refine casts

kripken:
If we see `(ref.cast $A)` but we have inferred that a more refined type will
be present there at runtime `$B` then we can refine the cast to `(ref.cast $B)`.

We could do the same even when a cast is not present, but that would
increase code size. This optimization keeps code size constant.

This affects a couple of casts on the j2wasm binary.


tlively (approved):
Nice!

PR #5807, opened 2023-07-10
Add new V8 flag for final types and remove old removed ones

kripken:


tlively (approved):
Does the final types flag help avoid problems?

kripken:
I don't see a difference as mentioned here:

https://bugs.chromium.org/p/v8/issues/detail?id=14153#c2

But it seems forward-looking to add that flag, as it should help? Or is there a downside?

tlively:
Got it. I would expect that V8 flag to be removed shortly after we complete the transition to final types, but I guess it might help in the meantime. LGTM.

PR #5808, opened 2023-07-11
Add a command line flag enabling standard final types

tlively:
To avoid breaking users who have not updated their frontends yet, Binaryen still
interprets binary and text type definitions that do not use `sub` or `sub final`
as non-final, even though the spec says that they should be final.

Add a flag, `--standard-final-types`, changing this behavior to match the spec
for use in validating updated frontends and to allow compatibility with
standard-compliant engines. Update the fuzzer to use this flag for compatibility
with V8.


tlively:
* main
  * **PR #5808** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5808" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5808?utm_source=stack-comment).

tlively:
The fuzzer quickly crashes on this because it picks up initial contents that are not valid under `--standard-final-types`. I think the best fix will be to fix all the tests to parse correctly under that flag.

test/lit/final-types.wast
```diff
@@ -0,0 +1,40 @@
...
+
+;; Test that the shorthand type declaration format is considered final if and
+;; only if --standard-final-types is supplied.
+
+(module
+  ;; CHECK:      (type $short (struct ))
+  (type $short (struct))
+  ;; CHECK:      (type $long (sub final (struct )))
+  ;; FINAL:      (type $long (struct ))
+  (type $long (sub final (struct)))
```
kripken, line 17: Can we have `sub` without final to show that case as well here?

kripken (approved):

tlively:
The fuzzer found a problem that was caused by two types that differed in finality being emitted into the text format with the same generated name, which caused the TrapsNeverHappen fuzz handler to parse the filtered wast incorrectly, leading to an unexpected trap. I'll fix that in a separate prerequisite PR that changes how type names are generated.

PR #5809, opened 2023-07-11
memory.init fixes for 64 bit

dakenf:
wasm-opt ignores pointer type for 64 bit memory and generates 32 bit local.get and memory.fill/init calls
memory.init is used with pthreads, so it fixes this issue https://github.com/emscripten-core/emscripten/issues/17485
and allows emscripten to complie code with MEMORY64 + PTHREADS

Here is original wasm code generated by llvm in 64 bit with pthreads:
```
(func $__wasm_init_tls (type 5) (param i64)
    local.get 0
    global.set $__tls_base
    local.get 0
    i32.const 0
    i32.const 520
    memory.init $.tdata
    call $__wasm_apply_global_tls_relocs)
```
And these errors are generated when wasm-opt is executed (i've included only 1 but there are about 8):
```
[wasm-validator error in function __wasm_init_tls] local.set's value type must be correct, on
(local.tee $1
(local.get $0)
)
[wasm-validator error in function __wasm_init_tls] i32 != i64: memory.fill dest must match memory index type, on
(memory.fill
(local.tee $1
(local.get $0)
)
(i32.const 0)
(i64.const 32)
)
```


src/passes/MemoryPacking.cpp
```diff
@@ -721,12 +722,12 @@ void MemoryPacking::createReplacements(Module* module,
...
           builder.makeConstPtr(c->value.getInteger() + bytesWritten, ptrType);
       } else {
-        auto* get = builder.makeLocalGet(-1, Type::i32);
+        auto* get = builder.makeLocalGet(-1, ptrType);
         getVars.push_back(&get->index);
         dest = get;
         if (bytesWritten > 0) {
-          Const* addend = builder.makeConst(int32_t(bytesWritten));
-          dest = builder.makeBinary(AddInt32, dest, addend);
+          Const* addend = builder.makeConst(is64 ? int64_t(bytesWritten) : int32_t(bytesWritten));
```
kripken, line 729: We have a helper for this:
```suggestion
          Const* addend = builder.makeConstPtr(bytesWritten, ptrType);
```

kripken (commented):
Thanks!

Please also add a test for this, maybe in a new file alongside `test/lit/passes/memory-packing_zero-filled-memory.wast`

test/lit/passes/memory-packing_zero-filled-memory64.wast
```diff
@@ -0,0 +1,342 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
+
+;; RUN: wasm-opt %s --memory-packing -all --zero-filled-memory --enable-memory64 -S -o - | filecheck %s
+
+(module
```
kripken, line 5: Is this file's input copied from an existing test? If so, it would be better to add a line ending in `filecheck %s --check-prefix WASM64`, in addition to the normal line ending in `filecheck %s`. Then the same input will be run in both modes in the same file.

I was thinking that a small new file would be good enough. But actually, full coverage in the main existing file might be better, to cover more things, with a second filecheck line.

kripken (commented):
Code looks good, just the remaining test issue.

test/lit/passes/memory-packing_zero-filled-memory64.wast
```diff
@@ -0,0 +1,38 @@
...
+ ;; CHECK-NEXT:   (i32.const 0)
+ ;; CHECK-NEXT:   (i64.const 8)
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT: )
+ (func $__wasm_init_tls (type 0) (param i64)
+     (memory.init $.tdata
+       (local.get 0)
+       (i32.const 0)
+       (i32.const 8)))
+ (data $.tdata "\00\00\00\00\00\00\00\00")
```
kripken, line 37: By convention we put all the `data` together, please move it up to the others.

kripken (commented):
Thanks, lgtm aside from a minor test comment.

kripken (approved):

PR #5812, opened 2023-07-12
[WasmGC] Add an experimental pass to pull call casts backwards, OptimizeCallCasts

kripken:
A Java/Kotlin/etc. class method gets a `this` pointer whose type is usually just "object",
and it then immediately casts it to the right type for the class. If we've devirtualized
the call, so we have a direct call to it, then we may end up with this:

```wat
(call $Foo.method
  (ref.cast ?
    (this)))

(func $Foo.method (param $this $object)
  (ref.cast $Foo
    (local.get $this))
```

The cast with `?` may cast to `Foo` or it may cast to an intermediate type, or there may be no cast there at all, depending on what optimizations ended up with there. But since we are calling `Foo.method` we can infer that it must be a `Foo`, and can cast to `Foo` right before the call. That then lets us refine `Foo.method`'s type. And then we may end up with fewer casts, since we have one cast before the call, and none after.

However, this has tradeoffs:

* We can have multiple callers of `Foo.method`. So we are adding more casts in the binary, even if not at runtime.
* Indirect calls to `Foo.method` may exist, and those calls may target `Foo.method` or a set of other functions, so we cannot refine them in the same way. One way to handle this is to duplicate `Foo.method` and use a refined version for direct calls; I noticed a few % of code size overhead from that. Another way is to add a "thunk": make the indirect call go to a tiny function that casts and then calls the final one (and let inlining cause duplication, if that makes sense). With that I see much less than 1% code size overhead, but it does mean indirect calls may end up adding a direct call through a thunk, which may actually make them slower.

Inspecting Kotlin and J2Wasm code with this, it seems to work properly and I do see some benefits sometimes in the callers, where the new cast ends up helping. But that seems rare, unfortunately, and so given the code size and runtime downsides I am not sure if we want this or not.


kripken:
@tlively 

> I would have expected very many functions to have unconditional casts, so I would have expected that simply copying them to handle indirect functions calls would have had unacceptable code size overhead.

What do you mean by "unacceptable"? :smile:  I did see a few % of code size overhead from it, as mentioned, enough to make me not want to copy...

> Does this optimization not apply to as many functions as I would have expected?

I measured this on J2Wasm now, and I see 4,278 functions with cast parameters out of 31,722. So it's 13% - significant but not huge. I guess those functions are just on average smaller than most, so the copying overhead ends up in single digits (or maybe function merging kicks in..?).

13% of functions being class methods might just be the right number. Quick measurement: 10,210 functions (32%) have no parameters at all (!), just results, which means they definitely don't take a `this` parameter. (Perhaps they were class methods once, but `this` was not used and we removed it..?)

> If so, is that fundamental or due to limitations in the current implementation?

As for limitations:

* We could look past the first basic block, but I don't think that would help much (Java/Kotlin put the casts in the first block).
* We could also propagate casts backwards past the current basic block (which could help the results of this pass cause more benefits), but I'm not sure how much that would help - I'd guess not hugely.
* We could also look not at direct calls like this pass does, but at groups of functions by type, which could help with `call_ref` in principle, but I don't think that would help in practice: If we end up doing `call_ref` that means more than one class is possible, which means more than one cast.

tlively:
I wouldn't have been surprised if 50% of functions had been optimizable, so I would have expected copying to have double-digit percentage code size bloat, but I'm glad it doesn't. I'm very surprised that so many functions don't take any parameters. Do you know what kind of functions those are?

Do we expect this to have measurable improvements?

I suppose a more sophisticated interprocedural analysis might be able to use more precise heuristics to determine when this optimization will be profitable to avoid some of the code size overhead, too.

kripken:
> I'm very surprised that so many functions don't take any parameters. Do you know what kind of functions those are?

Judging from their names, many are stringpool helpers, or clinits, or random computational stuff like hashing.

> Do we expect this to have measurable improvements?

We could try to measure this on real-world code, but just looking at it, I'm doubtful. We're either growing code size significantly (copies) or adding runtime overhead (thunks), and the benefits just don't look obvious.

> I suppose a more sophisticated interprocedural analysis might be able to use more precise heuristics to determine when this optimization will be profitable to avoid some of the code size overhead, too.

Yes, something more complex like an extension to GUFA might be able to use this information better. I might look into that.

kripken:
This has been replaced by #5850 which does a similar analysis but inside GUFA. The benefits of doing it there are that

1. It is simpler to do a lot more related things, like ignoring indirect calls to functions that trap.
2. GUFA infers content in its own IR, and after doing all that it can decide how to apply that to the main IR. In this PR we added casts, but GUFA also has an option to only refine existing ones, which is more cautious and avoids the code size issues in this PR. And even when GUFA does that careful mode it still benefits from what it inferred internally, which can lead to other benefits.

PR #5813, opened 2023-07-13
[wasm-merge] Handle chains of import/export

vouillon:
When a module item is imported and directly reexported by an intermediate module, we need to perform several name lookups and use its name in the initial module rather than the intermediate name when fusing imports and exports.


src/tools/wasm-merge.cpp
```diff
@@ -235,6 +235,19 @@ void updateNames(Module& wasm, KindNameUpdates& kindNameUpdates) {
     }
 
   private:
+    Name& resolveName(NameUpdates updates, Name& name) {
```
kripken, line 238: Looks like this takes name by reference - is that to memoize to make later lookups faster?

This code operates in parallel, however, so it is unsafe to modify the data structure. In fact it should probably be declared as const.

kripken, line 238:
```suggestion
    Name& resolveName(NameUpdates& updates, Name& name) {
```
Without this change we copy that map each time, which can be slow.

kripken (commented):
Thanks!

Please add a test of a loop as well.

src/tools/wasm-merge.cpp
vouillon: It's just that I took model on function `mapName` which is taking a name by reference. (I now see it's because it updates the name in place.)
I'm not a C++ expert...

kripken (approved):

PR #5814, opened 2023-07-13
WIP hack to support i8/i16 mutable arrays as public types for string interop

kripken:


src/wasm/wasm-type.cpp
```diff
@@ -2523,6 +2523,27 @@ TypeBuilder::BuildResult TypeBuilder::build() {
...
+  auto array16 = Array(Field(Field::i16, Mutable));
+  TypeBuilder builder(4);
+  builder.setHeapType(0, array8);
+  builder.setHeapType(1, array16);
+  builder.setHeapType(2, array8);
+  // We handle final and non-final here, but should remove one of them
+  // eventually TODO
+  builder.setFinal(2);
+  builder.setHeapType(3, array16);
+  builder.setFinal(3);
```
tlively, line 2537:
```suggestion
  builder[0] = array8;
  builder[1] = array16;
  builder[2] = array8;
  // We handle final and non-final here, but should remove one of them
  // eventually TODO
  builder[2].setFinal();
  builder[3] = array16;
  builder[3].setFinal();
```

tlively (approved):
LGTM, and I would be ok landing this if it is helpful to do so and the fuzzer doesn't complain.

kripken:
This looks fine on the fuzzer, so we could land it, but waiting on confirmation that this will likely be useful. cc @jakobkummerow No rush to decide either way, we can keep this open for now, or land it, however is more convenient.

src/wasm/wasm-type.cpp
```diff
@@ -2551,7 +2551,7 @@ TypeBuilder::BuildResult TypeBuilder::build() {
 
   return {results};
 }
-
+
```
kripken, line 2554:
```suggestion

```

src/wasm/wasm-type.cpp
```diff
@@ -2551,7 +2551,6 @@ TypeBuilder::BuildResult TypeBuilder::build() {
 
   return {results};
 }
```
kripken, line 2553:
```suggestion
}

```

kripken:
This has proven useful and so I'll land it later today.

kripken:
cc @jakobkummerow

Issue #3989, opened 2021-07-14
Identical function types not merged


kripken:
STR: `wasm-opt w.wasm -all --print --name-types --print`

The output is
```wat
(module
 (type $none_=>_none (func))
 (global $global$0 (ref null $none_=>_none) (ref.func $0))
 (global $global$1 i32 (i32.const 0))
 (global $global$2 (mut externref) (ref.null extern))
 (global $global$3 (mut i32) (i32.const -6567))
 (global $global$4 (mut anyref) (ref.null any))
 (global $global$5 (mut i32) (i32.const 10))
 (memory $0 (shared 16 17))
 (table $0 16 funcref)
 (table $1 28 funcref)
 (table $2 18 18 funcref)
 (table $3 0 0 externref)
 (elem $0 (table $0) (i32.const 0) func)
 (elem $1 (table $1) (i32.const 0) func)
 (elem $2 (table $1) (i32.const 1) func)
 (elem $3 (table $2) (i32.const 0) func)
 (elem $4 (table $2) (i32.const 2) func)
 (elem $5 func)
 (elem $6 func)
 (elem $7 (ref null $none_=>_none) (ref.func $0) (ref.func $0) (ref.null $none_=>_none))
 (elem $8 func)
 (func $0
  (nop)
 )
)
(module
 (type $type$0 (func))
 (global $global$0 (ref null $type$0) (ref.func $0))
 (global $global$1 i32 (i32.const 0))
 (global $global$2 (mut externref) (ref.null extern))
 (global $global$3 (mut i32) (i32.const -6567))
 (global $global$4 (mut anyref) (ref.null any))
 (global $global$5 (mut i32) (i32.const 10))
 (memory $0 (shared 16 17))
 (table $0 16 funcref)
 (table $1 28 funcref)
 (table $2 18 18 funcref)
 (table $3 0 0 externref)
 (elem $0 (table $0) (i32.const 0) func)
 (elem $1 (table $1) (i32.const 0) func)
 (elem $2 (table $1) (i32.const 1) func)
 (elem $3 (table $2) (i32.const 0) func)
 (elem $4 (table $2) (i32.const 2) func)
 (elem $5 func)
 (elem $6 func)
 (elem $7 (ref null $type$0) (ref.func $0) (ref.func $0) (ref.null $none_=>_none))
 (elem $8 func)
 (func $0
  (nop)
 )
)
```
[...]

PR #5816, opened 2023-07-13
Simplify and consolidate type printing

tlively:
When printing Binaryen IR, we previously generated names for unnamed heap types
based on their structure. This was useful for seeing the structure of simple
types at a glance without having to separately go look up their definitions, but
it also had two problems:

 1. The same name could be generated for multiple types. The generated names did
 not take into account rec group structure or finality, so types that differed
 only in these properties would have the same name. Also, generated type names
 were limited in length, so very large types that shared only some structure
 could also end up with the same names. Using the same name for multiple types
 produces incorrect and unparsable output.

 2. The generated names were not useful beyond the most trivial examples. Even
 with length limits, names for nontrivial types were extremely long and visually
 noisy, which made reading disassembled real-world code more challenging.

Fix these problems by emitting simple indexed names for unnamed heap types
instead. This regresses readability for very simple examples, but the trade off
is worth it.

This change also reduces the number of type printing systems we have by one.
Previously we had the system in Print.cpp, but we had another, more general and
extensible system in wasm-type-printing.h and wasm-type.cpp as well. Remove the
old type printing system from Print.cpp and replace it with a much smaller use
of the new system. This requires significant refactoring of Print.cpp so that
PrintExpressionContents object now holds a reference to a parent
PrintSExpression object that holds the type name state.

This diff is very large because almost every test output changed slightly. To
minimize the diff and ease review, change the type printer in wasm-type.cpp to
behave the same as the old type printer in Print.cpp except for the differences
in name generation. These changes will be reverted in much smaller PRs in the
future to generally improve how types are printed.


kripken:
I don't understand the last paragraph, sorry - maybe a typo in there somewhere?

kripken:
From what I gather from the first paragraphs, it sounds like this might let us remove this?

https://github.com/WebAssembly/binaryen/blob/417792e90feb81b8b27863814cb0b38f5fe10b1f/scripts/fuzz_opt.py#L1374-L1375

and close https://github.com/WebAssembly/binaryen/issues/3989?

test/passes/generate-stack-ir_optimize-stack-ir_print-stack-ir_optimize-level=3.txt
```diff
@@ -20,15 +20,15 @@
  (func $big_negative
   (local $temp f64)
   f64.const -2147483648
-  local.set $temp
+  local.set $0
```
kripken, line 23: Why did local names change?

tlively:
> I don't understand the last paragraph, sorry - maybe a typo in there somewhere?

I rewrote the last paragraph now. Let me know if it could still be clearer.

test/passes/generate-stack-ir_optimize-stack-ir_print-stack-ir_optimize-level=3.txt
tlively: Hmmm, I'm not sure. That does seem fishy.

tlively:
I don't know what's up with #3989. It doesn't really look related since there is only one function type in the input, but it also looks like perhaps it won't be able to be reproduced since we changed nulls to use bottom types.

kripken:
Thanks, last paragraph sgtm now!

About #3989, yeah, maybe that is moot. But if so then the only thing blocking the `RoundtripText` fuzzer from being re-enabled is that we sometimes emitted duplicate type names, which I think this PR fixes? So hopefully we can enable it here.

tlively:
* main
  * **PR #5816** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5816" target="_blank">Graphite</a>  👈
    * **PR #5808** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5808" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5816?utm_source=stack-comment).

tlively:
This should be ready for another look. I piped more context through to the StackIR printer, so the problem where it was not printing the correct local names is fixed, and in fact it now prints the correct type names, which it did not do before.

kripken:
Wow, this huge diff is hard to see on github :smile: I might view it locally.

Can you please ping me when tests pass so I just have to do that once?

tlively:
@kripken, this should be ready to go! Hopefully we can get it in before we get any more merge conflicts 😬

kripken (approved):
lgtm!

PR #5817, opened 2023-07-13
Substitute LocalGraph initial calculations with Reaching Definitions Analysis

zm2he:
This change creates a reaching definitions analysis, and makes some modifications in the `MonotoneCFGAnalyzer` class to facilitate these changes. In particular, this analysis is intended to eventually perform the same work that `LocalGraph` is currently doing.

I have reverted the changes to `LocalGraph` for the purpose of this pull request, but will restore them in the branch `local-graph-static-analysis` for fuzzing.

### DIFF
--- a/src/analysis/monotone-analyzer-impl.h
+++ b/src/analysis/monotone-analyzer-impl.h
@@ -44,6 +44,13 @@ inline MonotoneCFGAnalyzer<Lattice, TransferFunction>::MonotoneCFGAnalyzer(
 }
 
 template<typename Lattice, typename TransferFunction>
+inline void
+MonotoneCFGAnalyzer<Lattice, TransferFunction>::evaluateFunctionEntry(
+  Function* func) {
+  transferFunction.evaluateFunctionEntry(func, stateBlocks[0].inputState);
+}
+
+template<typename Lattice, typename TransferFunction>
 inline void MonotoneCFGAnalyzer<Lattice, TransferFunction>::evaluate() {
   std::queue<const BasicBlock*> worklist;
 
@@ -71,6 +78,19 @@ inline void MonotoneCFGAnalyzer<Lattice, TransferFunction>::evaluate() {
         worklist.push(&dep);
       }
     }
+  }
+}
+
+template<typename Lattice, typename TransferFunction>
+inline void MonotoneCFGAnalyzer<Lattice, TransferFunction>::collectResults() {
+  for (BlockState currBlockState : stateBlocks) {
+    typename Lattice::Element inputStateCopy = currBlockState.inputState;
+
+    // The transfer function generates the final set of states and uses it to
+    // produce useful information. For example, in reaching definitions
+    // analysis, these final states are used to populate a mapping of
+    // local.get's to a set of local.set's that affect its value.
+    transferFunction.collectResults(currBlockState.cfgBlock, inputStateCopy);
   }
 }
 
--- a/src/analysis/reaching-definitions-transfer-function.h
+++ b/src/analysis/reaching-definitions-transfer-function.h
@@ -1,0 +1,178 @@
+#ifndef wasm_analysis_reaching_definitions_transfer_function_h
+#define wasm_analysis_reaching_definitions_transfer_function_h
+
+#include "ir/find_all.h"
+#include "ir/local-graph.h"
+#include "visitor-transfer-function.h"
+
+namespace wasm::analysis {
+
+// A state contains all the "live" LocalSet instances at some particular
+// point in a program. The state tracks every LocalSet instance, not just
+// ones which influence a particular local index.
+
+// In addition to LocalSet expressions modifying a local index's value, we
+// must also account for the fact that the value of a local index could be
+// the default value at initialization or its value passed in as a parameter.
+// As a result, we use a fictitious LocalSet object held by the transfer
+// function to represent the local index obtaining its current value on
+// initialization or parameter passing.
+
+// Each local index gets an individual fictitious LocalSet, because the state
+// tracks all local indices, and setting one local index should not affect the
+// initial value of another.
+
+// When collecting results, the transfer function takes the states and converts
+// it into a map of LocalGets to LocalSets which affect it. The fictitious
+// inital value LocalSetes will be converted to nullptrs.
+
+class ReachingDefinitionsTransferFunction
+  : public VisitorTransferFunc<ReachingDefinitionsTransferFunction,
+                               FinitePowersetLattice<LocalSet*>,
+                               AnalysisDirection::Forward> {
+
+  // Number of locals in the function.
+  size_t numLocals;
+
+  // Maps a local index to a list of LocalSets that modify the index's value.
+  // The most common case is to have a single set; after that, to be a phi of 2
+  // items, so we use a small set of size 2 to avoid allocations there.
+  std::unordered_map<Index, SmallVector<LocalSet*, 2>> indexSetses;
+
+  // LocalGraph members we need to update.
+  LocalGraph::GetSetses& getSetses;
+
+  // Fictitious LocalSet objects to reprsent a local index obtaining its value
+  // from its default initial value or parameter value.
+  std::vector<LocalSet> fakeInitialValueSetses;
+
+  // Pointers to the fictitious LocalSet objects.
+  std::unordered_set<LocalSet*> fakeSetPtrs;
+
+  // Helper function which creates fictitious LocalSetses for a function,
+  // inserts them into fakeInitialValueSetses and fakeSetPtrs. It returns a
+  // vector of actual LocalSetses in the function and fictitious LocalSetses for
+  // use when instatitating the lattice.
+  static std::vector<LocalSet*>
+  listLocalSetses(Function* func,
+                  std::vector<LocalSet>& fakeInitialValueSetses,
+                  std::unordered_set<LocalSet*>& fakeSetPtrs) {
+    // Create a fictitious LocalSet for each local index.
+    for (Index i = 0; i < func->getNumLocals(); ++i) {
+      LocalSet currSet;
+      currSet.index = i;
+      fakeInitialValueSetses.push_back(currSet);
+    }
+
+    // Find all actual LocalSetses.
+    FindAll<LocalSet> setFinder(func->body);
+    std::vector<LocalSet*> setsesList = std::move(setFinder.list);
+
+    // Take a pointer for each fictitious LocalSet.
+    for (size_t i = 0; i < fakeInitialValueSetses.size(); ++i) {
+      setsesList.push_back(&fakeInitialValueSetses[i]);
+      fakeSetPtrs.insert(&fakeInitialValueSetses[i]);
+    }
+    return setsesList;
+  }
+
+public:
+  // Analysis lattice. Public, as the monotone analyzer may need it.
+  FinitePowersetLattice<LocalSet*> lattice;
+
+  // We actually don't update LocalGraph::Locations right now since the CFG we
+  // are working with doesn't contain the correct Expression**s, but this is
+  // left in for future improvements. TODO.
+  ReachingDefinitionsTransferFunction(Function* func,
+                                      LocalGraph::GetSetses& getSetses,
+                                      LocalGraph::Locations& locations)
+    : numLocals(func->getNumLocals()), getSetses(getSetses),
+      lattice(listLocalSetses(func, fakeInitialValueSetses, fakeSetPtrs)) {
+
+    // Map every local index to a set of all the local sets which affect it.
+    for (auto it = lattice.membersBegin(); it != lattice.membersEnd(); ++it) {
+      indexSetses[(*it)->index].push_back(*it);
+    }
+  }
+
+  void visitLocalSet(LocalSet* curr) {
+    assert(currState);
+
+    // This is only needed to derive states when solving the analysis, and
+    // not for collecting results from the final states.
+
+    // For a LocalSet, we remove all previous setses modifying its index and
+    // adds itself.
+    auto& currSetses = indexSetses[curr->index];
+    for (auto setInstance : currSetses) {
+      lattice.remove(currState, setInstance);
+    }
+
+    lattice.add(currState, curr);
+  }
+
+  // Only for collecting results. For curr, collects all of the sets to curr's
+  // index which are present in the state (i.e. those that set index's current
+  // value).
+  void visitLocalGet(LocalGet* curr) {
+    assert(currState);
+
+    // This is only to be run in the second phase where we iterate over the
+    // final (i.e. solved) states. Thus, only done when collectingResults is
+    // true.
+    if (collectingResults) {
+      auto& matchingSetses = indexSetses[curr->index];
+
+      for (auto setInstance : matchingSetses) {
+        if (lattice.exists(currState, setInstance)) {
+          // If a pointer to a real LocalSet, add it, otherwise add a nullptr.
+          if (fakeSetPtrs.find(setInstance) == fakeSetPtrs.end()) {
+            getSetses[curr].insert(setInstance);
+          } else {
+            getSetses[curr].insert(nullptr);
+          }
+        }
+      }
+    }
+  }
+
+  // At the entry point of the function, we must set the state to signify that
+  // the values in each local index are from either a parameter value or their
+  // default initialized value. This cannot be done normally, because the
+  // initial state of the entry block is initialized as the bottom element, but
+  // no one can modify it because this block doesn't depend on any other. Hence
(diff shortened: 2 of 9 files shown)


tlively (commented):
For the tests, it would be good to test _some_ programmatic way of access ing the analysis contents, even if it's not a mock of `LocalGraph` directly, rather than doing string matching on printed output.

(The reason it was more ok to use string matching to test the CFG utility was that the string matching was a quick way to test the _structure_ of the CFG, i.e. it matched up well with the queryable interface of the CFG, and the printing function itself used the intended interface. A similar testing strategy here would be to print only the results of the analysis using a separate print function that uses the programmatic interface to the analysis.)

src/analysis/reaching-definitions-transfer-function.h
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  // Sets the LocalGraph members we want to update, and signals that result
+  // collection is about to begin.
+  void beginResultCollection(LocalGraph::GetSetses* getSetses,
+                             LocalGraph::Locations* locations) {
+    this->getSetses = getSetses;
+    this->locations = locations;
+    collectingResults = true;
+  }
```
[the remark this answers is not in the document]
zm2he: This isn't used in this PR because it is used in `LocalGraph` to collect results into the `LocalGraph` members. I will change the test case to collect results in the same manner as `LocalGraph` instead of via printing.

src/analysis/reaching-definitions-transfer-function.h
```diff
@@ -0,0 +1,176 @@
...
+  void
+  evaluateFunctionEntry(Function* func,
+                        FinitePowersetLattice<LocalSet*>::Element& inputState) {
+    for (size_t i = 0; i < numLocals; ++i) {
+      inputState.set(lattice.getSetSize() - numLocals + i, true);
+    }
+  }
+
+  void transfer(const BasicBlock* cfgBlock,
+                FinitePowersetLattice<LocalSet*>::Element& inputState) {
```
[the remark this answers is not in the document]
zm2he: In progress.

src/analysis/reaching-definitions-transfer-function.h
```diff
@@ -0,0 +1,176 @@
...
+namespace wasm::analysis {
+
+class ReachingDefinitionsTransferFunction
+  : public Visitor<ReachingDefinitionsTransferFunction> {
+  // In addition to LocalSet expressions modifying a local index's value, we
+  // must also account for the fact that the local index's default value at
+  // initialization is used, or that it already has a value at the beginning.
+  // Therefore, when creating the powerset lattice from the list of all
+  // LocalSetses, we also append a list of nullptrs  numLocals long, one for
+  // each local index, at the end of the list of all LocalSetses. These nullptrs
```
kripken, line 18: Do we need a nullptr for each local index? The existing LocalGraph representation uses `nullptr` for all of them, which seems more efficient, and still unambiguous.

kripken: Also, what is this comment "attached" to? It doesn't seem to be `currState` right below it, so if it is a class-level comment, please put it outside of the class.

src/analysis/reaching-definitions-transfer-function.h
```diff
@@ -0,0 +1,176 @@
...
+  // represent the use of a default/param value.
+
+  FinitePowersetLattice<LocalSet*>::Element* currState = nullptr;
+  FinitePowersetLattice<LocalSet*>& lattice;
+
+  // Number of locals in a function.
+  size_t numLocals;
+
+  // Maps a local index to a list of LocalSets that modify the index's value.
+  std::unordered_map<Index, SmallVector<LocalSet*, 2>> indexSetses;
```
kripken, line 28: Please explain the constant 2 here. The old code has this, for example:
```cpp
  // The local.sets relevant for an index or a get. The most common case is to
  // have a single set; after that, to be a phi of 2 items, so we use a small
  // set of size 2 to avoid allocations there.
```

src/analysis/reaching-definitions-transfer-function.h
```diff
@@ -0,0 +1,176 @@
...
+  // Signals to the visit functions whether we are solving the analysis, or
+  // collecting results from the solved analysis.
+  bool collectingResults = false;
+
+public:
+  ReachingDefinitionsTransferFunction(FinitePowersetLattice<LocalSet*>& lattice,
+                                      size_t numLocals)
+    : lattice(lattice), numLocals(numLocals) {
+
+    // Populate indexSetses.
```
kripken, line 47: What is being populated here? The comment could be a little more explicit.

src/analysis/reaching-definitions-transfer-function.h
```diff
@@ -0,0 +1,176 @@
...
+  // LocalGraph members we need to update.
+  LocalGraph::GetSetses* getSetses = nullptr;
+
+  // We actually don't update locations right now since the CFG we are working
+  // with doesn't contain the correct Expression**s, but this is left in for
+  // future improvements. TODO.
+  LocalGraph::Locations* locations = nullptr;
+
+  // Signals to the visit functions whether we are solving the analysis, or
+  // collecting results from the solved analysis.
```
kripken, line 39: So there are two distinct phases in which this entire class is run? That would be worth making explicit. This comment by itself only makes immediate sense if the reader is already aware of that fact (I myself was only aware of it because you mentioned it offline).

src/analysis/reaching-definitions-transfer-function.h
zm2he: A nullptr is needed for each local index because one local index could get set, while another is not, meaning that one local index still has the initial value, while the other one has a value from a local.set. If I remove the single nullptr, that would imply that both values cannot have the initial default value, which would be incorrect.

And yeas, this is meant more as a class-level comment, so I will move it.

src/analysis/reaching-definitions-transfer-function.h
zm2he: Added more comments to make it clear, both in here, the analyzer, and the crtp utility.

src/analysis/reaching-definitions-transfer-function.h
zm2he: Added some more comments to clarify this.

src/analysis/visitor-transfer-function.h
```diff
@@ -0,0 +1,108 @@
...
+#include "cfg.h"
+#include "lattice.h"
+#include "wasm-traversal.h"
+
+namespace wasm::analysis {
+
+// Utility for visitor-based transfer functions for forward and backward
+// analysis. Forward analysis is chosen by default unless the template parameter
+// Backward is true.
+template<typename SubType, typename Lattice, bool Backward = false>
```
tlively, line 15: It would be good to use an enum rather than a bool for the analysis direction.

src/analysis/visitor-transfer-function.h
```diff
@@ -0,0 +1,108 @@
...
+
+public:
+  // Returns an iterable to all the BasicBlocks which depend on currBlock for
+  // information.
+  BasicBlock::BasicBlockIterable getDependents(const BasicBlock* currBlock) {
+    if constexpr (Backward)
+      return currBlock->preds();
+    else
+      return currBlock->succs();
+  }
```
tlively, line 41: This still needs to follow our braces style!
```suggestion
    if constexpr (Backward) {
      return currBlock->preds();
    } else {
      return currBlock->succs();
    }
  }
```

test/gtest/cfg.cpp
```diff
@@ -290,3 +292,189 @@ TEST_F(CFGTest, FinitePowersetLatticeFunctioning) {
...
+  LocalGraph::Locations locations;
+  FinitePowersetLattice<LocalSet*> lattice =
+    ReachingDefinitionsTransferFunction::getLatticeForFunc(func);
+  ReachingDefinitionsTransferFunction transferFunction(
+    lattice, func->getNumLocals(), getSetses, locations);
+
+  MonotoneCFGAnalyzer<FinitePowersetLattice<LocalSet*>,
+                      ReachingDefinitionsTransferFunction>
+    analyzer(lattice, transferFunction, cfg);
+  analyzer.evaluateFunctionEntry(func);
```
tlively, line 341: Probably worth a TODO about making this more automatic.

test/gtest/cfg.cpp
```diff
@@ -290,3 +292,189 @@ TEST_F(CFGTest, FinitePowersetLatticeFunctioning) {
...
+  analyzer.evaluateAndCollectResults();
+
+  FindAll<LocalSet> foundSets(func->body);
+  FindAll<LocalGet> foundGets(func->body);
+  LocalGraph::GetSetses expectedResult;
+  expectedResult[foundGets.list[0]].insert(foundSets.list[0]);
+  expectedResult[foundGets.list[1]].insert(foundSets.list[0]);
+  expectedResult[foundGets.list[2]].insert(nullptr);
```
tlively, line 349: It might be nice to use a list initializer for `expectedResult`, e.g.

```
LocalGraph::getSetses expectedResult = {{foundGets.list[0], {foundSets.list[0]}}, ...};
```

Further, it might be nice to give all the different gets and sets meaningful names (e.g. `getA`) to make it easier to match them up with the input program.

src/analysis/reaching-definitions-transfer-function.h
```diff
@@ -0,0 +1,167 @@
...
+    // adds itself.
+    auto& currSetses = indexSetses[curr->index];
+    for (auto setInstance : currSetses) {
+      lattice.remove(currState, setInstance);
+    }
+
+    // Removing all previous setses means that the default initial/parameter
+    // value is overwritten, so we need to remove this possibility from the
+    // state too.
+    currState->set(lattice.getSetSize() - numLocals + curr->index, false);
```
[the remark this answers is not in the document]
tlively: Ah, I don't think @kripken meant that we should actually insert the new LocalSet pointers into the CFG data structure. They're only "logically" associated with the entry block of the CFG. Keeping the fictitious pointers in the transfer function object SGTM.

test/gtest/cfg.cpp
zm2he: List initializer doesn't seem to be supported for the `SmallSet`, but I will use more meaningful names.

src/analysis/monotone-analyzer-impl.h
```diff
@@ -86,11 +86,11 @@ inline void MonotoneCFGAnalyzer<Lattice, TransferFunction>::collectResults() {
   for (BlockState currBlockState : stateBlocks) {
     typename Lattice::Element inputStateCopy = currBlockState.inputState;
 
-    // Calls collectResults from transfer function. In this method, the transfer
+    // The transfer
```
tlively, line 89: Please reformat the comment after this deletion.

src/analysis/visitor-transfer-function.h
```diff
@@ -9,10 +9,12 @@
 
 namespace wasm::analysis {
 
+enum TransferFuncDirection { FORWARD, BACKWARD };
```
tlively, line 12: It's good that you're already referring to these as `TransferFuncDirection::FORWARD` instead of just `FORWARD`. You can enforce that by making this an `enum class` instead of just an `enum`. 

For the naming, I think `AnalysisDirection` is more precise than `TransferFuncDirection`, and using PascalCase rather than all caps is more in line with other enums in Binaryen.


```suggestion
enum class AnalysisDirection { Forward, Backward };
```

tlively (approved):
Nice! LGTM % small remaining comments.

src/analysis/reaching-definitions-transfer-function.h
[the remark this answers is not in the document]
zm2he: Leaving this to a further PR.

src/analysis/reaching-definitions-transfer-function.h
```diff
@@ -0,0 +1,165 @@
...
+// In addition to LocalSet expressions modifying a local index's value, we
+// must also account for the fact that the value of a local index could be
+// the default value at initialization or its value passed in as a parameter.
+// As a result, we use a nullptr to represent this case for each local index,
+// as a "fictitious" set which provides the initial value of the index.
+
+// Each local index gets a separate nullptr, because the state tracks all
+// local indices, and setting one local index should not affect the initial
+// vlaue of another. Therefore, when creating the powerset lattice from the
+//  list of all LocalSetses, we also append a list of nullptrs numLocals long
```
kripken, line 23: Looks like some extra space at the start of this line and the next.

src/analysis/reaching-definitions-transfer-function.h
```diff
@@ -0,0 +1,165 @@
...
+//  at the end of the list. The ith local index is at:
+// lattice.setSize() - numLocals + i
```
kripken, line 25: Did the idea of using N "fake" local.sets instead of N nullptrs not work out?

zm2he: Added now.

zm2he:
Note about using "fake" `local.set`s. 

Using fictitious `local.set`s to represent the origin of the initial value of a local index in a function is convenient, as it allows them to be treated like any other `local.set` in the transfer function. The alternative was using `nullptr`s differentiated by position in the `FinitePowersetLattice`'s. This required special case logic with hard to understand index calculations. It also relied on the lattice implementation providing a specific ordering between lattice element members and the nullptrs being all appended at the end. This made ti less resilient to changes in lattice implementation.

The fictitious `local.set`s could be stored as members of the lattice. This was better in terms of separation of concerns, as it would put all the data relevant for solving the analysis in the lattice, while keeping only data mutation functionality in the transfer function. The main disadvantage would be that an entirely new lattice had to be created instead of relying on the existing `FinitePowersetLattice`, with new functions to generate fictitious `local.set`s and check if a pointer was a fictitious `local.set` or a real one. It would also mean that the new lattice would have many layers of composition.

Alternatively, they could be stored as members of the transfer function. This would require the lattice to be a member of the transfer function. The advantage of this was that we could use the existing `FinitePowersetLattice`. All the reaching-definitions specific logic would be confined to the transfer function, which is better in terms of maintainability. It also better hides the implementation detail that we use these fictitious `local.set`s to provide the necessary information to `LocalGraph`  from other analysis components, which is better. Disadvantages are that the transfer function is more tightly coupled to the lattice and that more complicated code is needed to initialize the lattice. However, as the transfer function by nature is already pretty tightly coupled, the first point is not too important.

A further alternative would be to insert these fictitious pointers in a block in the CFG. However, this has a lot of complexity which would make it too large for this PR, and furthermore would make the analysis CFG even more difficult than the regular IR CFG.

zm2he:
Note about `BasicBlock` iterable types.

This PR also unifies the `BasicBlock` iterable types `Predecessors` and `Successors` into a `BasicBlockIterable` iterable type. This is useful as it allows predecessors and successors to be treated interchangeably and generically, which is what we want for a generic analysis framework where we want to return any arbitrary iterable when getting dependent blocks.

Retaining two separate iterable types makes it difficult to write functions returning them, as it requires writing multiple overloaded functions with `enable_if` which will cause errors with static asserts checking the transfer function/lattice APIs as it sees them as two functions instead of a versions of the same functions, only one of which is ever used.

src/analysis/reaching-definitions-transfer-function.h
```diff
@@ -0,0 +1,178 @@
...
+  // The most common case is to have a single set; after that, to be a phi of 2
+  // items, so we use a small set of size 2 to avoid allocations there.
+  std::unordered_map<Index, SmallVector<LocalSet*, 2>> indexSetses;
+
+  // LocalGraph members we need to update.
+  LocalGraph::GetSetses& getSetses;
+
+  // Fictitious LocalSet objects to reprsent a local index obtaining its value
+  // from its default initial value or parameter value.
+  std::vector<LocalSet> fakeInitialValueSetses;
```
kripken, line 47: The suffix "es" in "Setses" is a (silly) way to indicate "a group of sets". That is, we go from one item to a set of items to a group of sets of items. For lack of a better term, we've been using "setses" for that in LocalGraph.

Specifically, `GetSetses getSetses` there means: each get has a set of sets for it, and this is the map of each of those get+set of sets items. Or, to put it another way, "getSets" sounds like "a map of gets to sets", but we want a map of gets to sets of sets, so we need to be even more plural.

This line, however, looks like a normal vector of sets, so it could be `fakeInitialValueSets`? Or am I missing something?

zm2he: No I misunderstood `Setses` as being a more unambiguous way to distinguish a local.set from a set (i.e. group) of items. I'll change it into `fakeInitialValueSets`.

PR #5820, opened 2023-07-17
wasm-merge: Error on import loops

kripken:
Please take a look @vouillon - are there cases where we do not want to error? If there are then this PR might not make sense. But I can't think of one atm.


vouillon:
No, I don't see any case either. It's better to error, indeed!

kripken:
Sounds good, thanks! I'll land this then.

PR #5821, opened 2023-07-17
[Outlining] LLVM Suffix Tree

ashleynh:
This PR adds LLVM's suffix tree data structure to Binaryen. This suffix tree is implemented using Ukkonen's algorithm for linear-time suffix tree construction, and is intended for fast substring queries.

Note: All of the .h and .cpp files included are from LLVM. These files were copied directly instead of imported into our existing LLVM integration (in third_party/) to avoid bumping the commit hash and avoid the potential for complications with upstream changes.

### DIFF
--- a/src/support/CMakeLists.txt
+++ b/src/support/CMakeLists.txt
@@ -10,6 +10,8 @@ set(support_SOURCES
   istring.cpp
   path.cpp
   safe_integer.cpp
+  suffix_tree.cpp
+  suffix_tree_node.cpp
   threads.cpp
   utilities.cpp
   ${support_HEADERS}
--- a/src/support/CMakeLists.txt
+++ b/src/support/CMakeLists.txt
@@ -10,10 +10,19 @@ set(support_SOURCES
   istring.cpp
   path.cpp
   safe_integer.cpp
-  suffix_tree.cpp
-  suffix_tree_node.cpp
   threads.cpp
   utilities.cpp
   ${support_HEADERS}
 )
+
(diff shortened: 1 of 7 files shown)


src/support/CMakeLists.txt
```diff
@@ -14,4 +14,15 @@ set(support_SOURCES
...
+#[[ The below condition is intended for removal once the suffix_tree and suffix_tree_node source files no longer depend on LLVM code in the third_party folder ]]
+if(EMSCRIPTEN)
 add_library(support OBJECT ${support_SOURCES})
+else()
+set(support_with_suffix_tree_SOURCES
+  suffix_tree.cpp
+  suffix_tree_node.cpp
+  ${support_SOURCES}
+)
+add_library(support OBJECT ${support_with_suffix_tree_SOURCES})
```
tlively, line 27: I _think_ that we usually have two spaces of indent for code inside the if-else.

src/support/CMakeLists.txt
```diff
@@ -14,4 +14,15 @@ set(support_SOURCES
   utilities.cpp
   ${support_HEADERS}
 )
+
+#[[ The below condition is intended for removal once the suffix_tree and suffix_tree_node source files no longer depend on LLVM code in the third_party folder ]]
```
tlively, line 18: Wrap to 80 chars, please.

tlively (approved):
LGTM % making the CMake beautiful, which it turns out I have opinions about after all.

src/support/CMakeLists.txt
ashleynh: done

src/support/CMakeLists.txt
ashleynh: done

PR #5822, opened 2023-07-18
Static Analysis: Add an API to get the block index of an expression

kripken:
I'd like to use this in the reverse-direction analysis for TrapsNeverHappen: I need to know
which expressions are in the same basic block.

### DIFF
--- a/src/analysis/cfg.h
+++ b/src/analysis/cfg.h
@@ -89,7 +89,7 @@ struct CFGBlockIndexes {
   CFGBlockIndexes(const CFG& cfg);
 
   // Gets the index of the basic block in which the instruction resides.
-  Index get(Expression* expr) {
+  Index get(Expression* expr) const {
     auto iter = map.find(expr);
     if (iter == map.end()) {
       // There is no entry for this, which can be the case for control flow
--- a/src/analysis/cfg.h
+++ b/src/analysis/cfg.h
@@ -26,7 +26,6 @@
 #include <iostream>
 #include <vector>
 
-#include "ir/properties.h"
 #include "wasm.h"
 
 namespace wasm::analysis {
@@ -88,8 +87,7 @@ struct CFG {
     auto iter = expressionBlockIndexMap.find(expr);
     if (iter == expressionBlockIndexMap.end()) {
       // There is no entry for this, which can be the case for control flow
-      // structures.
-      assert(Properties::isControlFlowStructure(expr));
+      // structures, or for unreachable code.
       return InvalidBlock;
     }
     return iter->second;
--- a/src/analysis/cfg.h
+++ b/src/analysis/cfg.h
@@ -26,6 +26,7 @@
 #include <iostream>
 #include <vector>
 
+#include "ir/properties.h"
 #include "wasm.h"
 
 namespace wasm::analysis {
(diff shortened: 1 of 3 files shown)


src/analysis/cfg.h
```diff
@@ -77,8 +79,29 @@ struct CFG {
 
   void print(std::ostream& os, Module* wasm = nullptr) const;
 
+  // This must be done before getBlockIndex is called on an expression.
+  void computeExpressionBlockIndexes();
+
+  // Gets the index of the basic block in which the instruction resides.
+  // computeExpressionBlockIndexes() must be called first to populate the map.
+  Index getBlockIndex(Expression* expr) {
```
tlively, line 87: Can we automatically call `computeExpressionBlockIndexes` here to make the API less stateful? Then we could make `computeExpressionBlockIndexes` private as well.

test/gtest/cfg.cpp
```diff
@@ -290,3 +290,43 @@ TEST_F(CFGTest, FinitePowersetLatticeFunctioning) {
...
+
+  Module wasm;
+  parseWast(wasm, moduleText);
+
+  auto* func = wasm.getFunction("foo");
+  CFG cfg = CFG::fromFunction(func);
+  cfg.computeExpressionBlockIndexes();
+
+  // The body of the function is an if. An if is a control flow structure and so
+  // it has no basic block (it can contain multiple ones).
```
tlively, line 321: For the analysis framework, we'll eventually want to do control-sensitive analyses and be able to quickly tell what conditions must be true given that we take a certain control flow edge. Knowing that we executed an `if` as opposed to something else will be important in that context. Also, `if` consumes a value from the stack and that will be important to model once we do analyses that track more than just values in locals. Will it cause problems if we update the CFG utility to include `if`s in basic blocks in the future?

src/analysis/cfg.h
kripken: We could, but always doing this computation would be significant extra work that is not needed in some (most?) cases. Personally I don't think requiring explicit opt-in to this feature makes the API noticeably worse.

test/gtest/cfg.cpp
kripken: It wouldn't be a problem in anything I current have planned. Yes, I can see how we'd want to connect ifs to basic blocks in some manner (and the same for try, I think).

src/analysis/cfg.h
tlively: If we defer the computation until the first call to `getBlockIndex`, then we would only be doing the work when it is needed.

kripken: That's an option. It does make the function no longer const, though, which is odd.

kripken: Another option could be to add a friend class, `BlockIndexer` which would operate on a `CFG`. What do you think of that?

tlively: > That's an option. It does make the function no longer const, though, which is odd.

This can be solved by making the index map `mutable`, but that's a gross hack.

> Another option could be to add a friend class, `BlockIndexer` which would operate on a `CFG`. What do you think of that?

This is my favorite option so far. I like the separation of concerns.

kripken: Yeah, I guess even friendship isn't needed there. Done, PTAL.

tlively (approved):
LGTM!

PR #5823, opened 2023-07-18
End the current basic block on a Call

kripken:
Before this PR, if a call had no paths to a catch in the same function then we skipped
creating a new basic block right after it. As a result, we could have a call in the middle
of a basic block.

That is ok, but I think it might be good to have the property that basic blocks end in
possibly-control-flow-transferring instructions. That is, anything that might branch -
even if just out of the whole function - ends a basic block, which is what this PR changes. 

This change makes a future PR simpler, as it wants to depend on that assumption.

I added a test in the analysis CFG file as that seems simplest: we can print out the
CFG there to verify what we want.

### DIFF
--- a/src/cfg/cfg-traversal.h
+++ b/src/cfg/cfg-traversal.h
@@ -295,12 +295,6 @@ struct CFGWalker : public ControlFlowWalker<SubType, VisitorType> {
   }
 
   static void doEndCall(SubType* self, Expression** currp) {
-    auto* module = self->getModule();
-    if (module && !module->features.hasExceptionHandling()) {
-      // EH is disabled, so there cannot be a branch here due to a throw.
-      return;
-    }
-
     doEndThrowingInst(self, currp);
     // Create a new basic block and link to it. We do this even if there are no
     // other edges leaving this call (no catch bodies in this function that we
@@ -402,7 +396,11 @@ struct CFGWalker : public ControlFlowWalker<SubType, VisitorType> {
       case Expression::Id::CallId:
       case Expression::Id::CallIndirectId:
       case Expression::Id::CallRefId: {
-        self->pushTask(SubType::doEndCall, currp);
+        auto* module = self->getModule();
+        if (module && module->features.hasExceptionHandling()) {
+          // This call might throw, so run the code to handle that.
+          self->pushTask(SubType::doEndCall, currp);
+        }
         break;
       }
       case Expression::Id::TryId: {
--- a/src/cfg/cfg-traversal.h
+++ b/src/cfg/cfg-traversal.h
@@ -297,9 +297,10 @@ struct CFGWalker : public ControlFlowWalker<SubType, VisitorType> {
   static void doEndCall(SubType* self, Expression** currp) {
     doEndThrowingInst(self, currp);
     // Create a new basic block and link to it. We do this even if there are no
-    // other edges leaving this call (if a throw would go entirely outside of
-    // the function), because we want to preserve the property that a basic
-    // block ends with an instruction that might branch.
+    // other edges leaving this call (no catch bodies we can reach if we throw),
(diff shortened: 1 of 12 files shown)


src/cfg/cfg-traversal.h
```diff
@@ -296,11 +296,13 @@ struct CFGWalker : public ControlFlowWalker<SubType, VisitorType> {
...
-      auto* last = self->currBasicBlock;
-      self->link(last, self->startBasicBlock());
-    }
+    // Create a new basic block and link to it. We do this even if there are no
+    // other edges leaving this call (no catch bodies in this function that we
+    // can reach if we throw), because we want to preserve the property that a
+    // basic block ends with an instruction that might branch, and the call
+    // might branch out of the entire function if it throws.
+    auto* last = self->currBasicBlock;
+    self->link(last, self->startBasicBlock());
```
aheejin, line 305: Hmm, I think I decided not to do this because it would create too many edges. What changed, and what's the next PR this helps?

And maybe we can at least check if EH is available?

kripken: Good point about checking for EH, that should help... I should also measure the overhead.

I think what changed is the new pass I am working on :smile: but I think it's a general principle that makes sense, that control flow can't move in the middle of a block (except for traps). LLVM etc. do that, I think?

The new pass is an analysis for TrapsNeverHappen that goes backwards. Imagine a function always casts the input parameter to type B, even though it is declared as a supertype A. That is common in Java etc., where class methods declare `this` to have a very generic type like `java.lang.Object`, while methods of a class `B` will immediately cast to that type. If we see a static call to such a function, then we can infer in the caller that the reference must have that type (if we assume traps never happen).

And that pass needs to take into account transfers of control flow:
```wat
(call $will-cast-to-B
  (block
    (local.get $ref) ;; we can't infer this far back
    (call $maybe-throw)
    (local.get $ref) ;; we can infer about this
  )
)

(func $will-cast-to-B (param $ref anyref)
  (ref.cast $B (local.get $ref))
```
Imagine that `maybe-throw` will throw if that local doesn't have the right type `B`. If so then we won't actually do the call, which means we can't infer the first get has type `B` (only the second). That is, we can only make this inference when the call will definitely execute from that point.

And I am checking if the call happens by looking just at the same basic block as it. That is, when I see a parameter that is in another basic block, I stop going backwards, as control flow might go elsewhere. For that, I need to know that a call can transfer control flow.

aheejin: Ah, makes sense. Thanks for explaining.

aheejin (approved):

tlively (approved):
LGTM! I think this is a step toward what we will need when we start constructing interprocedural CFGs for interprocedural analysis in the new framework.

kripken:
Looking into the speed, there is a very small change (1%) to execution time in my measurements, but there is a 10% increase in instruction and branch counts. I guess the time remains the same because of cache effects. So this does show there is a cost to adding these extra blocks, and we were right to worry about it before.

I pushed a commit now to not add the blocks when EH is disabled, like you suggested @aheejin . With that, when the feature is disabled, there is no regression. This should help most users at least, but we are paying some cost here.

That last commit adds some NFC code to provide the module to the CFG walker in more code paths, so that we can use the features from it.

kripken:
(That 10% cost is when measuring just the passes that use CFGWalker. Those are subset of the total passes, so the total cost is lower, of course. I measured with this:
```
bin/wasm-opt --precompute-propagate --optimize-added-constants
  --merge-locals --coalesce-locals --once-reduction --rse --ssa-nomerge
  --precompute-propagate --optimize-added-constants --merge-locals
  --coalesce-locals --once-reduction --rse --low-memory-unused
```

)

kripken:
Ok, we now have a plan to fix the slowdown this causes, in #5838, so this is ready to land.

PR #5825, opened 2023-07-19
SmallVector iteration improvements

kripken:
I needed these to experiment with replacing the vector in CFGWalker. However, I saw
no benefit to doing so, so I won't be landing that. But these iterator improvements
may help in the future (and I had to fight the C++ compiler quite a bit for them...).

### DIFF
--- a/src/support/small_vector.h
+++ b/src/support/small_vector.h
@@ -170,8 +170,9 @@ public:
       return *this;
     }
     Iterator operator++(int) {
+      Iterator self = *static_cast<Iterator*>(this);
       index++;
-      return *this;
+      return self;
     }
 
     Iterator& operator+=(difference_type off) {
--- a/src/support/small_vector.h
+++ b/src/support/small_vector.h
@@ -148,9 +148,11 @@ public:
   // iteration
 
   template<typename Parent, typename Iterator> struct IteratorBase {
+    using iterator_category = std::random_access_iterator_tag;
     using value_type = T;
-    using difference_type = long;
+    using difference_type = off_t;
     using reference = T&;
+    using pointer = T*;
 
     Parent* parent;
     size_t index;
@@ -162,6 +164,7 @@ public:
     }
 
     void operator++() { index++; }
+    void operator++(int) { index++; }
 
     Iterator& operator+=(difference_type off) {
       index += off;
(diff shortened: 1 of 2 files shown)


src/support/small_vector.h
```diff
@@ -162,6 +164,7 @@ template<typename T, size_t N> class SmallVector {
     }
 
     void operator++() { index++; }
+    void operator++(int) { index++; }
```
tlively, line 167: These operators need to return `Iterator&` and `Iterator`. See https://github.com/WebAssembly/binaryen/blob/main/src/support/parent_index_iterator.h#L58-L75 for reference.

src/support/small_vector.h
```diff
@@ -148,9 +148,11 @@ template<typename T, size_t N> class SmallVector {
   // iteration
 
   template<typename Parent, typename Iterator> struct IteratorBase {
+    using iterator_category = std::random_access_iterator_tag;
```
tlively, line 151: It looks like there are still a bunch of operations missing to make this a valid random access iterator. See https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator for the docs.

kripken: Thanks, I added a TODO for this.

src/support/small_vector.h
kripken: Added.

src/support/small_vector.h
```diff
@@ -161,7 +165,14 @@ template<typename T, size_t N> class SmallVector {
...
 
-    void operator++() { index++; }
+    Iterator& operator++() {
+      index++;
+      return *this;
+    }
+    Iterator operator++(int) {
+      index++;
+      return *this;
+    }
```
tlively, line 175: We should at least make this a correct postincrement implementation.
```suggestion
    Iterator operator++(int) {
      Iterator self = *static_cast<Iterator*>(this);
      index++;
      return self;
    }
```

tlively (approved):
I would be more comfortable if we implemented the missing functionality, but also doing that lazily once the compiler complains doesn't seem like the end of the world.

nunotexbsd:
@kripken 

Could you please take a look at https://github.com/WebAssembly/binaryen/issues/5928 ?
Thanks

PR #5827, opened 2023-07-20
[NFC] Avoid using ControlFlowWalker in CFGWalker

kripken:
ControlFlowWalker builds a stack of control flow items and allows finding
the target of a branch using it. But the only use CFGWalker made of that was
to map a block or a loop to its branches. We can just use the name of the block
or loop in the map, and avoid the extra overhead.

`perf stat` confirms fewer instructions and branches with this change, but
the difference in runtime is too small to measure.


tlively (approved):
Nice!

PR #5828, opened 2023-07-20
Add support for debug printing of functions

kripken:
cc @zm2he

### DIFF
--- a/test/gtest/printing.cpp
+++ b/test/gtest/printing.cpp
@@ -20,7 +20,7 @@ TEST_F(PrintingTest, Print) {
     )
   )wasm";
 
-  auto stringifyText = R"stringify(
+  auto printText = R"print(
 (func $a (result i32)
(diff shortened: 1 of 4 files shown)


test/gtest/printing.cpp
```diff
@@ -0,0 +1,47 @@
...
+  Module wasm;
+  parseWast(wasm, moduleText);
+
+  std::stringstream ss;
+  ss << '\n';
+  ss << *wasm.getFunction("a") << '\n';
+  ss << "And the other function:\n\n";
+  ss << *wasm.getFunction("b");
+
+  EXPECT_EQ(ss.str(), printText);
```
tlively, line 46: I have a slight preference for having two `EXPECT_EQ` statements, one for each function, with nothing else printed, since we don't need to test passing normal strings to stringstream.

tlively (approved):

PR #5830, opened 2023-07-21
SimplifyLocals: Refinalize after removing redundant tees

kripken:
A minor old bug that the fuzzer found just now. Removing a tee in the middle
can place a more refined type in a location, which requires updating parents.


tlively (approved):
Do you think it's worth refinalizing only if the type actually changes?

kripken:
Good idea, done.

PR #5831, opened 2023-07-21
Add a Fuzzer for Lattice and Transfer Function Properties

zm2he:
This change adds a fuzzer which checks the following properties:

- Transfer Function Monotonicity
- Lattice Element Reflexivity
- Lattice Element Transitivity
- Lattice Element Anti-Symmetry

This is accomplished by randomly generating a module and using its functions as transfer function inputs, along with randomly generated lattice element states. Lattice element properties are fuzzed from the randomly generated states also.

### DIFF
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -5,16 +5,11 @@
 #include <unordered_map>
 #include <vector>
 
-#include "tools/fuzzing/random.h"
 #include "wasm.h"
 
 namespace wasm::analysis {
 
 enum LatticeComparison { NO_RELATION, EQUAL, LESS, GREATER };
-
-std::string LatticeComparisonNames[4] = {
-  "No Relation", "Equal", "Less", "Greater"};
-std::string LatticeComparisonSymbols[4] = {"?", "=", "<", ">"};
 
 // If parameter "comparison" compares x and y, the function returns the opposite
 // direction comparison between y and x.
@@ -119,9 +114,6 @@ public:
 
   // Returns an instance of the bottom lattice element.
   Element getBottom();
-
-  // Returns a random lattice element. Used for fuzzing.
-  Element getRandom(Random& rand);
 };
 
 // A layer of abstraction over FiniteIntPowersetLattice which maps
@@ -181,9 +173,6 @@ public:
   }
 
   Element getBottom() { return intLattice.getBottom(); }
-
-  // Returns a random lattice element. Used for fuzzing.
-  Element getRandom(Random& rand) { return intLattice.getRandom(rand); }
 };
 
 } // namespace wasm::analysis
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -104,6 +104,7 @@ public:
   // Returns an instance of the bottom lattice element.
   Element getBottom();
 
+  // Returns a random lattice element. Used for fuzzing.
   Element getRandom(Random& rand);
 };
 
@@ -165,6 +166,7 @@ public:
 
   Element getBottom() { return intLattice.getBottom(); }
 
+  // Returns a random lattice element. Used for fuzzing.
   Element getRandom(Random& rand) { return intLattice.getRandom(rand); }
 };
 
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -12,6 +12,22 @@ namespace wasm::analysis {
 
 enum LatticeComparison { NO_RELATION, EQUAL, LESS, GREATER };
 
+std::string LatticeComparisonNames[4] = {
+  "No Relation", "Equal", "Less", "Greater"};
+std::string LatticeComparisonSymbols[4] = {"?", "=", "<", ">"};
+
+// If parameter "comparison" compares x and y, the function returns the opposite
+// direction comparison between y and x.
+inline LatticeComparison reverseComparison(LatticeComparison comparison) {
+  if (comparison == LatticeComparison::LESS) {
+    return LatticeComparison::GREATER;
+  } else if (comparison == LatticeComparison::GREATER) {
+    return LatticeComparison::LESS;
+  } else {
+    return comparison;
+  }
+}
+
 template<typename Lattice>
 constexpr bool has_getBottom =
   std::is_invocable_r<typename Lattice::Element,
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -5,6 +5,7 @@
 #include <unordered_map>
 #include <vector>
 
+#include "tools/fuzzing/random.h"
 #include "wasm.h"
 
 namespace wasm::analysis {
@@ -102,6 +103,8 @@ public:
 
   // Returns an instance of the bottom lattice element.
   Element getBottom();
+
+  Element getRandom(Random& rand);
 };
 
 // A layer of abstraction over FiniteIntPowersetLattice which maps
--- a/src/analysis/lattice.h
+++ b/src/analysis/lattice.h
@@ -164,6 +164,8 @@ public:
   }
 
   Element getBottom() { return intLattice.getBottom(); }
+
+  Element getRandom(Random& rand) { return intLattice.getRandom(rand); }
 };
 
 } // namespace wasm::analysis
--- a/src/tools/wasm-fuzz-lattices.cpp
+++ b/src/tools/wasm-fuzz-lattices.cpp
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2021 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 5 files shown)


src/tools/wasm-fuzz-lattices.cpp
```diff
@@ -0,0 +1,389 @@
...
+  // states for a CFG block on which the transfer function is applied.
+  template<typename Lattice, typename TransferFunction>
+  void check(analysis::CFG& cfg,
+             Lattice& lattice,
+             TransferFunction& transferFunction,
+             std::string transferFuncName) {
+    for (auto cfgIter = cfg.begin(); cfgIter != cfg.end(); ++cfgIter) {
+      typename Lattice::Element x = lattice.getRandom(rand);
+      typename Lattice::Element y = lattice.getRandom(rand);
+      typename Lattice::Element z = lattice.getRandom(rand);
```
tlively, line 252: It might make sense to move everything that doesn't depend on `cfgIter` out of the loop to reduce the amount of work done for a single seed and make debugging easier. Otherwise, if one of these checks fails, you need to know both the seed and how many times this loop has executed to step into the failing check. You could even go as far as to separate `check` out into `checkLattice` and `checkTransferFunction`.

src/tools/wasm-fuzz-lattices.cpp
```diff
@@ -0,0 +1,389 @@
...
+  bool verbose;
+
+  Random rand;
+
+  Fuzzer(bool verbose) : verbose(verbose), rand({}) {}
+
+  // Checks reflexivity of a lattice element, i.e. x = x.
+  template<typename Lattice>
+  void checkReflexivity(Lattice& lattice, typename Lattice::Element& element) {
+    analysis::LatticeComparison result = lattice.compare(element, element);
```
tlively, line 33: It would be fine to stick a `using analysis;` at the beginning of this file so you don't need to rewrite the namespace everywhere.

src/tools/wasm-fuzz-lattices.cpp
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    std::vector<char> bytes2 = bytes;
+
+    Module testModule;
+    TranslateToFuzzReader reader(testModule, std::move(bytes));
+    reader.build();
+    rand = Random(std::move(bytes2));
+
+    for (auto& func : testModule.functions) {
```
tlively, line 330: You can use `ModuleUtils::iterDefinedFunctions` to automatically skip imported functions if you want.

src/tools/wasm-fuzz-lattices.cpp
```diff
@@ -0,0 +1,389 @@
...
+      Function* currFunc = func.get();
+      // Imported functions can't be fuzzed here as they don't have a body.
+      if (currFunc->imported()) {
+        continue;
+      }
+      analysis::CFG cfg = analysis::CFG::fromFunction(currFunc);
+
+      checkLivenessTransferFunction(cfg, currFunc);
+      checkReachingDefinitionsTransferFunction(cfg, currFunc);
```
tlively, line 339: To reduce work per seed, perhaps it would make sense to randomly pick one of these to do. On the other hand, maybe that's not worth it until we have a lot more lattices and analyses.

src/tools/wasm-fuzz-lattices.cpp
```diff
@@ -0,0 +1,389 @@
...
+  options.add("--seed",
+              "",
+              "Run a single workload generated by the given seed",
+              WasmFuzzTypesOption,
+              Options::Arguments::One,
+              [&](Options*, const std::string& arg) {
+                seed = uint64_t(std::stoull(arg));
+              });
+
+  bool verbose = false;
```
tlively, line 365: It doesn't look like you use the verbose flag for anything yet, but perhaps you could. See https://github.com/WebAssembly/binaryen/blob/main/test/lit/fuzz-types.test for an example of how to test that a fuzzer works using the verbose flag.

tlively (commented):
Nice, this looks really good overall. Perhaps as a preliminary PR, you could refactor some of the infrastructure this shares with the type fuzzer into a shared location?

src/analysis/lattice.h
```diff
@@ -102,6 +119,9 @@ class FiniteIntPowersetLattice {
 
   // Returns an instance of the bottom lattice element.
   Element getBottom();
+
+  // Returns a random lattice element. Used for fuzzing.
+  Element getRandom(Random& rand);
```
kripken, line 124: Do we need this in the class? I'd hope the fuzzer can do this itself.

zm2he: Refactored fuzzer to do this instead. After some initial examination, I found that the interface of the current powerset lattices were sufficient to generate random lattice without accessing private memebers.

src/tools/wasm-fuzz-lattices.cpp
zm2he: In order to make the user less dependent on loop execution
- I only generate one triple of lattice elements per function instead of one triple per block. However, as the current lattices are dependent on function-specific information for initialization, I can't do less.
- I used the seed to generate per-function "seeds". These I expose to the user in error messages, as well as the name of failing functions. The user can use these seeds and function names to re-run specific function with a function-specific seed.

Doing the above is a compromise between making the tests deterministic enough for easy replication while keeping open the possibility to introduce more random tests for every module generated.

src/tools/wasm-fuzz-lattices.cpp
```diff
@@ -0,0 +1,479 @@
...
+      Fatal() << ss.str();
+    }
+  }
+
+  // Anti-Symmetry is defined as x <= y and y <= x imply x = y. Due to the
+  // fact that the compare(x, y) function of the lattice explicitly tells
+  // us if two lattice elements are <, =, or = instead of providing a
+  // <= comparison, it is not useful to check for anti-symmetry as it is defined
+  // in the fuzzer.
+
```
kripken, line 73:
```suggestion
  //
```
(This is all one big comment.)

src/tools/wasm-fuzz-lattices.cpp
```diff
@@ -0,0 +1,479 @@
...
+  // Instead, we check for a related concept that x < y implies y > x, and
+  // vice versa in this checkAntiSymmetry function.
```
kripken, line 75: You can also check that `x < y => x != y`.

src/tools/wasm-fuzz-lattices.cpp
```diff
@@ -0,0 +1,479 @@
...
+
+  Fuzzer(bool verbose) : verbose(verbose) {}
+
+  // Helper function to run per-function tests. latticeElementSeed is used to
+  // generate three lattice elements randomly. It is also used to select which
+  // analysis is to be tested for the function.
+  void runOnFunction(Function* func, uint64_t latticeElementSeed) {
+    RandEngine getFuncRand(latticeElementSeed);
+
+    // less bytes are needed to generate three random lattices.
```
kripken, line 306:
```suggestion
    // Less bytes are needed to generate three random lattices.
```

src/tools/wasm-fuzz-lattices.cpp
zm2he: The check already does that, but I can make that more clear in the comments.

src/tools/wasm-fuzz-lattices.cpp
tlively, line 306:
```suggestion
    // Fewer bytes are needed to generate three random lattices.
```
;)

src/tools/wasm-fuzz-lattices.cpp
```diff
@@ -0,0 +1,479 @@
...
+    for (size_t i = 0; i < funcBytes.size(); i += sizeof(uint64_t)) {
+      *(uint64_t*)(funcBytes.data() + i) = getFuncRand();
+    }
+
+    Random rand(std::move(funcBytes));
+
+    CFG cfg = CFG::fromFunction(func);
+
+    switch (rand.upTo(2)) {
+      case 0: {
```
tlively, line 317: Maybe each of these cases can be packaged up into a separate function, just to keep the `runOnFunction` routine shorter and easier to read.

tlively (commented):
It would be good to add a lit test as well, similar to the existing test for the type fuzzer.

src/tools/wasm-fuzz-lattices.cpp
zm2he: I could probably consolidate this under `AnalysisChecker` entirely if I passed a pointer to one the specific checker to it. However, I don't know if that would be too messy, since the `AnalysisChecker` is already the owner of the specific checker.

src/tools/wasm-fuzz-lattices.cpp
```diff
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2021 WebAssembly Community Group participants
```
tlively, line 2:
```suggestion
 * Copyright 2023 WebAssembly Community Group participants
```

src/tools/wasm-fuzz-lattices.cpp
```diff
@@ -90,6 +122,44 @@ template<typename Lattice, typename TransferFunction> struct AnalysisChecker {
     }
   }
 
+private:
+  // Prints the error message when a triple of lattice elements violates
+  // transitivity.
+  inline void printTransitivityError(std::ostream& os,
```
tlively, line 128: No need for `inline` here. Since these methods are defined inside the struct definition, they are implicitly `inline`.

src/tools/wasm-fuzz-lattices.cpp
tlively: This LGTM how it is 👍

tlively (approved):

PR #5832, opened 2023-07-21
[Outlining] Changing stringify values to 32-bit

ashleynh:
The LLVM suffix tree expects to be provided with a vector of 32-bit unsigned integers. This PR makes it easier to integrate our wasm program string with the suffix tree.

Because the range of possible values is reduced from 2^64 to 2^32, a signed integer was added to manage the next separator value and ensure we're using every possible negative number.

### DIFF
--- a/src/passes/hash-stringify-walker.cpp
+++ b/src/passes/hash-stringify-walker.cpp
@@ -55,7 +55,8 @@ bool StringifyEquator::operator()(Expression* lhs, Expression* rhs) const {
 void HashStringifyWalker::addUniqueSymbol() {
   // Use a negative value to distinguish symbols for separators from symbols
   // for Expressions
-  hashString.push_back((unsigned)nextSeparatorVal);
+  assert((uint32_t)nextSeparatorVal >= nextVal);
+  hashString.push_back((uint32_t)nextSeparatorVal);
   nextSeparatorVal--;
 }
 
--- a/src/passes/hash-stringify-walker.cpp
+++ b/src/passes/hash-stringify-walker.cpp
@@ -55,8 +55,8 @@ bool StringifyEquator::operator()(Expression* lhs, Expression* rhs) const {
 void HashStringifyWalker::addUniqueSymbol() {
   // Use a negative value to distinguish symbols for separators from symbols
   // for Expressions
-  hashString.push_back((uint64_t)-nextVal);
-  nextVal++;
+  hashString.push_back((unsigned)nextSeparatorVal);
+  nextSeparatorVal--;
 }
 
 void HashStringifyWalker::visitExpression(Expression* curr) {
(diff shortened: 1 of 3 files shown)


src/passes/stringify-walker.h
```diff
@@ -115,15 +115,18 @@ struct HashStringifyWalker : public StringifyWalker<HashStringifyWalker> {
   // After calling walkModule, this vector contains the result of encoding a
   // wasm module as a string of uint64_t values. Each value represents either an
   // Expression or a separator to mark the end of control flow.
-  std::vector<uint64_t> hashString;
+  std::vector<unsigned> hashString;
```
tlively, line 118: Can we use `uint32_t` instead of `unsigned`? The extra explicitness is nice, and they should be compatible on all systems we build on.

src/passes/stringify-walker.h
```diff
@@ -115,15 +115,18 @@ struct HashStringifyWalker : public StringifyWalker<HashStringifyWalker> {
...
   // A monotonic counter used to ensure that unique expressions in the
   // module are assigned a unique value in the hashString
-  uint64_t nextVal = 0;
+  unsigned nextVal = 0;
+  // A monotonic counter used to ensure that each separator in the
+  // module is assigned a unique value in the hashString
+  signed nextSeparatorVal = -1;
```
tlively, line 124: I've never seen someone use `signed` like this 😆

src/passes/hash-stringify-walker.cpp
```diff
@@ -55,8 +55,8 @@ bool StringifyEquator::operator()(Expression* lhs, Expression* rhs) const {
 void HashStringifyWalker::addUniqueSymbol() {
   // Use a negative value to distinguish symbols for separators from symbols
   // for Expressions
-  hashString.push_back((uint64_t)-nextVal);
-  nextVal++;
+  hashString.push_back((unsigned)nextSeparatorVal);
+  nextSeparatorVal--;
```
tlively, line 59: We could be extra safe here by asserting that the negative numbers and positive numbers have not crossed each other and started accidentally repeating values.

src/passes/stringify-walker.h
ashleynh: done

src/passes/stringify-walker.h
ashleynh: done

src/passes/hash-stringify-walker.cpp
ashleynh: done

tlively (approved):
LGTM! It would be good to shorten the commit title if possible. Usually 50 characters is a good target and 72 characters is taken as a hard limit.

kripken (approved):

PR #5837, opened 2023-07-25
wasm-merge: Fix locals in merged start

### LINKED ISSUE #5835: wasm-merge drops locals from merged start function
When using `wasm-merge` on the following minimal example there is a validation failure:

`wasm-merge mini/a.wast a mini/b.wast b --emit-text -o link.wast`

`a.wast`
```wast
(module
  (import "extern" "print" (func $print (param i32)))
  (func $start
    (local $x i32)
    (local.set $x (i32.const 1))
    (call $print (local.get $x)))
  (start $start))
```

`b.wast`
```wast
(module
  (import "extern" "print" (func $print (param i32)))
  (func $start
    (local $y i32)
    (local.set $y (i32.const 2))
    (call $print (local.get $y)))
  (start $start))
```

With the following error result.
The locals from the original start functions are dropped from the merged one.
```wast
(module
 (type $none_=>_none (func))
 (type $i32_=>_none (func (param i32)))
 (import "extern" "print" (func $print (param i32)))
 (import "extern" "print" (func $print_2 (param i32)))
 (start $merged.start)
 (func $start
  (local $x i32)
  (local.set $x
   (i32.const 1))
  (call $print
   (local.get $x)))
 (func $start_2
  (local $y i32)
  (local.set $y
   (i32.const 2))
  (call $print_2
   (local.get $y)))

 (func $merged.start
  (block
   (local.set $0
    (i32.const 1))
   (call $print
    (local.get $0)))
  (block
   (local.set $0
    (i32.const 2))
   (call $print_2
    (local.get $0)))))
```

kripken:
Start functions can have locals, which we previously ignored as we just
concatenated the bodies together. This makes us copy the second start
and call that, keeping them separate (the optimizer can then inline, if that
makes sense).

fixes #5835

### DIFF
--- a/src/tools/wasm-merge.cpp
+++ b/src/tools/wasm-merge.cpp
@@ -362,19 +362,19 @@ void copyModuleContents(Module& input, Name inputName) {
       // No previous start; just refer to the new one.
       merged.start = input.start;
     } else {
-      // Merge them, keeping the order. Note that we need to create a new
-      // function as there may be other references.
+      // Merge them, keeping the order. We copy both functions to avoid issues
+      // with other references to them, and just call the second one, leaving
+      // inlining to the optimizer if that makes sense to do.
+      auto copiedOldName = Names::getValidFunctionName(merged, "merged.start.old");
+      auto copiedNewName = Names::getValidFunctionName(merged, "merged.start.new");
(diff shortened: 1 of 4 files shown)


chambart:
I confirm. It does fix the problem. Thanks

PR #5838, opened 2023-07-26
CFGWalker: Allow users to not create "continuation" blocks [NFC]

kripken:
"Continuation" here means A->B with no other branches from A to another basic block, and
no other branches to B. A might only have a branch out of the entire basic block, if it throws.
We do want a create a new basic block in that case normally, so that each basic block ends in
a control flow transfer (and no transfer occurs in the middle), but users that don't care about
that just suffer overhead from it.

This PR makes the relevant passes 9% faster.

This plus #5827 offset the slowdown from #5823 and overall give an improvement compared
to before.


src/ir/LocalGraph.cpp
```diff
@@ -52,6 +52,13 @@ struct Flower : public CFGWalker<Flower, Visitor<Flower>, Info> {
 
   BasicBlock* makeBasicBlock() { return new BasicBlock(); }
 
+  void continueToNewBasicBlock() {
```
tlively, line 55: Are there other reasonable things this could be overridden with? I'm not sure there are, and if there were, they would require updating the internal state of the `CFGWalker`, which seems less than ideal. Maybe this could all be simplified if CFGWalker just took a bool (or similar enum) template parameter to configure this behavior?

kripken: A bool would work here, yeah. But in general another option here could be to check if the block is empty and merge it in that case. That is, if there are no other instructions a pass cares about, then it might be ok with merging the continuation, but not otherwise. That would require inspecting `block->contents` etc. So overall I think a hook like this makes sense.

kripken: (As a datapoint, I saw that 15% of blocks in LocalGraph are empty, that is, contain no LocalGet or Set instructions. That number is even higher for other passes, I'd bet, that focus on less common instructions. So there could be nice optimization opportunities here.)

kripken:
The test failure here suggests I have misunderstood something... investigating.

src/cfg/cfg-traversal.h
```diff
@@ -293,15 +293,36 @@ struct CFGWalker : public PostWalker<SubType, VisitorType> {
...
+  // it, with no other basic blocks relevant. That is, this is called when we
+  // are in block A and need to create a new block B with A -> B and A has no
+  // other branches from it, and B has no other branches to it, aside from A
+  // possibly exiting the entire function (e.g. via a throw). Some users of this
+  // class can override this behavior to not create a basic block here, if they
+  // don't need to preserve the property that a basic block ends with an
+  // instruction that might branch (that is, if we don't create a new basic
+  // block here then we might transfer control out of the entire function from
+  // the middle of the block).
+  void continueToNewBasicBlock() {
```
aheejin, line 306: Nit: How about `continueToUniqueSuccessor` or something? Without reading the comment, this name can be confusing, because we basically continue to a new basic block every time we create a new basic block.

aheejin (approved):

src/cfg/cfg-traversal.h
kripken: Yeah, good point. I've also been thinking about @tlively concerns from earlier about the complexity.

Perhaps it would be simpler to formulate this as "ignores branches to outside of the function." That would apply to all the relevant passes since they just look at locals (which go away anyhow if we branch out). How does that sound @aheejin @tlively ?

tlively: That sounds good to me. For the analysis framework we will an option like that to configure the correct behavior when building intraprocedural and interprocedural CFGs.

kripken: Ok, I made that change. I think it is simpler that way, PTAL.

Btw @aheejin - great test in `test/lit/passes/rse-eh.wast`, it caught my misunderstanding here and made it very easy to figure out!

tlively (approved):
LGTM, but how about making the bool a template parameter and using `if constexpr` when we check it to guarantee that it's resolved at compile time?

kripken:
@tlively Perhaps, but it will take more plumbing to achieve that. In particular the template parameter would need to be passed through LivenessWalker which is a subtype of CFGWalker (which is how e.g. CoalesceLocals uses CFGWalker indirectly), and maybe elsewhere. I'm not sure it's worth the extra boilerplate to try to avoid a missed optimization here, since it seems both very likely to be optimized but also not matter much if it isn't?

src/cfg/cfg-traversal.h
aheejin: Sounds good to me too!

aheejin (approved):

tlively:
Ok, no worries if making the configuration a template parameter would be more complex.

PR #5839, opened 2023-07-26
[Outlining] Integration test for suffix_tree and stringify

ashleynh:
Adds an integration test that identifies the substrings of a stringified wasm module using the suffix_tree.

### DIFF
--- a/src/support/suffix_tree.h
+++ b/src/support/suffix_tree.h
@@ -172,8 +172,15 @@ public:
     void advance();
 
   public:
+    using iterator_category = std::input_iterator_tag;
+    using value_type = RepeatedSubstring;
+    using difference_type = std::ptrdiff_t;
+    using pointer = const RepeatedSubstring*;
+    using reference = const RepeatedSubstring&;
+
     /// Return the current repeated substring.
     RepeatedSubstring& operator*() { return RS; }
+    RepeatedSubstring* operator->() { return &RS; }
 
     RepeatedSubstringIterator& operator++() {
       advance();
--- a/src/support/suffix_tree.h
+++ b/src/support/suffix_tree.h
@@ -60,8 +60,7 @@ public:
     std::vector<unsigned> StartIndices;
 
     bool operator==(const RepeatedSubstring& other) const {
-      return this->Length == other.Length &&
-             this->StartIndices == other.StartIndices;
+      return Length == other.Length && StartIndices == other.StartIndices;
     }
   };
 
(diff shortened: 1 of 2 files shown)


src/support/suffix_tree.h
```diff
@@ -58,6 +58,11 @@ class SuffixTree {
 
     /// The start indices of each occurrence.
     std::vector<unsigned> StartIndices;
+
+    bool operator==(const RepeatedSubstring& other) const {
+      return this->Length == other.Length &&
+             this->StartIndices == other.StartIndices;
```
tlively, line 64:
```suggestion
      return Length == other.Length &&
             StartIndices == other.StartIndices;
```

test/gtest/stringify.cpp
```diff
@@ -216,3 +217,26 @@ TEST_F(StringifyTest, Stringify) {
...
+  parseWast(wasm, dupModuleText);
+
+  HashStringifyWalker stringify = HashStringifyWalker();
+  stringify.walkModule(&wasm);
+
+  SuffixTree st(stringify.hashString);
+  std::vector<SuffixTree::RepeatedSubstring> substrings;
+  for (auto it = st.begin(); it != st.end(); it++) {
+    substrings.push_back(*it);
+  }
```
tlively, line 232:
```suggestion
  std::vector<SuffixTree::RepeatedSubstring> substrings(st.begin(), st.end());
```

tlively (approved):
Should suffix_tree.h say somewhere that it comes from LLVM?

src/support/suffix_tree.h
ashleynh: done!

test/gtest/stringify.cpp
ashleynh: done

test/gtest/stringify.cpp
```diff
@@ -216,3 +217,26 @@ TEST_F(StringifyTest, Stringify) {
...
+
+  EXPECT_EQ(
+    substrings,
+    (std::vector<SuffixTree::RepeatedSubstring>{
+      SuffixTree::RepeatedSubstring{3u, (std::vector<unsigned>{10, 23})},
+      SuffixTree::RepeatedSubstring{4u, (std::vector<unsigned>{9, 22})},
+      SuffixTree::RepeatedSubstring{2u, (std::vector<unsigned>{11, 24})},
+      SuffixTree::RepeatedSubstring{3u, (std::vector<unsigned>{18, 27})},
+      SuffixTree::RepeatedSubstring{2u, (std::vector<unsigned>{32, 19, 28})}}));
```
kripken, line 241: To verify this, should I be able to see the repeated substrings on line 173-217? If so, maybe add 0,1,2,3 etc. there to make that simpler?

PR #5840, opened 2023-07-26
TypeRefining: Add casts when we must

kripken:
See the example in the code and test for a situation that requires this for validation.

To fix validation we add a cast. That should practically always be removed by later
optimizations, and the fact it took the fuzzer this long to even find such a situation
also adds confidence that this won't be adding overhead (and in this situation, the
optimizer will definitely remove the cast).


tlively (approved):

tlively:
FWIW I expect to need this sort of thing more when we do optimizations based on the new analysis framework, since that will never take unreachable code into account, either.

PR #5842, opened 2023-07-26
Fix a crash in TypeRefining on bottom types

kripken:
This is what I get for not fuzzing #5840 enough beforehand.

### DIFF
--- a/src/passes/TypeRefining.cpp
+++ b/src/passes/TypeRefining.cpp
@@ -400,11 +400,16 @@ struct TypeRefining : public Pass {
 
       void visitStructSet(StructSet* curr) {
         if (curr->type == Type::unreachable) {
+          // Ignore unreachable code.
+          return;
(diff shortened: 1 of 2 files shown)


src/passes/TypeRefining.cpp
```diff
@@ -400,11 +400,16 @@ struct TypeRefining : public Pass {
 
       void visitStructSet(StructSet* curr) {
         if (curr->type == Type::unreachable) {
+          // Ignore unreachable code.
+          return;
+        }
+        auto type = curr->ref->type.getHeapType();
+        if (!type.isStruct()) {
```
tlively, line 407: This could be slightly more explicit:
```suggestion
        if (type.isBottom()) {
```

tlively (approved):
Disabled automerge in case you want to make that update, but LGTM otherwise.

PR #5843, opened 2023-07-27
[NFC] Port passes remove-unused-brs_all-features.wast to lit

tlively:
Port the test automatically using the port_passes_tests_to_lit.py script. As a
drive-by, fix a typo in the script as well.

PR #5844, opened 2023-07-27
Support sending additional values with BrOn

tlively:
Like the MVP (+multivalue) `br` and `br_if` instructions, `br_on*` instructions
are supposed to be able to send an arbitrary number of values to their target
labels. Add support for this along with tests for round-tripping through the
binary format, and also update RemoveUnusedBrs to handle these additional values
correctly.

The complexity here comes from the fact that the BrOn instructions already propagate
their inspected reference operands to their following code, their branch target,
or both, so adding additional values necessarily introduces multivalue. The use
of multivalue produces a significant amount of bloat after round-tripping and
also makes RemoveUnusedBrs significantly more complicated.

Despite this extra complexity, supporting arbitrary sent values is important for
spec compliance and because a follow-up PR will make Binaryen's output send
values on these branches in some cases as a fixup for other-wise invalid
br_on_cast instructions. Supporting these sent values in the IR and parsers will
be necessary for Binaryen to parse its own output.

See https://github.com/WebAssembly/gc/issues/381#issuecomment-1644489622 for
more context on the planned follow-up work.


tlively:
* main
  * **PR #5843** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5843" target="_blank">Graphite</a> 
    * **PR #5844** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5844" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5844?utm_source=stack-comment).

PR #5847, opened 2023-07-27
PossibleContents: Support more intersection types

kripken:
This is necessary for a future PR that will infer more things than full cones, which
were the thing we supported until this PR in intersections.

### DIFF
--- a/test/gtest/possible-contents.cpp
+++ b/test/gtest/possible-contents.cpp
@@ -325,7 +325,7 @@ TEST_F(PossibleContentsTest, TestIntersection) {
   assertHaveIntersection(exactI32, exactI32);
   assertHaveIntersection(i32Zero, i32Zero);
   assertHaveIntersection(exactFuncSignatureType, exactFuncSignatureType);
-  assertHaveIntersection(i32Zero, i32One); // TODO: this could be inferred false
+  assertLackIntersection(i32Zero, i32One);
 
   // Exact types only differing by nullability can intersect (not on the null,
   // but on something else).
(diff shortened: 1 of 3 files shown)


test/gtest/possible-contents.cpp
```diff
@@ -810,6 +809,21 @@ TEST_F(PossibleContentsTest, TestStructCones) {
   assertIntersection(
     literalNullA, PossibleContents::fullConeType(funcref), none);
 
+  // Computing intersections is also supported with a Literal.
+  assertIntersection(i32Zero, i32Zero, i32Zero);
+  assertIntersection(i32One, i32Zero, none);
+  assertIntersection(i32Global1, i32Zero, i32Zero);
```
tlively, line 815: Why is zero the intersection of a global and zero?

tlively (approved):
Looks good, but I'm concerned about the length and complexity of `PossibleContents::intersect`. Perhaps it's a good candidate for a separate refactoring PR?

I also still don't understand why GUFA needs intersection at all, but I know you've explained this to me multiple times in the past, so I can make my peace with this 🥲

test/gtest/possible-contents.cpp
kripken: Basically since we consider literal values the most precise thing possible. But we could in theory look at it another way. If we compare a Global to the intersection of a Global and Zero, then if the intersection was a Global we could evaluate it to true, but not if we had a Zero there - not unless we knew the Global contained a Zero. But I think if we do that intersection then we should be able to infer that about the Global in general.

Practically, I think it's useful to keep literals as the most precise/refined things possible.

kripken:
> I also still don't understand why GUFA needs intersection at all

GUFA is primarily a simple "flow" operation where values flow forward, but since the start it has also noted information about each location and filtered. For example, any write to an immutable Global is filtered to be, at the worst, a Global. And any write to an expression location is filtered to the type of the expression. We use intersections to do that filtering.

The new TrapsNeverHappen reverse analysis will provide another source of such filtering information, that is, it will tell us "this place can't contain anything worse than X, because TNH proves it so," so we'll intersect with X there.

tlively:
Aha, so intersection is used in the implementation of the transfer function, and does not occur at control flow merges. Makes sense, thanks!

kripken:
Yes, I guess that's the right terminology in the sense of abstract interpretation / static program analysis: the transfer function does filtering. (Which is also why this must still converge despite the flow increasing while the filtering reducing - each location remains monotonic, as the transfer function is applied in each before any flow is added.)

PR #5849, opened 2023-07-31
Lattice to model Stack

zm2he:
This change introduces a lattice `StackLattice` to model stack-related behavior. It is templated on a separate lattice which models some property of values on the stack. The `StackLattice` allows users to access the top of the stack, push abstract values, and pop them. Comparisons and least upper bound calculations are done on a value by value basis starting from the top of the stack and moving toward the bottom. This is done since it allows stacks from different scopes to be joined together easily. More information is in the comments.

One intended application of it is to use it to model the wasm value stack. The goal is to organize lattice elements representing individual stack values in a natural way which mirrors the wasm value stack. Transfer functions operate on each stack value individually. The stack lattice is an intermediate structure which is not intended to be directly operated on. Rather, it simply simulates the push and pop behavior of instructions.


src/analysis/stack-lattice.h
```diff
@@ -0,0 +1,139 @@
...
+  static_assert(is_lattice<StackFrameLattice>);
+  StackFrameLattice& stackFrameLattice;
+
+public:
+  StackLattice(StackFrameLattice& stackFrameLattice)
+    : stackFrameLattice(stackFrameLattice) {}
+
+  class Element {
+    std::deque<typename StackFrameLattice::Element> stackValue;
+    bool top = false;
```
tlively, line 20: How about using a `std::optional<std::deque<...>>` where top is represented by `std::nullopt`? That way we don't have to worry about the deque hanging out in top values.

src/analysis/stack-lattice.h
```diff
@@ -0,0 +1,139 @@
...
+    void pop() { stackValue.pop_back(); }
+
+    bool makeLeastUpperBound(const Element& other) {
+      bool modified = false;
+      auto otherIterator = other.stackValue.crbegin();
+      for (auto thisIterator = stackValue.rbegin();
+           thisIterator != stackValue.rend();
+           ++thisIterator) {
+        if (otherIterator != other.stackValue.crend()) {
+          modified |= (*thisIterator).makeLeastUpperBound(*otherIterator);
```
tlively, line 50:
```suggestion
          modified |= thisIterator->makeLeastUpperBound(*otherIterator);
```

src/analysis/stack-lattice.h
```diff
@@ -0,0 +1,139 @@
...
+      stackValue.push_back(frame);
+    }
+    void pop() { stackValue.pop_back(); }
+
+    bool makeLeastUpperBound(const Element& other) {
+      bool modified = false;
+      auto otherIterator = other.stackValue.crbegin();
+      for (auto thisIterator = stackValue.rbegin();
+           thisIterator != stackValue.rend();
+           ++thisIterator) {
```
tlively, line 48: You could check `otherIterator != other.stackValue.crend()` here in the loop condition and I believe you could do `++thisIterator, ++otherIterator` as the update statement to make this really nice and symmetrical. If that does work, I would pull out `thisIterator` and initialize it above the loop for symmetry with `otherIterator` as well.

src/analysis/stack-lattice.h
```diff
@@ -0,0 +1,139 @@
...
+          hasEqual = true;
+          break;
+        case LatticeComparison::LESS:
+          hasLess = true;
+          break;
+        case LatticeComparison::GREATER:
+          hasGreater = true;
+          break;
+        default:
+          hasUnrelated = true;
```
tlively, line 98: You could just return `NO_RELATION` at this point and save a line or two below. That would also let you get rid of `hasEqual` and infer it below from `!hasLess && !hasGreater`.

src/analysis/stack-lattice.h
```diff
@@ -0,0 +1,139 @@
...
+      ++leftIterator;
+      ++rightIterator;
+    }
+
+    if (left.stackValue.size() > right.stackValue.size()) {
+      if (!hasUnrelated && !hasLess) {
+        return LatticeComparison::GREATER;
+      } else {
+        return LatticeComparison::NO_RELATION;
+      }
```
tlively, line 109: I think you could model this with `hasGreater = true` and fall through to the logic below. Similar with `hasLess = true` when `right` is longer than `left`.

test/gtest/cfg.cpp
```diff
@@ -579,3 +581,67 @@ TEST_F(CFGTest, ReachingDefinitionsLoop) {
...
+    BitsLattice::Element temp = bitsLattice.getBottom();
+    temp.setValue(i);
+    secondStack.push(temp);
+  }
+
+  EXPECT_EQ(stackLattice.compare(firstStack, secondStack),
+            LatticeComparison::EQUAL);
+
+  secondStack.pop();
+  secondStack.pop();
```
tlively, line 608: It will probably be nice later to have `pop` return the element rather than having to do `stackTop` and `pop` separately.

tlively (commented):
Nice! A few top-level comments:

1. Please add newlines at the end of files
2. It would probably be good to split the two lattices into separate PRs so they can be reviewed separately. For testing the stack lattice, you can use the powerset lattice we already have as the stack element lattice.
3. It would be good to add fuzzer support as well as unit tests for each lattice.

src/analysis/bits-lattice.h
```diff
@@ -0,0 +1,83 @@
...
+
+// Lattice for used bits analysis. A flat uint32_t lattice, where
+// all numerical values are equal to each other in an abstract
+// sense, and there is a top above and a bottom below them.
+struct BitsLattice {
+  enum Status { BOTTOM, NUMBER, TOP };
+
+  class Element {
+    Status status = Status::BOTTOM;
+    uint32_t value = 0;
```
kripken, line 16: Why can't we just use `value`, in which top and bottom would be all zeros or all ones?

src/analysis/stack-lattice.h
```diff
@@ -0,0 +1,171 @@
...
+#define wasm_analysis_stack_lattice_h
+
+#include <deque>
+
+#include "lattice.h"
+
+namespace wasm::analysis {
+
+// A lattice for modelling stack-related phenomena. It is templated
+// on a lattice representing the contents of each stack frame.
```
kripken, line 11: Perhaps some examples would be useful here. Starting to read this file, I'm not sure if stack here means the call stack, or the wasm value stack, or a generic stack of anything unrelated to wasm, or otherwise.

test/gtest/cfg.cpp
```diff
@@ -579,3 +581,67 @@ TEST_F(CFGTest, ReachingDefinitionsLoop) {
 
   EXPECT_EQ(expectedResult, getSetses);
 }
+
+TEST_F(CFGTest, StackBitsLatticeFunctioning) {
+  BitsLattice bitsLattice;
+  StackLattice<BitsLattice> stackLattice(bitsLattice);
```
kripken, line 587: Can we test these two separately? Or are they only meant to be used with each other? If so that wasn't obvious to me from the headers.

zm2he:
I have removed the bits lattice from this PR, and am planning to introduce it in a separate PR, where I will address the comments related to it. The unit test has been modified to use an existing `FiniteIntPowersetLattice` to model stack elements.

test/gtest/cfg.cpp
zm2he: The `StackLattice` needs another lattice to model stack contents (i.e. each stack element). However, I will replace `BitsLattice` with an existing lattice.

src/analysis/stack-lattice.h
```diff
@@ -0,0 +1,187 @@
...
+template<typename StackElementLattice> class StackLattice {
+  static_assert(is_lattice<StackElementLattice>);
+  StackElementLattice& stackElementLattice;
+
+public:
+  StackLattice(StackElementLattice& stackElementLattice)
+    : stackElementLattice(stackElementLattice) {}
+
+  class Element {
+    // The top lattice is
```
kripken, line 23: Is this comment cut off?

src/analysis/stack-lattice.h
```diff
@@ -0,0 +1,187 @@
...
+#define wasm_analysis_stack_lattice_h
+
+#include <deque>
+#include <optional>
+
+#include "lattice.h"
+
+namespace wasm::analysis {
+
+// Models the WASM value stack
```
kripken, line 11:
```suggestion
// Models the Wasm value stack.
```

src/analysis/stack-lattice.h
```diff
@@ -0,0 +1,187 @@
...
+    typename StackElementLattice::Element pop() {
+      typename StackElementLattice::Element value = stackValue->back();
+      stackValue->pop_back();
+      return value;
+    }
+
+    // When taking the LUB, we take the LUBs of the elements of the suffixes of
+    // each stack. This is because when we have stack1 = [a] and stack2 = [b',
+    // a'] which is from a scope which encloses the scope of stack1, we want the
+    // result to be [b', LUB(a, a')]. We can alternatively imagine that each
```
kripken, line 65: Why do we want that? I feel there is another step between this and using it somewhere that I don't see explained yet. Maybe an example could help?

Also, what would we want to happen to these two stacks? `[b, a] ; [b', a']`

src/analysis/stack-lattice.h
kripken: Also, the two lines of comments below it now look a little too generic. I would expect to see something like this:
```
// A lattice to model the Wasm value stack.
//
// This is templated on a given lattice, and then represents stacks of elements of
// that one. These stacks can represent things on the Wasm value stack, such as
// ... For example, imagine that we want to optimize ... then we would use this
// on a lattice ... and then the optimization would work like ...
```

(I have no idea how to fill in those blanks, but that's the shape I'm looking for :smile: )

src/analysis/stack-lattice.h
zm2he: No, I'm not done writing the comments yet.

src/analysis/stack-lattice.h
tlively: OTOH, there's nothing in this lattice that ties it to the Wasm value stack specifically. That's just the first application we have in mind. I could also see this modeling exception handler stacks, for instance. I think it would be best to document the properties of this lattice (i.e. comparisons and joins being elementwise from the top of the stack rather than from the bottom of the stack) in more detail and use the connection to the Wasm value stack as an illustrative example.

src/analysis/stack-lattice.h
```diff
@@ -0,0 +1,189 @@
...
+    } else if (hasLess && hasGreater) {
+      return LatticeComparison::NO_RELATION;
+    } else {
+      return LatticeComparison::EQUAL;
+    }
+  }
+
+  Element getBottom() {
+    Element result;
+    return result;
```
tlively, line 183:
```suggestion
    return {};
```

or

```suggestion
    return Element{};
```

would also work.

src/tools/wasm-fuzz-lattices.cpp
```diff
@@ -363,6 +364,66 @@ struct ReachingDefinitionsChecker {
...
 };
 
+// Struct to set up and check the stack lattice. Uses the
+// FiniteIntPowersetLattice as to model stack contents.
+struct StackLatticeChecker {
+  FiniteIntPowersetLattice contentLattice;
+  StackLattice<FiniteIntPowersetLattice> stackLattice;
+
+  // Here only as a placeholder, since the checker utility currently wants a
+  // transfer function.
```
tlively, line 374: It would be good to fix this in a separate PR 👍

test/gtest/cfg.cpp
```diff
@@ -579,3 +580,84 @@ TEST_F(CFGTest, ReachingDefinitionsLoop) {
...
+    }
+    secondStack.push(std::move(temp));
+  }
+
+  EXPECT_EQ(stackLattice.compare(firstStack, secondStack),
+            LatticeComparison::EQUAL);
+
+  std::stringstream ss;
+  secondStack.pop().print(ss);
+  EXPECT_EQ(ss.str(), "1111");
```
tlively, line 614: Would it be easy to compare equality directly without converting to a string first? If so, that would simplify the test and make it test the real API more directly.

zm2he: I could do this. But one problem is, when comparing the `StackLattice`, I don't know how useful that would be for everything except the result of the LUB, since we are testing the correctness of the API of the lattice, but at the same time our "expected result" is also constructed using the exact same API. Of course, I could ditch the above test and focus on testing LUBs more.

src/analysis/stack-lattice.h
zm2he: Added some more clarification.

src/analysis/stack-lattice.h
zm2he: Added some more clarification, here, and at the top of the file. For the case you gave,
`LUB([b, a], [b', a']) = [LUB(b,b'), LUB(a,a')]`

src/analysis/stack-lattice.h
```diff
@@ -0,0 +1,249 @@
...
+// infinite stack of the form (bottom) [... BOTTOM, BOTTOM, b, a] (top). This
+// makes stacks in different scopes equivalent, with only their contents
+// different. Stacks in more "inner" scopes simply have more bottom elements in
+// the bottom portion.
+//
+// A common application for this lattice is modeling the Wasm value stack.
+// For instance, one could use this to analyze the maximum bit size of
+// values on the Wasm value stack. When new actual values are pushed
+// or popped off the Wasm value stack by instructions, the same is done
+// to abstract lattice elements in the StackLattice.
```
kripken, line 55: I'm still missing something here. Why would I want to know the maximum bit size of all values on the wasm stack? Wouldn't I care about a specific value being operated on?

E.g. if I have an `and` operation of `1` and something else, I want to look at that something else (and if it doesn't have the lowest bit set, the result is zero). I wouldn't care if other values that happen to be on the stack have that bit set or not.

I feel I'm missing some background for this class...

src/analysis/stack-lattice.h
```diff
@@ -8,9 +8,57 @@
...
+// of scopes. For instance, if we have the following program
+//
+// i32.const 0
+// i32.const 0
+// if (result i32)
+// i32.const 1
+// else
+// i32.const 2
+// endif
+// i32.add
```
tlively, line 37:
```suggestion
// i32.const 0
// i32.const 0
// if (result i32)
//   i32.const 1
// else
//   i32.const 2
// end
// i32.add
```

src/analysis/stack-lattice.h
```diff
@@ -8,9 +8,57 @@
...
+//
+// Before the if-else control flow, we have [] -> [i32], and after the if-else
+// control flow we have [i32, i32] -> [i32]. However, inside each of the if
+// and else conditions, we have [] -> [i32], because they cannot see the
+// stack elements pushed by the enclosing scope. In effect in the if and else,
+// we have a stack [i32 | i32], where we can't "see" left of the |.
+//
+// Conceptually, we can also imagine each stack [b, a] as being implicitly an
+// infinite stack of the form (bottom) [... BOTTOM, BOTTOM, b, a] (top). This
+// makes stacks in different scopes equivalent, with only their contents
```
tlively, line 47:
```suggestion
// makes stacks in different scopes comparable, with only their contents
```

src/analysis/stack-lattice.h
```diff
@@ -8,9 +8,57 @@
...
+// different. Stacks in more "inner" scopes simply have more bottom elements in
+// the bottom portion.
+//
+// A common application for this lattice is modeling the Wasm value stack.
+// For instance, one could use this to analyze the maximum bit size of
+// values on the Wasm value stack. When new actual values are pushed
+// or popped off the Wasm value stack by instructions, the same is done
+// to abstract lattice elements in the StackLattice.
+//
+// When two control flows a joined together, one with stack [b, a] and
```
tlively, line 57:
```suggestion
// When two control flows are joined together, one with stack [b, a] and
```

src/analysis/stack-lattice.h
```diff
@@ -20,14 +68,13 @@ template<typename StackElementLattice> class StackLattice {
     : stackElementLattice(stackElementLattice) {}
 
   class Element {
-    // The top lattice is
+    // The top lattice can be imagined as an infinitely high stack, which
```
tlively, line 71:
```suggestion
    // The top lattice can be imagined as an infinitely high stack of top elements, which
```

src/analysis/stack-lattice.h
```diff
@@ -20,14 +68,13 @@ template<typename StackElementLattice> class StackLattice {
...
+    // is unreachable in most cases. In practice, we make the stack an
+    // optional, and if we are the top lattice, the stack doesn't exist in
+    // the optional signalling the top lattice.
```
tlively, line 74: editing just for clarity
```suggestion
    // optional, and we represent top with the absence of a stack.
```

tlively (approved):
LGTM % small suggested edits

kripken (approved):

src/analysis/stack-lattice.h
zm2he: An example with used bits was added.

PR #5850, opened 2023-07-31
GUFA: Infer using TrapsNeverHappen

kripken:
This adds a TrapsNeverHappen oracle that is used inside the main PossibleContents
oracle of GUFA. The idea is that when traps never happen we can reason "backwards"
from information to things that must be true before it:
```cpp
temp = x.field;
x.cast_to<Y>(); // Y is a subtype of x's type X
```
Here we cast `x` to a subtype. If we assume traps never happen then the cast must
succeed, and that means we can assume we had a `Y` on the previous line, where
perhaps that information lets us infer the value of `x.field`.

This PR focuses on calls, which are the more interesting situation to optimize
because other passes do some work already inside functions. Specifically, we look
for things that will trap in the called function or the caller, such as if the called
function always casts a param to some type, we can assume the caller passes
such a type in. And if we have a call_ref then any target that would trap cannot be
called (at least in a closed world).

This has some benefits, in particular when combined with `--gufa-cast-all` since
that casts more things, which lets us apply the inferences made here. I see `3.3%`
fewer `call_ref` instructions on a Kotlin testcase, for example, but on Java the
improvement is a little less. That is on `-Os` where this helps more, as in `-O3`
we inline more, which means fewer calls. Overall, the situations this optimizes
occur enough to justify the effort I think, but the benefits are also not enormous.


src/ir/possible-contents.cpp
```diff
@@ -1321,12 +1323,504 @@ struct InfoCollector
...
+        // No fallthrough, we're done with this param.
+        break;
+      }
+
+      // There is a fallthrough. Check for a control flow transfer.
+      if (blockIndexes.get(next) != callBlockIndex) {
+        // Control flow might transfer; stop.
+        transferred = true;
+        break;
+      }
```
tlively, line 1805: Can we move this out to where we check `transferred` anyway below?

src/ir/possible-contents.cpp
```diff
@@ -1961,25 +2483,41 @@ void Flower::filterExpressionContents(PossibleContents& contents,
...
 
   // The caller cannot know of a situation where it might not be worth sending
   // more to a reference - all that logic is in here. That is, the rest of this
   // function is the only place we can mark |worthSendingMore| as false for a
   // reference.
-  assert(worthSendingMore);
+  bool isRef = type.isRef();
+  if (isRef) {
+    assert(worthSendingMore);
+  }
```
tlively, line 2500:
```suggestion
  assert(!isRef || worthSendingMore);
```

Can you remind me what "sending more" means in this context?

tlively (commented):
Comments so far. Still have to take a look at the tests.

src/ir/possible-contents.cpp
```diff
@@ -1321,12 +1323,504 @@ struct InfoCollector
...
+  //    (A)
+  //    (br_if ..)
+  //    (B)
+  //  )
+  //
+  // If we branch in the br_if then we might execute A and then something else
+  // entirely, and not reach B or the call. In that case we can't infer anything
+  // about A (perhaps, for example, we branch away exactly when A would fail the
+  // cast). Therefore in the optimization below we only optimize code that, if
+  // reached, will definitely reach the call, like B.
```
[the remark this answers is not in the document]
kripken: We have a dominator tree atm: https://github.com/WebAssembly/binaryen/blob/main/src/cfg/domtree.h (but this pass doesn't need it yet).

I'm open to moving this to the new framework, if we see there will benefits to that.

src/ir/possible-contents.cpp
```diff
@@ -1321,12 +1323,504 @@ struct InfoCollector
...
+          continue;
+        }
+
+        // If our operands will fail a cast, then we will trap.
+        bool traps = false;
+        for (auto& [castIndex, castType] : targetInfo.castParams) {
+          auto operandType = call->operands[castIndex]->type;
+          if (!Type::isSubType(operandType, castType) &&
+              !Type::isSubType(castType, operandType)) {
+            traps = true;
```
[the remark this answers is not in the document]
kripken: Very good point... fixing.

src/ir/possible-contents.cpp
```diff
@@ -1321,12 +1323,504 @@ struct InfoCollector
...
+          }
+        }
+        if (!traps) {
+          possibleTargets.push_back(target);
+        }
+      }
+
+      if (possibleTargets.empty()) {
+        // No target is possible.
+        info.inferences[call->target] = PossibleContents::none();
```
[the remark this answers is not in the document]
kripken: This inference will constrain the forward analysis. That is, if we say "hey, we've proven nothing can be here" then the forward pass will always avoid putting anything there (because it always takes the intersection with what we say).

src/ir/possible-contents.cpp
```diff
@@ -1321,12 +1323,504 @@ struct InfoCollector
...
+        // we know only their intersection can appear here.
+        auto declared = PossibleContents::fullConeType(curr->type);
+        auto intersection = PossibleContents::fullConeType(castType);
+        intersection.intersect(declared);
+        if (intersection.isConeType()) {
+          auto intersectionType = intersection.getType();
+          if (intersectionType != curr->type) {
+            // We inferred a more refined type.
+            info.inferences[curr] =
+              PossibleContents::fullConeType(intersectionType);
```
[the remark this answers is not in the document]
kripken: Right, yeah, this full cone is just equal to `intersection`, I'll improve that.

Using `info.inferences[curr]` for `declared` is dangerous, though - we'd have potential nondeterminism if we looked at what another thread was doing, as each thread builds up inferences as we go. We'd want multiple passes for that, I think.

src/ir/possible-contents.cpp
```diff
@@ -1321,12 +1323,504 @@ struct InfoCollector
...
+  // Maps expressions to the content we inferred there. If an expression is not
+  // here then expression->type (the type in Binaryen IR) is all we have.
+  std::unordered_map<Expression*, PossibleContents> inferences;
+
+  // Phase 1: Scan to find cast parameters and calls. This operates on a single
+  // function, and is called in parallel.
+  void scan(Function* func, TNHInfo& info, const PassOptions& options);
+
+  // Phase 2: Infer contents based on what we scanned.
+  void infer();
```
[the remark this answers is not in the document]
tlively: Hmm, maybe you're right. At least we have descriptive comments describing what happens in each function.

src/ir/possible-contents.cpp
kripken: In general we'd need a labelled break here (does C++ have those?) since we are nested inside a loop here and must break out of it first in order to reach the lower break. But I realize that we can just return from the entire function anyhow... I'll do that.

src/ir/possible-contents.cpp
kripken: As an optimization, when we know a location cannot increase further, we tell the thing sending it content to stop sending anything more. That avoids wasted work.

kripken:
Thanks for the feedback! Comments should be addressed.

Good catch on missing the case of the intersection being a null. That was not only a bug in two places but also a missed optimization opportunity, which I added now with a test - when only a null is possible, we can infer the value sent to the function is a null.

test/lit/passes/gufa-tnh.wast
```diff
@@ -0,0 +1,1905 @@
...
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $caller (export "out") (param $any anyref)
+    (call $called
+      (ref.cast $A
+        (local.get $any)
+      )
+      (ref.cast $A
+        ;; This call might transfer control flow (if it throws), so we
+        ;; can't optimize before it, but the last two casts will become $B.
```
tlively, line 906: It would be really nice if we could annotate functions `nothrow` in the IR so we could entirely skip creating CFG edges for exceptions that we know won't happen.

test/lit/passes/gufa-tnh.wast
```diff
@@ -0,0 +1,1905 @@
...
+  ;; CHECK-NEXT:    (local.get $x)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (call $called
+  ;; CHECK-NEXT:   (ref.cast null $C
+  ;; CHECK-NEXT:    (local.get $x)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $caller (export "out") (param $x anyref)
```
tlively, line 1505: It would be reasonable to combine the sibling cast cases in here (and into the next function), since those are really just more input/output combinations.

test/lit/passes/gufa-tnh.wast
```diff
@@ -0,0 +1,1905 @@
...
+  ;; CHECK:      (func $caller (type $anyref_=>_none) (param $x anyref)
+  ;; CHECK-NEXT:  (call $called
+  ;; CHECK-NEXT:   (ref.cast $A
+  ;; CHECK-NEXT:    (local.get $x)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $caller (export "out") (param $x anyref)
+    (call $called
+      ;; This will be refined to a non-nullable cast.
```
tlively, line 1681: Should we test with array heap type subtyping as well? Same for functions.

test/lit/passes/gufa-tnh.wast
```diff
@@ -0,0 +1,1905 @@
...
+      )
+    )
+    (array.copy $B $B
+      (local.get $array.copy.src)
+      (i32.const 4)
+      (local.get $array.copy.dest)
+      (i32.const 5)
+      (i32.const 6)
+    )
+    (array.fill $B
```
tlively, line 1791: Do we want `array.init_data` and `array.init_elem` as well?

test/lit/passes/gufa-tnh.wast
```diff
@@ -0,0 +1,1905 @@
...
+  ;; CHECK:      (func $caller (type $funcref_=>_none) (param $x funcref)
+  ;; CHECK-NEXT:  (call_ref $A
+  ;; CHECK-NEXT:   (ref.cast $A
+  ;; CHECK-NEXT:    (local.get $x)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $caller (export "out") (param $x funcref)
+    ;; In closed world this call must trap, as the only function of the right
+    ;; type will trap. But we are in open world so we do not optimize here.
```
tlively, line 1898: Perhaps worth a TODO about analyzing publicly exposed tables and functions to see whether it is possible for a non-trapping function to have arrived here from the outside.

tlively (commented):
Looking at the tests, it occurs to me that if parameters are passed on as arguments (which seems likely in an OOP setting!), then performing this analysis repeatedly may produce better results. Is that something we can implement and test easily? If not, it would also be something we get "for free" by using the analysis framework in the future.

kripken:
> performing this analysis repeatedly may produce better results. Is that something we can implement and test easily?

Yeah, shouldn't be too hard. The design here has one oracle take another as an input, so it's pluggable in that sense. The TNHOracle would need just a few lines of code to take an oracle as input. I can try that later.

test/lit/passes/gufa-tnh.wast
```diff
@@ -0,0 +1,1905 @@
...
+  ;; CHECK-NEXT:    (local.get $z)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $called (param $x funcref) (param $no-cast funcref) (param $y funcref) (param $z funcref)
+    ;; All but the second parameter are cast here, which allows some
+    ;; optimization in the caller. Nothing significant changes here in this
+    ;; function.
+    (drop
+      (ref.cast func
```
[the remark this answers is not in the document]
kripken: That's tested on line 1088.

test/lit/passes/gufa-tnh.wast
```diff
@@ -0,0 +1,1905 @@
...
+  ;; CHECK-NEXT:  (local $f funcref)
+  ;; CHECK-NEXT:  (local.set $f
+  ;; CHECK-NEXT:   (select (result funcref)
+  ;; CHECK-NEXT:    (global.get $unknown-funcref1)
+  ;; CHECK-NEXT:    (global.get $unknown-funcref2)
+  ;; CHECK-NEXT:    (global.get $unknown-i32)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (call $called
+  ;; CHECK-NEXT:   (ref.as_func
```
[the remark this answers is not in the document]
kripken: The test does have a `ref.cast`. I think what happens is that our printing code prefers this when possible: https://github.com/WebAssembly/binaryen/blob/465ebbf470e878cac6cf7a9629c413add97f9bb9/src/passes/Print.cpp#L2122-L2140 - when we remove that this test's output will be as you were expecting.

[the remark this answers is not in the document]
kripken: (Any reason not to remove that now?)

test/lit/passes/gufa-tnh.wast
kripken: And that would update function effects, something like that?

test/lit/passes/gufa-tnh.wast
kripken: Sorry, I don't understand what you mean here?

test/lit/passes/gufa-tnh.wast
kripken: There isn't any struct-specific code in the TNHOracle logic, so I was thinking that might be a little redundant. Even the array cast felt a little unneeded to me. But I guess it could be useful for future refactorings perhaps..?

test/lit/passes/gufa-tnh.wast
kripken: Oh, for some reason I thought those create new arrays... yeah, I'll add those.

test/lit/passes/gufa-tnh.wast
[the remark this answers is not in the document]
tlively: Nope, I'll make a PR.

test/lit/passes/gufa-tnh.wast
tlively: For this code, which detects control flow by checking whether the block index in the CFG has changed, simply not ending the basic block on a non-throwing function would be enough. Elsewhere, yes, we would want to update the function effects as well (or maybe updating the function effects would be how we mark a function as nothrow?).

test/lit/passes/gufa-tnh.wast
tlively: Rather than testing the sibling cases separately, I think it would make more sense to test them alongside all these other cases (but I don't feel strongly about it).

test/lit/passes/gufa-tnh.wast
tlively: Ah, if the code isn't struct-specific, then I'm ok not doing this.

test/lit/passes/gufa-tnh.wast
tlively: You were probably thinking of `array.new_data` and `arrray.new_elem` :)

test/lit/passes/gufa-tnh.wast
kripken: Thanks, I think I see. I think separately is a little clearer, since this function is fairly long already.

tlively (approved):

PR #5851, opened 2023-08-01
Fix a fuzz bug in TypeMapper

### LINKED ISSUE #5845: TypeMerging build error fuzz bug
```wat
(module
 (rec
  (type $A (array (mut f64)))
  (type $B (func (result (ref $A) (ref $E))))
  (type $C (func (param i64 f32 i32 f64 i64) (result f64)))
  (type $D (func (result (ref $A) (ref $C) f64 f32)))
  (type $E (sub $C (func (param i64 f32 i32 f64 i64) (result f64))))
  (type $F (sub final $B (func (result (ref $A) (ref $E)))))
 )

 (func $0 (type $F) (result (ref $A) (ref $E))
  (nop)
  (unreachable)
 )
)
```
```
$ bin/wasm-opt a.wat -all --closed-world --type-merging
Fatal: Internal GlobalTypeRewriter build error: Heap type has an invalid supertype at index 4
```

cc @tlively - looks like some interesting mutual recursion going on here...

If it's useful, here is the reduced testcase before my manual reduction, which is larger:

```wat
(module
 (rec
  (type $type$5 (array (mut i8)))
  (type $type$6 (sub final (struct (field (mut (ref null $type$0))) (field (mut funcref)) (field (mut f32)) (field (mut (ref null $type$10))))))
  (type $type$4 (func (result (ref $type$2) (ref $type$3) f64 f32)))
  (type $type$7 (sub final (array i32)))
  (type $type$8 (sub final (array (mut v128))))
  (type $type$9 (array (mut f64)))
  (type $type$10 (func (param i64 f32 i32 f64 i64) (result f64)))
  (type $type$0 (sub final $type$4 (func (result (ref $type$2) (ref $type$3) f64 f32))))
  (type $type$11 (sub final (array v128)))
  (type $type$12 (sub $type$10 (func (param i64 f32 i32 f64 i64) (result f64))))
  (type $type$13 (array i16))
  (type $type$14 (func (param v128 (ref null $type$7)) (result (ref null $type$5))))
  (type $type$2 (sub final $type$9 (array (mut f64))))
  (type $type$3 (sub $type$12 (func (param i64 f32 i32 f64 i64) (result f64))))
 )
 (type $type$1 (func (result (ref $type$2) (ref $type$3) f64 f32)))
 (func $0 (type $type$0) (result (ref $type$2) (ref $type$3) f64 f32)
  (nop)
  (unreachable)
 )
)
```

tlively:
TypeMapper is a utility used to globally rewrite types, mapping some eliminated
source types into destination types they should be replaced with. This was
previously done by first rewriting all the types in the IR according to the
given mapping, then rewriting the type definitions and updating all the types in
the IR again. Not only was doing the rewriting twice inefficient, it also
introduced a subtle bug where the set of private types eligible to be rewritten
could be inconsistent because updating types in the IR could change the types of
control flow structures. The fuzzer found a case where this inconsistency caused
the type rebuilding to fail.

Fix the bug by first building the new types with the mapping applied and only
then rewriting the IR a single time.

Fixes #5845.

### DIFF
--- a/src/ir/type-updating.h
+++ b/src/ir/type-updating.h
@@ -364,11 +364,6 @@ public:
   // not appear, it is mapped to itself.
   void mapTypes(const TypeMap& oldToNewTypes);
 
-  // Builds new types after updating their contents using the hooks below and
-  // returns a map from the old types to the modified types. Used internally in
-  // update().
-  TypeMap rebuildTypes();
-
   // Subclasses can implement these methods to modify the new set of types that
   // we map to. By default, we simply copy over the types, and these functions
   // are the hooks to apply changes through. The methods receive as input the
@@ -419,6 +414,12 @@ public:
     } rewriter(wasm, updates);
   }
 
+protected:
+  // Builds new types after updating their contents using the hooks below and
+  // returns a map from the old types to the modified types. Used internally in
+  // update().
+  TypeMap rebuildTypes();
+
 private:
   TypeBuilder typeBuilder;
 
@@ -441,24 +442,24 @@ public:
   void map() {
     // Update the internals of types (struct fields, signatures, etc.) to
     // refer to the merged types.
-    auto oldToNewTypes = rebuildTypes();
+    auto newMapping = rebuildTypes();
 
     // Compose the user-provided mapping from old types to other old types with
-    // the new mapping from old types to new types. `oldToNewtypes` will become
+    // the new mapping from old types to new types. `newMapping` will become
     // a copy of `mapping` except that the destination types will be the newly
     // built types.
     for (auto& [src, dest] : mapping) {
-      if (auto it = oldToNewTypes.find(dest); it != oldToNewTypes.end()) {
-        oldToNewTypes[src] = it->second;
+      if (auto it = newMapping.find(dest); it != newMapping.end()) {
+        newMapping[src] = it->second;
       } else {
         // This mapping was to a type that was not rebuilt, perhaps because it
         // is a basic type. Just use this mapping unmodified.
-        oldToNewTypes[src] = dest;
+        newMapping[src] = dest;
       }
     }
 
     // Map the types of expressions (curr->type, etc.) to the correct new types.
-    mapTypes(oldToNewTypes);
+    mapTypes(newMapping);
   }
 
   Type getNewType(Type type) {
--- a/src/ir/type-updating.h
+++ b/src/ir/type-updating.h
@@ -366,7 +366,7 @@ public:
 
   // Builds new types after updating their contents using the hooks below and
   // returns a map from the old types to the modified types. Used internally in
-  // update()`.
+  // update().
   TypeMap rebuildTypes();
 
   // Subclasses can implement these methods to modify the new set of types that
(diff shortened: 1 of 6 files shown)


tlively:
* main
  * **PR #5851** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5851" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5851?utm_source=stack-comment).

src/ir/type-updating.h
```diff
@@ -364,6 +364,11 @@ class GlobalTypeRewriter {
   // not appear, it is mapped to itself.
   void mapTypes(const TypeMap& oldToNewTypes);
 
+  // Builds new types after updating their contents using the hooks below and
+  // returns a map from the old types to the modified types. Used internally in
+  // update()`.
```
kripken, line 369:
```suggestion
  // update().
```

src/ir/type-updating.h
```diff
@@ -434,13 +439,23 @@ class TypeMapper : public GlobalTypeRewriter {
...
-    // Map the types of expressions (curr->type, etc.) to their merged
-    // types.
-    mapTypes(mapping);
-
     // Update the internals of types (struct fields, signatures, etc.) to
     // refer to the merged types.
-    update();
+    auto oldToNewTypes = rebuildTypes();
+
+    // Compose the given mapping from old types to old types with the new
```
kripken, line 446: "old types to old types"?

src/ir/type-updating.h
```diff
@@ -434,13 +439,23 @@ class TypeMapper : public GlobalTypeRewriter {
...
+    // mapping from old types to new types.
```
kripken, line 447: Maybe add some clarification as to how the composition works between `oldToNewTypes` and `mapping`? Something like `oldToNewTypes[mapping[x]] = y` or such.

test/lit/passes/type-merging.wast
```diff
@@ -918,6 +918,41 @@
   )
 )
 
+;; Regression test for bug where we updated module types before building the new
```
kripken, line 921:
```suggestion
;; Regression test for a bug where we updated module types before building the new
```

src/ir/type-updating.h
tlively: Yeah, the mapping the user provided. Will try to clarify.

src/ir/type-updating.h
```diff
@@ -364,6 +364,11 @@ class GlobalTypeRewriter {
   // not appear, it is mapped to itself.
   void mapTypes(const TypeMap& oldToNewTypes);
 
+  // Builds new types after updating their contents using the hooks below and
+  // returns a map from the old types to the modified types. Used internally in
+  // update().
+  TypeMap rebuildTypes();
```
kripken, line 370: If this is only useful internally, could it be private and placed down below?

src/ir/type-updating.h
```diff
@@ -434,13 +439,26 @@ class TypeMapper : public GlobalTypeRewriter {
...
-    update();
+    auto oldToNewTypes = rebuildTypes();
+
+    // Compose the user-provided mapping from old types to other old types with
+    // the new mapping from old types to new types. `oldToNewtypes` will become
+    // a copy of `mapping` except that the destination types will be the newly
+    // built types.
+    for (auto& [src, dest] : mapping) {
+      if (auto it = oldToNewTypes.find(dest); it != oldToNewTypes.end()) {
+        oldToNewTypes[src] = it->second;
```
kripken, line 452: I'm a little confused here. `mapping` is the user provided mapping of types existing before this operation, and `oldToNewTypes` maps old types to rebuilt types. The operation below can make `oldToNewTypes[..]` be equal to a type existing before this operation (since it copies `dest` from `mapping`), which sounds wrong? I'd expect something like
```suggestion
        oldToNewTypes[src] = oldToNewTypes[it->second];
```
That is, `mapping` tells us what original type we want, but we need to apply `oldToNewTypes` on the right in order to get the the rebuilt version of that type. Is that wrong?

Side note, could `oldToNewTypes` here be renamed to `rebuiltTypes` or such? (old/new has more than one meaning here.)

src/ir/type-updating.h
tlively: It's also used in the subclass, but I can make it protected at least.

src/ir/type-updating.h
tlively: Yeah, renaming `oldToNewTypes` to `newMapping` might make this even clearer, since it ends up taking the place of the original `mapping`.

tlively: Regarding the code you would expect, since `it` is result of looking up `dest` in `oldToNewTypes`, we're already doing the right-hand-side `oldToNewTypes` lookup you are expecting to see. Doing an additional lookup of `oldToNewTypes[it->second]` wouldn't make sense because `it->second` is already a newly built type.

kripken: Ah, thanks, I was indeed confused here.

kripken (approved):
lgtm with fuzzing

PR #5852, opened 2023-08-02
[Wasm GC] Stop printing deprecated cast etc. instructions

tlively:
Stop printing `ref.as_i31`, `br_on_func`, etc. because they have been removed
from the spec and are no longer supported by V8. #5614 already made this change
for the binary format. Like that PR, leave reading unmodified in case someone is
still using these instructions (even though they are useless). They will be
fully removed in a future PR as we finalize things ahead of standardizing
WasmGC.

### DIFF
--- a/src/passes/Print.cpp
+++ b/src/passes/Print.cpp
@@ -2092,20 +2092,6 @@ struct PrintExpressionContents
     printHeapType(o, curr->target->type.getHeapType(), wasm);
   }
   void visitRefTest(RefTest* curr) {
-    // TODO: These instructions are deprecated. Remove them.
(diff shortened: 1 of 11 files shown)


tlively:
* main
  * **PR #5852** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5852" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5852?utm_source=stack-comment).

kripken (approved):
lgtm if this passes fuzzing

PR #5854, opened 2023-08-02
[NFC] Remove some unneeded cruft from PossibleContents::intersect()

kripken:
In order of appearance in this diff:

* Use `heapType` rather than `otherHeapType` when either will do, for brevity.
* We checked `isNull` after checking for `isLiteral` (line 173), but every null is a literal.
* Calling `setNoneOrNull` simplifies some repetitive code.
* We checked `isLiteral` yet again lower down.


tlively (approved):
It would be nice to have some sort of coverage dashboard to making finding these things easier 🤔

PR #5857, opened 2023-08-04
Fix LinearExecutionWalker on calls

kripken:
Calls were simply not handled there, so we could think we were still in the same
basic block when we were not, affecting various passes (but somehow this went
unnoticed until the TNHOracle #5850 ran on some particular Java code).

One existing test was affected, and two new tests are added: one for TNHOracle
where I detected this, and one in OptimizeCasts which is perhaps a simpler way
to see the problem.

All the cases but the TNH one, however, do not need this fix for correctness
since they actually don't care if a call would throw. As a TODO, we should find a
way to undo this minor regression.


kripken:
The changes to wasm2js tests here worry me that this regression might be noticeable, so I'll try to think of a better option before landing this.

kripken:
I think I have an idea for how to avoid this regression, but it only makes sense to add after this lands.

To reduce the regression here, though, I added a small fix to actually set `Module` inside SimplifyLocals. With that, this PR has less of a downside if EH is not enabled, which is the common case for wasm2js for example. Though, in our test suite we do enable it, so the benefit is not obvious here.

src/passes/SimplifyLocals.cpp
```diff
@@ -1016,6 +1016,8 @@ struct SimplifyLocals
       }
 
       void visitLocalSet(LocalSet* curr) {
+        auto* module = this->getModule();
```
tlively, line 1019: Is the `this->` necessary here and below?

kripken:
Oh actually we have `-all --disable-exception-handling` so my last commit *does* remove the regression in the test suite here, nice.

PR #5860, opened 2023-08-07
LinearExecutionTraversal: Add an option to connect adjacent code, use in SimplifyLocals

kripken:
This addresses most of the minor regression from the correctness fix in #5857.
That PR makes us consider calls as branching, but in some cases it is ok to
ignore that branching (see the comment in the code), which this PR allows as
an option.

This undoes one test change from that PR (showing it undoes the regression).
More tests are added to cover this specifically as well.

### DIFF
--- a/src/ir/linear-execution.h
+++ b/src/ir/linear-execution.h
@@ -44,6 +44,39 @@ struct LinearExecutionWalker : public PostWalker<SubType, VisitorType> {
     self->noteNonLinear(*currp);
   }
 
+  // Optionally, we can connect adjacent basic blocks. "Adjacent" here means
+  // that the first branches to the second, and that there is no other code in
+  // between them. As a result, the first dominates the second, but it might not
+  // reach it.
+  //
+  // For example, a call may branch if exceptions are enabled, but if this
+  // option is flipped on then we will *not* call doNoteNonLinear on the call:
+  //
+  //  ..A..
+  //  call();
+  //  ..B..
+  //
+  // As a result, we'd consider A and B to be together. Another example is an
+  // if:
+  //
+  //  ..A..
+  //  if
+  //    ..B..
+  //  else
+  //    ..C..
+  //  end
+  //
+  // Here we will connect A and B, but *not* A and C (they are not adjacent) or
+  // B and C (they do not branch to each other).
+  //
+  // As the if case shows, this can be useful for cases where we want to look at
+  // dominated blocks with their dominator, but it only handles the trivial
+  // adjacent cases of such dominance. Passes should generally uses a full CFG
+  // and dominator tree for this, but this option does help some very common
+  // cases (calls, if without an else) and it has very low overhead (we still
+  // only do a simple postorder walk on the IR, no CFG is constructed, etc.).
+  bool connectAdjacentBlocks = false;
+
   static void scan(SubType* self, Expression** currp) {
     Expression* curr = *currp;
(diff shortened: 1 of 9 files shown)


test/lit/passes/type-ssa_and_merging.wast
```diff
@@ -22,6 +22,10 @@
   ;; NOP:      (type $none_=>_i32 (func (result i32)))
 
   ;; NOP:      (import "a" "b" (func $import (type $none_=>_i32) (result i32)))
+  ;; YES:       (type $A$2 (sub $A (struct )))
+
+  ;; YES:       (type $A$1 (sub $A (struct )))
+
```
kripken, line 28: This is unrelated, and is a change I've noticed for a while now but didn't have time to look into. I think perhaps we just updated this file manually and not with the auto updater. This is what the auto updater adds for it.

src/ir/linear-execution.h
```diff
@@ -44,14 +44,49 @@ struct LinearExecutionWalker : public PostWalker<SubType, VisitorType> {
     self->noteNonLinear(*currp);
   }
 
+  // Optionally, we can connect adjacent basic blocks. "Adjacent" here means
```
aheejin, line 47: This can be basically extended to any dominator-dominatee relationship, but given that this is a linear execution walker, we limit the scope to the adjacent blocks, correct?

I thought about whether "consecutive" sounds better, but usually this word doesn't contain the meaning that the first can branch out or return, so I'm not sure.

src/passes/SimplifyLocals.cpp
```diff
@@ -998,6 +998,11 @@ struct SimplifyLocals
     struct EquivalentOptimizer
       : public LinearExecutionWalker<EquivalentOptimizer> {
 
+      // It is ok to look at adjacent blocks together, as if a later part of a
+      // block is not reached that is fine - changes we make there would not be
+      // reached in that case.
+      bool connectAdjacentBlocks = true;
```
aheejin, line 1004: Are the cases where we need this the same as where we need `bool ignoreBranchesOutsideOfFunc = true;` for `cfg-traversal.h`?

aheejin (approved):

src/ir/linear-execution.h
kripken: Yes, exactly, it's the simple case during linear execution when we see dominator-dominatee.

src/passes/SimplifyLocals.cpp
kripken: I think it's a little different. I was actually thinking of saying in a comment "this is like ignoreBranchesOutsideOfFunc" but it seemed like it could be more confusing than helpful, though I'm not sure...

The main difference is there we ignore branches outside of the function. We never ignore a branch inside it. So we use that in cases where we just care about state in locals, for example, that goes away when we branch out of the function. We can use that for local coalescing for that reason. But we couldn't use this PR for local coalescing since we need to be aware of every branch inside the function, or else we could miss some interferences.

aheejin: Isn't this PR also about excluding branches out of a function? Ah, I guess it is branch within a function in case of `if`..?

kripken: Yes, exactly, an if would be inside. I think I can also handle `br_if` etc. but I'll leave that for later.

PR #5861, opened 2023-08-08
Remove legacy WasmGC instructions

tlively:
Remove old, experimental instructions and type encodings that will not be
shipped as part of WasmGC. Updating the encodings and text format to match the
final spec is left as future work.


tlively:
* main
  * **PR #5861** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5861" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5861?utm_source=stack-comment).

test/heap-types.wast
kripken, line 1: Do we have enough other coverage of all these opcodes?

Also if you want to delete this, please delete the other adjoining files with the same name + suffix.

test/lit/passes/optimize-instructions-gc-tnh.wast
```diff
@@ -156,39 +176,21 @@
...
-  ;; NO_TNH-NEXT: )
-  (func $ref.is_func (param $a funcref) (result i32)
-    ;; The check must succeed. We can return 1 here, and drop the rest, with or
-    ;; without TNH (in particular, TNH should not just remove the cast but not
-    ;; return a 1).
-    (ref.is_func
-      (ref.as_func
-        (local.get $a)
-      )
-    )
```
kripken, line 191: I don't see a replacement for this test (something with an inferred 1 in the output) immediately above or before it. Is there one?

test/lit/passes/remove-unused-brs-gc.wast
```diff
@@ -11,57 +11,6 @@
...
-     ;; An i31 is provided here, and so we will not branch.
-     (br_on_non_i31 $any
-      (i31.new (i32.const 0))
-     )
-    )
-    (ref.null any)
-   )
-  )
- )
-
```
kripken, line 64: Do these have replacements?

kripken (commented):
Not finished reading the tests, but here are some comments so far.

test/heap-types.wast
tlively: Mostly, but I can work on importing the spec tests first to be absolutely sure.

test/lit/passes/optimize-instructions-gc-tnh.wast
tlively: Will add a test back in here.

test/lit/passes/remove-unused-brs-gc.wast
tlively: `br_on_cast` and `br_on_cast_fail` are already well tested below, so these don't need replacements.

test/heap-types.wast
tlively: Ok, I just checked more thoroughly and we definitely do already have coverage for everything here, so I don't think this should be blocked on anything else.

test/lit/passes/remove-unused-brs_all-features.wast
```diff
@@ -115,111 +113,4 @@
...
-   )
-  )
-  (call $log (i32.const 6))
-  (drop
-   (block $non-null (result (ref func))
-    ;; a non-null reference is not null, and the br is always taken
-    (br_on_non_null $non-null (ref.func $br_on-to-br))
-    (call $log (i32.const 7))
-    (ref.func $br_on-to-br)
-   )
```
kripken, line 222: `br_on_func|i31` are legacy and I get why they are removed, but why remove `null|non_null`?

test/lit/passes/vacuum-gc.wast
```diff
@@ -18,21 +18,21 @@
...
+    ;; Without -tnh, we must assume all casts can have a trap effect, and so
     ;; we cannot remove anything here.
     (drop
       (ref.as_non_null
         (local.get $x)
       )
     )
     (drop
-      (ref.as_i31
+      (ref.cast null i31
```
kripken, line 35: Why change the nullability?
```suggestion
      (ref.cast i31
```

kripken: (this also happens in at least one other test below, so maybe I'm missing something...)

tlively: I don't think I remembered whether the result of `ref.as_i31` was supposed to be nullable or not, so I think I just took a guess 😅 It doesn't seem to have affected the test output unexpectedly, so I'm not sure it's worth changing. WDYT?

kripken (commented):
lgtm aside from https://github.com/WebAssembly/binaryen/pull/5861/files#r1288956276

test/lit/passes/remove-unused-brs_all-features.wast
tlively: I've added a test for `br_on_non_null` back in as well as a new test for `br_on_null`. It turns out we don't optimize `br_on_null` yet, so I left a TODO for that.

test/lit/passes/remove-unused-brs_all-features.wast
```diff
@@ -116,110 +118,38 @@
...
- ;; CHECK-NEXT:   )
- ;; CHECK-NEXT:  )
  ;; CHECK-NEXT: )
- (func $br_on-to-br (param $func (ref func))
-  (call $log (i32.const 0))
+ (func $br_on_null
   (block $null
-   ;; a non-null reference is not null, and the br is never taken
+   ;; a null reference to bottom is definitely null, and the br is always taken
+   ;; TODO: Optimize this.
```
kripken, line 133: Please add a test for the reverse case, which is optimized and which we had a test for before, where the input is non-null so the br is never taken.

test/lit/passes/remove-unused-brs_all-features.wast
kripken: I think we could use some more tests here. The i31 and func tests were placeholders for general `br_on` operations. How about aside from `br_on_null|non_null` we also replace the old tests with tests of `br_on_cast` which is definitely not or definitely taken?

kripken: (or do those exist already elsewhere?)

tlively: These are already thoroughly tested in test/lit/passes/remove-unused-brs-gc.wast

test/lit/passes/vacuum-gc.wast
tlively: Just for some rationale here: In the future when I update the text format, this will be able to be `ref.cast i31ref` rather than `ref.cast (ref i31)`, so it's a tiny simplification.

test/lit/passes/remove-unused-brs_all-features.wast
tlively: Updated. It seems we don't yet optimize in the known-null case for either br_on_null or br_on_non_null.

kripken (approved):

PR #5864, opened 2023-08-08
Add --js-call-indirect option.

juj:
Currently Chrome and Firefox are unhelpful in diagnosing `call_indirect` errors, but when one occurs, a blank error print

```
RuntimeError: null function or function signature mismatch
```
is logged.

This challenge was discussed in https://github.com/emscripten-core/emscripten/issues/16126 and there were workarounds presented there:

- one can build with `-Wbad-function-cast -Wcast-function-type` or even `-Werror=bad-function-cast -Werror=cast-function-type`. However, in our huge codebase, we get tons of diagnostics from these in 3rd party code paths that are never exercised, or are false positives, so it does not help one to target an immediate bug scenario - and would not help see into issues where the function pointers to call_indirect are either null, or corrupted.
- one can build with UBSAN: `-fsanitize=undefined`. However unfortunately our codebase is too large, and we get a `local count too large` error.
- in Chromium bug entry https://bugs.chromium.org/p/chromium/issues/detail?id=1350384# , Chrome authors did help Chrome's debugger to do better prints, but unfortunately our codebase is too large that Chrome dies attempting to debug it.

Hence I would like to improve the debuggability of `call_indirect`s from the toolchain side to help figure out when the function signature does not match, or when the function pointer is bad. 

To do that, my idea was to try to route all `call_indirect`s instead over to `js_call_indirect_sig` functions on the JS import side, which could do the calls by referring to the WebAssembly.Table (`wasmTable` object in Emscripten JS runtime). I got this far with that attempt, but then I run into these kinds of errors:

```
[wasm-validator error in function dynCall_jiji] unexpected false: non-final block elements returning a value must be drop()ed (binaryen's autodrop option might help you), on
(block (result i64)
 (call $js_call_indirect_jiji
  (local.get $1)
  (local.get $2)
  (local.get $3)
  (local.get $0)
 )
 (call_indirect (type $i32_i64_i32_=>_i64)
  (local.get $1)
  (local.get $2)
  (local.get $3)
  (local.get $0)
 )
)
(on index 0:
0000026E7512C988
), type: i64
[wasm-validator error in function dynCall_jii] unexpected false: non-final block elements returning a value must be drop()ed (binaryen's autodrop option might help you), on
(block (result i64)
 (call $js_call_indirect_jii
  (local.get $1)
  (local.get $2)
  (local.get $0)
 )
 (call_indirect (type $i32_i32_=>_i64)
  (local.get $1)
  (local.get $2)
  (local.get $0)
 )
)
(on index 0:
0000026E7514CC30
), type: i64
Fatal: error after opts
```
It looks like my attempt to use the `replaceCurrent()` function to replace the `call_indirect` with an import to `js_call_indirect_sig` function is not actually replacing the call, but rather prepending. Here I get stumped and am not sure what to do to remove the `call_indirect` altogether.

Any tips would be welcome!


src/passes/JsCallIndirect.cpp
```diff
@@ -0,0 +1,107 @@
...
+    Builder builder(*getModule());
+    std::vector<Expression*> callArgs;
+    callArgs.insert(
+      callArgs.end(), curr->operands.begin(), curr->operands.end());
+    callArgs.push_back(curr->target);
+    auto* block = builder.makeSequence(
+      builder.makeCall(getCallIndirectFromArgs(curr), callArgs, curr->type),
+      curr);
+
+    replaceCurrent(block);
```
kripken, line 69: The problem, I think, is that you append `curr` at the end here. This should fix it:
```suggestion
    replaceCurrent(
      builder.makeCall(getCallIndirectFromArgs(curr), callArgs, curr->type));
```

src/passes/JsCallIndirect.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+typedef std::unordered_map<std::string, std::unique_ptr<Function>>
+  JsCallIndirectMap;
+
+static std::string getCallIndirectFromArgs(CallIndirect* ci) {
+  std::string jsCallIndirectStr = std::string("js_call_indirect_");
+  jsCallIndirectStr += getSig(ci->type);
+  for (Expression*& operand : ci->operands) {
+    jsCallIndirectStr += getSig(operand->type);
+  }
```
kripken, line 47: I think you can do `return "js_call_indirect_" + getSigFromStructs(ci->type, ci->operands)` here but I'm not sure.

src/passes/JsCallIndirect.cpp
juj: Doh thanks! I was too fixated on looking at the wrong line. That indeed helped it.

src/passes/JsCallIndirect.cpp
juj: Nice, that worked.

PR #5865, opened 2023-08-08
SimplifyGlobals: Connect adjacent blocks in LinearExecutionWalker

kripken:
Followup to #5860, this does the same for SimplifyGlobals as for SimplifyLocals.

As there, this is valid because it's ok if we branch away. This part of the pass
applies a global value to a global.get based on a dominating global.set, so any
dominance is good enough for us.


test/lit/passes/simplify-globals-dominance.wast
```diff
@@ -0,0 +1,54 @@
...
+    )
+    (if
+      (i32.const 0)
+      (block
+        ;; This is dominated by the set, so we can apply 10 here.
+        (drop
+          (global.get $global)
+        )
+        (call $test)
+        ;; This is after a call, so we do nothing.
```
aheejin, line 43: Didn't we optimize across calls in #5860?

kripken: Ah, sorry, this might be confusing. This pass specifically cares about calls since the global might be set to another value inside it, so it has logic for that. I'll improve the comment.

aheejin (approved):

PR #5871, opened 2023-08-10
Add a Max Bits Analysis

zm2he:
This change introduces a prototype max bits analysis for values in the wasm value stack using the abstract interpretation static analysis framework. The implementation mirrors the existing `getMaxBits` function in `bits.h` but does not cover local indices and other values outside of the wasm value stack.

### DIFF
--- a/src/analysis/bits-bounds-lattice.h
+++ b/src/analysis/bits-bounds-lattice.h
@@ -1,0 +1,144 @@
+#ifndef wasm_analysis_bits_bounds_lattice_h
+#define wasm_analysis_bits_bounds_lattice_h
+
+#include <optional>
+
+#include "lattice.h"
+#include "wasm.h"
+
+namespace wasm::analysis {
+
+struct MaxBitsLattice {
+  enum LatticeState { BOTTOM, REGULAR_VALUE, TOP };
+
+  struct Element {
+    Index upperBound = 0;
+    std::optional<Literal> constVal;
+    LatticeState state = BOTTOM;
+
+    bool isTop() const { return state == TOP; }
+
+    bool isBottom() const { return state == BOTTOM; }
+
+    void setUpperBound(Index upperBound) {
+      if (state != TOP) {
+        state = REGULAR_VALUE;
+        this->upperBound = upperBound;
+        constVal.reset();
+      }
+    }
+
+    void setUpperBound(Index upperBound, Literal constVal) {
+      if (state != TOP) {
+        state = REGULAR_VALUE;
+        this->upperBound = upperBound;
+        this->constVal = constVal;
+      }
+    }
+
+    void setTop() {
+      state = TOP;
+      upperBound = UINT32_MAX;
+      constVal.reset();
+    }
+
+    bool makeLeastUpperBound(const Element& other) {
+      if (other.isBottom() || isTop()) {
+        return false;
+      } else if (other.isTop()) {
+        setTop();
+        return true;
+      } else if (isBottom()) {
+        if (other.constVal.has_value()) {
+          setUpperBound(other.upperBound, other.constVal.value());
+        } else {
+          setUpperBound(other.upperBound);
+        }
+        return true;
+      }
+
+      if (upperBound < other.upperBound) {
+        if (other.constVal.has_value()) {
+          setUpperBound(other.upperBound, other.constVal.value());
+        } else {
+          setUpperBound(other.upperBound);
+        }
+        return true;
+      } else if (upperBound == other.upperBound && other.constVal.has_value()) {
+        if (constVal.has_value()) {
+          constVal.reset();
+        } else {
+          constVal = other.constVal.value();
+        }
+        return true;
+      }
+
+      return false;
+    }
+
+    void print(std::ostream& os) {
+      if (state == TOP) {
+        os << "TOP";
+      } else if (state == BOTTOM) {
+        os << "BOTTOM";
+      } else {
+        os << upperBound;
+        if (constVal.has_value()) {
+          os << " " << constVal.value();
+        }
+      }
+    }
+
+    friend MaxBitsLattice;
+  };
+
+  LatticeComparison compare(const Element& left, const Element& right) {
+    if (left.isTop()) {
+      if (right.isTop()) {
+        return LatticeComparison::EQUAL;
+      } else {
+        return LatticeComparison::GREATER;
+      }
+    } else if (right.isTop()) {
+      return LatticeComparison::LESS;
+    } else if (left.isBottom()) {
+      if (right.isBottom()) {
+        return LatticeComparison::EQUAL;
+      } else {
+        return LatticeComparison::LESS;
+      }
+    } else if (right.isBottom()) {
+      return LatticeComparison::GREATER;
+    }
+
+    if (left.upperBound > right.upperBound) {
+      return LatticeComparison::GREATER;
+    } else if (left.upperBound < right.upperBound) {
+      return LatticeComparison::LESS;
+    } else {
+      if (left.constVal.has_value()) {
+        if (right.constVal.has_value()) {
+          if (left.constVal.value() == right.constVal.value()) {
+            return LatticeComparison::EQUAL;
+          } else {
+            return LatticeComparison::NO_RELATION;
+          }
+        } else {
+          return LatticeComparison::LESS;
+        }
+      } else {
+        if (right.constVal.has_value()) {
+          return LatticeComparison::GREATER;
+        } else {
+          return LatticeComparison::EQUAL;
+        }
+      }
+    }
+  }
+
+  Element getBottom() { return Element{}; }
+};
+
+} // namespace wasm::analysis
+
+#endif // wasm_analysis_bits_bounds_lattice_h
--- a/src/analysis/bits-bounds-transfer-function.h
+++ b/src/analysis/bits-bounds-transfer-function.h
@@ -1,0 +1,367 @@
+#ifndef wasm_analysis_bits_bounds_transfer_function_h
+#define wasm_analysis_bits_bounds_transfer_function_H
+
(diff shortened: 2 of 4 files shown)


src/analysis/bits-bounds-lattice.h
tlively, line 1: Please add the license comment to the top of this file.

src/analysis/bits-bounds-lattice.h
```diff
@@ -0,0 +1,144 @@
...
+
+namespace wasm::analysis {
+
+struct MaxBitsLattice {
+  enum LatticeState { BOTTOM, REGULAR_VALUE, TOP };
+
+  struct Element {
+    Index upperBound = 0;
+    std::optional<Literal> constVal;
+    LatticeState state = BOTTOM;
```
tlively, line 17: This might be a good opportunity to use a `std::variant` so you don't need to do so much work to manage the valid representations.

src/analysis/bits-bounds-lattice.h
```diff
@@ -0,0 +1,144 @@
...
+
+    bool isTop() const { return state == TOP; }
+
+    bool isBottom() const { return state == BOTTOM; }
+
+    void setUpperBound(Index upperBound) {
+      if (state != TOP) {
+        state = REGULAR_VALUE;
+        this->upperBound = upperBound;
```
tlively, line 26: What if this was previously a literal and now it's being joined with a max bits index, but the constant actually used more bits than the new max bits index?

src/analysis/bits-bounds-lattice.h
```diff
@@ -0,0 +1,144 @@
...
+        constVal.reset();
+      }
+    }
+
+    void setUpperBound(Index upperBound, Literal constVal) {
+      if (state != TOP) {
+        state = REGULAR_VALUE;
+        this->upperBound = upperBound;
+        this->constVal = constVal;
```
tlively, line 35: I'm surprised that both `upperBound` and `constVal` can be meaningful at once. Do you have an invariant that if both are present, then they must match? For example, is it the case that if `constVal` is `Literal(5)`, then `upperBound` must be 2 because the max bit of 5 is 2? If so, it might be simpler to only store the literal in that case and recalculate its max bits on the fly when necessary.

src/analysis/bits-bounds-lattice.h
```diff
@@ -0,0 +1,144 @@
...
+      if (upperBound < other.upperBound) {
+        if (other.constVal.has_value()) {
+          setUpperBound(other.upperBound, other.constVal.value());
+        } else {
+          setUpperBound(other.upperBound);
+        }
+        return true;
+      } else if (upperBound == other.upperBound && other.constVal.has_value()) {
+        if (constVal.has_value()) {
+          constVal.reset();
```
tlively, line 69: What if the two constants are the same? Then it should remain uncleared, right? We should add some checkers to the fuzzer to catch this kind of unnecessary loss of precision in LUB operations.

src/analysis/bits-bounds-lattice.h
```diff
@@ -0,0 +1,144 @@
...
+      return LatticeComparison::LESS;
+    } else if (left.isBottom()) {
+      if (right.isBottom()) {
+        return LatticeComparison::EQUAL;
+      } else {
+        return LatticeComparison::LESS;
+      }
+    } else if (right.isBottom()) {
+      return LatticeComparison::GREATER;
+    }
```
tlively, line 112: This could all be simplified to this:

```suggestion
    if (left.state != right.state) {
      return left.state < right.state ? LatticeComparison::LESS : LatticeComparison::GREATER;
    }
    if (left.isTop() || left.isBottom()) {
      return LatticeComparison::EQUAL;
    }
```

src/analysis/bits-bounds-lattice.h
```diff
@@ -0,0 +1,144 @@
...
+    } else {
+      if (left.constVal.has_value()) {
+        if (right.constVal.has_value()) {
+          if (left.constVal.value() == right.constVal.value()) {
+            return LatticeComparison::EQUAL;
+          } else {
+            return LatticeComparison::NO_RELATION;
+          }
+        } else {
+          return LatticeComparison::LESS;
```
tlively, line 127: If one side has a constant and the other does not, the correct thing to do is to determine the max bits of the constant and compare that to the max bits of the other side. `literal(8) < maxBits(3)`, but `literal(8) <> maxBits(2)` because we want `lub(literal(8), maxBits(2)) ==maxBits(3)`.

We've arrived at the part of the project where the Galois connections are actually relevant!

src/analysis/bits-bounds-transfer-function.h
```diff
@@ -0,0 +1,367 @@
...
+    MaxBitsLattice::Element left = currState->pop();
+
+    MaxBitsLattice::Element currElement = bitsLattice.getBottom();
+
+    bool addInformation = true;
+    switch (curr->op) {
+      case RotLInt32:
+      case RotRInt32:
+      case SubInt32: {
+        currElement.setUpperBound(32);
```
tlively, line 62: Looks like there's room for improvement here if the right side is known to be 0. Is that worth improving now, or are we trying to match 1:1 with the existing getMaxBits as much as possible?

src/analysis/bits-bounds-transfer-function.h
```diff
@@ -0,0 +1,367 @@
...
+      }
+      case DivSInt32: {
+        int32_t maxBitsLeft = left.upperBound;
+        int32_t maxBitsRight = right.upperBound;
+        if (maxBitsLeft == 32 || maxBitsRight == 32) {
+          currElement.setUpperBound(32);
+        } else {
+          currElement.setUpperBound(
+            std::max(0, maxBitsLeft - maxBitsRight + 1));
+        }
```
tlively, line 83: Why is this correct? Even if `rightMaxBits` is large, its value at runtime could still be 1, in which case the max bits of the result should be the same as `maxBitsLeft`, right? Same question for unsigned division below.

src/analysis/bits-bounds-transfer-function.h
```diff
@@ -0,0 +1,367 @@
...
+
+    if (collectingResults && addInformation) {
+      exprMaxBounds[curr] = currElement.upperBound;
+    }
+
+    currState->push(std::move(currElement));
+  }
+
+  void visitUnary(Unary* curr) {
+    MaxBitsLattice::Element unaryVal = currState->pop();
```
tlively, line 283:
```suggestion
    MaxBitsLattice::Element val = currState->pop();
```

No need to disambiguate between unary or anything else given that we're in `visitUnary`.

src/analysis/bits-bounds-transfer-function.h
```diff
@@ -0,0 +1,367 @@
...
+          unaryVal.upperBound >= 8 ? Index(64) : unaryVal.upperBound);
+        break;
+      }
+      case ExtendS16Int64: {
+        currElement.setUpperBound(
+          unaryVal.upperBound >= 16 ? Index(64) : unaryVal.upperBound);
+        break;
+      }
+      case ExtendS32Int64:
+      case ExtendSInt32: {
```
tlively, line 326: What's the difference between these two instructions? I'm surprised they get the same logic.

src/analysis/bits-bounds-transfer-function.h
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    currState->push(std::move(currElement));
+  }
+
+  void visitLocalSet(LocalSet* curr) {
+    MaxBitsLattice::Element val = currState->pop();
+
+    if (collectingResults && !val.isTop()) {
+      exprMaxBounds[curr] = val.upperBound;
```
tlively, line 347: This should only be the case if the local set is a tee, right?

tlively (commented):
Ideally it would be nice to have thorough unit tests for the transfer function, but maybe in the interest of time we can just add a few with a TODO to add more.

src/analysis/bits-bounds-transfer-function.h
zm2he: They get the same logic in `bits.h`.

src/analysis/bits-bounds-lattice.h
```diff
@@ -0,0 +1,144 @@
...
+#include "lattice.h"
+#include "wasm.h"
+
+namespace wasm::analysis {
+
+struct MaxBitsLattice {
+  enum LatticeState { BOTTOM, REGULAR_VALUE, TOP };
+
+  struct Element {
+    Index upperBound = 0;
```
kripken, line 15: We discussed various options here. From the name, I can't tell which one ended up used? Seems like a good place for a comment.

I also don't have a good guess about what "constVal" is right below it.

src/analysis/bits-bounds-transfer-function.h
zm2he: This is what's implementedin `bits.h`. Added a TODO there.

src/analysis/bits-bounds-lattice.h
zm2he: Added a comment explaining the values now. 

The two options we discussed are using a max bits range (and potentially a min bits range in the future), or using a literal value all the time. For cases where we don't know the literal value, we could set all bits right of the max bit to 1. We can "or" these literal bits together when joining or increasing the max bits count.

This implementation follows the max bits range scheme mostly. I added an option to put a literal as a value instead of a max bits range in the case of constants. This is because in `bits.h`, some of the calculations involve operations directly on constant values.

However, I think some of these operations like the ceil log 2 can probably be replaced with a max bits bound, so maybe we can even get rid of the literal value.

src/analysis/bits-bounds-transfer-function.h
zm2he: Added a TODO.

PR #5876, opened 2023-08-12
Improve cast optimizations

tlively:
Simplify the optimization of ref.cast and ref.test in OptimizeInstructions by
moving the loop that examines fallthrough values one at a time out to a shared
function in properties.h. Also simplify ref.cast optimization by analyzing the
cast result in just one place.

In addition to simplifying the code, also make the cast optimizations more
powerful by analyzing the nullability and heap type of the cast value
independently, resulting in a potentially more precise analysis of the cast
behavior. Also improve optimization power by considering fallthrough values when
optimizing the SuccessOnlyIfNonNull case.

### DIFF
--- a/src/ir/properties.h
+++ b/src/ir/properties.h
@@ -388,9 +388,18 @@ inline Type getFallthroughType(Expression* curr,
     if (type.isNullable() && next->type.isNonNullable()) {
       type = Type(type.getHeapType(), NonNullable);
     }
-    if (next->type.getHeapType() != type.getHeapType() &&
-        HeapType::isSubType(next->type.getHeapType(), type.getHeapType())) {
-      type = Type(next->type.getHeapType(), type.getNullability());
+    auto currType = type.getHeapType();
+    auto nextType = next->type.getHeapType();
+    if (nextType != currType) {
+      if (HeapType::isSubType(nextType, currType)) {
+        // We found a more refined type.
+        type = Type(nextType, type.getNullability());
+      } else if (!HeapType::isSubType(currType, nextType)) {
+        // The next type is neither more refined nor less refined; it is
+        // unrelated. No non-null value can be typed with two unrelated heap
+        // types types, so we can refine all the way down to the bottom type.
+        type = Type(nextType.getBottom(), type.getNullability());
+      }
     }
     curr = next;
   }
--- a/src/ir/properties.h
+++ b/src/ir/properties.h
@@ -379,27 +379,9 @@ inline Type getFallthroughType(Expression* curr,
     if (next == curr) {
       return type;
     }
-    if (next->type == Type::unreachable) {
-      return Type::unreachable;
-    }
-    assert(next->type.isRef());
-    // Improve nullability and heap type independently. This can lead to more
-    // precise results than if we checked subtyping on the type as a whole.
-    if (type.isNullable() && next->type.isNonNullable()) {
-      type = Type(type.getHeapType(), NonNullable);
-    }
-    auto currType = type.getHeapType();
-    auto nextType = next->type.getHeapType();
-    if (nextType != currType) {
-      if (HeapType::isSubType(nextType, currType)) {
-        // We found a more refined type.
-        type = Type(nextType, type.getNullability());
-      } else if (!HeapType::isSubType(currType, nextType)) {
-        // The next type is neither more refined nor less refined; it is
-        // unrelated. No non-null value can be typed with two unrelated heap
-        // types types, so we can refine all the way down to the bottom type.
-        type = Type(nextType.getBottom(), type.getNullability());
-      }
+    type = Type::getGreatestLowerBound(type, next->type);
+    if (type == Type::unreachable) {
+      return type;
     }
     curr = next;
   }
--- a/src/ir/properties.h
+++ b/src/ir/properties.h
@@ -362,6 +362,83 @@ inline Expression* getFallthrough(
   }
 }
 
+// Look at all the intermediate fallthrough expressions and return the most
+// precise type we know this value will have.
+inline Type getFallthroughType(Expression* curr,
+                               const PassOptions& passOptions,
+                               Module& module) {
+  Type type = curr->type;
+  if (!type.isRef()) {
+    // Only reference types can be improved (excepting improvements to
+    // unreachable, which we leave to refinalization).
+    // TOOD: Handle tuples if that ever becomes important.
+    return type;
+  }
+  while (1) {
+    auto* next = getImmediateFallthrough(curr, passOptions, module);
+    if (next == curr) {
+      return type;
+    }
+    if (next->type == Type::unreachable) {
+      return Type::unreachable;
+    }
+    assert(next->type.isRef());
+    // Improve nullability and heap type independently. This can lead to more
+    // precise results than if we checked subtyping on the type as a whole.
+    if (type.isNullable() && next->type.isNonNullable()) {
+      type = Type(type.getHeapType(), NonNullable);
+    }
+    if (next->type.getHeapType() != type.getHeapType() &&
+        HeapType::isSubType(next->type.getHeapType(), type.getHeapType())) {
+      type = Type(next->type.getHeapType(), type.getNullability());
+    }
+    curr = next;
+  }
+}
+
+// Find the best fallthrough value ordered by refinement of heaptype, refinement
+// of nullability, and closeness to the current expression. The type of the
+// expression this function returns may be nullable even if `getFallthroughType`
+// is non-nullable, but the heap type will definitely match.
+inline Expression** getMostRefinedFallthrough(Expression** currp,
+                                              const PassOptions& passOptions,
+                                              Module& module) {
+  Expression* curr = *currp;
+  if (!curr->type.isRef()) {
+    return currp;
+  }
+  auto bestType = curr->type.getHeapType();
+  auto bestNullability = curr->type.getNullability();
+  auto** bestp = currp;
+  while (1) {
+    curr = *currp;
+    auto** nextp =
+      Properties::getImmediateFallthroughPtr(currp, passOptions, module);
+    auto* next = *nextp;
+    if (next == curr || next->type == Type::unreachable) {
+      return bestp;
+    }
+    assert(next->type.isRef());
+    auto nextType = next->type.getHeapType();
+    auto nextNullability = next->type.getNullability();
+    if (nextType == bestType) {
+      // Heap types match: refine nullability if possible.
+      if (bestNullability == Nullable && nextNullability == NonNullable) {
+        bestp = nextp;
+        bestNullability = NonNullable;
+      }
+    } else {
+      // Refine heap type if possible, resetting nullability.
+      if (HeapType::isSubType(nextType, bestType)) {
+        bestp = nextp;
+        bestNullability = nextNullability;
+        bestType = nextType;
+      }
+    }
+    currp = nextp;
+  }
+}
+
 inline Index getNumChildren(Expression* curr) {
   Index ret = 0;
 
--- a/src/ir/properties.h
+++ b/src/ir/properties.h
@@ -371,7 +371,7 @@ inline Type getFallthroughType(Expression* curr,
   if (!type.isRef()) {
     // Only reference types can be improved (excepting improvements to
     // unreachable, which we leave to refinalization).
-    // TOOD: Handle tuples if that ever becomes important.
+    // TODO: Handle tuples if that ever becomes important.
     return type;
   }
   while (1) {
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -2004,16 +2004,23 @@ struct OptimizeInstructions
     Type refType =
       Properties::getFallthroughType(curr->ref, getPassOptions(), *getModule());
 
-    // If the cast admits null but we know the value is not null, then as a
-    // first step we can tighten up the cast to stop admitting nulls.
-    if (curr->type.isNullable() && refType.isNonNullable()) {
-      curr->type = Type(curr->type.getHeapType(), NonNullable);
+    // As a first step, we can tighten up the cast type to be the greatest lower
+    // bound of the original cast type and the type we know the cast value to
+    // have. We know any less specific type either cannot appear or will fail
+    // the cast anyways.
+    auto glb = Type::getGreatestLowerBound(curr->type, refType);
+    if (glb == Type::unreachable) {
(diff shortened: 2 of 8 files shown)


tlively:
* main
  * **PR #5876** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5876" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5876?utm_source=stack-comment).

src/ir/properties.h
```diff
@@ -362,6 +362,83 @@ inline Expression* getFallthrough(
   }
 }
 
+// Look at all the intermediate fallthrough expressions and return the most
+// precise type we know this value will have.
```
kripken, line 366: It would be good to comment how this handles the case of not having a most precise type. From the code it looks like nullability and heap types are treated independently, but the problem can happen among heap types too? And the code looks like it does a greedy search, so it might not find the "deepest" result,
```
    A
   / \
  B   C
 /
D
```
Here we might return C when D might be better for the caller. Though even between B and C it isn't clear which the caller would prefer, and the old code considered them both - can't that be better?

tlively: If we see a value typed with both `B` (or `D`) and `C`, we know the value cannot possibly exist, so if we observe that, we can jump straight to the bottom type. The commit I just pushed now optimizes this case.

src/ir/properties.h
```diff
@@ -362,6 +362,92 @@ inline Expression* getFallthrough(
...
+inline Type getFallthroughType(Expression* curr,
+                               const PassOptions& passOptions,
+                               Module& module) {
+  Type type = curr->type;
+  if (!type.isRef()) {
+    // Only reference types can be improved (excepting improvements to
+    // unreachable, which we leave to refinalization).
+    // TOOD: Handle tuples if that ever becomes important.
```
kripken, line 374:
```suggestion
    // TODO: Handle tuples if that ever becomes important.
```

src/passes/OptimizeInstructions.cpp
```diff
@@ -1995,39 +1995,60 @@ struct OptimizeInstructions
...
-          // cast.
+    // Given what we know about the type of the value, determine what we know
+    // about the results of the cast and optimize accordingly.
+    switch (GCTypeUtils::evaluateCastCheck(refType, curr->type)) {
+      case GCTypeUtils::Unknown:
+        // The cast may or may not succeed, so we cannot optimize.
+        break;
+      case GCTypeUtils::Success:
+      case GCTypeUtils::SuccessOnlyIfNonNull: {
+        // We know the cast will succeed, so we can try to optimize it out. Find
```
kripken, line 2029: This comment matches `Success` but not `SuccessIfNonNull`.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1995,39 +1995,60 @@ struct OptimizeInstructions
...
+        // the best-typed fallthrough value to propagate.
+        auto** refp = Properties::getMostRefinedFallthrough(
+          &curr->ref, getPassOptions(), *getModule());
+        auto* ref = *refp;
+        assert(ref->type.isRef());
```
kripken, line 2034: Must this be a ref? What about
```wat
(ref.cast $foo
  (block (ref $foo)
    (unreachable)
  )
)
```
?

src/passes/OptimizeInstructions.cpp
```diff
@@ -1995,39 +1995,60 @@ struct OptimizeInstructions
...
-        auto result = GCTypeUtils::evaluateCastCheck(ref->type, curr->type);
+    // Look at all the fallthrough values to get the most precise possible type
+    // of the value we are casting. local.tee, br_if, and blocks can all "lose"
+    // type information, so looking at all the fallthrough values can give us a
+    // more precise type than is stored in the IR.
+    Type refType =
+      Properties::getFallthroughType(curr->ref, getPassOptions(), *getModule());
+
+    // If the cast admits null but we know the value is not null, then as a
+    // first step we can tighten up the cast to stop admitting nulls.
```
kripken, line 2008: Can we not just apply `cast->type = refType`? That is, the cast must be as refined as the output of `getFallthroughType`, mustn't it? So we can apply not only the nullability but the heap type.

src/passes/OptimizeInstructions.cpp
```diff
@@ -1995,39 +1995,60 @@ struct OptimizeInstructions
...
+          // If the best value to propagate is the argument to the cast, we can
+          // simply remove the cast (or downgrade it to a null check if
+          // necessary).
           if (ref == curr->ref) {
-            replaceCurrent(ref);
+            if (needsNullCheck) {
+              replaceCurrent(builder.makeRefAs(RefAsNonNull, curr->ref));
+            } else {
+              replaceCurrent(ref);
+            }
```
kripken, line 2051: Can't this skip necessary code? `ref` might be a nested child with other stuff we need around it, like this:
```wat
(ref.cast
  (block
    (..side-effect..)
    (ref)
  )
)
```

src/passes/OptimizeInstructions.cpp
tlively: We can only do that if the cast would definitely succeed (i.e. when the value's type is a subtype of the cast type), but in that case we prefer to optimize the cast out entirely, not just improve its type. Otherwise we would be changing the semantics of the cast such that it always succeeds, even if that were not originally the case. This could also cause validation failures if we ended up changing the cast type to some unrelated type.

src/passes/OptimizeInstructions.cpp
tlively: Both `getFallthroughType` and `getMostRefinedFallthrough` return early if they encounter an unreachable, so they always produce referenced types and expressions with reference types as long as their original input had a reference type.

src/passes/OptimizeInstructions.cpp
tlively: This is guarded by a `ref == curr->ref` check, so we know there can be nothing else in the middle.

src/passes/OptimizeInstructions.cpp
tlively: Fixed.

src/passes/OptimizeInstructions.cpp
kripken: I don't understand. Maybe I'm confused about the situation here. Is this transformation not relevant and valid somehow?
```wat
(ref.cast null struct
  (block (result anyref) ;; 'for'get' about the better typing in the children
    (ref.as_non_null
      (ref.cast null $struct.foo ..))))
=>
(ref.cast $struct.foo ;; applied the refined nullability and heap
  (block (result anyref)
    (ref.as_non_null
      (ref.cast null $struct.foo ..))))
```
That seems valid since if we get to the cast then all the constraints above us must have been applied together, so we can use the intersection of them all.

src/passes/OptimizeInstructions.cpp
kripken: But note that in my example the type of the block is *not* unreachable, it is `(ref $foo)`, even though the contents are unreachable. I see a check in the code for the type of the block but not for its content, so I think it will return the last item in the block in that case, which is unreachable.

src/passes/OptimizeInstructions.cpp
kripken: Ah, good point, I missed that.

src/passes/OptimizeInstructions.cpp
tlively: Yes, that example's fine, but only because we know the cast will succeed by looking at all the fallthrough values. We already do even better by just replacing the cast entirely in this case.

Here's an example that would go wrong:

```
(ref.cast null $struct
  (block (result anyref)
    (ref.cast i31 ..)))
```

This cast will fail, so if we just replaced it with `ref.cast (ref i31)`, that would be a misoptimization. If this cast is the child of a struct.get, it would also cause a validation failure.

In contrast to arbitrarily changing the heap type here, it is valid to improve the nullability from nullable to nonnullable because that always refines the type and cannot be a misoptimization or cause a validation failure. Furthermore, performing this nullability optimization is not already handled somewhere else.

src/passes/OptimizeInstructions.cpp
tlively: `getFallthroughType` and `getMostRefinedFallthrough` will return `(ref $foo)` and a pointer to the block, respectively here.

kripken: I see, thanks, that's an API change compared to `getFallthrough` from before that confused me I guess.

src/passes/OptimizeInstructions.cpp
kripken: Oh, true, but I think we just need the intersection of it all, including the original cast? The intersection of `(ref i31)` and `(ref null $struct)` is empty, so we could put `(ref.cast none` (empty set) in your example. (And in my example, adding the intersection of the original cast has no effect.)

tlively: Yes, that would work, but it would be entirely superseded by the optimizations below. In the example you gave, the cast would be removed and the contents of the block would be teed out. In my example, the cast would be turned into a drop followed by `unreachable`.

kripken: It seems conceptually simpler to me, though, to just take the intersection of all the types (in fallthroughs + the original cast), rather than treat the nullability in a special way + add the code below. Do you disagree?

tlively: If we had a function in wasm-types.h for taking the meet/glb/intersection of two types, then maybe using that might be simpler, but we don't have a function like that ready to go. How would you feel if I updated the comment to point out that improving the heap type here wouldn't be useful due to the later optimizations?

kripken: I was thinking that the intersection is what is sort of already computed here - for nullability, improve to non-null, and for heap type, either go to nothing (if unrelated types) or pick the more refined of the two (if related). I see that code in this PR already, I think, so I was hoping there's a good way to refactor it to use that.

But maybe I'm not seeing some practical complication here that prevents it. You're closer to the code. If that seems impractical then sgtm to continue with this PR more or less as is and leave improvements for later?

tlively: Ok, I think you're right that this would be a nice refactoring in general, even if the refinement of the heap type isn't independently useful in this particular case.

tlively: PTAL at the last couple of commits ('glb' and 'fewer test changes') for this refactoring.

src/passes/OptimizeInstructions.cpp
```diff
@@ -2060,114 +2086,76 @@ struct OptimizeInstructions
...
-          // trap.
-          // Make sure to emit a block with the same type as us; leave updating
-          // types for other passes.
+        }
+        // If we get here, then we know that the heap type of the cast value is
+        // more refined than the heap type of best available fallthrough
+        // expression. The only way this can happen is if we were able to infer
+        // that the value has bottom heap type because it was typed with
+        // multiple, incompatible heap types. In this case, the cast succeeds
+        // only if the value is null, so we can fall through to that case.
```
kripken, line 2102: I don't follow this paragraph, sorry. What is "value" - the immediate value to the cast, i.e. `curr->ref`?

Can you perhaps give an example of this in code? Also, can we add an assert to verify the conclusion here?

test/lit/passes/optimize-instructions-gc.wast
```diff
@@ -2162,6 +2356,327 @@
...
+  ;; CHECK-NEXT:     )
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (ref.as_non_null
+  ;; CHECK-NEXT:   (local.get $1)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $compatible-cast-fallthrough-null-check (param $eqref eqref) (result (ref i31))
+    ;; Similar to above, but now the cast only succeeds if the value is non-null.
```
kripken, line 2417: The previous test also added a `ref.as_non_null`? Should that test have had a nullable cast? (I don't seem to see such a test.)

kripken (approved):
nice, lgtm % comments + fuzzing

src/passes/OptimizeInstructions.cpp
```diff
@@ -2060,114 +2086,76 @@ struct OptimizeInstructions
...
+      }
+        [[fallthrough]];
```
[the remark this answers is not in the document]
tlively: Yes, this is a bad interaction with the special indenting rules when a switch case has braces. I don't think there's anything we can change here without making either the compiler or the linter mad.

test/lit/passes/optimize-instructions-gc.wast
tlively: In the previous test, we know the value was non-null because of the `ref.as_non_null`. This test does not have the `ref.as_non_null`, so we don't know until runtime whether the value is non-null. The output is similar, but the logic to get there is slightly different.

tlively: I improved the comment here.

src/passes/OptimizeInstructions.cpp
tlively: Update the comment, replacing "value" with "input" to try to clarify. I also added assertions.

tlively: Now added an example as well.

tlively:
I made another small improvement to optimization of ref.test using the new getLeastUpperBound and added a test for it. I also have addressed all of the review comments now. A slightly outdated version of this PR made it through 280k fuzzer iterations before I stopped it, so I'll fuzz the latest version lightly and then land this.

src/passes/OptimizeInstructions.cpp
```diff
@@ -2060,114 +2086,80 @@ struct OptimizeInstructions
...
           return;
-        } else if (result == GCTypeUtils::Failure ||
-                   result == GCTypeUtils::Unreachable) {
-          // This cast cannot succeed, or it cannot even be reached, so we can
-          // trap.
-          // Make sure to emit a block with the same type as us; leave updating
-          // types for other passes.
+        }
+        // If we get here, then we know that the heap type of the cast input is
+        // more refined than the heap type of best available fallthrough
```
kripken, line 2098:
```suggestion
        // more refined than the heap type of the best available fallthrough
```

kripken (approved):

PR #5880, opened 2023-08-16
Factor IRBuilder utility out of the new wat parser

tlively:
Add an IRBuilder utility in a new wasm-ir-builder.h header. IRBuilder is
extremely similar to Builder, except that it manages building full trees of
Binaryen IR from a linear sequence of instructions, whereas Builder only builds
a single IR node at a time. To build full IR trees, IRBuilder maintains an
internal stack of expressions, popping children off the stack and pushing the
new node onto the stack whenever it builds a new node.

In addition to providing makeXYZ function to allocate, initialize, and finalize
new IR nodes, IRBuilder also provides a visit() method that can be used when the
user has already allocated the IR nodes and only needs to reconstruct the
connections between them. This will be useful in outlining both for constructing
outlined functions and for reconstructing functions around arbitrary outlined
holes.

Besides the new wat parser and outlining, this new utility can also eventually
be used in the binary parser and to convert from Poppy IR back to Binaryen IR if
that ever becomes necessary.

To simplify this initial change, IRBuilder exposes the same interface as the
code it replaces in the wat parser. A future change requiring more extensive
changes to the wat parser will simplify this interface. Also, since the new code
is tested only via the new wat parser, it only supports building instructions
that were already supported by the new wat parser to avoid trying to support any
instructions without corresponding testing. Implementing support for the
remaining instructions is left as future work.

### DIFF
--- a/src/wasm/wasm-ir-builder.cpp
+++ b/src/wasm/wasm-ir-builder.cpp
@@ -181,8 +181,8 @@ Result<> IRBuilder::visit(Expression* curr) {
   if (auto* block = curr->dynCast<Block>()) {
     block->finalize(block->type);
   } else {
-    // TODO: Call more efficient versions of finalize() for other kinds of nodes
-    // as well.
+    // TODO: Call more efficient versions of finalize() that take the known type
+    // for other kinds of nodes as well, as done above.
     ReFinalizeNode{}.visit(curr);
   }
   return push(curr);
--- a/src/wasm/wasm-ir-builder.cpp
+++ b/src/wasm/wasm-ir-builder.cpp
@@ -1,0 +1,921 @@
+/*
+ * Copyright 2023 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <cassert>
+
+#include "ir/names.h"
+#include "wasm-ir-builder.h"
+
+using namespace std::string_literals;
+
+namespace wasm {
+
+namespace {
+
+Result<> validateTypeAnnotation(HeapType type, Expression* child) {
+  if (child->type == Type::unreachable) {
+    return Ok{};
+  }
+  if (!child->type.isRef() ||
+      !HeapType::isSubType(child->type.getHeapType(), type)) {
+    return Err{"invalid reference type on stack"};
+  }
+  return Ok{};
+}
+
+} // anonymous namespace
+
+std::vector<Expression*>& IRBuilder::getExprStack() {
+  if (scopeStack.empty()) {
+    // We are not in a function, so push a dummy scope.
+    scopeStack.push_back({{}, Type::none});
+  }
+  return scopeStack.back().exprStack;
+}
+
+Result<Index> IRBuilder::addScratchLocal(Type type) {
+  if (!func) {
+    return Err{"scratch local required, but there is no function context"};
+  }
+  Name name = Names::getValidLocalName(*func, "scratch");
+  return Builder::addVar(func, name, type);
+}
+
+Result<> IRBuilder::push(Expression* expr) {
+  auto& exprStack = getExprStack();
+  if (expr->type == Type::unreachable) {
+    // We want to avoid popping back past this most recent unreachable
+    // instruction. Drop all prior instructions so they won't be consumed by
+    // later instructions but will still be emitted for their side effects, if
+    // any.
+    for (auto& expr : exprStack) {
+      expr = builder.dropIfConcretelyTyped(expr);
+    }
+    unreachable = true;
+    exprStack.push_back(expr);
+  } else if (expr->type.isTuple()) {
+    auto scratchIdx = addScratchLocal(expr->type);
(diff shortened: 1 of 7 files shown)


tlively:
* main
  * **PR #5880** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5880" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5880?utm_source=stack-comment).

src/wasm/wasm-ir-builder.cpp
```diff
@@ -0,0 +1,921 @@
...
+Result<> validateTypeAnnotation(HeapType type, Expression* child) {
+  if (child->type == Type::unreachable) {
+    return Ok{};
+  }
+  if (!child->type.isRef() ||
+      !HeapType::isSubType(child->type.getHeapType(), type)) {
+    return Err{"invalid reference type on stack"};
+  }
+  return Ok{};
+}
```
kripken, line 37: It seems a little odd to me to have this here, as it feels more of a concern for the parser. That is, non-parser users don't want this (except maybe as an assertion).

src/wasm/wasm-ir-builder.cpp
```diff
@@ -0,0 +1,921 @@
...
+Result<>
+IRBuilder::makeAtomicLoad(unsigned bytes, Address offset, Type type, Name mem) {
+  auto* curr = wasm.allocator.alloc<Load>();
+  curr->isAtomic = true;
+  curr->bytes = bytes;
+  curr->signed_ = false;
+  curr->offset = offset;
+  curr->align = bytes;
+  curr->type = type;
+  curr->memory = mem;
```
kripken, line 401: Reading long sequences like these makes me wonder if we can share code with Builder somehow. Something like this:
```cpp
Result<>
IRBuilder::makeAtomicLoad(unsigned bytes, Address offset, Type type, Name mem) {
  return visitLoad(builder.makeLoad(nullptr, bytes, offset, type, mem));
}
```
That is, call the Builder method and pass it nullptr for the `Expression*` children. Of course this won't work though as it calls `finalize` there.

The reverse might work, though?
```cpp
Result<>
IRBuilder::makeAtomicLoad(unsigned bytes, Address offset, Type type, Name mem) {
  auto* child = getChild();
  builder.makeLoad(child, bytes, offset, type, mem));
}
```
That is, get the children first using some new code (`getChild()` gets one child, another method could get an arbitrary number, etc.), and then pass that into Builder. It seems like that would also avoid the need for the delegates stuff in `visitExpression`?

A downside to that is we need to write out the number of children here explicitly. But it still seems like significantly less code overall. Does this make sense or am I missing something?

src/wasm/wasm-ir-builder.cpp
tlively: The benefit of performing validation here is that in the future the wat and binary parsers (which could both use this utility in principle) will be able to share this validation logic rather than duplicating it. Validation is done only in the `makeXYZ` methods; `visit` should already assume everything is valid.

src/wasm/wasm-ir-builder.cpp
tlively: Yeah, it would be nice to share this code. Unfortunately I couldn't figure out a good way to do so.

As you mentioned, using `builder` to do the initial allocation and field setting wouldn't work because the children would not yet be initialized.

Popping the children up front and then calling into `builder` would work, and is what wat-parser.cpp did before this change, but has the trade off that it doesn't benefit from being able to simply call `visit` (or `visitXYZ`) to install the children. Popping all the children manually is also a nontrivial amount of code that I was happy to be able to remove, and we benefit from having `visit` be well-tested because that's the API that @ashleynh will be using for outlining.

Somewhat interestingly, it would be more straightforward to reimplement `Builder` in terms of `IRBuilder`, but of course that would do a lot of unnecessary work compared to the current implementation.

tlively: One could imagine having a separate, shared utility called `ImmediateBuilder` that allocates nodes and sets their non-child fields and does nothing else. That could be shared between both `Builder` and `IRBuilder`, but is otherwise completely useless, so I don't think it would be worth it.

kripken: > Popping the children up front and then calling into builder would work, and is what wat-parser.cpp did before this change, but has the trade off that it doesn't benefit from being able to simply call visit (or visitXYZ) to install the children. Popping all the children manually is also a nontrivial amount of code that I was happy to be able to remove

Can we automate that by using delegations in a helper function that gets an Expression type and pops the right number of children, returning them in a vector? The only special cases would need to be optional children, but those already need to be special cased in this code.

src/wasm/wasm-ir-builder.cpp
tlively: FWIW I decided to delete most of the validation out of the `makeXYZ` methods as well. The only validation left is validation of type annotations that can only be done once we know a node's children.

src/wasm/wasm-ir-builder.cpp
tlively: PTAL at the latest version that `visit`s stack-allocated `Expression`s to get the children, then calls `builder.makeXYZ` to arena-allocate, initialize, and finalize the returned node.

src/wasm/wasm-ir-builder.cpp
```diff
@@ -0,0 +1,703 @@
...
+  return ret;
+}
+
+Result<> IRBuilder::visit(Expression* curr) {
+  UnifiedExpressionVisitor<IRBuilder, Result<>>::visit(curr);
+  if (auto* block = curr->dynCast<Block>()) {
+    block->finalize(block->type);
+  } else {
+    // TODO: Call more efficient versions of finalize() for other kinds of nodes
+    // as well.
```
kripken, line 185:
```suggestion
    // as well (passing in the known type, as done above).
```
(it took me a few seconds to realize this)

kripken (approved):

PR #5882, opened 2023-08-17
Further improve ref.cast during finalization

tlively:
We previously improved the nullability and heap type of the ref.cast target type
in RefCast::finalize() based on what we knew about its input type. Simplify the
code and make this improvement more powerful by using the greatest lower bound
of the original cast target and input type.

PR #5883, opened 2023-08-17
Fix finalization of call_ref to handle refined target types

tlively:
Previously CallRef::finalize() would never update the type of the CallRef, even
if the type of the call target had been refined to give a more precise result
type. Besides unnecessarily losing type information, this could also lead to
validation errors, since the validator checks that the type of CallRef matches
the result type of the target signature.

Fix the bug by updating CallRef's type based on its target signature in
CallRef::finalize() and add a test that depends on this refinalization.

### DIFF
--- a/src/wasm/wasm.cpp
+++ b/src/wasm/wasm.cpp
@@ -920,7 +920,9 @@ void I31Get::finalize() {
 }
 
 void CallRef::finalize() {
-  handleUnreachableOperands(this);
+  if (handleUnreachableOperands(this)) {
+    return;
+  }
   if (isReturn) {
     type = Type::unreachable;
     return;
--- a/src/wasm/wasm.cpp
+++ b/src/wasm/wasm.cpp
@@ -937,7 +937,6 @@ void CallRef::finalize() {
   type = target->type.getHeapType().getSignature().results;
 }
 
-
 void RefTest::finalize() {
   if (ref->type == Type::unreachable) {
     type = Type::unreachable;
--- a/src/wasm/wasm.cpp
+++ b/src/wasm/wasm.cpp
@@ -923,16 +923,20 @@ void CallRef::finalize() {
   handleUnreachableOperands(this);
   if (isReturn) {
     type = Type::unreachable;
+    return;
   }
   if (target->type == Type::unreachable) {
     type = Type::unreachable;
+    return;
   }
+  assert(target->type.isRef());
+  if (target->type.getHeapType().isBottom()) {
+    return;
+  }
+  assert(target->type.getHeapType().isSignature());
+  type = target->type.getHeapType().getSignature().results;
 }
 
-void CallRef::finalize(Type type_) {
-  type = type_;
-  finalize();
-}
 
 void RefTest::finalize() {
   if (ref->type == Type::unreachable) {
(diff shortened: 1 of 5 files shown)


tlively:
* main
  * **PR #5881** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5881" target="_blank">Graphite</a> 
    * **PR #5882** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5882" target="_blank">Graphite</a> 
      * **PR #5883** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5883" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5883?utm_source=stack-comment).

src/wasm/wasm.cpp
```diff
@@ -923,16 +923,20 @@ void CallRef::finalize() {
...
     type = Type::unreachable;
+    return;
   }
   if (target->type == Type::unreachable) {
     type = Type::unreachable;
+    return;
+  }
+  assert(target->type.isRef());
+  if (target->type.getHeapType().isBottom()) {
+    return;
```
kripken, line 934: Should we perhaps set the type to unreachable here? That would avoid the extra line in `wasm-binary`, and seems simpler, but maybe I'm missing a corner case...

test/lit/passes/local-subtyping.wast
```diff
@@ -229,6 +234,40 @@
...
+      (ref.func $ret-i31)
+    )
+    (local.set $x
+      ;; After $f is refined to hold $ret-i31 and the call_ref is refinalized,
+      ;; we will be able to refine $x to i31.
+      (call_ref $ret-any
+        (local.get $f)
+      )
+    )
+  )
```
kripken, line 262: Please add a test where `$f` is refined to a bottom type.

src/wasm/wasm.cpp
tlively: I tried doing this initially, but it led to extra complications because our validator checks that with very few exceptions, unreachable nodes must have unreachable children. It may make sense to overhaul this in the future so that all instructions that trap on null (e.g. struct.get, struct.set, etc.) all get type `unreachable` when their input has bottom type. For now, the current behavior is the most consistent with other trapping instructions.

kripken: Ah right... yeah, that makes sense.

test/lit/passes/local-subtyping.wast
tlively: Good idea, done.

src/wasm/wasm.cpp
```diff
@@ -948,23 +952,18 @@ void RefCast::finalize() {
...
-  if (ref->type.isRef() &&
-      HeapType::isSubType(ref->type.getHeapType(), type.getHeapType())) {
-    type = Type(ref->type.getHeapType(), type.getNullability());
-  }
+  // Do not unnecessarily lose type information. We could leave this for
+  // optimizations (and indeed we do a more powerful version of this in
+  // OptimizeInstructions), but doing it here as part of
+  // finalization/refinalization ensures that type information flows through in
+  // an optimal manner and can be used as soon as possible.
+  type = Type::getGreatestLowerBound(type, ref->type);
```
kripken, line 966: (code from the other PR?)

kripken (approved):

PR #5884, opened 2023-08-17
SignatureRefining: Handle updates to call.without.effects

kripken:
If we refine a signature type that is used in a `call.without.effects` then that call's
results may need to be updated. In the IR it looks like a normal call that happens to
pass a function reference as the last param, but it actually means that we call that
function (without side effects), so we need to have the same results, and the validator
already verified that (so the new testcase here fails without this fix).


src/passes/SignatureRefining.cpp
```diff
@@ -284,9 +285,76 @@ struct SignatureRefining : public Pass {
...
+    // is still a direct call to that function. (We could also defer this to
+    // later, if we relaxed validation here, but updating right now is better
+    // for followup optimizations.)
+
+    // Each time we update we create a new import with the proper type. Keep a
+    // map of them to avoid creating more than one for each type.
+    std::unordered_map<HeapType, Function*> newImports;
+
+    auto getImportWithNewResults = [&](Function* import, Type newResults) {
+      auto newType = Signature(import->getParams(), newResults);
```
tlively, line 315: Do we ever generalize the parameters as part of refining the signature? If so, we should use the generalized parameters here.

src/passes/SignatureRefining.cpp
```diff
@@ -284,9 +285,76 @@ struct SignatureRefining : public Pass {
...
+      if (auto iter = newImports.find(newType); iter != newImports.end()) {
+        return iter->second;
+      }
+
+      auto name = Names::getValidFunctionName(*module, import->name);
+      auto newImport =
+        module->addFunction(Builder(*module).makeFunction(name, newType, {}));
```
tlively, line 322: It seems that it would be possible for an import of the proper type to exist already. Would it make sense to initialize the `newImports` map with all the existing intrinsic imports?

kripken: Hmm, perhaps, but (1) other passes merge duplicate imports anyhow, and (2) these imports get removed entirely during the compilation process (a wasm can't ship with these intrinsics in it). So while it wouldn't be a lot of code to add, probably, it doesn't seem worth it.

src/passes/SignatureRefining.cpp
kripken: We refine parameters (not generalize - we want to improve, not preserve *variance, if I understand your intention here), but we finish that process entirely before getting to this function, which I think is simpler. That is, any change to the parameters will be seen here.

tlively (approved):

PR #5885, opened 2023-08-17
ReFinalize in TypeSSA

kripken:
This has been a bug for a while but it became noticeable after #5881 #5882
which do more work in refinalization.


tlively:
LGTM!

PR #5887, opened 2023-08-18
Improve br_on* optimizations

tlively:
Optimize both the known-null and known-non-null cases for BrOnNull and
BrOnNonNull and optimize for more cast behaviors such as SuccessOnlyIfNonNull
and Unreachable for BrOnCast and BrOnCastFail. Leave optimizing
SuccessOnlyIfNull to future work, since that's more complicated. Use type
information from fallthrough values to inform all the optimizations.

### DIFF
--- a/src/passes/RemoveUnusedBrs.cpp
+++ b/src/passes/RemoveUnusedBrs.cpp
@@ -699,7 +699,10 @@ struct RemoveUnusedBrs : public WalkerPass<PostWalker<RemoveUnusedBrs>> {
     }
 
     struct Optimizer : public PostWalker<Optimizer> {
+      PassOptions& passOptions;
       bool worked = false;
+
+      Optimizer(PassOptions& passOptions) : passOptions(passOptions) {}
 
       void visitBrOn(BrOn* curr) {
         // Ignore unreachable BrOns which we cannot improve anyhow. Note that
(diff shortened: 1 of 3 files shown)


tlively:
* main
  * **PR #5881** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5881" target="_blank">Graphite</a> 
    * **PR #5882** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5882" target="_blank">Graphite</a> 
      * **PR #5883** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5883" target="_blank">Graphite</a> 
        * **PR #5887** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5887" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5887?utm_source=stack-comment).

src/passes/RemoveUnusedBrs.cpp
```diff
@@ -712,56 +715,158 @@ struct RemoveUnusedBrs : public WalkerPass<PostWalker<RemoveUnusedBrs>> {
...
-          worked = true;
+          if (refType.isNull()) {
+            // Definitely not taken.
+            replaceCurrent(builder.makeDrop(curr->ref));
+            worked = true;
+            return;
+          }
+          if (refType.isNonNullable()) {
+            // Definitely taken. OptimizeInstructions will remove the null cast
+            // if possible.
```
kripken, line 774:
```suggestion
            // Definitely taken.
```
(that text seems covered already by the comment above? as on line 757 this seems unneeded)

test/lit/passes/remove-unused-brs-gc.wast
```diff
@@ -293,26 +616,34 @@
...
+ ;; CHECK-NEXT:   (select (result (ref null $struct))
+ ;; CHECK-NEXT:    (block (result (ref null $struct))
+ ;; CHECK-NEXT:     (block $something (result (ref null $struct))
+ ;; CHECK-NEXT:      (drop
+ ;; CHECK-NEXT:       (block (result nullref)
+ ;; CHECK-NEXT:        (br_on_non_null $something
+ ;; CHECK-NEXT:         (local.get $struct)
+ ;; CHECK-NEXT:        )
+ ;; CHECK-NEXT:        (ref.null none)
+ ;; CHECK-NEXT:       )
```
kripken, line 628: I understand why the `br_on_cast` in this existing test turned into `br_on_non_null + null`, but why did the `if` get selectified on new line 619? I guess because we consider a `br_on_non_null` to be much less work than `br_on_cast`?

test/lit/passes/remove-unused-brs_all-features.wast
```diff
@@ -142,33 +153,115 @@
...
+   )
+   (unreachable)
+  )
+ )
+
+ ;; CHECK:      (func $br_on_non_null-fallthrough (type $none_=>_funcref) (result funcref)
+ ;; CHECK-NEXT:  (local $x funcref)
+ ;; CHECK-NEXT:  (block $non-null (result (ref func))
+ ;; CHECK-NEXT:   (br $non-null
+ ;; CHECK-NEXT:    (ref.as_non_null
```
kripken, line 238: Similar question as before, we turned `br_on_non_null` into `br + ref.as_non_null` which seems larger and not faster?

test/lit/passes/remove-unused-brs-gc.wast
kripken: Looks like that's the case in `cost.h` - we consider casts very expensive ("infinity"), but null check pretty cheap ("2").

src/passes/RemoveUnusedBrs.cpp
```diff
@@ -712,56 +715,158 @@ struct RemoveUnusedBrs : public WalkerPass<PostWalker<RemoveUnusedBrs>> {
...
-        // Check if the type is the kind we are checking for.
+        // Improve the cast target type as much as possible given what we know
+        // about the input. Unlike in BrOn::finalize(), we consider type
+        // information from all the fallthrough values here.
+        auto glb = Type::getGreatestLowerBound(curr->castType, refType);
+        if (glb != Type::unreachable && glb != curr->castType) {
+          curr->castType = glb;
+          curr->finalize();
+          worked = true;
+        }
```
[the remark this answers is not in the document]
tlively: Not sure what I would say here besides "it is ok to continue to the next optimizations after this"

test/lit/passes/remove-unused-brs-gc.wast
```diff
@@ -43,28 +45,119 @@
...
+ ;; CHECK-NEXT:      (local.tee $any
+ ;; CHECK-NEXT:       (struct.new_default $struct)
+ ;; CHECK-NEXT:      )
+ ;; CHECK-NEXT:     )
+ ;; CHECK-NEXT:    )
+ ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:   (drop
+ ;; CHECK-NEXT:    (block (result nullref)
+ ;; CHECK-NEXT:     (br_on_non_null $block
+ ;; CHECK-NEXT:      (ref.cast null $struct
```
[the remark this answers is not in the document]
tlively: The idea is that OptimizeInstructions along with MergeBlocks and maybe other simple passes will almost certainly be able to optimize out the `ref.cast`, leaving only the `br_on_non_null`.

test/lit/passes/remove-unused-brs-gc.wast
tlively: Yep, that sounds right. Do you think there's anything worth changing here?

kripken: Ah, no, this looks right.

test/lit/passes/remove-unused-brs-gc.wast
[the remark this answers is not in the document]
kripken: That seems plausible, but even in cases we can do it I'm not sure we actually do. Maybe replace the drop with something that uses the value and see if even this testcase gets optimized? I would guess it does not atm.

I'm also not sure why it should almost certainly be optimizable in general?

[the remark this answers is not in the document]
tlively: This code is generated in the `SuccessOnlyIfNonNull` case. We know for a fact based on the fallthrough types that this new `ref.cast` will succeed at runtime, and we're emitting it only to maintain validity. OptimizeInstructions will  recognize that the `ref.cast` will certainly succeed and turn it into a tee+drop+get sequence at worst and simply remove it at best, if other passes were able to make improvements in the meantime.

test/lit/passes/remove-unused-brs_all-features.wast
tlively: Similarly, we know for a fact that the `ref.as_non_null` will succeed at runtime, so we are expecting other optimizations to be able to get rid of it. Even if they don't, making the branch unconditional will probably allow  following code to be DCE'd, so I would still expect this to be beneficial.

test/lit/passes/remove-unused-brs-gc.wast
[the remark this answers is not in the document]
kripken: Ah, I see, I forgot that part of OptimizeInstructions, thanks...

kripken (approved):

tlively:
Fuzzer looks ok after over 38k iterations.

PR #5889, opened 2023-08-21
Make the legacy parser follow more closely the standard GC text format

vouillon:
For now, the previous syntax is still accepted.


tlively:
Thanks for the patch! Can you add a test for the changes in test/lit? You can use `scripts/update_lit_checks.py -f path/to/test.wast` to automatically generate the checks for it if you add just a RUN line at the top of the test.

vouillon:
I have added some tests.

I should probably change the text output and all the tests to use the standard syntax. But maybe i can do that in another PR?

tlively (approved):
Awesome, thanks for working on this.

tlively:
> I should probably change the text output and all the tests to use the standard syntax. But maybe i can do that in another PR?

This would be very helpful. Doing it in a separate PR would be fine.

PR #5894, opened 2023-08-23
Use the standard syntax for ref.cast, ref.test and array.new_fixed

vouillon:
The text output and the tests have been updated.


tlively (approved):
🎉

PR #5897, opened 2023-08-23
Fix merging of unrelated types in TypeMerging

### LINKED ISSUE #5877: TypeMerging error on call param
```wat
(module
 (rec
  (type $A (func (param (ref null $Y )) (result f64)))
  (type $B (sub $A (func (param (ref null $B)) (result f64))))
  (type $C (sub $A (func (param (ref null $A)) (result f64))))

  (type $X (sub $B (func (param (ref null $A)) (result f64))))
  (type $Y (sub final $X (func (param funcref ) (result f64))))
 )

 (func $0 (type $X) (param $0 (ref null $A)) (result f64)
  (unreachable)
 )
 (func $1 (param $1 (ref null $B))
  (unreachable)
 )
 (func $2 (param $0 (ref $Y ))
  (call $1
   (ref.func $0)
  )
 )
 (func $4 (result (ref null $C))
  (unreachable)
 )
)
```
```bash
$ bin/wasm-opt a.wat -all --closed-world --type-merging
[wasm-validator error in function 2] call param types must match, on 
(call $1
 (ref.func $0)
)
(on argument 0)
Fatal: error after opts
```
cc @tlively

tlively:
Previously it was possible that the supertype merging phase would merge
unrelated types when DFA minimization would split a common supertype out of a
partition, leaving unrelated types behind in the same partition. Fix the problem
by post-processing the partitions in the supertype merging phase to split any
partitions that contain unrelated types.

Fixes #5877.

### DIFF
--- a/src/passes/TypeMerging.cpp
+++ b/src/passes/TypeMerging.cpp
@@ -387,7 +387,7 @@ bool TypeMerging::merge(MergeKind kind) {
     // together in the other new partition. Since we are only supposed to be
     // merging types into their supertypes, merging such unrelated types would
     // be unsafe. Post-process the refined partitions to manually split any
-    // partitions containing unrelated types
+    // partitions containing unrelated types.
     //
     // Normally splitting partitions like this would require re-running DFA
     // minimization afterward, but in this case it is not possible that the
--- a/src/passes/TypeMerging.cpp
+++ b/src/passes/TypeMerging.cpp
@@ -152,6 +152,10 @@ struct TypeMerging : public Pass {
   enum MergeKind { Supertypes, Siblings };
   bool merge(MergeKind kind);
 
+  // Split a partition into potentially multiple partitions for each
+  // disconnected group of types it contains.
+  std::vector<Partition> splitSupertypePartition(const std::vector<HeapType>&);
+
   CastTypes findCastTypes();
   std::vector<HeapType> getPublicChildren(HeapType type);
   DFA::State<HeapType> makeDFAState(HeapType type);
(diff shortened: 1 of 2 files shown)


tlively:
* main
  * **PR #5897** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5897" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5897?utm_source=stack-comment).

test/lit/passes/type-merging.wast
```diff
@@ -953,6 +953,80 @@
...
+  (type $D2 (sub $C (struct (ref null $B) (ref null $K))))
+ )
+
+ ;; CHECK:      (global $g1 (ref $B) (struct.new_default $D1))
+ (global $g1 (ref $B) (struct.new_default $D1))
+ ;; CHECK:      (global $g2 (ref $C) (struct.new_default $D2))
+ (global $g2 (ref $C) (struct.new_default $D2))
+)
+
+;; Same as above, but with some addition types that can be merged.
```
kripken, line 991:
```suggestion
;; Same as above, but with some additional types that can be merged.
```

src/passes/TypeMerging.cpp
```diff
@@ -361,16 +367,46 @@ bool TypeMerging::merge(MergeKind kind) {
...
+  dumpRefinedPartitions();
+#endif
+
+  if (kind == Supertypes) {
+    // It's possible that a partition has been split such that a common ancestor
+    // ended up in one of the new partitions, leaving unrelated types grouped
+    // together in the other new partition. Since we are only supposed to be
+    // merging types into their supertypes, merging such unrelated types would
+    // be unsafe. Post-process the refined partitions to manually split any
+    // partitions containing unrelated types
```
kripken, line 390:
```suggestion
    // partitions containing unrelated types.
```

kripken (approved):
Makes sense to me. But this is quite tricky code, I wonder if this might be the rare case where it makes sense to write a formal proof of an optimization. OTOH these rare cases might just be fuzzer situations so it might not matter much...

tlively:
> Makes sense to me. But this is quite tricky code, I wonder if this might be the rare case where it makes sense to write a formal proof of an optimization. OTOH these rare cases might just be fuzzer situations so it might not matter much...

Yeah, I considered adding a checker to the type fuzzer that would catch bugs in this pass. Maybe I can do that as a follow-up?

My initial patch was even more robust and ran the manual splitting followed by further DFA minimization in a loop until it detected a fixed point. You can see that code in the first commit here. But then when trying to come up with a test case, I managed to convince myself that the further DFA minimization and looping wasn't necessary.

PR #5898, opened 2023-08-23
GlobalStructInference: Add missing ReFinalize

kripken:


src/passes/GlobalStructInference.cpp
```diff
@@ -265,6 +268,11 @@ struct GlobalStructInference : public Pass {
           curr->ref = builder.makeSequence(
             builder.makeDrop(builder.makeRefAs(RefAsNonNull, curr->ref)),
             builder.makeGlobalGet(global, wasm.getGlobal(globals[0])->type));
+          // The struct.get will now read from something of the type of the
+          // global, which might be more refined compared to before, so the
+          // field being read might be refined as well, which could change the
+          // struct.get's type.
+          refinalize = true;
```
tlively, line 275: It looks like it would be fairly simple to check whether the type changed and only refinalize if so. Do you think that would be worth it?

tlively (approved):

src/passes/GlobalStructInference.cpp
kripken: Good idea, done.

kripken: (I checked the type of the global vs the ref, but I guess we could actually check if the field is refined as well... but that would be a bunch more work, and this is rare enough as it is I suspect.)

PR #5899, opened 2023-08-23
Fix a merge conflict on main

kripken:
Looks like https://github.com/WebAssembly/binaryen/pull/5894 raced with something else, as tests passed there but fail on `main`. Might be a new test was added since that PR was opened.

(this just applies an update to the existing tests)


tlively (commented):
Oops! LGTM

PR #5900, opened 2023-08-24
DebugInfo: Add a method to find debug info for replacements and use it in Precompute

kripken:
Many optimizations just move code around, and debug info is preserved anyhow as we
do so. Precompute however constantly creates new constant nodes, so we were losing
a significant amount of debug info.

Just copying debug info for the replacement is not enough, as we may replace a parent
with a child, like this:
```wat
(block
  ;; debug info on the const
  (i32.const 42)
)
=>
(i32.const 42)
```
When replacing the block we should look for debug info anywhere inside it, if we have
anything else, that is, debug info on a child is probably better than not having any
debug info at all.

This helps save debug info after inlining in some cases, see the new testcases.


aheejin:
Is this generally helpful even when the replacement is not a child of the original? In case it is not, is inaccurate info better than no info?

kripken:
I think it is generally helpful, but maybe we can think of an example that's not true? My thinking, at least, is that if we replace `A(B, C)` with some `D` then `D` does what all of `A(B, C)` did, so it's better to use the source file location for say `C` if we don't have one for `A` - it's at least representing the location of some of the source code that `D` implements.

aheejin (approved):

kripken:
This is probably still worth landing but I'll wait for other pending PRs in this area to land first, as this is the least important. We may also want a more general fix here, I'm not sure.

PR #5903, opened 2023-08-28
fix: debug info print with depth

JesseCodeBones:
Split from https://github.com/WebAssembly/binaryen/pull/5547.
Focus on debug info print optimization.

### DIFF
--- a/src/passes/Print.cpp
+++ b/src/passes/Print.cpp
@@ -170,6 +170,8 @@ struct PrintSExpression : public UnifiedExpressionVisitor<PrintSExpression> {
 
   std::vector<HeapType> heapTypes;
 
+  unsigned lastPrintIndent = 0;
+
   // Print type names by saved name or index if we have a module, or otherwise
   // by generating minimalist names. TODO: Handle conflicts between
   // user-provided names and the fallback indexed names.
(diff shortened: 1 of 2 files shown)


JesseCodeBones:
@kripken 
Hi Alon,
the PR is ready, please kindly review, thank you.
Regards

src/passes/Print.cpp
```diff
@@ -170,6 +170,8 @@ struct PrintSExpression : public UnifiedExpressionVisitor<PrintSExpression> {
 
   std::vector<HeapType> heapTypes;
 
+  unsigned lastPrintIndent = 0;
```
kripken, line 173: Please add a comment explaining the motivation for tracking the print indent, maybe with an example? I am imagining something like this:
```cpp
//  Track the print indent so that we can see when it changes. That affects how we print debug annotations. In particular, we don't want to print repeated debug locations for children, like this:
//
//  ;;@ file.cpp:20:4
// (block
//    ;; no need to annotate here
//    (nop)
//
// But we do want to print an annotation even if it repeats if it is not a child:
//
//  ;;@ file.cpp:20:4
// (block
// )
// ;;@ file.cpp:20:4 - this is worth annotating
// (nop)
```

kripken (commented):
Thanks! Looks good with some comments, see the suggestion. Is what I wrote there correct?

kripken (approved):
Actually let me land this - I have a followup to this code, and there would just be merge conflicts etc. I'll add a comment in my PR.

PR #5906, opened 2023-08-29
Source maps: Fix locations without debug info between ones that do

kripken:
Previously we did nothing for instructions without debug info. So if we had one
that did, followed by one that didn't, the one that didn't could get "smeared" with
the debug info of the first. Source map locations are the start of segments,
apparently, and so if we say a location has info then all others after it will as well,
until the next segment.

To fix that, add support for source map entries with just one field, the binary
location. Such entries have no debug info (no `file:line:col`), and though the source
maps spec is not very clear on this, this seems like the right way to prevent this
problem: to stop a segment with debug info by starting a new one without, when
we know we don't want that info any more. That is, before this PR we could have
this:
```wat
;; file.cpp:10:1
(nop) ;; binary offset 5 in wasm
(nop) ;; binary offset 6 in wasm
```
and the second `nop` would get the same debug annotation, since we just have
one source map segment,
```cpp
[5, file.cpp, 10, 1]  // start at offset 5 in wasm
```
With this PR, we emit:
```cpp
[5, file.cpp, 10, 1]  // start at offset 5 in wasm; file.cpp:10:1
[6]                   // start at offset 6 in wasm; no debug info
```

This causes no errors on the emscripten source map tests in addition to the ones
in this repo, so I am somewhat confident about it. It does add 7% to source map
sizes, however, since we add those 1-length segments now, but that seems
unavoidable to fix this bug. (This was reported by a Java user that was seeing an
obviously incorrect stack trace, which was due to such smearing.)

To implement this, add a new field that says if the next location in the source map
has debug info or not, and use that.


src/wasm/wasm-binary.cpp
```diff
@@ -2818,6 +2853,17 @@ void WasmBinaryReader::readNextDebugLocation() {
 
     int32_t positionDelta = readBase64VLQ(*sourceMap);
     uint32_t position = nextDebugLocation.availablePos + positionDelta;
+
+    nextDebugLocation.previousPos = nextDebugLocation.availablePos;
+    nextDebugLocation.availablePos = position;
```
kripken, line 2858: This is not new logic - just moved up from below (since we also want it for a 1-length entry).

kripken:
cc @JesseCodeBones

src/wasm/wasm-binary.cpp
JesseCodeBones: @kripken 
Hi Alon,
does this optimization only impact the debug info read process? how about the `BinaryenFunctionSetDebugLocation` API? I think AssemblyScript is using this API for debug info generation, it also have the smeared issue for the shadow stack check.

kripken: This part of the code, or this PR as a whole?

The PR as a whole affects binary read and write. We now write 1-length segments to prevent smearing, and we read them properly as well.

`SetDebugLocation` should still be fine. That sets debug info on a single expression. This PR fixes things in binary reading and writing, which due to smearing would cause things to get debug info when they should not. See the testcase in the PR for an example of a bug that is fixed.

aheejin (approved):

aheejin:
This seems to have broken `test_emsymbolizer`: https://ci.chromium.org/ui/p/emscripten-releases/builders/try/linux/b8771281787202226961/overview

kripken:
Sorry, I tried all the tests with `source` or `debug` in their name, and missed this one... looking now.

PR #5907, opened 2023-08-29
Do not prompt user on fuzz_opt.py --auto-initial-contents

tlively:
Remove the prompt for user confirmation when using the --auto-initial-contents
option with the fuzzer. It is not actionable, and it prevents me from going off
and doing something else when I build and start the fuzzer in the same command.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -274,14 +274,6 @@ def init_important_initial_contents():
         print('  ' + test)
     print()
 
-    # We prompt the user only when there is no seed given. This fuzz_opt.py is
-    # often used with seed in a script called from wasm-reduce, in which case we
-    # should not pause for a user input.
-    if given_seed is None:
-        ret = input('Do you want to proceed with these initial contents? (Y/n) ').lower()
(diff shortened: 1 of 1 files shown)


tlively:
* main
  * **PR #5907** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5907" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5907?utm_source=stack-comment).

kripken (approved):
lgtm.

Actually should we make auto initial contents the default as well? I never use anything else...

tlively:
I think it would be a more useful default, but we do have some logic for disallowing it when the current directory is not a git repo (for example, if someone just downloads a source tarball, I guess). Should we just print a warning and fall back to non-auto-initial-contents behavior when we detect that we're not in a git repo?

kripken:
If it's simpler I'd be ok with an error when not in git actually. This mode is so useful I wouldn't want to encourage anything else...

tlively:
I'll merge this and look into making this the default in a follow-up.

PR #5909, opened 2023-08-30
Validate and fix up tuples with non-nullable elements

tlively:
The code validating and fixing up non-nullable locals previously did not
correctly handle tuples that contained non-nullable elements, which could have
resulted in invalid modules going undetected. Update the code to handle tuples
and add tests.


tlively:
* main
  * **PR #5909** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5909" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5909?utm_source=stack-comment).

src/ir/type-updating.cpp
```diff
@@ -385,21 +387,48 @@ void handleNonDefaultableLocals(Function* func, Module& wasm) {
...
     return Builder(wasm).makeRefAs(RefAsNonNull, get);
   }
+  if (get->type.isTuple()) {
+    auto type = get->type;
+    get->type = getValidLocalType(type, wasm.features);
+    std::vector<Expression*> elems(type.size());
+    Builder builder(wasm);
+    for (Index i = 0, size = type.size(); i < size; ++i) {
+      auto* elemGet =
+        i == 0 ? get : builder.makeLocalGet(get->index, get->type);
```
kripken, line 424: I'm confused.  If the tuple has 3 elements, doesn't this look treat 0 as the (fixed up) tuple, and then 1 and 2 as the 2nd and 3rd indexes? What about the 1st index in the tuple?

That is, the first item in this loop has tuple type while all others do not..?

tlively: The original `get` and all the new `get`s have the same full tuple type. It's the tuple extract below that picks a single element out of it.

kripken: Oh, right... sorry, I was very confused here.

kripken (approved):

tlively:
The fuzzer did find a bug where I never updated the code that updates tees. I have a fix, but producing a minimized test case is taking a while.

tlively:
Going ahead and landing this. The new code is extremely similar to existing code and it is well-tested. The fuzzer seems happy with it so far.

PR #5911, opened 2023-08-30
[NFC] Document source maps and other debug support

kripken:
I thought it might be useful to document some things I learned about
source maps recently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,6 +321,10 @@ Advanced optimization techniques in the Binaryen optimizer include
 [Flat IR](https://github.com/WebAssembly/binaryen/blob/main/src/ir/flat.h), and
 [Stack/Poppy IR](https://github.com/WebAssembly/binaryen/blob/main/src/ir/stack-utils.h).
 
+See the
+[Optimizer Cookbook wiki page](https://github.com/WebAssembly/binaryen/wiki/Optimizer-Cookbook)
+for more on how to use the optimizer effectively.
+
 Binaryen also contains various passes that do other things than optimizations,
 like
 [legalization for JavaScript](https://github.com/WebAssembly/binaryen/blob/main/src/passes/LegalizeJSInterface.cpp),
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -814,6 +818,98 @@ a possible bug. See [the wiki page](https://github.com/WebAssembly/binaryen/wiki
...
+
+That is mostly straightforward, but one thing we need to do is to handle the
+lack of debug info in between things that have it. If we have `A B C` where `B`
+lacks debug info, then just emitting a segment for `A` and `C` would lead `A`'s
+segment to also cover `B`, since in source maps segments do not have a size -
+rather they end when a new segment begins. To avoid `B` getting smeared in this
+manner, we emit a source maps entry to `B` of size 1, which just marks the
+binary offset it has, and without the later 3 fields of the source file, line
+number, and column. (This appears to be the intent of the source maps spec, and
+works in browsers and tools.)
```
aheejin, line 903: Source map allows one-character segments but I wasn't able to find what that meant while searching. Is this used for this meaning in JS too?

aheejin (approved):
Thanks!

README.md
kripken: Yes, AFAICT that's what the source maps spec intends for all users. The spec doesn't have much in terms of explanation or details (an example would have saved me much time...), but I think there is no other purpose for a 1-length segment, i.e. a segment with just a binary location and no debug info. It has to mean just that, to mark a location as not having debug info.

(This ambiguity would not have been an issue if segments had endings, but maybe that's less compact to encode...)

aheejin: > (This ambiguity would not have been an issue if segments had endings, but maybe that's less compact to encode...)

Sourcemap has line endings `;`, so I guess the inventors thought the line would end soon so it would have been mostly fine..? Which I'm not sure because minified JS doesn't have many newlines either. Also it doesn't apply to us because we only use a single line for binary...

kripken: Interesting, yeah, maybe for wasm this is more of an issue since the binary is always a single "line".

PR #5912, opened 2023-08-30
[NFC] Source maps: Simplify the code and add comments

kripken:
Almost entirely trivial, except for this part:

```diff
-  if (nextDebugLocation.availablePos == 0 &&
-      nextDebugLocation.previousPos <= pos) {
+  if (nextDebugLocation.availablePos == 0) {
      return;
```

I believe removing the extra check has no effect. Removing it does not change
anything in the test suite, and logically, if we set `availablePos` to 0 then we
definitely want to return here - we set it to 0 to indicate there is nothing left
to read, which is what the code after it does.

As a result, we can remove the `previousPos` field entirely.

cc @JesseCodeBones - am I missing something about `previousPos`? Do you have
an example maybe that shows it is needed?

### DIFF
--- a/src/wasm-binary.h
+++ b/src/wasm-binary.h
@@ -1540,7 +1540,7 @@ class WasmBinaryReader {
   // and there is nothing left to read.
   size_t nextDebugPos;
 
-  // The debug location (file:line:col) corresponding to |nextDebugOffset|. That
+  // The debug location (file:line:col) corresponding to |nextDebugPos|. That
   // is, this is the next 3 fields in a source map entry that we have read, but
   // not used yet.
   //
--- a/src/wasm-binary.h
+++ b/src/wasm-binary.h
@@ -1531,25 +1531,29 @@ class WasmBinaryReader {
   std::istream* sourceMap;
 
   // The binary position that the next debug location refers to. That is, this
-  // is the first item in a source map entry that we have read, but have not
-  // used yet (we use it when we read the expression at this debug location).
+  // is the first item in a source map entry that we have read (the "column", in
+  // source map terms, which for wasm means the offset in the binary). We have
+  // read this entry, but have not used it yet (we use it when we read the
+  // expression at this binary offset).
   //
-  // This is set to 0 if we reach the end of the source and there is nothing
-  // left to read.
+  // This is set to 0 as an invalid value if we reach the end of the source map
+  // and there is nothing left to read.
   size_t nextDebugPos;
 
-  // The debug location (file:line:col) corresponding to |nextDebugOffset|. If
-  // that location has no debug info, then this contains the info from the
-  // previous one (because in a source map, these fields are relative to their
-  // last appearance (we "skip" over a place without debug info).
+  // The debug location (file:line:col) corresponding to |nextDebugOffset|. That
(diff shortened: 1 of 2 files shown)


src/wasm-binary.h
```diff
@@ -1526,34 +1526,48 @@ class WasmBinaryReader {
...
+  // is the first item in a source map entry that we have read (the "column", in
+  // source map terms, which for wasm means the offset in the binary). We have
+  // read this entry, but have not used it yet (we use it when we read the
+  // expression at this binary offset).
+  //
+  // This is set to 0 as an invalid value if we reach the end of the source map
+  // and there is nothing left to read.
+  size_t nextDebugPos;
+
+  // The debug location (file:line:col) corresponding to |nextDebugOffset|. That
```
aheejin, line 1543:
```suggestion
  // The debug location (file:line:col) corresponding to |nextDebugPos|. That
```
Which one (`nextDebugPos` vs. `nextDebugOffset`) did you intend to use?

src/wasm-binary.h
```diff
@@ -1526,34 +1526,48 @@ class WasmBinaryReader {
...
+  // If that location has no debug info (it lacks those 3 fields), then this
+  // contains the info from the previous one, because in a source map, these
+  // fields are relative to their last appearance, so we cannot forget them (we
+  // can't just do something like std::optional<DebugLocation> or such); for
+  // example, if we have line number 100, then no debug info, and then line
+  // number 500, then when we get to 500 we will see "+400" which is relative to
+  // the last existing line number (we "skip" over a place without debug info).
+  Function::DebugLocation nextDebugLocation;
+
+  // Whether debug info is present on |nextDebugOffset| (see comment there).
```
aheejin, line 1556:
```suggestion
  // Whether debug info is present on |nextDebugPos| (see comment there).
```
Same here

aheejin (approved):
Thanks! This is much easier to understand. By the way we had a similar data fields before #5504 - Then it was
```cpp
std::pair<uint32_t, Function::DebugLocation> nextDebugLocation;
```
And now it is 
```cpp
size_t nextDebugPos;
Function::DebugLocation nextDebugLocation;
```

Then I wonder what #5504 was for..? Given that #5504 seemed to fix something and this is NFC I guess I've missed something...

kripken:
> Given that https://github.com/WebAssembly/binaryen/pull/5504 seemed to fix something and this is NFC I guess I've missed something...

Yes, I think some parts of that PR were not needed, which are partially reverted here, unless I am mistaken. But other parts of that PR were good fixes, namely

https://github.com/WebAssembly/binaryen/pull/5504/files#diff-666d70a4550a2ccec6247c639d54231a05108efc40af830403dfbc6527b251f7R2762-R2768

and

https://github.com/WebAssembly/binaryen/pull/5504/files#diff-227e746a6f9adba03a777865b9bd74e49db780106f349eae1aaed0ca3d2eb91eL234-R261

src/wasm-binary.h
kripken: Thanks, good catch. I meant to use `pos` because it matches the variable `pos` we use to track the binary position.

PR #5918, opened 2023-09-01
Make final types the default

tlively:
Match the spec and parse the shorthand binary and text formats as final and emit
final types without supertypes using the shorthands as well. This is a
potentially-breaking change, since the text and binary shorthands can no longer
be used to define types that have subtypes.

Also make TypeBuilder entries final by default to better match the spec and
update the internal APIs to use the "open" terminology rather than "final"
terminology. Future changes will update the text format to use the standard "sub
open" rather than the current "sub final" keywords. The exception is the new wat
parser, which supporst "sub open" as of this change, since it didn't support
final types at all previously.


tlively:
* main
  * **PR #5918** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5918" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5918?utm_source=stack-comment).

tlively:
Oh neat, the coverage report strongly suggests that the TypeSSA tests have regressed, and indeed if you look at the diff, there should definitely not be so many final types.

Will fix that when I'm done moving.

tlively:
> (github errors on trying to review this PR, so trying a toplevel comment)
> 
> In `wasm-type.cpp`,
> 
> > bool isOpen = false;
> 
> Shouldn't that default to true?

No, we're changing the default to be final, i.e. not open.

test/ctor-eval/gc-2.wast.out
```diff
@@ -1,5 +1,5 @@
 (module
- (type $struct (struct (field i32)))
+ (type $struct (sub (struct (field i32))))
```
kripken, line 2: Why did this change?

tlively: This is a non-final type. Now that we interpret the shorthand `(type $... (struct ...))` as a final type, we need to add `(sub ...)` to the printed output to correctly communicate that this is a non-final type.

kripken: I'm still confused then, sorry. Why is this non-final? Final is the default, and I don't see "open" here, so isn't it final?

tlively: Right now Binaryen's text format does not include any of the changes from https://github.com/WebAssembly/gc/pull/413, so `(type (struct ...))` is final, `(type (sub (struct ...)))` is open, and `(type (sub final (struct ...)))` is final. Yes, this is confusing, so we're going to fix it in that PR.

src/passes/TypeSSA.cpp
```diff
@@ -255,6 +255,7 @@ struct TypeSSA : public Pass {
         builder[i] = oldType.getArray();
       }
       builder[i].subTypeOf(oldType);
+      builder[i].setOpen();
```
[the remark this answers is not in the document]
tlively: This was the easiest way to get the test for clashing rec groups to work out. It should be fine, since the optimal thing for us to be doing is to make all types open up front and then have another pass (not written yet) to close all the types we can at the end.

test/lit/passes/signature-pruning.wast
```diff
@@ -393,12 +393,12 @@
...
 
 (module
-  ;; CHECK:      (type $sig (func (param i32)))
+  ;; CHECK:      (type $sig (sub (func (param i32))))
   (type $sig (func_subtype (param i32) func))
 
   ;; As above, but now an import also uses this signature, which prevents us
   ;; from changing anything.
-  ;; CHECK:      (import "out" "func" (func $import (type $sig) (param i32)))
+  ;; CHECK:      (import "out" "func" (func $import (type $func.0) (param i32)))
```
kripken, line 401: Not specific to this PR, but this automatic type picking has always felt surprising and dangerous to me...

kripken (approved):

test/lit/passes/signature-pruning.wast
tlively: Oh yikes, this appears to be a real bug. `$func.0` does not exist.

tlively: Fixed separately in https://github.com/WebAssembly/binaryen/pull/5927.

tlively:
Going to land this before the separate bug fix instead of after because this is much more at risk of other merge conflicts.

PR #5920, opened 2023-09-06
Add bazel support for binaryen in open source

mollyibot:


generate_intrinsics.py
```diff
@@ -0,0 +1,30 @@
+"""Replicates CMake logic to generate WasmIntrinsics.cpp."""
```
kripken, line 1: Can this file be under `scripts/` or is there a bazel limitation there?

kripken (commented):
Thanks!

I'm not familiar with bazel itself but lgtm aside from one question.

generate_intrinsics.py
mollyibot: I think it needs to be in root directory and we also need to import to google3 to avoid some duplication. We will remove METADATA soon and send a new request. Thanks!

PR #5925, opened 2023-09-07
Add bazel support for binaryen in open source

mollyibot:


BUILD
```diff
@@ -0,0 +1,256 @@
...
+# py_binary(
+#     name = "wasm-opt",
+#     srcs = ["wasm-opt.py"],
+#     data = [":wasm-opt-impl"],
+#     deps = ["//pyglib:resources"],
+# )
+# copybara:insert_end
+
+cc_binary(
+    name = "wasm-opt-impl",
```
walkingeyerobot, line 134: let's name this wasm-opt

BUILD
```diff
@@ -0,0 +1,256 @@
...
+# version <version number> (version_<version number>-<latest commit hash>)
+# Ideally it would also include number of commits since the last version
+# was cut, but we don't easily have access to that number.
+# <version number> must contain only digits because emscripten will throw an
+# error otherwise.
+# copybara:strip_begin
+genrule(
+    name = "config",
+    outs = ["src/config.h"],
+    cmd = """echo '#define PROJECT_VERSION "HEAD"' >> $@""",
```
walkingeyerobot, line 233: sadly we can't use bash commands here; they're not portable for windows users. I recommend a python script.

BUILD
```diff
@@ -0,0 +1,256 @@
...
+)
+# copybara:strip_end_and_replace_begin
```
walkingeyerobot, line 235: I'm not a fan of including these in the open source version. internally we can find a better way to address this; the BUILD file in upstream binaryen should only contain what's useful for open source users.

walkingeyerobot (changes requested):

mobileink:
Strong recommendation: always use `BUILD.bazel`, not plain `BUILD` - avoid any possible name clash with other tools.  Ditto for `WORKSPACE.bazel` (althought that file will be going way with 7.0)

mobileink:
Second strong recommendation: you might as well start with Bazel modules.  It's vastly superior to WORKSPACE-based dependency management.  See my [fork](https://github.com/obazl/binaryen ) for an example.

BUILD
mollyibot: when this is pulled to google3 with git_to_third_party, the BUILD file will be inserted:
```
py_binary(name = "wasm-opt"..)
cc_library(name = "wasm-opt"..)
``` 
so I keep this name for avoiding confilict, however I think we can walk around this issue by , would this work for you?

cc_binary(
#copybara:strip_begin
name = "wasm-opt-impl",
#copybara:strip_end_and_replace_begin
# name = "wasm-opt-impl",
#copybara:replace_end
)

mobileink:
Recommendation: follow Bazel best practices; in particular, avoid toplevel build targets like `bazel build :foo`. Also prefer one BUILD.bazel file per source directory.  Better to avoid having file paths like `src/tools/wasmjs.cpp` etc.  in `srcs`. All those `cc_binary` targets can easily go in `src/tools/BUILD.bazel`, so you get build/run targets like `bazel build src/tools:wasm-js`.  Bazel labels make this easy and convenient.

BUILD
mobileink: if you put the cc_binary target in `src/tools/BUILD.bazel` then there is no name clash.

BUILD
mollyibot: we need to put something as placeholder similar to https://github.com/WebAssembly/binaryen/blob/main/config.h.in, and this is used here https://github.com/WebAssembly/binaryen/blob/main/src/support/command-line.cpp#L18, what do you recommend to replace it with? or do we want touch command-line.cpp to exclude config.h?

BUILD
mollyibot: when we sync the change to google3 to BUILD file, the BUILD file should stay the same as current BUILD of google3 version, and should the wasm-opt stay in BUILD.bazel of root directory?

mobileink:
Just for fun I wrote a very simple C program that generates `src/passes/WasmIntrinsics.cpp` directly from `src/passes/wasm-intrinsics.wat`. No need for a template file, and this also eliminates the need to pull in additional Python resources (absl-py).

Feel free to use it if you like, its at https://github.com/obazl/binaryen/blob/bazel/src/tools/textfile_to_binary.c .  To use it, there is a custom rule in `src/passes/BUILD.bzl`, used in target `//src/passes:wasm_intrinsics`.

BUILD
mobileink: Sorry, I don't understand "sync the change to google3 to BUILD file".  But in general there is no reason to put any targets in the root BUILD.bazel.  See [Packages](https://bazel.build/configure/best-practices#packages).

.bazelrc
```diff
@@ -0,0 +1 @@
+build --action_env=BAZEL_CXXOPTS=-std=c++17
```
mobileink, line 1: If binaryen is used as a dependency this will have no effect.  What is the purpose of BAZEL_CXXOPTS?  If all C++ targets require c++17, that can be added to each target using a global variable, e.g.

```
load("//:CONFIG.bzl", "CXXOPTS")
...
cc_library(
    ...
    copts = CXXOPTS + [ ... ],
   ...
)

File: BUILD.bazel [hunk 5]
```diff
@@ -0,0 +1,268 @@
...
+    python_version = "PY3",
+    deps = ["@com_google_absl_py//absl:app"],
+)
+
+genrule(
+    name = "config",
+    outs = ["src/config.h"],
+    cmd = "$(location :generate_config) $(location :src/config.h)",
+    tools = [":generate_config"]
+)
```
mobileink, line 246: Might want to use stamping for this. See [Stamping Bazel builds with selective delivery](https://www.aspect.dev/blog/stamping-bazel-builds-with-selective-delivery)

BUILD
mollyibot: The BUILD files refers to "src/support/xxx","src/tools/xxx" and "src/**/*" so I think we should still put it under root directory. We do not want to keep two versions of BUILD files in two different locations between google 3(third_party) and github(source of truth), so we use git_to_third_party and copybara:strip and copybara:insert to resolve the difference. In google3(third party) the build file is in root directory.

kripken:
> maybe this can be a follow up?

That makes sense to me. I suggest we do the simplest thing here for now, since eventually I think Binaryen can actually remove the need for any build integration for that file (we could handle it the same way we handle `gen-s-parser.py` for example).

.bazelrc
mollyibot: Is there any drawbacks using [bazelrc](https://bazel.build/run/bazelrc) configuration files ? i got some compilation errors  for example 
```
‘is_same_v’ is not a member of ‘std’; did you mean ‘is_same’?
```
relate this [post](https://stackoverflow.com/questions/70484433/is-trivial-v-is-not-a-member-of-std-while-build-c-project-with-conan) then I add BAZEL_CXXOPTS=-std=c++17.

BUILD.bazel
mollyibot: Thanks for the link, good to know the timestamp. Usually we used sha256 instead of timestamp. stamping sometimes may slow git operation.  I read the section `when to use timestamp` and looks like the stamp is not exposed to Bazel rules in any consistent way.  here we do not necessarily need to expose the timestamp/commit information to users, I add a dummy value `HEAD` to mimic https://github.com/WebAssembly/binaryen/blob/main/config.h.in to avoid additional changes to  where it is called in https://github.com/WebAssembly/binaryen/blob/main/src/support/command-line.cpp#L18

BUILD
walkingeyerobot: Sorry, I think I was unclear here. My comment is about the commented out rules. I don't think we should include them at all. Instead, we should write a BUILD.bazel file that works for open source and then apply any additional transformations that we need to internally. No need to clutter an open source BUILD.bazel file with things that aren't relevant outside Google.

.bazelrc
walkingeyerobot: I agree we want to avoid using a `.bazelrc`; putting flags here can create surprises for people integrating binaryen into their builds. If you're not able to figure out a way to put the c++17 flag it in the BUILD.bazel file, let me know and I'll poke at it in the next day or two.

mollyibot: I just uploaded a version that move this config to BUILD file, and I also moved traditional WORKSPACE file to bzlmod format :)

mobileink: Suggestion: put the c++17 arg in the variable:

```
BINARYEN_COPTS = [
    "-fexceptions",
    "-std=c++17"
]
```

mollyibot: Thanks for catching this, I found this as well when I just uploaded the change but I need to leave before making a change. Just uploaded a new version :).

gkdn:
@walkingeyerobot  Although I agree that more granular targets are better, I think we should try to keep this PR minimal and follow what we had in google3. I think all of this could be refactored later on as we have a setup in place and internal/external stuff becomes in sync. We don't really need to target the ideal scenario in the initial setup.

Similarly I recommend sticking to `BUILD` instead of `BUILD.bazel` since these build files intended to be shared with Blaze (which doesn't support later) and that's what we do in other projects. Again this is not a breaking change to do later if becomes a necessity.

BUILD
walkingeyerobot: Just want to reemphasize this comment. All copybara directives and commented out google3 code should be removed. We can find another way to transform this internally.

gkdn:
> > @walkingeyerobot Although I agree that more granular targets are better, I think we should try to keep this PR minimal and follow what we had in google3. I think all of this could be refactored later on as we have a setup in place and internal/external stuff becomes in sync. We don't really need to target the ideal scenario in the initial setup.
> 
> I am fine with doing this later, but I would like to get some commitment on this. For example, if we merge this PR and a community member submits a follow-up PR to make more granular targets, I would be inclined to merge that as well. At that point, would you or @mollyibot be able to commit to making the appropriate changes internally? I suspect it would be less work to simply do it correctly from the beginning, especially given we already have @mobileink 's [fork](https://github.com/obazl/binaryen) as an example.

The way I was thinking about this was;

If it is easy (my anticipation), it can be done as a follow up and wouldn't require our involvement and it would be still good to start with something close to google3 instead of bundling 2 refactorings.

And if it gets involved, then that decision should be done and executed separate from this and we wouldn't be right folks to take that over.

Does it sound reasonable?

> 
> > Similarly I recommend sticking to `BUILD` instead of `BUILD.bazel` since these build files intended to be shared with Blaze (which doesn't support later) and that's what we do in other projects. Again this is not a breaking change to do later if becomes a necessity.
> 
> I think renaming the file is a fairly straightforward transformation we can take on internally. I'll cite absl as a project that does this. See [this BUILD.bazel file](https://github.com/abseil/abseil-cpp/blob/master/absl/base/BUILD.bazel) as an example. We should strive to be the best open source citizens that we can be.

If you already follow that convention in other places handle it via transformation that being consistent with that sounds good to me.

PR #5927, opened 2023-09-09
Fix printing of types for imported functions

tlively:
Previously, the printer incorrectly reconstructed imported functions' types from
their signatures instead of printing their types directly. This could cause the
printer to print uses of types that were never defined and did not exist in the
module. Fix the bug by printing imported functions' heap types directly.


tlively:
* main
  * **PR #5927** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5927" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5927?utm_source=stack-comment).

kripken (approved):

PR #5941, opened 2023-09-14
Optimize tuple.extract of gets in BinaryInstWriter

tlively:
In general, the binary lowering of tuple.extract expects that all the tuple
values are on top of the stack, so it inserts drops and possibly uses a scratch
local to ensure only the extracted value is left. However, when the extracted
tuple expression is a local.get, local.tee, or global.get, it's much more
efficient to change the lowering of the get or tee to ensure that only the
extracted value is on the stack to begin with. Implement that optimization in
the binary writer.

### DIFF
--- a/src/wasm-stack.h
+++ b/src/wasm-stack.h
@@ -149,9 +149,9 @@ private:
   void countScratchLocals();
   void setScratchLocals();
 
-  // local.get and local.tee expressions that will be followed by
-  // tuple.extracts. We can optimize these by only getting only the local
-  // for the extracted index.
+  // local.get, local.tee, and glboal.get expressions that will be followed by
+  // tuple.extracts. We can optimize these by getting only the local for the
+  // extracted index.
   std::unordered_map<Expression*, Index> extractedGets;
 };
 
--- a/src/wasm-stack.h
+++ b/src/wasm-stack.h
@@ -148,6 +148,11 @@ private:
   InsertOrderedMap<Type, Index> scratchLocals;
   void countScratchLocals();
   void setScratchLocals();
+
+  // local.get and local.tee expressions that will be followed by
+  // tuple.extracts. We can optimize these by only getting only the local
+  // for the extracted index.
+  std::unordered_map<Expression*, Index> extractedGets;
 };
 
 // Takes binaryen IR and converts it to something else (binary or stack IR)
--- a/src/wasm/wasm-stack.cpp
+++ b/src/wasm/wasm-stack.cpp
@@ -87,6 +87,13 @@ void BinaryInstWriter::visitCallIndirect(CallIndirect* curr) {
 }
 
 void BinaryInstWriter::visitLocalGet(LocalGet* curr) {
+  if (auto it = extractedGets.find(curr); it != extractedGets.end()) {
+    // We have a tuple of locals to get, but we will only end up using one of
+    // them, so we can just emit that one.
+    o << int8_t(BinaryConsts::LocalGet)
+      << U32LEB(mappedLocals[std::make_pair(curr->index, it->second)]);
+    return;
+  }
   size_t numValues = func->getLocalType(curr->index).size();
   for (Index i = 0; i < numValues; ++i) {
     o << int8_t(BinaryConsts::LocalGet)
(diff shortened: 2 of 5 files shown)


tlively:
* main
  * **PR #5941** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5941" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5941?utm_source=stack-comment).

test/multivalue.wast.fromBinary.noDebugInfo
```diff
@@ -164,72 +157,13 @@
...
-       (local.get $1)
-      )
-      (drop
-       (local.get $3)
-      )
-      (local.get $12)
-     )
-    )
-    (local.get $13)
-   )
```
kripken, line 229: Great improvement!

src/wasm-stack.h
```diff
@@ -148,6 +148,11 @@ class BinaryInstWriter : public OverriddenVisitor<BinaryInstWriter> {
   InsertOrderedMap<Type, Index> scratchLocals;
   void countScratchLocals();
   void setScratchLocals();
+
+  // local.get and local.tee expressions that will be followed by
+  // tuple.extracts. We can optimize these by only getting only the local
```
kripken, line 153:
```suggestion
  // tuple.extracts. We can optimize these by getting only the local
```

src/wasm/wasm-stack.cpp
```diff
@@ -96,14 +103,28 @@ void BinaryInstWriter::visitLocalGet(LocalGet* curr) {
...
   }
   if (!curr->isTee()) {
+    // This is not a tee, so just finish setting the values.
     o << int8_t(BinaryConsts::LocalSet)
       << U32LEB(mappedLocals[std::make_pair(curr->index, 0)]);
+  } else if (auto it = extractedGets.find(curr); it != extractedGets.end()) {
+    // We only need to get the single extracted value.
+    if (it->second == 0) {
+      o << int8_t(BinaryConsts::LocalTee)
+        << U32LEB(mappedLocals[std::make_pair(curr->index, it->second)]);
```
kripken, line 119:
```suggestion
        << U32LEB(mappedLocals[std::make_pair(curr->index, 0)]);
```

src/wasm/wasm-stack.cpp
```diff
@@ -114,8 +135,14 @@ void BinaryInstWriter::visitLocalSet(LocalSet* curr) {
...
 void BinaryInstWriter::visitGlobalGet(GlobalGet* curr) {
-  // Emit a global.get for each element if this is a tuple global
   Index index = parent.getGlobalIndex(curr->name);
+  if (auto it = extractedGets.find(curr); it != extractedGets.end()) {
+    // We have a tuple of globals to get, but we will only end up using one of
+    // them, so we can just emit that one.
+    o << int8_t(BinaryConsts::GlobalGet) << U32LEB(index + it->second);
+    return;
+  }
+  // Emit a global.get for each element if this is a tuple global
```
kripken, line 145: Wait, we allow tuple globals? Does the wasm spec?

src/wasm/wasm-stack.cpp
```diff
@@ -2532,6 +2564,15 @@ void BinaryInstWriter::countScratchLocals() {
   for (auto& [type, _] : scratchLocals) {
     noteLocalType(type);
   }
+  // While we have all the tuple.extracts, also find extracts of local.gets and
+  // local.tees that we can optimize.
```
kripken, line 2568: Comment could mention global gets too.

test/multivalue.wast
```diff
@@ -53,14 +53,16 @@
...
   (global.set $g1
    (tuple.make
     (i32.const 42)
     (i64.const 7)
    )
   )
-  (global.get $g2)
+  (tuple.extract 1
+   (global.get $g2)
+  )
```
kripken, line 65: Should we also have a test of a `global.get` of a tuple that does not flow into an `extract`?

kripken (commented):
Nice!

src/wasm/wasm-stack.cpp
tlively: Yes, we weren't originally going to support tuple globals, but when we first implemented all this tuple business we found that some passes needed them. I don't remember which ones those were, though.

The Wasm spec supports tuple globals only as much as it supports tuple locals, i.e. not at all. But that's ok!

tlively:
Rebased on top of a separate PR moving the tests to lit, which also made sure we are testing both an extract and non-extract use of a tuple global.get. The other comments have been addressed in the latest commit.

src/wasm/wasm-stack.cpp
kripken: Hmm... a little odd, but ok I guess :smile: I suppose the passes would need to create more globals or such, otherwise.

kripken (approved):

kripken:
Fuzz bug:
```wat
(module
 (type $0 (func (result i64 f32 i32 i32 i32)))
 (type $1 (func (result f32)))
 (func $0 (type $0) (result i64 f32 i32 i32 i32)
  (local $3 (i64 f32 i32 i32 i32))
  (local $4 f32)
  (local.set $3
   (tuple.make
    (i64.const 0)
    (call $1)
    (i32.const 0)
    (i32.const 0)
    (i32.const 0)
   )
  )
  (local.set $4
   (tuple.extract 1
    (local.get $3)
   )
  )
  (tuple.make
   (i64.const 0)
   (f32.const 0)
   (i32.const 0)
   (i32.const 0)
   (i32.const 0)
  )
 )
 (func $1 (type $1) (result f32)
  (f32.convert_i32_s
   (i32.const 0)
  )
 )
)
```
STR:
```
bin/wasm-opt w.wat  --gufa-cast-all  --generate-stack-ir --optimize-stack-ir --roundtrip --shrink-level=1 -all
```

tlively:
Oof, I'm not sure how we should handle locals that have been optimized out in the stack IR, actually. Do you have an idea?

kripken:
I think it might be tricky to combine the two. Really the tuples should be lowered before Stack IR (is in Poppy IR), to make this clean. But for now, how about disabling local2stack on tuple locals in StackIR? The benefit of that optimization is around 1-2% at best, while your benefit here is quite large.

PR #5944, opened 2023-09-14
Add passes to finalize or unfinalize types

### LINKED ISSUE #5933: Treatment of finality in TypeSSA is too conservative
When support for final types was introduced in https://github.com/WebAssembly/binaryen/commit/73f24c0af330578e9fa01e3d5fc6391619baaa9e, The TypeSSA pass was adjusted to not create subtypes for final types.

I believe this is more strict than it needs to be when using `--closed-world`.

In a closed-world scenario, the finality marker is redundant. There is no reason to set it to anything other than final if the type has no subtypes. Thus, the following strategy should be sound:
1. On input, ignore the finality markers and treat all types as non-final (in particular in TypeSSA).
2. On output, mark types final that have no subtypes.

To test this strategy, I did a comparison between the same module with and without finality markers on struct types, optimized with
```
wasm-opt -all --closed-world -tnh -O3 --type-ssa --gufa -O3 --type-merging -O1
```
(using https://github.com/WebAssembly/binaryen/commit/b9f4e3609312473315cc401a2d8501e7a7731b04 to avoid the issue reported in https://github.com/WebAssembly/binaryen/issues/5932)
| Module | Size | Struct types | `call` | `call_ref` | `call_indirect` |
| :- | -: | -: | -: | -: | -: |
| [With final](https://github.com/WebAssembly/binaryen/files/12584550/ff.wasm.gz) | 1540124 | 55859 | 23507 | 257 | 3604 |
| [Without final](https://github.com/WebAssembly/binaryen/files/12584483/fs.wasm.gz) | 1520024 | 55008 | 23540 | 250 | 3538 |

As can be seen, ignoring the finality markers will reduce the eventual number of struct types and also improve the devirtualization of `call_ref` and `call_indirect` into `call`.

kripken:
TypeFinalization finalizes all types that we can, that is, all private types that have no
children. TypeUnFinalization unfinalizes (opens) all (private) types.

These could be used by first opening all types, optimizing, and then finalizing, as that
might find more opportunities.

Fixes #5933

TODO: document in wiki

TODO: run in default pipeline?


src/passes/TypeFinalizing.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+  void run(Module* module) override {
+    if (!module->features.hasGC()) {
+      return;
+    }
+
+    // To make a type final, it must have no subtypes.
+    std::unique_ptr<SubTypes> subTypes;
+    if (finalize) {
+      subTypes = std::make_unique<SubTypes>(*module);
```
tlively, line 50:
```suggestion
    std::optional<SubTypes> subTypes;
    if (finalize) {
      subTypes = SubTypes(*module);
```

src/passes/TypeFinalizing.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include "wasm-type.h"
+#include "wasm.h"
+
+namespace wasm {
+
+namespace {
+
+struct TypeFinalizing : public Pass {
+  // Whether we should finalize or unfinalize types.
+  bool finalize;
```
tlively, line 35: I wonder if the pass is small enough that it would be simpler to break it up into two separate pass classes rather than switching on this bool everywhere.

tlively (approved):

src/passes/TypeFinalizing.cpp
kripken: It is short, yeah, but the TypeRewriter hook stuff is kind of complex, so I'd rather keep the code shorter by sharing more.

PR #5945, opened 2023-09-15
Revert "Optimize tuple.extract of gets in BinaryInstWriter"

tlively:
Reverts WebAssembly/binaryen#5941

The binary writer optimization is not always correct when stack IR optimizations have run. Revert the change until we can fix it.


kripken (approved):
lgtm for now, let's talk more about this next week

PR #5949, opened 2023-09-16
Implement table.fill

### LINKED ISSUE #5939: wasm-opt missing support for table.fill
While wasm-opt appears to support `table.grow` correctly, any compiled module containing a `table.fill` instruction will fail to parse. This is true even if you pass the appropriate `--enable-reference-types` flag.

Sample code:
```c
void
mono_jiterp_grow_function_table (int count, int fill_value) {
	__asm(
		// table.grow expects stack layout [fill_ref] [count]
		// start by loading fill_value, then fetching that reference from the indirect fn table
		"local.get 1;"
		"table.get __indirect_function_table;"
		// now fill_ref is on the stack, load count and perform the actual table grow
		"local.get 0;"
		"table.grow __indirect_function_table;"
		// table.grow returns either the new size or an error code. i'm too lazy to deal with it
		"drop;"
	);
}

#if 0
void
mono_jiterp_fill_function_table (int offset, int count, int fill_value) {
	__asm(
		// table.fill expects stack layout [offset] [fill_ref] [count]
		// load offset
		"local.get 0;"
		// load fill_value, then fetch that reference from the indirect fn table
		"local.get 2;"
		"table.get __indirect_function_table;"
		// now fill_ref is on the stack, load count and perform the actual table grow
		"local.get 1;"
		"table.fill __indirect_function_table;"
	);
}
#endif
```

change the `#if 0` to `#if 1` and the resulting built module will cause wasm-opt to fail with something like:
```
[parse exception: invalid code after misc prefix: 17 (at 0:266808)]
  Fatal: error parsing wasm (try --debug for more info)
emcc : error : '/home/kate/Projects/dotnet-runtime-wasm/src/mono/wasm/emsdk/upstream/bin/wasm-opt --strip-dwarf --post-emscripten -O2 --low-memory-unused --zero-filled-memory --pass-arg=directize-initial-contents-immutable --strip-debug --strip-producers dotnet.native.wasm -o dotnet.native.wasm -g --mvp-features --enable-mutable-globals --enable-reference-types --enable-sign-ext --enable-simd' failed (returned 1) [/home/kate/Projects/dotnet-runtime-wasm/src/mono/wasm/wasm.proj]
```

From checking the spec, `17` is `table.fill`. Looking through the repo it seems like support for `table.init` and `table.copy` might be missing too, but I'm not planning to use them...

tlively:
This instruction was standardized as part of the bulk memory proposal, but we
never implemented it until now. Leave similar instructions like table.copy as
future work.

Fixes #5939.


tlively:
* main
  * **PR #5949** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5949" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5949?utm_source=stack-comment).

kripken (approved):

PR #5952, opened 2023-09-18
Make heap2local work through casts

vouillon:
I currently generate boxing and unboxing code like the following:
```wat
(local $x (ref eq)
...
(local.set $x
  (struct.new $float
    ...
  )
)
(struct.get $float 0
  (ref.cast (ref $float)
    (local.get $x)
  )
)
```
This PR allows them to cancel out.

The heap2local pass is performed before the optimize-casts pass [the local-subtyping pass, actually], which would have removed the cast. I first considered modifying the optimize-instruction pass, similarly to what is done for tuples in #5938, but this is more general.

It might make sense to handle tuples in a similar manner, by the way.

### DIFF
--- a/src/passes/Heap2Local.cpp
+++ b/src/passes/Heap2Local.cpp
@@ -430,6 +430,18 @@ struct Heap2LocalOptimizer {
       replaceCurrent(curr->value);
     }
 
+    void visitRefCast(RefCast* curr) {
+      if (!reached.count(curr)) {
+        return;
+      }
+
+      // It is safe to optimize out this RefCast, since we proved it
+      // contains our allocation and we have checked that the type of
+      // the allocation is a subtype of the type of the cast, and so
+      // cannot trap.
+      replaceCurrent(curr->ref);
+    }
+
     void visitStructSet(StructSet* curr) {
       if (!reached.count(curr)) {
         return;
@@ -526,7 +538,7 @@ struct Heap2LocalOptimizer {
         return;
       }
 
-      switch (getParentChildInteraction(parent, child)) {
+      switch (getParentChildInteraction(allocation, parent, child)) {
         case ParentChildInteraction::Escapes: {
           // If the parent may let us escape then we are done.
           return;
@@ -586,7 +598,8 @@ struct Heap2LocalOptimizer {
     rewriter.applyOptimization();
   }
 
-  ParentChildInteraction getParentChildInteraction(Expression* parent,
+  ParentChildInteraction getParentChildInteraction(StructNew* allocation,
+                                                   Expression* parent,
                                                    Expression* child) {
     // If there is no parent then we are the body of the function, and that
     // means we escape by flowing to the caller.
@@ -595,6 +608,7 @@ struct Heap2LocalOptimizer {
     }
 
     struct Checker : public Visitor<Checker> {
+      StructNew* allocation;
       Expression* child;
 
       // Assume escaping (or some other problem we cannot analyze) unless we are
@@ -647,6 +661,15 @@ struct Heap2LocalOptimizer {
         }
       }
 
+      void visitRefCast(RefCast* curr) {
+        // As it is our allocation that flows through here, we need to
+        // check that the cast will not trap, so that we can continue
+        // to (hopefully) optimize this allocation.
+        if (Type::isSubType(allocation->type, curr->type)) {
+          escapes = false;
+        }
+      }
+
       // GC operations.
       void visitStructSet(StructSet* curr) {
         // The reference does not escape (but the value is stored to memory and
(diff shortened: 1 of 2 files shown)


test/lit/passes/heap2local.wast
```diff
@@ -1945,4 +1980,56 @@
...
+  ;; CHECK-NEXT:     (local.get $1)
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:    (ref.null none)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (local.get $0)
+  ;; CHECK-NEXT: )
+  (func $cast-success (result anyref)
+    (struct.get $A 0
+      (ref.cast (ref $A)
```
kripken, line 2004: This cast will be removed by other optimization passes (the cast can simply be removed), so while it's a good testcase I'm still uncertain I understand the situation in which this PR helps. The example from the PR description would also be handled by other opts (the local type would be refined).

Is there a case other optimizations can't help with?

kripken: (To be clear, I do think it's very important we optimize these cases! It just seems like we might already do so.)

vouillon: You're right. This PR just makes the order in which passes are performed less relevant.
Currently, to get this optimization, since `heap2local` is performed before the casts are removed, i can repeat the `-O2` or `-O3` option (which can be slow on large programs) or perform the following passes at the end: `--heap2local --simplify-locals --reorder-locals --vacuum`.

kripken: I see, thanks. In that case, perhaps we can add another run of `heap2local` in the main pipeline, after the casts are removed. We should measure that this isn't a significant regression in compile time, but otherwise we usually prefer that as a solution to such issues rather than add code complexity.

vouillon: I understand. I think one needs this sequence of passes: `local-subtyping`, `optimize-instructions`, `heap2local`,  `simplify-locals`, `reorder-locals`,  `vacuum`. So, several passes need to be added, but they look cheap.

Note that this PR does not much more than duplicating the existing code that handles `ref.as_non_null`.
https://github.com/WebAssembly/binaryen/blob/8c7ae2d9413fc6880da195b483c5a4e111c191fa/src/passes/Heap2Local.cpp#L649-L650

vouillon: No problem! Thanks! I guess I should have been clearer about that from the start...

src/passes/Heap2Local.cpp
```diff
@@ -430,6 +430,16 @@ struct Heap2LocalOptimizer {
       replaceCurrent(curr->value);
     }
 
+    void visitRefCast(RefCast* curr) {
+      if (!reached.count(curr)) {
+        return;
+      }
+
+      // It is safe to optimize out this RefCast, since we proved it
+      // contains our allocation, and so cannot trap.
```
kripken, line 439: This comment is identical to the above for RefAs, but I think we need more here. That the RefAsNonNull receives an allocation is enough to prove it is not null, but here we need the extra type checking on line 663. So perhaps add `, and we verified already that it receives the correct type`.

src/passes/Heap2Local.cpp
```diff
@@ -647,6 +659,12 @@ struct Heap2LocalOptimizer {
         }
       }
 
+      void visitRefCast(RefCast* curr) {
+        if (Type::isSubType(allocation->type, curr->type)) {
```
kripken, line 663:
```suggestion
        // We only accept a value of the allocation's type (so that we know this
        // cast is removable later).
        if (Type::isSubType(allocation->type, curr->type)) {
```

test/lit/passes/heap2local.wast
```diff
@@ -1945,4 +1980,56 @@
...
+  ;; CHECK-NEXT:     (struct.new $A
+  ;; CHECK-NEXT:      (ref.null none)
+  ;; CHECK-NEXT:     )
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
+  (func $cast-failure (result anyref)
+    (struct.get $B 0
+      (ref.cast (ref $B)
```
kripken, line 2026:
```suggestion
      ;; The allocated $A arrives here, but the cast will fail, so we do not optimize.
      (ref.cast (ref $B)
```

kripken (approved):
lgtm with some comments + fuzzing.

For fuzzing, maybe run `./scripts/fuzz_opt.py` for a few thousand iterations, to be safe.

src/passes/Heap2Local.cpp
kripken: Or, rather, it's not about being removable later, but if we cast to something entirely different then we can't access the values of that thing (we only know the layout of the allocation), but the cast will trap anyhow, so we stop.

kripken (approved):

PR #5955, opened 2023-09-18
Reland "Optimize tuple.extract of gets in BinaryInstWriter"

tlively:
Reland "Optimize tuple.extract of gets in BinaryInstWriter"

In general, the binary lowering of tuple.extract expects that all the tuple
values are on top of the stack, so it inserts drops and possibly uses a scratch
local to ensure only the extracted value is left. However, when the extracted
tuple expression is a local.get, local.tee, or global.get, it's much more
efficient to change the lowering of the get or tee to ensure that only the
extracted value is on the stack to begin with. Implement that optimization in
the binary writer.

Also fix a bug in which the new optimization interacted poorly with StackIR
local2stack optimization by ignoring tuple locals in local2stack.

fix stack ir bug


tlively:
* main
  * **PR #5955** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5955" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5955?utm_source=stack-comment).

kripken:
What do you think about a separate PR first to disable tuple locals in local2stack? Then we can test that specifically.

kripken (approved):

PR #5956, opened 2023-09-18
TupleOptimization: Handle copies of different types in unreachable code

kripken:


tlively (approved):
It's a good thing we don't ever emit unreachable code into a binary because this would make a real validator very unhappy!

kripken:
With unreachable code, the only winning game is not to play.

kripken:
Fuzzing looks good, but it did take 600K iterations to find this bug, so we'll see...

PR #5959, opened 2023-09-18
Fix visitBlock and add visitBlockStart in IRBuilder

tlively:
Visiting a block should push it onto the stack just like visiting any other
expression, but we previously had a `visitBlock` that introduced a new scope
instead. Fix `visitBlock` to behave as expected and introduce a new
`visitBlockStart` method to introduce a new scope.

Unfortunately this cannot be fully tested yet because the wat parser uses the
`makeXYZ` API intead of the `visit` API, but at least I updated `makeBlock` to
call `visitBlockStart`, so that is tested.

### DIFF
--- a/src/wasm-ir-builder.h
+++ b/src/wasm-ir-builder.h
@@ -45,15 +45,13 @@ public:
   [[nodiscard]] Result<Expression*> build();
 
   // Call visit() on an existing Expression with its non-child fields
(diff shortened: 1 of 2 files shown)


ashleynh (approved):

ashleynh:
Nice header changes 👍🏽, thanks!

PR #5961, opened 2023-09-19
[NFC] RemoveUnusedModuleElements: Use delegations

kripken:
NFC, but fixes a current fuzz bug on `table.fill` not having an entry in this file. After
this PR, there is no need for such entries.


tlively (approved):
Nice 🎉

kripken:
10K fuzzer iters look good...

PR #5963, opened 2023-09-20
[Parser] Parse if-else in the new wat parser and IRBuilder

tlively:
Parse both the straight-line and folded versions of if, including the
abbreviations that allow omitting the else clause. In the IRBuilder, generalize
the scope stack to be able to track scopes other than blocks and add methods for
visiting the beginnings of ifs and elses.

### DIFF
--- a/scripts/gen-s-parser.py
+++ b/scripts/gen-s-parser.py
@@ -712,6 +712,8 @@ def instruction_parser(new_parser=False):
     trie = Node()
     inst_length = 0
     for inst, expr in instructions:
+        if new_parser and inst in {"then", "else"}:
+            continue
         inst_length = max(inst_length, len(inst))
         trie.insert(inst, expr)
 
--- a/scripts/gen-s-parser.py
+++ b/scripts/gen-s-parser.py
@@ -713,6 +713,7 @@ def instruction_parser(new_parser=False):
     inst_length = 0
     for inst, expr in instructions:
         if new_parser and inst in {"then", "else"}:
+            # These are not real instructions! skip them.
             continue
         inst_length = max(inst_length, len(inst))
         trie.insert(inst, expr)
--- a/src/parser/contexts.h
+++ b/src/parser/contexts.h
@@ -312,7 +312,12 @@ struct NullInstrParserCtx {
   InstrT makeBlock(Index, std::optional<Name>, BlockTypeT) {
     return Ok{};
   }
-  InstrT finishBlock(Index, InstrsT) { return Ok{}; }
+  template<typename BlockTypeT>
+  InstrT makeIf(Index, std::optional<Name>, BlockTypeT) {
+    return Ok{};
+  }
+  InstrT visitEnd(Index, InstrsT) { return Ok{}; }
+  InstrT visitElse(Index) { return Ok{}; }
 
   InstrT makeUnreachable(Index) { return Ok{}; }
   InstrT makeNop(Index) { return Ok{}; }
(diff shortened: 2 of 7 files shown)


tlively:
* main
  * **PR #5963** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5963" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5963?utm_source=stack-comment).

kripken (commented):
Can you explain the folded (required, not required etc.) stuff?

tlively:
> Can you explain the folded (required, not required etc.) stuff?

We have normal `instrs(Ctx&)` that parses an arbitrary sequence of instructions, including both the folded and unfolded formats, even when the formats are interleaved. The text format for folded ifs requires that the condition be a folded instruction, though, so it is not allowed to include unfolded instructions or multiple top-level folded instructions. It would be cleaner overall to split the parsing of a single tree of folded expressions into a separate function, but that would require a larger refactoring. I'm working on it as a follow-up.

scripts/gen-s-parser.py
```diff
@@ -712,6 +712,8 @@ def instruction_parser(new_parser=False):
     trie = Node()
     inst_length = 0
     for inst, expr in instructions:
+        if new_parser and inst in {"then", "else"}:
+            continue
```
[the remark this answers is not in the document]
kripken: Got it, thanks. Maybe a comment here?

src/parser/parsers.h
```diff
@@ -48,11 +48,14 @@ MaybeResult<typename Ctx::InstrT> unfoldedBlockinstr(Ctx&);
 template<typename Ctx> MaybeResult<typename Ctx::InstrT> blockinstr(Ctx&);
 template<typename Ctx> MaybeResult<typename Ctx::InstrT> plaininstr(Ctx&);
 template<typename Ctx> MaybeResult<typename Ctx::InstrT> instr(Ctx&);
-template<typename Ctx> Result<typename Ctx::InstrsT> instrs(Ctx&);
+template<typename Ctx>
+Result<typename Ctx::InstrsT> instrs(Ctx&, bool requireFolded = false);
```
kripken, line 52: Maybe this bool could be an enum?

kripken (approved):
lgtm with whichever of the last suggestions makes sense to you.

src/wasm/wasm-ir-builder.cpp
```diff
@@ -282,18 +282,26 @@ Result<> IRBuilder::visitArrayNew(ArrayNew* curr) {
...
+Result<> IRBuilder::visitIfStart(If* iff, Name label) {
+  auto cond = pop();
+  CHECK_ERR(cond);
+  iff->condition = *cond;
+  scopeStack.push_back(ScopeCtx::makeIf(iff, label));
+  return Ok{};
+}
+
+Result<Expression*> IRBuilder::finishScope(Block* block) {
+  if (scopeStack.empty() || scopeStack.back().isNone()) {
```
ashleynh, line 298: So in the case where I want to outline a sequence of 4 instructions, let's say all of them i32.const, I would need to call makeBlockStart() first? To be on the safe side, should I always intend to create a block for the sequence to be outlined so that the scopeStack is not empty (or doesn't have a last value of none)?

ashleynh (approved):

src/parser/contexts.h
```diff
@@ -994,10 +999,20 @@ struct ParseDefsCtx : TypeParserCtx<ParseDefsCtx> {
...
-  Result<> finishBlock(Index pos, InstrsT) {
+  Result<> makeIf(Index pos, std::optional<Name> label, HeapType type) {
+    // TODO: validate labels?
+    // TODO: Move error on input types to here?
+    return withLoc(
+      pos,
+      irBuilder.makeIf(label ? *label : Name{}, type.getSignature().results));
+  }
+
+  Result<> visitEnd(Index pos, InstrsT) {
```
ashleynh, line 1010: same point Alon made above, move visitEnd to be after makeBlock?

src/wasm/wasm-ir-builder.cpp
tlively: If you want those four i32.const instructions to be inside a block, then you would call either `visitBlockStart` or `makeBlock`, then `visit` on each of the i32.const instructions, then `visitEnd`. But if you're outlining the four instructions, then you may not know in advance whether you need a block, since it depends on whether you end up with one top-level instruction or multiple 🤔 

I think the best thing for me to do will be to add function scope contexts (which I'm realizing we need anyway for branches to top-level function scopes) so you can call `visitFunctionStart` at the beginning of a function (instead of `setFunction`) and `visitEnd` at the end of the function, which will then handle setting the function body for you.

src/parser/contexts.h
tlively: I think I ended up doing this later in this change of PRs, so I'll leave it for now to reduce churn.

PR #5968, opened 2023-09-21
NameTypes and TypeSSA : Prefer _ over $ in names, and lint away _N suffixes

kripken:
Apparently `$N` (e.g. `FooClass$5`) is a convention in Java for anonymous classes, so our
`$N` that we use to disambiguate could be confusing. As the way we disambiguate does
not matter, switch to using `_N`. This PR does that in both `TypeSSA` and `NameTypes`.

Also make `NameTypes` "lint" names as it goes. That pass tries to give types nice names,
leaving existing ones that seem ok, and renaming long or unnamed ones. This PR makes
it aware of the `_N` notation and it tries to remove it, if removing it does not cause a
collision. An example of how that helps is if TypeSSA creates a subtype `$Foo_0` and then
we manage to remove `$Foo`, then we can use the shorter name for the subtype.

### DIFF
--- a/src/passes/NameTypes.cpp
+++ b/src/passes/NameTypes.cpp
@@ -50,7 +50,7 @@ struct NameTypes : public Pass {
     // "Lint" the names a little. In particular a name with a "_7" or such
     // suffix, as TypeSSA creates, can be removed if it does not cause a
     // collision. This keeps the names unique while removing 'noise.'
-    for (auto& [type, names] : module->typeNames) {
+    for (auto& [_, names] : module->typeNames) {
       std::string name = names.name.toString();
       while (name.size() > 1 && isdigit(name.back())) {
         name.pop_back();
--- a/src/passes/NameTypes.cpp
+++ b/src/passes/NameTypes.cpp
@@ -34,13 +34,33 @@ struct NameTypes : public Pass {
     // Find all the types.
     std::vector<HeapType> types = ModuleUtils::collectHeapTypes(*module);
 
+    std::unordered_set<Name> used;
+
     // Ensure simple names. If a name already exists, and is short enough, keep
     // it.
     size_t i = 0;
     for (auto& type : types) {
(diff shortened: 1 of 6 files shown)


kripken:
Hmm, test error looks like nondeterminism, so there is more work to be done here.

kripken:
Looks ready now.

src/passes/NameTypes.cpp
```diff
@@ -34,13 +34,36 @@ struct NameTypes : public Pass {
...
+    std::unordered_set<Name> used;
+
     // Ensure simple names. If a name already exists, and is short enough, keep
     // it.
     size_t i = 0;
     for (auto& type : types) {
       if (module->typeNames.count(type) == 0 ||
           module->typeNames[type].name.size() >= NameLenLimit) {
-        module->typeNames[type].name = "type$" + std::to_string(i++);
+        module->typeNames[type].name = "type_" + std::to_string(i++);
```
tlively, line 45: I guess `type_0` will be linted to `type`. Is it worth special-casing "type" to avoid that?

tlively (approved):
LGTM % comment

src/passes/NameTypes.cpp
kripken: I don't see why? I feel like I'm not seeing what you find wrong with that linting...

tlively: I guess it's a more general QoL issue: if there is `somename_X` and `somename_Y`, there's not a principled way to determine which one deserves to be improved to `somename`. I would have a slight preference for improving neither in that case, but improving one arbitrarily also seems reasonable.

kripken: I see now, thanks. Yeah, maybe in some cases it might be better to avoid linting either name. I think in most cases however it will be worth it. Anyhow, if there's feedback on this we can iterate more later.

PR #5970, opened 2023-09-25
[Parser] Parse labels and br

tlively:
The parser previously parsed labels and could attach them to control flow
structures, but did not maintain the context necessary to correctly parse
branches. Support parsing labels as both names and indices in IRBuilder,
handling shadowing correctly, and use that support to implement parsing of br.

### DIFF
--- a/src/parser/contexts.h
+++ b/src/parser/contexts.h
@@ -940,9 +940,7 @@ struct ParseDefsCtx : TypeParserCtx<ParseDefsCtx> {
     return name;
   }
 
-  Result<Index> getLabelFromIdx(uint32_t idx) {
-    return idx;
-  }
+  Result<Index> getLabelFromIdx(uint32_t idx) { return idx; }
 
   Result<Index> getLabelFromName(Name name) {
     return irBuilder.getLabelIndex(name);
--- a/src/parser/contexts.h
+++ b/src/parser/contexts.h
@@ -279,6 +279,7 @@ struct NullInstrParserCtx {
   using GlobalIdxT = Ok;
   using MemoryIdxT = Ok;
   using DataIdxT = Ok;
+  using LabelIdxT = Ok;
 
   using MemargT = Ok;
 
@@ -298,6 +299,8 @@ struct NullInstrParserCtx {
   MemoryIdxT getMemoryFromName(Name) { return Ok{}; }
   DataIdxT getDataFromIdx(uint32_t) { return Ok{}; }
   DataIdxT getDataFromName(Name) { return Ok{}; }
+  LabelIdxT getLabelFromIdx(uint32_t) { return Ok{}; }
+  LabelIdxT getLabelFromName(Name) { return Ok{}; }
 
   MemargT getMemarg(uint64_t, uint32_t) { return Ok{}; }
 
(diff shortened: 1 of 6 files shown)


tlively:
* main
  * **PR #5970** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5970" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5970?utm_source=stack-comment).

tlively:
@ashleynh, would you have time to review this today (Friday)? Happy to answer any questions you have about the code.

src/parser/contexts.h
```diff
@@ -936,6 +940,12 @@ struct ParseDefsCtx : TypeParserCtx<ParseDefsCtx> {
     return name;
   }
 
+  Result<Index> getLabelFromIdx(uint32_t idx) { return idx; }
```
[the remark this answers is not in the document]
tlively: Yep, pretty much. The general pattern here is that the parser functions have `Ctx` template parameters and whenever they need to actually get a representation of some parsed data, they use representation types defined in the `Ctx` and call methods defined in the `Ctx`. In this `ParseDefsCtx`, we represent label indexes as `Index`, but up above in the `NullInstrParserCtx` we represent them as `Ok`, i.e. we don't really represent them at all. Since the representations are different for different parser contexts, we need to call into the contexts to produce them, even if that's trivial. We definitely hope that the compiler will inline these calls.

src/wasm-ir-builder.h
```diff
@@ -310,12 +349,24 @@ class IRBuilder : public UnifiedExpressionVisitor<IRBuilder, Result<>> {
     return scopeStack.back();
   }
 
+  Result<ScopeCtx*> getScope(Index label) {
+    Index numLabels = scopeStack.size();
+    if (!scopeStack.empty() && scopeStack[0].isNone()) {
```
ashleynh, line 354: What are the cases when a ScopeCtx has a type of None?

src/wasm-ir-builder.h
```diff
@@ -211,25 +220,29 @@ class IRBuilder : public UnifiedExpressionVisitor<IRBuilder, Result<>> {
...
       return ScopeCtx(BlockScope{block});
     }
-    static ScopeCtx makeIf(If* iff, Name label = {}) {
-      return ScopeCtx(IfScope{iff, label});
+    static ScopeCtx makeIf(If* iff, Name originalLabel = {}) {
+      return ScopeCtx(IfScope{iff, originalLabel});
     }
-    static ScopeCtx makeElse(If* iff, Name label = {}) {
-      return ScopeCtx(ElseScope{iff, label});
+    static ScopeCtx makeElse(If* iff, Name originalLabel, Name label) {
```
ashleynh, line 244: what's the relationship between the originalLabel and the label?

ashleynh (approved):

src/wasm-ir-builder.h
tlively: Yes, exactly 👍

src/wasm-ir-builder.h
tlively: This would be the case when parsing a global initializer or any other code that does not appear inside a function.

PR #5971, opened 2023-09-25
Outlining Test Improvements

ashleynh:
Abstracts repeat code

### DIFF
--- a/test/gtest/stringify.cpp
+++ b/test/gtest/stringify.cpp
@@ -2,6 +2,8 @@
 #include "passes/stringify-walker.h"
 #include "print-test.h"
 #include "support/suffix_tree.h"
+#include "wasm-io.h"
+#include <iostream>
 
 using namespace wasm;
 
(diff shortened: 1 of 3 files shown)


test/gtest/stringify.cpp
```diff
@@ -161,14 +161,18 @@ static auto dupModuleText = R"wasm(
    )
   )wasm";
 
+std::vector<uint32_t> hashStringifyModule(Module* wasm) {
+  HashStringifyWalker stringify = HashStringifyWalker();
+  stringify.walkModule(wasm);
+  return std::move(stringify.hashString);
```
kripken, line 167: Is the `move` here just for efficiency? Then as this is test code it's probably fine without it to keep things simpler.

(If we did care about speed we could also use `move` below in `repeatSubstrings` but it feels unnecessary in both places, to me.)

kripken (approved):

test/gtest/stringify.cpp
ashleynh: ah, thanks for noticing! I do want it in both places :)

ashleynh: oh, interesting. when I do std::move in repeatSubstrings, I get `error: moving a local object in a return statement prevents copy elision`

PR #5972, opened 2023-09-25
[Outlining] Adds SuffixTree::RepeatSubstring dedupe test

ashleynh:
This PR adds a StringProcessor struct intended to hold functions that filter vectors of SuffixTree::RepeatedSubstring, and a test of its first functionality, removing overlapping repeated substrings.

### DIFF
--- a/src/passes/CMakeLists.txt
+++ b/src/passes/CMakeLists.txt
@@ -11,13 +11,8 @@ string(APPEND WASM_INTRINSICS_EMBED "0x00")
 
 configure_file(WasmIntrinsics.cpp.in WasmIntrinsics.cpp @ONLY)
 
-set(passes_HEADERS
-  call-utils.h
-  intrinsics-module.h
-  opt-utils.h
-  param-utils.h
-  passes.h
-)
+FILE(GLOB passes_HEADERS *.h)
+
 set(passes_SOURCES
   param-utils.cpp
   pass.cpp
@@ -49,6 +44,7 @@ set(passes_SOURCES
   GlobalStructInference.cpp
   GlobalTypeOptimization.cpp
   GUFA.cpp
+  hash-stringify-walker.cpp
   Heap2Local.cpp
   I64ToI32Lowering.cpp
   Inlining.cpp
@@ -129,13 +125,8 @@ set(passes_SOURCES
 # suffix_tree_node source files no longer depend on LLVM code in the
 # third_party folder
 if(EMSCRIPTEN)
-  add_library(passes OBJECT ${passes_SOURCES})
-else()
-  set(passes_with_suffix_tree_SOURCES
-    stringify-walker.h
-    stringify-walker-impl.h
-  	hash-stringify-walker.cpp
-	${passes_SOURCES}
-  )
-  add_library(passes OBJECT ${passes_with_suffix_tree_SOURCES})
+  list(REMOVE_ITEM passes_SOURCES ${CMAKE_CURRENT_BINARY_DIR}/stringify-walker.h)
+  list(REMOVE_ITEM passes_SOURCES ${CMAKE_CURRENT_BINARY_DIR}/stringify-walker-impl.h)
+  list(REMOVE_ITEM passes_SOURCES "hash-stringify-walker.cpp")
 endif()
+add_library(passes OBJECT ${passes_SOURCES})
--- a/src/passes/CMakeLists.txt
+++ b/src/passes/CMakeLists.txt
@@ -11,7 +11,13 @@ string(APPEND WASM_INTRINSICS_EMBED "0x00")
 
 configure_file(WasmIntrinsics.cpp.in WasmIntrinsics.cpp @ONLY)
 
-FILE(GLOB passes_HEADERS *.h)
+set(passes_HEADERS
+  call-utils.h
+  intrinsics-module.h
+  opt-utils.h
+  param-utils.h
+  passes.h
+)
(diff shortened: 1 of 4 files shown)


src/passes/CMakeLists.txt
```diff
@@ -11,7 +11,13 @@ string(APPEND WASM_INTRINSICS_EMBED "0x00")
 
 configure_file(WasmIntrinsics.cpp.in WasmIntrinsics.cpp @ONLY)
 
-FILE(GLOB passes_HEADERS *.h)
```
tlively, line 14: It might be slightly nicer to keep using the glob up front, then use REMOVE_ITEM to remove selected headers in the emscripten build.

src/passes/CMakeLists.txt
```diff
@@ -120,4 +125,17 @@ set(passes_SOURCES
...
+# The below condition is intended for removal once the suffix_tree and
+# suffix_tree_node source files no longer depend on LLVM code in the
+# third_party folder
+if(EMSCRIPTEN)
+  add_library(passes OBJECT ${passes_SOURCES})
+else()
+  set(passes_with_suffix_tree_SOURCES
+    stringify-walker.h
+    stringify-walker-impl.h
+  	hash-stringify-walker.cpp
```
tlively, line 137: Indentation looks strange here.

src/passes/hash-stringify-walker.cpp
```diff
@@ -68,4 +68,33 @@ void HashStringifyWalker::visitExpression(Expression* curr) {
   }
 }
 
+std::vector<SuffixTree::RepeatedSubstring> StringifyProcessor::dedupe(
+  const std::vector<SuffixTree::RepeatedSubstring> substrings) {
+  std::set<uint32_t> seen;
```
tlively, line 73:
```suggestion
  std::unordered_set<uint32_t> seen;
```
Using an unordered_set (i.e. a hash set) should be faster when iteration order doesn't matter.

src/passes/hash-stringify-walker.cpp
```diff
@@ -68,4 +68,33 @@ void HashStringifyWalker::visitExpression(Expression* curr) {
...
+    for (auto startIdx : substring.StartIndices) {
+      // We are using the end index to ensure that each repeated substring
+      // reported by the SuffixTree is unique. This is because LLVM's SuffixTree
+      // reports back repeat sequences that are substrings of longer repeat
+      // sequences with the same endIdx, and we generally prefer to outline
+      // longer repeat sequences.
+      uint32_t endIdx = substring.Length + startIdx;
+      if (auto found = seen.find(endIdx); found != seen.end()) {
+        seenEndIdx = true;
+        continue;
```
tlively, line 87: It looks like this can be a `break` since `idxToInsert` is not used if we arrive here.

src/passes/hash-stringify-walker.cpp
```diff
@@ -68,4 +68,33 @@ void HashStringifyWalker::visitExpression(Expression* curr) {
...
+    std::vector<uint32_t> idxToInsert;
+    bool seenEndIdx = false;
+    for (auto startIdx : substring.StartIndices) {
+      // We are using the end index to ensure that each repeated substring
+      // reported by the SuffixTree is unique. This is because LLVM's SuffixTree
+      // reports back repeat sequences that are substrings of longer repeat
+      // sequences with the same endIdx, and we generally prefer to outline
+      // longer repeat sequences.
+      uint32_t endIdx = substring.Length + startIdx;
+      if (auto found = seen.find(endIdx); found != seen.end()) {
```
tlively, line 85:
```suggestion
      if (seen.count(endIdx)) {
```

src/passes/CMakeLists.txt
ashleynh: done

src/passes/hash-stringify-walker.cpp
ashleynh: done

src/passes/hash-stringify-walker.cpp
ashleynh: done

src/passes/CMakeLists.txt
ashleynh: done

src/passes/hash-stringify-walker.cpp
ashleynh: done

src/passes/hash-stringify-walker.cpp
```diff
@@ -68,4 +68,33 @@ void HashStringifyWalker::visitExpression(Expression* curr) {
   }
 }
 
+std::vector<SuffixTree::RepeatedSubstring> StringifyProcessor::dedupe(
```
tlively, line 71: It might be useful to add a comment summarizing the algorithm here and its properties. Something like this (if I'm understanding what the code does correctly):

"Deduplicate substrings by iterating through the list of substrings, keeping only those whose list of end indices is disjoint from the set of end indices for all substrings kept so far. Substrings that are contained within other substrings will always share an end index with those other substrings. Note that this deduplication may be over-aggressive, since it will remove substrings that are contained within any previous substring, even if they have many other occurrences that are not inside other substrings."

tlively (approved):

src/passes/hash-stringify-walker.cpp
ashleynh: done

PR #5980, opened 2023-09-27
Add more context to errors in legacy wat parser

tlively:
To make the error messages more useful, especially when parsing extremely large
wat files, print the erroneous fragments along with the existing error messages.

### DIFF
--- a/src/wasm/wasm-s-parser.cpp
+++ b/src/wasm/wasm-s-parser.cpp
@@ -27,6 +27,8 @@
 #include "wasm-binary.h"
 #include "wasm-builder.h"
 
+using namespace std::string_literals;
+
 #define abort_on(str)                                                          \
   { throw ParseException(std::string("abort_on ") + str); }
 #define element_assert(condition)                                              \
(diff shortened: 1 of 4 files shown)


tlively:
* main
  * **PR #5980** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5980" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5980?utm_source=stack-comment).

src/wasm/wasm-s-parser.cpp
```diff
@@ -516,7 +524,8 @@ Name SExpressionWasmBuilder::getElemSegmentName(Element& s) {
     // index
     size_t offset = parseIndex(s);
     if (offset >= elemSegmentNames.size()) {
-      throw ParseException("unknown elem segment", s.line, s.col);
+      throw ParseException(
+        "unknown elem segment"s + s.toString(), s.line, s.col);
```
kripken, line 528:
```suggestion
        "unknown elem segment: "s + s.toString(), s.line, s.col);
```

src/wasm/wasm-s-parser.cpp
```diff
@@ -503,7 +510,8 @@ Name SExpressionWasmBuilder::getTableName(Element& s) {
     // index
     size_t offset = parseIndex(s);
     if (offset >= tableNames.size()) {
-      throw ParseException("unknown table in getTableName", s.line, s.col);
+      throw ParseException(
+        "unknown table in getTableName"s + s.toString(), s.line, s.col);
```
kripken, line 514:
```suggestion
        "unknown table in getTableName: "s + s.toString(), s.line, s.col);
```

kripken (commented):
Would it be simpler perhaps to give `ParseException` the `Element& s` as a parameter, instead of `line,col`, and then it could append both the text and the `line,col`?

tlively:
> Would it be simpler perhaps to give `ParseException` the `Element& s` as a parameter, instead of `line,col`, and then it could append both the text and the `line,col`?

On the one hand, yes, but on the other hand, I don't want to touch this EOL code more 😢

kripken:
Would you like me to look into it? I wouldn't mind.

kripken:
Quick sketch of what I mean: https://gist.github.com/kripken/a8493f8944e2fc5d8882fdff92f55e22

Happy to fill that out into a PR, if you don't object.

tlively:
Perfect, thanks, that would be great.

PR #5981, opened 2023-09-28
wasm-s-parser: Add context in validation errors

kripken:
Instead of just reporting the reason and line + column, also log out the element
the error occurred at.

Supercedes #5980


tlively (approved):
Looks great, thanks!

PR #5982, opened 2023-09-29
Add an "unsubtyping" optimization

tlively:
Add a new pass that analyzes the module to find the minimal subtyping relation
that is necessary to maintain the validity and semantics of the program and
rewrites the types to use this minimal relation. Besides eliminating references
to otherwise-unused intermediate types, this optimization should unlock
significant additional optimizing power in other type optimizations that are
constrained by having to maintain supertype validity, since after this new
optimization there are fewer and more general supertypes.

The analysis works by visiting each expression and module element to collect the
subtypings that are required to maintain its validity, then using that as a
starting point, iteratively adding new subtypings required by type definitions
and casts until reaching a fixed point.


tlively:
* main
  * **PR #5982** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5982" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5982?utm_source=stack-comment).

tlively:
One of the new test cases is triggering a bug in the fuzzer with seed 15683576986865673243, but the optimization itself is looking pretty good.

src/passes/Unsubtyping.cpp
```diff
@@ -0,0 +1,482 @@
...
+#include "wasm.h"
+
+namespace wasm {
+
+namespace {
+
+struct Unsubtyping
+  : WalkerPass<PostWalker<Unsubtyping, OverriddenVisitor<Unsubtyping>>> {
+
+  // The original type depths, as computed by SubTypes.
```
kripken, line 35:
```suggestion
  // The original type depths, as computed by SubTypes (i.e., how deep the children are).
```

src/passes/Unsubtyping.cpp
```diff
@@ -0,0 +1,482 @@
...
+  // Map from cast source types to their destinations.
+  std::unordered_map<HeapType, std::unordered_set<HeapType>> castTypes;
+
+  void run(Module* wasm) override {
+    if (!wasm->features.hasGC()) {
+      return;
+    }
+    typeDepths = SubTypes(*wasm).getMaxDepths();
+    analyzePublicTypes(*wasm);
+    walkModule(wasm);
```
kripken, line 55: This will walk the module using a single core - probably worth optimizing that?

src/passes/Unsubtyping.cpp
```diff
@@ -0,0 +1,482 @@
...
+    analyzeTransitiveDependencies();
+    optimizeTypes(*wasm);
+    // Cast types may be refinable if their source and target types are no
+    // longer related. TODO: Experiment with running this only after checking
+    // whether it is necessary.
+    ReFinalize().run(getPassRunner(), wasm);
+  }
+
+  void noteSubtype(HeapType sub, HeapType super) {
```
kripken, line 64:
```suggestion
  // Notes that sub must be a subtype of super.
  void noteSubtype(HeapType sub, HeapType super) {
```
(is that right?)

src/passes/Unsubtyping.cpp
```diff
@@ -0,0 +1,482 @@
...
+    assert(!src.isTuple() && !dest.isTuple());
+    if (src == Type::unreachable) {
+      return;
+    }
+    assert(src.isRef() && dest.isRef());
+    noteCast(src.getHeapType(), dest.getHeapType());
+  }
+
+  void analyzePublicTypes(Module& wasm) {
+    for (auto type : ModuleUtils::getPublicHeapTypes(wasm)) {
```
kripken, line 127:
```suggestion
    // We cannot change supertypes for anything public.
    for (auto type : ModuleUtils::getPublicHeapTypes(wasm)) {
```

src/passes/Unsubtyping.cpp
```diff
@@ -0,0 +1,482 @@
...
+      noteSubtype(seg->type, getModule()->getTable(seg->table)->type);
+    }
+    for (auto init : seg->data) {
+      noteSubtype(init->type, seg->type);
+    }
+  }
+  void visitNop(Nop* curr) {}
+  void visitBlock(Block* curr) {
+    if (curr->name) {
+      BranchUtils::BranchSeeker seeker(curr->name);
```
kripken, line 239: This will be quadratic. Maintaining a stack of blocks would be more efficient. Using ControlFlowWalker might be a simple way to get such a stack.

src/passes/Unsubtyping.cpp
```diff
@@ -0,0 +1,482 @@
...
+            it != parent.supertypes.end() && !it->second.isBasic()) {
+          return it->second;
+        }
+        return std::nullopt;
+      }
+    };
+    Rewriter(*this, wasm).update();
+  }
+
+  void visitFunction(Function* func) {
```
kripken, line 218: Perhaps all the visitors could be in a separate class? That would be easier to read and also make the code reusable in the future, I think. That is, this separate class would just issue lots of calls to `parent.noteSubtype` (or it could store the results in itself, and then we could run those in parallel and then merge them, to use multiple cores)

src/passes/pass.cpp
```diff
@@ -495,6 +495,9 @@ void PassRegistry::registerPasses() {
   registerPass("type-unfinalizing",
                "mark all types as non-final (open)",
                createTypeUnFinalizingPass);
+  registerPass("unsubtyping",
+               "removes unnecessary subtyping relationships",
+               createUnsubtypingPass);
```
kripken, line 500: Should we run this in the default pipeline?

src/passes/Unsubtyping.cpp
```diff
@@ -0,0 +1,482 @@
...
+#include "pass.h"
+#include "wasm-traversal.h"
+#include "wasm-type.h"
+#include "wasm.h"
+
+namespace wasm {
+
+namespace {
+
+struct Unsubtyping
```
kripken, line 32: Some overview of the algorithm might help here. I spent a little time but it wasn't obvious to me why casts differ from other operations, for example, and hence why we need that double while loop on the work queue.

My lack of understanding may be because I am stuck in the general direction I thought could work, but perhaps can't (and so your solution might work but is more complex). I was thinking after you have gathered all the data from `noteSubtype` you are basically done, except from calling `propagateToSubTypes` (`struct-utils.h`). That is, whenever we see that a `B` was written to a place that is defined as `A` then `B` and all subtypes of `B` must be writable there, so they must be subtypes of `A`. What am I missing?

src/passes/Unsubtyping.cpp
tlively: This makes sense as a way to parallelize the pass, but I'm surprised that you think it would be easier to read if these were in a separate class. Why is that?

src/passes/pass.cpp
tlively: Maybe, but I'm worried that the cast analysis is going to make this extremely expensive. It would be nice to get some data on that first.

src/passes/Unsubtyping.cpp
kripken: Just as a separation of concerns. That other class would gather the raw data of what type is assigned to what, and nothing more. The user of the class would get that data and not care how it was gathered (all the visitors).

src/passes/Unsubtyping.cpp
tlively: If we have `C <: B <: A`, and we see that `B` is written to a location typed `A`, but `C` is never written to a `B` or `A` location, then your proposed solution would not be able to optimize anything, but the current implementation would be able to optimize `C` to not have any supertypes.

Casts do not introduce subtyping requirements on their own. If we have a cast from `A` to `C` (and no other casts) in the setup above, then it would be unnecessarily conservative to require `C <: A`, since we know that cast cannot succeed because we observed that `C` does not inhabit `A`.

I'm doing my best to get the most precise possible analysis (without making it a data flow analysis or anything) so we never have to come back and improve this ever again :)

kripken: I see, thanks. Eager propagation doesn't do that well, good point.

About your second paragraph, I wonder if we can actually have a separate pass that finds situations like a cast from `A` to `C` that will never succeed because no `C` is written to an `A`. That pass could replace such casts with failures.

Assuming such a pass, would that not handle casts entirely? And then that would leave a simple operation without propagation: for each type T, we see where it is written, and make it a subtype of the closest ancestor to itself among those, and as a second stage add more subtyping to keep validation working.

tlively: I think that could mostly work if we did a data flow analysis to find a set of types that can appear at each cast source, but even then, the data flow analysis would depend on subtyping information to be safe. For example, if a function parameter is a cast source, we would have to conservatively assume any of its subtypes could be the cast target unless we wanted to do an interprocedural analysis. No matter how precise we make the analysis, it still has to conservatively bottom out at depending on subtyping information.

So we have a circular dependency: we need subtyping information to analyze cast sources, and we need to analyze cast sources to get subtyping information. The only way to resolve that dependency optimally is to do both analyses in lockstep until reaching a fixed point, which is what the current implementation does.

kripken: Thanks for explaining, I think I see the general shape of your algorithm now.

I still don't get what constraints casts place on us, however (aside from us needing to remove failing casts):

> For example, if a function parameter is a cast source, we would have to conservatively assume any of its subtypes could be the cast target unless we wanted to do an interprocedural analysis.

I don't see why we need that yet. Consider this module:

```wat
(module
  (type $A)
  (type $B (sub $A))

  (func $get-A (param $A)
    (ref.cast $B (local.get $A))
  )
)
```

Now let's add one of the following functions to it, starting with this:
```wat
(func $caller
)
```
We added nothing here. We can turn that cast into a trap, since we've never seen
any $B written to an $A location, and we can remove the subtyping between $B and
$A.

Next, imagine we added this function instead:
```wat
(func $caller
  (call $get-A (new $A))
)
```
The cast will fail at runtime. Again, we can see it will trap at compile time,
and remove the subtyping.

Finally, imagine we added this function instead:
```wat
(func $caller
  (call $get-A (new $B))
)
```
The cast will succeed at runtime. Our analysis will see that we write a $B to an
$A location, so $B must be a subtype of $A, so we can optimize nothing here.

kripken: Hmm, I don't understand the testcase. I see a `local.set` of a `$bot.A` to `$top.A`, so we need that subtyping, and aside from that I see only casts, which as I said before, I don't think should actually constrain us (but we do need to fix up casts later). This is what I'd expect the optimization to emit for that module:
```wat
(module
 (rec
  (type $topC (sub (struct)))
  (type $midC (sub (struct)))
  (type $botC (sub (struct)))

  (type $topB (sub (struct (ref null $topC))))
  (type $midB (sub (struct (ref null $botC))))
  (type $botB (sub (struct (ref null $botC))))

  (type $topA (sub (struct (ref null $topB))))
  (type $midA (sub (struct (ref null $botB))))
  (type $botA (sub $topA (struct (ref null $botB)))) ;; only subtyping kept
 )

 (func $cast
  (local $l (ref null $topA))
  (local.set $l
   (struct.new $botA)
  )
  (drop
   (ref.cast (ref $botA) ;; rewritten to the only possibility, $botA
    (struct.new $topA)
   )
  )
  ;; Other two are outright impossible, so rewritten to trap.
  (unreachable)
  (unreachable)
 )
)
```
Is that output wrong somehow?

tlively: That transformation happens to be correct for that test because the casts will fail, but we can construct cases where the semantics would change if we disregarded subtyping constraints arising from casts:
```
(module
 (type $top (sub (struct)))
 (type $mid (sub $top (struct)))
 (type $bot (sub $mid (struct)))
 
 (func $test
  (local $top (ref $top))
  (local $mid (ref $mid))
  (local $bot (ref $bot))
  (local.set $top (struct.new $bot)) ;; Requires $bot <: $top
  (local.set $mid (ref.cast (ref $mid) (local.get $top))) ;; Cast $top -> $mid
 )
)
```
The cast here should succeed because the value of type `(ref $bot)` inhabits `(ref $mid)`, but if you disregard the subtyping constraints that casts introduce, you end up with `$bot <: $top` as the only constraint, so you would rewrite the types like this:

```
(type $top (sub (struct)))
(type $mid (sub (struct))) ;; no longer related to $top or $bot
(type $bot (sub $top (struct)))
```

With that type rewriting, the module would still validate, but the cast would fail because the value of type `(ref $bot)` no longer inhabits `(ref $mid)`.

Here's the key correctness property for casts, as documented in `analyzeCasts`:
```
    // For each cast (src, dest) pair, any type that remains a subtype of src
    // (meaning its values can inhabit locations typed src) and that was
    // originally a subtype of dest (meaning its values would have passed the
    // cast) should remain a subtype of dest so that its values continue to pass
    // the cast.
```

kripken: IIANM the last comment's output is maximally optimized, but over lunch I realized something that may be related to what you were trying to get me to understand before: while on that testcase what I'm saying can work , in other cases the "fix up casts" stuff that I've been handwaving can actually be difficult.

Specifically, say we have a cast to `$X` and there are two *sibling* types `$Y1, $Y2`. Then the cast could succeed if the input is either of those. I can think of three general approaches to that:

1. What I am guessing this PR does, which is to "reattach" `$Y1, $Y2` to `$X` as children (is that right?). Then you don't need to change the cast at all, but you do pay the cost of all of `$X`'s fields and their memory overhead.
2. Expand into casts: `(if (ref.test $Y1) (ref.cast $Y1) (ref.cast $Y2))`. This is larger in code size and may be slower, but it allows us to have any parent type for `$Y1, $Y2` so it could save memory. It is also "clean" in that we actually do remove the maximal amount of subtyping based on what is written where, and we can do that in a simple one-pass method. And then after that in another single pass we fix up all casts.
3. Slightly vague idea, but we can do a variation on the 1 and create `$shadow-X`, a "shadow hierarchy" of types that contain no fields and are only used for casts; `$Y1, $Y2` would be subtypes of the otherwise-empty `$shadow-X`, which has no memory overhead for them, and then we just need to make the cast be `(ref.cast $shadow-X)` rather than on `$X`. However, while this works for two siblings whose parents we want to remove, it's less obvious to me what to do if one is a subtype of `$X` anyhow and the other is not.

kripken: (sorry, our comments raced; reading your comment now after posting mine)

kripken: About the example from your last comment:
```wat
(module
 (type $top (sub (struct)))
 (type $mid (sub $top (struct)))
 (type $bot (sub $mid (struct)))
 
 (func $test
  (local $top (ref $top))
  (local $mid (ref $mid))
  (local $bot (ref $bot))
  (local.set $top (struct.new $bot)) ;; Requires $bot <: $top
  (local.set $mid (ref.cast (ref $mid) (local.get $top))) ;; Cast $top -> $mid
 )
)
```
In my idea (2 in my list) the output would be this:
```wat
(module
 (type $top (sub (struct)))
 (type $mid (sub (struct)))      ;; no super
 (type $bot (sub $top (struct))) ;; super improved
 
 (func $test
  (local $top (ref $top))
  (local $mid (ref $mid))
  (local $bot (ref $bot))
  (local.set $top (struct.new $bot))
  (local.set $mid (ref.cast (ref $bot) (local.get $top))) ;; cast type changed
 )
)
```
That almost works aside from the last `local.set`, so, yeah, as you said there is a problem here.

Separating the two problems, however, as I've suggested, we could first improve the cast, then optimize, and then remove supertypes. The optimization in the middle would get rid of `$mid`. That is, I think reusing the "what is written where" info from this pass to refine casts could help make this pass more effective. I was hoping it could also remove the cast handling complexity but that part might not be practical.

kripken: I realize that `--abstract-type-refining` does what I was thinking about regarding casts, so we do already have that first part already :smile: 

Anyhow, I'm mostly convinced I should review this PR for landing, but let me know if anything in the last ideas suggests anything worthwhile.

tlively: Yeah, optimizing casts and before and after this pass can be helpful, but we can't do it in the middle of this pass with incomplete type information.

kripken: We do have enough to do idea 2 from earlier ("expand casts"), don't we? I agree there are tradeoffs there but we have all the information we need at least.

kripken: Thinking more on this after we discussed it offline, I think I realized what was making it hard for me to understand before. I was expecting there to be a symmetry between all assignments, but there is actually a very unsymmetrical issue around casts. Say we have this:
```wat
;; $A :> B :> C
(local.set $A (new $B))
(local.set $B (new $C))
```
Then we have two constraints: B must be a subtype of A, and C of B. All we need is for data to be passable according to those rules, and in fact because of how subtyping works - it is recursive -  they imply `A :> B :> C`, so in particular C must be a subtype of A.

However, here:

```wat
;; $A :> B :> C
(local.set $A (new $C))
(cast $B (local.get $A))
```

Now we have two constraints, that C must be a subtype of A, and B must be a subtype of A. In theory we could fulfill those with both B and C being subtypes of A as siblings, but we need to be able to pass data around, and passing C to A and then that C-in-A to B will fail, because we no longer have that recursive property from before. That is, to be able to pass data around according to our two constraints, we need more than just those simple subtyping relations, but also some form of closure operation. If all we have are upcasts then we get it for free as we saw before; if we have downcasts then the symmetry is broken and we need more. Specifically we need everything that can reach A to be able to reach B, which implies subtyping between all subtypes of A to B.

A little verbose but maybe something like this in a comment in the pass could be helpful?

src/passes/Unsubtyping.cpp
tlively: As we discussed, the raw data is not that useful on its own because it is incomplete. Since we have to walk both function code and other module code, I found it easier to parallelize this pass without separating out the visitors.

src/passes/pass.cpp
tlively: The pass runs in 0.8 seconds on optimized calc.wasm after running TypeSSA or 0.6 seconds without TypeSSA, so I think it is fast enough that we could put it in the default pipeline.

tlively:
After #6002, the fuzzer is very happy with this. I have about 90k iterations so far and it's still going. Meanwhile I will continue trying to figure out this compilation error that happens on CI but not locally.

src/passes/pass.cpp
kripken: Great!

Let's leave the default pipeline for a followup, unless we are certain this has no downsides? Intuitively I think this can only improve optimizability, but there are theoretical cases like it preventing a merging of similar functions (if the merge depending on subtyping; we don't do that atm, but it is one idea we've had).

test/lit/passes/unsubtyping-casts.wast
```diff
@@ -0,0 +1,452 @@
...
+ ;; CHECK-NEXT:  (local.set $l
+ ;; CHECK-NEXT:   (struct.new_default $bot)
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT:  (drop
+ ;; CHECK-NEXT:   (ref.cast (ref none)
+ ;; CHECK-NEXT:    (struct.new_default $top)
+ ;; CHECK-NEXT:   )
+ ;; CHECK-NEXT:  )
+ ;; CHECK-NEXT: )
+ (func $cast-optmizable
```
kripken, line 295:
```suggestion
 (func $cast-optimizable
```

test/lit/passes/unsubtyping.wast
```diff
@@ -0,0 +1,1599 @@
...
+ )
+
+ ;; CHECK:      (func $callee (type $f) (result (ref $sub))
+ ;; CHECK-NEXT:  (unreachable)
+ ;; CHECK-NEXT: )
+ (func $callee (result (ref $sub))
+  (unreachable)
+ )
+)
+
```
kripken, line 970: IIANM, we are missing a test for the `call_ref` target's heap type itself (not its params and results)? I mean:
```wat
(call_ref $super
  (ref.func $sub_func)
)

(func $sub_func (type $sub))

(type $sub (sub $super (func)))
```

kripken: Regardless of that being a call_ref, we are writing a thing of type `$sub` into a lot of type `$super` (on function types).

kripken (commented):
Still reading the code but here are comments on the tests. Nice work!

src/passes/Unsubtyping.cpp
```diff
@@ -0,0 +1,580 @@
...
+      // and all its supertypes if we have already analyzed casts.
+      return;
+    }
+    // We already had a recorded supertype. The new supertype might be deeper,
+    // shallower, or identical to the old supertype.
+    auto oldSuper = it->second;
+    if (super == oldSuper) {
+      return;
+    }
+    if (HeapType::isSubType(super, oldSuper)) {
```
kripken, line 151: Might be worth a comment here, I had to work it out in my head, that since each type has a single super, if we know a type is a subtype of two others then all three must form a chain and the two supers must be related as well.

src/passes/Unsubtyping.cpp
```diff
@@ -0,0 +1,580 @@
...
+  // The new set of supertype relations.
+  std::unordered_map<HeapType, HeapType> supertypes;
+
+  // Map from cast source types to their destinations.
+  std::unordered_map<HeapType, std::unordered_set<HeapType>> castTypes;
+
+  // The set of subtypes that need to have their type definitions analyzed to
+  // transitively find other subtype relations they depend on. We add to it
+  // every time we find a new subtype relationship we need to keep.
+  std::unordered_set<HeapType> worklist;
```
kripken, line 117: Might be worth seeing if using `UniqueDeferredQueue` instead of `unordered_set` is faster here. It usually is in my experience, and as a bonus, there is a simple `push/pop` so using the API is more convenient later down.

src/passes/Unsubtyping.cpp
```diff
@@ -0,0 +1,580 @@
...
+        auto super = supertypes.at(type);
+        if (super.isBasic()) {
+          continue;
+        }
+
+        if (type.isStruct()) {
+          const auto& fields = type.getStruct().fields;
+          const auto& superFields = super.getStruct().fields;
+          for (size_t i = 0, size = superFields.size(); i < size; ++i) {
+            if (fields[i].mutable_ == Immutable) {
```
kripken, line 223: This check (and below for the others) is an optimization, IIUC? (since subtyping is only possible then). I would remove it, as the called function early exits if they are equal anyhow, and then we would not need to update this code in the future if wasm *does* start to allow subtyping of mutable fields in some manner. Even if not, it would keep this code simpler to remove.

src/passes/Unsubtyping.cpp
```diff
@@ -0,0 +1,580 @@
...
+      for (auto superIt = it; superIt != supertypes.end();
+           superIt = supertypes.find(superIt->second)) {
+        auto super = superIt->second;
+        auto destsIt = castTypes.find(super);
+        if (destsIt == castTypes.end()) {
+          continue;
+        }
+        for (auto dest : destsIt->second) {
+          if (HeapType::isSubType(type, dest)) {
+            noteSubtype(type, dest);
```
kripken, line 268: `dest` is used in the code, but the comment has `dest, src`, so it would be good to use `src` in the code as well to match the comment.

src/passes/Unsubtyping.cpp
```diff
@@ -0,0 +1,580 @@
...
+        }
+        return std::nullopt;
+      }
+    };
+    Rewriter(*this, wasm).update();
+  }
+
+  void doWalkModule(Module* wasm) {
+    // Visit the functions in parallel.
+    ModuleUtils::ParallelFunctionAnalysis<Unsubtyping> analysis(
```
kripken, line 293: It's a little awkward to create an entire Unsubtyping for each function, including the `worklist`. Though maybe that's the only field that isn't needed in both parallel and later work, so maybe it's ok?

src/passes/Unsubtyping.cpp
```diff
@@ -0,0 +1,580 @@
...
+          return it->second;
+        }
+        return std::nullopt;
+      }
+    };
+    Rewriter(*this, wasm).update();
+  }
+
+  void doWalkModule(Module* wasm) {
+    // Visit the functions in parallel.
```
kripken, line 292:
```suggestion
    // Visit the functions in parallel, filling in `supertypes` and `castTypes` on separate instances which will later be merged.
```

src/passes/Unsubtyping.cpp
kripken: As an example of a use I'd like to make of this information, the validator could look at the accumulated results of `noteSubtype` and issue an error for content that doesn't fit in those constraints. But I guess we can leave that to later refactoring.

src/passes/Unsubtyping.cpp
```diff
@@ -0,0 +1,580 @@
...
+  void visitBreak(Break* curr) {
+    if (curr->value) {
+      noteSubtype(curr->value->type, findBreakTarget(curr->name)->type);
+    }
+  }
+  void visitSwitch(Switch* curr) {
+    if (curr->value) {
+      noteSubtype(curr->value->type, findBreakTarget(curr->default_)->type);
+      for (auto name : curr->targets) {
+        noteSubtype(curr->value->type, findBreakTarget(name)->type);
```
kripken, line 361: Better to use `BranchUtils::getUniqueTargets`, as (1) it's shorter, and (2) it deduplicates the names first, leading to less work (often many switch targets are identical).

src/passes/Unsubtyping.cpp
```diff
@@ -0,0 +1,580 @@
...
+    auto* table = getModule()->getTable(curr->table);
+    auto tableType = table->type.getHeapType();
+    if (HeapType::isSubType(tableType, curr->heapType)) {
+      // Unlike other casts, where cast targets are always subtypes of cast
+      // sources, call_indirect target types may be supertypes of their source
+      // table types. In this case, the cast will always succeed, but only if we
+      // keep the types related.
+      noteSubtype(tableType, curr->heapType);
+    } else if (HeapType::isSubType(curr->heapType, tableType)) {
+      noteCast(table->type.getHeapType(), curr->heapType);
```
kripken, line 388:
```suggestion
      noteCast(tableType, curr->heapType);
```

test/lit/passes/unsubtyping.wast
kripken: Oh, wait, I forgot that in Binaryen IR we don't keep a separate (redundant) annotation. So the annotation is always equal to the type of the given target anyhow.

src/passes/Unsubtyping.cpp
```diff
@@ -0,0 +1,580 @@
...
+    handleCall(curr, curr->target->type.getHeapType().getSignature());
+  }
+  void visitRefTest(RefTest* curr) {
+    noteCast(curr->ref->type, curr->castType);
+  }
+  void visitRefCast(RefCast* curr) { noteCast(curr->ref->type, curr->type); }
+  void visitBrOn(BrOn* curr) {
+    if (curr->op == BrOnCast || curr->op == BrOnCastFail) {
+      noteCast(curr->ref->type, curr->castType);
+      noteSubtype(curr->getSentType(), findBreakTarget(curr->name)->type);
```
kripken, line 481: `BrOnNonNull` needs this line as well, I think?
```wat
(block (result $super)
  (br_on_non_null (local.get $sub)) ;; writes a sub to a super slot
)
```
(it's true the br by itself just affects nullability, but the heap type can still be a subtype of the target block)

src/passes/Unsubtyping.cpp
tlively: Yeah, I think leaving it to a later refactoring would be good.

src/ir/table-utils.cpp
```diff
@@ -72,14 +72,11 @@ bool usesExpressions(ElementSegment* curr, Module* module) {
...
-  // of using expressions.
-  bool hasTableOfSpecializedType =
-    curr->table.is() &&
-    module->getTable(curr->table)->type != Type(HeapType::func, Nullable);
+  // If the segment has a specialized (non-MVP) type, then it must use the
+  // post-MVP form of using expressions.
+  bool hasSpecializedType = curr->type != Type(HeapType::func, Nullable);
 
-  return !allElementsRefFunc || hasTableOfSpecializedType;
+  return !allElementsRefFunc || hasSpecializedType;
```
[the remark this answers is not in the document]
tlively: No, it's a GitHub artifact now that the previous PR is merged. It will be gone next time I push.

src/passes/pass.cpp
tlively: This pass still runs in 12 seconds and cannot optimize anything on the unoptimized binary. After the first -O3, it takes 7.5 seconds (and I don't know if it optimizes anything there).

kripken: 12 is maybe a lot. An option might be to add a limit on the number of iterations and do a conservative estimate from there, maybe.

kripken: Another option might be to only run this in `-O3` and such, and recommend the J2Wasm first pass be `-O1` or `-O2` only, which might make sense actually anyhow.

src/passes/Unsubtyping.cpp
tlively: Sure, but it's `super` that will be renamed to `dest`, not `type`. If that's surprising, then that further motivates aligning the names.

src/passes/Unsubtyping.cpp
tlively: Yes, and importantly it lets us reuse all the `noteSubtype` and `noteCast` helper functions rather than having to duplicate them.

tlively: ohhh, that's a good idea. I was struggling to come up with a good solution there. Still, I'm not convinced that anything less than the whole set of subtypes we calculate here can be useful.

src/passes/Unsubtyping.cpp
tlively: Indeed, this saves a second and a half on the unoptimized binary 👍

src/passes/pass.cpp
tlively: Happy to experiment (or have j2wasm experiment) with different pass placements after we get this landed. It would be interesting to see what optimizations unblock this one from working.

src/passes/Unsubtyping.cpp
tlively: It will be in the next push.

tlively:
All comments should be addressed!

kripken:
In the future, it would be great to not force-push to big PRs like this. The main source file and the smaller test file have changes but I don't have a good way to see a diff compared to what I read last, unless I'm missing something.

kripken (approved):
Great!

kripken:
Thanks, yeah, having just the last few PRs to look at makes this a lot faster.

PR #5985, opened 2023-10-01
Refine ref.test's castType during refinalization

tlively:
Just like we do with other casts, refine the cast type to be the greatest lower
bound of its previous cast type and its input type. The difference is that the
output type of ref.test remains i32, but it's still useful to retain more
precise type information.


tlively:
* main
  * **PR #5985** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/5985" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/5985?utm_source=stack-comment).

kripken (approved):

PR #5987, opened 2023-10-02
Asyncify: Improve comments

aheejin:
This fixes some outdated comments and typos in Asyncify and improves some other comments. This tries to make code comments more readable by making them more accurate and also by using the three state (normal, unwinding, and rewinding) consistently.

Drive-by fix: Typo fixes in SimplifyGlobals and wasm-reduce option.

### DIFF
--- a/src/passes/Asyncify.cpp
+++ b/src/passes/Asyncify.cpp
@@ -504,7 +504,7 @@ class ModuleAnalyzer {
     : public ModuleUtils::CallGraphPropertyAnalysis<Info>::FunctionInfo {
     // The function name.
     Name name;
-    // If this function can start an unwind/rewind and not the bottommost
+    // If this function can start an unwind/rewind and is not the bottommost
     // runtime.
     bool canChangeState = false;
     // If this function is part of the runtime that receives an unwinding
@@ -977,11 +977,13 @@ private:
     //      new_side2();
     //    }
     // where new_sideN is
-    // if (normal || check_call_index(callIndex))
+    // if (normal || check_call_index(callIndex)) {
     //   sideN();
+    // }
     // when sideN can change the state, and
-    // if (normal)
+    // if (normal) {
     //   sideN();
+    // }
     // when it does not. (See makeCallSupport() for details.)
     //
     // This way we will linearly get through all the code in the function,
--- a/src/passes/Asyncify.cpp
+++ b/src/passes/Asyncify.cpp
@@ -504,8 +504,10 @@ class ModuleAnalyzer {
     : public ModuleUtils::CallGraphPropertyAnalysis<Info>::FunctionInfo {
     // The function name.
     Name name;
-    // If this function can start an unwind/rewind and is not the bottommost
-    // runtime.
+    // If this function can start an unwind/rewind. We only set this in cases
+    // where we need to know that fact and also that we need to instrument code
+    // to handle it (as a result, we do not set it for the bottommost runtime,
+    // which needs no instrumentation).
     bool canChangeState = false;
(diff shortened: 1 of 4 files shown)


src/passes/Asyncify.cpp
```diff
@@ -503,7 +504,8 @@ class ModuleAnalyzer {
     : public ModuleUtils::CallGraphPropertyAnalysis<Info>::FunctionInfo {
     // The function name.
     Name name;
-    // If this function can start an unwind/rewind.
+    // If this function can start an unwind/rewind and not the bottommost
```
kripken, line 507:
```suggestion
    // If this function can start an unwind/rewind and is not the bottommost
```

kripken: Is this right? Or did I misunderstand your intention?

But skimming the code below, I'm not sure how this relates to the bottommost runtime?

src/passes/Asyncify.cpp
```diff
@@ -962,21 +964,25 @@ struct AsyncifyFlow : public Pass {
...
-    //      side1();
-    //    }
-    //    if (rewinding || !temp) {
-    //      side2();
+    //    if (rewinding || !cond) {
+    //      new_side2();
     //    }
+    // where new_sideN is
+    // if (normal || check_call_index(callIndex))
+    //   sideN();
```
kripken, line 981:
```suggestion
    // if (normal || check_call_index(callIndex)) {
    //   sideN();
    // }
```

src/passes/Asyncify.cpp
```diff
@@ -962,21 +964,25 @@ struct AsyncifyFlow : public Pass {
...
+    // when sideN can change the state, and
+    // if (normal)
+    //   sideN();
```
kripken, line 984:
```suggestion
    // if (normal) {
    //   sideN();
    // }
```

src/passes/Asyncify.cpp
aheejin: So the current definition is `canChangeState` is true whenever a function can start unwinding or rewinding. But later, we set this to false when `isBottomMostRuntime` is true:
https://github.com/WebAssembly/binaryen/blob/0fe08192d866f72d0f4e680c862bb7af48dec1ac/src/passes/Asyncify.cpp#L629-L630

This was confusing to me so I added this. Let me know if I'm misunderstanding something.

kripken: Oh, then maybe the better thing is to clarify with a comment there.

What's going on there is that we do not mark the bottom-most runtime as changing the state,

https://github.com/WebAssembly/binaryen/blob/0fe08192d866f72d0f4e680c862bb7af48dec1ac/src/passes/Asyncify.cpp#L515-L517

The top-most runtime can start a pause/resume, but the bottom is just where we unwind to and resume from. (That technically does modify the state, but we care about places we need to instrument for state changes, and the bottom runtime needs no such instrumentation - we don't need to "handle" those state changes like in user code.)

kripken: I think it makes more sense to remove the change here and add a comment below on that operation.

But I don't feel strongly. I'd say though that if we want the comment right here then I'd replace it with something like this:

```diff
- If this function can start an unwind/rewind and not the bottommost runtime
+ If this function can start an unwind/rewind. We only set this in cases where we need to know that fact and also that we need to instrument code to handle it (as a result, we do not set it for the bottommost runtime, which needs no instrumentation).
```

aheejin: Thanks. Fixed as suggested.

kripken (approved):

PR #5988, opened 2023-10-02
Asyncify: Simpify if into i32.or

aheejin:
```wast
(if (result i32)
  (expr0)
  (i32.const 1)
  (expr1)
)
```
can be written as
```wast
(i32.or
  (expr0)
  (expr1)
)
```

Also this removes some unused variables and methods.

This also adds an optimization for
```wast
(i32.eqz                                                                   
  (global.get $__asyncify_state)                                            
)
```
in `--mod-asyncify-always-and-only-unwind` to fix an unexpected regression caused by this.

### DIFF
--- a/src/passes/Asyncify.cpp
+++ b/src/passes/Asyncify.cpp
@@ -530,10 +530,9 @@ public:
                  const String::Split& removeListInput,
                  const String::Split& addListInput,
                  const String::Split& onlyListInput,
-                 bool asserts,
                  bool verbose)
     : module(module), canIndirectChangeState(canIndirectChangeState),
-      fakeGlobals(module), asserts(asserts), verbose(verbose) {
+      fakeGlobals(module), verbose(verbose) {
 
     PatternMatcher removeList("remove", module, removeListInput);
     PatternMatcher addList("add", module, addListInput);
@@ -784,7 +783,6 @@ public:
   }
 
   FakeGlobalHelper fakeGlobals;
-  bool asserts;
   bool verbose;
 };
 
@@ -840,11 +838,6 @@ public:
     return makeBinary(EqInt32,
                       makeGlobalGet(ASYNCIFY_STATE, Type::i32),
                       makeConst(Literal(int32_t(value))));
-  }
-
-  Expression* makeNegatedStateCheck(State value) {
-    return makeUnary(Abstract::getUnary(pointerType, Abstract::EqZ),
-                     makeStateCheck(value));
   }
 };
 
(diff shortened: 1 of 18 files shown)


test/lit/passes/asyncify_mod-asyncify-always-and-only-unwind_O.wast
```diff
@@ -41,7 +41,16 @@
...
+  ;; CHECK-NEXT:  (local.set $0
+  ;; CHECK-NEXT:   (block $__asyncify_unwind (result i32)
+  ;; CHECK-NEXT:    (block
+  ;; CHECK-NEXT:     (call $import)
+  ;; CHECK-NEXT:     (br $__asyncify_unwind
+  ;; CHECK-NEXT:      (i32.const 0)
+  ;; CHECK-NEXT:     )
+  ;; CHECK-NEXT:    )
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
```
kripken, line 53: Why did this change?

aheejin: In old code, after some optimization passes, this happened:
```wast
(select
  ...
  ...
  (global.get $__asyncify_state)
)
```
Because this code is checking whether the current state is normal (0) or rewinding (2), this can be optimized by https://github.com/WebAssembly/binaryen/blob/0fe08192d866f72d0f4e680c862bb7af48dec1ac/src/passes/Asyncify.cpp#L1925-L1937

But now generated code is different and after optimization passes this ended up like this:
```wast
(i32.eqz                                                                   
  (global.get $__asyncify_state)                                            
)
``` 
And we didn't have the same optimization for `Unary`. So added a similar optimization for `i32.eqz`, so that the whole `i32.eqz` can be replaced with `(i32.const 0)`.

src/passes/Asyncify.cpp
```diff
@@ -1936,6 +1927,24 @@ struct ModAsyncify
...
+    }
+    // This is a comparison of the state to zero, which means we are checking
+    // "if running normally, run this code, but if rewinding, ignore it". If
+    // we know we'll never rewind, we can optimize this.
+    if (neverRewind) {
+      Builder builder(*this->getModule());
+      // The whole expression will be 1 because it is (i32.eqz (i32.const 0))
+      this->replaceCurrent(builder.makeConst(int32_t(1)));
+    }
+  }
```
aheejin, line 1946: This is repeating the same existing optimization in `visitSelect`. This was added in this PR because this PR unintentionally caused some regression in a test with `--mod-asyncify-always-and-only-unwind`. See https://github.com/WebAssembly/binaryen/pull/5988#discussion_r1343616394

kripken (approved):

PR #5992, opened 2023-10-03
Compute full transitive closure in GlobalEffects

kripken:


kripken:
I am not seeing a benefit on Java with this, but this pass is now complete and it could help in principle, as the tests show, so I think it's worth landing and experimenting with more.

src/passes/GlobalEffects.cpp
```diff
@@ -22,60 +22,184 @@
...
+    // including which functions they call directly (so that we can compute
+    // transitive effects later).
 
-    auto& funcEffectsMap = getPassOptions().funcEffectsMap;
-
-    // First, clear any previous effects.
-    funcEffectsMap.reset();
+    struct FuncInfo {
+      // Effects in this function.
+      std::unique_ptr<EffectAnalyzer> effects;
```
tlively, line 38: Would it make sense to use `std::optional` here?

src/passes/GlobalEffects.cpp
```diff
@@ -22,60 +22,184 @@
...
+      if (!info.effects) {
+        // Add no entry to funcEffectsMap, since nothing is known.
+        continue;
+      }
+
+      // Only allocate a new funcEffectsMap here when we actually have data for
+      // it (which might make later effect computation slightly faster, to
+      // quickly skip the funcEffectsMap code path).
+      if (!funcEffectsMap) {
+        funcEffectsMap = std::make_shared<FuncEffectsMap>();
```
tlively, line 200: Why is this a shared pointer? I would have assumed we would use unique ownership and anything that needs to reference it would do so via a normal pointer or reference.

tlively (approved):
An alternative approach would be to apply effects from callees to callers until reaching a fixed point. This could evaluate fewer callee-caller pairs than eagerly computing the transitive closure of the call graph because new pairs would go on the work queue only if the caller's effects were actually changed, but on the other hand, evaluating each pair would be more expensive, so it's unclear whether it would always be a win.

kripken:
> An alternative approach would be to apply effects from callees to callers until reaching a fixed point. [..] so it's unclear whether it would always be a win.

Good thinking, yeah, see earlier commits in this branch. It turns out it was slower to do the closure with effects, I think because we track so many different effects in each EffectAnalyzer, and a single update (one modified global) can trigger a whole set of updates. It was over 2x faster to do the current way, which is to find the closure over the call graph first and do effects later (measured on Java). Doing it that way also avoids needing extra work to figure out infinite recursion (which closing over effects doesn't do; but if we close over the call graph, we get it for free).

src/passes/GlobalEffects.cpp
kripken: Good idea, done.

I actually didn't know how to do late construction with optional, but it turns out you can do `optional.emplace(..)` for that...

src/passes/GlobalEffects.cpp
kripken: I think the reason (predating this PR of course) is that we store this on a `PassOption`, and we do allow copying those around.

tlively:
Neat, I'm glad you tried it out and measured.

> Doing it that way also avoids needing extra work to figure out infinite recursion (which closing over effects doesn't do; but if we close over the call graph, we get it for free).

The effects can't keep increasing forever, so shouldn't it have Just Worked with recursion if you were only re-enqueuing 
work when the effects changed?

src/passes/GlobalEffects.cpp
tlively: Makes sense, thanks 👍

kripken:
I mean, how can you detect infinite recursion? Even if you compute the closure of Effects, all you get is "all the globals written by this function are X,Y,Z, and it can throw, etc." but we don't have a list there of the functions it can directly call (that isn't an effect atm).

Whereas if you compute the closure of the call graph, then you just check "can x reach x" to know if it can infinitely recurse.

tlively:
You would still use mappings from callers to callees as your work list entries, so you would still be flowing information around the call graph until finding a fixed point, but there would be no need to eagerly compute the transitive closure of the call graph.

kripken:
But we also need to know which functions can infinitely recurse, so we do need to compute the transitive closure of the call graph somehow. Closing on effects by itself doesn't do that, even if the work list entries are based on the call graph of course.

tlively:
Ohhh, I see what you mean. Thanks!

PR #5997, opened 2023-10-05
Typed Continuations: Allow result types on tags

frank-emrich:

This PR relaxes the restriction that tags must not have `result`s , only `param`s. Tags with results must not be used for exception handling and are only allowed if the typed continuations feature is enabled.

As a minor point, this PR also changes the printing of tags without `param`s: To make the presentation consistent, `(param)` is omitted when printing a tag.


frank-emrich:
CC @tlively

kripken (approved):

PR #5998, opened 2023-10-05
Typed Continuations: Add cont type

frank-emrich:

This PR adds continuation types, of the form `(cont $foo)` for some function type `$foo`.

The only notable changes affecting existing code are the following:
- This is the first `HeapType` which has another `HeapType` (rather than, say, a `Type`) as its immediate child. This required fixes to certain traversals that have a flag for being at the toplevel of a type.
- Some shared logic for parsing `HeapType`s has been factored out.


frank-emrich:
CC @tlively

tlively (commented):
This looks great, thanks. Could you move the non-spec tests to test/lit? We've been trying to move off of the legacy test infrastructure as much as possible.

Also, my guess is that many of our type optimizations are not prepared to deal with a new heap type. We don't need to fix that immediately, but we need to make sure our fuzzer does not start crashing. That's a risk because it will pick up our tests as initial fuzzing contents.

If you haven't already, it would be best to make sure that `./scripts/fuzz_opt.py` gets to about 10k iterations without crashing before landing this.

If there are bugs and they turn out to be significant work to fix, then we can have the fuzzer ignore the new tests for now.

src/wasm-type.h
```diff
@@ -470,6 +476,14 @@ struct Signature {
   std::string toString() const;
 };
 
+struct Continuation {
+  HeapType ht;
```
kripken, line 480: Naive question from someone unfamiliar with the Type Continuations spec: If a continuation is just a heap type, why do we need a Continuation type?

frank-emrich: From a spec perspective, continuation types are built from existing function types, but they are different from them: Given an already defined function type `$f`, we have that `(cont $f)` is the type of continuations whose signature is like that of `$foo`, but `(cont $f)` and `$f` are different types. In particular, in the types section, you would separately have the following:
```
(type $f (func (param ...) (result ...)))
(type $c (cont $f))
```

Speaking just from an implementation perspective, having a dedicated `Continuation` type is mostly a matter of consistency and convenience: For example, in `HeapTypeInfo`, we may have a `HeapType continuation_func` field instead of `Continuation continuation` in the `union`. But that seems a bit dangerous: We must not accidentally mix up the `HeapType` that denotes the continuation itself vs. the function `HeapType` it is built from.

Further, having `Continuation` as its own type allowed me to add overloads for, say, the existing printing and hashing functions, and follow the same pattern as for other types when dispatching to these overloads.

Does that clarify things?

frank-emrich:
Sorry for the delay!

>  Could you move the non-spec tests to test/lit? 

Sure! Do these lit tests also test roundtripping to and from binary? I was happy to see that this is automatically done for the non-lit tests.

> If you haven't already, it would be best to make sure that `./scripts/fuzz_opt.py` gets to about 10k iterations without crashing before landing this.

Good point! I had completely overlooked checking what the fuzzer does. Right now, the fuzzer crashes when using any file as root that contains `cont` types. There is just a bunch of logic in the fuzzer of the shape `if(t.isSignature()) {...} else if (t.is...) {...} else {WASM_UNREACHABLE("oops")}`. I'll have to understand the fuzzer a bit better first, but would prefer adding support as a separate PR, once the other instructions are there as well.

tlively:
> Sorry for the delay!
> 
> > Could you move the non-spec tests to test/lit?
> 
> Sure! Do these lit tests also test roundtripping to and from binary? I was happy to see that this is automatically done for the non-lit tests.

Not automatically, but it's easy to add. See https://github.com/WebAssembly/binaryen/blob/main/test/lit/types-function-references.wast for a very thorough example that tries to test everything the non-lit test runner tests. Often we don't go quite that far and just add an extra RUN line with an additional --roundtrip option instead.

> > If you haven't already, it would be best to make sure that `./scripts/fuzz_opt.py` gets to about 10k iterations without crashing before landing this.
> 
> Good point! I had completely overlooked checking what the fuzzer does. Right now, the fuzzer crashes when using any file as root that contains `cont` types. There is just a bunch of logic in the fuzzer of the shape `if(t.isSignature()) {...} else if (t.is...) {...} else {WASM_UNREACHABLE("oops")}`. I'll have to understand the fuzzer a bit better first, but would prefer adding support as a separate PR, once the other instructions are there as well.

Sounds good. In that case, you can add your new test files to the `INITIAL_CONTENTS_IGNORE` list in fuzz_opt.py.

tlively:
Yep, just applying the diff should be fine.

tlively (approved):
LGTM! Did you get a chance to fuzz this to make sure the fuzzer is happy even after telling it to ignore the new test?

src/wasm-type.h
kripken: Thanks! Now I understand.

frank-emrich:
Yes, I let it run to 19k iterations, no issues.

tlively:
Great, thank you!

PR #6002, opened 2023-10-08
Fix a bug printing and emitting empty, passive element segments

tlively:
Empty, passive element segments were always emitted as having `func` type
because all their elements trivially were RefFunc (because they have no
elements) and because we were incorrectly checking table types if they existed
instead of the element segment's type directly to see if it was non-func.

Fix the bug by checking each element segment's type directly and add a test.


tlively:
* main
  * **PR #6002** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6002" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/6002?utm_source=stack-comment).

kripken (approved):

PR #6004, opened 2023-10-10
AddThrows experiment

tlively:


tlively:
* main
  * **PR #6004** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6004" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/6004?utm_source=stack-comment).

vouillon:
Note that for signed divisions, dividing by `-1` can also trap.

src/passes/AddThrows.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+  template<typename T>
+  void handleIndex(T* curr, std::optional<Index> refLocal) {
+    Builder builder(*getModule());
+    if (!refLocal) {
+      refLocal = builder.addVar(getFunction(), curr->ref->type);
+      curr->ref = builder.makeLocalTee(*refLocal, curr->ref, curr->ref->type);
+    }
+    auto scratch = builder.addVar(getFunction(), Type::i32);
+    curr->index = builder.makeIf(
```
vouillon, line 59: I have been using `br_if` since inserting a conditional makes a lot of optimizations (like Local CSE) less effective.

tlively: Interesting! I think wasm-opt generally prefers to emit conditionals because they end up being fewer bytes than a block + br_if. I don't have an intuition for why a conditional would be worse for CSE. Can you explain that?

vouillon: As far as I understand, CSE currently does not work across basic blocks, except when they are adjacent. The expressions before and after a conditional belong to different non-adjacent basic blocks, while the basic blocks before and after a `br_if` are considered adjacent. See file `ir/linear-execution.h`.

I'm doing a non-local transformation, actually: I'm putting the target of the `br_if` at the end of the current function (or at the end of a `try` block).

PR #6005, opened 2023-10-11
Add getGeneralSuperType() that includes basic supers, and use in fuzzer

kripken:
With this, the fuzzer can replace e.g. an `eq` expression with a specific struct type,
because now it is away that struct types have `eq` as their ancestor.

### DIFF
--- a/src/ir/subtypes.h
+++ b/src/ir/subtypes.h
@@ -122,7 +122,6 @@ struct SubTypes {
     }
 
     // Add the max depths of basic types.
-    // TODO: update when we get structtype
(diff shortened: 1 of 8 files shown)


tlively (approved):
I don't love the name `getGeneralSuperType`, but I'm not sure it's worth doing a preliminary PR to rename the current function to `getDeclaredSuperType` so this one can be `getSuperType`.

kripken:
Nice idea, that does sound better than "general". I think it's worth the work, I'll look into it.

PR #6006, opened 2023-10-12
README.md: eliminate spurious import in wasm-merge sample output

keithw:
The `first` module's `(import "second" "bar"...` gets satisfied by the `second` module's `(export "bar" (func $func))` and goes away.


kripken (approved):
Thanks!

PR #6007, opened 2023-10-12
Fuzzer: Mark runs where over 50% of functions trap as ignored

kripken:
The number of ignored functions is logged out, so this can help us avoid getting
into a situation where many testcases just trap most of the time rather than
doing anything useful.

50% seems a reasonable cutoff. Even if 50% of functions trap, at least we are
getting 50% that don't, so lots of useful work.

We don't seem to have a problem with this atm, e.g.
```
ITERATION: 32278 [..] 91 ignored
```
So less than 100 ignored out of 32 K there. Still, this could regress with any
fuzzer change in theory.

### DIFF
--- a/scripts/fuzz_opt.py
+++ b/scripts/fuzz_opt.py
@@ -727,6 +727,11 @@ class CompareVMs(TestCaseHandler):
                         # simply trapped, and was not very useful, so mark it
                         # as ignored. Ideally the fuzzer testcases would be
                         # improved to reduce this number.
+                        #
+                        # Note that we don't change output=IGNORE as there may
+                        # still be useful testing here (up to 50%), so we only
+                        # note that this is a mostly-ignored run, but we do not
+                        # ignore the parts that are useful.
                         note_ignored_vm_run(f'(testcase mostly ignored: {calls} calls, {errors} errors)')
                 return output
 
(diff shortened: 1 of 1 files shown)


scripts/fuzz_opt.py
```diff
@@ -718,7 +718,17 @@ class BinaryenInterpreter:
...
+                if output != IGNORE:
+                    calls = output.count(FUZZ_EXEC_CALL_PREFIX)
+                    errors = output.count(TRAP_PREFIX) + output.count(HOST_LIMIT_PREFIX)
+                    if errors > calls / 2:
+                        # A significant amount of execution on this testcase
+                        # simply trapped, and was not very useful, so mark it
+                        # as ignored. Ideally the fuzzer testcases would be
+                        # improved to reduce this number.
+                        note_ignored_vm_run(f'(testcase mostly ignored: {calls} calls, {errors} errors)')
+                return output
```
aheejin, line 731: Should we also modify output to `IGNORE`?

aheejin (approved):

scripts/fuzz_opt.py
kripken: I think it's better not to, as then we'd still get some useful testing even if most functions trap. Though if it's 100% that's not much, obviously... but even 50% seems rare luckily.

PR #6008, opened 2023-10-12
Fuzzer: Move logic for adding a new local on demand to local.get

kripken:
Previously `makeTrappingRefUse` would add a local on demand if one was
missing for the type, and add a tee for it. This PR moves that logic to
`makeLocalGet` so that we get those benefits any time we want to emit a
`local.get` of a local type that does not exist (including from
`makeTrappingRefUse` which calls `makeLocalGet`).


aheejin (approved):

kripken:
@tlively what's our plan for updating `test/lit/fuzz-lattices.test`? That fails in this PR because fuzzer output changes. But the file contains both autogenerated and hand-written portions, so it can't be auto-updated.

I can try to update it manually somehow, but I'm not sure how to tell if the new output is actually testing the same thing as before?

Can we either remove the manual parts, or remove this test?

tlively:
Yeah, let's delete that file.

kripken:
Sounds good, done.

PR #6010, opened 2023-10-13
GUFA: Add missing set of optimized boolean

kripken:
Without marking us as having optimized we didn't refinalize, and broke validation.


test/lit/passes/gufa-refs.wast
```diff
@@ -5730,3 +5730,53 @@
...
+      (struct.new $A)
+    )
+    (drop
+      (block $label (result anyref)
+        (drop
+          (br_on_cast $label anyref (ref struct)
+            ;; This cast can be refined since we know the input is $A. After we
+            ;; do that, we must refinalize, as the br_on_cast's types must be
+            ;; valid - specifically, we can't end up with the input type being
+            ;; $A and the output type still being (ref struct), as the input
```
tlively, line 5771:
```suggestion
            ;; $A and the output type still being (ref struct), as the output
```

tlively (approved):

PR #6015, opened 2023-10-17
Rename getSuperType to getDeclaredSuperType

kripken:
A later PR will add `getSuperType` which will mean "get the general super type -
either declared, or not".

See discussion in #6005


tlively (approved):
Thanks!

PR #6020, opened 2023-10-17
SimplifyGlobals: Run on function code in missing places

kripken:
The pass was written before we had relevant code in module locations, but now
with GC we can have `global.get`s of more things.

The scanner did not run on global code in a way that is not a problem yet, but
will be for a later PR I'll open. It will be tested there. That is, right now there is no
optimization that is confused by the fact that we did not scan code at the module
level, but the next PR will do that.

The use modifier did not run on global code either, which was an actual missed
optimization opportunity: There are cases where we want to modify a `global.get`
to point to another one, and such a get might be in global code, not just in a
function. A test is added for that.


test/lit/passes/simplify-globals-prefer_earlier.wast
```diff
@@ -0,0 +1,53 @@
...
+  (import "a" "b" (global $global1 i32))
+
+  ;; CHECK:      (global $global2 i32 (global.get $global1))
+  (global $global2 i32 (global.get $global1))
+
+  ;; CHECK:      (global $global3 i32 (global.get $global1))
+  (global $global3 i32 (global.get $global2))
+
+  ;; CHECK:      (global $global4 i32 (global.get $global1))
+  (global $global4 i32 (global.get $global3))
```
kripken, line 23: These `global.get`s in global code (inits of globals) are what is now optimized that was not before.

tlively (approved):
lgtm!

PR #6022, opened 2023-10-18
Reuse existing function types for blocks

tlively:
Type annotations on multivalue blocks (and loops, ifs, and trys) are type
indices that refer to function types in the type section. For these type
annotations, the identities of the function types does not matter. As long as
the referenced type has the correct parameters and results, it will be valid to
use.

Previously, when collecting module types, we always used the "default" function
type for multivalue control flow, i.e. we used a final function type with no
supertypes in a singleton rec group. However, in cases where the program already
contains another function type with the expected signature, using the default
type is unnecessary and bloats the type section.

Update the type collecting code to reuse existing function types for multivalue
control flow where possible rather than unconditionally adding the default
function type. Similarly, update the binary writer to use the first heap type
with the required signature when emitting annotations on multivalue control flow
structures. To make this all testable, update the printer to print the type
annotations as well, rather than just the result types. Since the parser was not
able to parse those newly emitted type annotations, update the parser as well.


tlively:
* main
  * **PR #6022** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6022" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/6022?utm_source=stack-comment).

src/passes/Print.cpp
```diff
@@ -242,6 +243,22 @@ struct PrintSExpression : public UnifiedExpressionVisitor<PrintSExpression> {
...
+  std::ostream& printBlockType(Signature sig) {
+    assert(sig.params == Type::none);
+    if (sig.results == Type::none) {
+      return o;
+    }
+    if (sig.results.isTuple()) {
+      if (auto it = signatureTypes.find(sig); it != signatureTypes.end()) {
+        o << "(type ";
+        printHeapType(it->second);
+        o << ") ";
```
kripken, line 255: Do we need to print `(type $foo) (result i32 i32)`? Can't we just print `(result i32 i32)` without the type's name?

src/passes/Print.cpp
```diff
@@ -242,6 +243,22 @@ struct PrintSExpression : public UnifiedExpressionVisitor<PrintSExpression> {
     return printPrefixedTypes("param", type);
   }
 
+  std::ostream& printBlockType(Signature sig) {
```
kripken, line 246: Maybe `printControlFlowType`?

test/lit/wasm-split/instrument-memory64.wast
```diff
@@ -33,7 +91,7 @@
 ;; CHECK-NEXT:   (block
 ;; CHECK-NEXT:    (i64.store align=1
 ;; CHECK-NEXT:     (local.get $addr)
-;; CHECK-NEXT:     (i64.const {{.*}})
+;; CHECK-NEXT:     (i64.const -6733849689813251748)
```
kripken, line 94: I think this file has automatic changes we don't want?

kripken (commented):
Nice, basically lgtm but I have some questions.

test/lit/wasm-split/instrument-memory64.wast
tlively: Yep, undoing 👍

src/passes/Print.cpp
tlively: Technically, if we only print `(result i32 i32)`, then that implicitly declares and uses this type if it does not otherwise exist in the module: `(rec (type (sub final (func (result i32 i32)))))).0`. By printing the real heap type we want to refer to, we avoid defining that new type.

This doesn't actually matter in practice because we don't store a heap type on control flow structures, just a signature, but it's good to emit text that more precisely matches what our binaries will look like. In particular, it would make a difference if you had some other tool parse our text format.

src/passes/Print.cpp
tlively: In the spec this is called `blocktype`, so I think it is useful to match that terminology.

kripken (approved):
Makes sense, thanks.

PR #6023, opened 2023-10-18
SimplifyGlobals: Fold single-use globals to their use

kripken:
For example,
```wat
(global $a (struct.new $A))
(global $b (struct.new $B (global.get $A)))

=>

(global $b (struct.new $B (struct.new $A)))
```
and the global `$a` is now unused. This is valid if `$a` has no other uses.

This saves a little in code size, but should not really help otherwise, as we
already look through immutable `global.get` operations in important
optimizations. But the code size may matter if there are many such single-
use globals.


src/passes/SimplifyGlobals.cpp
```diff
@@ -672,6 +677,62 @@ struct SimplifyGlobals : public Pass {
...
+
+      void visitGlobalGet(GlobalGet* curr) {
+        // If this is a get of a global with a single get and no sets, then we
+        // can fold that code into here.
+        auto name = curr->name;
+        auto& info = infos[name];
+        if (info.written == 0 && info.read == 1) {
+          auto* global = wasm.getGlobal(name);
+          if (global->init) {
+            // Steal that global's code. To keep that global valid, make it an
```
tlively, line 713: Why do we need to keep the global valid? It seems like it would be pretty simple to remove it here, too.

tlively (approved):

src/passes/SimplifyGlobals.cpp
kripken: Yeah, I guess it would just be a loop on `module->globals` and a decision on which to keep. Still, we'd need to make sure the passes after us don't get confused by that (e.g. there is an info entry for a global that doesn't exist), and that's more code than the one-line operation here.

We do remove such unused globals in the very next pass after this one anyhow, so that seems simplest to me.

tlively: sgtm

kripken:
I tested on some real-world code for more coverage as I fuzz, and that found a real problem. It's not always ok to turn such a global into an import, because it makes the type public... so we should remove them here as you said @tlively

kripken:
I ended up just copying the code, which is even simpler than what we had before. Slightly more work, but really not much.

After the fix, on a Dart testcase I have I see this PR remove 2/3 of the 1,402 globals, which ends up reducing code size by 6.5%. cc @askeksa

I don't see any change on Java, I guess they don't emit a pattern that benefits from this yet.

tlively (approved):

PR #6024, opened 2023-10-18
[Outlining] Filter branches & returns

ashleynh:
Adds a function to filter branch and return instructions from being included in potential outlining candidates.

### DIFF
--- a/src/passes/stringify-walker.h
+++ b/src/passes/stringify-walker.h
@@ -234,11 +234,20 @@ struct HashStringifyWalker : public StringifyWalker<HashStringifyWalker> {
 // Functions that filter vectors of SuffixTree::RepeatedSubstring
 struct StringifyProcessor {
   static std::vector<SuffixTree::RepeatedSubstring>
-  dedupe(const std::vector<SuffixTree::RepeatedSubstring>&& substrings);
+  dedupe(const std::vector<SuffixTree::RepeatedSubstring>& substrings);
+  // Filter is the general purpose function backing subsequent filter functions.
+  // It can be used directly, but generally prefer a wrapper function
+  // to encapsulate your condition and make it available for tests
   static std::vector<SuffixTree::RepeatedSubstring>
-  filter(const std::vector<SuffixTree::RepeatedSubstring>&& substrings,
+  filter(const std::vector<SuffixTree::RepeatedSubstring>& substrings,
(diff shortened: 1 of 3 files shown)


src/passes/stringify-walker.h
```diff
@@ -235,10 +235,19 @@ struct HashStringifyWalker : public StringifyWalker<HashStringifyWalker> {
...
+  // Filter is the general purpose function backing subsequent filter functions.
+  // It can be used directly, but generally prefer a wrapper function
+  // to encapsulate your condition and make it available for tests
   static std::vector<SuffixTree::RepeatedSubstring>
   filter(const std::vector<SuffixTree::RepeatedSubstring>&& substrings,
          const std::vector<Expression*> exprs,
          std::function<bool(const Expression*)> condition);
+  static std::vector<SuffixTree::RepeatedSubstring>
+  filterLocalSets(const std::vector<SuffixTree::RepeatedSubstring>&& substrings,
+                  const std::vector<Expression*> exprs);
```
tlively, line 247: Actually, if these functions don't re-use the space for the input vector, then taking them as `const vector&` should be sufficient. `vector&&` would allow you to safely reuse the input vector as the output vector, but `const vector&&` isn't really useful.

test/gtest/stringify.cpp
```diff
@@ -335,14 +335,55 @@ TEST_F(StringifyTest, FilterLocalSets) {
...
     result,
     (std::vector<SuffixTree::RepeatedSubstring>{
       // sequence i32.const 0, i32.const 1 appears at idx 6 and again at 16
       SuffixTree::RepeatedSubstring{2u, (std::vector<unsigned>{6, 16})}}));
 }
+
+TEST_F(StringifyTest, FilterBranches) {
+  static auto branchesModuleText = R"wasm(
+  (module
+  (func $a (result i32)
```
tlively, line 351: It would be good to indent everything inside the module.
```suggestion
  (module
   (func $a (result i32)
```

tlively (approved):
LGTM % comments

src/passes/stringify-walker.h
ashleynh: done!

test/gtest/stringify.cpp
ashleynh: done!

PR #6026, opened 2023-10-18
RemoveUnusedModuleElements: Make exports skip trampolines

kripken:
If we export a function that just calls another function, we can export that one
instead. Then the one in the middle may be unused,

```js
function foo() {
  return bar();
}

export foo; // can be an export of bar
```

This saves a few bytes in rare cases, but probably more important is that it saves
the trampoline, so if this is on a hot path, we save a call.

Context: https://github.com/emscripten-core/emscripten/issues/20478#issuecomment-1767386125

In general this is not needed as inlining does it: in the example above, `bar` would be
inlined into `foo` anyhow usually, and then they'd be identical and merged. But in
some cases we don't inline, like if the target is an import, which is how this came up.
But now that I think of it, there are other cases, so I'll mark this draft to keep thinking
about it.


tlively (commented):
Beware the case where the callee is a subtype of the caller, which may cause this transformation to make new types public.

kripken:
That should be handled, this only does the optimization if the types are equal,

https://github.com/WebAssembly/binaryen/pull/6026/files#diff-f0c968562d115b75fb1671c7b72222e08776d01059355efc9271c96e7b720c30R755-R760

kripken:
After thinking on this, I think it does handle all the cases we need. Exports are just special, and we need to handle them, and all other cases end up optimized in the caller of the trampoline (not necessarily the trampoline itself). Exports are special since the caller is outside the module, so inlining won't work, and we need this PR.

RReverser (approved):
I'm not familiar with test structure so don't really understand why some CHECKs were moved, but the code itself looks reasonable. As you said, this is likely a rare edge case, but still nice to see it eliminated.

kripken:
Funny how various tests across the test suite have benefited from this. It's all minor, and in artificial testcases, but this might be more common than I'd thought.

src/passes/RemoveUnusedModuleElements.cpp
```diff
@@ -712,6 +714,65 @@ struct RemoveUnusedModuleElements : public Pass {
...
+      if (exp->kind != ExternalKind::Function) {
+        continue;
+      }
+
+      auto* func = module->getFunction(exp->value);
+      if (!func->body) {
+        continue;
+      }
+
+      auto* call = func->body->dynCast<Call>();
```
RReverser, line 750: Should this also check for `ReturnCall`? (assuming that tail calls are represented via different node in Binaryen)

kripken: ReturnCall is the same as Call, just with `isReturn` set to true, so it's already handled. I'll add a test though for coverage.

test/lit/passes/remove-unused-module-elements_all-features.wast
```diff
@@ -761,3 +753,166 @@
...
+    (call $import-A)
+  )
+
+  (func $middle-B-B (type $B)
+    (call $import-B)
+  )
+)
+
+;; As above, but checking for parameters: It's ok to pass values through, but
+;; not to do anything else.
```
tlively, line 851: Maybe add one where the order of the parameters is changed?

tlively (approved):

src/passes/RemoveUnusedModuleElements.cpp
RReverser: Ah good to know.

tlively:
This seems to break emscripten dynamic loading. Perhaps the linking section information is no longer correct after this transformation? I've paused the auto-roller pending a fix: https://chromium-review.googlesource.com/c/emscripten-releases/+/4955360

RReverser:
Unless I'm missing something in my sample, it seems Clang (wasm-ld?) in general can't handle re-exporting imported function, e.g.:

```c
__attribute__((used))
__attribute__((import_name("foo")))
__attribute__((export_name("bar")))
int baz();
```

with pure Clang results in:

```
temp.wasm:      file format wasm 0x1

Section Details:

Memory[1]:
 - memory[0] pages: initial=2
Export[1]:
 - memory[0] -> "memory"
Custom:
 - name: "producers"
Custom:
 - name: "target_features"
  - [+] mutable-globals
  - [+] sign-ext
```

I then tried adding another inline exported function just in case I messed up attribute syntax, but that one ends up exported fine:

```c
__attribute__((used))
__attribute__((import_name("foo")))
__attribute__((export_name("bar")))
int baz();

__attribute__((export_name("bar2")))
int baz2() {
	return baz();
}
```

```
temp.wasm:      file format wasm 0x1

Section Details:

Type[1]:
 - type[0] () -> i32
Import[1]:
 - func[0] sig=0 <env.foo> <- env.foo
Function[1]:
 - func[1] sig=0 <bar2>
Memory[1]:
 - memory[0] pages: initial=2
Export[2]:
 - memory[0] -> "memory"
 - func[1] <bar2> -> "bar2"
Code[1]:
 - func[1] size=4 <bar2>
Custom:
 - name: "producers"
Custom:
 - name: "target_features"
  - [+] mutable-globals
  - [+] sign-ext
```

Maybe this is Clang-specific issue where it ignores `export_name` attribute if `import_name` is already provided, but in combination with the failure you linked seems it might be somewhere deeper, likely in wasm-ld.

kripken:
That part definitely looks like a `wasm-ld` bug.

Testcase:
```wat
(module
 (type $0 (func (param i32) (result i32)))
 (type $1 (func (param i32)))
 (type $2 (func))
 (import "env" "malloc" (func $fimport$0 (param i32) (result i32)))
 (import "env" "free" (func $fimport$1 (param i32)))
 (export "__wasm_call_ctors" (func $0))
 (export "__wasm_apply_data_relocs" (func $0))
 (export "mallocproxy" (func $fimport$0))
 (export "freeproxy" (func $fimport$1))
 (func $0
 )
 ;; dylink section
 ;;   memorysize: 0
 ;;   memoryalignment: 0
 ;;   tablesize: 0
 ;;   tablealignment: 0
)
```
```bash
$ wasm-ld liblib.so -o out
wasm-ld: error: liblib.so: invalid function export
```
That's simplified from `./test/runner core2.test_dlfcn_mallocs`.

I assume we should file an LLVM bug with this @tlively @sbc100 ? If so I can do that.

I also see this uncovered some other bugs in Binaryen, so we definitely should disable this optimization for now. I'll open PRs for that.

sbc100:
I guess we should make it so that trying to set `export_name` on an undefined symbol fails at compile time.

kripken:
@sbc100 To be sure we're on the same page, this isn't about the object file format. This is in dynamic linking. That is, we optimize the dynamic library with binaryen, causing this import-that-is-exported, and then we have a dynamic library, which we give to `wasm-ld` during the link of the main executable (I'm guessing that's to know how to error on unresolved symbols?), and it errors at that time.

> I guess we should make it so that trying to set export_name on an undefined symbol fails at compile time.

IIUC, that would help create such import-that-is-exported in clang, but not by another tool.

In any case this is not urgent as we reverted the optimization, but it would be nice to support this eventually in wasm-ld. It's just a bit of code size improvement but it saves a call, so it might matter at runtime.

sbc100:
I see, yes it would be good if wasm-ld could accept the optimized side module.  I worry it will be hard to create a test for that on the wasm-ld side though.. perhaps we could need to resort to checking in a binary :-/

PR #6032, opened 2023-10-19
fix: Avoid segfault in catch validator

peblair:
While working on testing some new [binaryen.ml](https://github.com/grain-lang/binaryen.ml) bindings for Grain, I came across a segmentation fault in the Binaryen WASM validator. The problem is that, if you construct a `try` expression which references a nonexistent tag in one of its `catch` blocks, the validation code successfully identifies the `null` pointer but then proceeds to try to read from it. I have updated the validation code to mitigate this problem and added in a unit test. I verified that backing out the change to `wasm-validator.cpp` causes the unit test file to segfault.

Note that this is not an issue that can be reproduced with the Python `check.py` test suite, since the s-expression parser will reject the malformed `try` block.

Apologies in advance for any faux pas in the test; I don't have much C++ experience.

### DIFF
--- a/src/wasm/wasm-validator.cpp
+++ b/src/wasm/wasm-validator.cpp
@@ -2372,38 +2372,36 @@ void FunctionValidator::visitTry(Try* curr) {
     auto* tag = getModule()->getTagOrNull(tagName);
     if (!shouldBeTrue(tag != nullptr, curr, "")) {
       getStream() << "tag name is invalid: " << tagName << "\n";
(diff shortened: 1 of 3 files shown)


kripken (approved):
Thanks!

Looks like there is a style change that needs to be applied, see the CI error (at the top there it says how to do that).

peblair:
@kripken Thanks for the quick approval! The style check should pass now.

PR #6033, opened 2023-10-20
[analysis][NFC] Create a TransferFunction concept

tlively:
Factor the static assertions for transfer functions out into a new
transfer-function.h header. The concept requires the `getDependents` method to
return an input range of basic blocks, and to satisfy that requirement, fix up
_indirect_ptr_iterator in cfg-impl.h so that it is a proper iterator. Remove
part of the lattice fuzzer that was using a placeholder transfer function in a
way that does not satisfy the new type constraints; most of that code will be
overhauled in the future anyway.

PR #6035, opened 2023-10-21
[analysis][NFC] Rename `makeLeastUpperBound` to `join` and move it to lattice

tlively:
In general, the operation of taking the least upper bound of two lattice
elements may depend on some state stored in the lattice object rather than in
the elements themselves. To avoid forcing the elements to be larger and more
complicated in that case (by storing a parent pointer back to the lattice), move
the least upper bound operation to make it a method of the lattice rather than
the lattice element. This is also more consistent with where we put e.g. the
`compare` method.

While we are at it, rename `makeLeastUpperBound` to `join`, which is much
shorter and nicer. Usually we avoid using esoteric mathematical jargon like
this, but "join" as a normal verb actually describes the operation nicely, so I
think it is ok in this case.

### DIFF
--- a/src/analysis/lattices/powerset.h
+++ b/src/analysis/lattices/powerset.h
@@ -71,11 +71,6 @@ public:
     bool isTop() const { return count() == bitvector.size(); }
     bool isBottom() const { return count() == 0; }
 
-    // Calculates the LUB of this element with some other element and sets
-    // this element to the LUB in place. Returns true if this element before
-    // this method call was different than the LUB.
-    bool makeLeastUpperBound(const Element& other) noexcept;
-
     // Prints out the bits in the bitvector for a lattice element.
     void print(std::ostream& os);
 
@@ -89,6 +84,11 @@ public:
 
   // Returns an instance of the bottom lattice element.
   Element getBottom() const noexcept;
+
+  // Calculates the LUB of this element with some other element and sets
+  // this element to the LUB in place. Returns true if this element before
+  // this method call was different than the LUB.
+  bool join(Element& self, const Element& other) const noexcept;
 };
 
 // A layer of abstraction over FiniteIntPowersetLattice which maps
@@ -149,6 +149,10 @@ public:
   }
 
   Element getBottom() const noexcept { return intLattice.getBottom(); }
+
+  bool join(Element& self, const Element& other) const noexcept {
+    return intLattice.join(self, other);
+  }
 };
 
 } // namespace wasm::analysis
--- a/src/analysis/lattices/powerset.h
+++ b/src/analysis/lattices/powerset.h
@@ -85,9 +85,9 @@ public:
   // Returns an instance of the bottom lattice element.
   Element getBottom() const noexcept;
 
-  // Calculates the LUB of this element with some other element and sets
-  // this element to the LUB in place. Returns true if this element before
-  // this method call was different than the LUB.
+  // Modifies `self` to be the join (aka least upper bound) of `self` and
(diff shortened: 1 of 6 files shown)


tlively:
* main
  * **PR #6033** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6033" target="_blank">Graphite</a> 
    * **PR #6034** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6034" target="_blank">Graphite</a> 
      * **PR #6035** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6035" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/6035?utm_source=stack-comment).

src/analysis/lattices/powerset.h
```diff
@@ -89,6 +84,11 @@ class FiniteIntPowersetLattice {
 
   // Returns an instance of the bottom lattice element.
   Element getBottom() const noexcept;
+
+  // Calculates the LUB of this element with some other element and sets
+  // this element to the LUB in place. Returns true if this element before
+  // this method call was different than the LUB.
```
kripken, line 90: Maybe comment on how the name is a shorthand, as explained in the PR description?

kripken (approved):

src/analysis/lattices/powerset.h
ashleynh: +1 yes please

src/analysis/lattices/stack.h
```diff
@@ -258,7 +215,50 @@ template<Lattice L> class StackLattice {
...
+  // the bottom of the higher stack will be kept, while the LUB of the
+  // corresponding tops of each stack will be taken. For instance,
+  // LUB([d, c, b, a], [b', a']) is [d, c, LUB(b, b'), LUB(a, a')].
+  //
+  // We start at the top because it makes taking the LUB of stacks with
+  // different scope easier, as mentioned at the top of the file. It also
+  // fits with the conception of the stack starting at the top and having
+  // an infinite bottom, which allows stacks of different height and scope
+  // to be easily joined.
+  bool join(Element& self, const Element& other) const noexcept {
```
ashleynh, line 230: Nit: Seems a bit weird to call the first element "self" as it implies the instance of the StackLattice class. Wdyt?

ashleynh (approved):

src/analysis/lattices/stack.h
tlively: Fair enough. Not sure what better to call them than `a` and `b` or something similarly unimaginative, though. Do you have ideas?

ashleynh: The Obj-C dev in me quite liked the previous naming of `makeLeastUpperBound` because it encoded mutability at the function's call site. 

Since the first argument always has to match the LUB, maybe we can call it LUB?

tlively: Hmm, the first argument is the LUB at the end of the method, but not necessarily at the beginning. It would be nice to have names that make sense both before and after.

Maybe `joinee` and `joiner`?

tlively: Will do this globally as a follow-up.

tlively:
### Merge activity

* **Oct 25, 1:00 PM**: @tlively started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/6035).

PR #6048, opened 2023-10-24
Fix the case of an unreachable entry in LocalGraph

kripken:
Followup to #6046 - the fuzzer found we missed handling the case of the entry itself
being unreachable and there being a loop after it, see code and testcase.

Also add a "dump" impl in that file which helps debugging.

### DIFF
--- a/src/ir/LocalGraph.cpp
+++ b/src/ir/LocalGraph.cpp
@@ -110,11 +110,6 @@ struct Flower : public CFGWalker<Flower, Visitor<Flower>, Info> {
     auto numLocals = func->getNumLocals();
     std::vector<FlowBlock*> work;
 
-    // Track if we have unreachable code anywhere, as if we do that may inhibit
-    // certain optimizations below. If there are fewer live blocks than the set
-    // of all blocks, then some are dead.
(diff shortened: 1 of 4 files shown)


kripken:
Unfortunately this is not enough either... it's possible depending on unreachability here is not the best idea.

kripken:
A proper analysis of unreachability may not be worth it, as it would add overhead that could easily overtake the benefits of this optimization. So I am considering just reverting the previous PR, sadly.

kripken:
I pushed a proper fix, but still need to measure if the slowdown is worth it.

kripken:
Sadly this does make us a little slower. Just 1-2% or so, but as the whole point here was an optimization, it makes it not worthwhile. I'll experiment with the other idea of relaxing those assertions.

kripken:
I pushed a version that relaxes the assertion as @tlively suggested, and am fuzzing it now. It may be the best option here.

kripken:
This passed 12K fuzzer iterations, so I think it is stable, and is the best solution we have AFAICT.

tlively (approved):

PR #6040, opened 2023-10-21
[analysis] Implement a Lift lattice

tlively:
This lattice "lifts" another lattice by inserting a new bottom element
underneath it.

PR #6039, opened 2023-10-21
[analysis] Implement Flat lattice

tlively:
Given a type `T`, `Flat<T>` is the lattice where none of the values of `T` are
comparable except with themselves, but they are all greater than a common bottom
element not in `T` and less than a common top element also not in `T`.

PR #6038, opened 2023-10-21
[analysis] Add a FullLattice concept and Inverted lattice

tlively:
The FullLattice concept extends the base Lattice with `getTop` and `meet`
operations. The `Inverted` lattice uses these operations to reverse the order of
an arbitrary full lattice, for example to create a lattice of integers ordered
by `>` rather than by `<`.

PR #6037, opened 2023-10-21
[analysis] Implement an Int lattice

tlively:
Implement a generic lattice template for integral types ordered by `<`.

PR #6050, opened 2023-10-25
[analysis] Improve lattice fuzzer

tlively:
Implement new `RandomLattice` and `RandomFullLattice` utilities that are
lattices randomly created from other lattices. By recursively using themselves
as the parameter lattices for lattices like `Inverted` and `Lift`, these random
lattices can become arbitrarily nested.

Decouple the checking of lattice properties from the checking of transfer
function properties by creating a new, standalone `checkLatticeProperties`
function.

### DIFF
--- a/src/tools/wasm-fuzz-lattices.cpp
+++ b/src/tools/wasm-fuzz-lattices.cpp
@@ -50,8 +50,9 @@ uint64_t getSeed() {
 
 // Actually a pointer to `L::ElementImpl`, but we erase the type to avoid
 // getting into a situation where `L` satisfying `Lattice` or `FullLattice`
-// circularly requires that `L` satisfies `Lattice` or `FullLattice`. Also make
-// the pointer copyable to satisfy that constraint on lattice elements.
+// circularly requires that `L` satisfies `Lattice` or `FullLattice`. C++ does
+// not allow concepts to depend on themselves. Also make the pointer copyable to
+// satisfy that constraint on lattice elements.
 template<typename L>
 struct RandomElement : std::unique_ptr<void, void (*)(void*)> {
   RandomElement() = default;
@@ -91,6 +92,9 @@ struct RandomElement : std::unique_ptr<void, void (*)(void*)> {
 };
 
 struct RandomFullLattice {
+  // The inner lattice and lattice element types. These must be defined later
+  // because they depend on `RandomFullLattice` satisfying `FullLattice`, but
+  // that requires the type to be complete.
   struct L;
   struct ElementImpl;
   using Element = RandomElement<RandomFullLattice>;
@@ -104,6 +108,7 @@ struct RandomFullLattice {
                     size_t depth = 0,
                     std::optional<uint32_t> maybePick = std::nullopt);
 
+  // Make a random element of this lattice.
   Element makeElement() const noexcept;
 
   Element getBottom() const noexcept;
@@ -114,6 +119,9 @@ struct RandomFullLattice {
 };
 
 struct RandomLattice {
+  // The inner lattice and lattice element types. These must be defined later
+  // because they depend on `RandomLattice` satisfying `Lattice`, but that
+  // requires the type to be complete.
   struct L;
   struct ElementImpl;
   using Element = RandomElement<RandomLattice>;
@@ -125,6 +133,7 @@ struct RandomLattice {
 
   RandomLattice(Random& rand, size_t depth = 0);
 
+  // Make a random element of this lattice.
   Element makeElement() const noexcept;
 
   Element getBottom() const noexcept;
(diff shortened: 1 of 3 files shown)


tlively:
* main
  * **PR #6033** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6033" target="_blank">Graphite</a> 
    * **PR #6034** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6034" target="_blank">Graphite</a> 
      * **PR #6035** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6035" target="_blank">Graphite</a> 
        * **PR #6036** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6036" target="_blank">Graphite</a> 
          * **PR #6037** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6037" target="_blank">Graphite</a> 
            * **PR #6038** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6038" target="_blank">Graphite</a> 
              * **PR #6039** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6039" target="_blank">Graphite</a> 
                * **PR #6040** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6040" target="_blank">Graphite</a> 
                  * **PR #6050** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6050" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/6050?utm_source=stack-comment).

kripken (commented):
The diff here seems to show a lot of code I've already reviewed. Should this be rebased on that chain of PRs that is landing now?

tlively:
@kripken, the diff should now be fixed.

src/tools/wasm-fuzz-lattices.cpp
```diff
@@ -35,14 +42,421 @@ using namespace analysis;
...
   return std::uniform_int_distribution<uint64_t>{}(rand);
 }
 
+// Actually a pointer to `L::ElementImpl`, but we erase the type to avoid
+// getting into a situation where `L` satisfying `Lattice` or `FullLattice`
+// circularly requires that `L` satisfies `Lattice` or `FullLattice`. C++ does
+// not allow concepts to depend on themselves. Also make the pointer copyable to
+// satisfy that constraint on lattice elements.
+template<typename L>
+struct RandomElement : std::unique_ptr<void, void (*)(void*)> {
```
kripken, line 58: To be sure I follow, this is a unique pointer to a void (!), and also has a destructor of type `void *(void*)`?

src/tools/wasm-fuzz-lattices.cpp
```diff
@@ -35,14 +42,421 @@ using namespace analysis;
...
+  RandomElement(typename L::ElementImpl&& other)
+    : std::unique_ptr<void, void (*)(void*)>(
+        new typename L::ElementImpl(std::move(other)),
+        [](void* e) { delete static_cast<typename L::ElementImpl*>(e); }) {}
+
+  RandomElement(const RandomElement& other)
+    : RandomElement([&]() {
+        auto copy = *other;
+        return copy;
+      }()) {}
```
kripken, line 70: Can't `other` be unallocated before the lambda here is called? The lambda doesn't keep it alive AFAIK.

src/tools/wasm-fuzz-lattices.cpp
tlively: Correct!

src/tools/wasm-fuzz-lattices.cpp
tlively: This lambda is only called immediately as part of the initialization expression, so `other` will definitely still be alive. The only reason we have a lambda here is that I don't know how to make a copy of a referenced value without assigning it to some lvalue. We need a copy of `other` in the first place because we can't pass `other` directly to the other constructor because we can't move out of it.

tlively:
The test failure on alpine is unrelated and fixed by #6054.

src/tools/wasm-fuzz-lattices.cpp
```diff
@@ -35,14 +42,421 @@ using namespace analysis;
...
+
+  const typename L::ElementImpl* operator->() const { return &*(*this); }
+};
+
+struct RandomFullLattice {
+  // The inner lattice and lattice element types. These must be defined later
+  // because they depend on `RandomFullLattice` satisfying `FullLattice`, but
+  // that requires the type to be complete.
+  struct L;
+  struct ElementImpl;
```
kripken, line 100: Wait, C++ lets you do a forward declaration in the struct definition, and you can define the forward-declared thing later outside? So you could even define it in another file..? How can that work?

tlively: Yep, just like for a method. The type isn't complete until you define it, of course, which is why many of the methods below also have to be defined out of line, and after `L` and `ElementImpl` have been defined.

tlively: Unfortunately you can't forward-declare a `using` statement, which is why these have to be expressed with structs and trivial inheritance.

kripken (approved):
lgtm but I worry about some day needing to debug this intricate C++ code my self :smile: but as it is in fuzzer code I worry less.

Issue #5271, opened 2022-11-17
Asyncify lists does not work correctly with dots


caiiiycuk:
Hi. I am using asyncify for jsdos project. And I found that `...` is not handled correctly, for example I have function:
```
Program::WriteOut(char const*, ...)
```

When I compile I have warning:
```
warning: Asyncify onlylist contained a non-matching pattern: Program::WriteOut(char const*,    ) (Program::WriteOut\28char\20const*\2c\20\20\20\20\29)
```

As you can see dots are replaced with spaces:
```
expected (from wasm dis): Program::WriteOut\28char\20const*\2c\20...\29
actual                  : Program::WriteOut\28char\20const*\2c\20\20\20\20\29
```

kripken:
Hmm, that could be a problem in wasm-ld where it demangles names into the names section, or it could be a problem in binaryen where it reads it. Can you look in the names section in the wasm binary (before binaryen runs) and see what appears there?

kripken:
(I mean, to look with a hex editor - `wasm-dis` may be too late, if the bug is in binaryen parsing.)

caiiiycuk:
I have this in wdosbox-x.wasm:
```
F#Program::WriteOut(char const*, ...)Ø^F+Program::WriteOut(char const*, char const*)Ù^F.Program::WriteOut_NoParsing(char const*, bool)Ú^
```

kripken:
Looks correct, thanks... ok, I dug into this now since it's clear it's either an emscripten or a binaryen issue. Turns out it's on the emscripten side,

https://github.com/emscripten-core/emscripten/pull/13477

The function `unmangle_symbols_from_cmdline` replaces `.` with spaces. It looks like `.` is part of the escaping approach there, but the problem is that `.` is also a necessary character sometimes, with things like `...` that appear in signatures.

cc @juj - what do you think the best solution is? One idea might be to not run `unmangle_symbols_from_cmdline` if the input arrives from a response file. Then at least response files would be a way to know exactly what you are getting. But maybe the escaping can be tweaked?

juj:
> cc @juj - what do you think the best solution is? One idea might be to not run unmangle_symbols_from_cmdline if the input arrives from a response file. Then at least response files would be a way to know exactly what you are getting. But maybe the escaping can be tweaked?

Ouch, ellipsis never occurred to me, I thought I was taking three characters that were all never used in the LLVM name mangling.

[...]

PR #6051, opened 2023-10-25
Fix #6047, #5271: Support one-line-one-function file format for asyncify lists

### LINKED ISSUE #6047: handleBracketingOperators (src/support/string.h) is broken, and list parsing in asyncify is not functioning correctly.
Hi, I've come across a corner case where asyncify list parsing is failing with an error:
```
 Fatal: Asyncify: failed to parse lists
````

The list parsing mechanism is attempting to calculate nesting operators from the code:

> Handles bracketing in a list initially split by ",", but the list may contain nested ","s. For example,
> **void foo(int, double)**
> must be kept together because of the "(". Likewise, "{", "<", "[" are handled.

The implementation performs simple nesting counting:
```c++
    for (const char c : part) {
      if (c == '(' || c == '<' || c == '[' || c == '{') {
        nesting++;
      } else if (c == ')' || c == '>' || c == ']' || c == '}') {
        nesting--;
      }
    }
```

This doesn't work for the function `XStream::operator>(int&)`. Obviously, the nesting will be -1 at the end, and a fatal error will occur.

I don't have an idea on how to easily fix this. Moreover, I've been using asyncify for a long time, and this list format is hard to maintain. I think that using a one-line-one-function file format is much easier to maintain and will not have this issue. 

What do you think? Any suggestion how to fix this?

caiiiycuk:
Fixes:
- #6047
- #5271


kripken (approved):
lgtm with some minor suggestions.

kripken:
`delim != "\n";` seems better to me, as newlines are special: they are the one case that we can just split naively using them, without scoping/bracketing.

kripken:
@caiiiycuk it looks like this breaks the auto-updater script,
```
$ ./scripts/auto_update_tests.py lit
[..]
Fatal: Failed opening '%S/asyncify-foo,bar-nl.txt'
[..]
subprocess.CalledProcessError: Command 'binaryen/scripts/foreach.py binaryen/test/lit/passes/asyncify_pass-arg=asyncify-onlylist@foo,bar.wast /tmp/tmptkbv0v5t wasm-opt --asyncify --pass-arg=asyncify-onlylist@@%S/asyncify-foo,bar-nl.txt -S -o -' returned non-zero exit status 1.
[..]
subprocess.CalledProcessError: Command '['/usr/bin/python3', 'binaryen/scripts/update_lit_checks.py', '--binaryen-bin=binaryen/bin', 'binaryen/test/lit/**/*.wast', 'binaryen/test/lit/**/*.wat']' returned non-zero exit status 1.
```

We should run that in CI, but don't atm, which is how this was missed.

What is `%S` trying to do there?

caiiiycuk:
@kripken 
`asyncify-foo,bar-nl.txt` is located in the same folder as a test. However, if you just name it without specifying `%S`, it may not work correctly. This is because the current working directory (cwd) might be the root of the project or even some other folder. After briefly checking the documentation, it seems that `%S` refers to the source directory (the directory of the file currently being run), and this could potentially resolve the issue. I apologize for not testing it in different folders to confirm.

caiiiycuk:
Ah, it's not related to different folders. I didn't imagine that I could break the auto-update, but now it seems obvious. Sorry

kripken:
@caiiiycuk I'm not sure I understand, but it sounds like you have an idea of what went wrong here - do you know how to fix it?

caiiiycuk:
@kripken https://github.com/WebAssembly/binaryen/pull/6075/files

PR #6057, opened 2023-10-27
[analysis] Implement an array lattice

tlively:
The elements of `Array<L, N>` lattice are arrays of length `N` of elements of
`L`, compared pairwise with each other. This lattice is a concrete
implementation of what would be written L^N with pen and paper.

PR #6056, opened 2023-10-27
[analysis][NFC] Rename parameters to join and meet methods

tlively:
Since these methods, which operate on lattice elements, moved to the lattice
types, it no longer makes much sense for their parameters to be called `self`
and `other`. Rename them to `joinee` and `joiner` for joins and `meetee` and
`meeter` for meets.

PR #6058, opened 2023-10-27
[analysis] Implement a vector lattice

tlively:
The vector lattice is nearly identical to the array lattice, except that the
size of the elements is specified at runtime when the lattice object is created
rather than at compile time. The code and tests are largely copy-pasted and
fixed up from the array implementation, but there are a couple differences.
First, initializing vector elements does not need the template magic used to
initialize array elements. Second, the obvious implementations of join and meet
do not work for vectors of bools because they might be specialized to be bit
vectors, so we need workarounds for that particular case.

### DIFF
--- a/src/analysis/lattices/vector.h
+++ b/src/analysis/lattices/vector.h
@@ -47,7 +47,8 @@ template<Lattice L> struct Vector {
     return Element(size, lattice.getTop());
   }
 
-  // `a` <= `b` if their elements are pairwise <=, etc.
+  // `a` <= `b` if their elements are pairwise <=, etc. Unless we determine
+  // that there is no relation, we must check all the elements.
   LatticeComparison compare(const Element& a, const Element& b) const noexcept {
     assert(a.size() == size);
     assert(b.size() == size);
--- a/src/analysis/lattices/vector.h
+++ b/src/analysis/lattices/vector.h
@@ -14,11 +14,10 @@
  * limitations under the License.
  */
 
-#ifndef wasm_analysis_lattices_array_h
-#define wasm_analysis_lattices_array_h
+#ifndef wasm_analysis_lattices_vector_h
+#define wasm_analysis_lattices_vector_h
 
-#include <array>
-#include <utility>
+#include <vector>
 
 #include "../lattice.h"
 #include "bool.h"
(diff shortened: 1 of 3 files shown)


tlively:
* main
      * **PR #6058** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6058" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/6058?utm_source=stack-comment).

src/analysis/lattices/vector.h
```diff
@@ -0,0 +1,137 @@
...
+          result = LESS;
+          continue;
+        case GREATER:
+          if (result == LESS) {
+            // Cannot be both greater and less.
+            return NO_RELATION;
+          }
+          result = GREATER;
+          continue;
+      }
```
kripken, line 75: That we copy-paste this switch logic is kind of disappointing. I'd have hoped with all the template magic there would be a way to avoid duplication? The same for the functions below.

src/tools/wasm-fuzz-lattices.cpp
```diff
@@ -235,6 +250,14 @@ RandomFullLattice::Element RandomFullLattice::makeElement() const noexcept {
...
   }
+  if (const auto* l = std::get_if<Vector<RandomFullLattice>>(lattice.get())) {
+    std::vector<typename RandomFullLattice::Element> elem;
+    elem.reserve(l->size);
+    for (size_t i = 0; i < l->size; ++i) {
+      elem.push_back(l->lattice.makeElement());
+    }
+    return ElementImpl{std::move(elem)};
+  }
   WASM_UNREACHABLE("unexpected lattice");
```
kripken, line 261: I won't ask for a change here, but I really do think `virtual` would be a nicer way to write code like this, rather than have to keep modifying RandomFullLattice for each type. That is, I would personally gone the other route and had a virtual method that this class calls, each new lattice just implements a "makeRandom" interface, so it all plugs in.

src/analysis/lattices/vector.h
tlively: You mean the copy-paste between the array and vector lattices? I could probably try to use even more template magic to reduce the duplication, but I think at that point we could start calling it template obfuscation 😅

src/tools/wasm-fuzz-lattices.cpp
tlively: I plan to at least get rid of the distinction between `RandomFullLattice` and `RandomLattice`, which should simplify things.

I can also clean up this giant if-else chain by using `std::visit` combined with overloaded free functions, which will be closer to what you're envisioning, I think. I don't think that going all the way to creating an inheritance-based wrapper for each of the lattices would be worth it, though.

src/analysis/lattices/vector.h
kripken: Can they both inherit from something?

Or perhaps they can both simply call a standalone utility class that does these operations?

tlively: We could perhaps have a CRTP parent class, but I don't think it's worth it. These methods are simple enough that I think there's more value in being able to read their implementations directly without an extra layer of source indirection than there would be in deduplicating them.

kripken: What would be the harm in adding a utility function (not a CRTP parent class) "compareSingleLatticeElement" that contains the switch, and is called from those loops? Those functions would become much smaller and more readable, and we'd avoid the duplication.

tlively: The types are different, so at least that utility would have to be another template. It would also have to go in some new header file but would not be part of any public API, which is unfortunate. I don't anticipate the utility would be used anywhere except these two places, either.

I can do it if you feel strongly about it, but I don't think the extra complexity is worth the deduplication.

kripken: I don't feel that strongly about it, I suppose.

kripken (approved):

tlively:
### Merge activity

* **Oct 30, 10:45 PM**: @tlively started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/6058).

PR #6059, opened 2023-10-27
[analysis] Add a generic powerset lattice

tlively:
Add `Powerset2<Set>` that implements the lattice of sets concretely represented
as `Set` ordered by subset as well as the full lattice `FinitePowerset2<Set>`
that is additionally configurted with the universe of set elements so it can
produce a top element, which is the set of everything in the universe.

The "2" in the name serves to temporarily differentiate these new powerset
lattices from the existing powerset lattices, which are less flexible and will
be replaced with the new powerset lattices in a future PR.

Also add a new `BitSet` utility that implements a set interface matching e.g.
`std::set<size_t>` or `std::unordered_set<size_t>` in terms of a
`std::vector<bool>` bit vector. Use `BitSet` as the concrete set implementation
used to fuzz `Powerset2` and `FinitePowerset2`.


tlively:
* main
      * **PR #6058** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6058" target="_blank">Graphite</a> 
        * **PR #6059** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6059" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/6059?utm_source=stack-comment).

src/analysis/lattices/powerset2.h
```diff
@@ -0,0 +1,106 @@
...
+#include "../lattice.h"
+#include "support/bitset.h"
+
+#ifndef wasm_analysis_lattices_powerset2_h
+#define wasm_analysis_lattices_powerset2_h
+
+namespace wasm::analysis {
+
+// A powerset lattice whose elements are sets (represented concretely with type
+// `Set`) ordered by subset.
```
kripken, line 28: Perhaps we could call this a "Set Lattice", that is, a lattice over sets and ordered by the natural subset relation? "Powerset" is a specific mathematical operation to generate all subsets but this lattice here does not create a powerset nor does it require that the elements actually be a powerset IIANM.

If so then the one below could be "Powerset Lattice" (without "finite" in the name, which feels odd there to me).

src/analysis/lattices/powerset2.h
```diff
@@ -0,0 +1,106 @@
...
+    bool result = false;
+    for (const auto& val : joiner) {
+      result |= joinee.insert(val).second;
+    }
+    return result;
+  }
+};
+
+// A powerset lattice initialized with a list of all elements in the universe,
+// making it possible to produce a top elements that contains all of them.
```
kripken, line 67:
```suggestion
// making it possible to produce a top element that contains all of them.
```

src/support/bitset.h
```diff
@@ -0,0 +1,216 @@
...
+#define wasm_support_bitset_h
+
+#include <cassert>
+#include <cstddef>
+#include <iterator>
+#include <vector>
+
+namespace wasm {
+
+// Represent a set of integers in [0, N) with a bitvector of size N, where a 1
```
kripken, line 27:
```suggestion
// Represent a set of integers 0 <= k < N with a bitvector of size N, where a 1
```

kripken: Because the notation `[0, N)` is used in math to imply a half-open interval of real numbers. It's technically fine either way though.

src/support/bitset.h
```diff
@@ -0,0 +1,216 @@
...
+
+  class iterator {
+    const std::vector<bool>* parent = nullptr;
+    size_t index = 0;
+
+    iterator(const std::vector<bool>* parent) : parent(parent) {
+      const auto size = parent->size();
+      while (index < size && !(*parent)[index]) {
+        ++index;
+      }
```
kripken, line 52: Looks like this code advances to the first item in the set? A comment might help.

src/support/bitset.h
```diff
@@ -0,0 +1,216 @@
...
+
+  iterator begin() const noexcept { return iterator({this}); }
+  iterator end() const noexcept { return iterator({this, vec().size()}); }
+
+  bool empty() const noexcept { return begin() == end(); }
+
+  size_t size() const noexcept {
+    size_t result = 0;
+    for (auto it = begin(); it != end(); ++it, ++result) {
+    }
```
kripken, line 137: Wouldn't it be possibly simpler and faster to iterate on the underlying `std::vector<bool>`?
```cpp
auto* super = (std::vector<bool>*)this;
for (auto x : super) result += x;

kripken: (possibly simpler & faster since the current code will have nested loops)

kripken (commented):
Maybe pull out bitset to its own PR and add dedicated testing for it?

src/support/bitset.h
tlively: Oh, I didn't know the interval notation typically implied real numbers. Will update.

src/support/bitset.h
tlively: Yeah, I'll pull this out into a separate PR and add comments, as you suggest.

src/support/bitset.h
tlively: I would hope it would optimize to the same thing, but yes, I can rewrite the code to use the underlying vector directly.

src/support/bitset.h
kripken: Link for the curious: https://en.wikipedia.org/wiki/Interval_(mathematics)

src/analysis/lattices/powerset2.h
tlively: It's true we never materialize the full powerset, but this lattice does represent a powerset. The lattice itself is the powerset, since its elements are subsets of the set of all `T`.  For that reason, and to be consistent with the literature, we should continue calling this a powerset lattice.

kripken: Fair enough, but why is the latter called "finite" then, if both are finite?

tlively: The first one is not necessarily finite. There is no bound on the size of the elements it supports.

PR #6061, opened 2023-10-27
OnceReduction: Optimize bodies of trivial "once" functions

kripken:
In particular, if the body just calls another "once" function, then we can
skip the early-exit logic (see details in the comment in the new code).

Followup/replacement of #6055 

cc @gkdn

### DIFF
--- a/src/passes/OnceReduction.cpp
+++ b/src/passes/OnceReduction.cpp
@@ -344,38 +344,70 @@ struct Optimizer
       // This is a "once" function, that is, it has this shape:
       //
       //  function foo() {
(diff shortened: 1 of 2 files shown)


tlively (approved):

gkdn:
Logic LGTM, validating on my branch.

gkdn:
> Logic LGTM, validating on my branch.

👍

kripken:
Great, thanks for confirming @gkdn 

Meanwhile this has been fuzzed to almost 1M iterations, so I'm fairly confident in it (though this pattern is quite hard to fuzz).

PR #6065, opened 2023-10-31
[analysis][NFC] Refactor lattice unit tests

tlively:
Many of the lattice tests were essentially copy-pasted from one lattice to the
next because they all tested isomorphic subsets of the various lattices,
specifically in the shape of a diamond. Refactor the code so that all lattices
that have tests of this shape use the same utility test functions.

PR #6064, opened 2023-10-30
[analysis] Add a lattice for value types

tlively:
Add a lattice that is a thin wrapper around `wasm::Type` giving it the interface
of a lattice. As usual, `Type::unreachable` is the bottom element, but unlike in
the underlying API, we uniformly treat `Type::none` as the top type so that we
have a proper lattice.

PR #6062, opened 2023-10-28
[analysis] Add a tuple lattice

tlively:
This lattice combines any number of other lattices into a single lattice whose
elements are tuples of elements of the other lattices. This will be one of the
most important lattices in the analysis framework because it will be used to
combine information about different parts of the program, e.g. locals and the
value stack, into a single lattice.

PR #6067, opened 2023-10-31
[analysis] Add a "Shared" lattice to represent shared state

tlively:
The analysis framework stores a separate lattice element for each basic block
being analyzed to represent the program state at the beginning of the block.
However, in many analyses a significant portion of program state is not
flow-sensitive, so does not benefit from having a separate copy per block. For
example, an analysis might track constraints on the types of locals that do not
vary across blocks, so it really only needs a single copy of the constrains for
each local. It would be correct to simply duplicate the state across blocks
anyway, but it would not be efficient.

To make it possible to share a single copy of a lattice element across basic
blocks, introduce a `Shared<L>` lattice. Mathematically, this lattice represents
a single ascending chain in the underlying lattice and its elements are ordered
according to sequence numbers corresponding to positions in that chain.
Concretely, though, the `Shared<L>` lattice only ever materializes a single,
monotonically increasing element of `L` and all of its elements provide access
to that shared underlying element.

`Shared<L>` will let us get the benefits of having mutable shared state in the
concrete implementation of analyses without losing the benefits of keeping those
analyses expressible purely in terms of the monotone framework.

### DIFF
--- a/src/analysis/lattices/shared.h
+++ b/src/analysis/lattices/shared.h
@@ -17,7 +17,7 @@
 #ifndef wasm_analysis_lattices_shared_h
 #define wasm_analysis_lattices_shared_h
 
-#include <cstddef>
+#include <cstdint>
 
 #include "../lattice.h"
 #include "bool.h"
--- a/src/analysis/lattices/shared.h
+++ b/src/analysis/lattices/shared.h
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2023 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef wasm_analysis_lattices_shared_h
+#define wasm_analysis_lattices_shared_h
+
+#include <cstddef>
+
+#include "../lattice.h"
+#include "bool.h"
+
+namespace wasm::analysis {
+
+// A lattice whose elements are a single ascending chain in lattice `L`.
+// Internally, there is only ever a single monotonically increasing element of L
+// materialized. Dereferencing any element of the Shared lattice will produce
+// the current value of that single element of L, which is generally safe
+// because the current value always overapproximates the value at the time of
+// the Shared element's construction.
+template<Lattice L> struct Shared {
+  class Element {
+    const typename L::Element* val;
+    uint32_t seq = 0;
+
+    Element(const typename L::Element* val) : val(val) {}
+
+  public:
+    Element() = default;
+    Element(const Element&) = default;
+    Element(Element&&) = default;
+    Element& operator=(const Element&) = default;
+    Element& operator=(Element&&) = default;
+
+    // Only provide const references to the value to force all updates to go
+    // through our API.
+    const typename L::Element& operator*() const noexcept { return *val; }
+    const typename L::Element* operator->() const noexcept { return val; }
+
+    bool operator==(const Element& other) const noexcept {
+      assert(val == other.val);
+      return seq == other.seq;
+    }
+
+    bool operator!=(const Element& other) const noexcept {
+      return !(*this == other);
+    }
+
+    friend Shared;
+  };
+
+  L lattice;
+
+  // The current value that all elements point to and the current maximum
+  // sequence number. The sequence numbers monotonically increase along with
+  // `val` and serve to provide ordering between elements of this lattice.
+  mutable typename L::Element val;
+  mutable uint32_t seq = 0;
+
+  Shared(L&& l) : lattice(std::move(l)), val(lattice.getBottom()) {}
+
+  Element getBottom() const noexcept { return Element{&val}; }
+
+  LatticeComparison compare(const Element& a, const Element& b) const noexcept {
+    assert(a.val == b.val);
+    return a.seq < b.seq ? LESS : a.seq == b.seq ? EQUAL : GREATER;
+  }
+
+  bool join(Element& joinee, const Element& joiner) const noexcept {
+    assert(joinee.val == joiner.val);
+    if (joinee.seq < joiner.seq) {
+      joinee.seq = joiner.seq;
+      return true;
+    }
+    return false;
+  }
+
+  bool join(Element& joinee, const typename L::Element& joiner) const noexcept {
+    if (lattice.join(val, joiner)) {
+      // We have moved to the next value in our ascending chain. Assign it a new
+      // sequence number and update joinee with that sequence number.
+      joinee.seq = ++seq;
+      return true;
+    }
+    return false;
+  }
+};
+
+#if __cplusplus >= 202002L
+static_assert(Lattice<Shared<Bool>>);
+#endif // __cplusplus >= 202002L
+
+} // namespace wasm::analysis
+
+#endif // wasm_analysis_lattices_shared_h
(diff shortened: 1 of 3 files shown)


tlively:
* main
      * **PR #6058** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6058" target="_blank">Graphite</a> 
        * **PR #6062** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6062" target="_blank">Graphite</a> 
          * **PR #6064** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6064" target="_blank">Graphite</a> 
            * **PR #6065** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6065" target="_blank">Graphite</a> 
              * **PR #6067** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6067" target="_blank">Graphite</a>  👈
          * **PR #6059** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6059" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/6067?utm_source=stack-comment).

kripken:
> Mathematically, this lattice represents a single ascending chain in the underlying lattice

I followed up to here, and I got stuck.

Can you maybe give a concrete example of this?

As context, after reading the first paragraph I was expecting to see something like a "wrapper" for a lattice element that allows us to refer to an element without copying it. But I don't know why we need some new kind of structure here (the new chain) or even what that precisely means.

tlively:
Unfortunately we can't just have a `std::shared_ptr` to a mutable underlying lattice element and call it a day, even though that's kind of what we would like to have. This new lattice needs to _be_ a lattice so that it fits properly into the framework. This isn't just to make the compiler happy; it's important that `join` returns true to the analysis worklist algorithm when and only when a block needs to be re-analyzed because its input has changed since the last time it was analyzed.

To determine whether a block needs to be re-analyzed, we keep a sequence number alongside the pointer to the shared underlying element. When we propagate new results to the beginning of the block, we know the block needs to be re-analyzed if and only if the propagated sequence number is higher than the old sequence number. Every time the value of the shared underlying element increases, we increase the current sequence number so that any block that was analyzed based on a previous value will be re-analyzed.

The sequence numbers are an important part of getting the new lattice to do what we need it to do, but they're also mathematically interesting because each successive sequence number corresponds to a higher value that the shared underlying element has taken on at some point in time. The sequence of those values through time form an increasing chain in the underlying lattice and the sequence numbers are both timestamps and indices into that chain.

It's also nice to know that this `Shared` utility isn't just a hack to get shared, mutable state into the framework. It's actually a mathematically justifiable lattice. That gives me more confidence that it won't end up violating core assumptions and requiring hacky workarounds or special casing in the future.

src/analysis/lattices/shared.h
```diff
@@ -0,0 +1,108 @@
...
+#include "../lattice.h"
+#include "bool.h"
+
+namespace wasm::analysis {
+
+// A lattice whose elements are a single ascending chain in lattice `L`.
+// Internally, there is only ever a single monotonically increasing element of L
+// materialized. Dereferencing any element of the Shared lattice will produce
+// the current value of that single element of L, which is generally safe
+// because the current value always overapproximates the value at the time of
```
kripken, line 32: What does "overapproximate" mean here? A "higher" value in the lattice L?

src/analysis/lattices/shared.h
```diff
@@ -0,0 +1,108 @@
...
+// the Shared element's construction.
+template<Lattice L> struct Shared {
+  class Element {
+    const typename L::Element* val;
+    uint32_t seq = 0;
```
kripken, line 37: It would be great to see some of the comments in the PR itself added as comments here, to explain what `seq` is etc.

src/analysis/lattices/shared.h
```diff
@@ -0,0 +1,108 @@
...
+    friend Shared;
+  };
+
+  L lattice;
+
+  // The current value that all elements point to and the current maximum
+  // sequence number. The sequence numbers monotonically increase along with
+  // `val` and serve to provide ordering between elements of this lattice.
+  mutable typename L::Element val;
+  mutable uint32_t seq = 0;
```
kripken, line 71: This will be the 2nd and 3rd time we use `mutable` in this entire codebase... can we avoid that?

If we can't please add a comment explaining why this dangerous pattern is necessary here.

src/analysis/lattices/shared.h
kripken, line 37:
```suggestion
    Index seq = 0;
```
Index seems appropriate as this number is tied to the number of basic blocks (and iterations on them), which is tied to maximum sizes in wasm files. That is, if wasm supports more then 32 bits here, we'd want this type to change, which `Index` will do for us.

test/gtest/lattices.cpp
```diff
@@ -546,3 +547,79 @@ TEST(ValTypeLattice, Meet) {
...
+  analysis::Shared<analysis::UInt32> shared{analysis::UInt32{}};
+  EXPECT_EQ(*shared.getBottom(), 0u);
+}
+
+TEST(SharedLattice, Compare) {
+  analysis::Shared<analysis::UInt32> shared{analysis::UInt32{}};
+
+  auto zero = shared.getBottom();
+
+  auto one = zero;
```
kripken, line 561: I am having trouble reading a line that says literally `1 == 0`...

(And later down I see `2 == 1`. By induction, all natural numbers are apparently `0`... :thinking: :laughing: )

How about `a, b, c`?

test/gtest/lattices.cpp
```diff
@@ -546,3 +547,79 @@ TEST(ValTypeLattice, Meet) {
...
+  shared.join(two, 2);
+
+  EXPECT_EQ(shared.compare(zero, zero), analysis::EQUAL);
+  EXPECT_EQ(shared.compare(zero, one), analysis::LESS);
+  EXPECT_EQ(shared.compare(zero, uno), analysis::LESS);
+  EXPECT_EQ(shared.compare(zero, two), analysis::LESS);
+
+  EXPECT_EQ(shared.compare(one, zero), analysis::GREATER);
+  EXPECT_EQ(shared.compare(one, one), analysis::EQUAL);
+  EXPECT_EQ(shared.compare(one, uno), analysis::EQUAL);
```
kripken, line 577: To make sure I follow, `uno` had an extra `join` done on it, but it was of the same value as we had before, so the sequence number was not increased, hence it is equal to `one`?

kripken (commented):
Thanks, the idea now makes sense to me, and lgtm % comments

src/analysis/lattices/shared.h
tlively: Yep 👍

src/analysis/lattices/shared.h
tlively: This number is not tied to the number of basic blocks. It is the number of times the shared value is increased, which can be dramatically higher than the number of basic blocks. Since analysis continues to a fixed point, it can visit basic blocks an unbounded number of times and each visit to a basic block may potentially increase the value multiple times.

Since this isn't actually an index into anything that is ever materialized, I would prefer to keep it a uint32_t for now, or possibly make it a uint64_t right now.

test/gtest/lattices.cpp
tlively: The only way to increase a shared value is via a `join`, so to create our `one` value we have to start from zero and add one via the join on the next line. I'll clarify this in the code.

test/gtest/lattices.cpp
tlively: That's right 👍

src/analysis/lattices/shared.h
kripken: sgtm. How about using a name for it for convenience later if we need to change it, `using Seq = uint32_t;` ?

kripken (approved):

tlively:
### Merge activity

* **Nov 2, 2:30 PM**: @tlively started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/6067).

PR #6068, opened 2023-10-31
[NFC] Improve OnceReduction comment

kripken:
Followup to #6061 in which we only optimized the case of a "once" function immediately
calling another and doing nothing else. It is ok to do more things afterwards, so long as
we do nothing else before, by the same logic as already mentioned in the pass (but
clarified as to the meaning of "called" - all we need is for the function to have been
"entered" in the proof there, IIANM).

@gkdn am I missing something?

### DIFF
--- a/src/passes/OnceReduction.cpp
+++ b/src/passes/OnceReduction.cpp
@@ -486,14 +486,16 @@ struct OnceReduction : public Pass {
         ExpressionManipulator::nop(body);
         continue;
       }
-      if (list.size() != 3) {
-        // Something non-trivial; too many items for us to consider.
-        continue;
-      }
-      auto* payload = list[2];
-      if (auto* call = payload->dynCast<Call>()) {
+      // The early-exit logic is the first item, followed by the global setting,
+      // and we've ruled out the case of there being nothing else after those,
+      // so there are at least 3 items.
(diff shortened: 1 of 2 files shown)


gkdn:
Are you proposing; "we can still remove the condition, if the once function does more than calling another once function"?

kripken:
Yes, exactly. Concretely:

```wat
(func $once
  (if
    (global.get $once)
    (return)
  )
  (global.set $once (i32.const 1))
  ;; Call another once function. This lets us remove the if and global.set.
  (call $once.1)
  ;; Do other stuff. This does not stop us, since this PR.
  (call $anything)
)
```

gkdn:
if `anything` calls back `$once`, we would no longer have the guard so `anything` would be called again.

kripken:
Oh right... how silly of me!

I'll update this PR to fix the proof text and remove the change.

kripken:
This PR now only adds a comment to explain why we can only optimize in the case that we do.

tlively (approved):

gkdn (approved):

kripken:
This weird ASan error seems like a random flake... it happens once every few PRs now. Very strange. I wish we had a way to ssh into the CI here to investigate...

PR #6072, opened 2023-11-01
Fuzzer: Better handling of globals from initial content

kripken:
Previously the fuzzer never added gets or sets of globals from initial content. That was
an oversight, I'm pretty sure - it's just that the code that sets up the lists from which we
pick globals for gets and sets was in another place. That is, any globals in the initial
content file were never used in new random code the fuzzer generates (only new
globals the fuzzer generated were used there).

This PR allows us to use those globals, but also ignores them with some probability,
to avoid breaking patterns like "once" globals (that we want to only be used from
initial content, at least much of the time).

Also simplify the code here: we don't need `isInvalidGlobal` just to handle the hang
limit global, which is already handled by not being added to the lists we pick names
from anyhow.


tlively (approved):
Nice 👍

PR #6077, opened 2023-11-02
[analysis] Make it easier to implement a transfer function

tlively:
Combine the `transfer` and `getDependents` methods of a transfer function so
that a transfer function only has to implement `transfer`, which now returns a
range of basic blocks that may need to be re-analyzed.

To make it easier to implement the returned basic block range, change the
requirement so that it provides iterators to `const BasicBlock*` rather than
`BasicBlock`. This allows us to entirely remove cfg-impl.h.


tlively:
* main
  * **PR #6077** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6077" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/6077?utm_source=stack-comment).

kripken (approved):

Issue #5951, opened 2023-09-18
Parse Exception with Block Type But WABT Successfully Parses It


mobsceneZ:
Hello there, I recently came across a Wasm module which cannot be parsed by Binaryen: 

```wasm
$ wabt/bin/wasm2wat test.wasm
(module
  (type (;0;) (func (result i32)))
  (type (;1;) (func (param i32) (result i32)))
  (func (;0;) (type 0) (result i32)
    i32.const 1
    block (param i32) (result i32)  ;; label = @1
      i32.const 2
      i32.add
    end)
  (export "param" (func 0)))
```

When using wasm-dis to parse above wasm module, it encountered following errors:

```shell
$ wasm-dis test.wasm
[parse exception: Block requires more values than are available (at 0:48)]
Fatal: error parsing wasm (try --debug for more info)
```

However, when using wasm2wat provided by WABT, it parsed successfully and wasm-interp runs `param` function without error:

```shell
$ wasm-interp test.wasm -r "param"
param() => i32:3
```

The binaryen version I tested is 16a59938563c93d8459bf36679c83497aeba7cc7, which is currently the newest commit version.

zapashcanon:
FYI this is accepted by the reference interpreter and by Owi.

kripken:
The issue is probably the same as in #5950 - Binaryen does not support multivalue input params to control flow structures (here, `block (param i32)`). Unfortunately it seems we don't have an error on this atm.

mobsceneZ:
> The issue is probably the same as in #5950 - Binaryen does not support multivalue input params to control flow structures (here, `block (param i32)`). Unfortunately it seems we don't have an error on this atm.

Thanks for your kind response, kripken. Sorry to bother but I have two more questions that I'm not sure:
1. Does Binaryen not support table.init, elem.drop, table.copy, table.fill yet? Because I witnessed similar parse errors when dealing with these instructions.
2. The following wat file works just fine with WABT's `wat2wasm` and `wasm2wat`, however, when I convert it into wasm binary form using `wat2wasm` and later use `wasm-dis` to parse it, error occurs: 
```wasm
$ cat test.wat
(module
        (type (;0;) (func))
        (func (;0;) (type 0))
        (func (;1;) (type 0))
        (table (;0;) 10 funcref)
        (elem (;53;) declare func)
        (elem (;54;) declare funcref (ref.func 0) (ref.func 0) (ref.null func) (ref.func 1))
        (elem (;55;) declare func)
[...]

PR #6078, opened 2023-11-02
Implement table.copy

kripken:
With this, the cool Scheme game in https://davexunit.itch.io/strigoform can be optimized by wasm-opt. We reduce code size by 20% and improve the quality pretty substantially (e.g. we remove most of the casts).

Helps #5951


src/passes/Unsubtyping.cpp
```diff
@@ -446,6 +446,10 @@ struct Unsubtyping
   void visitTableFill(TableFill* curr) {
     noteSubtype(curr->value->type, getModule()->getTable(curr->table)->type);
   }
+  void visitTableCopy(TableCopy* curr) {
```
tlively, line 449: It would be nice to add a test case for this in test/lit/passes/unsubtyping.wast.

src/wasm-interpreter.h
```diff
@@ -3027,6 +3032,57 @@ class ModuleRunnerBase : public ExpressionRunner<SubType> {
...
+    Address sizeVal(size.getSingleValue().getUnsigned());
+
+    auto destInfo = getTableInterfaceInfo(curr->destTable);
+    auto sourceInfo = getTableInterfaceInfo(curr->sourceTable);
+    auto destTableSize = destInfo.interface->tableSize(destInfo.name);
+    auto sourceTableSize = sourceInfo.interface->tableSize(sourceInfo.name);
+    if (sourceVal + sizeVal > sourceTableSize ||
+        destVal + sizeVal > destTableSize ||
+        // FIXME: better/cheaper way to detect wrapping?
+        sourceVal + sizeVal < sourceVal || sourceVal + sizeVal < sizeVal ||
```
tlively, line 3063: I don't think you need to check for wrapping against both `sourceVal` and `sizeVal`. Even the most wrapping possible cannot wrap around to greater than the minimum of the two, so just one of them will serve to detect the wrapping.

src/wasm/wasm-s-parser.cpp
```diff
@@ -2706,6 +2706,24 @@ Expression* SExpressionWasmBuilder::makeTableFill(Element& s) {
   return Builder(wasm).makeTableFill(tableName, dest, value, size);
 }
 
+Expression* SExpressionWasmBuilder::makeTableCopy(Element& s) {
+  auto destTableName = s[1]->str();
```
tlively, line 2710: Should we support raw indices as well, or do we not support those in other places? The new wat parser will support them in either case (eventually).

test/spec/table_copy.wast
tlively, line 1: I assume this is copied from upstream? I saw the TODO at the end, but did anything here require modifications? Are the failures mentioned in the TODO due to the fact that we re-evaluate table initializers whenever the table is accessed (IIRC)?

tlively (approved):

src/wasm-interpreter.h
kripken: This is copied from the parallel logic in MemoryCopy. I agree that it looks unoptimal. I think the TODO covers improving them.

src/wasm/wasm-s-parser.cpp
kripken: I don't think we do in other places, as this is based on TableFill right above us. That actually might explain why some spec tests didn't pass...

test/spec/table_copy.wast
kripken: No modifications. Just the TODO at the end, for tests that don't pass yet. I didn't investigate those in detail, but at least one factor was a failure to parse, which might be what you mentioned above.

tlively (approved):

PR #6079, opened 2023-11-03
Update CFGWalker to generate consolidated exit blocks

tlively:
Previously CFGWalker designated a particular block as the "exit" block, but it
was just the block that happened to appear at the end of the function that
returned values by implicitly flowing them out. That exit block was not tied in
any way to other blocks that might end in returns, so analyses that needed to
perform some action at the end of the function would have had to perform that
action at the end of the designated exit block but also separately at any return
instruction.

Update CFGWalker to make the exit block a synthetic empty block that is a
successor of all other blocks that implicitly or explicitly return from the
function in case there are multiple such blocks, or to make the exit block the
single returning block if there is only one. This means that analyses will only
perform their end-of-function actions at the end of the exit block rather than
additionally at every return instruction.

### DIFF
--- a/src/cfg/cfg-traversal.h
+++ b/src/cfg/cfg-traversal.h
@@ -53,7 +53,7 @@ struct CFGWalker : public PostWalker<SubType, VisitorType> {
   // The exit block for the function: either the single block that returns or
   // flows values out of the function, or an empty synthetic block that is a
   // successor of all such blocks. This block may not exist if a function
-  // traps or infinitely loops and therefore never exits.
+  // traps, infinitely loops, throws, or otherwise never exits normally.
   //
   // Analyses that care about reaching the end of the function can just look at
   // this block instead of all the individual returns.
--- a/src/cfg/cfg-traversal.h
+++ b/src/cfg/cfg-traversal.h
@@ -46,22 +46,18 @@ struct CFGWalker : public PostWalker<SubType, VisitorType> {
     std::vector<BasicBlock*> out, in;
   };
 
-  // The entry block at the function's start. This always exists.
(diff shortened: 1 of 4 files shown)


tlively:
* main
  * **PR #6079** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6079" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/6079?utm_source=stack-comment).

src/cfg/cfg-traversal.h
```diff
@@ -46,22 +46,18 @@ struct CFGWalker : public PostWalker<SubType, VisitorType> {
...
-  // a return or an unreachable.
-  BasicBlock* exit;
+  // The entry block at the function's start. This always exists, although it
+  // might be empty if the function is empty.
+  BasicBlock* entry = nullptr;
+
+  // The exit block for the function: either the single block that returns or
+  // flows values out of the function, or an empty synthetic block that is a
+  // successor of all such blocks. This block may not exist if a function
+  // traps or infinitely loops and therefore never exits.
```
kripken, line 56: Might be worth mentioning throwing here, which is not considered an "exit". That is, the exit is the "return to caller, possibly with a value" block, not a general "exit from function somehow" block (which would have included exceptions).

kripken (approved):
I'm pretty sure this will not regress performance, but it might be worth doing some measurements just to be safe.

tlively:
I confirmed that this does not change performance on --merge-locals, which uses local graph.

PR #6083, opened 2023-11-05
Typed continuations: resume instructions

frank-emrich:

This particular PR adds support for the `resume` instruction. The most notable missing feature is validation, which is not implemented, yet.


frank-emrich:
There are currently a few `FIXME`s which are basically questions of mine, one of them currently causing the new lit test to fail.

src/ir/branch-utils.h
```diff
@@ -78,6 +78,24 @@ void operateOnScopeNameUsesAndSentTypes(Expression* expr, T func) {
...
+      // could get it from the Module (to look up the tag), but this would
+      // require additional plumbing at all the use sites of this function, such
+      // as users of BranchSeeker, which is in turn used by functions such as
+      // Block::finalize(). Plumbing the module through to that point would
+      // require chancing the re-finalisation code, for example.
+      //
+      // We could also store this information in the Resume node itself, but
+      // that would mean storing the signature of all the Tags that we have a
+      // handle clause for.
+      func(name, Type::none);
```
frank-emrich, line 98: The issue described here is what's causing the lit test suite to fail:
Enabling `BINARYEN_PASS_DEBUG` means that `validateBinaryenIR` eventually re-finalises each block by calling `Block::finalize()`. This can cause situations where the types collected by `BranchSeeker` fully determine the new type given to the block, ignoring the type annotation on it. This means that the code snippet here, where `Type::none` is passed as a placeholder for the real type causes problems, because `Block::finalize()` will obtain this from `BranchSeeker` and uses it as the overall type of the block. This then makes `validateBinaryenIR` fail.

This failure seems to be a relatively recent change, since the test suite was passing a few weeks ago when I did the majority of the work on this PR... But I guess it just shows that my "solution" of passing `Type::none` here was never quite right to begin with.

tlively: cc @aheejin, I guess the new EH design will also run into the same problem here: the types of values carried on the branches from the `catch_table` instruction will depend on the tags on that instruction, so we will need the module here to look up the types of those tags.

@frank-emrich, I don't think we have a choice but to plumb the module through to here sooner or later, and probably sooner, since I can't think of a good workaround 🤔

src/ir/branch-utils.h
```diff
@@ -97,6 +115,10 @@ void operateOnScopeNameUsesAndSentValues(Expression* expr, T func) {
       func(name, sw->value);
     } else if (auto* br = expr->dynCast<BrOn>()) {
       func(name, br->ref);
+    } else if (auto* res = expr->dynCast<Resume>()) {
+      // FIXME(frank-emrich) We can't really determine the values being sent
+      // here because they come from suspend instructions elsewhere. Is that a
+      // problem?
```
tlively, line 121: New EH will have the same problem. I would just pass `nullptr` for now and see what breaks. Hopefully nothing!

src/ir/effects.h
```diff
@@ -958,6 +958,20 @@ class EffectAnalyzer {
...
     }
+
+    void visitResume(Resume* curr) {
+      // This acts as a kitchen sink effect.
+      parent.calls = true;
+
+      // FIXME(frank-emrich) We should probably set parent.trap or
+      // parent.implicitTrap, because the resume instruction itself may trap:
+      // The continuation reference is nullable, and we trap if it is indeed
+      // null.
```
tlively, line 969: `implicitTrap` is the right thing to set here.

src/ir/branch-utils.h
frank-emrich: I've worked on this for a bit, which mostly involved changing all use sites of `BranchSeeker`, and have a question:

Currently, `BranchSeeker` is mostly used  by calling its `has` and `count` functions. These don't care about the types sent to the branches types at all. In fact, there are few uses of `BranchSeeker` that *do* inspect the `types` field afterwards. And we only need access to the `Module` in order to correctly calculate the types sent to the branch. That means we would have to change `Branchseeker::count` and `BranchSeeker::has` to take a `Module` parameter, even though the `types` field is ignored.

This suggests that it may be useful to change the existing `BranchSeeker` such that it continues offering `has` and `count` with the existing signatures but remove the `types` field. This could be implemented entirely using `operateOnScopeNameUses`, no type (or module) information needed.

We then define a new `BranchTypeSeeker` that behaves more like the old  `BranchSeeker`, which uses `operateOnScopeNameUsesAndSentTypes`, does accumulate the types sent to the branch, and needs access to the `Module` to do so to implement typed continuations and the new style EH.

 This question came up when changing the signature of the various `finalize` methods of `Block`: The existing `finalize()` must definitely take a `Module` in the future, because it uses `BranchSeeker` to actually inspect the types sent to branches.
However, `Block::finalize(Type type_)` and`Block::finalize(Type type_, Breakability breakability)` only use `BranchUtils::BranchSeeker::has` and don't care about the types sent to the block. Thus, if we go through with the splitting of `BranchSeeker` mentioned above, we could avoid also needing to add a `Module` parameter to these two.

I've mostly gone through with changing the signatures of all three `Block::finalize` functions now, so I don't care much which option we take here, but it felt a bit unfortunate to change all these use sites of `Branchseeker::count` and `BranchSeeker::has`, even though they don't actually care about the extra typing information.

tlively: This refactoring sounds good to me.

One option might be for the `finalize` methods that would need to take a Module to instead conservatively keep the type unchanged, then have special logic in the `ReFinalize` that takes the module into account to do better refinalization.

frank-emrich: I have created a separate PR with the Module parameter plumbing: #6096. I will rebase this PR once the other one has landed.

frank-emrich: After the discussion in #6096 I will now implement storing the sent types directly in `Resume` nodes, instead of using the `Module` here to determine them.

The most straightforward implementation is an `ArenaVector<Type>` whose i-th entry contains the type sent to the i-th block (i.e, `curr->handlerBlocks[i]`). Let's assume this vector is called `sentTypes`.


There is one different issue here that affects the overall implementation of `operateOnScopeNameUsesAndSentTypes`: Since it is implemented using `operateOnScopeNameUses`, a `Resume` instruction with n handler clauses results in n calls to the lambda defined in `operateOnScopeNameUsesAndSentTypes`, with the name of the corresponding handler block as the `name` argument each time. However, in the lambda we then have would run code like this:

```
  ...
  } else if (auto* res = expr->dynCast<Resume>()) {
    for (usize i = 0; < res->handlerBlocks.size(); i++) {
      auto handlerBlock = res->handlerBlocks[i];
      if (handlerBlock == name) {
        func(name, curr->sentTypes[i]);
      }
    }
  } else { ...
```

In total, this is quadratic in the number of handler clauses.

Which of the following solutions would you prefer:
1. `operateOnScopeNameUses` takes a lambda whose parameter is a `vector` of names. We would effectively call the lambda "in bulk" for those node types that have more than one scope name use. But this would require creating singleton vector for all the other node types, so this doesn't seem efficient.
2. We could pass some kind of index to the lambda, indicating that the currently passed `Name` is the i-th block name used by the particular node. But that doesn't seem great either, the index would be meaningless for all node types other than `Resume`.
3. `sentTypes` could be a map from block names to types instead of just a vector of types. That seems relatively clean, but uses more space, and it would need to be integrated into the logic that updates tag names in case a naming section is encountered in a binary.
4. We may implement `operateOnScopeNameUsesAndSentValues` without using `operateOnScopeNameUses` at all: Instead, it could just be a visitor on Expressions. In the case for `Resume` nodes, we would then process all the blocks at once and invoke `func` for each handler block.

I'm personally in favor of the last option: After all, the lambda passed to `operateOnScopeNameUses` in `operateOnScopeNameUsesAndSentValues` *already* disambiguates the different node types, `operateOnScopeNameUses` is basically just used to pre-filter the nodes and extract the name of the used block.

Note that this entire discussion equally applies to `Try` nodes once the updated EH proposal is implemented.

kripken: 4 makes sense to me as well.

aheejin: How many handler clauses do you expect to have with `resume`? With `try_table`, at least LLVM at the moment will generate instructions with only one clause (it is either for the C++ tag/label or the `catch_all` label. We don't generate both at the same time. But even if we do it's gonna be only two), so we are not really gonna be affected by the quadratic loop. If what you would have in average is less than 5, I'm not sure if this is something we should worry about?

frank-emrich:
I've now addressed the first round of comments.

One thing that I've noticed is that `updated_lit_checks.py` cannot distinguish between tag definitions `(tag $mytag ...)` in the tags section of a module and handler clauses appearing in `resume` instructions: `(resume $ct (tag $mytag $myblock) ...)`. As result, the script messes up the CHECK clauses. As temporary workaround, I've removed the "Assertions have been generated by update_lit_checks.py ..." header line from my new test file and removed the erroneously added CHECK lines manually.

I've had a quick look at the logic in `updated_lit_checks.py`. In general, we could distinguish the two kinds of `(tag ...)` expressions based on whether or not something like a `(resume` or `(func` appears beforehand. Anything indicating that we're not in the tag section any more should suffice. I've tried doing this with a negative lookahead before `tag`, but Python only accepts them if they have a constant length, which doesn't quite work for what I tried to do.

src/ir/branch-utils.h
frank-emrich: I've now re-implemented `operateOnScopeNameUsesAndSentTypes` using a `Visitor`.

@aheejin There are some compilation schemes where you may create `resume` instructions with `tag` clauses that handle more or less every single tag defined. However, even then I'd surprised to see more than a dozen or two. However, given that there are source languages with effect handlers that would like to target our instructions directly, it's basically up to what people do in those source languages. Having the quadratic behavior here is probably fine unless someone translates some crazy generated code using effect handlers into wasm.

But given that the solution using a visitor was relatively simple to implement, I'd say let's just err on the side of caution.

tlively:
I think the `update_lit_checks.py` issue will be solved if you put the `(tag ...)` clause on the same line as the `(resume ...` rather than on its own line.

frank-emrich:
> I think the `update_lit_checks.py` issue will be solved if you put the `(tag ...)` clause on the same line as the `(resume ...` rather than on its own line.

That doesn't seem to have worked if I just move it on its own line in the input function. Did you mean changing the printer such that it goes on its own line in the output? That seems tricky, because there may be many such clauses, so at some point they have to go on their own lines, rather than unconditionally keeping them all on the same line as the `resume`.

src/ir/branch-utils.h
frank-emrich: Thanks, that did work!

tlively:
Yes, I think you'd have to change the printer to put the `tag` clauses on the same line as `(resume` as well. Even though this does not visually scale to large numbers of tag clauses, it is at least consistent with what we do with e.g. function parameters and results, `br_table` labels, etc. In general our printer never tries to be clever and break up long lines.

frank-emrich:
I've changed the printing now so that all the `(tag ...)` clauses appear on the same line as the `resume`. That did indeed fix the issue with `update_lit_checks.py`.

In general, things should be ready now for you to have a look at.

tlively (commented):
Generally looks good, just a few final comments.

src/wasm/wasm.cpp
```diff
@@ -1318,6 +1319,71 @@ void StringSliceIter::finalize() {
...
+  if (this->sentTypes.size() != this->handlerBlocks.size()) {
+    Fatal() << "Resume node was not finalized";
+  }
+}
+
+void Resume::finalize(Module* wasm) {
+  if (!(this->contType.isContinuation() &&
+        this->contType.getContinuation().type.isSignature())) {
+    Fatal() << "ill-formed Resume expression";
+  }
```
aheejin, line 1347: Can we move this (validation in `Resume::finalize()` and `Resume::finalize(Module* wasm)` to `wasm-validator.cpp`? Also if we do
```cpp
void finalize(Module* wasm = nullptr);
```
in `wasm.h`, we wouldn't need two methods.

test/lit/typed_continuations_resume.wast
```diff
@@ -0,0 +1,94 @@
...
+
+ ;; CHECK-BINARY:      (type $3 (func (param (ref $ct)) (result i32)))
+
+ ;; CHECK-BINARY:      (tag $t (result i32))
+ ;; CHECK-TEXT:      (type $2 (func (result i32)))
+
+ ;; CHECK-TEXT:      (type $3 (func (param (ref $ct)) (result i32)))
+
+ ;; CHECK-TEXT:      (tag $t (result i32))
+ (tag $t (result i32))
```
aheejin, line 29: Can we have a test with `resume` using a tag with non-empty `params` as well?

src/wasm/wasm.cpp
```diff
@@ -1318,6 +1319,71 @@ void StringSliceIter::finalize() {
...
+    if (tgps.size() > 0) {
+      TypeList sentValueTypes;
+      sentValueTypes.reserve(tgps.size() + 1);
+
+      sentValueTypes.insert(sentValueTypes.begin(), tgps.begin(), tgps.end());
+      sentValueTypes.push_back(ctPrimeRef);
+      this->sentTypes[i] = Type(sentValueTypes);
+    } else {
+      this->sentTypes[i] = ctPrimeRef;
+    }
```
aheejin, line 1383: Do we need these separate cases? The body within `if` would work fine even if `tpgs` is empty.

aheejin (commented):
Looking good; mostly nits. I appreciate your discussions on the tag types in https://github.com/WebAssembly/binaryen/pull/6083#issuecomment-1854634679 and #6096, which I referred to before I submitted #6181.

src/wasm/wasm.cpp
frank-emrich: Good point! The `else` branch is probably the more common case, where the handler block only receives the continuation, but no additional payloads. So separating the two cases just boils down to avoiding the allocation for a singleton `std::vector<Type>` (namely, `sentValueTypes`). Would you prefer avoiding the separate `else` case?

src/wasm/wasm-binary.cpp
```diff
@@ -4040,6 +4040,12 @@ BinaryConsts::ASTNodes WasmBinaryReader::readExpression(Expression*& curr) {
       visitCallRef(call);
       break;
     }
+    case BinaryConsts::Resume: {
+      auto resume = allocator.alloc<Resume>();
+      curr = resume;
+      visitResume(resume);
```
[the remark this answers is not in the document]
frank-emrich: Done

src/wasm-delegations-fields.def
```diff
@@ -899,6 +899,17 @@ switch (DELEGATE_ID) {
     DELEGATE_END(StringSliceIter);
     break;
   }
+
+  case Expression::Id::ResumeId: {
+    DELEGATE_START(Resume);
```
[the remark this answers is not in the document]
frank-emrich: Done

src/passes/Print.cpp
```diff
@@ -2728,6 +2730,36 @@ void PrintSExpression::visitTry(Try* curr) {
...
+  // We deliberate keep all (tag ...) clauses on the same line as the resume
+  // itself to work around a quirk in update_lit_checks.py
+  for (size_t i = 0; i < curr->handlerTags.size(); i++) {
+    o << maybeSpace << '(';
+    printMedium(o, "tag ");
+    printName(curr->handlerTags[i], o);
+    o << " ";
+    printName(curr->handlerBlocks[i], o);
+    o << ")";
+  }
```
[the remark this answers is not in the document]
frank-emrich: Done

src/wasm-interpreter.h
```diff
@@ -3913,6 +3914,7 @@ class ModuleRunnerBase : public ExpressionRunner<SubType> {
     multiValues.pop_back();
     return ret;
   }
+  Flow visitResume(Resume* curr) { return Flow(NONCONSTANT_FLOW); }
```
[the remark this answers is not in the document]
frank-emrich: Done

src/wasm.h
```diff
@@ -42,6 +42,8 @@
 
 namespace wasm {
 
+class Module;
+
```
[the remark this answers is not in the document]
frank-emrich: Yes, exactly

src/ir/branch-utils.h
```diff
@@ -69,19 +69,35 @@ template<typename T> void operateOnScopeNameUses(Expression* expr, T func) {
...
-    // if-else chain. This will need to be updated with new br variants.
-    if (auto* br = expr->dynCast<Break>()) {
-      func(name, br->value ? br->value->type : Type::none);
-    } else if (auto* sw = expr->dynCast<Switch>()) {
-      func(name, sw->value ? sw->value->type : Type::none);
-    } else if (auto* br = expr->dynCast<BrOn>()) {
-      func(name, br->getSentType());
-    } else {
-      assert(expr->is<Try>() || expr->is<Rethrow>()); // delegate or rethrow
+  struct SentTypesVisitor : public Visitor<SentTypesVisitor> {
```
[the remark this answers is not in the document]
frank-emrich: Done

src/ir/subtype-exprs.h
```diff
@@ -331,6 +331,8 @@ struct SubtypingDiscoverer : public OverriddenVisitor<SubType> {
   void visitStringIterMove(StringIterMove* curr) {}
   void visitStringSliceWTF(StringSliceWTF* curr) {}
   void visitStringSliceIter(StringSliceIter* curr) {}
+
+  void visitResume(Resume* curr) { WASM_UNREACHABLE("not implemented"); }
```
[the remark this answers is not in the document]
frank-emrich: I think It's slightly more complicated than that: We may also need to report subtyping relations induced by values being sent to handler blocks: If a `resume` instruction has a `(tag $t $h)` clause, then the _n_ param types of `$t` must match the first _n_ result types of block `$h`. I guess this means that we would have to record these relationships here and skipped it so far.

src/ir/effects.h
```diff
@@ -963,6 +963,19 @@ class EffectAnalyzer {
...
+
+    void visitResume(Resume* curr) {
+      // This acts as a kitchen sink effect.
+      parent.calls = true;
+
+      // resume instructions accept nullable continuation references and trap
+      // on null.
+      parent.implicitTrap = true;
+
+      if (parent.features.hasExceptionHandling() && parent.tryDepth == 0) {
```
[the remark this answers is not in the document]
frank-emrich: This was just inspired by the following happening at the end of `visitCall` [here](https://github.com/WebAssembly/binaryen/blob/82057de7666b84424f8d2bd90e941762e4e14f0e/src/ir/effects.h#L494) :
```
// When EH is enabled, any call can throw.
if (parent.features.hasExceptionHandling() && parent.tryDepth == 0) {
    parent.throws_ = true;
}
```

I'm not sure if it's really necessary for `Resume`. @aheejin would you mind having a look?

src/wasm/wasm-s-parser.cpp
```diff
@@ -2921,6 +2921,38 @@ Expression* SExpressionWasmBuilder::makeCallRef(Element& s, bool isReturn) {
...
 
+Expression* SExpressionWasmBuilder::makeResume(Element& s) {
+  auto ret = allocator.alloc<Resume>();
+
+  ret->contType = parseHeapType(*s[1]);
+  if (!ret->contType.isContinuation()) {
+    throw ParseException("expected continuation type", s[1]->line, s[1]->col);
+  }
+
+  Index i = 2;
```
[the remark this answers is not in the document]
frank-emrich: Good catch! Fortunately, this is just a bug in the explainer document. The heap type was not part of the instruction in earlier revisions of the proposal, but it is now (unconditionally) required. The description of `resume` instructions in the [overview document](https://github.com/wasmfx/specfx/blob/main/proposals/continuations/Overview.md#instructions) has the correct syntax.

src/wasm/wasm-s-parser.cpp
```diff
@@ -2921,6 +2921,38 @@ Expression* SExpressionWasmBuilder::makeCallRef(Element& s, bool isReturn) {
...
+  while (i < s.size() && elementStartsWith(*s[i], "tag")) {
+    Element& inner = *s[i++];
+    if (inner.size() < 3) {
+      throw ParseException("invalid tag block", inner.line, inner.col);
+    }
+    Name tag = getTagName(*inner[1]);
+    if (!wasm.getTagOrNull(tag)) {
+      throw ParseException("bad tag name", inner[1]->line, inner[1]->col);
+    }
+    ret->handlerTags.push_back(getTagName(*inner[1]));
```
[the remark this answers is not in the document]
frank-emrich: Done

src/wasm/wasm-binary.cpp
```diff
@@ -7701,6 +7707,52 @@ void WasmBinaryReader::visitRefAs(RefAs* curr, uint8_t code) {
...
+    auto handler = getBreakTarget(handlerIndex).name;
+
+    curr->handlerTags[i] = tag;
+    curr->handlerBlocks[i] = handler;
+
+    // We don't know the final name yet
+    tagRefs[tagIndex].push_back(&curr->handlerTags[i]);
+  }
+
+  curr->cont = popNonVoidExpression();
```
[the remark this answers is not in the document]
frank-emrich: Please note that neither the operands nor the continuation are actually part of the resume instruction, it is only the folded textual representation (and binaryen's AST-like representation of instructions) that makes it look that way.
The type of a `resume` instruction (documented [here](https://github.com/wasmfx/specfx/blob/main/proposals/continuations/Overview.md#instructions)) is of the form `[t1* (ref null? $ct)] -> [t2*]`, where `t1*` are the operands. That's why the continuation should be on top of the stack at this point.

Only the continuation type and list of handler clauses are actually part of the instruction.

src/wasm/wasm-binary.cpp
```diff
@@ -7701,6 +7707,52 @@ void WasmBinaryReader::visitRefAs(RefAs* curr, uint8_t code) {
...
+  BYN_TRACE("zz node: Resume\n");
+
+  auto contTypeIndex = getU32LEB();
+  curr->contType = getTypeByIndex(contTypeIndex);
+  if (!curr->contType.isContinuation()) {
+    throwError("non-continuation type in resume instruction " +
+               curr->contType.toString());
+  }
+
+  auto handlerNum = getU32LEB();
```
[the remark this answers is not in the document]
frank-emrich: Done

src/wasm/wasm.cpp
```diff
@@ -1318,6 +1319,71 @@ void StringSliceIter::finalize() {
   }
 }
 
+ArenaVector<Type>& Resume::getSentTypes() {
+  if (this->sentTypes.size() != this->handlerBlocks.size()) {
+    Fatal() << "Types sent to blocks have not been determined, yet.";
+  }
+  return this->sentTypes;
+}
```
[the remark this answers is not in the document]
frank-emrich: Done

src/wasm.h
```diff
@@ -1956,6 +1959,31 @@ class StringSliceIter
...
+  ArenaVector<Name> handlerTags;
+  ArenaVector<Name> handlerBlocks;
+
+  ExpressionList operands;
+  Expression* cont;
+
+  ArenaVector<Type>& getSentTypes();
+
+  void finalize();
+  void finalize(Module* wasm);
```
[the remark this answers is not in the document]
frank-emrich: Done

src/wasm/wasm.cpp
frank-emrich: Done

test/lit/typed_continuations_resume.wast
```diff
@@ -0,0 +1,94 @@
+;; NOTE: Assertions have been generated by update_lit_checks.py --all-items and should not be edited.
```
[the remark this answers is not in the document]
frank-emrich: Done

test/lit/typed_continuations_resume.wast
```diff
@@ -0,0 +1,94 @@
...
+
+
```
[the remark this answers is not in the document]
frank-emrich: Done

test/lit/typed_continuations_resume.wast
frank-emrich: Done, I've just changed `$t` in this test.

frank-emrich:
Thanks for the reviews, @tlively and @aheejin. I've addressed all comments, with most of them being resolved now.

src/wasm/wasm-validator.cpp
```diff
@@ -3285,6 +3286,24 @@ void FunctionValidator::visitStringSliceIter(StringSliceIter* curr) {
...
+  // TODO implement actual type-checking
+  shouldBeTrue(
+    !getModule() || getModule()->features.hasTypedContinuations(),
+    curr,
+    "resume requires typed-continuatons [--enable-typed-continuations]");
+
+  shouldBeTrue(
+    curr->sentTypes.size() == curr->handlerBlocks.size(),
+    curr,
+    "sentTypes cache in Resume instruction has not been initialised");
```
aheejin, line 3299:
```suggestion
    "sentTypes cache in Resume instruction has not been initialized");
```
Nit: Binaryen codebase seems to use American spelling, so...

src/wasm/wasm.cpp
aheejin: The current code looks good to me as is.

aheejin (approved):
Thanks!

tlively:
Oops, I saw your spelling nit just after I merged this, @aheejin. I'll fix it.

src/ir/effects.h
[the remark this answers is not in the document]
kripken: Can a resume instruction throw an exception? If not, we can remove this. (Calls can, which is why they have it.)

PR #6085, opened 2023-11-06
Fix build failure on older Ubuntu

tlively:
Update the C++20 builder to use Ubuntu 20.04 to catch problems building with its system compiler. Also fix such a problem in wasm-fuzz-lattices.cpp.

### DIFF
--- a/src/tools/wasm-fuzz-lattices.cpp
+++ b/src/tools/wasm-fuzz-lattices.cpp
@@ -862,13 +862,14 @@ struct ReachingDefinitionsChecker {
 // Uninteresting implementation details for RandomFullLattice and RandomLattice.
 
 RandomFullLattice::Element RandomFullLattice::getBottom() const noexcept {
-  return std::visit([](const auto& l) { return ElementImpl{l.getBottom()}; },
-                    *lattice);
+  return std::visit(
+    [](const auto& l) -> Element { return ElementImpl{l.getBottom()}; },
+    *lattice);
 }
 
 RandomFullLattice::Element RandomFullLattice::getTop() const noexcept {
-  return std::visit([](const auto& l) { return ElementImpl{l.getTop()}; },
-                    *lattice);
+  return std::visit(
+    [](const auto& l) -> Element { return ElementImpl{l.getTop()}; }, *lattice);
 }
 
 // TODO: use std::remove_cvref_t from C++20 instead.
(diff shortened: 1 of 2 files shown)


tlively:
* main
  * **PR #6085** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6085" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/6085?utm_source=stack-comment).

tlively:
Huzzah! The build is fixed! Shall we keep the change to the C++20 builder here to avoid similar problems in the future?

sbc100 (commented):
What was wrong with the current code exactly?  Was it using some new feature?  Or is the old ubuntu compiler just buggy/wrong?

src/tools/wasm-fuzz-lattices.cpp
```diff
@@ -158,39 +158,47 @@ using ArrayLattice = analysis::Array<RandomLattice, 2>;
...
-                 typename Shared<RandomLattice>::Element> {};
+using FullLatticeVariant = std::variant<Bool,
+                                        UInt32,
+                                        ValType,
+                                        Inverted<RandomFullLattice>,
+                                        ArrayFullLattice,
+                                        Vector<RandomFullLattice>,
+                                        TupleFullLattice>;
+
+struct RandomFullLattice::LatticeImpl : FullLatticeVariant {};
```
kripken, line 169: Is this "subclass without adding anything" a workaround for a compiler bug, or a C++ spec requirement, or something else?

tlively: Even before this PR, we had to subclass without adding anything because we had to defer defining these nested types until their outer types (`RandomFullLattice` and `RandomLattice`) were complete. Since `using` statements cannot be forward declared, we had no choice but to forward declare `LatticeImpl` (previously `L`) as separate structs.

This PR just gives a name to the raw `std::variant<...>` superclasses as well and uses those newly-named types down below where we call `std::visit`.

tlively:
> What was wrong with the current code exactly? Was it using some new feature? Or is the old ubuntu compiler just buggy/wrong?

It seems that the implementation of `std::visit` on the older ubuntu did not Just Work when passed subclasses of `std::variant` rather than the `std::variant` itself. Stack Overflow suggests this was a gcc bug: https://stackoverflow.com/questions/51309467/using-stdvisit-on-a-class-inheriting-from-stdvariant-libstdc-vs-libc

PR #6086, opened 2023-11-07
[Parser] Parse `call` and `return_call`

tlively:
To support parsing calls, add support for parsing function indices and building
calls with IRBuilder.

### DIFF
--- a/src/wasm/wasm-ir-builder.cpp
+++ b/src/wasm/wasm-ir-builder.cpp
@@ -313,6 +313,17 @@ Result<> IRBuilder::visitBreak(Break* curr, std::optional<Index> label) {
   return Ok{};
 }
 
+Result<> IRBuilder::visitCall(Call* curr) {
+  auto numArgs = wasm.getFunction(curr->target)->getNumParams();
+  curr->operands.resize(numArgs);
+  for (size_t i = 0; i < numArgs; ++i) {
+    auto arg = pop();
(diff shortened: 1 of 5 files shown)


tlively:
* main
  * **PR #6086** <a href="https://app.graphite.dev/github/pr/WebAssembly/binaryen/6086" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/WebAssembly/binaryen/6086?utm_source=stack-comment).

src/wasm/wasm-ir-builder.cpp
```diff
@@ -546,7 +557,14 @@ Result<> IRBuilder::makeBreak(Index label) {
 
 // Result<> IRBuilder::makeSwitch() {}
 
-// Result<> IRBuilder::makeCall() {}
+Result<> IRBuilder::makeCall(Name func, bool isReturn) {
```
ashleynh, line 560: Should we add a comment documenting that makeCall is a departure from the `wasm-builder.h` `makeCall`, in that there is an assumption the Function already exists before creating the call to it.

ashleynh (approved):

src/wasm/wasm-ir-builder.cpp
tlively: Can do 👍

PR #6088, opened 2023-11-08
Move --separate-data-segments into a pass so it can be run from wasm-opt

sbc100:
Because we currently strip some data segments (i.e. EM_JS strings) during `--post-emscripten` this is too late `--separate-data-segments` always runs in `wasm-emscripten-finalize`.

Once emscripten switches over to using the pass directly we can remove the support from `wasm-emscripten-finalize`

### DIFF
--- a/src/passes/SeparateDataSegments.cpp
+++ b/src/passes/SeparateDataSegments.cpp
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2023 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+//
+// Emits the data segments to a file. The file contains data from address base
+// onwards (we must pass in base, as we can't tell it from the wasm - the
+// first segment may start after a run of zeros, but we need those zeros in
+// the file.
+//
+
+#include "pass.h"
+#include "support/file.h"
+#include "wasm-features.h"
+#include "wasm.h"
+
+namespace wasm {
+
+struct SeparateDataSegments : public Pass {
+  bool modifiesBinaryenIR() override { return false; }
+
+  void run(Module* module) override {
+    std::string outfileName =
+      getPassOptions().getArgument("separate-data-segments",
+                                   "SeparateDataSegments usage: wasm-opt "
+                                   "--separate-data-segments@FILENAME");
+    Output outfile(outfileName, Flags::Binary);
+    std::string baseStr = getPassOptions().getArgument(
+      "separate-data-segments-global-base",
+      "SeparateDataSegments usage: wasm-opt "
+      "--pass-arg=separate-data-segments-global-base@NUMBER");
+    Address base = std::stoi(baseStr);
+    size_t lastEnd = 0;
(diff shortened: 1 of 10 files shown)


test/lit/passes/separate-data-segments.wast
```diff
@@ -0,0 +1,29 @@
...
+  ;; CHECK-NEXT:  (call $foo)
+  ;; CHECK-NEXT: )
+  (func $bar
+    (call $foo)
+  )
+
+  (data (i32.const 1024) "hello world\n")
+)
+
+;; CHECK-DATA: {{^}}hello world{{$}}
```
sbc100, line 29: Is there some way ti can keep this hand-written check for the `cat %t.data | filecheck %s --check-prefix=CHECK-DATA` command above while still using `update_lit_checks.py` to update the rest of the checks?

kripken: I don't think so - each test file is either automatic or not. Might need separate files then. But @tlively might have a better idea?

src/passes/SeparateDataSegments.cpp
```diff
@@ -0,0 +1,75 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Emits the data segments to a file. The file contains data from address base
+ * onwards (we must pass in base, as we can't tell it from the wasm - the
+ * first segment may start after a run of zeros, but we need those zeros in
+ * the file).
```
kripken, line 19: Please move this to a `//` comment below. We don't merge the description into the license - we keep the license in a uniform way.

src/passes/SeparateDataSegments.cpp
```diff
@@ -0,0 +1,75 @@
...
+  void run(Module* module) override {
+    std::string outfileName =
+      getPassOptions().getArgument("separate-data-segments",
+                                   "SeparateDataSegments usage: wasm-opt "
+                                   "--separate-data-segments@FILENAME");
+    Output outfile(outfileName, Flags::Binary);
+    std::string baseStr =
+      getPassOptions().getArgument("global-base",
+                                   "SeparateDataSegments usage: wasm-opt "
+                                   "--global-base=NUMBER");
```
kripken, line 40: If this argument is only going to be used here, please prefix it, so `--separate-data-segments-global-base`. That's the convention we use in e.g. Asyncify which has several pass arguments.

test/lit/passes/separate-data-segments.wast
```diff
@@ -0,0 +1,29 @@
...
+  )
+
+  ;; CHECK:      (func $bar
+  ;; CHECK-NEXT:  (call $foo)
+  ;; CHECK-NEXT: )
+  (func $bar
+    (call $foo)
+  )
+
+  (data (i32.const 1024) "hello world\n")
```
kripken, line 26: Please move this to below the `memory` which is the usual place for this.

src/passes/SeparateDataSegments.cpp
sbc100: That seems odd.. but will do for conformance sake.

test/lit/passes/separate-data-segments.wast
tlively: No, unfortunately the updater script does not support mixing hand-written and generated checks. We could potentially add that as a feature, though. I don't think it would be too hard to add a flag telling the updater to ignore preserve checks with a particular `CHECK` prefix. OTOH, the updater script is already rather complicated.

src/passes/SeparateDataSegments.cpp
sbc100: Done.   Quite a mouthful but I guess that doesn't really matter since its all internal usage.

tlively (approved):

kripken (approved):
Yeah, the name is a little verbose as you said, but just internal. We could in theory shorten the pass name to reduce the verbosity, but I think it's good as is.

PR #6091, opened 2023-11-09
[wasm-emscripten-finalize] Remove --separate-data-segments

sbc100:
See #6088


kripken (approved):

sbc100:
I the test failure here genuine?  It looks unrelated maybe?

kripken:
Looks like that's the flake we fixed on `main` recently. Merging `main` should fix it.

PR #6106, opened 2023-11-10
[NFC] Refactor out subtyping discovery code

kripken:
This implements an idea I mentioned in the past, to extract the subtyping discovery
code out of Unsubtyping so it could be reused elsewhere. Example possible uses:
the validator could use to remove a lot of code, and also a future PR of mine will
need it. Separately from those, I think this is a nice refactoring as it makes Unsubtyping
much smaller.

This just moves the code out and adds some C++ template elbow grease as needed.

### DIFF
--- a/src/ir/subtype-exprs.h
+++ b/src/ir/subtype-exprs.h
@@ -224,12 +224,7 @@ struct SubtypingDiscoverer : public OverriddenVisitor<Parent> {
   void visitTupleMake(TupleMake* curr) {}
   void visitTupleExtract(TupleExtract* curr) {}
   void visitRefI31(RefI31* curr) {}
-  void visitI31Get(I31Get* curr) {
-
-    // XXX XXX XXX added past NFC!
-
-    self()->noteSubtype(curr->i31, Type(HeapType::i31, Nullable));
-  }
+  void visitI31Get(I31Get* curr) {}
   void visitCallRef(CallRef* curr) {
     if (!curr->target->type.isSignature()) {
       return;
--- a/src/ir/subtype-exprs.h
+++ b/src/ir/subtype-exprs.h
@@ -62,12 +62,12 @@ namespace wasm {
 // Note that noteCast(Type, Type) and noteCast(Type, Expression) never occur and
 // do not need to be implemented.
 //
-// The parent must also inherit from ControlFlowWalker (for findBreakTarget).
+// The class must also inherit from ControlFlowWalker (for findBreakTarget).
 //
 
 template<typename SubType>
 struct SubtypingDiscoverer : public OverriddenVisitor<SubType> {
-  SubType* self() { return (SubType*)this; }
+  SubType* self() { return static_cast<SubType*>(this); }
 
   void visitFunction(Function* func) {
     if (func->body) {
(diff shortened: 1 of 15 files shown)


src/ir/subtype-exprs.h
```diff
@@ -0,0 +1,339 @@
...
+//                                    a fixed type, for example, a Call operand
+//                                    must be a subtype of the signature's
+//                                    param.
+//  * noteSubType(Expression, Expression) - An expression's type must be a
+//                                          subtype of anothers, for example,
+//                                          a block and its last child.
+//
+//  * noteCast(HeapType, HeapType) - A fixed type is cast to another, for
+//  example,
+//                                   in a CallIndirect.
```
tlively, line 57: Looks like the formatting could be more consistent here.

src/ir/subtype-exprs.h
```diff
@@ -0,0 +1,339 @@
...
+//  * noteCast(Expression, Type) - An expression's type is cast to a fixed type,
+//                                 for example, in RefTest.
+//  * noteCast(Expression, Expression) - An expression's type is cast to
+//                                       another, for example, in RefCast.
```
tlively, line 61: Perhaps I'll find out below, but what is the benefit of this over just extracting the type of the cast expression and using `noteCast(Expression, Type)` in this case?

src/ir/subtype-exprs.h
```diff
@@ -0,0 +1,339 @@
...
+//
+// Note that noteCast(Type, Type) and noteCast(Type, Expression) never occur and
+// do not need to be implemented.
+//
+// The parent must also inherit from ControlFlowWalker (for findBreakTarget).
+//
+
+template<typename Parent>
```
tlively, line 69: Do we have any other cases where we take a parent as a parameter instead of a subclass as a parameter?

src/ir/subtype-exprs.h
```diff
@@ -0,0 +1,339 @@
...
+struct SubtypingDiscoverer : public OverriddenVisitor<Parent> {
+  Parent* self() { return (Parent*)this; }
```
tlively, line 71: I believe this will help the compiler give more helpful errors in case something is wrong with the setup.
```suggestion
  Parent* self() { return static_cast<Parent*>(this); }
```

tlively (commented):
This LGTM % some small comments. It would be very exciting if this let us simplify the validator.

src/ir/subtype-exprs.h
kripken: This PR tries to keep the info as "high-level" as possible, that is, not to lose information. We might want the extra information here to say something about the expression that fails to validate, if we use this in the validator some day (though, if we have the cast anyhow, we can do `cast->value`).

src/ir/subtype-exprs.h
kripken: Oh, maybe "Parent" is a confusing name here, yeah. It's the same as the usual pattern.

tlively:
I think that if we wanted to use this in the validator, we would want to note the "subtyping" for non-ref values as well.

tlively (approved):

PR #6107, opened 2023-11-10
Add missing subtyping discovery for I31Get

kripken:
This was not needed in Unsubtyping, but other users of this code will need it (it
failed a test on a future PR of mine, so it will get tested there).


tlively (approved):

kripken:
Closing as this is bundled in #6108 and I've added a few more small fixes there. I'll split them all out later.

PR #6115, opened 2023-11-13
OptimizeAddedConstants: Handle a final added constant properly

kripken:
We had an assert there that was wrong. In fact the assert is just in one of two code paths,
and an optional one: the end situation is we have an expression and a constant to add to it,
and the assert was in the case that the expression is a Const so we can do the add at
compile time (the other code path does the add at runtime). This code path is optional as
Precompute would do such compile-time addition anyhow, but it is nice to fix and leave that
path so that this pass emits fully optimal code.

### DIFF
--- a/src/passes/OptimizeInstructions.cpp
+++ b/src/passes/OptimizeInstructions.cpp
@@ -2951,8 +2951,8 @@ private:
     }
     // Add the total constant value we computed to the value remaining here.
     // Note that if the value is 32 bits then |makeFromInt64| will wrap to 32
-    // bits for us; as all the operations before us, and the add below us, are
-    // adds and subtracts then any overflow is not a problem.
+    // bits for us; as all the operations before us and the add below us are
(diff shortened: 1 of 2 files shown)


src/passes/OptimizeInstructions.cpp
```diff
@@ -2949,19 +2949,21 @@ struct OptimizeInstructions
     if (constant == 0ULL) {
       return walked; // nothing more to do
     }
+    // Add the total constant value we computed to the value remaining here.
+    // Note that if the value is 32 bits then |makeFromInt64| will wrap to 32
+    // bits for us; as all the operations before us, and the add below us, are
+    // adds and subtracts then any overflow is not a problem.
```
tlively, line 2955: comma nits for readability:

```suggestion
    // bits for us; as all the operations before us and the add below us are
    // adds and subtracts, any overflow is not a problem.
```

tlively (approved):

PR #6117, opened 2023-11-14
[Outlining] Adding more tests

ashleynh:
Checking a couple of testing TODOs off and adding more tests of the outlining pass for outlining:
- a sequence at the beginning of an existing function
- a sequence that is outlined into a function that takes no arguments
- multiple sequences from the same source function into different outlined functions

### DIFF
--- a/src/passes/Outlining.cpp
+++ b/src/passes/Outlining.cpp
@@ -246,6 +246,9 @@ struct Outlining : public Pass {
     // relative to the entire program.
     auto sequences = makeSequences(module, substrings, stringify);
     outline(module, sequences);
+    // Position the outlined functions first in the functions vector to make
+    // the outlining lit tests far more readable.
+    moveOutlinedFunctions(module, substrings.size());
   }
 
   Name addOutlinedFunction(Module* module,
@@ -253,26 +256,16 @@ struct Outlining : public Pass {
                            const std::vector<Expression*>& exprs) {
     auto startIdx = substring.StartIndices[0];
     // The outlined functions can be named anything.
-    Name outlinedFunc =
-      Names::getValidFunctionName(*module, std::string("outline$"));
+    Name func = Names::getValidFunctionName(*module, std::string("outline$"));
     // Calculate the function signature for the outlined sequence.
     StackSignature sig;
     for (uint32_t exprIdx = startIdx; exprIdx < startIdx + substring.Length;
          exprIdx++) {
       sig += StackSignature(exprs[exprIdx]);
     }
-    // Purposefully inserting the outlined functions at the beginning of the
-    // functions vector, so that outlined function assertions are positioned
-    // within the test module. This greatly improves readability of the
-    // outlining lit tests. Because of the direct manipulation of the functions
-    // vector, instead of using the addFunction() helper method, a call to
-    // updateFunctionsMap() is required.
-    module->functions.insert(
-      module->functions.begin(),
-      Builder::makeFunction(
-        outlinedFunc, Signature(sig.params, sig.results), {}));
-    module->updateFunctionsMap();
-    return outlinedFunc;
+    module->addFunction(
+      Builder::makeFunction(func, Signature(sig.params, sig.results), {}));
+    return func;
   }
 
   using Sequences =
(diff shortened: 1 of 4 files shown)


test/lit/passes/outlining.wast
```diff
@@ -66,35 +64,37 @@
...
-  (func $b
+  (func $d
     (drop (i32.const 0))
     (drop (i32.const 5))
     (drop (i32.const 1))
     (drop (i32.const 2))
   )
 )
+
+;; Tests that outlining occurs properly when the sequence is at the beginning of a function. Also tests that the outlined function has no arguments.
```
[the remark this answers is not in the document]
ashleynh: Great idea! done

test/lit/passes/outlining.wast
```diff
@@ -106,3 +106,136 @@
...
+    (drop (i32.const 0))
+    (drop (i32.add
+      (i32.const 0)
+      (i32.const 1)))
+    (drop (i32.const 1))
+    (drop (i32.const 7))
+  )
+)
+
+;; Tests multiple sequences being outlined from the same source function into different outlined functions.
```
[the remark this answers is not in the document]
ashleynh: Talked offline. No need to now that the outlined function is printing first.

test/lit/passes/outlining.wast
```diff
@@ -106,3 +106,136 @@
...
+;; CHECK:      (func $outline$
+;; CHECK-NEXT:  (drop
+;; CHECK-NEXT:   (i32.const 0)
+;; CHECK-NEXT:  )
+;; CHECK-NEXT:  (drop
+;; CHECK-NEXT:   (i32.add
+;; CHECK-NEXT:    (i32.const 0)
+;; CHECK-NEXT:    (i32.const 1)
+;; CHECK-NEXT:   )
+;; CHECK-NEXT:  )
```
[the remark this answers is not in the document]
ashleynh: done

src/passes/Outlining.cpp
```diff
@@ -261,8 +261,17 @@ struct Outlining : public Pass {
...
+    // functions vector, so that outlined function assertions are positioned
+    // within the test module. This greatly improves readability of the
+    // outlining lit tests. Because of the direct manipulation of the functions
+    // vector, instead of using the addFunction() helper method, a call to
+    // updateFunctionsMap() is required.
+    module->functions.insert(
+      module->functions.begin(),
+      Builder::makeFunction(
+        outlinedFunc, Signature(sig.params, sig.results), {}));
+    module->updateFunctionsMap();
```
tlively, line 274: It would be better to update the function map just once at the end rather than for each outlined function. How about we rename this function `createOutlinedFunction` and have it return the `std::unique_ptr<Function>` it gets from `Builder::makeFunction`. Then in the caller we could insert all the new functions into `module->functions` and update the functions map in one go.

test/lit/passes/outlining.wast
```diff
@@ -49,60 +59,152 @@
...
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (i32.add
+  ;; CHECK-NEXT:    (i32.const 0)
+  ;; CHECK-NEXT:    (i32.const 1)
+  ;; CHECK-NEXT:   )
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT:  (drop
+  ;; CHECK-NEXT:   (i32.const 1)
+  ;; CHECK-NEXT:  )
+  ;; CHECK-NEXT: )
```
tlively, line 127: I think we could simplify the outlined sequence here.

test/lit/passes/outlining.wast
```diff
@@ -49,60 +59,152 @@
...
+;; CHECK-NEXT:    (i32.const 1)
+;; CHECK-NEXT:   )
 ;; CHECK-NEXT:  )
 ;; CHECK-NEXT: )
+
+;; CHECK:      (func $g
+;; CHECK-NEXT:  (call $outline$)
+;; CHECK-NEXT:  (call $outline$_4)
+;; CHECK-NEXT: )
+  (func $g
```
tlively, line 186: Is this the most recent output of the update script? Usually it lines up the checks with the functions.

ashleynh: done

test/lit/passes/outlining.wast
ashleynh: done, also simplified the second test in the file

src/passes/Outlining.cpp
ashleynh: great idea, thanks

tlively (approved):
LGTM with outlining sequenced shortened and standard s-expression formatting.

